PR #180, opened 2018-05-30
Various fixes for JS (ES6) generator

delenius:
### PR checklist

### Description of the PR

1. Don't use property initializers

The ES6 Javascript generator used property initializers, as
described [here](https://tc39.github.io/proposal-class-public-fields/).

This is not standard ES6. In particular, create-react-app chokes on it
(see [this issue](https://github.com/swagger-api/swagger-codegen/issues/8024),
which is fixed by this commit).

This commit instead reverts back to using prototypes, as in the non-es6
JS generator.

2. Add missing pom.xml files

This allows us to run `mvn integration-test` in these folders.

3. Add missing double-quotes around enum field names.

This avoids errors due to fields that contain illegal characters.

4. Fix support for multiple inheritance

The previous implementation was broken. I added static "initialize"
methods, which supports multiple inheritance.

5. Comment out some broken tests.

There were tests for OuterBoolean etc, which are not generated
as part of index.js in the generated petstore clients. Hence, these
tests threw an error. The actual tests were already commented out
anyway. I just commented out one more line to avoid an error.

6. Remove some empty lines in generates JS files.

7. Fixed doubly generated superclass initializer calls.

8. Fixed missing single quotes in returnType in the generated "Api" classes.

The es6 and promise-es6 integration tests now pass (they didn't before).

Furthermore, the generated clients now work with create-react-app (see [this open swagger-codegen issue](https://github.com/swagger-api/swagger-codegen/issues/8024)).

Copying JS technical committee: @CodeNinjai  @frol  @cliffano


wing328:
@delenius thanks for the fixes and enhancements. What about including the the JS Petstore ES6 clients in the Travis test by adding the folders to https://github.com/OpenAPITools/openapi-generator/blob/master/pom.xml#L861 so that the CI will run the tests for us moving forward?

delenius:
@wing328 , sure! A couple of questions:

- Should I add them both under `profiles`, and also under `modules`?
- Should the `env` variable under the profile be `javascript` for all variants? I don't know what this does.

wing328:
> Should I add them both under profiles, and also under modules?

Yes

> Should the env variable under the profile be javascript for all variants? I don't know what this does.

I don't think it's used but plesae still set it to javascript for consistency.

wing328:
> BTW, item (7) in this PR may need more discussion. Previously, the model classes generated calls to superclass constructors, e.g. in Cat.js you would get

I would suggest going with your approach to move forward with this PR.

delenius:
Ok, did that, rebased against latest master, and rebuilt the samples.

wing328:
If no one has further feedback/question on this PR, I'll merge it into master over the weekend.

PR #182, opened 2018-05-30
Docker: use correct MAVEN_CONFIG

jaypea:
the variable wasn't used correctly. i always fell back to use `.` as homedir.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -14,7 +14,7 @@ pattern="@Command(name = \"$1\""
 if expr "x$1" : 'x[a-z][a-z-]*$' > /dev/null && fgrep -qe "$pattern" "$cmdsrc"/*.java || expr "$1" = 'help' > /dev/null; then
     # If ${GEN_DIR} has been mapped elsewhere from default, and that location has not been built
     if [[ ! -f "${codegen}" ]]; then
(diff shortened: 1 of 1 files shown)


jaypea:
this is a quick test in your console:
```
$ MAVEN_CONFIG=/some/dir
$ dirname MAVEN_CONFIG
.
$ dirname $MAVEN_CONFIG
/some
```

in the first case, dirname is called with string `'MAVEN_CONFIG'` 
in the second run its called with the contents of the variable `$MAVEN_CONFIG`

jimschubert:
@jaypea this is obviously a correct fix for the script as sit is written, and I'm going to merge it.

I was just wondering, what is your use case where you discovered this?

If you build the Dockerfile directly, it looks like this would result in the setting at `/opt/openapi-generator`. It seems like run-in-docker.sh sets this to `/var/maven/.m2`, then maps to your local .m2 repo. I have concerns that this change may cause host m2 repository entries to have root permissions.  Sharing your use case would help me evaluate the impact, and whether we might need to update the script(s).

jimschubert:
oh, at first glance I missed that we are in fact setting user/group in run-in-docker.sh. I'm still interested in the use case, though.

jaypea:
Though, i had to remove the userid setting in docker run because i didn't want to change the docker-entrypoint.sh to this extend. I would rather chown the mounted repository, currently its owned by root in the container.

PR #201, opened 2018-06-02
[gradle] Plugin release management

jimschubert:
This applies steps necessary for publishing to Sonatype, including
sources, javadoc, jar and signing. Also includes full POM details, per
Sonatype requirements.

In my original implementation of the plugin, I hadn't realized `maven-publish` was an experimental update to `maven`. I had falsely assumed that it applied all necessary logic in the current diff of this PR, but a key requirement (POM signing) is not available in `maven-publish`.

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

PR #202, opened 2018-06-02
Build improvements

jmini:
* Docs: set version to 3.0.0  in the Readme of `openapi-generator-maven-plugin`
* Changes "openapi-generator-gradle-plugin" 
  - Change artifactId to avoid confusion between the real gradle plugin and the maven wrapper that calls gradle.
  - Disable maven-install-plugin for the wrapper
  - Disable maven-deploy-plugin for the wrapper

### DIFF
--- a/modules/openapi-generator-gradle-plugin/pom.xml
+++ b/modules/openapi-generator-gradle-plugin/pom.xml
@@ -65,57 +65,14 @@
                     </execution>
                 </executions>
             </plugin>
-            <!-- 3) copy the gradle artifacts to '/target/gradle-artifacts'. -->
-            <plugin>
-                <artifactId>maven-resources-plugin</artifactId>
-                <version>3.1.0</version>
-                <executions>
-                    <execution>
-                        <id>copy-gradle-artifacts</id>
-                        <phase>install</phase>
-                        <goals>
-                            <goal>copy-resources</goal>
-                        </goals>
-                        <configuration>
-                            <outputDirectory>${project.basedir}/target/gradle-artifacts</outputDirectory>
-                            <resources>
-                                <resource>
-                                    <directory>${settings.localRepository}/org/openapitools/openapi-generator-gradle-plugin/${project.version}/</directory>
-                                    <filtering>false</filtering>
-                                </resource>
-                            </resources>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
-            <!-- 4) change maven-deploy-plugin. -->
+            <!-- 3) disable maven deploy. This wrapper is not needed. -->
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
                 <version>2.8.2</version>
-                <executions>
-                    <!-- deploying this wrapper is not needed -->
-                    <execution>
-                        <id>default-deploy</id>
-                        <configuration>
-                            <skip>true</skip>
-                        </configuration>
-                    </execution>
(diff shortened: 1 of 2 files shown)


jmini:
I have noticed that we have also the "wrong pom and wrong jar for `openapi-generator-gradle-plugin`" problem  on our snapshot repository (https://oss.sonatype.org/content/repositories/snapshots/) and I have tried to address this the maven way.

---

This is what I have tried with commit 4106857:

After the run of `gradle-maven-plugin` I have added two steps:

First step:
Copy the artifact from the maven local repository into an other one `target/gradle-artifacts` like this:

https://github.com/OpenAPITools/openapi-generator/blob/4106857fc9e2e2c001882bccd0441d5a8718a7eb/modules/openapi-generator-gradle-plugin/pom.xml#L68-L90

This is requested for the next step. `deploy-file` request the file to be outside of the maven local repository. Otherwise this error is displayed:
> Cannot deploy artifact from the local repository

Next step:
I have used the [deploy-file](http://maven.apache.org/plugins/maven-deploy-plugin/deploy-file-mojo.html) goal of the `maven-deploy-plugin` to upload the artifacts generated by gradle:

https://github.com/OpenAPITools/openapi-generator/blob/4106857fc9e2e2c001882bccd0441d5a8718a7eb/modules/openapi-generator-gradle-plugin/pom.xml#L104-L117

I see two problems with this approach:

* First we need to configure one distant maven repository. I have configured `sonatype-nexus-snapshots` but this means that this config is not valid when we deploy a release. The normal `deploy` mojo seems to work with the `<distributionManagement>` section (defined in the effective pom). This seems to support the fact that the release repo and the snapshot repos are not the same.
* Second issue: the credentials are missing. I am not sure how this is solved.

Log from the [travis build](https://travis-ci.org/OpenAPITools/openapi-generator/builds/387087170):
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy-file (custom-deploy) on project openapi-generator-gradle-plugin-mvn-wrapper: Failed to deploy artifacts: Could not transfer artifact org.openapitools:openapi-generator-gradle-plugin:jar:3.0.1-20180602.132316-5 from/to remote-repository (https://oss.sonatype.org/content/repositories/snapshots/): Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/org/openapitools/openapi-generator-gradle-plugin/3.0.1-SNAPSHOT/openapi-generator-gradle-plugin-3.0.1-20180602.132316-5.jar. Return code is: 401, ReasonPhrase: Unauthorized. -> [Help 1]
```

I will revert the two steps described here in order to have a PR that can be merged. We need to continue to investigate how we can upload the graddle jars to the distant maven repository. We can use either maven or gradle. We need something that works for release and for snapshots. Related PR #201.
 
cc @jimschubert

jmini:
In my opinion this is ready to be merged. CI is ok (CircleCI is unrelated to this change), description was edited to represent the content of this PR.

This is an improvement in regard to what we have now.
The deploy part (discussed in the middle of this PR) can be fixed in a follow-up PR or as part of #201.

jimschubert:
@jmini to release the gradle plugin, we'll need to make sure we have the configuration values mentioned in #201 in place and then we just need to call `./gradlew uploadArchives` at the end of the CI script. The `build.gradle` handles everything for snapshot vs release maintenance (including signing of all artifacts and pom and publishing of jar/sources/javadoc artifacts).  The only things we really need hooked into the maven lifecycle are clean and build/test.

PR #203, opened 2018-06-02
Update typescript-angularjs to generate top level enums

achew22:
Enums previously were generated as empty interfaces. They are now generated
as proper first class enum objects in typescript.

### PR checklist

cc: @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny

### Description of the PR

Adds support for generating top level enums in TypeScript.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-angularjs/model.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angularjs/model.mustache
@@ -8,6 +8,16 @@ import * as models from './models';
  * {{{description}}}
  */
 {{/description}}
+{{#isEnum}}
+export enum {{{classname}}} {
+{{#allowableValues}}
+{{#enumVars}}
+    {{{name}}} = {{{value}}}{{^-last}},{{/-last}}
+{{/enumVars}}
+{{/allowableValues}}
+}
+{{/isEnum}}
+{{^isEnum}}
 export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
 {{#vars}}
 {{#description}}
@@ -26,7 +36,7 @@ export namespace {{classname}} {
     export enum {{enumName}} {
     {{#allowableValues}}
     {{#enumVars}}
-        {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
+        {{{name}}} = {{{value}}}{{^-last}},{{/-last}}
     {{/enumVars}}
     {{/allowableValues}}
     }
@@ -34,5 +44,6 @@ export namespace {{classname}} {
 {{/vars}}
 }
 {{/hasEnums}}
+{{/isEnum}}
 {{/model}}
 {{/models}}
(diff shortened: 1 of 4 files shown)


kenisteward:
Thanks for your contribution!

Regarding this PR, did you use the typescript angular mustache as an example? We have solved the problems for enums there and it was just never rolled over to this one.

kenisteward:
Also please regenerate the samples :)

wing328:
I also want to add that we're working on consolidating TS generators into one with the library option similar to what we've done for the Java generator.

achew22:
PTAL

kenisteward:
It is as I expected. This doesn't work if users are using typescript 2.3 or under. 

Only typescript 2.4 or higher supports string enums 

Please take a look at what we do for the typescript-angular one if you want to continue this PR.

achew22:
Out of curiosity, what is your threshold for using features? 2.3 is 6 versions back dating to April 2017. When do you consider a new language feature ready to use?

kenisteward:
It's more of the fact that this library supports all the way down to
angular 2 which base angular 2 only supported typescript 2.2 if I remember
correctly.

Couple that with the fact that we don't specifically allow users to choose
their own ts version and it could cause issues.

The best thing would be when we have this as base changes to ts instead
then we can say

1. Choose ts version in options
2. Choose lib
3 choose lib version

Then if someone chooses something that is incompatible from their lib
choose that's on them.

achew22:
> It's more of the fact that this library supports all the way down to
> angular 2 which base angular 2 only supported typescript 2.2 if I remember
> correctly.

This is the library component for AngularJS. Do you mean this supports all the way back to Angular 1.2?

> [...] if someone chooses something that is incompatible from their lib choose that's on them.

Seems like a reasonable path forward. You also have the escape hatch of the `-t` option. I've vendored these into my project now so I'm using a local copy. I just wanted to push my changes upstream so others didn't have to figure it out, but it does serve as a template for how you could keep the library alive using the latest and greatest features, but still provide that backup plan for people who can't migrate quite yet.

kenisteward:
Oh woops haha. I don't know where my brain went. Since that's the case you
could do it but force usage of typescript 2.4 or suggest it by setting a
peer dependency on ts 2.4

kenisteward:
I just rechecked our the mustache templates and it looks like the angularjs ones are not as well maintained and don't have a "create npm package" generation.

Because of this I'd say we can go a few routes. 

1. We merge as is since it should work fine that way.
2. You do what we did with the typescript-angular mustache which gives us confidence it will work with any typescript ( 2+ at least I think)
3. You create the package.json generation and add the package.json peer dependency 

I think I'd prefer 2 as it makes the most sense to stay compatible across all but would like to have opinions from another member.

 @TiFu @taxpon @sebastianhaas @Vrolijkx @macjohnny
@wing328

wing328:
Closing this as there's no update.

Issue #186, opened 2018-05-30
Output of the generator is not working in the cli


JFCote:
##### Description

When I try to put the output of the cli in a text file when doing a generation, it doesn't work. If I do a `validate`, it works.

##### openapi-generator version

Latest master

##### OpenAPI declaration file content or url

Any

##### Command line used for generation

```
java -jar openapi-generator-cli.jar generate -i swagger.yaml --generator-name
java-play-framework -o generatedServer -DhideGenerationTimestamp=true > lol.txt
```

##### Steps to reproduce

Just run the above command line and notice the lol.txt is 0 byte. If you do a `validate `instead, the file is modified as expected.

##### Related issues/PRs

None that I know of.

##### Suggest a fix/enhancement

I really don't know for the moment.

jimschubert:
I've noticed this as well. The issue is that some streams used by the generator project don't output to the same file descriptor as stdout/stderr for the CLI. I didn't have a chance to investigate, unfortunately. 

The reason `validate` works is because it to uses the system stdout and stderr streams. I believe with `validate` if you set log levels to debug, the logs from Swagger Parser are also not redirected.

The fix should be a matter of changing root level logging to a ConsoleLogger in the CLI module. There are some `println` calls in the generator project which also don't seem to be redirected, so it seems to me that `System.setOut` may be called somewhere.

jimschubert:
I looked into this. The issue was that we are using slf4j-simple in the core generator project. CLI uses this as well, and that implementation only writes to STDERR.

You'd need to do:

```
java -jar openapi-generator-cli.jar generate -i swagger.yaml --generator-name \
    java-play-framework -o generatedServer -DhideGenerationTimestamp=true 2>&1 > lol.txt
```

It's confusing that non-error messages would be written to stdout. I've opened #207 to have intuitive usage of these output streams.

jimschubert:
@JFCote #207 has been merged, and should fix this issue. Can you verify?

JFCote:
@jimschubert Will validate this today and let you know

JFCote:
@jimschubert Confirmed fixed!

PR #207, opened 2018-06-03
[cli] Write to stdout/stderr, allow redirection

jimschubert:
Previously, slf4j-simple from generator core was being used. This writes
to only a single stream (STDERR) and is confusing from a CLI tooling
perspective.

This consumes logback in CLI, and excludes core's slf4j-simple
dependency. This allows us to define multiple appenders, one for STDOUT
and one for STDERR.

WARN messages and lower are written to STDOUT. ERROR is written to
STDERR.

See #186

### PR checklist


wing328:
@jimschubert thanks for the PR. I'll review shortly.

jimschubert:
@wing328 I'll look at the default levels. Previously, they were whatever the simple plugin defaulted to, not what was in logback.xml (which isn't really in use in the core package).

jimschubert:
@wing328 I've fixed this. The previous root logger was `INFO`, which wasn't ideal. The `ch.qos.logback.classic.*` logs occurred because we had `layout` rather than `encoder`. Layout is deprecated (https://logback.qos.ch/codes.html#layoutInsteadOfEncoder) and caused those logs.

jmini:
Does it make sense to have `slf4j-simple` as `compile` scope dependency in the `org.openapitools:openapi-generator` ?

In my opinion the core should only depend from the slf4j api and let consumers (cli, other...) define the implementation they want to use (`logback`, `slf4j-simple`, ...).

jimschubert:
Should I just remove it? We can't run core on it's own, and without it we would have the noop logger.

I also think the logback.xml in the core project can be removed. It's a little presumptuous that the consumer will use logback, and if they're creating a fat jar it most likely needs to be removed anyway.

jmini:
I did not look in details into it, I have the feeling that the unit tests from the core might need to have a logger. We might just need to change the scope of `slf4j-simple` to `test`.

jimschubert:
@jmini When you say tests "might need to have a logger" do you mean that some tests may be instantiating a logger and outputting logs rather than doing a println? I think we can just change those to println rather than pulling in an additional dependency. After all, logging isn't really a common thing to do in tests. slf4j doesn't need a binding defined in a module (it defaults to `NOPLogger` there no-operation logger if no binding is found on the classpath). This could improve build times a little if we remove all unnecessary dependencies.

I think later this evening, I'll remove the extra dependency and check tests for use of loggers.

jimschubert:
@jmini I've removed slf4j-simple from the core project completely.  There was a single test which instantiated (but did not use) a logger.

This removal results in a single warning in log output as I'd expect:

```
Running TestSuite
objc[28394]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_151.jdk/Contents/Home/jre/bin/java (0x1060f24c0) and /Library/Java/JavaVirtualMachines/jdk1.8.0_151.jdk/Contents/Home/jre/lib/libinstrument.dylib (0x10b1c24e0). One of the two will be used. Which one is undefined.
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Tests run: 374, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.911 sec - in TestSuite
```

This is acceptable as it doesn't pull an unnecessary artifact, and doesn't open us to the potential of having conflicting logger implementations for additional tools moving forward. I've left the exclusion of slf4j-simple in the dependency declaration in openapi-generator-cli so we'd have record of this exclusion in the master branch.

modules/openapi-generator-cli/pom.xml
```diff
@@ -78,6 +77,12 @@
             <groupId>org.openapitools</groupId>
             <artifactId>openapi-generator</artifactId>
             <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
```
jmini: @jimschubert:
> I've left the exclusion of slf4j-simple in the dependency declaration in openapi-generator-cli so we'd have record of this exclusion in the master branch.

For me this is similar to unnecessary or commented code, it will be very difficult to understand from the outside or for us in few weeks or months. I would remove those lines.

jmini:
> I've removed slf4j-simple from the core project completely. There was a single test which instantiated (but did not use) a logger.

I consider this change as not really user friendly.

Personally, when I run a unit test, I do not use the log output to debug or understand what is happening. But I have the feeling that other developers in this project do use logs.

The logger is used when a unit test is executed, not inside `src/test/java` but inside `src/main/java`.
I would have let a `test` scope for `slf4j-simple` because this way:

* you do not have a logger implementation when you reuse the `openapi-generator` core library in an other projects (I did this yesterday and I excluded `slf4j-simple` because we are using logback) 
* you have one when you run the tests (inside an IDE or using maven).

But we can also do it your way: remove it completely and restore it when developers complains about not seeing the logs and some SLF4J warnings when they run unit tests.

modules/openapi-generator-cli/pom.xml
```diff
@@ -91,9 +96,9 @@
             <version>2.3.3</version>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
-            <version>${slf4j-version}</version>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.0.13</version>
```
jmini: Is there a reason why you have picked version `1.0.13` for `logback-classic`?

It seems to me that this is an implementation for `org.slf4j:slf4j-api:1.7.5` maybe it is compatible with the version `1.7.12` that we have defined (I guess also without any good reason)

jimschubert: I used the version documented as the native implementation on slf4j: https://www.slf4j.org/manual.html

jmini: This page seems outdated to me... See: https://logback.qos.ch/download.html

Or on maven central:
http://mvnrepository.com/artifact/ch.qos.logback/logback-classic

Do we want to update to:
`logback-classic` version `1.2.3` and `slf4j` to it corresponding version `1.7.25`. This way we will be in sync with `swagger-core`.

jimschubert:
@jmini if you run tests on this branch, you should still see a lot of logs. It seems that most of what people would use during testing is output using `println`. This PR doesn't clean those up, but we should do that at some point.

My assumptions with testing are that people are unit testing, and are therefore focusing on a small amount of code. If things are so complex that a contributor has to rely on log messages to understand the code flow during that test, I think it's on us to reduce that complexity rather than focus on logging as a use case for testing.

Enabling the simple logger binding for the test scope would be hiding that issue. I'm not against it, just very cautious about the reasoning.

modules/openapi-generator-cli/pom.xml
jimschubert: If we remove this, it can result in a transient dependency pulling in the simple binding, and a warning.

modules/openapi-generator-cli/pom.xml
jimschubert: @jmini this won't necessarily work. SLF4J's site is current.

Let me explain how slf4j works, as there seems to be some confusion.

slf4j provides an abstraction atop common logging frameworks. It does this via bindings that just need to exist in your classpath. It natively includes bindings to logback. This is why you don't have to add a binding jar, only the logback implementation. The logback usage is locked to logback 1.0.13 (see https://github.com/qos-ch/slf4j/blob/master/pom.xml) because it uses interfaces specific to that version. If you upgrade to a newer version of logback, you'd have to check if it is ABI compatible with 1.0.13. If it isn't, you can have exceptions thrown when code is JIT'd. That's obviously not ideal. It *might* just work with some luck, but I see no reason to chance it.

Similarly, if we don't exclude the simple logger binding explicitly, it opens us to the possibility of pulling this in as a transitive dependency. This simple logging package is the only one I've found that writes to stderr only, so allowing that to be included would result in a regression in this bug.

jmini: Thank you a lot for the explanations... `1.0.13` seems to be the version that we need to use.

jmini:
About logger implementation in tests, I am OK with the fact that we do not provide one.

Personally I do not need it. I just consider that having one is more user friendly than having this when a unit test is executed:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

As @wing328 told, this is something we can document:
> previously i put down https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-debug-openapi-generator. given that the community has grown a lot in size, we can create a new page/section on how to debug openapi-generator. how to setup dev environment, how to test locally , etc

wing328:
I did anther test with ./bin/ruby-petstore.sh and the result looks good.

If no more feedback/question, I'll merge this PR tomorrow (Thur)

Issue #197, opened 2018-06-01
[JAVA] oneOf implementation: wrong discriminator values and discriminator field


jmzc:
##### Description

When JSON.java code is generated for an oneOf property, it is set wrong discriminator values,
and discriminator field is set empty 

##### openapi-generator version

openapi-generator-cli-3.0.0   (SNAPSHOT 20180531.120113-109) 

##### OpenAPI declaration file content or url

```
Person:
      type: object
      properties:
         genre:
            oneOf:
               - $ref: '#/components/schemas/Male'
               - $ref: '#/components/schemas/Female'
            discriminator:
               propertyName: genretype
               mapping:
                  M: '#/components/schemas/Male'
                  F: '#/components/schemas/Female'

Genre:
      type: object
      required:
        - genretype
      properties:
         genretype:
           type: string 
           enum:
            - M
            - F

Male:
     allOf:
      - $ref: '#/components/schemas/Genre'
      - type: object
    
Female:
     allOf:
      - $ref: '#/components/schemas/Genre'
      - type: object
```

##### Command line used for generation

```
java -jar openapi-generator-cli.jar generate \
-i openapi-3.0.yaml \
-l java \
-c java-config.json \
-o ./output

```

```
> cat java-config.json
{
  "modelPackage" : "com.test.ws.client.model",
  "apiPackage"   : "com.test.ws.client.api",
  "groupId" : "com.test.ws",
  "invokerPackage" : "com.test.ws.client",
  "artifactId"  : "testws-client",
  "artifact-version": "0.0.1-SNAPSHOT",
  "dateLibrary" : "threetenbp"
}
```

##### Steps to reproduce

After generation according step before , JSON.java class includes the next code 

```
GsonFireBuilder fireBuilder = new GsonFireBuilder()
          .registerTypeSelector(Genre.class, new TypeSelector() {
            @Override
            public Class getClassForElement(JsonElement readElement) {
                Map classByDiscriminatorValue = new HashMap();
                classByDiscriminatorValue.put("Male".toUpperCase(), Male.class);
                classByDiscriminatorValue.put("Female".toUpperCase(),Female.class);
                classByDiscriminatorValue.put("Genre".toUpperCase(), Genre.class);
                return getClassByDiscriminator(
                                           classByDiscriminatorValue,
                                           getDiscriminatorValue(readElement, ""));
            }
          })
```

##### Related issues/PRs

Improve handling of oneOf #15

##### Suggest a fix/enhancement

[...]

PR #209, opened 2018-06-03
Delete unused methods in DefaultCodegen

ackintosh:
### PR checklist

### Description of the PR

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1535,36 +1535,6 @@ public class DefaultCodegen implements CodegenConfig {
         return m;
     }
 
-    /**
-     * Recursively look for a discriminator in the interface tree
-     *
-     * @param schema         composed schema
-     * @param allDefinitions all schema defintion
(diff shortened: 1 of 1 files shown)


jmini:
Thank you for this PR, it makes me aware of dead code.

---

Before removing those method, did we make an analysis about where they were used in swagger-generator v2?

For example #197 is about `oneOf`, maybe the method `isDiscriminatorInInterfaceTree(..)` could be useful (this is just a guess, I did not do an analysis yet).

ackintosh:
Sounds like a good idea.

But I think that it is better to delete the codes that is not currently in use as it is difficult to predict whether or not to use that codes. (My point is making the codes in this project more simply and keep.)

Even if we need the deleted code, it is easy to copy and use it.

What do you think?

jmini:
> But I think that it is better to delete the codes that is not currently in use as it is difficult to predict whether or not to use that codes. (My point is making the codes in this project more simply and keep.)

dead/unused/commented code is not a good thing... I totally agree with you.

wing328:
@ackintosh thanks for the PR to clean up unused methods.

@jmini very good point about the need to understand how these methods were used before and we can always revisit these methods when we've time by checking out a particular commit.

PR #213, opened 2018-06-03
[cli] Completions command for suggestions

jimschubert:
This takes airlift's 'suggest' command and reuses it as a different
command name, 'completion'. This gives us in-built CLI completions which
are useful in the repo-level container's docker-entrypoint.sh. This
previously parsed Java files for conventional usage of Command
annotations, which is potentially buggy. The new implementation relies
only on CLI to provide command completion suggestions.

As part of this, we can prepare for bash completion scripts which can be
added to our homebrew formula.  The new completion command will also
complete on command options, for example:

```
cli completion generate
```

This will provide all short and long form switches available to the
generate command.

Example (commands available globally):

```
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar completion
list
generate
meta
langs
help
config-help
validate
version
completion
```

Example (switches available to `generate` command):

```
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar completion generate
-c
--config
--invoker-package
--git-repo-id
--additional-properties
-t
--template-dir
--import-mappings
-v
--verbose
--model-name-suffix
--library
-s
--skip-overwrite
--artifact-id
--api-package
-D
--ignore-file-override
--model-package
--instantiation-types
--model-name-prefix
--git-user-id
-o
--output
--group-id
-l
--lang
-g
--generator-name
--language-specific-primitives
--artifact-version
--reserved-words-mappings
-a
--auth
-i
--input-spec
--release-note
--remove-operation-id-prefix
--type-mappings
--http-user-agent
```

Example (switches available to `meta` command):

```
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar completion meta
-n
--name
-p
--package
-o
--output
```

This also moves the exit code of 1 for executing the CLI with no args to after the `run` call, allowing CLI to execute the default command before exiting as an error.

/cc @jmini @wing328 (I don't know who is considered technical committee for CLI, maven plugin, etc)

### PR checklist

- [ ] Filed the PR against the correct branch: Default: `master`.

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2010 the original author or authors.
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2010 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * NOTICE: File originally taken from:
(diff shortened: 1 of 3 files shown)


modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (C) 2010 the original author or authors.
```
wing328: Shall we update the above with the following instead?
```
 * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
 * Copyright 2018 SmartBear Software
```

wing328:
@jimschubert the enhancement looks awesome. I'll give it a try tomorrow and let you know if I've any feedback.

modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/CompletionCommand.java
jimschubert: @wing328 we can't. Per comment below the header, I took this from the framework's suggest command and renamed it with a description. It's unfortunate that they have such a generic license text, because it makes it look like something that needs to be updated.

jimschubert: Oh, or do you think we should add our line as a second line in the header?

jimschubert: @wing328 I've added
```
* Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
```
Below the line you've commented on, as this licensed file is originally from airlift/airline, not SmartBear.

wing328: Looks good to me

Issue #282, opened 2018-06-11
[typescript-angular] service configuration/basePath modification after construction


macjohnny:
##### Description

See original issue here: https://github.com/swagger-api/swagger-codegen/issues/7949

##### openapi-generator version

3.0.0

macjohnny:
@chrismellard @akehir would you like to implement the PR https://github.com/swagger-api/swagger-codegen/pull/8305 here at OpenAPI Generator (a community fork of swagger codegen)?

PR #289, opened 2018-06-12
[typescript][angular] basePath through injected config

akehir:
### PR checklist

### Description of the PR

@macjohnny : This resolves https://github.com/OpenAPITools/openapi-generator/issues/282 , implementing my swagger-codegen pull request on the fork.

I notice, that there are a few tests missing, notably the ones I've previously added for angular 6. 

CI is here: https://travis-ci.org/akehir/openapi-generator
The original PR here: https://github.com/swagger-api/swagger-codegen/pull/8305

### Original description of the PR
Instead of using the this.basepath, we use this.configuration.basepath - allowing us to inject a different configuration (after loading a configuration file over http during APP_INITIALIZER).
We've been using this setup at a project for a while and it has been working well for us (different api url's in different environments, provided by the webserver on runtime).

This would close the problems discussed here: #3221 (comment) and #7949.

This the distinction to the existing PR #7965 is, that it is not required to generate / handle an observable, but rather just by changing the injected configuration, another basePath would be used. Do you think this would solve your concerns @chrismellard?

Since the rest of the swagger api client is always referring to the this.configuration.{...}, I believe this approach might be more consistent with the general handling of the configuration service.

To keep compatibility with implementations that don't inject a configuration, but just the base path, I set the base path on the internal configuration object to the injected basepath (when they don't exist).


macjohnny:
@akehir thanks a lot for your PR!

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -67,6 +67,8 @@ export class {{classname}} {
         if (configuration) {
```
macjohnny: wouldn't you want to set 
```
        if (configuration) {
            this.configuration = configuration;
            this.basePath = basePath || configuration.basePath || this.basePath;
        }
        this.configuration.basePath = this.basePath;
```
in order to account for the situation where you pass a `configuration` object that does not contain a base path plus a `basePath` parameter, since the `basePath` property of `Configuration` is optional, see
https://github.com/openapitools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-angular/configuration.mustache#L15 
?

macjohnny: I would even mark the `basePath` property of the api service as deprecated, since only `this.configuration.basePath` should be used. what do you think about that?

macjohnny:
cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01)

macjohnny:
@akehir which tests are missing? I think we should add them if they went missing...

akehir:
@macjohnny Although, the angular6 tests are not tested by the CI as far as I can see (there's no pom.xml / the tests & all the angular modules / builds except 4 & 4.3 are commented out).

Therefore, maybe the typescript test cases would need to be overhauled in general =/

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
macjohnny: looks good now

macjohnny:
> As for the tests, they might have been removed intentionally (especially, the angular 5 test, which was previously there). However, the scripts for generating the angular 6 tests were missing from the unix .bin/ and only existed in the windows .bat - which was the main reason for my confusion.

> I've created the unix scripts for generating the angular 6 test cases, and updated those with my newly generated codegen.

> Therefore, maybe the typescript test cases would need to be overhauled in general =/

@akehir @wing328 I think we should consider running those tests in CI, what do you think?

wing328:
@macjohnny totally agreed running more tests to have better coverage on the auto-generated code.

macjohnny:
@akehir would you be willing to give it a try and include your test case in the CI configuration?

wing328:
A good starting point is https://github.com/OpenAPITools/openapi-generator/blob/master/pom.xml#L921

You will need to add a pom.xml similar to https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-angular-v4.3/npm/pom.xml

wing328:
You may also reference the Petstore integration tests for TS Fetch client: https://github.com/OpenAPITools/openapi-generator/tree/master/samples/client/petstore/typescript-fetch/tests/default

(Typescript Angular petstore samples do not have integration tests at the moment and we would welcome contribution to add those)

Let us know if you need any help. Thanks again for your contribution.

akehir:
I can definitely look into writing these tests. However I'm quite busy at the moment and can't make any promises as to when I will be able to finish.

Creating the pom & including the tasks is the smaller issue than creating the end to end test - especially since on my system I had issues running the fetch petstore client.

```
[INFO] 
[INFO] Building TS Fetch Petstore Test Client 1.0-SNAPSHOT
[INFO] 
[INFO] --- maven-dependency-plugin:2.8:copy-dependencies (default) @ TypeScriptFetchPestoreClientTests ---
[INFO] 
[INFO] --- exec-maven-plugin:1.2.1:exec (npm-install) @ TypeScriptFetchPestoreClientTests ---
npm ERR! path /app/cloned/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/@swagger/typescript-fetch-petstore-6b055bd4/node_modules/encoding
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall rename
npm ERR! enoent ENOENT: no such file or directory, rename '/app/cloned/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/@swagger/typescript-fetch-petstore-6b055bd4/node_modules/encoding' -> '/app/cloned/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/encoding-93df19ef'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/akehir/.npm/_logs/2018-06-14T14_10_37_216Z-debug.log
[INFO]
```

akehir:
@macjohnny @wing328 @tomvangreen

I have created test cases for Angular 6 (provided in root), and added them to the samples maven profile.
During the creation, I had some issues, which is why the test structure is not quite as in the fetch client

If the api client is generated as a 'local' npm module (```/builds/with-npm```), similar to the fetch example, it is possible to import it as a npm linked dependency. However, ```npm link``` will create a symlink. Trying to use the service by simply importing the .ts file will lead to errors, as the ts compiler can't find the angular dependencies. Installing angular in the client directory will not help, because then the injected classes (httpclient) will be different from the ones in the main angular test app.

Due to that, I moved the node_modules / angular cli project 2 levels up, and just create the sample app under ```/tests/default``` (same as with the fetch client). If further tests / projects are added, they can be managed as multiple projects under angular-cli.

One open question is also, if it makes sense to instanciate a complete app component, under which to run the tests. I've created a very simple sample (basically just adapted the angular-cli standard project, to show how that could be done. However, for most tests, located under /tests/ folder, I've just used the TestBed/ Testingmodule without actually creating a component. There, the generated api clients can be injected and tested.

macjohnny:
@akehir @wing328 any idea which steps are still missing to let these test cases automatically run with circleCI?

akehir:
@macjohnny : CircleCI seems to be failing due to a scala test (sbt being the scala build tool). The version 0.13.xx is quite old, though.

The travis build is still running for me ( https://travis-ci.org/akehir/openapi-generator ), I'm waiting on that to see what's up.

macjohnny:
@akehir awesome! so it seems the tests are being run on CircleCI, that's great!

akehir:
@macjohnny : On Travis, the tests failed due to chrome not being available. I switched to PhantomJS for the maven builds / tests. I hope that will work now, we'll see in 30 Minutes ;-)

macjohnny:
@akehir Why did you change it to PhantomJS? PhantomJS is discontinued, see [http://phantomjs.org/](http://phantomjs.org/).
I would strongly opt for Chrome. Maybe the Image just needs to be adapted to have chrome installed? There are CircleCI images with chrome

akehir:
@macjohnny : I'm aware that PhantomJS is deprecated; however you'd probably need to install chrome directly, in such a way: https://hackernoon.com/running-karma-tests-with-headless-chrome-inside-docker-ae4aceb06ed3

I was not sure that changing the docker images just for these test cases is in the scope of what I'm trying to achieve.

For circleci, this would mean that chrome should be started directly in the virtual machine, not the docker.
The docker images / commands would need to be adapted.

And most notably, my test's wouldn't be included in circleci and shippable at the moment - both environments use different pom.xml (```CI/pom.xml.shippable``` and ```CI/pom.xml.circleci```). 

Furthermore, circleci 1.0 which we're currently using is also being deprecated, and we'd need to start using a new circleci 2.0 config - see here: https://circleci.com/docs/2.0/migration/

Therefore, to keep the scope of my changes minimal, PhantomJS seemed best. If @wing328 agrees with you on ChromeHeadless, I can try to update the pom files (to include my tests), and to update the images / commands to install chrome headless on the images.

akehir:
@macjohnny : Your example is quite different - mainly, your complete tests run within docker. In the circleci here, a docker image is first built on the ci vm, and then started.

akehir:
@macjohnny : I've tried to enable the tests with chrome headless for circleci and travis. Let's see what happens ;-)

akehir:
@macjohnny : CircleCI seems to work now with the test (on the machine itself).

macjohnny:
Awesome!

macjohnny:
@akehir the angular-tests seem to be executed successfully.

CircleCI seems to fail due to
`[error] (*:update) sbt.ResolveException: download failed: org.scala-lang#scala-library;2.11.11!scala-library.jar`
@wing328 could you please re-trigger CircleCI?

wing328:
Looks like a rebase (on master) has not been done correctly resulting in a lot of commits incorrectly added to this PR/branch.

I would suggest opening a new PR instead (by cherry-picking the commits)

macjohnny:
the circleCI build fails again:
`[ERROR] Plugin org.apache.maven.plugins:maven-javadoc-plugin:2.10.4 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-javadoc-plugin:jar:2.10.4: Could not transfer artifact org.apache.maven.plugins:maven-javadoc-plugin:pom:2.10.4 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]`

Issue #191, opened 2018-05-31
String/Boolean map using additionalProperties is not generated


JFCote:
##### Description

I have a simple object declared like this:

```
StringBooleanMap:
      additionalProperties:
          type: boolean
```

It was used in another model:

```
  GetFrontEndPermissionsResponse:
    type: object
    required:
      - meta
    properties:
      meta:
        $ref: "#/definitions/ResponseMeta"
      data:
        $ref: "#/definitions/StringBooleanMap"
```

With swagger-codegen, it was simply generated and used in the models needing it. It was looking like that when used in a model:

```
@JsonProperty("data")
  private StringBooleanMap data = null;
```

With openapi-generator, it is not generated at all. Instead, it simply generate a "map" with no type.

```
@JsonProperty("data")
  private Map data = null;
```

This is a very bad regression since we cannot validate that the map really is a String/Boolean, we lose type safety.

##### openapi-generator version

Latest master

##### Command line used for generation

`java -jar openapi-generator-cli.jar generate -i swagger.yaml --generator-name java-play-framework -o generatedServer -DhideGenerationTimestamp=true`

##### Steps to reproduce

Just run the above command line and make sure you have and object using the `StringBooleanMap` object in one of its property.

##### Related issues/PRs

I think it is related to https://github.com/OpenAPITools/openapi-generator/issues/185

##### Suggest a fix/enhancement

I think if we don't want to lose all the current swagger-codegen users that would want to switch, we need to generate stuff that was previously generated and it just make sense. If I create a model and that this model is referenced somewhere else, we should generate it "as is". Converting it to "basic type" is a very dangerous assumption that do not follow the spec in my opinion. And as I said earlier, this break type safety when dealing with collection (Map in this case)

jmini:
A lot has changed in the OpenAPI instance (model classes defined in swagger-core) we get when we parse a spec. This is why we needed to change stuff in the code that interprets the model and prepare the objects that will be used in templates.

Reporting cases like this one is exactly the way we can improve the quality of OpenAPI-Generator. It has a simple minimal example. Some information to understand the context. This is perfect. Thank you a lot!

I hope we will get those issue fixed for `3.0.1` or `3.0.2` in order for you to be able to migrate from swagger-codegen v2 to OpenAPI-Generator.

JFCote:
@jmini Thanks. Will keep a look at these issue and will gladly test the PR when available.

wing328:
[...]

PR #301, opened 2018-06-13
Fix getter methods for Java model with additionalProperties

stianlik:
See [merged PR for swagger-codegen](https://github.com/swagger-api/swagger-codegen/pull/8245)

### DIFF
--- a/samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
+++ b/samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
@@ -64,7 +64,8 @@ public class JSON {
             }
           })
         ;
-        return fireBuilder.createGsonBuilder();
+        GsonBuilder builder = fireBuilder.createGsonBuilder();
+        return builder;
     }
 
     private static String getDiscriminatorValue(JsonElement readElement, String discriminatorField) {
--- a/samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
+++ b/samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
@@ -64,8 +64,7 @@ public class JSON {
             }
           })
         ;
-        GsonBuilder builder = fireBuilder.createGsonBuilder();
-        return builder;
+        return fireBuilder.createGsonBuilder();
     }
 
     private static String getDiscriminatorValue(JsonElement readElement, String discriminatorField) {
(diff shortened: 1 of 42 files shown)


wing328:
cc @bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) for review

samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
```diff
@@ -64,7 +64,8 @@ public Class getClassForElement(JsonElement readElement) {
             }
           })
         ;
-        return fireBuilder.createGsonBuilder();
+        GsonBuilder builder = fireBuilder.createGsonBuilder();
```
JFCote: why this change?

samples/client/petstore/java/rest-assured/src/main/java/org/openapitools/client/JSON.java
```diff
@@ -64,7 +64,8 @@ public Class getClassForElement(JsonElement readElement) {
             }
           })
         ;
-        return fireBuilder.createGsonBuilder();
+        GsonBuilder builder = fireBuilder.createGsonBuilder();
```
JFCote: why?

jmini:
> Shippable — Run 770 status is FAILED. 

It seems that other java clients under `samples/` are affected by this change.

* run `mvn clean verify` locally
* run `bin/java-petstore-all.sh`
* commit changes to this PR.

(instead of `bin/java-petstore-all.sh` you can also run `bin/ensure-up-to-date` -- the same script than the one executed by Shippable CI. But this will take longer to run).

samples/client/petstore/java/okhttp-gson/src/main/java/org/openapitools/client/JSON.java
stianlik: This is a change related to one of my other PRs, not sure why it was included here, removed after regenerating samples. Sorry about that.

stianlik:
It looks like this issue has been resolved differently in openapi-generator. When `additionalProperties` is defined, the java client does not implement a domain class, but returns a generic `Map` instead. If that is the case, this PR is useless.

@jmini @JFCote Can you confirm that the expected behavior is as described above?

JFCote:
I have create a bug for this here:
https://github.com/OpenAPITools/openapi-generator/issues/191

stianlik:
I'm closing this PR as it's a bugfix for an obsolete solution to #191. Assuming #191 is solved properly, it is no need for an additional bugfix.

PR #305, opened 2018-06-13
CI: prevent ELM to run on Shippable

jmini:
Reported by @wing328 on Gitter:

> I wonder if you can take a look at https://app.shippable.com/github/OpenAPITools/openapi-generator/runs/772/1/console
the error is related to elm
but there's no elm in `CI/pom.xml.shippable`...

Error:

```
Compiling src/Main.elm
./elm-compile-test: line 7: elm: command not found
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (bundle-test) on project ElmClientTests: Command execution failed.: Process exited with an error: 127 (Exit value: 127) -> [Help 1]
```

This is probably a side effect of the change I have made in #280 to solve this error:

> [ERROR] [FATAL] Non-resolvable parent POM for org.openapitools:openapi-generator-cli:[unknown-version]: Could not find artifact org.openapitools:openapi-generator-project:pom:3.0.1 in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots) and 'parent.relativePath' points at wrong local POM @ line 3, column 11

Change made in #280:

I needed to call the main POM: `/pom.xml` from `CI/pom.xml.shippable` in order to build the missing module.

It probably introduces the observed side-effect in https://app.shippable.com/github/OpenAPITools/openapi-generator/runs/772/1/console

---

This PR changes the profile name for Shippable from `samples` to `samples.shippable`.

Changed files:

* `CI/pom.xml.shippable` : new definition
* `shippable.yaml` : activate the correct profile with `-P`

PR #319, opened 2018-06-14
Remove Shippable CI specific pom file

jmini:
As discussed in #305, this PR is about removing the different pom files used for CI.


jmini:
**For Shippable:**

I have compared `/pom` and `/CI/pom.xml.shippable`

* jmustache version
  - `/pom.xml` defines `jmustache-version` == `1.14`, 
  - `/CI/pom.xml.shippable` defines  `jmustache-version` == `1.12`
* `/pom.xml` has some additional configurations (like `distributionManagement` section) and a `release` profile to perform a release
* `maven-checkstyle-plugin`  and `formatter-maven-plugin` are commented `/pom.xml`
* the profile name used to run tests for some samples are not the same:
  - `samples` in `/pom.xml`
  - `samples.shippable` in `/CI/pom.xml.shippable` (introduced with #305)
* some paths are different

My take is that the small differences do not justify to maintain two version of the pom file.
With cfae3f768239b4ec290752bd21be92b6b14bac45 I have removed `/CI/pom.xml.shippable` and of course I have merged the `samples.shippable` profile section into `/pom.xml`

PR #325, opened 2018-06-15
[Golang][client] fix RFC-3339 date-time query param

grokify:
### PR checklist

@antihax
@bvwells

### Description of the PR

The OpenAPI 3.0 Spec and Swagger 2.0 Spec define the `date-time` string format as:

> dateTime | string | date-time | As defined by date-time - RFC3339

* Spec 3.0: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md
* Spec 2.0: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

An example of RFC-3339 `date-time` format is `2017-07-21T17:32:28Z`. Here is the full spec:

* RFC-3339: https://tools.ietf.org/html/rfc3339#section-5.6

The current SDK will create in an incorrectly formatted time for a `query` parameter. This happens because `client.go` `parameterToString` function uses `fmt.Sprintf("%v", obj)` which produces the following time format which is not a RFC-3339 time:

`2018-01-01 00:00:00 +0000 UTC`

This can be resolved by adding the following code to `parameterToString(obj interface{}, collectionFormat string) string`.

```
} else if t, ok := obj.(time.Time); ok {
    return t.Format(time.RFC3339)
}
```

This approach for testing time has been verified to work and is mentioned here:

https://stackoverflow.com/questions/41483505/golang-check-if-a-data-is-time-time

An example is on Go Playground here: https://play.golang.org/p/Jq16Shapio8

A test is not provided because it appears a `date-time` format `string` `query` parameter does not exist in the current Petstore API.

This is a breaking change since the date time format produced by the Client SDK changes.

More info can be seen here where the issue was encountered:

* https://github.com/grokify/go-ringcentral/issues/19
* https://github.com/swagger-api/swagger-codegen/issues/8039

This PR updates `client.mustache`. Other changed files are from a re-built `go-petstore` SDK in the sample folder.


antihax:
Looks good to me also.

PR #212, opened 2018-06-03
[PHP-Lumen] Lumen 5.6 support

wing328:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the correct branch: Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Cherry-picking https://github.com/OpenAPITools/openapi-generator/pull/114

Credits: @Lv-Yi 

I'll update the samples tomorrow and do some tests.

PR #343, opened 2018-06-18
[Lumen] fixed Output-Formats [*/*] injection in Lumen routes/web.php

Lv-Yi:
### PR checklist

### Description of the PR

fixed Output-Formats [ * / * ] injection in Lumen routes/web.php as discussed in #212 .
removed  MaxPermSize option in php-lumen-petstore-server.sh and security/lumen-petstore-server.sh which was obsolete in java8.
updated petstore server examples.


wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3902,8 +3902,8 @@ private void addProducesInfo(OpenAPI openAPI, ApiResponse inputResponse, Codegen
 
         int count = 0;
         for (String key : produces) {
-            // escape quotation to avoid code injection, "*/*" is a special case, do nothing
-            String encodedKey = "*/*".equals(key)? key : escapeText(escapeQuotationMark(key));
+            // escape quotation to avoid code injection, "*/*" is a special case, do nothing except adding blank
+            String encodedKey = "*/*".equals(key)? "* / *" : escapeText(escapeQuotationMark(key));
```
wing328: @Lv-Yi thanks for the fix but I'm afraid we need to keep it as previously reported by other users "*/*" should not be escaped/changed in order for it to function properly in the auto-generated code.

I think we'll need to create additional vendor extensions (e.g. x-consumes-doc) in postProcessOperations to handle "*/*" specifically in  the PHP documentation.

wing328: @Lv-Yi suggestion: what about doing it in the postProcessOperations of PHP Lumen generator (java class) instead so that other generators won't be impacted by the fix? Here is an example:

https://github.com/OpenAPITools/openapi-generator/pull/354/files#diff-be240257b80dee7947c87e9322ba4950R332

PR #353, opened 2018-06-19
Fix #352 - [PHP] [Improvement] Make Api return object instead of array

### LINKED ISSUE #352: [PHP] [Improvement] Make Api return object instead of array
##### Description

Today a given "myOperation" operation will have four methods declared on an API:
- myOperation()
- myOperationWithHttpInfo()
- myOperationAsync()
- myOperationAsyncWithHttpInfo()

The two last ones will return promisses, containing the same answer as the first two.
The first one returns either {{returnType}} or void, depending on if there is a returnType. (This is actually wrong, but it is being targeted in PR128, linked to issue #125)

But my problem is with the myOperatorWithHttpInfo() method: it returns an array with three elements:
- the message itself (response)
- the status code
- and an array with headers.

Considering the best practices for OOP, and comparing with other language generators that compound this project (for example, the C# generator), this method should rather return an Object that describes the API Response. Such an object should contain getters for each of the three elements (`getMessage()`, `getStatus()`, `getHeaders()` ).

My improvement suggestion (I'll push a PR as soon as I finish this description) is to add another class to the client, called ApiResponse (just like we already have in the C# client generator). This class will be the return for a class of methods ending with *WithApiResponse(), which will replace the existing *WithHttpInfo() ones and offer a more OOP-like interface for interacting with the API Response.

I can see two challenges with this approach:

1. Typing the message (return). This is already a problem for the *WithHttpInfo() methods, and is being solved with the addition of the following phpDoc in moustache:
  ```
  * @return array of {{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}null{{/returnType}}, HTTP status code, HTTP response headers (array of strings)
  ```
  , which results in something like: 
  ```
   * @return array of MyResponseClass, HTTP status code, HTTP response headers (array of strings)
  ```
I suggest using the same approach:
```
* @return APIResponse with "message" property of type MyResponseClass
```
, which is not a perfect solution, but can be used until the [Generics RFC](https://wiki.php.net/rfc/generics) gets an implementation and becomes approved.

2. Backwards compatibility. I have no idea of what is your current approach to breaking changes, though my previous experience with swagger tells me that it is not a big concern. Just in case, my PR will keep the existing *WithHttpInfo() methods, but will change its implementation to use the *WithApiResponse() ones internally.

 Prepare 3.0.3-SNAPSHOT

not applicable.

./bin/php-petstore.php, from the project root folder

##### Steps to reproduce

not applicable

##### Related issues/PRs

not applicable/not found

will push a PR on this issue

thomasphansen:
### PR checklist

- @jebentier 
- @dkarlovi 
- @mandrean 
- @jfastnacht 
- @ackintosh 

### Description of the PR
This PR will:
- add the class ApiResponse to the client
- implement *WithApiResponse methods in the Api classes, returning ApiResponse
- to preserve compatilibility, keep the *WithHttpInfo methods

For more details, please check issue #352.


wing328:
@thomasphansen thanks for the PR. From what I understand, this is not a breaking change as you're simply introducing another method.

Do you know if there'a any PHP guideline or best practice in terms of what's more preferrable to be returned by method: object vs array?

I'm not against the idea of using `ApiResonse`, which is what we use for Java API client for example, and would like to know which way (object or array) is more common among PHP developers.

thomasphansen:
Hello William, sorry for the late reply!

There are some articles on the internet where you may find general guidelines - even though this is actually a matter of coding OOP or not. Here goes what I found:

- https://softwareengineering.stackexchange.com/questions/296752/php-when-to-use-arrays-and-when-to-use-objects-for-mostly-data-storing-code-co  _(the first answer is by far the most clear and straight-forward answer for this object vs array discussion)_.
- https://www.brandonsavage.net/stop-returning-arrays-use-objects-instead/
- https://stackoverflow.com/questions/6710967/why-return-object-instead-of-array _(look at the two first answers, the first one offers a list of articles talking about the subject, and the second one is pretty detailed on why it is better to use objects instead of arrays)_.

As a matter of fact, I couldn't find any article defending the use of arrays instead of objects, except of course for not structured data (which is not the case). The first article has actually a very nice argument for using objects in our case:

> Are you building a library or shared code base?
>If so, use a class! Think about some new developer who is using your library, or another developer somewhere else in the company who has never used your code before. It will be much easier for them to look at a class definition and understand what that class does, or see a number of functions in your library which accept objects of a certain class, than to try and guess what structure they need to generate in a number of arrays. Also, this touches on the data consistency issues with #1: if you're developing a library or shared code base, be nice to your users: give them classes which enforce data consistency and protect them from making errors with the design of your data.

Hope it helps! :)

I just noticed you have added the PR on return types to master - I'll rebase my PR as soon as I have time, ok? :)

ackintosh:
@thomasphansen Thanks for the suggestion!
I think object is not just a container, it has a responsibility. What do you think is the responsibility of our ApiResponse?

thomasphansen:
Hello Akihito,

Thanks for your asking! :)

IMO, I don't think this is the case: REST and SOAP interfaces should return [DTO's](https://en.wikipedia.org/wiki/Data_transfer_object), which by definition have no logic associated to them - their only responsibility is to deliver the data obtained from the service. 

Also, just remember that all the generated Model classes also follow DTO pattern, since their only responsibility (like ApiResponse) is to contain data, with no logic involved... :)

(DTO pattern need is described by Martin Fowler [here](https://martinfowler.com/eaaCatalog/dataTransferObject.html) :) )

Hope it makes sense! :)

ackintosh:
Sorry for late reply! 💦 

https://martinfowler.com/eaaCatalog/dataTransferObject.html

There are 2 points I think important in the article.

- > Although the main reason for using a Data Transfer Object is to batch up what would be multiple remote calls into a single call, 
- > it's worth mentioning that another advantage is to encapsulate the serialization mechanism for transferring data over the wire. 

That describes the DTO's responsiblity. 💡 

Take the generated Model classes for example: Model has [serialization](https://github.com/ackintosh/openapi-generator-1/blob/c223f281713d9955911b2a762c4ac6061b09ac58/samples/client/petstore/php/OpenAPIClient-php/lib/Model/Pet.php#L478) and [validation](https://github.com/ackintosh/openapi-generator-1/blob/c223f281713d9955911b2a762c4ac6061b09ac58/samples/client/petstore/php/OpenAPIClient-php/lib/Model/Pet.php#L235) mechanism (and the logic to execute its responsiblity). And these are used in the process transfering data (e.g. Pet) between client and server.

Array can't alternative that as Array is just container.

I can't find the reason we should use Class instead of Array in the return value yet. 🤔💦

thomasphansen:
If you all agree with that, then I can code the change to make it an option! :)

wing328:
@thomasphansen quick question: have you evaluated other SDKs such as Facebook, Twitter, Slack, etc to see if these PHP SDK return a list or object?

This is not to say these SDKs should be our reference so good to know if they follow a particular pattern when returning data from functions.

wing328:
@thomasphansen if you've time for a quick chat, please ping me via https://gitter.im. Thanks.

thomasphansen:
Hello @wing328 

Sorry for the late reply (too busy here! :) ).

William, I am really sorry, but I don't have time to go through other API's and do such a check. The main reason for this pull request was (and it is) to make it easier to work with, and to make it offer an interface that would be coherent to what you already offer in other generators (such as the Java one, as you mentioned before). And even though I really love discussions on subjects like what a perfect API interface model would be, I also see that such a discussion should then affect all the current generators, not only the PHP one. IMHO, out of a major refactoring scenario, it is a bit pointless to argue if the APIResponse concept is valid or not, since it is already present in the openapi project.

Which means: if openapi is not enforcing a standard interface model through the many generators you have, then each generator is, in last analysis, "free" to do what they want - and thus, accepting or not this push request is actually more about if the php generator team agrees or not with it. And I am perfectly fine with that! You are the ones taking care of the code, and it should be your decision if a PR is a valid contribution or not. And if someone does not agree, well, that's what forks were created for! :)

For now, in our own projects, I'm using the fork we did, and we will do our best to keep it up to date with your master branch. If you decide that this contribution is relevant, or if you think there is anything I could do to make it more adequate to openapi, just let me know, ok? :)

Have a nice day! :)

ybelenko:
@thomasphansen 
One day after I'm not sure that I've suggested good solution. Let's take a look at current code:
```php
/**
 * Operation fakeOuterNumberSerialize
 *
 * @param  float $body Input number as post body (optional)
 *
 * @throws \OpenAPI\Client\ApiException on non-2xx response
 * @throws \InvalidArgumentException
 * @return float
 */
public function fakeOuterNumberSerialize($body = null)
```
and with a quick look you perfectly knows what you'll recieve and how to extract data(don't need to extract anything at all)
Let's imagine with PSR-7 response instead:
```php
/**
 * Operation fakeOuterNumberSerialize
 *
 * @param  float $body Input number as post body (optional)
 *
 * @throws \OpenAPI\Client\ApiException on non-2xx response
 * @throws \InvalidArgumentException
 * @return \Psr\Http\Message\ResponseInterface
 */
public function fakeOuterNumberSerialize($body = null)
```
now you know that you'll receive `ResponseInterface` and no clue how to extract `float` data that you looking for. 

It seems to me that we need to generate something like `ApiResponseForFakeOuterNumberSerialize` which implements PSR-7 interface and have getters and setters. If so user can see that he'll recieve this class as a return and he can read setters and getters from it's documentation to get data. I'm afraid that it's a huge extra work.

wing328:
> William, I am really sorry, but I don't have time to go through other API's and do such a check. The main reason for this pull request was (and it is) to make it easier to work with, and to make it offer an interface that would be coherent to what you already offer in other generators (such as the Java one, as you mentioned before). And even though I really love discussions on subjects like what a perfect API interface model would be, I also see that such a discussion should then affect all the current generators, not only the PHP one. IMHO, out of a major refactoring scenario, it is a bit pointless to argue if the APIResponse concept is valid or not, since it is already present in the openapi project.

@thomasphansen that's perfectly fine. To be clear I'm not trying argue if APIResponse is valid or not. I just want to know which way is more preferred by PHP developers.

> Which means: if openapi is not enforcing a standard interface model through the many generators you have, then each generator is, in last analysis, "free" to do what they want - and thus, accepting or not this push request is actually more about if the php generator team agrees or not with it. And I am perfectly fine with that! You are the ones taking care of the code, and it should be your decision if a PR is a valid contribution or not. And if someone does not agree, well, that's what forks were created for! :)

At the very beginning, we tried to enforce a standard so that all API clients behave more or less the same way (e.g. all clients should return an object). However, we got feedback later what make senses in one generator (e.g. Java API client) may not welcome by developers using other languages (e.g. Perl). As a result, we let the users of the generator decide what works for them. Sometime we may not get a strict answer and we will need to provide an option to make both parties having different opinions happy.

In this particular case, I can't speak on behalf of all PHP developers. Based on my guess, some may prefer returning a list while others may prefer returning an object (e.g. ApiResponse). I still think providing an option is the best middle ground to make everyone happy.

Issue #350, opened 2018-06-19
[TypesScript] Models are sorted incorrectly in output


gbrown-ce:
##### Description

When generating the client, the model classes in the resulting api.ts file are sorted alphabetically.  In many cases this is fine and doesn't matter.  The problem comes in when you are using inheritance.  In TypeScript you cannot put an inherited class before its parent.  If for instance you have a parent "Section" and a child "ButtonSection", this will cause the Typescript compilation to fail since the child comes before the parent alphabetically.

##### openapi-generator version

3.0.2

##### OpenAPI declaration file content or url

```yaml
swagger: "2.0"
info:
  version: "1.0.0"
  title: Sections Service
basePath: /
# 
schemes:
  - http
consumes:
  - application/json
produces:
  - application/json
paths:
  /sections:
    x-swagger-router-controller: sections
    get:
      description: Retrieves sections
      operationId: getSections
      parameters:
        - name: skip
          in: query
          description: Number of items to skip
          required: false
          type: integer
          format: int64
          default: 0
          minimum: 0
        - name: take
          in: query
          description: Number of items to return
          required: false
          type: integer
          format: int64
          default: 20
          minimum: 1
          maximum: 100
      responses:
        200:
          description: Success
          headers:
            X-Total-Count:
              description: Total number of sections
              type: integer
              format: int64
              minimum: 0
          schema:
            type: array
            items:
              $ref: "#/definitions/Section"
  /swagger:
    x-swagger-pipe: swagger_raw
definitions:
  Section:
    type: object
    discriminator: type
    required:
      - name
      - type
    properties:
      name:
        type: string
      type:
        type: string
        enum:
          - ButtonSection
  ButtonSection:
    type: object
    allOf:
      - $ref: '#/definitions/Section'
      - type: object
        required:
          - buttons
        properties:
          buttons:
            type: array
            items:
              $ref: '#/definitions/Button'
  Button:
    type: object
    required:
      - index
      - label
    properties:
      index:
        type: integer
        format: int64
        minimum: 0
      label:
        type: string
```

##### Command line used for generation

docker run -d --rm -v $WORKSPACE:/input:ro -v $WORKSPACE/output:/output openapitools/openapi-generator-cli generate -i /input/swagger.yaml -g typescript-node -o /output

##### Steps to reproduce

Run the above command on reference file

##### Suggest a fix/enhancement

[...]

PR #363, opened 2018-06-20
Restructure TypeScript Node generation into separate files (PHNX-1041)

gbrown-ce:
### PR checklist

### Description of the PR

The current version of TypeScript node generation puts all of the generated code in a single file. This causes issues when using inheritance as each model may depend on another and they must be included in the proper order.

Split the mustache template into pieces so that it can properly separate the models and different apis to allow for proper compilation.  I also removed some compiled JS files from the petstore example generation that shouldn't have been in the repository.

Related issue: #350 

@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny


macjohnny:
@gbrown-ce thanks for the PR! will have a look at it soon

gbrown-ce:
@macjohnny thanks! And don't be afraid of the 50+ files changed.  Only the first 7 are really changes.  The rest are just generated samples.

gbrown-ce:
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny I believe the necessary CI step has passed here and this should be ready for review/merge.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -141,9 +243,54 @@ private void addNpmPackageGeneration() {
...
         supportingFiles.add(new SupportingFile("tsconfig.mustache", getPackageRootDirectory(), "tsconfig.json"));
     }
+    
+    private String getIndexDirectory() {
+        String indexPackage = modelPackage.substring(0, Math.max(0, modelPackage.lastIndexOf('.')));
+        return indexPackage.replace('.', File.separatorChar);
+    }
+
+    @Override
+    public String getSchemaType(Schema p) {
```
macjohnny: could you please explain in the comments why these type mapping methods are needed / what they do?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -141,9 +243,54 @@ private void addNpmPackageGeneration() {
...
+        String openAPIType = super.getSchemaType(p);
```
macjohnny: could you please explain in the comments why these type mapping methods are needed / what they do?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -141,9 +243,54 @@ private void addNpmPackageGeneration() {
...
+            type = typeMapping.get(type);
+        }
+        return type;
+    }
+    
+    private boolean isLanguagePrimitive(String type) {
+        return languageSpecificPrimitives.contains(type);
+    }
+
+    private boolean isLanguageGenericType(String type) {
```
macjohnny: could you give an example input in the method docs?
I guess you expect something like `Array<string>`, right?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -141,9 +243,54 @@ private void addNpmPackageGeneration() {
...
+        return false;
+    }
 
     private String getPackageRootDirectory() {
         String indexPackage = modelPackage.substring(0, Math.max(0, modelPackage.lastIndexOf('.')));
         return indexPackage.replace('.', File.separatorChar);
     }
+    
+    private String getApiFilenameFromClassname(String classname) {
+        String name = classname.substring(0, classname.length() - "Api".length());
```
macjohnny: how about setting the "Api" suffix as a member property, which later could possibly be customized with a configuration option?

macjohnny:
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx 
I would be happy to get a second opinion

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
gbrown-ce: @macjohnny I believe the idea for this is to allow the "Buffer" type to be used properly. Initially when I did this, it tried to import it as ./models/Buffer which is wrong. There is similar work done in the angular client for the "Blob" type.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
gbrown-ce: @macjohnny I'm not sure of the exact purpose of this one, again, it is taken from the angular generator.  But yes, I'd assume you're correct

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
gbrown-ce: Sure, I can move that

gbrown-ce:
@macjohnny yes I tested it with my own project and it works great.

gbrown-ce:
@macjohnny I have updated the PR with your requested changes.  Even went so far as to add the api suffix as a parameter. I also went through and ran all 5 of our swagger documents through it and the newly generated clients worked perfectly.

macjohnny:
@gbrown-ce thanks for the changes, I think now it should be good to go.

gbrown-ce:
@macjohnny I'm about to head on a week long vacation. Can I trust you to get this merged before I get back? That'd be a huge help.

macjohnny:
@gbrown-ce This PR will eventually be merged by @wing382, I am not able to do that.

macjohnny:
Anyway, you could also temporarily build from your branch and publish the artifact in your local repository

gbrown-ce:
@macjohnny good to know. Can you take a look @wing328 ?

As far as local artifacts, yes I could do that (and have for testing), but we are integrating our Jenkins pipeline with the openapi-generator-cli docker image, so to be fully finished on my end, I need to final code to be in (or make my own docker image, which I'd rather avoid)

gbrown-ce:
Oh and the failures in the Travis build seem to be related to a problem that already exists in master.

wing328:
Please ignore the failure related to rust-server for the time being as the failure is due to CI pulling a newer version of the dependency.

wing328:
For "/" in the file path, please replace it with `File.separator` so that the generator works with Windows platform as well.

wing328:
If I understand the PR correctly, this enhancement is a breaking change without fallback so the PR should target the `4.0.x` branch instead.

macjohnny:
@wing328 @gbrown-ce 
>For "/" in the file path, please replace it with File.separator so that the generator works with Windows platform as well.


This change should not be applied, since the imports in typescript files always use `/` as a file separator. This also works e.g. in the `typescript-angular` generator, see e.g.
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java#L445

macjohnny:
> If I understand the PR correctly, this enhancement is a breaking change without fallback so the PR should target the 4.0.x branch instead.

this PR could be turned into a non-breaking change by simply re-exporting the files in the `api.ts`, where they were originally located:

https://github.com/OpenAPITools/openapi-generator/pull/363/files#diff-808fb691981e56904bccf3634e507ce2R11

@gbrown-ce would you like to apply this change?

gbrown-ce:
@macjohnny Just got back from vacation and made the suggested changes.  Just waiting for CI to run through.

gbrown-ce:
@macjohnny @wing328 looks like CI is happy.  Do the changes look good to you guys?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -110,10 +211,12 @@ public void setNpmRepository(String npmRepository) {
     @Override
     public void processOpts() {
         super.processOpts();
-        supportingFiles.add(new SupportingFile("api.mustache", null, "api.ts"));
+        supportingFiles.add(new SupportingFile("models.mustache", modelPackage().replace('.', File.separatorChar), "models.ts"));
+        supportingFiles.add(new SupportingFile("apis.mustache", apiPackage().replace('.', File.separatorChar), "api.ts"));
```
macjohnny: isn't there a collision for the destination filename `api.ts`, since both `apis.mustache` and `index.mustache` are generated to `api.ts`?
anyway I would suggest to rename `api.mustache` to `api-single.mustache` and `apis.mustache` to `api.mustache` in order to avoid confusions.

gbrown-ce: No collision there, the apis.mustache goes into a subfolder, but yes, I can clean that up a bit.

gbrown-ce: @macjohnny 

I'm going to go with this:

api-all.mustache = all of the apis re-exported together (output as /apis/apis.ts)
api-single.mustache = single api (output as /apis/{name}Api.ts)
api.mustache = entry point (output as /api.ts to prevent breaking change)

I thin this should avoid any confusion

gbrown-ce: @macjohnny done.  Just awaiting CI again

macjohnny:
@gbrown-ce could you please merge the current `master` and resolve the conflict?
@wing328 could you please add this to the `3.1.0` milestone? it should not be a breaking change, since the `api.ts` re-exports all classes that are now in separate files.
@TiFu could you please have a quick look at this?

jmini:
@macjohnny: thank you a lot for the coordination and all the feedback on this PR.

It is nice that you found a way to make this a non-breaking change. With this it can be merged into `master` for `3.1.0-SNAPSHOT`. This week would be the right timing to do it.

gbrown-ce:
@macjohnny @wing328 I sorted out the merge conflicts.  Ready to merge into master.

jmini:
Maybe the change is important enough to be mentioned in the [Migration guide between OpenAPI-Generator versions](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/migration-guide.adoc). => This can be filed as a separated PR.

Issue #364, opened 2018-06-21
DockerHub: consume a specific version ("3.0.x" or "3.1.x")


jmini:
I am not a DockerHub user myself, now that we have introduced branches to work on different version of the project:

(current state 21.06.2018)
* `master` for version `3.0.3-SNAPSHOT`
* `3.1.x` for version `3.1.0-SNAPSHOT`
* `4.0.x` for version `4.0.0-SNAPSHOT`

See also: https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches

I am wondering how this works in DockerHub:
https://hub.docker.com/r/openapitools/openapi-generator-cli/tags/

"latest" tag seems to be the last built, regardless of the branch.

What would make sense?

* "latest" for latest from the `master` branch
* "latest_<branch name>" in the other cases (for our protected branches like `3.1.x` and `4.0.x` - not for the working branches)

etherealjoy:
`latest` would be `master` 
`3.1.x` would be `3.1.0-SNAPSHOT`
This is usually the norm but is not binding

jmini:
According to those lines:

https://github.com/OpenAPITools/openapi-generator/blob/12ab7e27521dce18b7fb27e40ce2eb636f4fac29/.travis.yml#L125-L128

We are not pushing the Docker images for the `3.1.x` and `4.0.x` branches.

On the Docker-Hub `latest` Tag, you can the continuously built artifact from the `master` branch.

---

I am not sure if there is an interest to have other GitHub-tags for the other branches.

wing328:
What about using tags like "SNAPSHOT" (e.g. 3.1.1-SNAPSHOT) as @etherealjoy suggested? (I don't think in general SNAPSHOT is exclusively used for Java JAR and developers should be able to understand what SNAPSHOT mean for a docker image release)

etherealjoy:
> latest would be master
3.1.x would be 3.1.0-SNAPSHOT
This is usually the norm but is not binding

This is like @jimschubert  mentioned is the norm. But of course not every docker repo follows this.
tag `latest` --> `master` branch
tag `3.0.x` --> `3.0.x` branch if it exists
tag `3.0` --> last released `3.0.x` branch if it exists, or 3.0.0 for the particular version.
tags has to reference a persistent branch not something which will be merged and deleted.
Multiple tags can be applied to an image.

jmini:
@etherealjoy 

[...]

PR #390, opened 2018-06-25
Improve Docker Tags

jmini:
This PR improve the way Docker-Tags are made. See the discussion in #364.

In addition to what we have now (`latest` and tags for version), new tags for `SNAPSHOT` version are introduced:

* `3.0.x` Tag: contains first `3.0.2-SNAPSHOT` and after the `3.0.2` release it contains `3.0.3-SNAPSHOT` and so on.
* `3.1.x` Tag: contains first `3.1.0-SNAPSHOT` and after the `3.1.0` release it contains `3.1.1-SNAPSHOT` and so on.
* `4.0.x` Tag: contains first `4.0.0-SNAPSHOT` and after the `4.0.0` release it contains `4.0.1-SNAPSHOT` and so on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,10 +122,27 @@ after_success:
         popd;
       fi;
     fi;
-  ## docker: build and push openapi-generator-online to DockerHub
-  - if [ $DOCKER_HUB_USERNAME ]; then echo "$DOCKER_HUB_PASSWORD" | docker login --username=$DOCKER_HUB_USERNAME --password-stdin && docker build -t $DOCKER_GENERATOR_IMAGE_NAME ./modules/openapi-generator-online && if [ ! -z "$TRAVIS_TAG" ]; then docker tag $DOCKER_GENERATOR_IMAGE_NAME:latest $DOCKER_GENERATOR_IMAGE_NAME:$TRAVIS_TAG; fi && if [ ! -z "$TRAVIS_TAG" ] || [ "$TRAVIS_BRANCH" = "master" ]; then docker push $DOCKER_GENERATOR_IMAGE_NAME && echo "Pushed to $DOCKER_GENERATOR_IMAGE_NAME"; fi; fi
-  ## docker: build cli image and push to Docker Hub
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -122,10 +122,27 @@ after_success:
...
+          docker build -t $DOCKER_CODEGEN_CLI_IMAGE_NAME -t $DOCKER_GENERATOR_IMAGE_NAME:$MVN_VERSION_FOR_DOCKER_TAG ./modules/openapi-generator-cli
+        else
+          docker build -t $DOCKER_GENERATOR_IMAGE_NAME:$MVN_VERSION_FOR_DOCKER_TAG ./modules/openapi-generator-online
+          docker build -t $DOCKER_GENERATOR_IMAGE_NAME:$MVN_VERSION_FOR_DOCKER_TAG ./modules/openapi-generator-cli
+        fi;
+        if [ ! -z "$TRAVIS_TAG" ] || [ "$TRAVIS_BRANCH" == "master" ] || [ "$TRAVIS_BRANCH" == "3.1.x" ] || [ "$TRAVIS_BRANCH" == "4.0.x" ];
+          docker push $DOCKER_GENERATOR_IMAGE_NAME
+          echo "Pushed to $DOCKER_GENERATOR_IMAGE_NAME"; 
+          docker push $DOCKER_CODEGEN_CLI_IMAGE_NAME
+          echo "Pushed to $DOCKER_CODEGEN_CLI_IMAGE_NAME";
```
wing328: If `docker push $DOCKER_GENERATOR_IMAGE_NAME` fails but `docker push $DOCKER_CODEGEN_CLI_IMAGE_NAME` succeeds, is it correct to say that the CI job will still build successfully?

(I think such situation rarely occurs but still want to confirm the behavior in case that happens)

jmini: Good question.
Did my modification changed this?

---

Was this the reason why there was 2 items in the old version?

```
- if [ $DOCKER_HUB_USERNAME ]; ... => DOCKER_GENERATOR_IMAGE_NAME
- if [ $DOCKER_HUB_USERNAME ];  ... => DOCKER_CODEGEN_CLI_IMAGE_NAME
```

---

My idea to bring both together was to reduce copy-paste code, but I can split it back to 2 items.

wing328: > Was this the reason why there was 2 items in the old version?

I wasn't the one who added it but that could be the reason.

> My idea to bring both together was to reduce copy-paste code, but I can split it back to 2 items.

I definitely like your idea and formatting to make the 1-liner bash command more readable.

What about checking the exit status of line 141 before proceeding further?

jmini: > What about checking the exit status of line 141 before proceeding further?

Good Idea, am not not really an expert in this domain.

jmini:
From the doc https://docs.travis-ci.com/user/customizing-the-build

> Note the `set -ev` at the top. The `-e` flag causes the script to exit as soon as one command returns a non-zero exit code. This can be handy if you want whatever script you have to exit early. It also helps in complex installation scripts where one failed command wouldn’t otherwise cause the installation to fail.
>
>The `-v` flag makes the shell print all lines in the script before executing them, which helps identify which steps failed.

We are using the flag:

https://github.com/OpenAPITools/openapi-generator/blob/ed1ce3732504e14c60d7a7ec94b809e3bb74491f/.travis.yml#L92-L94

So I think we the case described by @wing328 seems to be covered.

wing328:
👌 

Merging this into master. Thanks for the enhancement.

jmini:
The change introduced by this PR did not work on travisCI. It was reverted with commit d43801a9b72e9ecee760b9ede031aa355f1f8e9e

PR #360, opened 2018-06-20
Fix alias to map in the model's properties

wing328:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.1.x`, `4.0.x`. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

To fix #255, #191 

This is the first fix for "alias". Please expect more PRs to fix various usage of alias (e.g. alias of alias, alias of array, alias handling in response, parameters, etc)

In the upcoming PRs, I'll also add more Java unit tests to cover the issue moving forward.

PR #82, opened 2018-05-17
Revert PR #45

jmini:
Discussed in #79

Reverts:
Consider minLength, maxLength and pattern in referenced schema (#45)
This reverts commit 6b8079808b0d1d730324caf26489030a0cd960bd.

PR #45, opened 2018-05-15
Consider minLength, maxLength and pattern in referenced schema

jmini:
Reported by @FatCash on https://github.com/swagger-api/swagger-codegen/issues/8001

OAS2 example:

```
definitions:
  SomeObj:
    type: object
    properties:
      nick:
        $ref: "#/definitions/NickName"
  NickName:
    type: string
    minLength: 1
    maxLength: 3
    pattern: "^[A-Z]+$"
```

---

The solution proposed here is really temporary. When we will work on #20, I think that a lot of the method will be refactored. For example the OpenAPI instance will be propagated instead of `Map<String, Schema>` everywhere and use `ModelUtils` to resolve referenced Schema.

PR #401, opened 2018-06-26
Restore unit test that consider minLength, maxLength and pattern in referenced schema

jmini:
This PR restores the unit test that consider `minLength`, `maxLength` and `pattern` when they are defined in a in referenced schema.

OAS2 example:

```
definitions:
  SomeObj:
    type: object
    properties:
      nick:
        $ref: "#/definitions/NickName"
  NickName:
    type: string
    minLength: 1
    maxLength: 3
    pattern: "^[A-Z]+$"
```

This was filed as originally as swagger-api/swagger-codegen#8001

I have tried to implement a solution for this with #45, but my solution had too many side effects and was reverted with #82.

This PR brings back the tests without changing anything to the implementation. I think the changes made with #360 by @wing328 solves also issue swagger-api/swagger-codegen#8001.


jmini:
Unit tests are green, this PR is just about new assertions in one unit test, no changes to the implementation. => Self approving.

Issue #234, opened 2018-06-06
typescript-angular: Make the api suffix configurable


macjohnny:
See original issue at https://github.com/swagger-api/swagger-codegen/issues/8281

macjohnny:
@tomvangreen would you like to implement this here at OpenAPI Generator?

tomvangreen:
I have to implement it for swagger codegen, as that is what we use in our project(s) and I cannot change it at this time. But I can see how hard it is to also provide it for openapi-generator. Can't guarantuee anything, but I will see.

macjohnny:
@tomvangreen by the way, migration to openapi-generator is as simple as replacing the dependency in your `pom.xml`:
```xml
<dependency>
    <groupId>org.openapitools</groupId>
    <artifactId>openapi-generator-maven-plugin</artifactId>
    <version>${openapi-generator-version}</version>
</dependency>
```
I did that recently in our project, and it worked like a charm.

I would be happy to review your PR with the changes.

tomvangreen:
I have added two parameters for the typescript angular code generator, serviceSuffix and serviceFileSuffix. I tried running the tests and I get an error, that those two parameters have not been tested. I looked through the repo to try to find out where or how I can test those new parameters, but I am a bit lost.

Can anybody point me in the right direction for this?

macjohnny:
@tomvangreen what is the error message?
@wing328 any idea where to fix this?

tomvangreen:
@macjohnny Sorry, I could have added the message right away. I get this error message:
`Failed tests: 
  TypeScriptAngularClientOptionsTest>AbstractOptionsTest.checkOptionsHelp:61 These options weren't checked: serviceSuffix, serviceFileSuffix.
`

tomvangreen:
I think I got it. I needed to add the parameters in the `TypeScriptAngularClientOptionsProvider`.

tomvangreen:
@macjohnny I think this issue can be closed, as my pull request has been merged. 

But before you close it, maybe I could hijack this issue for another related topic/questsion. One other point of pain for us is the naming of the model classes. The swagger api we are using is also used by other clients and we don't have much control over how the models are tagged/named. 

[...]

PR #418, opened 2018-06-29
typescript-angular: Feature/model suffix

tomvangreen:
### PR checklist

### Description of the PR
I introduced a `modelSuffix` and `modelFileSuffix` parameter in order to add suffixes to the generated models and model file names. Taking the samples from the sample, with this setting you could for example change the model `Pet` in the file `pet.ts` to `PetVm` in the file `pet.vm.ts` by defining the model suffix `Vm` and the model file suffix `.vm`.

The motivation behind this change is to have a clear distincation between generated models and models that are created by a developer. Having the ability to add a suffix to the generated models makes it immediatly obvious of what kind a model is.

I currently see one issue and that is that the class suffix gets also appended to enumeration types. The `StatusEnum` in the `Pet` class will become `StatusVmEnum` if I set the `modelSuffix` option to `Vm`. I will have a look if I find how I can avoid that. 

@TiFu @taxpon  @sebastianhaas @kenisteward @Vrolijkx @macjohnny


tomvangreen:
#234

tomvangreen:
Yes, you are right. You could set the suffix to something invalid. I think this could also be a problem with the filename.

From what I see we have following options:
1. Validate and throw an exception if the name is invalid.
2. Sanitize the input and strip every invalid character and show a warning.
3. Let it run and fail.

I would say because of the filename probably approach 1 or 2 should be used. Better to catch this problem than to let it fail uncontrolled. Do you know of any similar cases in this project and how it is handled there?

tomvangreen:
I added validation for class and file suffixes. As my earlier change with the service suffix has the same problem, I also added the validation for those configurations.

Currently I only allow alpha numeric characters for class suffixes. For file suffixes I additionally allow the characters '.' and '-'. 

If any of the four config values is invalid I throw an IllegalArgument exception and display a message.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -486,7 +505,44 @@ private String getApiFilenameFromClassname(String classname) {
...
+    }
+
+    private String removeModelSuffixIfNecessary(String name) {
+        if (modelSuffix.length() == 0 || !name.endsWith(modelSuffix)) {
+            return name;
+        }
+        return name.substring(0, name.length() - modelSuffix.length());
     }
 
+    private void validateFileSuffixArgument(String argument, String value) {
```
macjohnny: please add a method description. describe what `argument` and `value` are or rename them to better understand their meaning

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -486,7 +505,44 @@ private String getApiFilenameFromClassname(String classname) {
...
+        if (!value.matches(FILE_NAME_SUFFIX_PATTERN)) {
+            throw new IllegalArgumentException(
+                    String.format("%s file suffix only allows '.', '-' and alphanumeric characters.", argument)
+            );
+        }
+    }
+
+    private void validateClassSuffixArgument(String argument, String value) {
```
macjohnny: same here

TiFu:
Agreed. Thanks for the PR!

macjohnny:
@wing328 could you please add this to the `3.1.0` milestone?

PR #183, opened 2018-05-30
Revise how to obtain the example value

ackintosh:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This is the first fix for `example`.

PR #419, opened 2018-06-29
Obtain the example value from examples

jmini:
### PR checklist

### Description of the PR

After #183 made by @ackintosh, I think it would also be good to support `examples` in addition to `example` (in a similar way).

If both (`example` and `examples` properties) are set, we keep the `example` value, if multiple `examples` are set we keep the first one.

In the future we might want to update codegen classes so that they support multiple examples values (OAS3 feature)

### DIFF
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
@@ -274,7 +274,7 @@ public class DefaultCodegenTest {
         final OpenAPI openAPI = new OpenAPIParser().readLocation("src/test/resources/3_0/examples.yaml", null, new ParseOptions()).getOpenAPI();
         final DefaultCodegen codegen = new DefaultCodegen();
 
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
```diff
@@ -262,23 +268,35 @@ public void testExample3() {
...
+        Assert.assertEquals(codegenParameter2.example, "example3: parameter value");
     }
 
     @Test
     public void testExample4() {
         final OpenAPI openAPI = new OpenAPIParser().readLocation("src/test/resources/3_0/examples.yaml", null, new ParseOptions()).getOpenAPI();
         final DefaultCodegen codegen = new DefaultCodegen();
 
-        Operation operation = openAPI.getPaths().get("/example4").getGet();
+        Operation operation = openAPI.getPaths().get("/example4/singular").getPost();
```
ackintosh: Thanks for the follow-up!

PR #431, opened 2018-07-01
Mark postProcessOperations(Map) as deprecated

jmini:
Instead of postProcessOperations(Map)

### PR checklist

### Description of the PR

On Codegen Classes we have 2 methods that are similar and redundant:

* `Map<String, Object> postProcessOperations(Map<String, Object> objs)`
* `Map<String, Object> postProcessOperationsWithModels(Map<String, Object> objs, List<Object> allModels)`

They are called at the exact same place with the difference that `postProcessOperationsWithModels(..)` also gives access to the list of models.

This PR marks `postProcessOperations(..)` as deprecated and recommend to use `postProcessOperationsWithModels(..)` instead.

In a future major release (like `4.0.0`) we might want to remove `postProcessOperations(..)` in order to avoid redundant similar methods in our API.


wing328:
Yup, `postProcessOperationsWithModels` was introduced as developers may still be using `postProcessOperations`. In the future major release, we should just keep one and notify developers accordingly about the deprecation.

PR #435, opened 2018-07-02
[Java] Add support for Spring 5 WebClient

0v1se:
### PR checklist

- [-] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [-] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [-] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.1.x`, `4.0.x`. Default: `master`.
- [-] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Basic support for Spring 5 WebClient with project reactor's Mono/Flux


jmini:
> Shippable — Run 1154 status is FAILED. 

The new sample output `samples/client/petstore/java/webclient` is not up-to-date with the generator.

Please run:

* `mvn clean verify` on the root
* run `bin/java-petstore-webclient.sh`
* commit the changes to this PR

jmini:
If you like you can credit yourself in the root `README.md` under template creator:
https://github.com/OpenAPITools/openapi-generator#template-creator

jmini:
For the CI, can you please add the new generated sample to `/CI/pom.xml.circleci`:
```xml
 <module>samples/client/petstore/java/webclient</module> 
```

Here:
https://github.com/OpenAPITools/openapi-generator/blob/001f5ae50d96e9b69e377f136b7fab20e8feba85/CI/pom.xml.circleci#L843-L856

(the list is not really sorted... You can add your project somewhere near the other java client tests)

jmini:
Thank you for the update.

---

I have fetch your branch on my computer to perform some tests

I have tried to run:
```
mvn verify -f samples/client/petstore/java/webclient
```

An I got a lot of compile errors, about `org.threeten.bp` missing. Maybe you should generate the sample with java8 date lib.

I also have compile error like this one:
```
[ERROR] /Users/jbr/Git/openapi-generator/samples/client/petstore/java/webclient/src/main/java/org/openapitools/client/api/PetApi.java:[144,72] cannot find symbol
[ERROR]   symbol:   variable CollectionFormat
[ERROR]   location: class org.openapitools.client.ApiClient
```

0v1se:
@jmini thanks
will check this and update

0v1se:
@jmini fixed some bugs.
Sample project compiles ok now.

I disabled threetenbp this way (no need for threetenbp cause java8 is required):
```java
        if (WEBCLIENT.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) {
            dateLibrary = "java8";
        }
```
Is it ok to disable threeten this way? Or is there better method?

jmini:
> Is it ok to disable threeten this way? Or is there better method?

Good question. For the moment we do not have really a way to define that some options are only available with certain lib or other config.

I had a quick look at the code, `vertx` requires java8 by doing it `setJava8Mode(true)` and `additionalProperties.put("java8", "true")` in the big if-elseif block. Maybe you can do it in a similar manner:

https://github.com/OpenAPITools/openapi-generator/blob/960412a9b4b60c597ebab5b11871b5848f5c97cb/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java#L283-L287

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
```diff
@@ -151,6 +153,10 @@ public String getHelp() {
 
     @Override
     public void processOpts() {
+        if (WEBCLIENT.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) {
```
jmini: I guess that this check is too early. I am not sure, but I think that if the value is set to something on the command line, the value will be set with `super.processOpts()` and will override your check.

If possible I would override the wrong/unsupported values in the "else if" block line 290.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
```diff
@@ -280,6 +286,8 @@ public void processOpts() {
         } else if (RESTTEMPLATE.equals(getLibrary())) {
             additionalProperties.put("jackson", "true");
             supportingFiles.add(new SupportingFile("auth/Authentication.mustache", authFolder, "Authentication.java"));
+        } else if (WEBCLIENT.equals(getLibrary())) {
+            additionalProperties.put("jackson", "true");
```
jmini: here I would set other values like `java8Mode`, `dateLibrary`...

0v1se: At first I added this check at line 290, but it was not working, because dateLibrary is used in AbstractJavaCodegen.processOpts(), so setting it on line 290 doesn't work.

As you stated, setting it on line 156 doesn't always work either (because it's set inside AbstractJavaCodegen.processOpts if parameter about date library is present).

I think, the only solution is to divide processOpts(). Currently it does some things:
* gets params and sets fiels (e.g dateLibrary)
* does some actions according to fields
For example, in the end of processOpts:
```java
        if ("threetenbp".equals(dateLibrary)) {
            additionalProperties.put("threetenbp", "true");
            additionalProperties.put("jsr310", "true");
            typeMapping.put("date", "LocalDate");
            typeMapping.put("DateTime", "OffsetDateTime");
            importMapping.put("LocalDate", "org.threeten.bp.LocalDate");
            importMapping.put("OffsetDateTime", "org.threeten.bp.OffsetDateTime");
        } else if ("joda".equals(dateLibrary)) {
```

Possibly, this method (processOpts) should do only one thing (should be divided)
and the method for checking/updating fields should be added.

Actually, it's not a big deal now. I can add back threetenbp support (but it's useless cause java8 support only), then there is no reason to do all these things now.

What do you think?

jmini: Thank you a lot for the analysis and the explanations.

I think that we can do the refactoring you have proposed (splitting `processOpts()`) in a separated PR.

jmini:
I have tested the sample locally:

```
$ mvn verify -f samples/client/petstore/java/webclient/
[INFO] Scanning for projects...
[INFO] 
[INFO] Building petstore-webclient 1.0.0
[INFO] 
...
[INFO] BUILD SUCCESS
[INFO] Total time: 3.694 s
[INFO] Finished at: 2018-07-03T16:01:27+02:00
[INFO] Final Memory: 21M/256M
```

jmini:
@0v1se: Thank you a lot for this contribution.

jmini:
Twitt: https://twitter.com/oas_generator/status/1014415775766073344

Issue #443, opened 2018-07-03
[jaxrs-resteasy] generated code doesn't compile when @Produces contains multiple values


tnmtechnologies:
##### Description

In case of multiple possible Content-Type values for a response, the values for the Produces annotation are not separated by a coma.
It leads to a compilation error.

##### openapi-generator version
openapi-generator version is 3.0.3

##### OpenAPI declaration file content or url

```
   responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/LocationData'
        default:
          description: Unexpected error
          content:
            application/problem+json:
              schema:
                $ref: './TS29571_CommonData.yaml#/components/schemas/ProblemDetails'
```

##### Command line used for generation

 java  -jar ./openapi-generator-cli-3.0.3.jar generate -i TS29572_Nlmf_Location.yaml -g jaxrs-resteasy -o samples.openapi && (cd samples.openapi; mvn package)

##### Steps to reproduce

Compilation errors are:
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[39,35] '}' expected
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[39,63] class, interface, or enum expected
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[45,12] class, interface, or enum expected
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[45,76] class, interface, or enum expected
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[45,105] class, interface, or enum expected
[ERROR] ...../samples.openapi/src/gen/java/org/openapitools/api/DetermineLocationApi.java:[48,5] class, interface, or enum expected
[INFO] 6 errors
```

Generated source code looks like:
```
@io.swagger.annotations.Api(description = "the determine-location API")
@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaResteasyServerCodegen", date = "2018-07-03T11:18:06.912088700+02:00[Europe/Paris]")
public class DetermineLocationApi  {

    @Inject DetermineLocationApiService service;

    @POST
    
[...]

PR #445, opened 2018-07-03
fix: The values for @Produces annotation were not separated by a comma.

tnmtechnologies:
###Bug found with jaxrs-resteay source code.
Issue #443

### Description of the PR

Before adding a new mediaType into the produces list, we put a ["hasMore", "true"] to the last mediaType in order to be sure "hasMore" value is "true" because in some situations, "hasMore"  value in the map was null.


jmini:
@tnmtechnologies: thank you a lot for this PR! Great job.

I took the liberty to add a unit test that correspond to the case you have fixed (inspired by what you have posted in #443).

I will merge this PR when the CI Job are green.

tnmtechnologies:
Thank you too. Happy to know it helps.

PR #470, opened 2018-07-05
[Python][Client] pure library client package

rienafairefr:
### PR checklist

**-> Caveat** I wasn't sure if I needed to commit the changes that happened after running /bin/python-petstore.sh, I commited them just in case
- [X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.1.x`, `4.0.x`. Default: `master`.
@taxpon @frol @mbohlool @cbornet @kenjones-cisco

### Description of the PR

A few times already, when using Swagger-codegen before also, I was thinking that all the supporting files for Python packages are a lot of the time unecessary. If you're building a package around auto-generated code, you'll be using it as a library and have your own setup.py, etc. 

This PR adds a "onlyPackage" cli options ([...]generate [...] -l python -DonlyPackage [...]) (couldn't think of a better name, possibilities "libraryClient", "pureLibrary", ... IDK)

When this option is activated, 

* docs and tests are generated in subfolders of the package itself
* README.mustache is generated to \<packageName\>_README.md to not overwrite the overarching project's README.md
* all the supporting file at the root of the project (git_push, requirements, setup.py) are not generated


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -198,10 +200,21 @@ public void processOpts() {
             setPackageVersion("1.0.0");
         }
 
+        if (additionalProperties.containsKey(CodegenConstants.ONLYPACKAGE_GENERATION)) {
+            setOnlyPackage(true);
```
kenjones-cisco: As this is a new feature, and is not overriding an inherited functionality. The pattern used for `excludeTests` would seem to be a better match. ie using a local method variable instead of having a class variable with getter/setter.

kenjones-cisco:
Thanks for the PR!
Overall this looks good. Just a minor change is recommended.

kenjones-cisco:
LGTM!

rienafairefr:
@kenjones-cisco you're right, a local variable makes more sense. I've pushed that change up :+1:

rienafairefr:
the fail in the CI is normal I think, something related to the spring-mvc-j8-localdatetime sample

cbornet:
Is it not already possible to generate this wuth the existing selector options ?

rienafairefr:
I usually don't want to fully skip them, just generate them somewhere else. For example, if there is already a docs/ folder in my project, and I auto-generate there, openapi-generator would interfere with it. docs/ inside the package folder is good, same for tests, same for the readme.md. I've tested a lot of setups (ignoring files, copying the files after generation), and I feel like the simplest is having a flag for the generator. All in all, I feel like generating a library instead of a Pypi package should be an option :-)

wing328:
@rienafairefr as discussed, please name the option as "generateSourceCodeOnly" (default: false).

rienafairefr:
OK, I just renamed the option to generateSourceCodeOnly :+1:

wing328:
@rienafairefr thanks for the PR, which has been merged into the master.

rienafairefr:
@wing328 You're welcome :+1:

fizxmike:
FWIW, having the library code in the same repo as your implementation code seems bad pattern (maybe okay for POC). Why not just put the python client in it's own repo the do `pip install git+https://github.com/somebodys/pythonclient` ??

PR #474, opened 2018-07-05
[cli] Don't log to STDOUT if debug flags are set

### LINKED ISSUE #473: [cli] Generator logs to STDOUT when debug options are set
##### Description
Since #207, log messages are written to STDOUT and STDERR respectively based on the level of the message, which makes sense.

However, when debugging the models / operations passed to the templates, it would be nice to be able to pipe the JSON directly into a file / program which could parse it directly.
Currently the piped output will include any of the messages written to STDOUT (which is everything except for errors), which means that it is unparsable.

Latest from `master`

N/A

`java -DdebugModels -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g java -i cellar.json -o /tmp/cellar > generate.json`

##### Steps to reproduce
Pipe the output of any `generate` command with a debug flag set (such as `debugModels` or `debugOperations`)

##### Related issues/PRs
This is in direct response to #207.

See PR #474

devplayer0:
### PR checklist

### Description of the PR
It makes sense that error messages should be written to STDERR and
all others should be written to STDOUT (as shown in #207). However, it
would be convenient to parse the debugging output when the relevant
flags are set.

This change will disable logging to STDOUT and redirect all log messages
to STDERR when any of the debug flags are set. (Resolves #473)

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -34,7 +34,6 @@ import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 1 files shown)


wing328:
@devplayer0 thanks for the PR. Shall we let the users control the behavior using `1>&2` instead so that the users have the option to decide where the log messages go?

devplayer0:
@wing328 Normally that's what I'd do, but the problem is that there's no way to filter out the log output from the JSON when flags like `-DdebugOperations` are set.

With this change, I can do `java -DdebugModels -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g java -i cellar.json -o /tmp/cellar > generate.json` and the contents `generate.json` will be parsed correctly by a JSON viewer, while the log messages will be printed to STDERR.

jimschubert:
I think this is a good solution, but I think it would make more sense to put this behind an opt-in switch on the command line (something like `--debug-to-stdout` or `--debug-print-only`).

I use the debug options often, and while I understand the use case… I would find it annoying to not see the other STDOUT messages when using the debug options.

devplayer0:
@wing328 @jimschubert I have changed the behaviour so that log messages will only be redirected to STDERR if the `--log-to-stderr` flag is set. (Instead of just checking for the presence of the debug options)

For example, see the following commands:
```bash
java -DdebugModels -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g java -i cellar.json -o /tmp/cellar > generate.log
java -DdebugModels -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate --log-to-stderr -g java -i cellar.json -o /tmp/cellar > generate.json
```
The contents of `generate.log` would be all (non-error) log messages alongside the JSON output of the debug options as normal. `generate.json` would contain only the debug options output. (with __all__ log messages being written to STDERR)

jimschubert:
I've removed the breaking change label, as this is now a new option with no breaking change.

modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
```diff
@@ -199,8 +202,23 @@
             description = "Skips the default behavior of validating an input specification.")
     private Boolean skipValidateSpec;
 
+    @Option(name = {"--log-to-stderr"},
```
jimschubert: I think this option is much clearer, and has the benefit of not being a breaking change.

jimschubert:
Updated README via a3e5edc367946111f224af83378f4e91627ea973

wing328:
@devplayer0 thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

PR #492, opened 2018-07-07
[haskell-http-client] regen samples

jonschoning:


wing328:
The shippable failure is not related to the change.

Issue #505, opened 2018-07-08
[Golang][client] Posting file limitations - MIME part name, multiple files, etc.


grokify:
##### Description

(1) MIME part name
Customized `multipart/form-data` `name` parameter. Right now this is hardcoded to `file`. This means APIs that expect a particular part name will not recognize the file.

https://github.com/OpenAPITools/openapi-generator/blob/a14e757059e8db02260f06263675d581dcdc33fb/modules/openapi-generator/src/main/resources/go/client.mustache#L210

(2) Multiple files

Currently, only one file is supported, required or optional given the interface for `APIClient.prepareRequest`

(3) Multiple MIME part names

Ability to have multiple part names for files.

##### openapi-generator version

3.1.0 and `master`

##### OpenAPI declaration file content or url

https://github.com/grokify/go-ringcentral/blob/master/codegen/swagger_spec.yaml

```yaml
          -
            name: "attachment"
            in: "formData"
            description: "File to upload"
            required: false
            type: "file"
```

##### Command line used for generation

Any

##### Steps to reproduce

1. Examine `client.mustache`.
1. Create client and send request.

##### Related issues/PRs

https://github.com/grokify/go-ringcentral/issues/16

##### Suggest a fix/enhancement

Simple fix for existing one file with correct MIME part name.

* Add a parameter `localVarFormFileName` to `APIClient.prepareRequest`
* Set `localVarFormFileName` in every generated API call that calls `APIClient.prepareRequest`
* Set `localVarFormFileName` to part name when present

Consider larger fix to support multiple part names and multiple files per part name.

PR #506, opened 2018-07-08
[Golang][client] fix `formData` `file` parameter name use

grokify:
### PR checklist

@antihax
@bvwells

### Description of the PR

When a parameter in `formData` with type `file` is sent, the parameter name is hardcoded to `file` but it should be set to the parameter name because it needs to be used in the MIME `multipart/form-data` `Content-Disposition` header. As an example, if an API requires the following and the parameter name is set to `attachment` in the spec, `APIClient.prepareRequest` will still set `name` to be hardcoded to `file`.

```
Content-Disposition: form-data; name="attachment"; filename="myfile.pdf"
```

This issue is described in more detail here:

1. https://github.com/OpenAPITools/openapi-generator/issues/505
1. https://github.com/grokify/go-ringcentral/issues/16

`localVarFormFileName` was select as the variable name to hold this info because it aligns with `localVarFormParams` and the name is specifically associated with the file, e.g. `localVarFileName` and `localVarFileBytes`. It is also more inline with Go's `FormFile` in `CreateFormFile`.

This PR solves the immediate issue of not using the parameter name. A future enhancement will be to support an array of files per parameter and multiple file parameters.


modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
```diff
@@ -573,7 +571,6 @@ paths:
       parameters:
         - name: enum_header_string_array
           in: header
-          explode: true
```
wing328: Seems like there's still an issue with the online converter. I'll open an issue tomorrow.

Issue #226, opened 2018-06-05
Add a C client generator


wing328:
##### Description

In certain scenarios in which the C++ generator cannot be used, the developers may need to use a C API client instead to access the REST API.

##### openapi-generator version

Latest master

##### Suggest a fix/enhancement

If anyone wants to work on the new generator for C API client, please reply to let us know

PR #516, opened 2018-07-09
Add C client generator

wing328:
### PR checklist

- [ ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.1.x`, `4.0.x`. Default: `master`.

### Description of the PR

For https://github.com/OpenAPITools/openapi-generator/issues/226

Still WIP

cc @PowerOfCreation


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CClientCodegen.java
```diff
@@ -0,0 +1,700 @@
...
+import org.openapitools.codegen.utils.ModelUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.Map;
+
+public class CClientCodegen extends DefaultCodegen implements CodegenConfig {
```
stkrwork: Wouldn't it be better to name it CCUrlClientCodegen, so that the name reflects the library used as well? maybe in the future we have more than one c client

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CClientCodegen.java
```diff
@@ -0,0 +1,700 @@
...
+    public CClientCodegen() {
+        super();
+
+        // clear import mapping (from default generator) as ruby does not use it
+        // at the moment
+        importMapping.clear();
+
+        modelPackage = "models";
+        apiPackage = "api";
+        outputFolder = "generated-code" + File.separator + "C";
```
stkrwork: maybe generated-code/C-LibCurl

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CClientCodegen.java
wing328: Good idea 👍

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CClientCodegen.java
wing328: Sure

wing328: Thanks for the feedback. I've switched it to `generated-code/C-libcurl`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CClientCodegen.java
wing328: I've renamed it as "CLibcurlClientCodegen.java" to follow the {language}{framework|library}{Client|Server}Codegen pattern.

Ref: https://github.com/curl/curl

samples/client/petstore/C/model/api_response.c
```diff
@@ -19,13 +19,34 @@ api_response_t *api_response_create(
...
 	free(api_response->message)
 
 	free(api_response);
 }
+
+cJSON *api_response_convertToJSON(api_response_t *api_response) {
+	cJSON *item = cJSON_CreateObject();
+	// api_response->code
+	if(cJSON_AddNumberToObject(item, "code", api_response->code) == NULL) {
+    	goto fail;
```
stkrwork: why use goto? Why not insert the "cJSON_Delete(item);" and "return NULL;" in the if statement?

stkrwork: also there seems to be an indentation error here.

PowerOfCreation: What do you want to refactor there? The checks are made at each method that allocated new memory to avoid a Segmentation Fault, they can't be combined. By combining and only doing the checks once you will already have a Segmentation Fault before the combined checks. There could only be written a new method that handles the cJSON_Delete(item); return NULL; and would be used somewhat like this:

if(cJSON_AddNumberToObject(item, "code", api_response->code) == NULL) {
    	return cJSON_deleteJSONObjectAndReturnNull(item);
...
cJSON *cJSON_deleteJSONObjectAndReturnNull(cJSON *item) {
    cJSON_Delete(item);
    return NULL;
}

This is unnecessary complicated and the goto is a lot easier and cleaner to read in my opinion. I really don't see a point to remove the goto's there as they are also used like that in the example code of the library.

wing328: I think we can go with the current implementation for the beta release and consider @stkrwork suggestion as a day-2 requirement for improvement to the auto-generated C code.

etherealjoy: @wing328 
I agree. The community could have the chance to check and add PR to improve.

zhemant:
List of issues which I faced while compiling the sample. Some check are to see if its okay to generate variable like this

- [ variable generation //to check] 

- [file keyvaluepair.h //to check ] 

- [for status enum, how to generate enum in mustache ] 

- [ In imports there is a statement looking for file.h which is not generated] 

- [ tag issue, in api, baseName is 'tags' where as in model baseName is tag due to which when compiling methods in tag.h are not found] 

- [ Major part of Api body is yet to complete e.g. list declaration are there but not sure how to use them]

samples/client/petstore/C/api/UserAPI.c
```diff
@@ -0,0 +1,319 @@
...
+                    "pet",
+                    localVarPath,
+                    localVarQueryParameters,
+                    localVarHeaderParameters,
+                    localVarFormParameters,
+                    localVarBodyParameters,
+                    "GET");
+
+   free(apiClient->dataReceived);
+   free(UsernameString);
```
zhemant: UsernameString is not defined at all

samples/client/petstore/C/api/UserAPI.c
```diff
@@ -0,0 +1,319 @@
...
+   return;
+
+}
+
+// Get user by user name
+//
+user_t *UserAPI_getUserByName(apiClient_t *apiClient, char* Username) {
+    list_t    *localVarQueryParameters,
+    list_t    *localVarHeaderParameters,
+    list_t    *localVarFormParameters,
```
zhemant: How are these used?

wing328: These variables are later passed to`apiClient_invoke`:  https://github.com/OpenAPITools/openapi-generator/pull/516/files#diff-5632e1467c11db52bdf6ca9036c23f3dR34

zhemant: What is the purpose of them?

samples/client/petstore/C/api/UserAPI.c
wing328: I think it should free the input argument "UserName" in https://github.com/OpenAPITools/openapi-generator/pull/516/files/1fd318ce0f99617c906966d8a32ec34e29c08dc5#diff-e8a9562fab1f78d7c89817ed972e01e2R160 instead. Let me submit a fix.

wing328: Fixed via 77e3c3611

PowerOfCreation:
I have tried the generator and I'm getting a few errors when trying to build it.

The errors: https://hastebin.com/zuhifemopu.cs
Our specification file: https://hastebin.com/zesohuyabu.bash

Ignore the php thing, this is only the folder name.

PowerOfCreation:
I'm still facing a lot of errors when compiling the generated code:
Output: https://hastebin.com/doniyezicu.cs
Our specification file: https://hastebin.com/zesohuyabu.bash

zhemant:
I am facing issue with baseName, E.g. in file event_type.c base name is "event_type" but which it is used in other files it comes as "eventType" It seems like issue with mix of camel case and snake case

zhemant:
This is OAS3 modified for petstore example. I have created a schema for enum that how it is generally used in OAS3 and given reference for the enum. It generates files for this schema, but they dont deal with the data of enums.  only empty function are generated. 
https://hastebin.com/rixotiroqe.bash

wing328:
The Travis failure (timeout) seems to be caused by the change in the build image (not by this PR):
```
Changed 62 dependencies!
Precompiling executables...
Precompiled build_runner:graph_inspector.
Precompiled build_runner:build_runner.
[INFO] Generating build script...
[INFO] Generating build script completed, took 280ms
The job exceeded the maximum time limit for jobs, and has been terminated.
```

wing328:
Tweet: https://twitter.com/oas_generator/status/1063488038284587008

Reddit: https://www.reddit.com/r/programming/comments/9xoe7f/welcome_c_client_generator_as_the_100th_generator/

Hacker News: https://news.ycombinator.com/item?id=18470389

wing328:
For discussion on the future enhancement of the C generator, please refer to https://github.com/OpenAPITools/openapi-generator/issues/1473

Issue #538, opened 2018-07-11
[rust-server] how to cope with loss of `file` type?


bjgill:
##### Description

Swagger v2 had three types to represent binary data: `{type: string, format: binary}`, `{type: string, format: byte}`, and `{type: file}`. In OpenAPI v3, we lost the last of these.

In `rust-server`, binary and byte were treated identically - as `swagger::ByteArray`. File was `Box<Future<Item=Option<Box<Stream<Item=Vec<u8>, Error=Error> + Send>>, Error=Error> + Send>`.

I'm not sure what the correct approach for OpenAPI v3 is. Should we keep one of binary/byte as `swagger::ByteArray` and move the other to use the same type as file used to? Do we keep binary/byte as-is and add in some `rust-server`-specific workaround for those wanting to retain the old file behaviour?

##### Related issues/PRs

Discovered in https://github.com/OpenAPITools/openapi-generator/issues/307

bjgill:
Right - I've spoken with some of the people using swagger-codegen at the moment, and it turns out that there are currently people using the `file` type. It's necessary when transferring large files to avoid excessive memory usage.

Broadly, I think there are 3 options:
1. Drop support for `swagger::ByteArray`, and use the stream option for `binary`/`byte`. This has the advantage of simplicity, but is really clunky to use for those who just want a `Vec<u8>`
2. Keep the stream-based mechanism for large files, but hide it behind a format or custom property. Simple to implement. However, given that this choice is completely transparent to peers, the openAPI spec seems like the wrong place to be making this choice.
3. Define a new type/trait for `binary`/`byte` APIs that can be converted to/from both streams and `Vec<u8>`. This gives control over to the person integrating with the autogenerated code. Having a new type/trait will also give us a natural place to document this behaviour.

I personally prefer the 3rd option. However, I'll be talking to a number of other users to get some feedback. I'm expecting to have a design ready for implementation by the end of the week.

bjgill:
I've now had some discussions. This revealed a few things.

1. First, the old `Stream`-based file handling wasn't actually streaming on the server - we buffered the file in its entirety, and then presented the in-memory file as a stream. 
2. Second, there is a rather gnarly design problem server-side. Things are simple for the client - it can just push files out over the wire in whatever format it prefers. However, if there is more than 1 parameter, the server needs to know what order to expect the parameters in to avoid needing to buffer the whole request until it can work out the order for itself. 
[...]

Issue #307, opened 2018-06-13
[rust-server] features broken from swagger-codegen


bjgill:
##### Description

In #290, we got rust-server mostly working in openapi-generator. However, there are still a few missing features. See the commented out sections of https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/rust-server/petstore-with-fake-endpoints-models-for-testing.yaml, but in summary:

- [x] Responses with `schema: {type: <something>}` and >0 headers **(got a fix - https://github.com/OpenAPITools/openapi-generator/pull/549)**
- [x] Enum strings (and arrays of strings) in form data params **(we never supported this; see #545 for the fix for the compilation error)**
- [x] Byte strings in form data parameters **(we never supported this properly; see #538)**
- [x] Binary string properties **(we never supported this properly; see #538)**

I suspect the former might be related to the terrible hack I did in `postProcessModels`, whilst the latter 3 seem likely to be related to `processParam`.

~EDIT: I have no plans to work on this for the next fortnight, so contributions gratefully welcomed.~

bjgill:
Right. I'm planning to pick this up this week.

bjgill:
My plan is to add notes to this issue as I go. I'm likely to be working on this over a few days, and don't want to forget things.

The schema problem is interesting. From my testing, it appears that the problem is any base schema in a response with at least one header defined. both the body and first header get the type of the first header. The type is correct in the generated `openapi.yaml`, so maybe this is a problem with the template? It looks as if we're doing this correctly in client/mod.rs, so presumably I can copy the logic from there.

Hmm. It looks as if the problem is that inside `{{#headers}}{{#-first}}`, `{{{dataType}}}` refers to the type of the first header, rather than the type of the body as a whole. This is the only time we need both the body type and the fact that there are some headers at the same time.

bjgill:
On the binary/byte form params - I don't think this ever worked.

What happens is that if `hasFile` is set, we generate `entry.parse::<{{{dataType}}}>()`, but "the trait `std::str::FromStr` is not implemented for `swagger::ByteArray`". In the old sample, `hasFile` was not set, so we fell through to the other implementation, which was just to set it to the example value.

So:
* if the API `hasFile`, the server code won't compile if the API has binary/byte format parameters (but should work otherwise).
* if the API doesn't `hasFile`, form parameters just don't work server-side (we throw away the request value, using the example value instead).

[...]

PR #547, opened 2018-07-12
[rust-server] Drop `file` support

bjgill:
### PR checklist

### Description of the PR

In swagger v2, we had 'binary', 'byte', and 'file'. OpenAPI v3 only has
the former two. This commit drops the old 'file' handling. This has the
side-effect of removing a half-complete implementation of form parameter handling.

This removes the ability to send files as streams, so will make life
harder for those wishing to send large files without running out of
memory.

See https://github.com/OpenAPITools/openapi-generator/issues/307 and https://github.com/OpenAPITools/openapi-generator/issues/538 for the motivation behind this PR.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
@@ -710,8 +710,6 @@ public class RustServerCodegen extends DefaultCodegen implements CodegenConfig {
                 }
                 header.nameInCamelCase = toModelName(header.baseName);
             }
-
-            // additionalProperties.put("apiHasFile", true);
         }
 
         return objs;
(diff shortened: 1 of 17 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -164,8 +164,8 @@ public RustServerCodegen() {
         typeMapping.put("date", "chrono::DateTime<chrono::Utc>");
         typeMapping.put("DateTime", "chrono::DateTime<chrono::Utc>");
         typeMapping.put("password", "String");
-        typeMapping.put("File", "Box<Stream<Item=Vec<u8>, Error=Error> + Send>");
-        typeMapping.put("file", "Box<Stream<Item=Vec<u8>, Error=Error> + Send>");
+        typeMapping.put("File", "swagger::ByteArray");
+        typeMapping.put("file", "swagger::ByteArray");
```
wing328: For other generators, we're also mapping `file` and `binary` to the same type.

bjgill:
Withdrawing - we need to retain support for transferring large files in some form.

bjgill:
Re-opening as per https://github.com/OpenAPITools/openapi-generator/issues/538#issuecomment-406236888.

The apparent former support for asynchronous file transfer wasn't actually asynchronous (we still buffered the whole file in memory). Thus, we lose nothing by removing the file mechanism now. We can defer the problem of large file transfer until someone actually needs the feature.

wing328:
> We can defer the problem of large file transfer until someone actually needs the feature.

Totally agree.

bjgill:
OK - we've not had any objections. On that basis, I think this is now ready to be merged (modulo anyone wanting to review this?)

wing328:
I'll merge it this weekend if no further question/feedback on this change.

cc @frol @farcaller

Issue #550, opened 2018-07-12
[rust-server] allow Metaswitch to migrate to openapi-generator


bjgill:
##### Description

Once https://github.com/OpenAPITools/openapi-generator/issues/307 is fixed, `rust-server` will be usable in openapi-generator. At @metaswitch, we'd like to migrate across.

Currently, we're working on a branch of swagger-codegen. We need to work on a branch due to having some changes to `html2` we've not yet upstreamed, as well as having added some `rust-server`-specific CI.

Now that Travis is testing `rust-server`, we no longer need to be working in a branch. The work to allow us to migrate across is therefore:

- [x] upstream our `html2` changes: https://github.com/OpenAPITools/openapi-generator/pull/643
- [x] upstream a few contributions that have got stuck waiting for the migration. These enhance the current authentication/context handling: https://github.com/OpenAPITools/openapi-generator/pull/645 and https://github.com/OpenAPITools/openapi-generator/pull/612
- [x] we need reproducible builds, so create our own docker containers. There'll be some internal work to allow us to continue doing this for openapi-generator
- [x] test our internal swagger specs against openapi-generator and raise/fix any issues discovered
- [x] a bunch of internal education

My hope is that once we've finished all that, we'll be able to start contributing directly to the open source repo, rather than having to contribute to our internal fork and then upstream. This should mean that the community gets more visibility into what we're doing.

##### Suggest a fix/enhancement

EDIT: I'm planning to work on this over the next couple of weeks.

jmini:
This sounds like a good plan... Thank you a lot for sharing it.

bjgill:
OK - we're making progress. I'm now waiting on the following PRs:

- [ ] #612
- [ ] #643
- [ ] #645

bjgill:
I've started regression testing of internal specs. Bugs found:

- [ ] #659

bjgill:
OK - we've worked around one regression, and documented the other. Thanks for all the help - I think we're now ready to use openapi-generator.

PR #552, opened 2018-07-12
[rust-server] enhance support for middlewares

bjgill:
### PR checklist

- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @frol @farcaller 

### Description of the PR

This enhances codegen to generate the new `RequestParser` trait, so middlewares can get the operation ID from a hyper `Request`. This PR also tweaks the autogenerated client to be generic over the future type that the underlying hyper Client returns.  This is needed to allow passing wrapped clients - the `FutureResponse` type used originally is private to hyper and can't be constructed.  This change allows us to wrap the hyper client below the autogen layer but above hyper.

I've not done any manual testing of this PR - we've been using this in production for the past month, though, so low risk.

Thanks to @mthebridge, who originally contributed these changes.

Part of the fix for https://github.com/OpenAPITools/openapi-generator/issues/550


wing328:
@bjgill thanks for the PR. Is it correct to say that this PR is a breaking change for the Rust client?

If yes, we should include it in 3.2.x branch instead.

bjgill:
@wing328 - this will only be a breaking change if the user has put the `Client`'s type signature in their code. Otherwise, no changes are needed - e.g. for the example client integration. Any migration should be straightforward - they'll just need to update their type signature to match the new one.

Given that `rust-server` is in the list of generators not yet fully migrated, do we actually need to abide by the stability guarantees implied by the versioning of openapi-generator as a whole? There is going to be at least one more breaking PR before I'm ready to do any serious testing of `rust-server`.

However, if it's not acceptable to keep `rust-server` unstable for the moment, we probably do need to move active development onto the `3.2.x` branch. Do you know what the expected release date for that is?

(Also, I've noticed that we weren't previously checking that the example integration compiled, so have just fixed that in the latest commit.)

wing328:
> Given that rust-server is in the list of generators not yet fully migrated, do we actually need to abide by the stability guarantees implied by the versioning of openapi-generator as a whole? There is going to be at least one more breaking PR before I'm ready to do any serious testing of rust-server.

In the migration guide, I've updated `rust-server` as "migrated". Only the `apex` generator is pending migration and testing at the moment.

wing328:
>  this will only be a breaking change if the user has put the Client's type signature in their code. Otherwise, no changes are needed - e.g. for the example client integration. Any migration should be straightforward - they'll just need to update their type signature to match the new one.

Thanks for the explanation. 

If this PR is ready to merge, please kindly let me know.

bjgill:
Thanks. This is ready to merge.

PR #555, opened 2018-07-13
[Golang][client] Add option for standard Go generated code comment

grokify:
### PR checklist

@antihax 
@bvwells 

### Description of the PR

Update the generated code comment to the following official Go pattern:

`^// Code generated .* DO NOT EDIT\.$`

This allows the auto-generated code to be differentially processed by Go Lint as discussed here:

* https://github.com/golang/go/issues/13560#issuecomment-288457920
* https://medium.com/@dgryski/five-nice-things-for-machine-generated-code-5335e67c1e36

This is also used by [Go Report Card](https://goreportcard.com/report/github.com/grokify/go-ringcentral).


etherealjoy:
will this prevent warning of un-commented Exported symbols?

grokify:
> will this prevent warning of un-commented Exported symbols?

Standard Go behavior is that uncommented, exported symbol warnings will be suppressed. In fact, `golint` will ignore the files. It's debatable to me whether this Go behavior is desirable or not. I personally think comments are a good thing but a number of people have presented cases where it is not desirable here: https://github.com/golang/lint/issues/263.

Perhaps we could add a flag to allow the builder to decide whether they want a recognized, or not-recognized string?

etherealjoy:
> I think it's worthwhile for code humans need to use.
This is what I meant. It is not always ideal to suppress warnings.

grokify:
I just updated the samples so `go-petstore-withXml` sample has `-DwithGoCodegenComment=true`. `go-petstore` continues to have the default behavior. That way users can see both behaviors in the samples.

> > I think it's worthwhile for code humans need to use.
> This is what I meant. It is not always ideal to suppress warnings.

Agree. It seems like there are lots of opinions on this on both sides for machine-generated code. The updated PR supports this feature as a optional flag so people of both persuasions can be supported.

The following looks like simple improvement opportunities for our templates:

```
client/response.go:16:6: exported type APIResponse should have comment or be unexported
client/response.go:33:1: exported function NewAPIResponse should have comment or be unexported
client/response.go:39:1: exported function NewAPIResponseWithError should have comment or be unexported
```

The following feels like it could be a bit more challenging for us to fix:

```
examples/scim_users_get/get_scim_users.go:217:3: var userId should be userID
```

modules/openapi-generator/src/main/resources/go/partial_header.mustache
```diff
@@ -13,5 +13,11 @@
  {{#infoEmail}}
  * Contact: {{{infoEmail}}}
  {{/infoEmail}}
+{{^withGoCodegenComment}}
  * Generated by: OpenAPI Generator (https://openapi-generator.tech)
```
wing328: @grokify thanks for the PR. What about replacing this line directly with the following to avoid adding another option?
```
// Code generated by OpenAPI Generator (https://openapi-generator.tech); DO NOT EDIT.
```

grokify: My original commit f31ecadf24a3e82596020793bb626d65f4736116 was a simple replace, however, in discussion with @etherealjoy, I agreed it's not ideal to remove the original behavior and that the original behavior should be the default. This is because this change trigger's Go's special linter handling for machine generated code and disables `golint` and thus `gometalinter`. The result is that:

1. Warnings for exported symbols without comments will be suppressed
2. Warnings for non-standard MixedCase symbol names will be suppressed

For exported symbol comments, it seems generally better to have comments than not since this are expected to be used by humans. Depending on the specification, this can be the majority of warnings. The cases where people indicated a preference for suppression of these warnings seem to be related to when there are no documentation resources available for a legacy / deprecated API. It doesn't seem like a legacy / deprecated API use case should be the default behavior.

MixedCase is a different issue since sometimes I personally feel MixedCase linter is too strict. I like things like `userId` and `serverUrl` because it makes the symbol name more easily machine parseable, but I recognize the standard in the Go community here.

There is a way to separate this behavior, but that involves having `golint` issue the warnings and then using `gometalinter` to selectively remove the warnings you are not interested in.

So, the thought is to make the comment a non-default option because it will:

1. allow users to see comment errors which should be addressed most of the time
1. allow more customized behavior using a combination of `golint` and `gometalinter`
1. allow a different behavior for scenarios like legacy / deprecated API use cases

wing328: @grokify thanks for the explanation. My feedback was to avoid adding too many options for a generator. I'm ok with this new option given the reason you provided.

Issue #531, opened 2018-07-11
[typescript-fetch] Add support for file (application/octet-stream) response types


macjohnny:
See https://github.com/swagger-api/swagger-codegen/issues/8409

Solution: adapt PR https://github.com/swagger-api/swagger-codegen/pull/8410

Place1:
I've created a fork of typescript-fetch that fixes this issue and a bunch of others such as model property naming etc. I've already mentioned it in #347 but there seems to be a little bit of interest. Would this fork solve the issue for you? I'm keen to gauge interest :)

macjohnny:
@Place1 thanks for the fork. I haven't tried it, yet. 
> but there seems to be a little bit of interest.

I wouldn't say that. I think any improvement is welcome and wanted.
I think it would be good to file a PR here that incorporates all your code, don't you think so @wing328 ?

Place1:
i'll start working on a PR. Something I've done in my fork is make sure the generated client is a single `.ts` file which is quite different to what typescript-fetch currently does. What's the best practice for openapi-generator in this situation? Should I make my PR introduce a new generator to avoid causing a breaking change in the current typescript-fetch generator?

macjohnny:
> single .ts file which is quite different to what typescript-fetch currently does

We recently included a change to https://github.com/OpenAPITools/openapi-generator/pull/363 to split the api up into separate files, as it can cause issues with inheritance if the files are all in the same file and in the wrong order, see https://github.com/OpenAPITools/openapi-generator/issues/350

Place1:
Perhaps the generator could alternatively support hoisting of all parent classes. It partially gives up on alphabetical ordering but there's a benefit in a single file output in terms of simplicity. This suggestion is for typescript-fetch

macjohnny:
@Place1 I personally think that splitting the apis into multiple files does not add complexity, it adds simplicity, but apart from personal preference, if any technical problems can be sorted out, why not.

macjohnny:
maybe @TiFu can give his opinion about this, too

TiFu:
PRs which improve the structure of the generators are always welcome. Especially, the change proposed by @Place1  would make it easier to implement the redesign of all ts-generators, so I'd be happy to review it (which I'll do, later see the PR above).

Background: We want to unify all ts-generators into one generator, so that we don't have to copy bug fixes to 5 other generators. Having a better basis would make this work so much easier.

Place1:
[...]

Issue #347, opened 2018-06-19
Generated client with typescript-fetch generator is very different from swagger-codegen-2.2.3


megaboich:
The problem: When using latest openapi-generator there is a lot of differences in generated client codebase. 

We have quite a big code base generated automatically using swagger-codegen-2.2.3. There are some 35 of different REST controllers and we have build automation script that generates typescript-fetch API clients for them during the build process and in general it works very well and everyone is happy.

But I think in version 3 of swagger codegen there were introduced lots of changes.
The most annoying change is switching from object-structured parameters to plain parameters.
Compare this code which invokes generated api:
**Before**
```
    await this.api.getWorkspacePermissions({
        authorization: this.apiHelper.authToken,
        workspaceId: workspaceId
    })
```

**After**
```
    await this.api.getWorkspacePermissions(this.apiHelper.authToken, workspaceId)
```

I do not like that names of parameters are not presented now. In some cases we have lots of string parameters and here I see several disadvantages in this approach:
1. When I do not see the names then it is not readable anymore.
2. Error could be easily added when I put them in wrong order.
3. Also if someone at backend decide to change parameter name and order (like remove something called `projectId` and add different required parameter called `parentObjectId`) then new generated client will still be compatible because there is just a bunch of string parameters in API call without information about the name.

Any ideas why this change was introduced? Probably I just do not see the benefits from it?

Place1:
I've created a fork of the typescript-fetch generator over at https://github.com/Place1/swagger-codegen-typescript-browser

I'd be very interested in using it as the foundation for a rework of typescript-fetch over here. It follows the params object as api method argument and fixes a bunch of other issues present in the generator related to form data and model naming conversions (camelCase), additionally, it supports middleware.

macjohnny:
@Place1 thanks for sharing!
@TiFu could you have a look at this?

TiFu:
Ah that's a very good idea. I'll look into it.

I didn't have any time to start working on it yet (due to another project running sideways) but it should become more relaxed now.

macjohnny:
@TiFu if the parameters of the api methods are to be combined into a single object (maybe as an option), https://github.com/swagger-api/swagger-codegen/issues/8435 of @sschoeb could also be considered. maybe @sschoeb would want to help?

aaronbeall:
[...]

PR #162, opened 2018-05-27
WIP (feedback requested): [gradle-plugin] Initial commit

jimschubert:
### PR checklist

- [ ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the correct branch: Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

See #141 

This is an initial implementation of a gradle-plugin for OpenAPI Generator. I've opened this PR to solicit feedback. Please see the initial commit of README.adoc for details and examples.

To test this out, you'll first need to build the generator:

```
mvn clean package install
```

Then, you'll want to build and publish the gradle plugin locally (update `inputSpec` references to your own locally available specs):

```
cd gradle
./gradlew build
./gradlew publishToMavenLocal
```

From here, you can create a simple `build.gradle` and execute tasks:

```
buildscript {
  repositories {
   mavenLocal()
   mavenCentral()
    maven {
        url "https://plugins.gradle.org/m2/"
    }
  }
  dependencies {
    classpath "org.openapitools:openapi-generator-gradle-plugin:3.0.0-SNAPSHOT"
  }
}

apply plugin: 'org.openapi.generator'

openApiMeta {
   generatorName = "Jim"
   packageName = "us.jimschubert.example"
}

openApiValidate {
   inputSpec = "/Users/jim/projects/openapi-generator/modules/openapi-generator/src/test/resources/3_0/petstore.yaml"
}

openApiGenerate {
    generatorName = "kotlin"
    inputSpec = "$rootDir/specs/petstore-v3.0.yaml".toString()
    outputDir = "$buildDir/generated".toString()
    apiPackage = "org.openapi.example.api"
    invokerPackage = "org.openapi.example.invoker"
    modelPackage = "org.openapi.example.model"
    modelFilesConstrainedTo = [
            "Error"
    ]
    configOptions = [
        dateLibrary: "java8"
    ]
}
```

[...]

PR #145, opened 2018-05-24
Multiple enhancements to typescript fetch generator 

JFCote:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01)

That is a copy from my own PR in https://github.com/swagger-api/swagger-codegen/pull/7914 that has never been merged. @wing328 ask me to redo it here

PR #569, opened 2018-07-15
Typescript fetch fork

Place1:
### PR checklist

### Description of the PR
This PR applies a fork of typescript-fetch that I originally created here: https://github.com/place1/swagger-codegen-typescript-browser

The notable differences to the original typescript-fetch generator are outlines in the [README.md](https://github.com/Place1/swagger-codegen-typescript-browser/blob/master/README.md)

The changes in this fork also address the following open issues on this repo:
- #531
- #347

### TODO:
Confirm the this PR fixes the following known issues:
- [x] swagger-api/swagger-codegen#8410
      ensure that application/octet-stream responses come back as a blob
- [x] swagger-api/swagger-codegen#2776
      ensure that dates are deserialized
- [x] swagger-api/swagger-codegen#7801
      ensure that date params and date-array params are serialized

Add support for:
- [x] swagger-api/swagger-codegen#7828
      add method overload to allow caller to receive native Response
- [x] break models and operations out into multiple files

General improvements:
- [x] Find a better implementation for `modelPropertyNaming`.
       The current runtime conversion has known edge cases. Only a perfect implementation (i.e. based on the spec itself) should be acceptable for an option like modelPropertyNaming to exist.
- [x] improve middleware api interface. use a single options object rather than many positional arguments


Place1:
Sorry for messing up the first PR. I use gitlab all day at work and they have the source/target selection the other way around :rofl:

Place1:
@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01)

modules/openapi-generator/src/main/resources/typescript-fetch/licenseInfo.mustache
```diff
@@ -5,7 +5,7 @@
  * {{#version}}OpenAPI spec version: {{{version}}}{{/version}}
  * {{#infoEmail}}Contact: {{{infoEmail}}}{{/infoEmail}}
  *
- * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
- * https://openapi-generator.tech
+ * NOTE: This class is auto generated by the swagger code generator program.
+ * https://github.com/swagger-api/swagger-codegen.git
```
TiFu: Could you please revert this change?

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
```diff
@@ -1,12 +1,8 @@
...
- * {{{description}}}
- * @export
- * @enum {string}
- */
+{{#description}}// {{description}}{{/description}}
 export enum {{classname}} {
 {{#allowableValues}}
 {{#enumVars}}
-    {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
+    {{{name}}} = {{{value}}}{{^-last}},{{/-last}} {{#description}}// {{description}}{{/description}}
```
TiFu: Is there a reason why `<any>` was removed? 

This was not removed in `modelGeneric.mustache` (see below)

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
```diff
@@ -1,32 +0,0 @@
-{
```
TiFu: I think we should still have a `package.json` and `tsconfig.json`

modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
```diff
@@ -1,38 +1,20 @@
-/**
- * {{{description}}}
- * @export
- * @interface {{classname}}
- */
```
TiFu: I think this comment section should remain instead of the new one.

modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
```diff
@@ -1,38 +1,20 @@
...
 {{#additionalPropertiesType}}
     [key: string]: {{{additionalPropertiesType}}}{{#hasVars}} | any{{/hasVars}};
-
 {{/additionalPropertiesType}}
 {{#vars}}
-    /**
-     * {{{description}}}
-     * @type {{=<% %>=}}{<%&datatype%>}<%={{ }}=%>
-     * @memberof {{classname}}
-     */
```
TiFu: See above - i'd prefer this comment over the new one

modules/openapi-generator/src/main/resources/typescript-fetch/licenseInfo.mustache
Place1: sure thing

modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
Place1: sure thing

modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
Place1: sure thing

modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -1,364 +1,442 @@
...
-    {{/required}}
-    {{/allParams}}
-            const localVarPath = `{{{path}}}`{{#pathParams}}
-                .replace(`{${"{{baseName}}"}}`, encodeURIComponent(String({{paramName}}))){{/pathParams}};
-            const localVarUrlObj = url.parse(localVarPath, true);
-            const localVarRequestOptions = Object.assign({ method: '{{httpMethod}}' }, options);
-            const localVarHeaderParameter = {} as any;
-            const localVarQueryParameter = {} as any;
-    {{#hasFormParams}}
-            const localVarFormParams = new url.URLSearchParams();
```
Place1: @TiFu my fork does away with the dependency on the URL module and uses browser native APIs such as `FormData` and `encodeURIComponent`.

This solves issues such as https://github.com/swagger-api/swagger-codegen/issues/6403

From memory some users encountered issues with `@types` definitions for isomorphic fetch as well, but I can't find the bug reports :/

modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -1,364 +1,442 @@
...
+            body: requestParameters.{{paramName}},
+            {{/bodyParam}}
+            {{/hasBodyParam}}
+            {{#hasFormParams}}
+            body: formData,
+            {{/hasFormParams}}
+            {{#returnType}}
+            responseType: this.getResponseType('{{{returnType}}}'),
+            {{/returnType}}
+            modelPropertyNaming: '{{modelPropertyNaming}}',
```
Place1: @TiFu We use the `modelPropertyNaming` flag to do response model property renaming. Currently it uses a runtime method to do the conversion but a better implementation would modify this to use a property name map.

The existing codegen, using the default `camelCase` for model property naming just causes broken code for clients. This is a major barrier for new users IMO when their first experience is a generator results in a broken default.

see: https://github.com/swagger-api/swagger-codegen/issues/3818

modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -1,364 +1,442 @@
...
+            {{#hasBodyParam}}
+            {{#bodyParam}}
+            body: requestParameters.{{paramName}},
+            {{/bodyParam}}
+            {{/hasBodyParam}}
+            {{#hasFormParams}}
+            body: formData,
+            {{/hasFormParams}}
+            {{#returnType}}
+            responseType: this.getResponseType('{{{returnType}}}'),
```
Place1: @TiFu this fork uses the response type of the request to support handling "string" response types. The original generator just calls `response.json()` which crashes when the response is plain text.

I'm sure this issue has been recoded in the original swagger codegen issue tracker but someone picked it up on my fork as well: https://github.com/Place1/swagger-codegen-typescript-browser/issues/2

modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -1,364 +1,442 @@
...
-        if (configuration) {
-            this.configuration = configuration;
-            this.basePath = configuration.basePath || this.basePath;
+    private middleware: Middleware[];
+
+    constructor(protected configuration = new Configuration()) {
+        this.middleware = configuration.middleware;
+    }
+
+    withMiddleware<T extends BaseAPI>(this: T, ...middlewares: Middleware[]) {
```
Place1: @TiFu our generator supports middleware. This is a very useful feature and something that most HTTP clients contain. My colleagues have been able to make good use of this feature across different projects to implement logging, automatic request retries, timeouts, circuit breakers, automatic token refresh logic, automatic logout logic, complex authentication/authorization challenges that are custom or out-of-scope for the built in security definitions, etc.

modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -1,364 +1,442 @@
...
- * @interface FetchAPI
- */
-export interface FetchAPI {
-    (url: string, init?: any): Promise<Response>;
+{{#apiInfo}}
+{{#apis}}
+{{#operations}}
+{{#operation}}
+{{#allParams.0}}
+export interface {{operationIdCamelCase}}Request {
```
Place1: @TiFu this fork also generates interfaces for all API parameters. This is very helpful as a consumer because it allows you to reference them in your own methods/variables. The original generator uses inline type signatures which aren't reusable.

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
Place1: No I think I must have just sweept it away when changeing things. I can't see a reason for it though, enum values should be true to what they really are without type erasure.

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
Place1: @TiFu I'm happy to put these back in if required. I removed them because i've found that there's a general level of simplicity gained from just having a single file output. But perhaps with [typescript project references](https://blogs.msdn.microsoft.com/typescript/2018/07/12/announcing-typescript-3-0-rc/#project-references) coming in TS 3.0 it might be worth while to have.

TiFu:
Oh thanks for the write up! No idea how I missed these changes earlier. Gonna do an in-depth review tonight.

Place1:
@macjohnny hey i've had a look at that PR, my changes here implement all the changes over there so I think it's covered (hopefully I didn't miss anything).

macjohnny:
@Place1 great, so it should work out

TiFu:
Alright the PR looks good to me (and I realized why I missed all those changes - `api.mustache` was hidden by default and I overlooked it). Seems to need some fixing so that the CI tests pass (see below). 

Shippable complains that the samples aren't up to date:
```
 Perform git status
HEAD detached from FETCH_HEAD

	modified:   samples/client/petstore/typescript-fetch/builds/default/api.ts
	modified:   samples/client/petstore/typescript-fetch/builds/es6-target/api.ts
	modified:   samples/client/petstore/typescript-fetch/builds/with-interfaces/api.ts
	modified:   samples/client/petstore/typescript-fetch/builds/with-npm-version/api.ts

Please run 'bin/utils/ensure-up-to-date' locally and commit changes (UNCOMMITTED CHANGES ERROR)
```
Appveyor seems to be missing a README.mustache?

```
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found, returning only the first one
[main] INFO  o.o.codegen.AbstractGenerator - writing file C:\projects\openapi-generator\samples\client\petstore\typescript-fetch\builds\es6-target\api.ts
[main] ERROR o.o.codegen.AbstractGenerator - typescript-fetch\README.mustache (The system cannot find the path specified)
[main] ERROR o.o.codegen.AbstractGenerator - typescript-fetch\README.mustache (The system cannot find the path specified)
[main] ERROR o.o.codegen.AbstractGenerator - can't load template typescript-fetch\README.mustache
[main] ERROR o.o.codegen.AbstractGenerator - can't load template typescript-fetch\README.mustache
Exception in thread "main" java.lang.RuntimeException: Could not generate supporting file 'SupportingFile:
	templateFile: README.mustache
	folder: 
	destinationFilename: README.md
'
	at org.openapitools.codegen.DefaultGenerator.generateSupportingFiles(DefaultGenerator.java:656)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:797)
	at org.openapitools.codegen.cmd.Generate.run(Generate.java:315)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:59)
Caused by: java.lang.RuntimeException: can't load template typescript-fetch\README.mustache
	at org.openapitools.codegen.AbstractGenerator.readTemplate(AbstractGenerator.java:68)
	at org.openapitools.codegen.DefaultGenerator.generateSupportingFiles(DefaultGenerator.java:623)
	... 3 more
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=2g; support was removed in 8.0
```
travis complains that the code doesn't compile

```
> @swagger/typescript-fetch-petstore@1.0.0 build /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version
> tsc --outDir dist/
index.ts(16,1): error TS2308: Module "./api" has already exported a member named 'Configuration'. Consider explicitly re-exporting to resolve the ambiguity.
index.ts(16,1): error TS2308: Module "./api" has already exported a member named 'ConfigurationParameters'. Consider explicitly re-exporting to resolve the ambiguity.
npm ERR! code ELIFECYCLE
```

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
TiFu: `package.json` just makes it simpler to include in packages just as a dependency.

`tsconfig.json` should definitely be included, e.g. specify EcmaScritp version.

macjohnny:
@Place1 could you please update this PR and resolve the conflicts due to #145 ?

macjohnny:
@emreavsar the additional headers in you are asking for in https://github.com/swagger-api/swagger-codegen/issues/8444 could be solved by using this changed typescript-fetch generator and provide an appropriate middleware, i.e. an interceptor that sets your headers.

Place1:
If anyone else has time i'd be great to hear some feedback for the API around middleware. One concern I have is that there's too many positional arguments:
```
export interface Middleware {
    pre?(fetch: FetchAPI, url: string, init: RequestInit): Promise<FetchParams | undefined | void>;
    post?(fetch: FetchAPI, url: string, init: RequestInit, response: Response): Promise<Response | undefined | void>;
}
```
Does anyone have an opinion on a better API?

TiFu:
Maybe try to move the positional arguments into a separate class or interface? 
e.g. for `pre`

```
// (i'm not very happy with this interface name, maybe you can think of a better one?)
interface Request {
      fetch: FetchAPI;
      url: string;
      init: RequestInit;
}
```

and then the same thing for `post`.

jmini:
> Shippable — Run 1769 status is FAILED. 

I think this was fixed. Please merge `master` into this PR.

> continuous-integration/travis-ci/pr — The Travis CI build failed

https://travis-ci.org/OpenAPITools/openapi-generator/builds/408467081
There I have the feeling that some of the Typescript modules have changed... Let me see if I discover something.

jmini:
I have merged master into this PR.
I have also restored the `pom.xml` in the `samples/` project. This might appear wired, but the CI need those pom to run the typescript sample. I have also [documented](https://github.com/OpenAPITools/openapi-generator/wiki/Samples-folder#samplesci-folder) the approach I follow with the `sample.ci` folder.

Place1:
Cheers @jmini

jmini:
> Shippable — Run 2134 status is FAILED 

This is probably related to #779 (at least it seems to be related).
Maybe a merge from `master` branch into this PR will solve the problem?

> continuous-integration/travis-ci/pr — The Travis CI build could not complete...

Build 414786868

This seems to be related to C++ stuff, but last time I did something on this PR, there was a webpack issue in the travis CI.

jmini:
I have merged master into this PR, I also have reworked my `pom.xml` files in `samples.ci`. I hope this will fixed the Shippable build.

jmini:
Ok last travis build ([#2042](https://travis-ci.org/OpenAPITools/openapi-generator/builds/415151271)) have the webpack problem that needs to be fixed:

```
Output filename not configured.
npm ERR! code ELIFECYCLE
npm ERR! errno 255
npm ERR! typescript-fetch-test@1.0.0 webpack: `webpack`
npm ERR! Exit status 255
npm ERR! 
npm ERR! Failed at the typescript-fetch-test@1.0.0 webpack script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Place1:
@jmini I'm not sure what the webpack issue is to be honest

Place1:
@TiFu other than the webpack CI issue, is there anything that needs to be done on this PR to get it moving forward?

jmini:
@Place1:

I try to help you with the CI, but my only typescript experience is an hello-world and I never configure webpack myself (setup is made by one of my colleague). So it is really hard to help you.

I have merged `master` in your PR.
I have also restored `samples/client/petstore/typescript-fetch/tests/default`.

Let see how this works.

---

I hope members of the typescript team like @TiFu or @macjohnny  will be able to judge the content of your PR.

In all cases it can not be merged if it breaks the CI.

jmini:
> continuous-integration/travis-ci/pr — The Travis CI build failed 

I have the feeling that now the error in build [`2147`](https://travis-ci.org/OpenAPITools/openapi-generator/builds/417294328) is a real compile error:

```
> typescript-fetch-test@1.0.0 build /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default
> tsc
test/PetApi.ts(18,24): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(23,24): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(29,24): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(31,28): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(32,32): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(40,24): error TS2554: Expected 1 arguments, but got 2.
test/PetApi.ts(44,24): error TS2554: Expected 1 arguments, but got 2.
test/PetApiFactory.ts(2,10): error TS2305: Module '"/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version/dist/index"' has no exported member 'PetApiFactory'.
test/PetApiFactory.ts(9,5): error TS2322: Type '"foobar"' is not assignable to type '(name: string, scopes?: string[]) => string'.
test/StoreApi.ts(16,24): error TS2554: Expected 0 arguments, but got 1.
test/StoreApiFactory.ts(2,10): error TS2305: Module '"/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version/dist/index"' has no exported member 'StoreApiFactory'.
test/StoreApiFactory.ts(9,5): error TS2322: Type '"foobar"' is not assignable to type '(name: string, scopes?: string[]) => string'.
```

@Place1 : why did you remove the tests in the `samples/client/petstore/typescript-fetch/tests/default`, I have the feeling that they need to be updated to reflect your changes.

TiFu:
@Place1 CI passing is definitely required. We also always appreciate additional test cases if you have the time to add some.

Place1:
Is anyone able to give me a hand with the browserify tests? I'm not familiar with browserify. @jmini do you have any idea what the CI error means: `ParseError: 'import' and 'export' may appear only with 'sourceType: module'`?

macjohnny:
> ParseError: 'import' and 'export' may appear only with 'sourceType: module'?

@Place1 this seems to be due to the module resolution setting in the `tsconfig.json`, see https://www.typescriptlang.org/docs/handbook/module-resolution.html

wing328:
@Place1 I would also like to point out that your commit (as shown in the Commits tab) is not linked to your Github account, which means this PR won't count as your contribution in https://github.com/OpenAPITools/openapi-generator/graphs/contributors.

Place1:
CI is passing now :)

wing328:
@Place1 👍 

@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig please have another look. If no further feedback, I'll merge it into master on coming Saturday so that it will be included in the 3.3.0 minor release scheduled on Oct 1st.

modules/openapi-generator-gradle-plugin/bin/main/org/openapitools/generator/gradle/plugin/extensions/OpenApiGeneratorGenerateExtension.kt
```diff
@@ -0,0 +1,286 @@
...
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.listProperty
+import org.gradle.kotlin.dsl.property
+
+/**
+ * Gradle project level extension object definition for the generate task
+ *
+ * @author Jim Schubert
+ */
+open class OpenApiGeneratorGenerateExtension(project: Project) {
```
macjohnny: why is this file part of this PR?

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
```diff
@@ -2,27 +2,15 @@
...
     "build": "tsc --outDir dist/",
     "prepublishOnly": "npm run build"
   },
-  "dependencies": {
-    "portable-fetch": "^3.0.0"
-  },
   "devDependencies": {
-    "@types/node": "^8.0.9",
-    "typescript": "^2.4"
+    "typescript": "^2.9"
```
macjohnny: is typescript 2.9 really required?

macjohnny:
@Place1 the method signatures should be backwards-compatible, right?

modules/openapi-generator-gradle-plugin/bin/main/org/openapitools/generator/gradle/plugin/extensions/OpenApiGeneratorGenerateExtension.kt
Place1: no idea. should i revert it? my git blame is telling me it was added 4 months ago by Jim Schubert in #162

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
Place1: good question. probably not. i'll revert it.

Place1:
> 
> 
> @Place1 the method signatures should be backwards-compatible, right?

No. The this PR is named as a "fork" because it's a rewrite and has breaking changes in it.

From the original post: "The notable differences to the original typescript-fetch generator are outlines in the README.md"

Here's a summary:
- only generates class based APIs (no functional programming APIs as the original generator called them)
- usage of FormData to support multipart/form-data (i.e. file upload) requests
  - this is a breaking change to NodeJS users who will need to polyfill FormData
- removal of 3rd party dependencies including isomorphic-fetch
  - nodejs users will have to install isomorphic-fetch or node-fetch (or other) themselves and pass it like follows: `new Configuration({ fetchApi: require('node-fetch') })` [see example from tests](https://github.com/OpenAPITools/openapi-generator/blob/d770f24ff57a37c8ffbe746dacd4d322399efb15/samples/client/petstore/typescript-fetch/tests/default/configuration.ts)
- changed the API operation signature to address #347
  - essentially 2.2.3 had operations with 1 param signatures e.g. `listPets({ name: 'a', age: 1 })` and that parameter was an anonymous type declaration (not an interface).
  - in 3.x.x this was broken and we had operations with multiple param signatures like this `listPets(name: string, age: number)` which meant some operations has really big signatures.
  - in this fork we go back to the 2.2.3 style of having 1 param, but we go a step better and actually export it as an interface i.e:
    ```
    export interface ListPetRequest {
      name: string;
      age: number;
    }
    // ...
    listPet(requestParameters: ListPetRequest)
    ```
  - this should not be a breaking change to 2.2.3 users, although there are probably other changes made in this fork or the original 3.x.x changes that cause them breaks
- implementing model property naming conversions may be considered a breaking change
  - all previous versions implemented the `modelPropertyNaming` configuration option but there was no runtime implementation.
  - this lead to runtime errors if the API/generated code didn't having matching property names (a likely result)
  - this fork correctly implements modelProperyNaming, including runtime support, but as a result of this, if there are properties that the API accepts/returns that aren't documented in the swagger spec then the generated code will not pass them through. This is a result of an implementation detail that you can see in [the generated sample](https://github.com/OpenAPITools/openapi-generator/blob/d770f24ff57a37c8ffbe746dacd4d322399efb15/samples/client/petstore/typescript-fetch/builds/default/models/Pet.ts)
    - the PetFromJSON and PetToJSON methods do the naming conversions
    - from [the generator implementation](https://github.com/OpenAPITools/openapi-generator/blob/d770f24ff57a37c8ffbe746dacd4d322399efb15/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache) you can see that we define the conversion by reading the spec - this is important because only the spec knows the correct original name - but as a result of this implementation we don't copy unknown/undocumented properties from the to/from the API side.

Place1:
@macjohnny My company has been using this fork on 4 projects since i first created it over at https://github.com/place1/swagger-codegen-typescript-browser

3 of these projects were originally on swagger-generator 2.2.3 and 2.0.1 and were updated for the purpose of testing this fork in prod (we found and fixed a few bugs this way 👍). The other project was green field; as a result I have confidence in the lib but i'd love others to test it out on their projects because there may be some edge cases i've missed.

Place1:
apparently travis isn't happy with typescript-node now? what's with that

macjohnny:
@Place1 thanks for the excellent change summary! I think this change greatly enhances the typescript fetch generator.

If I could wish for one more thing: how about having an option to choose between
```typescript
fooOperation(fooId: FooId, otherParam: OtherParamType) {
```
and
```typescript
fooOperation({fooId, otherParam}: {fooId: FooId, otherParam: OtherParamType}) {
```
as suggested in https://github.com/OpenAPITools/openapi-generator/issues/865#issuecomment-415082772? maybe @cspotcode would want to help implementing this?
I think it should be as easy as adding a configuration option and introduce a condition in the template to decide whether to add `{` and `}` or not and separating the types. So 3.x.x users would face less breaking changes.

macjohnny:
@Place1 
> apparently travis isn't happy with typescript-node now? what's with that

Since this is the only error occuring and typescript node was not changed here, it should not matter in this PR.

Place1:
@macjohnny I think the following idea from #865 would be a quick and easy improvement

> When none of the params are required, then append a default value so that the option-bag can be omitted entirely:

The other idea, to add the `CollectParameters` option imo doesn't provide any advantages and additional configuration options add to the feature set to test/maintain.

@TiFu followed up the comment you linked by saying: 

> Another option would be to create interfaces for these parameter bags

and this is what this PR introduces.

The reason for using a single param bag rather than positional args is because it's not unusual for an operation to have many optional arguments leading to some pretty lame to use code like:
```
listPet('name', undefined, undefined, 'some-param-i-care-about', undefined, undefined, undefined, 10, 100)
```
where the hypothetical param bag version would look like
```
listPet({
  filter: 'name',
  search: 'some-param-i-care-about',
  page: 10,
  pageSize: 100,
});
```

Additionally, positional arguments throw some long term maintenance headaches into the mix. Given some params such as:
```yaml
- name: filter
  in: query
  required: true
  schema:
    type: string
```
we'd get the following operation
```
myOperation(filter?: string)
```
now the API is updated and they support a new parameter
```
- name: newParam
  in: query
  required: false
  schema:
    type: string
- name: filter
  in: query
  required: false
  schema:
    type: string
```
and we end up with
```
myOperation(newFilter?: string, filter?: string)
```
and you can probably see the issue here. The spec maintainers may not realize there's a dependency on the order of parameters in the spec (and why should they) and the client after regenerating the library will have code that compiles and runs but results in logical errors if they forget to update their calls to the correct `myOperation(undefined, 'filter')`. There are other issues along these lines that cause issues even if we sort the params or try other smarts in the generator.

I'm happy if others disagree with me and would like to show me some of the advantages. The only one that comes to my mind is the ability to do currying or partial applications more easily, but these can still be done with the param bag albeit you have to curry/pa the param bag itself and make use of property omission, something an FP library might not ship with out-of-the-box.

macjohnny:
@Place1 I absolutely agree with you, my only concern is backwards-compatibility. Since we could provide it rather easily here, I think we should do it.

Place1:
@macjohnny I agree, I'm worries about backwards compat as well. From personal experience, I found it a little frustrating when there were breaks between 2.0.1, 2.2.3 and 3.x.x :( I don't want to cause issues for people

wing328:
Travis CI reports the following errors:
```
> @swagger/typescript-fetch-petstore@1.0.0 test /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/es6-target
> echo 'Error: no test specified'

Error: no test specified
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN @swagger/typescript-fetch-petstore@1.0.0 No repository field.
npm WARN @swagger/typescript-fetch-petstore@1.0.0 No license field.

added 1 package from 1 contributor and audited 1 package in 0.844s
found 0 vulnerabilities


> @swagger/typescript-fetch-petstore@1.0.0 build /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version
> tsc --outDir dist/


> @swagger/typescript-fetch-petstore@1.0.0 test /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version
> echo 'Error: no test specified'

Error: no test specified
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/duplexer3-a19ffb2b/package.json'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/duplexer3-a19ffb2b/README.md'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/duplexer3-a19ffb2b/index.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/duplexer3-a19ffb2b/LICENSE.md'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/create-hash.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/create-hmac.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/index.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/md5.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/rng.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/helpers.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/pbkdf2.js'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/crypto-browserify-d10f92d6/.travis.yml'
npm WARN tar ENOENT: no such file or directory, open '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/typescript-91f70a85/lib/typescript.js'
npm ERR! path /home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/@swagger/typescript-fetch-petstore-c75f09ff/node_modules/ansi-styles
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall rename
npm ERR! enoent ENOENT: no such file or directory, rename '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/@swagger/typescript-fetch-petstore-c75f09ff/node_modules/ansi-styles' -> '/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/tests/default/node_modules/.staging/ansi-styles-bfd5c2c6'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2018-09-29T04_09_26_098Z-debug.log
[[1;31mERROR[m] Failed to execute goal [32morg.codehaus.mojo:exec-maven-plugin:1.2.1:exec[m [1m(npm-install)[m on project [36mTypeScriptFetchPestoreClientTests[m: [1;31mCommand execution failed.[m: Process exited with an error: 254 (Exit value: 254) -> [1m[Help 1][m
[[1;31mERROR[m] 
```

cspotcode:
Regarding option-bag parameters and the `CollectParameters` option:

For backwards compatibility, we need an option for users to opt-in or opt-out of the option-bag. 
 I proposed `CollectParameters` because it seems to have some traction already.  Even though it's a third-party extension, I've seen it pop up in OpenAPI specs at work.  Although it's an additional configuration option, it only needs to be set once at the root of the spec file.  (or you can override it for individual operations)

```yaml
swagger: 2.0
info:
  x-codegen-settings:
    collectParameters: false
```

[APIMATIC has a bunch of others.](https://blog.apimatic.io/swagger-2-0-extension-for-code-generation-settings-abf602c3869d)  I'm not saying we need to match them, but if they've already picked names, why not use them?

The alternative is to require opting out of option-bag.

I agree that 9 times out of 10, option-bags are the way to go.  However, for API endpoints like "getUser" with only a single argument, some people may prefer to use positional arguments.  I agree that this carries the risks you mentioned, if additional arguments are added later, they may break the function signature.

wing328:
@Place1 I suggest adding `CollectParameters` support in another PR instead.

Did you have a chance to review the issues reported by the CI? Did the tests work for you locally?

Place1:
Ok CI is now passing. I'm working on a commit for the last check list item `swagger-api/swagger-codegen#7828
add method overload to allow caller to receive native Response`. It should be done by tonight. Tomorrow i'll give it all 1 last big test on a few major projects and then i'm happy for it to be merged (assuming no other changes are required) :D

samples/client/petstore/typescript-fetch/builds/default/apis/PetApi.ts
```diff
@@ -0,0 +1,396 @@
...
+     */
+    async deletePet(requestParameters: DeletePetRequest): Promise<void> {
+        await this.deletePetRaw(requestParameters);
+    }
+
+    /**
+     * Multiple status values can be provided with comma separated strings
+     * Finds Pets by status
+     */
+    async findPetsByStatusRaw(requestParameters: FindPetsByStatusRequest): Promise<runtime.ApiResponse<Array<Pet>>> {
```
Place1: i've added support for retrieving the raw API response with the typed JSON value.

samples/client/petstore/typescript-fetch/builds/default/apis/PetApi.ts
```diff
@@ -0,0 +1,396 @@
...
+        });
+
+        return new runtime.JSONApiResponse(response, (jsonValue) => jsonValue.map(PetFromJSON));
+    }
+
+    /**
+     * Multiple status values can be provided with comma separated strings
+     * Finds Pets by status
+     */
+    async findPetsByStatus(requestParameters: FindPetsByStatusRequest): Promise<Array<Pet>> {
```
Place1: the non-raw version of the operation also exists, it just unwraps the raw ApiResponse for convenience

samples/client/petstore/typescript-fetch/builds/default/apis/PetApi.ts
```diff
@@ -0,0 +1,396 @@
...
+export interface UpdatePetWithFormRequest {
+    petId: number;
+    name?: string;
+    status?: string;
+}
+
+export interface UploadFileRequest {
+    petId: number;
+    additionalMetadata?: string;
+    file?: Blob;
```
Place1: support for blobs has been added. the pet store example doesn't include any blob responses, but binary body params and binary responses now work; using the following spec for testing: `modules/openapi-generator/src/test/resources/2_0/binaryDataTest.json`

Place1:
@wing328 any updates?

wing328:
If no further question/feedback from anyone, I'll merge this PR into 4.0.x branch (breaking changes without fallback) this coming weekend.

wing328:
@Place1 thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

JFCote:
@Place1 @TiFu @wing328  I'm not sure I understand this new fetch generator. It is breaking my entire code and I might not be able to upgrade to latest version of OpenApi Generator because of this.

It seems all the signature have changed.

For exemple, before this, my function `addAffiliate` would simply needs a `AffiliateCore` object in the body. **It is defined that way in the spec.**

With the new generator, I need to pass a `AddAffiliateRequest` object type.

The new interface is defined that way:

```
export interface AddAffiliateRequest {
    newAffiliate: AffiliateCore;
}
```

It looks to me like a wrapper. But by doing this, we do not respect the API spec and new object type needs to be passed. Objects that the user of the API / Client might not even know.

I suggest we rollback this merge completely or that we remove the part that create new object type out of thin air. Unless there is a way or an option in the generator to use the real method signature but it doesn't seems that way in the generator code.

mhidalgop:
> @emreavsar the additional headers in you are asking for in [swagger-api/swagger-codegen#8444](https://github.com/swagger-api/swagger-codegen/issues/8444) could be solved by using this changed typescript-fetch generator and provide an appropriate middleware, i.e. an interceptor that sets your headers.

How could I do this??? I need to send the version API into the header for all request.

macjohnny:
@mhidalgop see e.g. https://github.com/OpenAPITools/openapi-generator/blob/fc4563baef8a48cd6084f39279f3d9404fe32681/samples/client/petstore/typescript-fetch/builds/with-npm-version/src/runtime.ts#L37-L40

PR #580, opened 2018-07-16
[cpp-restsdk] Support external supporting files

Peaches491:
### PR checklist

### Description of the PR

Currently, all executions of `openapi-generator` produce a set of `supportingFiles` alongside the `model/` and `api/` files. If you have more than one `openapi-generator` spec target, these `supportingFiles` are completely identical, except for the namespace in which they are generated. 

This PR allows users to specify `supportingFilesDirectory` and `supportingFilesNamespace`, the two parameters necessary to point the generated `model/*` and `api/*` files to an external set of `supportingFiles`.

This enables the end-user to create useful tooling around this common set of `supportingFiles`.

Left unset, the generated code is nearly equivalent to that in `master`.


etherealjoy:
> Currently, all executions of openapi-generator produce a set of supportingFiles alongside the model/ and api/ files. If you have more than one openapi-generator spec target, these supportingFiles are completely identical, except for the namespace in which they are generated.

This a problem we need to solve.
Thanks for the problem description. We have solved the same problem for Qt5 C++ Client.
https://github.com/swagger-api/swagger-codegen/pull/7206

The way it is, is that there is the modelPrefix and 
`qt5cpp --additional-properties cppNamespace=k8s --additional-properties modelNamePrefix=K8S`
So you can plug any amount of Clients.
I think this is a better way way to handle multiple clients so that you always know which file you are opening and including from.

wing328:
The following error found in the AppVeyor job:
```
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=2g; support was removed in 8.0
Exception in thread "main" java.lang.RuntimeException: Could not process model 'Order'.Please make sure that your schema is correct!
	at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:400)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:790)
	at org.openapitools.codegen.cmd.Generate.run(Generate.java:315)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:59)
Caused by: java.nio.file.InvalidPathException: Illegal char <<> at index 18: ..\model\#include <cpprest/details/basic_types.h>.h
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at java.nio.file.Paths.get(Paths.java:84)
	at org.openapitools.codegen.languages.CppRestSdkClientCodegen.toModelImport(CppRestSdkClientCodegen.java:263)
	at org.openapitools.codegen.DefaultGenerator.processModels(DefaultGenerator.java:1036)
	at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:395)
	... 3 more
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=2g; support was removed in 8.0
```

Looks like there's an issue with the Mustache template:
```
Caused by: java.nio.file.InvalidPathException: Illegal char <<> at index 18: ..\model\#include <cpprest/details/basic_types.h>.h
```

Peaches491:
@etherealjoy I've filed an issue for the duplicate supporting files problem description you've highlighted: https://github.com/OpenAPITools/openapi-generator/issues/586

etherealjoy:
@Peaches491 
In this case it would appear we have some options

This is a breaking change so not always good for everyone
CMakeLists.txt
─ common `all Files will have common namespace but not prefixed but common to all clients`
     ├── JsonBody.h
     ├── ModelBase.cpp
     ├── ModelBase.h
     ├── MultipartFormData.cpp
     ├── MultipartFormData.h
     ├── Object.cpp
     ├── Object.h
─ model `all Files will be namespaced and prefixed, unique to single client`
       └── ...
─ api `all Files will be namespaced and prefixed, unique to a single client`
        └── ...

Or using current implementation we can do `add_subdirectory` from a higher level `CMakeLists.txt` with all client folders. `cmake` can do target specific include so we can explore this option as well to make this current client include only from this specific related files.
─ client  `all Files will be namespaced and prefixed to a single client`
     ├── CMakeLists.txt
     ├── HttpContent.cpp 
     ├── HttpContent.h 
     ├── IHttpBody.h
     ├── JsonBody.cpp
     ├── JsonBody.h
     ├── ModelBase.cpp
     ├── ModelBase.h
     ├── MultipartFormData.cpp
     ├── MultipartFormData.h
     ├── Object.cpp
     ├── Object.h
     ├── api
         └── ...
     └── model
        └── ...

In both cases we can solve duplicate linking.
Models having the same name is not always an error because of spec updates same model will have  different members. 

These are breaking changes so we have to consider

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppRestSdkClientCodegen.java
```diff
@@ -176,6 +205,19 @@ public void processOpts() {
...
+        if (additionalProperties.containsKey(SUPPORTING_FILES_DIRECTORY)) {
+            supportingFilesDirectory = additionalProperties.get(SUPPORTING_FILES_DIRECTORY).toString();
+
+        }
+        if (additionalProperties.containsKey(SUPPORTING_FILES_NAMESPACE)) {
+            supportingFilesNamespace = additionalProperties.get(SUPPORTING_FILES_NAMESPACE).toString();
+        }
+
+        // Ensure that the supportingFilesNamespace always ends with "::"
+        if(!supportingFilesNamespace.isEmpty() && !supportingFilesNamespace.endsWith("::")) {
```
rienafairefr: I'm wondering why not add the :: in the templates ? The `{{{supportingFilesNamespace}}}ApiClient` was unexpected, I think.

PR #633, opened 2018-07-24
[java] add useNullForUnknownEnumValue option

### LINKED ISSUE #625: [Java] new option: Throw an exception for invalid enum values instead of using null
I have a similar problem than the one reported in https://github.com/swagger-api/swagger-codegen/issues/5950

When working with enum in a java server, if the enum value sent by the client is unknown, after deserialization, the value is set to `null`.

This might be useful in some cases (the OpenAPI Spec does not always contains all the possible values) but is not the behavior expected in most of the cases.

---
After having looked at the thread https://github.com/swagger-api/swagger-codegen/issues/5950
I propose to add a new option: `useNullForUnknownEnumValue`:

* `true` (current behavior `3.1.x`)
* `false` 

In my opinion (and as other participants in the original thread) starting with `3.2.0` the new default should be `false`.

With  `useNullForUnknownEnumValue == true`: 

https://github.com/OpenAPITools/openapi-generator/blob/af3ca293e4ff3defcbfd4a32b82f7d3229a29d48/samples/server/petstore/jaxrs/jersey2/src/gen/java/org/openapitools/model/EnumTest.java#L52-L61

With `useNullForUnknownEnumValue == false`:

```java
   @JsonCreator 
   public static EnumStringEnum fromValue(String text) { 
     for (EnumStringEnum b : EnumStringEnum.values()) { 
       if (String.valueOf(b.value).equals(text)) { 
         return b; 
       } 
     } 
     throw new IllegalArgumentException("Unexpected value '" + text + "'");
   } 
} 
```

---
In the thread there is also a discussion about removing `fromValue` and the `@JsonCreator` to let Jackson handling deserialisation.

This is not possible in my opinion:

* There are cases where the Java enum name is not the enum value. (Lowercase/uppercase, values with spaces are allowed in OAS3, …)
* The type of the value is not always a string.

---

cc: @TehBakker, @slarti-b, @wing328, @cbornet 

Java Technical Committee: @bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01)

jmini:
### PR checklist

- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.2.x`, `4.0.x`. Default: `master`.
- [x] Java Technical Committee: @bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01)

cc: @TehBakker, @slarti-b, @wing328, @cbornet

### Description of the PR

Fixes #625

This PR introduce a new option `useNullForUnknownEnumValue`.

* When `false` the generated code throws an exception when the value is unknown (not present in the enum)
* When `true` (`3.1.x` and `3.0.x` behavior): unknown values are mapped to `null`

Migration guide is updated with this PR.

2 Java samples are still using the old behavior:

* `bin/java-petstore-jersey1.sh`
* `bin/java-petstore-jersey2.sh`

### DIFF
--- a/bin/java-petstore-jersey2.json
+++ b/bin/java-petstore-jersey2.json
@@ -1,7 +1,7 @@
 {
     "library": "jersey2",
     "artifactId": "petstore-jersey2",
-	"additionalProperties" : {
-		"useNullForUnknownEnumValue" : true
-	}
+        "additionalProperties" : {
+            "useNullForUnknownEnumValue" : true
+        }
 }
--- a/bin/java-petstore-jersey2.json
+++ b/bin/java-petstore-jersey2.json
@@ -1,4 +1,7 @@
 {
     "library": "jersey2",
-    "artifactId": "petstore-jersey2"
+    "artifactId": "petstore-jersey2",
+	"additionalProperties" : {
+		"useNullForUnknownEnumValue" : true
+	}
 }
--- a/bin/java-petstore-jersey1.sh
+++ b/bin/java-petstore-jersey1.sh
@@ -27,7 +27,7 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate --artifact-id petstore-java-client-jersey1 -t modules/openapi-generator/src/main/resources/Java -i modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -g java -o samples/client/petstore/java/jersey1 -DhideGenerationTimestamp=true --library=jersey1 $@"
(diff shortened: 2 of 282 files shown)


jmini:
Changed base branch: `3.2.0-SNAPSHOT` is now on `master` branch

bin/java-petstore-jersey2.json
```diff
@@ -1,4 +1,7 @@
 {
     "library": "jersey2",
-    "artifactId": "petstore-jersey2"
+    "artifactId": "petstore-jersey2",
+	"additionalProperties" : {
+		"useNullForUnknownEnumValue" : true
```
wing328: Minor suggestion: What about using spaces instead of tabs?

jmini: Implemented. Thank you for the review.

PR #640, opened 2018-07-24
[all] Fix inconsistent parameter names when ensuring uniqueness.

Jauler:
### PR checklist

- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.2.x`, `4.0.x`. Default: `master`.

### Description of the PR

If parameter renaming takes place while ensuring unique names in DefaultCodegen.java fromOperation method, parameter lists queryParams, pathParams, headerParams, cookieParams are not updated accordingly which causes wrong parameters to be selected when using {{#pathParams}}, {{#queryParams}} etc. in mustache templates.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2351,6 +2351,13 @@ public class DefaultCodegen implements CodegenConfig {
 
                 CodegenParameter p = fromParameter(param, imports);
 
+                // ensure unique params
+                if (ensureUniqueParams) {
+                    if (!isParameterNameUnique(p, allParams)) {
+                        p.paramName = generateNextName(p.paramName);
+                    }
+                }
+
                 allParams.add(p);
 
                 if (param instanceof QueryParameter || "query".equalsIgnoreCase(param.getIn())) {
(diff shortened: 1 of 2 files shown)


jmini:
Can you provide a small test case or example spec to reproduce the issue?

Jauler:
I was using spec from https://github.com/kubernetes/kubernetes/blob/release-1.10/api/openapi-spec/swagger.json which is rather big...

The important part is:
```
......
    "parameters": [
     {
      "uniqueItems": true,
      "type": "string",
      "description": "path to the resource",
      "name": "path",
      "in": "path",
      "required": true
     },
     {
      "uniqueItems": true,
      "type": "string",
      "description": "Path is the URL path to use for the current proxy request to pod.",
      "name": "path",
      "in": "query"
     }
    ]
......
```
Notice parameter named "path" in both, query params and in path params, which caused the generation of new name for `allParams` list which was not propogated to `pathParams` and `queryParams` lists.

P.S. If You believe that this is not enough to demonstrate the issue, later I could try to prepare a small test case for that.

jmini:
Thank you for the explanations.
I just added a unit test for your fix 34cf828.
If the CI is green, I will merge it...

---

That said, the rename logic has a lot of bugs... For example with 3 params named `myparam`, they get renamed `myparam`, `myparam2` and `myparam2`...
This can be solved in a follow up PR.

PR #661, opened 2018-07-26
[Elm] Template improvements

### LINKED ISSUE #458: [elm-client] Update dependency: elm-date-extra
Update dependency `"justinmimbs/elm-date-extra": "2.0.0 <= v < 3.0.0"`.

There is a breaking change in the api.

```
The pattern matches things of type:

    Maybe a

But the values it will actually be trying to match are:

    Result String Date.Date
```

### See

https://github.com/OpenAPITools/openapi-generator/blob/b5942624d6714f38c0bde9ae9c7400df1d48b1ed/modules/openapi-generator/src/main/resources/elm/DateTime.mustache

https://github.com/OpenAPITools/openapi-generator/blob/b5942624d6714f38c0bde9ae9c7400df1d48b1ed/modules/openapi-generator/src/main/resources/elm/elm-package.mustache#L15

laughedelic:
### PR checklist

### Description of the PR

This PR fixes #458 and improves doc comments and line spacing in the Elm template. (see also https://github.com/swagger-api/swagger-codegen/issues/8152 for the context).

There are still some extra newlines that are hard to get rid of. They appear because the `{{>blah}}` includes bring the last newline in the parent template. I don't know if this can be somehow improved on the Mustache side, but in general it's quite hard to produce some code that immediately complies with the formatting guideline (elm-format).


laughedelic:
e4fa432 also solves https://github.com/swagger-api/swagger-codegen/issues/8151

wing328:
> They appear because the {{>blah}} includes bring the last newline in the parent template.

@laughedelic thanks for the PR 👍 

Would removing the last newline in the parent (partial) template help? If you can point me to the line number in the template and show me the actual vs expected output, I can help take a look.

cc @trenneman

laughedelic:
> Would removing the last newline in the parent (partial) template help? If you can point me to the line number in the template and show me the actual vs expected output, I can help take a look.

As I understand it, the extra newline comes from the child template. So if you have an `{{>child}}` at the last line of the parent template, the result will have `\n\n` at the end, coming from the parent and the child.

Here is an example piece of template:

https://github.com/OpenAPITools/openapi-generator/blob/e4fa4329a71a7538fed5dbfc67c38fe937fd5a1f/modules/openapi-generator/src/main/resources/elm/modelTypeAlias.mustache#L13-L16

And the produced result:

https://github.com/OpenAPITools/openapi-generator/blob/e4fa4329a71a7538fed5dbfc67c38fe937fd5a1f/samples/client/petstore/elm/src/Data/ApiResponse.elm#L30-L47

One extra newline is added after inlining the `aliasDecoder`, same with the `aliasEncoder`, and then one more in the end, because the `modelTypeAlias` is itself inlined in the `model` template.

bin/elm-petstore.sh
```diff
@@ -27,6 +27,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g elm -o samples/client/petstore/elm $@"
+ags="generate -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g elm -t modules/openapi-generator/src/main/resources/elm -o samples/client/petstore/elm $@"
```
trenneman: do we need this?

modules/openapi-generator/src/main/resources/elm/DateOnly.mustache
```diff
@@ -24,11 +25,8 @@ dateOnlyEncoder model =
...
+        Result.Ok date ->
             Decode.succeed date
 
-        Nothing ->
-            Decode.fail <|
-                "Cannot convert "
-                    ++ str
-                    ++ " to DateOnly"
+        Result.Err msg ->
+            Decode.fail msg
```
trenneman: Something like `Decode.fail <| "Cannot decode ISO string: " ++ msg` makes clear that it's the decoder that failed

modules/openapi-generator/src/main/resources/elm/DateTime.mustache
```diff
@@ -24,11 +25,8 @@ dateTimeEncoder model =
...
+        Result.Ok date ->
             Decode.succeed date
 
-        Nothing ->
-            Decode.fail <|
-                "Cannot convert "
-                    ++ str
-                    ++ " to DateTime"
+        Result.Err msg ->
+            Decode.fail msg
```
trenneman: The same here

bin/elm-petstore.sh
wing328: Yes, I think so as using -t allows us to regenerate the samples without doing another "mvn clean package" after updating the templates.

modules/openapi-generator/src/main/resources/elm/DateOnly.mustache
laughedelic: The error message in the new date-extra is

> Failed to create a Date from string 'foo': Invalid ISO 8601 format

Do you still want to add `"Cannot decode ISO string: "`?

bin/elm-petstore.sh
laughedelic: Yes, I also checked that it's used in other scripts.

trenneman: It feels a bit weird that it sort of overloads the build, but it makes sense to add it if that's also the case for the other generators.

modules/openapi-generator/src/main/resources/elm/DateOnly.mustache
trenneman: Well the thing is this is a `Decoder` error, not a `Date` error. But the type already indicates that so I think we can stick with the `Decode.fail msg`.

laughedelic:
Hi! I replied to the review comments and now I'm not sure if there's still anything left to change.

wing328:
If no more question/feedback on this PR, I'll merge it into master tomorrow (Monday)

andys8:
Nice!

wing328:
Hi all, I've filed https://github.com/OpenAPITools/openapi-generator/pull/879 to improve the code format of the Elm client by removing the EOF from 2 mustache files. Please review to see if it looks good to you.

wing328:
#879 has been reviewed and merged into master.

Shall we use https://github.com/jfmengels/elm-lint to improve the auto-generated Elm code?

trenneman:
@wing328 You typically use `elm-format` for that. You want to add this as a post-processing step?

andys8:
`elm-format` does formatting and minor changes (like changing import order). `elm-analyse` is a longer (static code analysis). We could avoid the two dependencies by making sure the templates themselves are formatted correctly. But it could also be a benefit to make it sure by using `elm-format`.

wing328:
Thanks for the suggestion guys. I will see if I can come up with a solution to run `elm-format` for each auto-generated Elm files. Will keep you guys posted.

trenneman:
That would help. However, my preference would be to format it correctly from the start, without any post-processing.

wing328:
> That would help. However, my preference would be to format it correctly from the start, without any post-processing.

Totally agree but we found that it's not easy to do so with mustache.

For Go, we're trying to use `gofmt -w` and the result looks good so far: https://github.com/OpenAPITools/openapi-generator/pull/929

I've created https://github.com/OpenAPITools/openapi-generator/issues/933 to track the code format improvement

Issue #663, opened 2018-07-27
[PHP] Generate SDK into root directory not into a folder 


developersteve:
##### Description

Currently PHP generation is one of the only libs to generate into a folder inside the nominated root directory

##### openapi-generator version

3.1.1 but has been doing it since 2

##### Steps to reproduce

Generate a standard PHP SDK

##### Suggest a fix/enhancement

Generation to occur in the base directory rather than the named folder inside the base directory

wing328:
@developersteve thanks for the suggestion. I also agree we should make the generators more consistent.

I've filed #681 to remove package path from the default output folder (path)

wing328:
@developersteve PR merged into master and you can find the change in the v3.2.0 released yesterday.

developersteve:
Awesome thank you

PR #681, opened 2018-07-29
[PHP] Decommission "packagePath", add new option "packageName"

wing328:
### PR checklist

### Description of the PR

- Decommission "packagePath"
- Users will need to use -o (--output) to control the destination folder instead to make it consistent with other generators
- add new option "packageName"

To close https://github.com/OpenAPITools/openapi-generator/issues/663

cc @jebentier (2017/07) @dkarlovi (2017/07) @mandrean (2017/08) @jfastnacht (2017/09) @ackintosh (2017/09) @ybelenko (2018/07)


wing328:
@ybelenko what about keeping getPackagePath() or packagePath in all PHP server generators and put the following in the constructor instead?

```
        packagePath = ""; // empty packagePath (top folder)
```

so that the package path (root folder) can be changed easily by the users if they need to customize it somehow?

ybelenko:
@wing328 It was my first approuch. I don't have full knowledge of your path generation, but this way I can generate code in a specific folder. I think we need to ask @ackintosh about that.

wing328:
@ybelenko sure let's wait for @ackintosh to review.

Btw, the following approach is already used in other PHP server generators:
```
        packagePath = ""; // empty packagePath (top folder)
```

ackintosh:
I think `packagePath = ""` is good idea. 💡 

It sets the destination to root folder as default and be able to change if user want.

wing328:
@ackintosh thanks for the feedback. I'll change the rest accordingly.

wing328:
@ybelenko @ackintosh I've decommissioned packagePath and add a new option "packageName". The samples are exactly the same as before. Please review when you've time.

wing328:
@ackintosh good catch. Fixed via https://github.com/OpenAPITools/openapi-generator/pull/681/commits/0d45184fd8bef57ac5bfcc02653bcb6092b17f2a

wing328:
@ybelenko can you merge it if it looks good to you? Thanks

ybelenko:
Now I don't understand what we need `packageName` option for? We don't use it in path construction.

The only reason I can imagine for `packageName`("GeneratedPetstore" packageName as example):
```php
/**
 * FakeApi Class Doc Comment
 * PHP Version 5
 *
 * @category Class
 * @package  GeneratedPetstore
 * @author   OpenAPI Generator team
 */
class FakeApi {
```
Do we keep `packageName` option to avoid breaking changes as fallback solution?
Or do you expect that user wants to use `packageName` in mustaches somehow?

> The main package name for classes. e.g. GeneratedPetstore

If I understand this description correct, then mustache should look like:
```django
class {{packageName}}{{classname}}
```
and result:
```php
class GeneratedPetstoreFakeApi
```

In total `packageName` is a prefix to generated class names, isn't it?

wing328:
The {{packageName}} is used in the README: https://github.com/OpenAPITools/openapi-generator/pull/681/files#diff-5763a353c6e95ff8967e09ce0e52ba53R1

wing328:
## Upgrade Note

The `packagePath` option has been deprecated. Please use `-o` or `--output` to specify the output folder.

To specify the package name in the auto-generated README.md, please use the new option `packageName` instead.

PR #691, opened 2018-07-30
Better handling of operationID starting with numbers

wing328:
### PR checklist

### Description of the PR


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -603,6 +603,12 @@ public String toOperationId(String operationId) {
             operationId = "call_" + operationId;
         }
 
+        // operationId starts with a number
+        if (operationId.matches("^\\d.*")) {
+            LOGGER.warn(operationId + " (starting with a number) cannot be used as method name. Renamed to " + camelize(sanitizeName("call_" + operationId)));
+            operationId = camelize("call_" + operationId);
```
jimschubert: I think "operation" as a prefix makes more sense. "Call" suggests an operation is being performed, which may not always be the case (for example if the templates are modified to create infrastructural services rather than endpoint invokers).

stkrwork: also you are logging the sanitized name here, and then don't sanitize it when assigning it to the variable operationId

jimschubert: Nice catch. This should probably assign, then log the assigned value to avoid issues in the future.

wing328: @jimschubert yup, agree there may be a better choice of word other than "call_". My goal is to make the output compiles without issues due to method names starting with numbers. Users need to rename the operationId if they want complete control.

wing328: @stkrwork good catch. Let me revise that.

wing328:
@jimschubert @stkrwork revised and pushed some changes. Please have a look.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -552,7 +552,7 @@ public String toOperationId(String operationId) {
             operationId = "call_" + operationId;
         }
 
-        // model name starts with a number
+        // operationId starts with a number
         if (operationId.matches("^\\d.*")) {
             LOGGER.warn(operationId + " (starting with a number) cannot be used as method name. Renamed to " + underscore(sanitizeName("call_" + operationId)));
             operationId = "call_" + operationId;
```
stkrwork: still not sanitizing here

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PerlClientCodegen.java
```diff
@@ -377,8 +377,14 @@ public String toOperationId(String operationId) {
...
-            LOGGER.warn(operationId + " (reserved word) cannot be used as method name. Renamed to " + underscore("call_" + operationId));
-            return underscore("call_" + operationId);
+            LOGGER.warn(operationId + " (reserved word) cannot be used as method name. Renamed to " + underscore(sanitizeName("call_" + operationId)));
+            return underscore(sanitizeName("call_" + operationId));
+        }
+
+        // operationId starts with a number
+        if (operationId.matches("^\\d.*")) {
+            LOGGER.warn(operationId + " (starting with a number) cannot be used as method name. Renamed to " + underscore(sanitizeName("call_" + operationId)));
+            operationId = "call_" + operationId;
```
stkrwork: still not sanitizing here

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -603,6 +603,12 @@ public String toOperationId(String operationId) {
             operationId = "call_" + operationId;
```
stkrwork: still not sanitizing here

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -603,6 +603,12 @@ public String toOperationId(String operationId) {
...
         }
 
+        // operationId starts with a number
+        if (operationId.matches("^\\d.*")) {
+            LOGGER.warn(operationId + " (starting with a number) cannot be used as method name. Renamed to " + camelize(sanitizeName("call_" + operationId)));
+            operationId = "call_" + operationId;
```
stkrwork: still not sanitizing here

jmini: I am not sure to understand, this is just about adding the `call_` prefix. The value will be sanitized on line 612: `return camelize(sanitizeName(operationId));`

stkrwork: oh my bad, didnt see that

wing328: @jmini is correct. It was already sanitized a few lines above and that's the reason why I didn't call the function `sanitize` again when the operationId starts with a number.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PerlClientCodegen.java
stkrwork: didnt see sanitization on returning value

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
stkrwork: nvm

wing328:
@jmini @jimschubert @stkrwork thanks for reviewing the change. Let's go with "call_" first and see if we need a better prefix later.

PR #709, opened 2018-08-01
[haskell-http-client] regen samples

jonschoning:


wing328:
@jonschoning shall we add the following to [./bin/utils/ensure-up-to-date](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/utils/ensure-up-to-date#L33) to ensure the haskell-http-client samples are always up-to-date?
```
./bin/haskell-http-client-petstore.sh
```

jonschoning:
Actually the example app for haskell-http-client is built against the `OAS3` yaml file instead of the OAS2 yaml file, so due to that, I would be ok with adding `bin/openapi3/haskell-http-client-petstore.sh` to ensure-up-todate (if we only wanted the OAS2 input file e.g. `bin/haskell-http-client-petstore.sh` then i'd have to fix the example app in the sample folder to use types from the OAS2 input yaml file)

wing328:
Feel free to use scripts in `bin/openapi3/scripts`, e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/bin/utils/ensure-up-to-date#L27

jonschoning:
done

Issue #734, opened 2018-08-03
[Slim] Security samples script fails


ybelenko:
##### Description

Following instruction for every PR I've noticed that I cannot generate security samples anymore.

##### openapi-generator version

3.2.0-SNAPSHOT

##### OpenAPI declaration file content or url

[petstore-security-test.yaml](../blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-security-test.yaml)

##### Command line used for generation

```shell
$ bin/security/php-slim-server-petstore.sh
```

##### Steps to reproduce
```shell
$ bin/security/php-slim-server-petstore.sh
# START SCRIPT: bin/security/php-slim-server-petstore.sh
Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with 
the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec 
(CLI).
 | Error count: 1, Warning count: 1
Errors:
        -attribute paths.'/fake'(put).descriptions is unexpected
Warnings:
        -Unused model: Return

        at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:541)
        at org.openapitools.codegen.cmd.Generate.run(Generate.java:323)
        at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:59)
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
```

wing328:
@ybelenko thanks for reporting the issue. I've filed @739 to fix it.

Issue #710, opened 2018-08-01
Parser optionalVars.hasMore in models


ybelenko:
##### Description

Found weird parser bug.
Last variable in `optionalVars` array shows `hasMore: true` which is obviously wrong.
It's even more weird that this behavior noticed only with `FormatTest`(definition: `format_test`) model withing all spec.
This bug breaks constructor function in models generation.
Mustache
```php
public function construct(
    {{#hasRequired}}
    {{#requiredVars}}
    ${{name}}{{#hasMore}},{{/hasMore}}{{^hasMore}}{{#hasOptional}},{{/hasOptional}}{{/hasMore}}
    {{/requiredVars}}
    {{/hasRequired}}
    {{#hasOptional}}
    {{#optionalVars}}
    ${{name}} = null{{#hasMore}},{{/hasMore}}
    {{/optionalVars}}
    {{/hasOptional}}
) {

}
```
Output:
```php
public function __construct(
    $number,
    $byte,
    $date,
    $password,
    $integer = null,
    $int32 = null,
    $int64 = null,
    $float = null,
    $double = null,
    $string = null,
    $binary = null,
    $dateTime = null,
    $uuid = null,
) {

}
```

Link to parser output:
https://gist.github.com/ybelenko/89f11291565ed4f8719061c89f082f39#file-formattest-json-L1296-L1345
```json
{
  "optionalVars": [
    {
      "baseName": "uuid",
      "getter": "getUuid",
      "setter": "setUuid",
      "dataType": "string",
      "datatypeWithEnum": "string",
      "dataFormat": "uuid",
      "name": "uuid",
      "defaultValueWithParam": " = data.uuid;",
      "baseType": "string",
      "jsonSchema":
        "{\r\n  \"type\" : \"string\",\r\n  \"format\" : \"uuid\"\r\n}",
      "exclusiveMinimum": false,
      "exclusiveMaximum": false,
      "hasMore": true,
      "required": false,
      "secondaryParam": false,
      "hasMoreNonReadOnly": true,
      "isPrimitiveType": true,
      "isContainer": false,
      "isNotContainer": true,
      "isString": true,
      "isNumeric": false,
      "isInteger": false,
      "isLong": false,
      "isNumber": false,
      "isFloat": false,
      "isDouble": false,
      "isByteArray": false,
      "isBinary": false,
      "isFile": false,
      "isBoolean": false,
      "isDate": false,
      "isDateTime": false,
      "isUuid": true,
[...]

PR #735, opened 2018-08-03
[Slim] Models constructor

ybelenko:
### PR checklist

### Description of the PR
New model template generates file like:
```php
class Animal
{

    /** @var string $className */
    private $className;

    /** @var string $color (optional) Default 'red' */
    private $color = 'red';

    /**
     * Animal constructor
     *
     * @param string $className
     * @param string|null $color (optional) Default 'red'
     */
    public function __construct(
        $className,
        $color = 'red'
    ) {
        $this->className = $className;
        $this->color = $color;
    }

    /**
     * Alternative static class constructor
     *
     * @param mixed[]|null $data Associated array of property values initializing the model
     * @throws InvalidArgumentException when $data doesn't contain required constructor arguments
     * @example $animal = Animal::createFromObject([ 'className' => 'foobar' ]);
     *
     * @return Animal
     */
    public static function createFromObject(array $data = null)
    {
        if ($data['className'] === null) {
            throw new InvalidArgumentException("'className' can't be null");
        }
        $className = $data['className'];
        $color = (isset($data['color'])) ? $data['color'] : 'red';
        return new Animal(
            $className,
            $color
        );
    }
}
```

### Why that way and not another

**Why I think that `private $className` and `private $color` is better than single `$container` from PHPClient codegen:**
1. It's more common and expected for developers, more descriptive, you can say what properties exists with a quick look
2. Most of time developer needs to query database and map result to related model class. For this case there is [PDOStatement::fetchObject()](http://php.net/manual/en/pdostatement.fetchobject.php). Other database drivers(mysqli) has similar methods too.
3. I've quickly checked popular ORMs. [Eloquent](https://laravel.com/docs/5.1/eloquent) have builtin generator to produce models and doesn't follow approuch from this PR. [Doctrine](https://www.doctrine-project.org/) however, creates similar models, example [User.php](https://github.com/1ma/Slim-Doctrine-Demo/blob/master/src/Domain/User.php). The last one [Atlas](http://atlasphp.io/) claims own persistence models with optional fallback to domain models. That domain models looks similar to that PR, example [Map From Persistence To Domain](http://atlasphp.io/cassini/orm/domain.html#1-1-11-5)
4. PhpSymfonyCodegen already produces that kind of [models](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/php-symfony/SymfonyBundle-php/Model/User.php) and it looks like a stable implementation, correct me if I'm wrong

**Why I've added both traditional and associative array constructor in the same class:**
1. I cannot choose one, so I've decided to add `createFromObject` static method when you want to parse JSON and get new model instance as result. I like classic constructor with arguments list more, but I understand that sometimes it doesn't fit.
2. With classic constructor you can see required and optional variables right away.

### Unresolved question to discuss
I can't say for sure that force user to set required arguments is a good idea. I mean sometimes you receive messy JSON object without required fields, if you'll try to parse it via `createFromObject` method your code will die with uncatched exception. From another point, it's very usefull during development to notice missed properties instantly. 
PHPClient throws exceptions when you're trying to set property with setter method, but it gives no exceptions within constructor even if you don't provide any props at all. It doesn't make sense to me. :confused: 
**The question is, should we split arguments to required and optional and throw exceptions right from constructor method OR set all arguments optional and validates them silently with special method `valid()` only when user ask for it?**

### Issues found in process
1. Parser optionalVars.hasMore in models, #710 :bug: 
2. [Slim] Security samples script fails, #734

### DIFF
--- a/modules/openapi-generator/src/main/resources/php-slim-server/model.mustache
+++ b/modules/openapi-generator/src/main/resources/php-slim-server/model.mustache
@@ -3,20 +3,22 @@
  * {{classname}}
  */
 namespace {{modelPackage}};
+
 use \InvalidArgumentException;
 
 /**
  * {{classname}}
  */
-class {{classname}} {
+class {{classname}}
+{
 
     {{#requiredVars}}
-    /** @var {{dataType}} ${{name}} {{#description}}{{description}} {{/description}}*/
+    /** @var {{dataType}} ${{name}}{{#description}} {{description}}{{/description}} */
     private ${{name}};
 
     {{/requiredVars}}
     {{#optionalVars}}
-    /** @var {{dataType}} ${{name}} (optional) {{#description}}{{description}} {{/description}}{{#defaultValue}}Default {{{defaultValue}}} {{/defaultValue}}*/
+    /** @var {{dataType}} ${{name}} (optional){{#description}} {{description}}{{/description}}{{#defaultValue}} Default {{{defaultValue}}}{{/defaultValue}} */
     private ${{name}}{{#defaultValue}} = {{{defaultValue}}}{{/defaultValue}};
 
     {{/optionalVars}}
@@ -25,10 +27,10 @@ class {{classname}} {
     {{#hasVars}}
      *
     {{#requiredVars}}
-     * @param {{dataType}} ${{name}} {{#description}}{{description}}{{/description}}
+     * @param {{dataType}} ${{name}}{{#description}} {{description}}{{/description}}
     {{/requiredVars}}
     {{#optionalVars}}
-     * @param {{dataType}}|null ${{name}} (optional) {{#description}}{{description}} {{/description}}{{#defaultValue}}Default {{{defaultValue}}} {{/defaultValue}}
+     * @param {{dataType}}|null ${{name}} (optional){{#description}} {{description}}{{/description}}{{#defaultValue}} Default {{{defaultValue}}}{{/defaultValue}}
     {{/optionalVars}}
     {{/hasVars}}
      */
@@ -43,7 +45,8 @@ class {{classname}} {
     ) {
     {{/hasVars}}
     {{^hasVars}}
-    public function __construct() {
+    public function __construct()
+    {
     {{/hasVars}}
         {{#allVars}}
         $this->{{name}} = ${{name}};
@@ -59,9 +62,12 @@ class {{classname}} {
      *
      * @return {{classname}}
      */
-    public static function createFromObject(array $data = null) {
+    public static function createFromObject(array $data = null)
+    {
     {{#requiredVars}}
-        if ($data['{{baseName}}'] === null) throw new InvalidArgumentException("'{{baseName}}' can't be null");
+        if ($data['{{baseName}}'] === null) {
+            throw new InvalidArgumentException("'{{baseName}}' can't be null");
+        }
         ${{name}} = $data['{{baseName}}'];
     {{/requiredVars}}
     {{#optionalVars}}
--- a/modules/openapi-generator/src/main/resources/php-slim-server/model.mustache
+++ b/modules/openapi-generator/src/main/resources/php-slim-server/model.mustache
@@ -11,13 +11,13 @@ use \InvalidArgumentException;
 class {{classname}} {
 
     {{#requiredVars}}
-    /** @var {{dataType}} ${{name}} {{#description}}{{description}}{{/description}}*/
+    /** @var {{dataType}} ${{name}} {{#description}}{{description}} {{/description}}*/
     private ${{name}};
 
     {{/requiredVars}}
     {{#optionalVars}}
-    /** @var {{dataType}} ${{name}} (optional) {{#description}}{{description}}{{/description}}*/
-    private ${{name}};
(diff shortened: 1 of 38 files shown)


ackintosh:
> Why I think that `private $className` and `private $color` is better than single `$container` from PHPClient codegen:

My understanding that why PHPClient uses `$container` is, to make implementing [ArrayAccess](http://php.net/manual/ja/class.arrayaccess.php) interface easy. So `private $className, private $color` is good if the models of Slim doesn't need to implement ArrayAccess. 👍

ybelenko:
> My understanding that why PHPClient uses `$container` is, to make implementing ArrayAccess interface easy. So `private $className, private $color` is good if the models of Slim doesn't need to implement ArrayAccess. +1

@ackintosh Btw, why do you need ArrayAccess in ClientSDK? For HttpGuzzle? Give as an example, when ArrayAccess is convenient.

And you didn't answered to main question:
> **The question is, should we split arguments to required and optional and throw exceptions right from constructor method OR set all arguments optional and validates them silently with special method valid() only when user ask for it?**

wing328:
> @ackintosh Btw, why do you need ArrayAccess in ClientSDK? For HttpGuzzle? Give as an example, when ArrayAccess is convenient.

My understanding is that it's more like preference. Consider the following object definition:
```
  AnimalFarm:
    type: array
    items:
      $ref: '#/definitions/Animal'
```

How would you generate that in your proposed model generation?

ybelenko:
> How would you generate that in your proposed model generation?

Now I see your point. Hmm... Isn't it breaks programming basics? Nobody think about model object as an Array. I can imagine more ridiculous example:
```yaml
definitions:
  Currency:
    type: string
    enum:
      - usd
      - eur
      - gbp
```
And there is no way to create working model for example above. It doesn't have properties it's property by itself. As fallback solution I can do:
```php
class AnimalFarm implements ArrayAccess {}
```
for models which is arrays, but I don't know what to do with silly `Currency` example and others.

ybelenko:
@wing328 
Know what's even more funny? I've used Slim codegen a lot, and I didn't use models at all. I mean I didn't even think that they useful. I've enhanced APIs and middlewares, request validation, authentication, even UnitTests was my primary goal... but models. 

Models very useful for client SDK, no doubts, because you don't know how request object looks like. I would say in most cases you don't even read documentation for API when there is SDK. But when you're provider of API and develop server, then you perfectly know models and properties. In this case I would change `AnimalFarm` definition to object which is more friendly to use. I would change it, because I can and I don't want to create headaches to myself :smile:

ybelenko:
Well, I don't see any way, beside to add generator options to customize models. That's what I think is minimal set of options for models:

### `modelProperties`
Possible values:
* `privateVars` - generate private var for every model prop, eg `private $color = 'red'`
* `protectedVars` - generate protected var for every model prop, eg `protected $color = 'red'`
* `publicVars` - generate public var for every model prop, eg `public $color = 'red'` 
* `protectedContainer` - PHPClient generator way. Generate single `protected $container = []` var for all props
* `none` - don't generate properties

Default: `privateVars`

### `modelConstructorArguments`
Possible values: 
* `listRequiredAndOptional` - generate constructor with list of required and optional arguments. PHP throws error when required aguments of constructor not provided
* `listAllOptional` - generate constructor with list of optional arguments. All of them are optional. This way you can create new instance of model without props and assign them afterwards and not get any PHP errors
* `singleOptionalObject` - PHPClient generator approach. `public function __construct(array $data = null)`
* `none` - don't generate constructor. Can't imagine any real example, but maybe somebody needs it

Default: `listRequiredAndOptional`

### `modelSettersAndGetters`
Possible values: 
* `magicMethods` - `__set` and `__get` PHP magic methods to prevent creating of 100 of setters 
* `gettersAndSetters` - PHPClient approach, create setter and getter function for every prop
* `none` - don't generate setters at all. User can choose `'modelProperties': 'publicVars'` and assign them directly without any methods, or you can set them via constructor only

Default: `magicMethods`

### `modelThrowExceptions`
Possible values:
* `inConstructor` - generated model throws `InvalidArgumentException` when constructor required arguments are invalid or `null`
* `inValidMethod` - generated model throws `Exception` within `valid` public method
* `inSetters` - generated model throws `InvalidArgumentException` when value assigned to required prop is invalid or `null`
* `inConstructorAndSetters` - `inConstructor` and `inSetters` both options applied
* `none` - don't throw exceptions in models

Default: `inValidMethod`

### `modelJsonSerialization`
Possible values:
* `objectSerializer` - use PHPClient `ObjectSerializer` class for models serialization in `jsonSerialize` method
* `jsonMapper` - use [jsonmapper](https://github.com/cweiske/jsonmapper) package for models serialization in `jsonSerialize` method
* `custom` - my own implementation of `jsonSerialize` method, don't know how to call it yet(maybe this option is useless, user can set `none` value and implement own method in mustache)
* `none` - don't generate `jsonSerialize` method

Default: `objectSerializer`

cc @ackintosh

ybelenko:
Doesn't make sense to enhance outdated Slim version anymore. New models will be added to Slim4 generator after launch #3658.

PR #251, opened 2018-06-08
Validate spec on generation by default

jimschubert:
Adds a validation parameter to CodegenConfigurator, and passes through
options from CLI, Maven Plugin and Gradle Plugin to that property.

Default is to validate the spec during generation. If spec has errors,
we will output errors as well as warnings to the user.

Option can be disabled by passing false to validateSpec (Maven/Gradle)
or --validate-spec (CLI).

This was brought up in #208.  We want to validate the spec on generation to prevent OpenAPI Spec 3.x options from creeping into OpenAPI Spec 2.0 support.  I would also consider this a bug fix, because as a user I would be surprised to find that generation succeeds when my spec is invalid by default.

### PR checklist

- [ ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the correct branch: Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

PR #740, opened 2018-08-05
Update --skip-validate-spec option with "--validate-spec"

wing328:
### PR checklist

### Description of the PR

Update --skip-validate-spec option with "--validate-spec" to make consistent with the maven/gradle plug-in option. An improvement to https://github.com/OpenAPITools/openapi-generator/pull/251

@jimschubert please review to see if this change looks good to you.

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -194,10 +194,10 @@ public class Generate implements Runnable {
             description = CodegenConstants.REMOVE_OPERATION_ID_PREFIX_DESC)
     private Boolean removeOperationIdPrefix;
 
-    @Option(name = {"--skip-validate-spec"},
-            title = "skip spec validation",
-            description = "Skips the default behavior of validating an input specification.")
-    private Boolean skipValidateSpec;
+    @Option(name = {"--validate-spec"},
+            title = "enable spec validation (true or false)",
+            description = "Enable validating an input specification and terminate if errors found. (default: true)")
+    private String validateSpec;
 
     @Override
     public void run() {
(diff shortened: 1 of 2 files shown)


jimschubert:
After some discussion, I'm going to close this for a handful of reasons which I'll enumerate here.

1. CLI and Maven/Gradle plugins have slightly different use cases. Although all three end up in code being generated, the CLI is meant to be executed directly by the user or user scripts while the plugins are meant to be declaratively configured and executed by a build tool. As such, CLI options are user-focused while plugin options are value-focused.
2. Changing to an `--option=string` option rather than a `--switch` option would make this inconsistent with how other switches are applied at the CLI level.
3. The current switch `--skip-validate-spec` follows common Linux/Unix command format for switches, for example from `man git-commit`:

   ```
          -n, --no-verify
            This option bypasses the pre-commit and commit-msg hooks. See also 
   githooks(5).
   
          --allow-empty
              Usually recording a commit that has the exact same tree as its sole 
   parent commit is a mistake, and the command prevents you from making such a
              commit. This option bypasses the safety, and is primarily for use by 
   foreign SCM interface scripts.
   ```
    In the above example, git allows skipping with the more common `--no-` prefix. I chose `--skip-` because you can find this in some CLI tools, and we have a large non-English speaking user base; I felt that `--skip-` would translate more easily than simply `--no-`. I could see us using the same option names as above for opt-out and opt-in behaviors.
4. Changing from a switch to a string-valued option doesn't add any apparent value. I could see someone who is intimately familiar with options in the plugin trying to apply those same options to the CLI tool, but they'd immediately be presented with an error and then run `cli help` and find the correct option. I've contributed to the codebase for over 2 years, and I don't know all options well enough to pass them effortlessly between the plugins and the CLI… so I think the likelihood of this is pretty low. More commonly, I'd assume people work similarly to how I work (open plugin README for options, execute `cli help` or `cli config-help` for options).

We could definitely change the command line option name if the community thinks it's necessary. We could also include both options (`--skip-validate-spec` as boolean and `--validate-spec=[true|false]` as string option). Changing the existing option name for a new option name would be a minor breaking change for the CLI.

Issue #551, opened 2018-07-12
Unable to use systemProperties in gradle plugin


FatCash:
##### Description

Hi, I'm struggling to configure the gradle plugin to generate only model.
I can't seem to pass any property at all in systemProperties. Is my syntax incorrect?

```
openApiGenerate {
    generatorName = "java"
    ....
    configOptions = [
            dateLibrary: "java8",
            useBeanValidation: true
    ]
    systemProperties = [
            modelDocs: "false"
    ]
}
```

gives: 
`Caused by: java.lang.IllegalStateException: System.getProperty(key) must not be null
        at org.openapitools.generator.gradle.plugin.tasks.GenerateTask.doWork(GenerateTask.kt:350)
`
If sent as boolean it gives:
`Caused by: java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String
        at org.openapitools.generator.gradle.plugin.tasks.GenerateTask.doWork(GenerateTask.kt:543)
`

How to send these flags via gradle plugin?
java -Dmodels -DmodelDocs=false -DmodelTests=false

##### openapi-generator version

org.openapitools:openapi-generator-gradle-plugin:3.1.0

wing328:
@FatCash thanks for reporting the issue.

cc @jimschubert

madhead:
Does not work with Kotlin DSL either:

```
openApiGenerate {
    generatorName.set("java")
    inputSpec.set("$projectDir/src/main/swagger/schema.json")
    outputDir.set("$buildDir/generated")
    modelPackage.set("a.b.c.model")
    systemProperties.set(mapOf(
            "models" to "true"  // ←here
    ))
}
```

demtnt:
there problem is in line 
`355: originalEnvironmentVariables[key] = System.getProperty(key)`
Map originalEnvironmentVariables could not contains nulls but getProperty returns null.

And for me it is unclear what we would like to achieve. If remember initial state and return this state after, then we need to remember not existing properties, store it in set and remove it after as well(not only change properties values from map)

jimschubert:
@FatCash I've opened #764 to address this issue. Sorry it took a while- it's been a busy month.

wing328:
THE fix (#764) by @jimschubert has been merged into master. Please give it a try using the SNAPSHOT version, docker image or building the project locally.

FatCash:
Thank you @jimschubert and @wing328 
[...]

PR #764, opened 2018-08-07
[gradle] Support nullable system property values

jimschubert:
### PR checklist

### Description of the PR

The gradle plugin sets all System properties before generation, then
reverts them back to their original state.

System.getProperty/setProperty return null if the property was not
previously set. The Kotlin map was defined with non-nullable key/value
constraints, so setting something not commonly set (modelDocs: "false")
would result in an runtime exception.

This changes the map to support nullable values, and rather than setting
a null System property at the end, it clears those which previously had
no value.

See #551

### DIFF
--- a/modules/openapi-generator-gradle-plugin/samples/local-spec/build.gradle
+++ b/modules/openapi-generator-gradle-plugin/samples/local-spec/build.gradle
@@ -41,6 +41,9 @@ openApiGenerate {
     configOptions = [
             dateLibrary: "java8"
     ]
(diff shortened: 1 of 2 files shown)


jimschubert:
I've added `modelDocs: "false"` as one of the sample's properties.

Compile and install the project and the gradle plugin, then build the sample with:

```
gradle -PopenApiGeneratorVersion=3.2.1-SNAPSHOT openApiGenerate --info
```

PR #769, opened 2018-08-08
[PHP] Upgrade php-cs-fixer to 2.12, enables PHP >= 7.2 support

mattmelling:
### PR checklist

### Description of the PR

The PHP client uses friendsofphp/php-cs-fixer v1.12 for PSR2 analysis. 1,12 doesn't supports PHP up to < 7.2.

PHP >= 7.2 support was added in php-cs-fixer v2 which made some changes to the API and default fixers/rules.

This PR upgrades php-cs-fixer to v2.12. The fixers/rules have been aligned for PSR2 as desired in the original PR on swagger-codegen (#3863). I've changed defaults for some of the rules to prevent lots of whitespace warnings from php-cs-fixer, these typically seem to be down to blank template values and are not raised when using the configuration in master.

#### Notes
php-cs-fixer added the idea of "risky" fixers in v2. Since the `strict_param` and `strict_comparison` fixers are considered risky, we must alter the command line to `vendor/bin/php-cs-fixer fix --dry-run --diff -vvv --allow-risky yes` even when using `--dry-run`.

/cc @jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @ybelenko


modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -58,7 +58,7 @@ class ObjectSerializer
                 if ($value !== null
                     && !in_array($openAPIType, [{{&primitives}}], true)
                     && method_exists($openAPIType, 'getAllowableEnumValues')
-                    && !in_array($value, $openAPIType::getAllowableEnumValues())) {
+                    && !in_array($value, $openAPIType::getAllowableEnumValues(), true)) {
```
ybelenko: Can you explain, please, how this change is related to `phpcs` fixer?

mattmelling: `php-cs-fixer` raises a `strict-param` warning for in_array calls where the `$strict` parameter is false (which is the default). Passing true ensures we use strict comparison, i.e. value + type checking. 

From my reading of the code I couldn't see any reason why we would *need* loose comparisons, especially since they can result in weird behaviour as documented [here](http://php.net/manual/en/function.in-array.php#106319) and [here](http://php.net/manual/en/function.in-array.php#121479).

ackintosh:
Oh, sorry 💦

LGTM ✨ I'll merge this PR  later today if no further feedback.

ackintosh:
## Upgrade Note

php-cs-fixer `2.12` requires PHP5.6. To use php client in PHP5.5, please set `~1.12` to composer.json as below:

```diff
    "require-dev": {
        "phpunit/phpunit": "^4.8",
        "squizlabs/php_codesniffer": "~2.6",
-       "friendsofphp/php-cs-fixer": "~2.12"
+       "friendsofphp/php-cs-fixer": "~1.12"
    },
```

PR #799, opened 2018-08-13
Fixes for meta and list command

jmini:
### PR checklist

### Description of the PR

To reproduce:

Create a `my-codegen-openapi-generator` artefact using the `meta` command as described here:
https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md#creating-a-new-template

**Change 1/**: If you create a generator that have no type (`CLIENT`, `SERVER`, `DOCUMENTATION`, `CONFIG`, `OTHER`) it should be displayed when you run the command:

 (classpath in the windows format, for linux or Mac replace the `;` with a `:`)

```
java -cp out/generators/my-codegen/target/my-codegen-openapi-generator-1.0.0.jar;modules/openapi-generator-cli/target/openapi-generator-cli.jar org.openapitools.codegen.OpenAPIGenerator
```

The change introduces the category `UNDIFINED` for this case.
It also skip a category if not generator is present for it.

**Change 2/**: the meta command is now from type `OTHER`:

Example  ` MyCodegenGenerator`:

```java
  public CodegenType getTag() {
    return CodegenType.OTHER;
  }
```
**Change 3/**: meta script is added to ensure-up-to-date


jimschubert:
@jmini what is the problem you're seeing that you're trying to resolve with this?

If you're adding the UNDEFINED category for custom generators that return `null` for `getType()`, I have to ask why. For a generator project, I don't think we should ever want to consider an "undefined" generator type like this (especially when we already have an "OTHER" category.

Also, I don't think the meta generator should default to `OTHER`. What's the reason for that change? Are you trying to force users to define their generator types after invoking `meta`? Why not just add a command to the task and allow users to define the `CodegenType` value at generation time?

A side note… while looking at the customization doc, I found that the last command before `Selective generation` is incorrect. Would you mind changing it as part of your branch? It should be:

```
java -cp out/generators/my-codegen/target/my-codegen-openapi-generator-1.0.0.jar:modules/openapi-generator-cli/target/openapi-generator-cli.jar \ 
  org.openapitools.codegen.OpenAPIGenerator generate -g my-codegen   \
  -i https://raw.githubusercontent.com/openapitools/openapi-generator/master/modules/openapi-generator/src/test/resources/2_0/petstore.yaml \
  -o ./out/myClient
```

(notice the output path and `org.openapitools` instead of `io.openapitools`.

jimschubert:
Also, directions in the README for this generator are incorrect. Would you mind changing them on your branch as well?

https://github.com/OpenAPITools/openapi-generator/blob/3a5b223b87bd69f17d0d9940d8c84c65dd3c270a/modules/openapi-generator/src/main/resources/codegen/README.mustache#L49

Should match what I posted above:

```
java -cp /path/to/openapi-generator-cli.jar:/path/to/your.jar \
    org.openapitools.codegen.OpenAPIGenerator generate -g {{name}} \
    -i /path/to/openapi.yaml -o ./test 
```

Notice: entry class is incorrect, and example is missing `generate` command.

jimschubert:
@jmini I see your point about the null types, although I really think this is an edge case.

Could we name it "UNSPECIFIED" or "UNTAGGED" instead. My concern is mainly that "UNDEFINED" looks like the CLI or the generator module are exposing a programming error.

jmini:
See https://github.com/OpenAPITools/openapi-generator/pull/817 for the docs.

samples/meta-codegen/lib/pom.xml
```diff
@@ -116,7 +116,7 @@
     </dependencies>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <openapi-generator-version>3.0.0-SNAPSHOT</openapi-generator-version>
+        <openapi-generator-version>3.2.2-SNAPSHOT</openapi-generator-version>
```
wing328: I'll take note of this during the release process.

jmini: `lib/` is generated (output of the `meta` command of the CLI). It is updated with ensure-up-to-date.

jmini:
WIP: not all the changes requested here, are implemented.

jmini:
Everything is implemented, this PR is ready for review.

jmini:
I will merge this later today if nobody else wants to comment.

Issue #819, opened 2018-08-15
[KOTLIN] Spring Boot Server Generator


dr4ke616:
##### Description

Hey everyone,

I'd like to propose a new server generator for Kotlin using the Spring Boot framework.

##### openapi-generator version

Based off `master` - next minor release

mikezx6r:
There is a kotlin-spring language generator under development. Have you checked it out? Depending on your needs, it may be adequate.

I'm watching it right now, but there's still a few issues/changes I'd like to see. Going to provide feedback/proposals to people working on it.

If you have `allOf` in an object reference, and depending on which commit you're on, it either attempts to use inheritance of a data class, or it produces duplicate properties.

I also like the delegate pattern that the Java Spring generator uses where everything is defined in one interface (as I only want API and Models generated, and manage the Controllers etc myself).

Or was this to track the effort currently taking place?

dr4ke616:
@mikezx6r this issue can actually be closed now, as the first implementation of a Kotlin spring-boot was implemented in PR #820

PR #820, opened 2018-08-15
[KOTLIN] Spring Boot Server Generator

dr4ke616:
### PR checklist

### Description

This PR aims to add support for a Spring Boot Kotlin sever generator.

#### Sample Generation Scripts

- `bin/kotlin-springboot-petstore-server.sh`
- `bin/openapi3/kotlin-springboot-petstore-server.sh`
- `bin/windows/kotlin-springboot-petstore-server.bat`
- `bin/windows/openapi3/kotlin-springboot-petstore-server.bat`

#### Outstanding Tasks Related To This PR:

- [x] Additional Configuration options:
    - [x] Swagger annotations
    - [x] Service interfaces
    - [x] Service implementations
    - [x] Default exception handlers
- [x] Generate README.md
- [x] Enable optional bean validation using `javax.validation.Valid`
- [x] ~~Use `RestController` instead of `Controller` annotation (see PR #571 for more info with regard to how it is achieved using the Spring Java generator)~~ Edit: Outside of scope to this PR, will add in later PR.

CCing relevant contributors:
- @jimschubert (2017/09)


wing328:
Linked to #819

dr4ke616:
@wing328 yup, this PR will only support JDK8 for now. I may submit another PR later with support for JDK7.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinSpringServerCodegen.java
```diff
@@ -0,0 +1,442 @@
...
+    public Map<String, Object> postProcessOperationsWithModels(Map<String, Object> objs, List<Object> allModels) {
+        Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
+        if (operations != null) {
+            List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
+            ops.forEach(operation -> {
+                List<CodegenResponse> responses = operation.responses;
+                if (responses != null) {
+                    responses.forEach(resp -> {
+
+                        if ("0".equals(resp.code))
```
jimschubert: I love that this setting a "good" response code if none was defined in the spec. 

And it makes me wonder if we should loop responses and verify that response codes are valid as part of the `cli validate --recommend` option.

@jmini @ackintosh any thoughts on adding that as a feature (because you both worked with me most on the feature and requirements)?

modules/openapi-generator/src/main/resources/kotlin-spring/libraries/spring-boot/README.mustache
```diff
@@ -0,0 +1,21 @@
...
+```bash
+mvn package && java -jar target/openapi-spring-1.0.0.jar
+```
+
+To build the project using gradle, run:
+```bash
+gradle build && java -jar build/libs/openapi-spring-1.0.0.jar
+```
+
+If all builds successfully, the server should run on [http://localhost:8080/](http://localhost:8080/)
```
jimschubert: I think this needs to template out the `port` option.

modules/openapi-generator/src/main/resources/kotlin-spring/libraries/spring-boot/README.mustache
```diff
@@ -0,0 +1,21 @@
...
+
+## Getting Started
+
+This document assumes you have either maven or gradle available, either via the wrapper or otherwise. This does not come with a gradle / maven wrapper checked in.
+
+By default a [`pom.xml`](pom.xml) file will be generated. If you specified `gradleBuildFile=true` when generating this project, a `build.gradle.kts` will also be generated. Note this uses [Gradle Kotlin DSL](https://github.com/gradle/kotlin-dsl).
+
+To build the project using maven, run:
+```bash
+mvn package && java -jar target/openapi-spring-1.0.0.jar
```
jimschubert: `openapi-spring` should be replaced with `{{artifactId}}` here and a couple lines below (the version possibly as well).

modules/openapi-generator/src/main/resources/kotlin-spring/beanValidationQueryParams.mustache
```diff
@@ -0,0 +1 @@
+{{#required}}@NotNull {{/required}}{{>beanValidationCore}}
```
jimschubert: `beanValidationCore` template is missing in this PR, and causes generation to fail.

modules/openapi-generator/src/main/resources/kotlin-spring/libraries/spring-boot/buildGradleKts.mustache
```diff
@@ -0,0 +1,45 @@
...
+    id("org.springframework.boot") version "2.0.3.RELEASE"
+    id("io.spring.dependency-management") version "1.0.5.RELEASE"
+}
+
+dependencies {
+    compile("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
+    compile("org.jetbrains.kotlin:kotlin-reflect")
+    compile("org.springframework.boot:spring-boot-starter-web")
+{{#swaggerAnnotations}}
+    compile("io.swagger:swagger-annotations:1.5.14")
```
jimschubert: Is this old swagger-annotations version due to compatibility issues with Spring Boot? Can it be bumped to latest pre-2.0 version (1.5.21)?

modules/openapi-generator/src/main/resources/kotlin-spring/libraries/spring-boot/buildGradleKts.mustache
```diff
@@ -0,0 +1,45 @@
...
+    jcenter()
+    mavenCentral()
+}
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions.jvmTarget = "1.8"
+}
+
+plugins {
+    val kotlinVersion = "1.2.41"
```
jimschubert: Are newer versions of kotlin (1.2.{50,51,60}) unsupported by Spring Boot?

modules/openapi-generator/src/main/resources/kotlin-spring/libraries/spring-boot/defaultBasePath.mustache
```diff
@@ -0,0 +1 @@
+{{contextPath}}
```
jimschubert: Is `defaultBasePath.mustache` here for compatibility or extension reasons? If not, can `{{contextPath}}` just be inserted where `{{>defaultbasePath}}` exists, to simplify template maintenance and customization?

I don't see `{{>defaultbasePath}}` anywhere, and I haven't looked too in-depth into the template inclusions to understand if it's potentially in the missing file.

.gitignore
```diff
@@ -188,6 +188,10 @@ samples/client/petstore/kotlin-string/build
 samples/server/petstore/kotlin-server/ktor/build
 \?
 
+# Kotlin Spring Boot
+samples/server/petstore/kotlin-springboot
+samples/server/openapi3/petstore/kotlin-springboot
+
```
wing328: @dr4ke616 Can you please remove these lines so that the project will include these files to be tested by the CI?

bin/kotlin-springboot-petstore-server.sh
```diff
@@ -0,0 +1,36 @@
...
+executable="./modules/openapi-generator-cli/target/openapi-generator-cli.jar"
+
+if [ ! -f "$executable" ]
+then
+  mvn clean package
+fi
+
+# if you've executed sbt assembly previously it will use that instead.
+export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
+ags="$@ generate -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g kotlin-spring -o samples/server/petstore/kotlin-springboot --additional-properties=library=spring-boot"
```
jimschubert: target `-t` with the templates under resource, and you don't need to rebuild the jar each time only a template is changed.

.gitignore
jimschubert: I was going to ask you about this, then I saw that you commented on it.

Putting these directories into .gitignore is a great way to reduce noise while doing local development and awaiting feedback, though!

dr4ke616:
@jimschubert I've merged your suggested changes

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinSpringServerCodegen.java
```diff
@@ -0,0 +1,442 @@
...
+        addSwitch(GRADLE_BUILD_FILE, "generate a gradle build file using the Kotlin DSL", gradleBuildFile);
+        addSwitch(SWAGGER_ANNOTATIONS, "generate swagger annotations to go alongside controllers and models", swaggerAnnotations);
+        addSwitch(SERVICE_INTERFACE, "generate service interfaces to go alongside controllers. In most " +
+                "cases this option would be used to update an existing project, so not to override implementations. " +
+                "Useful to help facilitate the generation gap pattern", serviceInterface);
+        addSwitch(SERVICE_IMPLEMENTATION, "generate stub service implementations that extends service " +
+                "interfaces. If this is set to true service interfaces will also be generated", serviceImplementation);
+        addSwitch(USE_BEANVALIDATION, "Use BeanValidation API annotations to validate data types", useBeanValidation);
+
+        supportedLibraries.put(SPRING_BOOT, "Spring-boot Server application.");
```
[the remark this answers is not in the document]
dr4ke616: Yes, initially I was thinking to have this extended to potentially support other libraries as well, Spring MVC being one example. 

If I'm honest, I was a little unsure about what overall direction to take, either have this as a supported library under `kotlin-server`, or standalone as is with the java spring implementation. I'm happy to go with which ever direction you feel to be most consistent and suitable.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinSpringServerCodegen.java
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    @Override
+    public CodegenType getTag() {
+        return CodegenType.SERVER;
+    }
+
+    @Override
+    public String getName() {
+        return "kotlin-spring";
```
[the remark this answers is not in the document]
dr4ke616: If it's desired overall to have this as a library under `kotlin-server`, id be happy to make those changes. However, I'd probably continue to extend `AbstractKotlinCodegen`, rather than extending `KotlinServerCodegen` (if possible).

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinSpringServerCodegen.java
```diff
@@ -0,0 +1,442 @@
...
+        writePropertyBack(USE_BEANVALIDATION, useBeanValidation);
+
+        modelTemplateFiles.put("model.mustache", ".kt");
+        apiTemplateFiles.put("api.mustache", ".kt");
+        supportingFiles.add(new SupportingFile("README.mustache", "", "README.md"));
+
+        if (this.serviceInterface) {
+            apiTemplateFiles.put("service.mustache", "Service.kt");
+        } else if (this.serviceImplementation) {
+            additionalProperties.put(SERVICE_INTERFACE, true);
```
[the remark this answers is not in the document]
dr4ke616: My thinking was that you can't generate the implementation classes with missing interfaces, as it would lead to compile errors due to the missing generated interfaces.

I will however re-visit this logic, and perhaps add appropriate logging. Or perhaps if `serviceInterface` is set to `false`, just generate the standalone classes that don't implement anything.

jimschubert:
@dr4ke616 thanks. I think the only remaining big question is about the missing `beanValidationCore.mustache`. The inclusion causes generate to fail for me, and my fix was just to remove that. I didn't attempt compiling the generated server, and will look into it further a little later today or tomorrow.

I asked @wing328 about the `kotlin-spring` generator name and we don't think this needs to change at the moment. The Java server has a fully separate `spring` generator name, and from that perspective I think `kotlin-spring` is the preferred name.

We have an open feature request to split options for locating generators into framework, language, CodegenType. At that point, I think the `kotlin-spring` would change.

dr4ke616:
@jimschubert great, RE `beanValidationCore.mustache` I've a fix for that on the way, plus some minor changes regarding other comments made. Will push later today or tomorrow and let you know.

Regarding library naming, i will leave as `kotlin-spring` for the time being.

dr4ke616:
@jimschubert suggested changes made in ae9d383

jimschubert:
Tested against the new changes, and things look good. I added a stub object in the generated getPetById and verified the response. So the PR looks good to me. Thanks @dr4ke616 for the awesome contribution!

I have a few non-blocking questions, though…

I don't work with Spring, personally, and found it weird that it fails on finding an appropriate response type when making a GET request from the browser. The browser sends the following Accept header:

```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
```

So, Spring should be falling back to application/json for the final `*/*` option. Instead, it returns the following any time xml/xhtml+xml are defined:

```
<html>
    <body>
        <h1>Whitelabel Error Page</h1>
        <p>This application has no explicit mapping for /error, so you are seeing this as a fallback.</p>
        <div id='created'>Mon Aug 20 22:20:16 EDT 2018</div>
        <div>There was an unexpected error (type=Not Acceptable, status=406).</div>
        <div>Could not find acceptable representation</div>
    </body>
</html>
```

Is this because the spring starter dependency includes Thymeleaf or whatever the Spring templating for HTML is?

Also, I've found that the implementation won't construct a valid response unless I have exactly `application/json` or `application/json; charset=utf-8`. If I have `application/xml` as documented in the `RequestMapping` attribute, it also fails. Does this indicate that we'll need some additional configuration based on the `xml` template property?

/cc @wing328 or @jmini if one of you are familiar with Spring Boot would you mind looking this over to make sure I haven't missed something, and then merging?

dr4ke616:
Ive made one final commit to this PR which I think it may be worth mentioning (see af871fd). Initially the mapping from `date-time` was to `LocalDateTime`, I updated to `OffsetDateTime`, similarly `date` was also mapped to `LocalDateTime`, now it is mapped to `LocalDate`.

I made this change to the `AbstractKotlinCodegen`, therefore i feel it worth mentioning in the case it has any negative impact on some of the existing Kotlin code generators which I may have overlooked. If there are issues, i can revert the commit and add the type mappings straight to `KotlinSpringServerCodegen`.

dr4ke616:
@jimschubert regarding your first question, I believe spring is responding with a HTML like response because within your Accept header is `text/html`. This is handled with Spring default error handler, which behaves differently depending on what that is in your `Accept` header. Some more info available [here](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-web-applications.html#boot-features-error-handling).

Regarding your second question, this is due to a missing `jackson-dataformat-xml` dependency (Fixed here - ba20ef7).

wing328:
> Ive made one final commit to this PR which I think it may be worth mentioning (see 530ab3f). Initially the mapping from date-time was to LocalDateTime, I updated to OffsetDateTime, similarly date was also mapped to LocalDateTime, now it is mapped to LocalDate.

I suggest not to change the mapping in the abstract Kotlin class as it will impact the`kotlin` and `kotlin-server` generators. This is not to say `OffsetDateTime` is not appropriate.

For `kotlin` client to support `OffsetDateTime`, we can add another dateLibrary option in addition to what we provide today:
```
	dateLibrary
	    Option. Date library to use
	        string - String
	        java8 - Java 8 native JSR310
	        threetenbp - Threetenbp
```

For kotlin spring boot generator, I would suggest overriding the mapping locally instead of updating the default mapping in the abstract Kotlin class.

@jimschubert any concern to change the default mapping in the abstract Kotlin class, which will impact the Kotlin server generator you created?

dr4ke616:
Should be ready for final review.
- Rebased changes on top of master
- Reverted date time mappings in `AbstractKotlinCodegen` and added them locally to `KotlinSpringServerCodegen` (as @wing328 suggested)
- Regenerated all samples

jimschubert:
OffsetDateTime is part of Java8, and isn't another date library.

It was originally implemented as LocalDateTime, if I remember correctly, to match how other generators work. It looks like maybe _all_ JVM generators default to LocalDateTime. I agree that it should probably be OffsetDateTime because OAS defines date-time as targeting 
RFC 3339 (https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14).

The issue is that `LocalDateTime#parse` fails on standard ISO 8601 formats whereas OffsetDateTime does not. And while OAS requires rfc3339 for date-time, in reality most APIs support ISO 8601 in which zone information is optional.

An example is when the value is provided by JavaScript:

```
var d = new Date();
"2018-08-23T03:00:33.254Z"
```

Try this in a REPL, and you get failures on default parsing with `LocalDateTime#parse`:

```

scala> import java.time._
import java.time._

scala> import java.time.format._
import java.time.format._

scala> val input = "2018-08-23T03:00:33.254Z"
input: String = 2018-08-23T03:00:33.254Z

scala> val local = LocalDateTime.parse(input)
java.time.format.DateTimeParseException: Text '2018-08-23T03:00:33.254Z' could not be parsed, unparsed text found at index 23
  at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1952)
  at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
  at java.time.LocalDateTime.parse(LocalDateTime.java:492)
  at java.time.LocalDateTime.parse(LocalDateTime.java:477)
  ... 28 elided

scala> val off = OffsetDateTime.parse(input)
off: java.time.OffsetDateTime = 2018-08-23T03:00:33.254Z
```

Parse the same with formatting options and you get consistent parsing:

```
scala> val off = OffsetDateTime.parse(input, DateTimeFormatter.ISO_OFFSET_DATE_TIME)
off: java.time.OffsetDateTime = 2018-08-23T03:00:33.254Z

scala> val local = LocalDateTime.parse(input, DateTimeFormatter.ISO_OFFSET_DATE_TIME)
local: java.time.LocalDateTime = 2018-08-23T03:00:33.254
```

Now, suppose your input date time is the following (valid ISO 8601 without zone information):

```
scala> val input = "2018-08-23T03:00:33"
input: String = 2018-08-23T03:00:33
```

This fails to parse for everything _except_ `LocalDateTime#parse`:

```
scala> val local = LocalDateTime.parse(input, DateTimeFormatter.ISO_OFFSET_DATE_TIME)
java.time.format.DateTimeParseException: Text '2018-08-23T03:00:33' could not be parsed at index 19
  at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
  at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
  at java.time.LocalDateTime.parse(LocalDateTime.java:492)
  ... 28 elided

scala> val off = OffsetDateTime.parse(input, DateTimeFormatter.ISO_OFFSET_DATE_TIME)
java.time.format.DateTimeParseException: Text '2018-08-23T03:00:33' could not be parsed at index 19
  at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
  at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
  at java.time.OffsetDateTime.parse(OffsetDateTime.java:402)
  ... 28 elided

scala> val local = LocalDateTime.parse(input)
local: java.time.LocalDateTime = 2018-08-23T03:00:33

scala> val off = OffsetDateTime.parse(input)
java.time.format.DateTimeParseException: Text '2018-08-23T03:00:33' could not be parsed at index 19
  at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
  at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
  at java.time.OffsetDateTime.parse(OffsetDateTime.java:402)
  at java.time.OffsetDateTime.parse(OffsetDateTime.java:387)
  ... 28 elided
```

As far as I can tell, the assumption across JVM languages that this should be LocalDateTime may have been due to misreading the ABNF

> date-time       = full-date "T" full-time

We do allow for modifying these types at generation time, so I don't think it's a huge deal to replace how this is done across all generators. After all, clients _should_ understand the date-time formats being used in their targeted APIs.

If anything, I think all JVM languages should be updated to support both parse methods with `DateTimeFormatter.ISO_OFFSET_DATE_TIME` as this is more consistent and works in both cases when inputs have time zone suffixes.

@wing328 do you have any historical context as to why the default for date-time was chosen as `LocalDateTime` over `OffsetDateTime`? Would you be opposed to changing this for all generators in the 4.0.0 release?

wing328:
> @wing328 do you have any historical context as to why the default for date-time was chosen as LocalDateTime over OffsetDateTime? Would you be opposed to changing this for all generators in the 4.0.0 release?

Sorry, I don't recall the history/reason of why LocalDateTime was used instead.

If no one has further feedback on this PR, I'll merge it tomorrow (Friday)

cbornet:
Some history : https://github.com/swagger-api/swagger-codegen/issues/2138

cbornet:
Before that the use of LocalDateTime had been a bug, not a deliberate choice

wing328:
Tweet to promote the new generator: https://twitter.com/oas_generator/status/1034438177782546433

wing328:
FYI. I've filed https://github.com/OpenAPITools/openapi-generator/pull/918 with some minor improvements (license, README, etc)

wing328:
@dr4ke616 thanks again for the new generator, which has been included in the v3.2.3 release: https://twitter.com/oas_generator/status/1035200785066254336

PR #837, opened 2018-08-18
Data object classes for templating context

rienafairefr:
Sorry in advance for the massive PR, though I can't seem to see how it could be made with less 

### PR checklist

@jimschubert @wing328 

### Description of the PR

in some discussion with @jimschubert [here](https://github.com/OpenAPITools/openapi-generator/pull/690#issuecomment-413080607) I've seen that the current plethora of Map<String, Object> pseudo-classes should be avoided. Indeed, it's making us do a lot of casting back and forth between Object and Map<String, Object>, a lot of unchecked casting warnings are suppressed, type safety is not used, overall it's something that could be changed for the better

I've implemented a bunch of data-only classes (suffixed `Bundle`s for now but I could see using `DTO` or `Bag` if whatever terminology is preferred) for replacing all these Map<String, Object> that could be replaced. Of course we need to keep the vendorExtensions, additionalProperties maps, and probably others I'm forgetting of the top of my head, unchanged, because these really need to be truly dynamic (they can be changed by the spec or the user). 

There is a bit of conundrum with the additionalProperties there because when executing a template, we want to pass it a context that can be a POJO, but we also want the values from additionalProperties to be available (e.g. `{{licenseInfo}}` or `{{appDescription}}`. I added a AdditionalPropertiesBundle base class to handle that, but it feels hacky to me: I convert the POJO to a map, and add the additionalproperties to it, just before it goes into the `template.execute(context...)`, it needs a better solution I think.


jimschubert:
I've created a [Separation of Concerns](https://github.com/OpenAPITools/openapi-generator/projects/5) project to track some of the larger refactoring efforts. I've added this PR and #690 to that project, because I think they're both directly related to the efforts I've documented in the project.

As I discussed in #690, `Map<String, Object>` is quite a pain to work with. I was originally only envisioning the model passed to `template.execute(context...)` as being a strong typed object. This is because I'd like to see the generation workflow redone as well. That is, rather than DefaultGenerator converting an OpenAPI spec into our `Codegen*` models and dispatching to the `DefaultConfig` or derived `CodedgenConfig` instances at each step, I think it would be more maintainable and intuitive to parse the spec, transform it to a "ready" API abstraction, then pass this abstraction into the workflow object, handing off to the framework generator only once for post-processing right before passing it to templating. This reduces the back-and-forth processing and repeated mutations along the way. See #841 and #843 for more in-depth details.

If we change all input parameters to these `*Bundle` types, this is a major breaking change and would need to retarget merging into the 4.0.0 branch.

jimschubert:
@rienafairefr sorry I've just seen your response. I've had a lot of stuff come up lately (I'm building a house and will be moving in about 6 weeks). I'll try to get this reviewed over the next week.

rienafairefr:
@jimschubert I've rewritten and removed the changes in public interface altogether, the generators still operate on `Map<String, Object>` as was the case before, it is just the final step (file generation) that uses an almost-strongly-typed object. I say almost because the *Bundle classes are still operable as Map<String, Object> (because we need to add `additionalProperties` to it for example)

jimschubert:
@rienafairefr awesome. I'll try to take a look this week. It's there anything you'd like me to try and tackle?

Issue #372, opened 2018-06-21
openAPI v3 - Callbacks feature


CyrilleBenard:
##### Description

New feature :

I was wondering when the callbacks feature (openAPI v3) will be planned to be developed in the openapi-generator ?
The 2 "languages" I'm interesting in are :
C++ (cpp-pistache-server and cpp-restsdk)
Java (jaxrs-resteasy)

Below a small openapi file useful to check this feature

##### openapi-generator version
N/A

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
  version: 1.0.0
  title: Simple Callback example
  description: Validate a simple callback
paths:
  /subscribe:
    post:
      summary: Subscribe to a notification
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                callbackUrl: 
                  type: string
                  format: uri
                  example: http://client_domain/the_call_back_endpoint
              required:
                - callbackUrl
      responses:
        '201':
          description: Subscription created
        '403':
          description: Forbidden

      callbacks: 
        onEventReport: 
          '{$request.body#/callbackUrl}': 
            post:
              requestBody:
                required: true
                content:
                  application/json:
                    schema:
                      type: object
                      properties:
                        message:
                          type: string
                          example: Event 01 raised
                      required:
                        - message
              responses: # Expected responses (from the client) to the callback message
                '204':
                  description: Successful acknowledgement
                default:
                  description: Unexpected error

```

```json
{
  "openapi": "3.0.0",
  "info": {
    "version": "1.0.0",
    "title": "Simple Callback example",
    "description": "Validate a simple callback"
  },
  "paths": {
    "/subscribe": {
      "post": {
        "summary": "Subscribe to a notification",
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "callbackUrl": {
                    "type": "string",
                    "format": "uri",
                    "example": "http://client_domain/the_call_back_endpoint"
                  }
                },
                "required": [
                  "callbackUrl"
                ]
              }
            }
          }
        },
        "responses": {
          "201": {
            "description": "Subscription created"
          },
          "403": {
            "description": "Forbidden"
          }
        },
        "callbacks": {
          "onEventReport": {
            "{$request.body#/callbackUrl}": {
              "post": {
                "requestBody": {
                  "required": true,
                  "content": {
                    "application/json": {
                      "schema": {
[...]

PR #861, opened 2018-08-21
Add callback model

devplayer0:
### PR checklist

### Description of the PR
This adds a new `CodegenCallback` class, a list of which is now present in `CodegenOperation`. `CodegenOperation` now also includes a `isCallbackRequest` boolean since `fromCallback()` (the method added to `DefaultCodegen` to process callbacks in OpenAPI operations) uses CodegenOperation as the model for a callback request.

A `CodegenOperation` which represents a callback request will have a `null` operation id.

A test is included for this new model.

(See #372)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -24,6 +24,7 @@ import com.samskivert.mustache.Mustache.Compiler;
 import io.swagger.v3.core.util.Json;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
+import io.swagger.v3.oas.models.callbacks.Callback;
 import io.swagger.v3.oas.models.examples.Example;
 import io.swagger.v3.oas.models.headers.Header;
 import io.swagger.v3.oas.models.media.ArraySchema;
@@ -50,6 +51,7 @@ import io.swagger.v3.parser.util.SchemaTypeUtil;
 import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.openapitools.codegen.CodegenDiscriminator.MappedModel;
 import org.openapitools.codegen.examples.ExampleGenerator;
 import org.openapitools.codegen.serializer.SerializerUtils;
@@ -75,6 +77,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import java.util.stream.Collectors;
 
 public class DefaultCodegen implements CodegenConfig {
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2220,29 +2223,35 @@ public CodegenOperation fromOperation(String path,
...
+
         CodegenOperation op = CodegenModelFactory.newInstance(CodegenModelType.OPERATION);
         Set<String> imports = new HashSet<String>();
         if (operation.getExtensions() != null && !operation.getExtensions().isEmpty()) {
             op.vendorExtensions.putAll(operation.getExtensions());
-        }
 
-        if (operation == null)
-            throw new RuntimeException("operation cannot be null in fromOperation");
+            Object isCallbackRequest = op.vendorExtensions.remove("x-callback-request");
```
wing328: @devplayer0 Curious, why remove the vendor extension "x-callback-request"?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2220,29 +2223,35 @@ public CodegenOperation fromOperation(String path,
...
+            String operationId = getOrGenerateOperationId(operation, path, httpMethod);
+            // remove prefix in operationId
+            if (removeOperationIdPrefix) {
+                int offset = operationId.indexOf('_');
+                if (offset > -1) {
+                    operationId = operationId.substring(offset + 1);
+                }
             }
+            operationId = removeNonNameElementToCamelCase(operationId);
+            op.operationId = op.isCallbackRequest ? null : toOperationId(operationId);
```
wing328: @devplayer0 may I know why the `operationId` is not set if callback is defined in the spec?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
devplayer0: It's removed since `isCallbackRequest` is set on the operation just below this line - I was just using the vendor extension as a way of distinguishing between a normal operation and one that represents a callback request in `fromOperation()` (this extension is set to `true` in `fromCallback()` without changing the existing method signature.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
devplayer0: @wing328 My initial thought was that since it's not a server side method, callbacks aren't really "operations" on an API per se. In the [OpenAPI example for callbacks](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/callback-example.yaml) operation IDs are not used when defining the callback request.

If I were to not set the operation ID to null, I feel like it would be important to change the behaviour of generating a default operation ID if the user does not provide one - the runtime expression parts should probably be removed and the callback identifier (`onData` in the OpenAPI example) should probably be used as a prefix. What do you think?

wing328: > In the OpenAPI example for callbacks operation IDs are not used when defining the callback request.

`operationId` (optional) is for uniquely identifying the operation:

```
Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
```

In that example, there's only one operation so it's ok not to specify the operationId.

>  the runtime expression parts should probably be removed and the callback identifier (onData in the OpenAPI example) should probably be used as a prefix.

Perfectly fine with me that you want to change the default naming when the callback is specified. My experience told me that the default will not meet everyone's requirement so the users need to specify the "operationId" if they want a better `operationId`

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
wing328: Please leave a coment in https://github.com/OpenAPITools/openapi-generator/blob/6544528df7238caeb559c57b0e5dcb3614063940/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2671 about x-callback-request being removed later in the process.

I'll see I've a better to do it without using vendor extensions (to make it consistent with the rest of the program)

devplayer0:
@wing328 Done.

wing328:
@devplayer0 I'll run some tests with this PR tomorrow and let you know if I've further feedback.

wing328:
UPDATE: I ran some tests and the result looks good. I'll merge it tomorrow if no one has further feedback/question on this PR.

wing328:
@devplayer0 thanks for the PR, which has been included in the v3.2.3 release: https://twitter.com/oas_generator/status/1035200785066254336

devplayer0:
@wing328 Great!

PR #60, opened 2018-05-16
[go] Fix issue #51, change object mapping from *interface{} to map[st…

ilk:
### PR checklist
- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

To fix #51  I propose the following changes:

1. map swagger object type to map[string]interface{} instead of *interface{} 
2. removing pointer since a map is already a reference type

### Technical Commitee

@antihax 
@wing328

Issue #51, opened 2018-05-15
[go] map type object *interface{} convert problem


ilk:
##### Description

Definition property "type": "object" gets converted to *interface{} which returns an error when set to a struct or pointer pointer.

##### openapi-generator version

3.0.0
##### Suggest a fix/enhancement

Instead of an *interface{} map the object to map[string]interface{}

wing328:
@ilk thanks for creating the issue. May I know if you can also share a spec to more easily reproduce the issue?

When using `type: object`, are you referring to an arbitrary JSON payload?

ilk:
**For example:**
```json
...
"Provider" : {
  "type" : "object",
  "properties" : {
    "Foobar" : {
      "type" : "object",
      "description" : "Foobar-specific configuration object"
    },
...
```

**Results to:**
```go
type Provider struct {
  Foobar *interface{}
  ...
}
```

**Better would be:**
```go
type Provider struct {
  Foobar map[string]interface{}
  ...
}
```

louis77:
@antihax Which works properly if definitions are used in the OpenAPI Spec. 

However, arbitrary inline objects are currently represented as *interface{}'s. To use @ilk 's example, `Foobar` is an arbitrary inline object, most likely there will be no corresponding struct definition in the Go code that will use the go-client.

wing328:
Is it correct to say that the proposed change will also work with XML payload (since Go client supports the "withXML" option) ?
```
	withXml
	    whether to include support for application/xml content type and include XML annotations in the model (works with libraries that provide support for JSON and XML) (Default: false)
```

PR #888, opened 2018-08-23
RFC: [go] define "any-type" schema as an empty interface in generated go code.

ashanbrown:
I could use some advice on this one.  The fix for https://github.com/OpenAPITools/openapi-generator/issues/51 (PR https://github.com/OpenAPITools/openapi-generator/pull/60) broke our use of the go generator because it now assumes unspecified (i.e. "any-type") schemas will be `map[string]interface{}` instead of `*interface{}`.  This change tries to restore the empty interface, albeit as `interface{}` instead of `*interface{}`.  I'm not sure what the intent of using a pointer to an interface was in the original code but maybe that would be preferred .  This code does not change the generated petshop code (but maybe petshop should be updated to include an "any-type" schema).

The "any-type" schema is mentioned at https://swagger.io/docs/specification/data-models/data-types/ which claims to be a guide to openapi 3.0, but I don't see it mentioned by name in the "open-api" spec.

Thanks for your help and for keeping this project moving forward.

### PR checklist

### Description of the PR

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
@@ -83,7 +83,9 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
                         "complex64",
                         "complex128",
                         "rune",
-                        "byte")
+                        "byte",
+                        "interface{}"
+                        )
         );
 
         instantiationTypes.clear();
@@ -298,12 +300,19 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
 
         if (ref != null && !ref.isEmpty()) {
             type = openAPIType;
-        } else if (typeMapping.containsKey(openAPIType)) {
-            type = typeMapping.get(openAPIType);
-            if (languageSpecificPrimitives.contains(type))
-                return (type);
-        } else
-            type = openAPIType;
+        } else {
+            // Handle "any type" as an empty interface
+            if (openAPIType == "object" && p != null && !ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
+                return "interface{}";
+            }
+
+            if (typeMapping.containsKey(openAPIType)) {
+                type = typeMapping.get(openAPIType);
+                if (languageSpecificPrimitives.contains(type))
+                    return (type);
+            } else
+                type = openAPIType;
+        }
         return type;
     }
 
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
@@ -302,7 +302,7 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
             type = openAPIType;
         } else {
             // Handle "any type" as an empty interface
-            if (openAPIType == "object" && p != null && !ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
+            if (openAPIType.equals("object") && p != null && !ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
                 return "interface{}";
             }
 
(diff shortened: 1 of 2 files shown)


wing328:
@ashanbrown thanks for the PR. cc the technical committee for review:

@antihax (2017/11) @bvwells (2017/12) @grokify (2018/07)

ashanbrown:
I was also thinking that maybe we should push the problem upward any introduce an "anytype" `openAPIType` that would default to "object" but could be mapped to something by typeMapping on a per-generator basis.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
```diff
@@ -298,12 +300,19 @@ public String getSchemaType(Schema p) {
...
             type = openAPIType;
-        } else if (typeMapping.containsKey(openAPIType)) {
-            type = typeMapping.get(openAPIType);
-            if (languageSpecificPrimitives.contains(type))
-                return (type);
-        } else
-            type = openAPIType;
+        } else {
+            // Handle "any type" as an empty interface
+            if (openAPIType == "object" && p != null && !ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
```
wing328: @ashanbrown Please use `"object".equals(openAPIType)` instead

ashanbrown:
Looking through the code I noticed that a number of other (not go) generators have a concept of an "any" type.  I'm wondering if if we should just always return "any" for all code generators and let the individual generators map the type to the a primitive, as shown in:

https://github.com/OpenAPITools/openapi-generator/compare/master...launchdarkly:ashanbrown/any-primitive

eclipse1985:
Hey there! Any update on this? We really need it :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
eclipse1985: @ashanbrown can you please fix it? Here in Arduino we really appreciate to have this PR merged :)

ashanbrown: Sure @eclipse1985, I've made that change.  It wasn't sure that the team was really offering to merge this though because I didn't see further comments from @antihax, @bvwells or  @grokify about this or my follow-up comment at https://github.com/OpenAPITools/openapi-generator/pull/888#issuecomment-415777718.  I'd still be happy to have this merged as is but my impression was there might be a larger issue here.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
```diff
@@ -298,12 +300,19 @@ public String getSchemaType(Schema p) {
...
             type = openAPIType;
-        } else if (typeMapping.containsKey(openAPIType)) {
-            type = typeMapping.get(openAPIType);
-            if (languageSpecificPrimitives.contains(type))
-                return (type);
-        } else
-            type = openAPIType;
+        } else {
+            // Handle "any type" as an empty interface
+            if (openAPIType.equals("object") && p != null && !ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
```
zippolyte: Hmm, it seems this would only work with the following suggestion.

```suggestion
            if (openAPIType.equals("object") && p != null && ModelUtils.isObjectSchema(p) && !ModelUtils.isMapSchema(p)) {
```

 Indeed, a schema like the following will be an `ObjectSchema`:
```yaml
AnyType:
  type: object
```

ashanbrown:
@zippolyte @bkabrda I don't have much to add on this as I haven't been working with openapi a lot lately.  I don't know if you also looked at the alternate proposal at https://github.com/OpenAPITools/openapi-generator/pull/3878 by @mattiabertorello.  I suspect he has a deeper (and more up-to-date) understanding of how this code works than I do.  Any change that allows specification of an arbitrary type is would be fine by me.  I don't completely follow the change suggested by @zippolyte and would be reluctant to implement it myself without a corresponding test for the behavior we are expecting.

wing328:
This is already done: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java#L133

Please give the latest master a try.

Closing this PR for the time being.

PR #894, opened 2018-08-24
Create override for array schema default value when using the spring …

### LINKED ISSUE #890: [Java][Spring] Optional query parameter of type array-string doesn't respect defaults
##### Description

Using OAS 3, I have a query parameter defined as follows:
```yaml
    exampleSort:
      name: sort
      description: Sort parameters for list response
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
        default: 'a'
```
However after running the generator, my controller class has the following defined as `defaultValue`:
```
@ApiParam(value = "Sort parameters for list response", defaultValue = "new ArrayList<>()") @Valid @RequestParam(value = "sort", required = false, defaultValue="new ArrayList<>()") List<String> sort
```

When a request is made to the service with the optional parameter missing, Spring creates a new string list and adds `new ArrayList<>()` as the single element of the list, so the defaulted string becomes:

`List<String> sort = ["new ArrayList<>()"]`

When I leave `required: false` but remove the default entry:

```yaml
    exampleSort:
      name: sort
      description: Sort parameters for list response
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
```

Then the result is 

```
@ApiParam(value = "Sort parameters for list response") @Valid @RequestParam(value = "sort", required = false) List<String> sort
```

But in this case, Spring does not create a default value (since none was specified), so sort becomes null in the controller:

`List<String> sort = null`

Spring MVC expects a String as default value, but if it's meant to be a list, it could be a comma delimited list as a string `"a"` or at the very least, an empty string (which is parsed as an empty list).

3.2.2

```yaml
openapi: 3.0.0
paths:
  /v1/examples:
    get:
      tags:
        - Example
      summary: Get a list of Examples
      operationId: getExamples
      parameters:
        - $ref: '#/components/parameters/exampleSort'
...
components:
  parameters:
    exampleSort:
      name: sort
      description: Sort parameters for list response
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
        default: 'a'
...
```

Gradle task

```
openApiGenerate {
    inputSpec = openApiInput.path
    outputDir = openApiOutputDir.path
    generatorName = 'spring'
    invokerPackage = ...
    modelPackage = ...
    apiPackage = ...
    additionalProperties = [
        'interfaceOnly'     : 'true',
        'basePackage'       : ...,
        'configPackage'     : ...,
        'sourceFolder'      : 'src/generated/java/',
        'java8'             : 'true',
        'delegatePattern'   : 'true',
        'implicitHeaders'   : 'true',
        'useTags'           : 'true',
        'booleanGetterPrefix': 'is',
    ]
}
```

##### Steps to reproduce

See description ^^

##### Related issues/PRs

None

I've already deduced where this value is coming from:

`AbstractJavaCodegen.toDefaultValue(Schema p) {...}`

Perhaps a resolution could be to override `toDefaultValue(Schema p)` in `SpringCodegen` to implement the necessary functionality, without having to worry about the implications on other types of code generators.

jason-cohen:
…codegen.

### PR checklist

- [ ] @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger

### Description of the PR

Fix #890

### Note

There is an issue with the swagger-parser (https://github.com/swagger-api/swagger-parser/issues/814) where an empty string for the default field is parsed as null. This fix won't achieve the desired effect until that issue is resolved and released.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
@@ -21,6 +21,7 @@ import com.samskivert.mustache.Mustache;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.PathItem;
+import io.swagger.v3.oas.models.media.Schema;
 import org.apache.commons.lang3.tuple.Pair;
 import org.openapitools.codegen.CliOption;
 import org.openapitools.codegen.CodegenConstants;
@@ -34,6 +35,7 @@ import org.openapitools.codegen.CodegenType;
 import org.openapitools.codegen.SupportingFile;
 import org.openapitools.codegen.languages.features.BeanValidationFeatures;
 import org.openapitools.codegen.languages.features.OptionalFeatures;
+import org.openapitools.codegen.utils.ModelUtils;
 import org.openapitools.codegen.utils.URLPathUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


jason-cohen:
@JFCote I had a bit of a back-and-forth with the members of swagger-parser, and I think this PR would need to be revised. Simply performing `getDefault().toString()` will probably not work, as the parsed schema will probably be of type `List<>` and not `String`. This is my presumption based on what I believe the fix for the swagger-parser will be, but I could be wrong.

PR #899, opened 2018-08-26
[CI] use regex to check if TRAVIS_BRANCH is a working branch

jmini:
For our working branches `3.3.x`, `4.0.x`...
A `mvn clean deploy` should be performed to push the artifacts to the remote maven repository. 

This PR changes the check: instead of looking for a specific name, a regex pattern is used.


jmini:
`master` merged to `3.3.x` with 267ffc0e2ede6c503392e41c236315502197d368.

Verified: [3.3.0-SNAPSHOT](https://oss.sonatype.org/content/repositories/snapshots/org/openapitools/openapi-generator/3.3.0-SNAPSHOT/) is published on sonatype snapshots.

PR #906, opened 2018-08-27
[PHP] Add interface/abstract/trait helpers

ybelenko:
### PR checklist

### Description of the PR

It's obvious that some of PHP generators use `Interface`/`Abstract`/`Trait` classes. I've added few functions to generate names for this classes.
Defaults are set to follow "PSR Naming Conventions".
If user will ask we can add prefix/suffix generator options in future. I don't see `modelPrefix` generator option, so I assume it's not important now.
I didn't refresh samples, because they are the same.

Refs:
[PSR Naming Conventions](https://www.php-fig.org/bylaws/psr-naming-conventions/)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPhpCodegen.java
```diff
@@ -481,6 +492,39 @@ public String toModelTestFilename(String name) {
...
 
+    /**
+     * Output the proper interface name (capitalized).
+     * In case the name belongs to the TypeSystem it won't be renamed.
+     *
+     * @param name the name of the interface
+     * @return capitalized model name
+     */
+    public String toInterfaceName(final String name) {
+        return initialCaps(interfaceNamePrefix + name + interfaceNameSuffix);
```
wing328: @ybelenko Instead of using `initalCaps`, what about using `camelize` instead?

PR #911, opened 2018-08-27
[java][webclient] Register Jackson modules on ObjectMapper

### LINKED ISSUE #910: [Java][Webclient] Unable to deserialize OffsetDateTime
##### Description
If we declare a date-time string, the deserialization causes an exception:

```
Type definition error: [simple type, class java.time.OffsetDateTime]; nested exception is com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `java.time.OffsetDateTime` (no Creators, like default construct, exist): no String-argument constructor/factory method to deserialize from String value ('2018-08-27T15:13:30.892Z')
```

3.2.3-SNAPSHOT

```yaml
openapi: 3.0.0

paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            format: int32
      responses:
        '200':
          description: A paged array of pets
          headers:
          content:
            application/json:    
              schema:
                $ref: "#/components/schemas/Pets"
components:
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        lastUpdate:
          type: string
          format: date-time
        tag:
          type: string
    Pets:
      type: array
      items:
        $ref: "#/components/schemas/Pet"
```

Jackson jsr310 module is listed as a dependency but it's not registered inside the objectMapper instance.

We should add: 
```
mapper.registerModule(new JavaTimeModule());
mapper.findAndRegisterModules();
```
during `buildWebClient()` as stated in https://github.com/FasterXML/jackson-modules-java8#registering-modules (see the warning)

Will submit a PR soon.

marcoreni:
### PR checklist

### Description of the PR
Fixes #910

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
@@ -2,6 +2,7 @@ package {{invokerPackage}};
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.core.ParameterizedTypeReference;
@@ -128,6 +129,8 @@ public class ApiClient {
     */
     public static WebClient buildWebClient(ObjectMapper mapper, DateFormat dateFormat) {
         mapper.setDateFormat(dateFormat);
+        mapper.registerModule(new JavaTimeModule());
+        mapper.findAndRegisterModules();
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         ExchangeStrategies strategies = ExchangeStrategies
             .builder()
--- a/samples/client/petstore/java/webclient/src/main/java/org/openapitools/client/ApiClient.java
+++ b/samples/client/petstore/java/webclient/src/main/java/org/openapitools/client/ApiClient.java
@@ -2,6 +2,7 @@ package org.openapitools.client;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.core.ParameterizedTypeReference;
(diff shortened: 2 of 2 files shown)


modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
```diff
@@ -128,6 +129,8 @@ public class ApiClient {
     */
     public static WebClient buildWebClient(ObjectMapper mapper, DateFormat dateFormat) {
         mapper.setDateFormat(dateFormat);
+        mapper.registerModule(new JavaTimeModule());
+        mapper.findAndRegisterModules();
```
cbornet: Are you sure you need to do this ? We don't do it for the other jackson-based generators

marcoreni: Hello @cbornet, I pushed a repro scenario here: https://github.com/marcoreni/openapi-generator/tree/repro/910 

You can see the issue by launching samples\client\echo\java\issue910\src\test\java\org\openapitools\client\api\Issue910Test

cbornet: No I mean we register the `JavaTimeModule` but we don't call `findAndRegisterModules`

marcoreni: No, that's not mandatory. I added it because it would allow to autoregister other modules that may be added in the future.

cbornet: If it's not mandatory I would remove it. And give the possibility to customize the ObjectMapper to the client if needed.

marcoreni: OK, I'll push a fix soon. Do you think that `mapper.registerModule()` should be applied only if the ObjectMapper has NOT been provided ( https://github.com/OpenAPITools/openapi-generator/blob/ab803e59e3c97f329800a1d520dea76e9307a792/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache#L92 )?

cbornet: Indeed. I would create a default mapper if none is provided and otherwise take the mapper given by the user as is assuming he knows what he does.

marcoreni: Then I think we should also move
https://github.com/OpenAPITools/openapi-generator/blob/ab803e59e3c97f329800a1d520dea76e9307a792/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache#L131
and
https://github.com/OpenAPITools/openapi-generator/blob/ab803e59e3c97f329800a1d520dea76e9307a792/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache#L134

Will push soon.

jmini:
Please update `samples/client/petstore/java/webclient` to reflect the latest template changes.

cbornet:
LGTM

PR #923, opened 2018-08-29
[GO] add application/octet-stream content-type support in request

xiachengjia:
### PR checklist

### Description of the PR


wing328:
@xiachengjia thanks for the PR. I've updated the title to replace Python with Go instead

cc @antihax (2017/11) @bvwells (2017/12) @grokify (2018/07)

modules/openapi-generator/src/main/resources/go/client.mustache
```diff
@@ -306,17 +306,22 @@ func (c *APIClient) prepareRequest(
 	return localVarRequest, nil
 }
 
-func (c *APIClient) decode(v interface{}, b []byte, contentType string) (err error) {
+func (c *APIClient) decode(v interface{}, reader io.ReadCloser, contentType string) (err error) {
```
antihax: the body must always close regardless of content type or it will leak file handles, so we likely need a catch all for any other case. For the case of streams, should make it clear in the docs that the developer must close the body later.

xiachengjia: yes, but this template was hard to do it
I thing add middleware to cache response is best way e.g
if I want origin response:  
`rsp, err := client.MyTag.SomeRequest(context, data_body).Response()`
if I want package level response
`rsp, err := client.MyTag.SomeCall(context, data_body).Process()`

wing328: > yes, but this template was hard to do it

Can you elaborate on the challenge? We can add an additional mustache tag for your use case.

wing328:
Closing this as there's no update. Please feel free to reopen or submit a new one when the change is ready for review.

PR #787, opened 2018-08-11
Set swagger-parser to version 2.0.4

jmini:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.0.x`. Default: `master`.
- [x] Copied the technical committee: @OpenAPITools/generator-core-team 

### Description of the PR

With #696 swagger-parser was updated to a custom version.
This PR set it to the official `2.0.2` version (see https://github.com/swagger-api/swagger-parser/issues/780)

PR #951, opened 2018-09-01
Update swagger-parser to v2.0.3-OpenAPITools.org-1

jmini:
### PR checklist

### Description of the PR

With our `3.3.0` release we would like to update Swagger-Parser. 

Version [`2.0.3-OpenAPITools.org-1`](https://github.com/OpenAPITools/swagger-parser/releases/tag/v2.0.3-OpenAPITools.org-1) was published and is intended to be used until the official `2.0.3` is published (see #787)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1134,7 +1134,7 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <swagger-parser-version>2.0.2-OpenAPITools.org-1</swagger-parser-version>
+        <swagger-parser-version>2.0.3-OpenAPITools.org-1</swagger-parser-version>
         <swagger-core-version>2.0.1</swagger-core-version>
         <scala-version>2.11.1</scala-version>
         <felix-version>3.3.0</felix-version>


jmini:
For a reason that I ignore, the new swagger-parser version has an impact on `samples/server/petstore/rust-server/output/rust-server-test/`. Samples updated with  d3fd57d

CC: @farcaller @frol, @bjgill

jmini:
Following change was observed when no `host` is defined in the OASv2 spec like [rust-server-test.yaml](https://github.com/OpenAPITools/openapi-generator/blob/8a034ce06388c08f09ce2552986224cc55d05ed1/modules/openapi-generator/src/test/resources/2_0/rust-server/rust-server-test.yaml) 

```diff
--- with version 2.0.2-OpenAPITools.org-1
+++ with version 2.0.3-OpenAPITools.org-1
openapi: 3.0.1
info:
  title: rust-server-test
  description: This spec is for testing rust-server-specific things
  version: 1.0.0
+servers:
+- url: /
paths:
  /dummy:
    get:
      summary: A dummy endpoint to make the spec valid.
      responses:
        200:
          description: Success
          content: {}
components:
  schemas:
    additionalPropertiesObject:
      type: object
      additionalProperties:
        type: string
      description: An additionalPropertiesObject
      example: foo
```

A new server containing "/" is defined. 

--

Having no `host` seems to be valid according to [this page](https://swagger.io/docs/specification/2-0/api-host-and-base-path/):

> If host is not specified, it is assumed to be the same host where the API documentation is being served.

The new value seems correct to me ([OAS3 specification server object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#serverObject)) 

>  This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the OpenAPI document is being served.

The change in the generated code might be unexpected now, but I think that we will need to fix it.

jmini:
@bjgill: I have the feeling you are mixing basepath and host.

In OASv2 you need to provide 3 elements:

```yaml
    host: localhost:8080
    basePath: /v2
    schemes:
      - http
```

The result is one server entry in an OASv3.

---

In the `rust-server-test.yaml` used as input to generate the sample `host` is not defined.

In previous Swagger-Parser version: not having an `host` resulted in not having a server entry after the conversion from OASv2 to OASv3.

In my opinion they have improved the conversion.

---

Notice that the discussion is about checking if Swagger-Parser is doing a better job now or if they have introduced a bug. This should be the criteria to accept or not the Pull-Request.

What we do based on a given input is up to our project and might need some fixes (can be made in separated PR).

jmini:
> What's the urgency on this? Is this something you were planning to fix? I could probably have a look sometime later this week.

Doing a Swagger-Parser update with a minor release (second digit in the version number) is a good practice.

Other users are waiting for other fixes in the parser.

---

The discussion was triggered here because, while I was updating the lib, I have noticed some changes in the generated rust sample and I was trying to figure out if this is due to a swagger-parser issue (like a regression).

---
For me the discussion pointed out that the rust-generator might need to better handle cases like (OASv3):

```yaml
openapi: 3.0.1
info:
  title: test
  description: Test API
  version: 1.2.3
servers:
  - url: /
```

Independently of this Swagger-Parser update.

---

With the new parser version the same case in OASv2 can now be also considered:

```yaml
swagger: '2.0'
info:
  description: Test API
  version: 1.2.3
  title: test
schemes:
- http
```

This was not possible before due to an issue in the parser.

---

Keep in mind that this is a sort of corner case. Most of the spec out there will define a host (OASv2) or an absolute server location (OASv3). Not fixing this case right now is not a big deal at all.

---

@bjgill:

This is why I hope you agree with this PR that does the Swagger-Update.
A feedback from your side would be appreciated.

wing328:
> Keep in mind that this is a sort of corner case. Most of the spec out there will define a host (OASv2) or an absolute server location (OASv3). Not fixing this case right now is not a big deal at all.

I agree with you that we can move forward with this PR

wing328:
> Keep in mind that this is a sort of corner case. Most of the spec out there will define a host (OASv2) or an absolute server location (OASv3). Not fixing this case right now is not a big deal at all.

I agree with you that we can move forward with this PR

bjgill:
Was busy writing a comment agreeing, but clearly not fast enough. 

I'll have a look at getting rust-server fixed up, then.

PR #959, opened 2018-09-04
Use elm-format to format Elm API client

wing328:
### PR checklist

### Description of the PR

If ELM_FORMAT_PATH (environment variable) is set to `/usr/local/bin/elm-format`, Elm generator will format the Elm code automatically.

cc @trenneman


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -48,21 +54,20 @@
 import java.util.Set;
 import java.util.TreeSet;
 
-
 public class ElmClientCodegen extends DefaultCodegen implements CodegenConfig {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DartClientCodegen.class);
```
trenneman: `ElmClientCodegen.class`

wing328: Fixed. Sorry for the copy-and-paste error.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -624,4 +633,44 @@ private void addEncoderAndDecoder(Map<String, Object> vendorExtensions, String d
...
+        if (file == null) {
+            return;
+        }
+
+        String elmFmtPath = System.getenv("ELM_FORMAT_PATH");
+        if (StringUtils.isEmpty(elmFmtPath)) {
+            return; // skip if ELM_FORMAT_PATH env variable is not defined
+        }
+
+        // only procees the following type (or we can simply rely on the file extension to check if it's a elm file)
```
trenneman: I'd skip this part and just check for `.elm`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -624,4 +633,44 @@ private void addEncoderAndDecoder(Map<String, Object> vendorExtensions, String d
...
+                        "api-test",
+                        "api"));
+        if (!supportedFileType.contains(fileType)) {
+            return;
+        }
+
+         // only process files with elm extension
+        if ("elm".equals(FilenameUtils.getExtension(file.toString()))) {
+            // currently only support "elm-format -w yourcode.elm"
+            String command = elmFmtPath + " --yes " + file.toString();
```
trenneman: I don't think we need `--yes`.
I think we do need `--elm-version=0.18` for Elm 0.18 (see https://github.com/avh4/elm-format)

samples/client/petstore/elm-0.18/src/Data/Order_.elm
```diff
@@ -65,20 +63,21 @@ orderEncoder model =
 statusDecoder : Decoder Status
 statusDecoder =
     Decode.string
-        |> Decode.andThen (\str ->
-            case str of
-                "placed" ->
-                    Decode.succeed Placed
+        |> Decode.andThen
```
trenneman: This I will update in the templates as well (no need to that in this PR as more templates will change)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -160,6 +165,10 @@ public ElmClientCodegen() {
     public void processOpts() {
         super.processOpts();
 
+        if (StringUtils.isEmpty(System.getenv("ELM_FORMAT_PATH"))) {
```
trenneman: Do we need a system variable for this? Can we not run `elm-format` from the PATH?

trenneman:
I think we should also add `elm-format --validate` to the CI as well to validate it was run.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
wing328: Sure. I can remove that, which should have no impact on the result.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
wing328: I've to keep `--yes`. Otherwise, the command execution won't complete

trenneman: OK, and it seems `elm-format` did not succeed in your last commit. Are you running the latest version?

bin/elm-petstore-all.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+./bin/elm-0.18-petstore.sh
```
trenneman: I think we should also add elm-format --validate to the CI as well to validate it was run.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
trenneman: I feel it makes more sense to use it if it is on the PATH, but if this is consistent with other generators I would keep it this way.

bin/elm-petstore-all.sh
wing328: Sure. I'll do that in another PR by updating the CI config

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
wing328: Updated via d3183af

bin/elm-petstore-all.sh
trenneman: OK, great

trenneman:
@wing328 If you create another PR for validating the CI this one looks good to go.
Though I still feel we should default to using `elm-make` directly from your PATH. But that's probably not an Elm-specific discussion.

PR #963, opened 2018-09-04
[rust-server] Cope with base paths with trailing '/'

bjgill:
### PR checklist

- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @frol @farcaller 

### Description of the PR

Since we're concatenating URL path segments as strings, we need to strip any trailing '/'. 

Of course, the correct fix would be to use the `Url` type and perform the concatenation correctly. This is just a quick fix to ensure that `rust-server` stops mangling the URL when no base_path is provided (we've been exposed by the recent update to swagger-parser - see #951).

I've also renamed `base_path` to `base_url` in a file where it was actually a URL being referenced, and tidied up some imports.

The renaming is a change to the public API. However, it's fully back-compatible.


bjgill:
#967 does this better

PR #966, opened 2018-09-04
Generate a Gemfile.lock for the Ruby client

ndbroadbent:
This keeps the versions consistent when a developer is testing the client locally or running some automated tests on CI.

I ran `./bin/ruby-client-petstore.sh`, and it produces an identical `Gemfile.lock` at `samples/client/petstore/ruby/Gemfile.lock`.


wing328:
cc @cliffano (2017/07) @zlx (2017/09)

PR #850, opened 2018-08-20
[JAXRS] use contextPath variable for RestApplication templates

michbeck100:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.3.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

The mustache templates for the JAXRS application define `/` in the `@ApplicationPath` annotation. This PR replaces this with `{{{contextPath}}}`, which contains the path of the server url. If no additional path is given and the server url contains no trailing / the value is an empty string. This will be replaced again with / by JAXRS.

PR #967, opened 2018-09-04
Trim trailing '/' from basePathWithoutHost

bjgill:
### PR checklist

### Description of the PR

`basePath` has its trailing `/` trimmed if present. This PR does the same thing for `basePathWithoutHost` as well. This fixes the common use of `{{{basePathWithoutHost}}}{{{path}}}` in the presence of trailing `/`.

This is now necessary as a result of #951, which causes `basePathWithoutHost` to otherwise default to `/`. This breaks `rust-server` (and probably other generators as well), which assumes that it can safely perform a string concatenation of `basePathWithoutHost` and `path`.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -193,8 +193,8 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
         }
 
         URL url = URLPathUtils.getServerURL(openAPI);
-        contextPath = config.escapeText(url.getPath());
-        basePathWithoutHost = contextPath.replaceAll("/$", ""); // for backward compatibility
+        contextPath = config.escapeText(url.getPath()).replaceAll("/$", ""); // for backward compatibility
+        basePathWithoutHost = contextPath;
         basePath = config.escapeText(URLPathUtils.getHost(openAPI)).replaceAll("/$", "");
     }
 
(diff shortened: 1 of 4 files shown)


jmini:
I have opened https://github.com/OpenAPITools/openapi-generator/pull/997 to add the discussed additional validation.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -194,14 +194,8 @@ private void configureGeneratorProperties() {
...
         contextPath = config.escapeText(url.getPath());
-        basePathWithoutHost = contextPath; // for backward compatibility
-        basePath = config.escapeText(URLPathUtils.getHost(openAPI));
-        if ("/".equals(basePath.substring(basePath.length() - 1))) {
-            // remove trailing "/"
-            // https://host.example.com/ => https://host.example.com
-            basePath = basePath.substring(0, basePath.length() - 1);
-        }
-
+        basePathWithoutHost = contextPath.replaceAll("/$", ""); // for backward compatibility
```
wing328: @bjgill for `contextPath`, I think you will need to do the same by removing the trailing "/". I remember a few generators are still using `contextPath` in the templates

jmini:
a5b6486 (removing trailing slash from `contextPath`) impacts the `@ApplicationPath` value in some JAXRS example.

See https://github.com/OpenAPITools/openapi-generator/pull/850#issuecomment-414937786 for reference.

I think the samples should be regenerated (and the java files containing the change need to be deleted before generation)

wing328:
@bjgill thanks again for the PR, which is included in the v3.3.0 minor release: https://twitter.com/oas_generator/status/1046941449609068544

PR #987, opened 2018-09-06
Fix issue with Ruby client where strings from example properties are not wrapped with quotes

ndbroadbent:
This fixes an issue with examples for the Ruby client, where example strings are not properly quoted.

Before, it would generate code like this:

```
order_id = order_id123 # String | 
```

After this change, the example string is wrapped with single quotes:

```
order_id = 'order_id123' # String | 
```

This change might be useful for some other languages too, but not sure. Maybe I should update the code in `DefaultCodegen`.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
@@ -18,6 +18,8 @@
 package org.openapitools.codegen.languages;
 
 import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.parameters.Parameter;
+import io.swagger.v3.oas.models.examples.Example;
 import org.apache.commons.lang3.StringUtils;
 import org.openapitools.codegen.*;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


ndbroadbent:
I'm working on a Python client now, and it has the same issue with unquoted example strings. I was also able to fix this by overriding the `setParameterExampleValue` method. I have a class that extends `PythonClientCodegen`, and dropped this code at the bottom:

```python

  /**
   * Return the example value of the parameter. - Fixes issue with unquoted
   * strings from x-example
   *
   * @param codegenParameter Codegen parameter
   * @param parameter        Parameter
   */
  public void setParameterExampleValue(CodegenParameter codegenParameter, Parameter parameter) {
    if (parameter.getExample() != null) {
      codegenParameter.example = parameter.getExample().toString();
    } else if (parameter.getExamples() != null && !parameter.getExamples().isEmpty()) {
      Example example = parameter.getExamples().values().iterator().next();
      if (example.getValue() != null) {
        codegenParameter.example = example.getValue().toString();
      }
    } else {
      Schema schema = parameter.getSchema();
      if (schema != null && schema.getExample() != null) {
        codegenParameter.example = schema.getExample().toString();
      }
    }

    setParameterExampleValue(codegenParameter);
  }
```

Maybe this should go into `DefaultCodegen`?

wing328:
> Maybe this should go into DefaultCodegen?

I would prefer keeping the [existing implementation in DefaultCodegen](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1122) and let generator override it to meet its requirement.

wing328:
@ndbroadbent thanks for the PR, which has been included in the v3.3.3 release: https://twitter.com/oas_generator/status/1062929948191510528

PR #988, opened 2018-09-06
[Java][Feign] Add support for Feign 10 (#977)

### LINKED ISSUE #977: Java Feign template doesn't work with Feign 10
Hello,

Currently Java Feign template generates code for Feign 9.4.0. 
I'd like the generated client to work with Feign 10.0.1 since I already use that version in my project.
However that not possible currently because of [breaking change in Feign](https://github.com/OpenFeign/feign/releases/tag/10.0.1): 
> #744 [Core] Added HttpMethod to RetryableException

https://github.com/OpenFeign/feign/pull/744/files#diff-64c69768b214213c25fd1246bae90adaR33 - `RetryableException` constructor wants HTTP method now.

Looks like it's only thrown here in this file: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/feign/auth/OAuth.mustache#L95

So I had to overwrite original template with my own to generate working code right now.

But is it possible to update the template to support newer Feign out of box?

ihrankouski:
### PR checklist

### Description of the PR

Fixes #977

### DIFF
--- a/bin/java-petstore-feign.json
+++ b/bin/java-petstore-feign.json
@@ -1,4 +1,7 @@
 {
     "library": "feign",
-    "artifactId": "petstore-feign"
+    "artifactId": "petstore-feign",
+    "additionalProperties": {
+        "useFeign10": true
+    }
 }
--- a/docs/generators/java.md
+++ b/docs/generators/java.md
@@ -137,7 +137,7 @@ CONFIG OPTIONS for java
 	library
 	    library template (sub-template) to use (Default: okhttp-gson)
 	        jersey1 - HTTP client: Jersey client 1.19.4. JSON processing: Jackson 2.8.9. Enable Java6 support using '-DsupportJava6=true'. Enable gzip request encoding using '-DuseGzipFeature=true'.
-	        feign - HTTP client: OpenFeign 9.4.0. JSON processing: Jackson 2.8.9
+	        feign - HTTP client: OpenFeign 9.4.0. JSON processing: Jackson 2.8.9. Enable OpenFeign 10.0.1 using '-DuseFeign10=true'.
 	        jersey2 - HTTP client: Jersey client 2.25.1. JSON processing: Jackson 2.8.9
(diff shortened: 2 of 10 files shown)


cbornet:
Can't we just upgrade to feign 10 and not add an option ? From what I see there is no change to the API from a user perspective so it stays backward compatible.

ihrankouski:
@cbornet there's a breaking change in 10: constructor of `feing.RetryableException` now requires HTTP method as argument.

cbornet:
Yes but it's not visible to the SDK user

jmini:
> Can't we just upgrade to feign 10 and not add an option ?

Well @wing328 suggested to do it with an option https://github.com/OpenAPITools/openapi-generator/issues/977#issuecomment-419006428

---

I think we have no clear rule, when we can just update the library used by the generated code, or when we need to do it with switches

In general, each time we support both versions of a lib with a switch, everything gets more completed (the templates, the testing effort, the options list that the user can set).

I do not know this specific library, but I would tend to think that if the `9.x` stream of Feign is no longer supported, we can update without supporting an old version in the generator.

The other point explained by @cbornet is absolutely correct: 
if someone who has generated code with OpenAPI-Generator `3.2.3` (relying on `9.4` of Feign) has nothing to change in its code (built on top of the generated code) when he switches to a new version of OpenAPI-Generator (relying on Feign `10.0.1`) then the need for a flag is really small.

wing328:
Right. That's my suggestion. https://github.com/OpenFeign/feign/releases/tag/release-10.0.0 was released less than a month ago so I think a majority of the users are still using the Feign 9.x.

Given that Feign 10.x no longer supports JDK 7 if I understand it correctly, I prefer to have an option to keep JDK 7 users happy

wing328:
> Then maybe keep v9 only for JDK7 ?

Yup, the current approach still uses v9 by default. When more users have switched to v10, we will change the default to use v10.

wing328:
@ihrankouski thanks for the contribution, which has been merged into master.

wing328:
@ihrankouski thanks again for the PR, which is included in the v3.3.0 minor release: https://twitter.com/oas_generator/status/1046941449609068544

Issue #994, opened 2018-09-07
[Java] InputStream is preferable to File from oas v3 with format binary


fujigon:
##### Description

Currently, `File` type is employed in generated codes from oas v3 with format binary.
I think InputStream type is more suitable for it.

I locally test with jersey, and found that jersey accepts `File` type as request body entity, but it internally writes the binary stream to local disk, and passes the wrote file to the resource class argument.
It is not good for performance, so I think we should change type into `InputStream`

##### openapi-generator version

current master (abe741f)

##### OpenAPI declaration file content or url
https://gist.github.com/fujigon/dd00a99b79a2bc811761b29e63e08a39#file-openapi-yaml

##### Command line used for generation

```
# git clone, etc
mvn package
java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i openapi.yaml -g jaxrs-jersey
```

##### Steps to reproduce
execute the command, and it successfully runs.
I get
- https://gist.github.com/fujigon/dd00a99b79a2bc811761b29e63e08a39#file-generated_testsapi-java
- https://gist.github.com/fujigon/dd00a99b79a2bc811761b29e63e08a39#file-generated_testsapiservice-java
and so on.

##### Suggest a fix/enhancement
https://github.com/fujigon/openapi-generator/tree/feature/fix-primitive-type-resolve

ndjensen:
I would like to +1 this idea.  My scenario is slightly different in that I could use an InputStream on the Java client side.  The client generated method returns a File object.  I want the bytes to be streamed and read in, so it's inefficient to write the bytes to disk then immediately read them back into memory.

ajeans:
This seems to be related to #6715 
`File` is replaced with `Resource` for the `resttemplate`

Cloudmersive:
Any update this?

PavelZamriy:
In my case I also have a problem with the client side (spring-mvc) that expects File for `format: binary` but I have a spring error because of the `content-type: application/pdf` 
_"Content type 'application/pdf' not supported for bodyType=java.io.File_

sebastianblesgen:
In my case I also have a problem with the client side (spring-mvc) that expects File for format: binary but I have a spring error because of the content-type: application/pdf
"Content type 'application/pdf' not supported for bodyType=java.io.File

Drevsh:
Same here, seems like the webclient generation has a problem. The generated code tries to convert the response into a file object but fails:


```
[...]

PR #995, opened 2018-09-07
[WIP] InputStream is preferable as the format binary's type in java

fujigon:
### PR checklist

@wing328

### Description of the PR
This patch fixes https://github.com/OpenAPITools/openapi-generator/issues/994


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -888,67 +888,6 @@ public String toApiImport(String name) {
      * returns string presentation of the example path (it's a constructor)
      */
     public DefaultCodegen() {
-        defaultIncludes = new HashSet<String>(
```
fujigon: these type mappings are java specific.
I moved to AbstractJavaCodegen

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1339,10 +1278,10 @@ private static String getPrimitiveType(Schema schema) {
             return "BigDecimal";
         } else if (ModelUtils.isByteArraySchema(schema)) {
             return "ByteArray";
-        } else if (ModelUtils.isFileSchema(schema)) {
```
fujigon: `isFileSchema` method returns true even if schema is binary. 
so in order to detect "binary", place `isBinarySchema` sentence prior to `isFileSchema`.

fujigon: cf. https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L461

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -108,6 +108,65 @@
 
     public AbstractJavaCodegen() {
         super();
+
+        defaultIncludes = new HashSet<String>(
```
fujigon: moved

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -150,8 +209,6 @@ public AbstractJavaCodegen() {
         );
         instantiationTypes.put("array", "ArrayList");
         instantiationTypes.put("map", "HashMap");
-        typeMapping.put("date", "Date");
```
fujigon: moved

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -150,8 +209,6 @@ public AbstractJavaCodegen() {
...
-        typeMapping.put("file", "File");
```
fujigon: duplicated, so deleted

modules/openapi-generator/src/test/java/org/openapitools/codegen/swift3/Swift3ModelTest.java
```diff
@@ -98,10 +98,10 @@ public void simpleModelTest() {
 
         final CodegenProperty property4 = cm.vars.get(3);
         Assert.assertEquals(property4.baseName, "binary");
-        Assert.assertEquals(property4.dataType, "URL");
+        Assert.assertEquals(property4.dataType, "Data");
```
fujigon: because changing https://github.com/OpenAPITools/openapi-generator/pull/995/files#r215861936, need to fix test.
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift3Codegen.java#L142
says that  "binary" returns "Date", so I thinks this change is reasonable.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
wing328: Thanks for the explanation. In OAS3, `file` type (in OAS2) has been deprecated and replaced with `type: string, format: binary` instead.

fujigon: Thanks for reply.
Yes, I known it by https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L461 . 

The problem is that return type is "File" from the aspect of performance, for `file` type (in OAS2) and `type: string, format: binary` (in OAS3). cf. https://github.com/OpenAPITools/openapi-generator/issues/994 description.

So I think there're 2 strategies.

1. distinguish BinarySchema and FileSchema
In this case, my patch works.
For backward compatibility, ModelUtils' `isFileSchema` still should detect not only "FileSchema " but also "BinarySchema", as you mention and https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L461 says.
Then change type mapping of both `binary` (https://github.com/OpenAPITools/openapi-generator/pull/995/files#diff-f7863d157cb04293ab52ff741f420c0cR146) and `file` (https://github.com/OpenAPITools/openapi-generator/pull/995/files#diff-f7863d157cb04293ab52ff741f420c0cR147) to `ImputStream`

2. handle BinarySchema and FileSchema samely
In this case, we should merge `isFileSchema` method and `isBinarySchema` method into 
`isFileSchemaOrBinarySchema`.

```
else if (ModelUtils.isFileSchemaOrBinarySchema(schema)) {
  return "binary"
}
// delete isFileSchema sentence and isBinarySchema sentence
```

and then keep https://github.com/OpenAPITools/openapi-generator/pull/995/files#diff-f7863d157cb04293ab52ff741f420c0cR146.

I feel like 1 is preferable. 
How do you think?

dkellenb:
Any chance that this MR gets updated?

jvs64893:
Is this still relevant or do we need something adapted to #6715?

sliw39:
Hi ! 
Any chance this issue will be resolved ?
#6715 only applies to spring's webclients library.
As a result, I still have to use a temp file to be able to upload my data as form multipart data (using "java" generator)

PR #997, opened 2018-09-08
[validation] add error if path do not start with a slash

jmini:
### PR checklist

- [x] Copied the technical committee: @OpenAPITools/generator-core-team  @OpenAPITools/openapi-generator-collaborators (change in core)

### Description of the PR

From https://github.com/OpenAPITools/openapi-generator/pull/967#issuecomment-418602467

This spec is not valid:

```yaml
openapi: '3.0.1'
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/api/'
paths:
  ping/me:
    get:
      operationId: pingGet
      responses:
        '201':
          description: OK
```

Because the path `ping/me` do not conform to this rule:

> The field name MUST begin with a slash.

See the [OpenAPI Specification 3.0.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#patterned-fields).

Some editors like [Swagger online editor](https://editor.swagger.io/) or the [KaiZen-Editor](https://github.com/RepreZen/KaiZen-OpenApi-Editor/) are reporting errors, but Swagger-Parser do not.

This PR add additional errors to the Swagger-Parser error message. If the user decide to by-pass the validation he might get invalid generated code.

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Validate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Validate.java
@@ -46,6 +46,7 @@ public class Validate implements Runnable {
         SwaggerParseResult result = new OpenAPIParser().readLocation(spec, null, null);
         List<String> messageList = result.getMessages();
         Set<String> errors = new HashSet<String>(messageList);
+        errors.addAll(ModelUtils.getErrorMessages(result.getOpenAPI()));
         Set<String> warnings = new HashSet<String>();
 
         StringBuilder sb = new StringBuilder();
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
@@ -543,6 +543,7 @@ public class CodegenConfigurator implements Serializable {
 
         Set<String> validationMessages = new HashSet<>(result.getMessages());
         OpenAPI specification = result.getOpenAPI();
+        validationMessages.addAll(ModelUtils.getErrorMessages(specification));
 
         // NOTE: We will only expose errors+warnings if there are already errors in the spec.
         if (validationMessages.size() > 0) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -692,4 +692,21 @@ public class ModelUtils {
         }
         return null;
     }
+
+    /**
+     * Compute additional error messages that are not detected by the parser
+     * @param openAPI specification being checked
+     * @return a list of errors (items that are not conform)
+     */
+    public static List<String> getErrorMessages(OpenAPI openAPI) {
+        List<String> errors = new ArrayList<>();
+        if (openAPI.getPaths() != null) {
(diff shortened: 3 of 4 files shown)


jmini:
> Is there some reason why we can't fix this upstream in swagger-parser?

No, I like the idea. I did not think about it.

Lets add this now in OpenAPI-Generator, file an issue and PR in Swagger-Parser. If it is accepted there, we can revert this change when we do the next Swagger-Parser Update in our code.

jimschubert:
> Is there some reason why we can't fix this upstream in swagger-parser, however?

The fact that this doesn't exist in swagger-parser means that adding it could potentially break many consumers. I'd be surprised if swagger-parser accepted this before a major release version.

jimschubert:
@jmini I've labeled this as breaking change without fallback because any users with invalid paths objects in their spec will no longer be able to generate code. This could happen if one has created a custom generator which resolves paths and automatically prefixes them, or if the user has written custom template specific to their invalid spec (in which case, our recommendation to add the preceding slash will invalidate the custom templates).

Is your intention for this to be a hard error, and therefore a breaking change?

Is there reasoning that:
* we shouldn't allow an override to ignore errors?
* we shouldn't just prepend the forward slash on any path which is missing the required slash?

jmini:
@jimschubert thank you for this valuable feedback.

---

> The fact that this doesn't exist in swagger-parser means that adding it could potentially break many consumers.

For me, it more means that nobody has brought the case before. From what I see, this lib is more issue-driven (stuff is improved when someone submit an issue and a PR).

I am not sure that they decided on purpose to support a feature or not or in this case to report an error or not. Checking for the rule "Paths must begin with a `/`" might just not be implemented yet in the Swagger-Parser.

In fact the kind of error message they report, are more technical (the parser was failing somewhere) than for an input not being conform to the OpenAPI Specification.

---

I am more interested by the way OpenAPI-Generator should react in case of an input to being invalid, like this one:

```yaml
openapi: '3.0.1'
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/api'
paths:
  ping/me:
    get:
      operationId: pingGet
      responses:
        '201':
          description: OK
```

I did not check for each generator provided by the project, but it is easy to find at least some of them that are generating something that do not work out of this input.

I have proposed to add a validation, because I had the impression that this is the pattern you recommended here https://github.com/OpenAPITools/openapi-generator/pull/251#issue-193491385:

> I would also consider this a bug fix, because as a user I would be surprised to find that generation succeeds when my spec is invalid by default.

Now if I understood you correctly, in this specific case you would prefer a solution where we correct the spec by our self treating the path as it was named `/ping/me` (which would be the correct form in this case). 

---

No problem for me to rewrite the PR in this direction. I do not really have a strong intention or opinion for this PR and I am open for better solutions.

Please confirm first that I understood your message correctly.

jimschubert:
@jmini I'm sorry if my previous comment was confusing. To clarify, I would think this should be implemented similarly to the issue you've linked:

* default to the "correct" behavior
* allow users to override this behavior, i.e. skipping errors
* recommendation: if behavior is overridden or not, we should normalize the path to avoid inconsistencies with generators and templates

jmini:
@jimschubert:

> if behavior is overridden or not, we should normalize the path to avoid inconsistencies with generators and templates

I have opened #1034 for this specific part (is not a breaking change, can be added to `3.3.0`).

jimschubert:
Closing this as I implemented it locally only to realize that swagger-parser also  results in an error now.

Issue #989, opened 2018-09-06
CLI option to control logging level


skydoctor:
##### Description

There doesn't seem to be a way to control the verbosity of the output logs printed by the generator-cli, even though each line has a level defined. There should be a way to limit the output logs to errors or warnings only as desired.

##### openapi-generator version

3.1.2

##### OpenAPI declaration file content or url

N/A

##### Command line used for generation

java -jar /usr/share/java/openapi-generator-cli-3.1.2.jar generate -t cpprest -g cpp-restsdk

##### Steps to reproduce

Run any CLI command to generate code from yaml/json

ackintosh:
Thanks for reporting this issue. I'll file a PR later.

ackintosh:
@skydoctor 
The enhancement will be released on Oct 1st (v3.3.0). Thanks for reporting this issue!

kimamula:
Is this available with the docker image?
I tried the following command which does not work.

```sh
$ docker run --rm -v ${PWD}:/local -e JAVA_OPTS="-Dlog.level=error" \
openapitools/openapi-generator-cli:v3.3.2 generate \
-i /local/openapi.yaml \
-g typescript-angular \
-o /local/out \
-c /local/openapitools/config.json
```

ackintosh:
hmm...

As @kimamula said, the option does not work on the docker image. 👀 

On the other hand, it works on the image which is built locally. 🤔 

```bash
$ cd openapi-generator
$ docker build -t openapi-generator-test .
...
$ docker run --rm -v (PWD):/local -e JAVA_OPTS="-Dlog.level=warn" \
      openapi-generator-test generate \
      ...
```

kimamula:
I have created [a new issue](https://github.com/OpenAPITools/openapi-generator/issues/1410) for this problem.
I am guessing that this is because https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-cli/Dockerfile is not using `JAVA_OPTS` environment variable.

PR #1000, opened 2018-09-09
Add a CLI option to control logging level

ackintosh:
### PR checklist

### Description of the PR

closes https://github.com/OpenAPITools/openapi-generator/issues/989

This PR makes logging level controllable with the property as below.

```
$ java -Dlog.level=error ....
```

### DIFF
--- a/modules/openapi-generator-cli/src/main/resources/logback.xml
+++ b/modules/openapi-generator-cli/src/main/resources/logback.xml
@@ -24,7 +24,7 @@
         <appender-ref ref="STDOUT"/>
         <appender-ref ref="STDERR"/>
     </logger>
-    <logger name="org.openapitools" level="${log-level:-info}">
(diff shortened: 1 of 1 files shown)


ackintosh:
This PR is not targeting a particular programming language.

wing328:
@ackintosh thanks for the enhancement, which definitely makes it easier to debug OpenAPI Generator (in addition to providing a configuration file)

Instead of using `log_level`, may I suggest naming the system property as `log.level` instead as it seems to be more common to use dots instead of underscores in system property? (but I couldn't find any official style guide for naming system properties)

Let's see if others have comments on this as well.

jimschubert:
I agree that it's probably better as `log.level` so other logger implementations that we might not be wrapping could pick this up.

ackintosh:
Thanks for the suggestion. It looks better to me as well. ✨

Issue #1002, opened 2018-09-09
[Symfony] Make controllers explicitly public


Addvilz:
##### Description

Yet another Symfony4 compatibility fix. 

Since Symfony 3.4, services in container are private by default. Because of that, doing a resource import in root routing (for example, to prefix paths or do something similar) will not work, as the controllers are inaccessible, resulting in an error.

##### Suggest a fix/enhancement

The only backwards compatible fix is to mark these controller services public by default.

I will open a PR to address this issue.

Addvilz:
P.S if anyone is interested, a monkey patch I have been using so far is to manually override controller services in the app services. Works fine for small api's with 1.3 controllers total, not so well if there are 100+...

BenjaminHae:
I think this can be solved by adding the tag `controller.service_arguments` to the services.

PR #1003, opened 2018-09-09
Make Symfony generated controllers explicitly public

Addvilz:
### PR checklist

### Description of the PR

Fix for #1002

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko

### DIFF
--- a/modules/openapi-generator/src/main/resources/php-symfony/services.mustache
+++ b/modules/openapi-generator/src/main/resources/php-symfony/services.mustache
@@ -24,6 +24,7 @@ services:
 {{#operations}}
     {{bundleAlias}}.controller.{{pathPrefix}}:
         class: {{controllerPackage}}\{{baseName}}Controller
+        public: true
         calls:
          - [setSerializer, ['@{{bundleAlias}}.service.serializer']]
          - [setValidator,  ['@{{bundleAlias}}.service.validator']]
--- a/samples/server/petstore/php-symfony/SymfonyBundle-php/Resources/config/services.yml
+++ b/samples/server/petstore/php-symfony/SymfonyBundle-php/Resources/config/services.yml
@@ -21,6 +21,7 @@ services:
 
     open_apiserver.controller.pet:
         class: OpenAPI\Server\Controller\PetController
+        public: true
         calls:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]
@@ -28,6 +29,7 @@ services:
 
     open_apiserver.controller.store:
         class: OpenAPI\Server\Controller\StoreController
+        public: true
         calls:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]
@@ -35,6 +37,7 @@ services:
 
     open_apiserver.controller.user:
         class: OpenAPI\Server\Controller\UserController
+        public: true
         calls:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]


wing328:
@dkarlovi thanks for the feedback. I believe you're referring to https://symfony.com/doc/current/best_practices/business-logic.html:

> Services can also be public or private. If you use the default services.yaml configuration, all services are private by default.
> 
> BEST PRACTICE
> Services should be private whenever possible. This will prevent you from accessing that service via $container->get(). Instead, you will need to use dependency injection.

dkarlovi:
> I believe you're referring to https://symfony.com/doc/current/best_practices/business-logic.html:

Indeed.

> Have you looked into the reason provided here by @Addvilz ?

If that use case doesn't work with private controllers (being a common use case), it looks more like an issue with Symfony or maybe misconfiguration on @Addvilz part. Symfony itself is promoting private controllers, having them not work like this wouldn't make that recommendation viable.

@Addvilz can you provide a small reproducer for this? At least your routing config and the exact error.

Addvilz:
@dkarlovi the reproducer is step1...step 3 of README for symfony generated bundle, in any fresh Symfony 4.x installation.

```
foo:
    prefix: /bar
    host: "{host}"
    requirements:
        host: example.com|other.example.com
    resource: "@SomeGeneratedBundle/Resources/config/routing.yml"
```
This simply does not work in Symfony 4.

I agree however that this is by no means a good fix. But this seems like the only fix that would not break backwards compatibility.

If you have any other way to make this work, I would be happy to implement the fix. I am not really happy about this either.

wing328:
> I agree however that this is by no means a good fix. But this seems like the only fix that would not break backwards compatibility.

To move this forward, may I suggest the following?

- Add logger.warn to warn about the use of public controller
- Show instructions on how to customize the template to make the private controller
- Inform the user that the next major release (4.0.0) will have private controller as the default

I think providing all these information to our users will help understand what we're doing and what's the upcoming change we plan to make.

Addvilz:
I have an alternative dating back from changes in Symfony 3.3, I need to test it for backwards compatibility (relevant: https://symfony.com/doc/current/service_container/3.3-di-changes.html#controllers-are-registered-as-services)

Seems that adding a tag to the service does some magic and allows controller to be used the way we are without making the controller public.

`tags: ['controller.service_arguments']`

If this works as I expect, I will update the PR to reflect the changes.

Addvilz:
Quick thumbs up, I am supper occupied at this time, and can't get to address this properly. If someone want's to take over, feel free and leave a note here, otherwise I will try and address this as soon as I can.

BenjaminHae:
As far as I can tell this can indeed be solved by adding the tag `controller.service_arguments` to the services.

ossinkine:
Symfony Controllers should always be public.
Adding `controller.service_arguments` also solves the issue because it makes tagged controller public: https://github.com/symfony/http-kernel/blob/d7c74c617834c69b5150259c00e8ab0f5f04027a/DependencyInjection/RegisterControllerArgumentLocatorsPass.php#L56

Also `controller.service_arguments` tag also make possible to inject dependency into controller action directly but this is not needed right now.
So I think this MR can be merged.

modules/openapi-generator/src/main/resources/php-symfony/services.mustache
```diff
@@ -24,6 +24,7 @@ services:
 {{#operations}}
     {{bundleAlias}}.controller.{{pathPrefix}}:
         class: {{controllerPackage}}\{{baseName}}Controller
+        public: true
```
dkarlovi: Tagging the controllers with `controller.service_arguments` is the superior solution since it's a tag tailor-made for that use case. This solution goes about it from the wrong side.

ossinkine:
> If adding the tag which is intended for controllers solves the issues, why aren't we adding that tag then, instead of just declaring controllers public?

Symfony requires controllers to be public and we can solve this by adding `public: true` into the config.
`controller.service_arguments` tag solves another issue - dependecy injection directly into action.
Making a controller public is a side effect.

Besides the tag was introduced in Symfony 3.3. Which Symfony versions do we want to support?

dkarlovi:
> Symfony requires controllers to be public and we can solve this by adding `public: true` into the config.
`controller.service_arguments` tag solves another issue - dependecy injection directly into action.
Making a controller public is a side effect.

Yes, but if you have one tag which
1. solves both use cases
2. is officially recommended by Symfony for the task

why would you ever choose a second one just because the first one also does what the second one does? The correct solution is to tag correctly, why choose just "pretty close"?

> Besides the tag was introduced in Symfony 3.3. Which Symfony versions do we want to support?

IMO this tool should support versions which are supported by Symfony themselves. Which is currently 3.4 LTS. As always, if you need support for unsupported Symfony versions, you can use the older version of the tool.

Addvilz:
Closed in favor of #2059.

PR #1009, opened 2018-09-10
fix generation of jaguar api with auth

jaumard:


wing328:
FYI. The Travis test has finished without issues (instead of pending)

https://travis-ci.org/OpenAPITools/openapi-generator/builds/426693014

PR #1016, opened 2018-09-11
Last fixes for jaguar dart

jaumard:


bin/dart-jaguar-petstore.sh
```diff
@@ -41,3 +41,9 @@ java $JAVA_OPTS -jar $executable $ags
 
 # The current petstore test will then work for both: the browser library and the vm library.
 
+cd samples/client/petstore/dart-jaguar/openapi
```
wing328: We'll move these commands to pom.xml instead as this is how we test/build the output.

jaumard: Ho ok ! didn't know that :) `Allow edits from maintainers` is checked you can do it directly here if you want

wing328: Let me merge it into `dart-jaguar` first.

PR #1044, opened 2018-09-16
#1043 - [Ada] Operation security scopes are ignored when generating the server

### LINKED ISSUE #1043: [Ada] Operation security scopes are ignored when generating the server
##### Description

When you declare an operation with a security section, this defines the security models and
scopes which are required by the operation. This information is partially used, and in particular
the operation required scopes are not available to the model generator.

For example, if we define two OpenAPI 3.0 operations with same
security but two different scopes:

```
 paths:
  /agents:
      ...
      security:
        - oauth: ['agent:register']
  /hosts
      security:
        - oauth: ['host:create']
```

and the security scheme would define "oauth":

```
components:
 securitySchemes:
   agent_auth:
     type: oauth2
     flows:
       password:
         tokenUrl:
'https://hyperion.vacs.fr/hyperion/api/v1/oauth/token'
         scopes:
           'agent:register': Register a new monitoring agent
            'host:create': Register and create a host
```

Then, the generated server operation will require that all the scopes are available.
(Instead of only checking for **agent:register** for the first operation and **host:create** for the second.

3.3.0

```
java -jar openapi-generator-cli.jar generate --generator-name ada-server -i hyperion.yaml \
            -DprojectName=hyperion --model-package Hyperion.Rest
```

The information is available internally within the **SecurityRequirement** list associated with each operation. This is not available to model templates.

By updating the language code generator we can access this information from **fromOperation** and create a vendor specific extension attribute that provides the information.  I've done it in two steps:

1. In fromOperation, add a x-scopes attribute that lists the operation security required scopes,
2. In postProcessOperationsWithModels, build a new list of authMethods that only contain the operation required scopes (and not all of them), and put that list in the x-auth-scopes attribute.
3. Update the server templates to use the x-auth-scopes instead of authMethods.

stcarrez:
This fixes #1043 

### PR checklist

### Description of the PR

- Update fromOperation() to keep the operation required scopes for each auth
  method and store that information in the x-scopes vendor extensions attribute
- Update postProcessOperationsWithModels() to process the operation required
  scopes and build a list of authMethods which only contain the required scopes
  for the operation and store these authMethods in the x-auth-scopes attribute.
- Update postProcessAuthMethod() to handle the logic of filtering and building
  the operation authMethod (new instances are created because we must not
  modify the global authMethod definitions)
- Update the Ada server templates to use the x-auth-scopes instead of authMethods
  Add a URL prefix parameter for the Ada server instantiation

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractAdaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractAdaCodegen.java
@@ -24,6 +24,8 @@ import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.media.ArraySchema;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.responses.ApiResponse;
+import io.swagger.v3.oas.models.security.SecurityRequirement;
+import io.swagger.v3.oas.models.security.SecurityScheme;
 import org.openapitools.codegen.CodegenConfig;
 import org.openapitools.codegen.CodegenConstants;
 import org.openapitools.codegen.CodegenModel;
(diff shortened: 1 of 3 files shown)


wing328:
@stcarrez thanks again for the fix, which is included in the v3.3.0 minor release: https://twitter.com/oas_generator/status/1046941449609068544

PR #1070, opened 2018-09-20
Kotlin spring server codegen improvements

herojan:
### PR checklist

@dr4ke616
### Description of the PR

- Fixes a bug where lists were instantiated with the arrayOf function.
- If required and readonly, a field should be optional. This is to allow people to do a GET and easily PUT it back with small changes.
- Use List type for the OpenApi array. This is because kotlin data classes should not use kotlin.Array since arrays are compared by reference rather than structure. Different arrays with the same content should return true but return false.
- Uses String instead of kotlin.String, Int instead of kotlin.Int, etc for built-in kotlin types.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
@@ -173,7 +173,7 @@ public abstract class AbstractKotlinCodegen extends DefaultCodegen implements Co
         typeMapping.put("DateTime", "java.time.LocalDateTime");
 
         instantiationTypes.put("array", "kotlin.arrayOf");
-        instantiationTypes.put("list", "kotlin.arrayOf");
+        instantiationTypes.put("list", "kotlin.listOf");
         instantiationTypes.put("map", "kotlin.mapOf");
 
         importMapping = new HashMap<String, String>();
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
@@ -173,7 +173,7 @@ public abstract class AbstractKotlinCodegen extends DefaultCodegen implements Co
         typeMapping.put("DateTime", "java.time.LocalDateTime");
 
         instantiationTypes.put("array", "kotlin.arrayOf");
-        instantiationTypes.put("list", "kotlin.listOf");
+        instantiationTypes.put("list", "kotlin.arrayOf");
         instantiationTypes.put("map", "kotlin.mapOf");
 
         importMapping = new HashMap<String, String>();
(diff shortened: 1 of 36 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -173,7 +173,7 @@ public AbstractKotlinCodegen() {
         typeMapping.put("DateTime", "java.time.LocalDateTime");
 
         instantiationTypes.put("array", "kotlin.arrayOf");
-        instantiationTypes.put("list", "kotlin.arrayOf");
+        instantiationTypes.put("list", "kotlin.listOf");
```
wing328: This will also impact other Kotline generators (e.g. client).

@jimschubert @dr4ke616, are you ok with that?

samples/server/petstore/kotlin-springboot/src/main/kotlin/org/openapitools/model/Pet.kt
```diff
@@ -20,22 +20,22 @@ import io.swagger.annotations.ApiModelProperty
...
 data class Pet (
 
-        @get:NotNull 
+    @get:NotNull 
         @ApiModelProperty(example = "doggie", required = true, value = "")
-        @JsonProperty("name") val name: kotlin.String,
+        @JsonProperty("name") val name: String,
 
-        @get:NotNull 
+    @get:NotNull
```
dr4ke616: Formatting here looks a little inconsistent

herojan: Fixed that now

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
jimschubert: Array was being used over List because of the serializer. That's since been changed, so this shouldn't be an issue as far as language support is concerned. We could always just generate the client to hit a service created by the Springboot generator to verify.

However, this will break client consumers who have coded to array.

Another thing to consider is that Kotlin arrays are compiled to Java arrays. These are more lightweight than List. So this change may also have an impact on resource limited consumers (Android).

herojan: @jimschubert Ah ok, I'll change it back in that case. Out of interest though is list being used as a type in OpenApi specs? I thought this map was from openapi types but I don't see list anywhere

jimschubert: @herojan these types represent openapi data types as well as other common types that have been normalized in `DefaultCodegen`. I'm on mobile, so it's hard to search GitHub, but if I remember correctly the DefaultCodegen logic (originally targeting mainly Java) treated OpenAPI array types as array of they were primitives or list if not. It's been about a year since I looked through that code, so that's pretty off memory. But, for example, you could have a type `File` and instantiate it with some custom infrastructure type like `MyFileMaker`.

herojan: @jimschubert Ah right that makes sense, good to know. Thanks!

PR #1076, opened 2018-09-21
fix model generator for snake case open api definition

jaumard:


modules/openapi-generator/src/main/resources/dart-jaguar/class.mustache
```diff
@@ -2,6 +2,7 @@ part '{{classFilename}}.jser.dart';
 
 class {{classname}} {
   {{#vars}}{{#description}} /* {{{description}}} */{{/description}}
+  @Alias('{{baseName}}')
```
wing328: @jaumard thanks for the PR. I would suggest using `{{{baseName}}}` as `{{baseName}}` may escape special characters in HTML.

wing328:
cc @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09)

modules/openapi-generator/src/main/resources/dart-jaguar/class.mustache
jaumard: But here we want them to be escaped anyway. It's should be plain text only as it's the name of the JSON field

jaumard:
@wing328 `@Alias` is specific to jaguar for the serialization so I'll say it's not needed for other dart generator

wing328:
@jaumard ah ok. Thanks for the explanation.

modules/openapi-generator/src/main/resources/dart-jaguar/class.mustache
jaumard: No problem :) it's done ^^

Issue #1126, opened 2018-09-27
[C#] Child models in allOf clause not compiling because parent does not implement BaseValidate


rubms:
##### Description

The auto-generated C# models that are children in allOf clauses don't compile because they are attempting to invoke a `BaseValidate` method that should be implemented by the parent object, but is not implemented.

##### openapi-generator version

v3.3.0 (master snapshot)

##### OpenAPI declaration file content or url

```yaml
swagger: '2.0'
info:
  description: "This spec is for testing the wrong generation of childrend models in allOf clauses."
  version: 1.0.0
  title: Child allOf model test case
  license:
    name: Apache-2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
schemes:
  - http
paths:
  /test/:
    get:
      tags:
        - Test
      summary: Test
      description: Test
      operationId: getTest
      responses:
        '200':
          description: successful operation
          schema:
            $ref: '#/definitions/Child'

definitions:
  Parent:
    type: object
    properties:
      parentProperty:
        type: string
  Child:
    allOf:
    - $ref: "#/definitions/Parent"
    - type: object
      properties:
        childProperty:
          type: string

```

##### Command line used for generation

java -jar openapi-generator-cli.jar generate -g csharp -i .\test.yml -o test

##### Steps to reproduce

Auto-generate the OpenAPI specification above using the `csharp` generator. The generated `Child` model does not compile because it uses a `BaseValidate` method that should be implemented in `Parent`, but is not.

##### Suggest a fix/enhancement

The problem is originated in 2 points. 

On the one hand, `DefaultCodegen.java` fails to correctly mark the `hasChildren` property of `CodegenModel` when the model has children:
```java
            // Let parent know about all its children
            for (String name : allModels.keySet()) {
                CodegenModel cm = allModels.get(name);
                CodegenModel parent = allModels.get(cm.getParent());
                // if a discriminator exists on the parent, don't add this child to the inheritance hierarchy
                // TODO Determine what to do if the parent discriminator name == the grandparent discriminator name
                while (parent != null) {
                    if (parent.getChildren() == null) {
                        parent.setChildren(new ArrayList<CodegenModel>());
                    }
                    parent.getChildren().add(cm);
                    parent.hasChildren = true; // <- IT SHOULD MARK THAT THE PARENT HAS CHILDEN HERE
                    if (parent.getDiscriminator() == null) {
                        parent = allModels.get(parent.getParent());
                    } else {
[...]

PR #1127, opened 2018-09-27
[C#] Fixed issue #1126 with invalid generation of C# children models in allOf clauses

rubms:
### PR checklist

### Description of the PR

Fixed the #1126 issue, that causes the auto-generated C# children models not to compile because they attempt to invoke a `BaseValidate` method in the parent model that is not implemented. In order to fix the issue:

- `DefaultCodegen.java` now sets the `hasChildren` of `CodegenModel` to `true` in models that have children. 
- The `modelGeneric.mustache` now uses `hasChildren`, instead of `discriminator`, in order to decide whether to implement a `BaseValidate` method (for children to run the parent model validations). In addition, now the `BaseValidate` method of the parent is explicitly invoked (in order to prevent infinite recursion in models that are both, children and parents.

c.c. Core (as `DefaultCodegen.java` has been modified).
    @wing328 (2015/07)
    @jimschubert (2016/05)
    @cbornet (2016/05)
    @jaz-ah (2016/05)
    @ackintosh (2018/02)
    @JFCote (2018/03)
    @jmini (2018/04)

c.c. C# @mandrean (2017/08) @jimschubert (2017/09)


jimschubert:
@rubms Thanks for the contribution. I've tested it locally with the spec in the linked issue and verified the compilation errors before the fix are now resolved with this code change.

@wing328 before merging, I wanted to get your take on the change. Moving from `discriminator` to `hasChildren` is potentially a breaking change because it would require anyone with customized templates to update their customizations from `discriminator` to `hasChildren`. However, I would assume that anyone affected by this bug "in the wild" would have some local changes to accommodate, so I'm on the fence of whether this is a change to the template contract. What's your take on this being considered more a bug fix than a breaking change? Also, do you think this would potentially need to be fixed across other generators using `discriminator` in the same way? With this change, I feel like the two seemingly identical options will have diverged.

wing328:
@jimschubert as discussed, it's a bug fix but we don't classify it as a breaking change.

PR #1180, opened 2018-10-05
[rust-server] Allow text/html plaintext responses

colelawrence:
### PR checklist

### Description of the PR

Hello @frol, @farcaller, and @bjgill,

This addition to the rust-server generator enables the use of `text/html` responses as plaintext.

@bjgill, I'm sorry I did not add an issue to gather feedback.

I can run shell scripts in the checklist in a few moments

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
@@ -477,6 +477,10 @@ public class RustServerCodegen extends DefaultCodegen implements CodegenConfig {
         return mimetype.toLowerCase(Locale.ROOT).startsWith("text/plain");
     }
 
+    boolean isMimetypeHtmlText(String mimetype) {
+        return mimetype.toLowerCase(Locale.ROOT).startsWith("text/html");
+    }
+
     boolean isMimetypeWwwFormUrlEncoded(String mimetype) {
         return mimetype.toLowerCase(Locale.ROOT).startsWith("application/x-www-form-urlencoded");
     }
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -547,6 +551,8 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
                     consumesXml = true;
                 } else if (isMimetypePlainText(mimeType)) {
                     consumesPlainText = true;
+                } else if (isMimetypeHtmlText(mimeType)) {
```
colelawrence: Consuming html plaintext may not make sense (I don't know what is supposed to happen), we can adjust this.

bjgill:
Thanks for the contribution. This seems like a sensible way of handling HTML initially. We can then integrate a full parser/generator later if needed.

It would, however, be nice to see the effect this PR has - would it be possible to add a sample HTML endpoint to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/rust-server/rust-server-test.yaml? That should give us some confidence that the change has the right effect, as well as making it harder to regress this function in future.

colelawrence:
@bjgill yes, I can add a test for this!

bjgill:
Thanks. I suspect you'll want to add something like:

```yaml
  /html:
    post:
      summary: Test HTML handling
      consumes: [text/html]
      produces: [text/html]
      parameters:
      - in: body
        name: body
        required: true
        schema:
          type: string
      responses:
        200:
          description: Success
          schema:
            type: string
```

bjgill:
I've now got this working (with sample) at https://github.com/Metaswitch/openapi-generator/tree/rust-server-html. Unless you want to fold those changes into this PR, I'll open a new PR tomorrow.

bjgill:
Thanks for your contribution. I've created https://github.com/OpenAPITools/openapi-generator/pull/1329, which adds the sample update.

colelawrence:
Thanks Benjamin!

PR #1197, opened 2018-10-08
[Spring] Add an option to return success code

ackintosh:
### PR checklist

### Description of the PR

~~This PR makes the stub server returns status code defined at Response Object instead of 501(Not implemented).~~

This PR adds an option to make the stub server return success code. ( https://github.com/OpenAPITools/openapi-generator/pull/1197#issuecomment-442300124 )


ackintosh:

samples/client/petstore/spring-stubs/src/main/java/org/openapitools/api/StoreApi.java
```diff
@@ -71,19 +73,20 @@
...
-            for (MediaType mediaType: MediaType.parseMediaTypes(request.getHeader("Accept"))) {
-                if (mediaType.isCompatibleWith(MediaType.valueOf("application/json"))) {
-                    ApiUtil.setExampleResponse(request, "application/json", "{  \"petId\" : 6,  \"quantity\" : 1,  \"id\" : 0,  \"shipDate\" : \"2000-01-23T04:56:07.000+00:00\",  \"complete\" : false,  \"status\" : \"placed\"}");
-                    break;
-                }
-                if (mediaType.isCompatibleWith(MediaType.valueOf("application/xml"))) {
-                    ApiUtil.setExampleResponse(request, "application/xml", "<Order>  <id>123456789</id>  <petId>123456789</petId>  <quantity>123</quantity>  <shipDate>2000-01-23T04:56:07.000Z</shipDate>  <status>aeiou</status>  <complete>true</complete></Order>");
-                    break;
-                }
+        for (MediaType mediaType: MediaType.parseMediaTypes(request.getHeader("Accept"))) {
```
cbornet: Bad indent

modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache
```diff
@@ -1,35 +1,38 @@
...
-    {{#-last}}
-{{#async}}        {{/async}}{{^async}}{{#jdk8}}    {{/jdk8}}{{/async}}        }
-        {{#jdk8}}
-{{#async}}    {{/async}}        });
-        {{/jdk8}}
-{{#async}}    {{/async}}        return new ResponseEntity<>(HttpStatus.NOT_IMPLEMENTED);
-        {{#jdk8}}
-            {{#async}}
+{{#-first}}
+{{#jdk8}}
```
cbornet: Indentation is broken

cbornet:
I'm not sure about this one : why would we pick a specific status code over another ?
At least 501 describes the situation quite well : the endpoint was not implemented yet.

ackintosh:
Thanks for your review!
I'm considering that the case of frontend development (e.g. SPA, mobile app) with a stub server.
In that case, a stub server returns a value defined at spec, but a client interprets the response as an error as the response code is 501.

cbornet:
Then I guess you would rather have a `200` for all endpoints in that case, no ?

cbornet:
We could have an application property where you set the default HTTP code you want to get.

ackintosh:
The following is the spec that I'm working on. The each endpoint differently have the HTTP code a stub server should returns. (`GET /posts` -> 200, `DELETE /posts/{postId}` -> 204)
So I think we need to pick a specific status defined at `responses`. 

```yaml
openapi: 3.0.1
info:
  title: 匿名掲示板API
  version: 1.0.0
paths:
  /posts:
    get:
      description: 投稿をすべて取得する
      responses:
        '200':
          description: 投稿の配列
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Post'
    post:
...
      responses:
        '201':
          description: 作成した投稿
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Post'
        '422':
          $ref: '#/components/responses/UnprocessableEntity'
  /posts/{postId}:
    delete:
      description: 投稿を削除する
      parameters:
        - $ref: '#/components/parameters/postIdParam'
      responses:
        '204':
          description: 投稿の削除に成功
        '404':
          $ref: '#/components/responses/NotFound'
...

```

cbornet:
Is it really important for your client ? Clients should accept any 2xx codes as successful responses  (as per Postel's law)

ackintosh:
As you say from the view of client, it may not important. Client can accept 2xx codes as successful responses.
On the other hand, from the view of a stub server, I think we should return properly codes as possible as.

cbornet:
Ok. Then we would need some kind of "success code" resolution. I wonder if we don't already have that for some gens.

ackintosh:
I've fixed the broken indentation. Please have a look when you have time.

ackintosh:
ping @cbornet 😉

ackintosh:
> I still think we shouldn't send codes like 400 by default. As I see it, OAI response codes are mostly hints for special codes that you want to highlight. 

That's a thought.

---

I've noticed two use cases of the generated server codes:

1. Scaffolding

- I understand that it is the use case you mentioned in https://github.com/OpenAPITools/openapi-generator/pull/1197#issuecomment-427901112 and https://github.com/OpenAPITools/openapi-generator/pull/1197#issuecomment-433313560 .
  - Is my understanding correct?
- The generated codes will be modified to implement application logic and will be able to return correct status code.
  - so generated codes returning `501 Not Implemented` is reasonable.

2. Just for stubbing

- The use case mentioned in [this comment](https://github.com/OpenAPITools/openapi-generator/pull/1197#issuecomment-428128776).
- The generated codes are used for just stubbing in frontend development.
  - When server-side development has finished the stub is replaced with the concrete application.
  - This contract-first approach proceeds frontend and server-side parallely
- In this use case the generated server should return a status code defined in the OpenAPI spec.
  - I think `400` is not problem in this use case if `400` is defined at the spec.
  - If `500`/`400` is inconvenient we can put `200` at top of `responses` definition.

cbornet:
Yes, for scaffolding `501` is better, for stubbing I guess 2xx would be better with resolution of the code if there are codes different than `200` defined in the OAI.

ackintosh:
> for stubbing I guess 2xx would be better with resolution of the code if there are codes different than 200 defined in the OAI.

Good idea. 💡 

👇In my mind, mustache file will be updated like below. Are we on the same page?

`MethodBody.mustache`
```diff
-    return new ResponseEntity<>(HttpStatus.valueOf(HttpStatus.NOT_IMPLEMENTED));
+    return new ResponseEntity<>(HttpStatus.valueOf(
+        ({{#responses}}{{#-first}}{{{code}}}{{/-first}}{{/responses}} >= 400) ? 200 : {{#responses}}{{#-first}}{{{code}}}{{/-first}}{{/responses}}
+    ));
```

cbornet:
I think we could do the resolution in the Java gen (postProcessOperation ?)

ackintosh:
Ah 💡 Thanks!

ackintosh:
Hmm... ideally [`example`](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache#L2) should have its status code. [example has its "content-type" and "response body"](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache#L10-L13).
(related class: [ExampleGenerator.java](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/examples/ExampleGenerator.java))

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2321,7 +2321,13 @@ public CodegenOperation fromOperation(String path,
                     }
 
                     // generate examples
-                    op.examples = new ExampleGenerator(schemas, openAPI).generateFromResponseSchema(responseSchema, getProducesInfo(openAPI, operation));
+                    String exampleStatusCode = "200";
+                    for (String key : operation.getResponses().keySet()) {
+                        if (operation.getResponses().get(key) == methodResponse && !key.equals("default")) {
```
ackintosh: `methodResponse` is ["2xx" or "default"](https://github.com/ackintosh/openapi-generator-1/blob/2e4f484af9906aa78c0f7bd7cb7c470a0de66ac6/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2278).

ackintosh:
Please review the changes when you have time. 🙏

ackintosh:
As be commented at gitter, I'll make this changes an option. Making the changes an option makes sense to preserve backward compat and covers [the two use cases](https://github.com/OpenAPITools/openapi-generator/pull/1197#issuecomment-439052539).

cc @cbornet

ackintosh:
We have 2 ways how make the changes an option: Generation or Spring config. I'm trying to make a "Generation" option as I'm not familiar with Spring. 🤔💦

ackintosh:
#### Test `--additional-properties returnSuccessCode=true`

```
$ bin/springboot-petstore-server.sh --additional-properties returnSuccessCode=true
...
...
$ git diff
```
image

ackintosh:
@cbornet I've updated this PR, make success code optional. Generated server returns 501 as default. Please review 🙏

modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache
```diff
@@ -25,14 +25,14 @@ return CompletableFuture.supplyAsync(()-> {
     {{/-last}}
 {{/examples}}
 {{^examples}}
-return {{#jdk8}}{{#async}}CompletableFuture.completedFuture({{/async}}{{/jdk8}}new ResponseEntity<>(HttpStatus.NOT_IMPLEMENTED){{#jdk8}}{{#async}}){{/async}}{{/jdk8}};
+return {{#jdk8}}{{#async}}CompletableFuture.completedFuture({{/async}}{{/jdk8}}new ResponseEntity<>({{#returnSuccessCode}}HttpStatus.OK{{/returnSuccessCode}}{{^returnSuccessCode}}HttpStatus.NOT_IMPLEMENTED{{/returnSuccessCode}}){{#jdk8}}{{#async}}){{/async}}{{/jdk8}};
```
cbornet: Shouldn't it be `HttpStatus.valueOf({{{statusCode}}})` instead of `HttpStatus.OK`

modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache
```diff
@@ -43,5 +43,8 @@ exchange.getResponse().setStatusCode(HttpStatus.NOT_IMPLEMENTED);
         }
         {{/-last}}
     {{/examples}}
+{{^examples}}
+        exchange.getResponse().setStatusCode({{#returnSuccessCode}}HttpStatus.OK{{/returnSuccessCode}}{{^returnSuccessCode}}HttpStatus.NOT_IMPLEMENTED{{/returnSuccessCode}});
```
cbornet: Shouldn't it be `HttpStatus.valueOf({{{statusCode}}})` instead of `HttpStatus.OK` ?

cbornet:
LGTM :+1:

modules/openapi-generator/src/main/resources/JavaSpring/methodBody.mustache
ackintosh: Yes, it should be `HttpStatus.OK`  as [the status code is contained by example](https://github.com/OpenAPITools/openapi-generator/pull/1197/files#diff-48c5051e9bbbf89ae664cd5918204e52R65). The code above is in `{{^examples}}` section.

ackintosh:
Thank you. ✨

ackintosh:
To be clear the purpose, I've updated the description of this PR. 😌 
https://github.com/OpenAPITools/openapi-generator/pull/1197#issue-221122063

PR #1236, opened 2018-10-14
Update generator option

wing328:
### PR checklist

### Description of the PR

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/LuaClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/LuaClientCodegen.java
@@ -48,11 +48,11 @@ public class LuaClientCodegen extends DefaultCodegen implements CodegenConfig {
     private static final Logger LOGGER = LoggerFactory.getLogger(LuaClientCodegen.class);
 
     protected String specFolder = "spec";
-    protected String packageName = "openapi-client";
+    protected String packageName = "openapiclient";
     protected String packageVersion = "1.0.0-1";
     protected String apiDocPath = "docs/";
     protected String modelDocPath = "docs/";
-    protected String luaRocksFilename = "openapi-client-1.0.0-1.rockspec";
+    protected String luaRocksFilename = "openapiclient-1.0.0-1.rockspec";
 
     public CodegenType getTag() {
         return CodegenType.CLIENT;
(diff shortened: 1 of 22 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -219,7 +219,7 @@ public void processOpts() {
         if (additionalProperties.containsKey(CodegenConstants.PACKAGE_NAME)) {
             setPackageName((String) additionalProperties.get(CodegenConstants.PACKAGE_NAME));
         } else {
-            setPackageName("swagger_client");
+            setPackageName("openapi_client");
```
wing328: @bjgill FYI. I've renamed the default value to "openapi_client" (similar to what we've done in other generators)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/LuaClientCodegen.java
```diff
@@ -138,8 +138,8 @@ public LuaClientCodegen() {
         importMapping.put("os", "io/ioutil");
 
         cliOptions.clear();
-        cliOptions.add(new CliOption(CodegenConstants.PACKAGE_NAME, "Lua package name (convention: lowercase).")
-                .defaultValue("swagger-client"));
+        cliOptions.add(new CliOption(CodegenConstants.PACKAGE_NAME, "Lua package name (convention: kebab-case).")
+                .defaultValue("openapi-client"));
```
wing328: @daurnimator FYI. I've renamed the default value to "openapi_client" (similar to what we've done in other generators)

daurnimator: Have you though? Seems to be using hypens rather than underscores?
Generally both hypens and underscores should *both* be avoided in lua module names. Anything after a hyphen should be a version number. Anything after a "_" is stripped if not found.
Thought you'll find lua modules in the wild that break this: we shouldn't encourage it.

wing328: @daurnimator I picked kebab-case (e.g. openapi-client) to start with as I checked https://luarocks.org/ and found packages (top download) named with hyphens.

So what's the suggested naming convention for Lua package name? Just `openapiclient`?

wing328: @daurnimator thanks for the clarification. I've updated the wording via https://github.com/OpenAPITools/openapi-generator/pull/1236/commits/11280274d99e48851fa7b4d35489ae5f326a96de

PR #1223, opened 2018-10-12
fix export_generator shell script

fujigon:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

`./bin/utils/export_docs_generators.sh` (and its depending `./bin/utils/export_generators.sh`) has string pattern bug.

PR #1239, opened 2018-10-14
[cli] config-help writes doc-compat output

jimschubert:
This updates config-help to have more control over how the output is
written for the user. We dump config-help output for per-generator
documentation, and this cleans up some cross-platform compatibility
issues that might arise from tweaking the output for clarity in the
target file.

Previously, we'd pipe config-help output to sed, then insert the
generator name, which we then redirected to an output file. The sed
syntax had to include a trailing newline so our tabbed configs would
automatically become code tags in markdown. Inserting newlines into sed
replacement strings doesn't work the same across platforms, mostly
because of Apple's customizations to GNU programs.

This commit moves the generator name and newline insertion into the
command itself. It also includes a new --output option, allowing the
user to specify the output location of the config-help.

Currently, we only dump in plain-text, and it is only coincidental that
our plaintext output results in a desirable Markdown output. If
tabbed lines did not automatically convert to a code style block, some
generators like C# would end up with broken text (`List<T>` would become
just `List`, for example).

I had previously discussed extending config-help to output to other
formats like asciidoc. This commit does not introduce any steps toward
that end.

### PR checklist

### Description of the PR

### DIFF
--- a/bin/utils/export_docs_generators.sh
+++ b/bin/utils/export_docs_generators.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 SCRIPT="$0"
-echo "# START SCRIPT: $SCRIPT"
+echo "# START SCRIPT: ${SCRIPT}"
 
 executable="./modules/openapi-generator-cli/target/openapi-generator-cli.jar"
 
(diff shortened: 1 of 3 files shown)


jimschubert:
Extending on contribution done in #1223

/cc @wing328

PR #1244, opened 2018-10-15
Enable error handling in Java WebClient library, fixes #1243

### LINKED ISSUE #1243: [Java] [WebClient] Custom error handling of HTTP responses not possible
##### Description

Custom error handling of HTTP responses not possible, because of the catch-all line in:

https://github.com/OpenAPITools/openapi-generator/blob/e22c7c1c8aca41d7135bb7f3c9e54de173bd50f7/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache#L486

3.3.0

any

standard

##### Steps to reproduce

generate any client

lukstei:
### PR checklist

### Description of the PR

Enable error handling in Java WebClient library, fixes #1243.
Also now the generated Exception class `ApiException` is used.

Custom error handling behaviour can be defined in an overwriting class e.g.:

```java
public class MyApiClient extends ApiClient {
  @Override
  protected <T> Mono<T> handleUnsuccessfulResponse(ClientResponse response) {
    return super.<T>handleUnsuccessfulResponse(response)
      .onErrorMap(errorcode(404), x -> new MyNotFoundException("Not found"));
  }

  private static Predicate<Throwable> errorcode(int errorCode) {
    return x -> x instanceof ApiException && ((ApiException) x).getCode() == errorCode;
  }
}
```

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
@@ -453,6 +453,23 @@ public class ApiClient {
     }
 
     /**
+     * Handles an unsuccessful server response.
+     * This method is called when the server returns a HTTP status code != 2xx.
+     *
+     * The default behaviour is to wrap the unsuccessful response in an {@code ApiException}
+     * and terminate the reactive stream with the exception.
+     *
+     * Can be overwritten to implement custom error handling.
+     *
+     * @param response the unsuccessful HTTP response
+     * @param <T>
+     * @return reactive stream in error state
+     */
+    protected <T> Mono<T> handleUnsuccessfulResponse(ClientResponse response) {
+        return Mono.error(new ApiException(response.statusCode().toString(), null, response.statusCode().value(), response.headers().asHttpHeaders()));
+    }
+
+    /**
      * Invoke API by sending HTTP request with the given options.
      *
      * @param <T> the return type to use
@@ -483,7 +500,7 @@ public class ApiClient {
                         return response.bodyToMono(returnType);
                     }
                 } else {
-                    return Mono.error(new RestClientException("API returned " + statusCode + " and it wasn't handled by the RestTemplate error handler"));
+                    return handleUnsuccessfulResponse(response);
                 }
         });
     }
@@ -520,7 +537,7 @@ public class ApiClient {
                         return response.bodyToFlux(returnType);
                     }
                 } else {
-                    return Flux.error(new RestClientException("API returned " + statusCode + " and it wasn't handled by the RestTemplate error handler"));
+                    return handleUnsuccessfulResponse(response);
                 }
             });
     }
--- a/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
@@ -453,23 +453,6 @@ public class ApiClient {
     }
 
     /**
-     * Handles an unsuccessful server response.
-     * This method is called when the server returns a HTTP status code != 2xx.
-     *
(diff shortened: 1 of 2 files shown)


lukstei:
> Thanks for the PR but your commit (as shown in the Commits tab) is not linked to your Github account, which means this PR won't count as your contribution in https://github.com/OpenAPITools/openapi-generator/graphs/contributors.
> 
> Let me know if you need help fixing it.
> 
> Ref: https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-can-i-update-commits-that-are-not-linked-to-my-github-account

Thanks for the info - fixed it

cbornet:
Spring has a built-in ResponseStatusException which should be used instead of ApiException

cbornet:
Actually I think this all should be simplified to:
```java
return requestBuilder.retrieve()..bodyToMono(returnType);
```
which should thow a [WebClientResponseException](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/reactive/function/client/WebClientResponseException.html) if the status code is 4xx or 5xx

lukstei:
> Actually I think this all should be simplified to:
> 
> ```java
> return requestBuilder.retrieve().bodyToMono(returnType);
> ```
> 
> which should throw a [WebClientResponseException](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/reactive/function/client/WebClientResponseException.html) if the status code is 4xx or 5xx

@cbornet yes you are right  it is handled when calling retrieve() instead of exchange() (see [DefaultWebClient:374](https://github.com/spring-projects/spring-framework/blob/v5.1.3.RELEASE/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L374)). So this custom error handling can be removed altogether. I will do this.

Two further questions arised: 

1) Should we wrap the `WebClientResponseException` in an `ApiException`?
2) [ApiClient:495](https://github.com/OpenAPITools/openapi-generator/blob/b6a937883969da0345c34d4c4aa0301fffcfb647/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache#L495): This `return Mono.empty()` is not optimal because it does not consume the body (see note in [ClientResponse](https://docs.spring.io/spring/docs/5.0.5.RELEASE_to_5.0.6.RELEASE/Spring%20Framework%205.0.6.RELEASE/org/springframework/web/reactive/function/client/ClientResponse.html)). 
This is fixed when using your suggested solution because it then calls the drainBody() method (see [DefaultWebClient.java:448](https://github.com/spring-projects/spring-framework/blob/v5.1.3.RELEASE/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L448
)) which consumes the body. 
The question for me is, if we can safely remove this custom handling logic altogether. I think for the second case yes (`returnType == null`), because this is already handled in the [`BodyExtractor`](https://github.com/spring-projects/spring-framework/blob/v5.1.3.RELEASE/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyExtractors.java#L186), but I'm not sure about the `response.statusCode() == HttpStatus.NO_CONTENT` case, what do you think?

cbornet:
IMO we should remove all the logic and not use ApiException as it doesn't add any value. Also I think WebClient users would expect WebClientResponseException if something went wrong.

lukstei:
I removed all the logic, which makes it much more cleaner now and also solves the initial issue #1243.

The question whether we should use the `ApiException` is for me still not clear, because I think it would never add any value regardless of the underlying library, and I suppose every library has its own underlying Exception, but this should be discussed on a more general level and not in this issue I think.

@cbornet thank you for the input.

cbornet:
LGTP :+1:

cbornet:
As for ApiException it could be used to replace the RuntimeException that are thrown at some places of the code.

PR #1258, opened 2018-10-17
[Rust] Client library choice between hyper and reqwest

bcourtine:
### PR checklist

CC @frol @farcaller @bjgill for review.

### Description of the PR

Port of PR https://github.com/swagger-api/swagger-codegen/pull/8804

Propose an alternative library (reqwest) to generate a Rust client. This library is much easier to use than hyper (especially to deal with https services).


wing328:
@bcourtine thanks for the PR. I'd a quick look and overall it looks good.

Can you please add `samples/client/petstore/rust-reqwest` to the following line in pom.xml so that Travis will test the auto-generated code for Rust (reqwest) client?

https://github.com/OpenAPITools/openapi-generator/blob/aac88a5c093e39c78b69224fe98f5037ba11e608/pom.xml#L1027

bin/rust-petstore.sh
```diff
@@ -27,6 +27,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust -DpackageName=petstore_client $@"
+ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust -DpackageName=petstore_client --library=hyper $@"
```
bjgill: I don't think this is back-compatible. As such, I would suggest creating this against the branch for the next major release - 4.0.x, I believe. See https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches

modules/openapi-generator/src/main/resources/rust/Cargo.mustache
```diff
@@ -4,14 +4,21 @@ version = "{{{packageVersion}}}"
...
+futures = "0.1.23"
+{{/hyper}}
+{{#reqwest}}
+reqwest = "~0.8"
+{{/reqwest}}
 
 [dev-dependencies]
+{{#hyper}}
 tokio-core = "*"
+{{/hyper}}
```
bjgill: My highest-level question - why make the decision at codegen time rather than at compile time? You could presumably achieve the same effect with [Rust features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section). I'm not saying you're wrong - merely curious as to why you chose as you did.

bcourtine: I don't have a good reason for that. I am a beginner in Rust development and learning.

I don't know what purpose this dev lib deserves. Since I could generate my project without it in the Cargo file, I dropped it, simply to keep the file as tiny as possible.

If you think it is a mistake, I can put it back.

bin/rust-petstore.sh
bcourtine: This "library" modification is optional. If no library is chosen for generation, Hyper remains the default library. I may be wrong, but I think it is backward compatible.

wing328: If `hyper` is the default, the change should be backward-compatible.

If `reqwest` is the default, the change should target 3.4.x branch as it's a breaking change with fallbacks.

bcourtine:
Shippable failed tests seems unrelated to current PR. It fails in "ElixirPetstoreClientTests" which is not impacted by this branch commits.

modules/openapi-generator/src/main/resources/rust/Cargo.mustache
bcourtine: I probably misunderstood your question (I thought it was about `tokio-core` lib only).

For the whole library choice, the main reason is that since I'm learning Rust, I don't know how to use it yet.

Another reason whould be that `hyper` and `reqwest` templates are quite different: API main methods currently don't have the same return type (`Result` vs `Box<Future<Result>>>`. So I presume that generating a client managing both libs with features would have much more complex code (but since I don't know features, I may be wrong).

bjgill: That seems reasonable. I could well imagine that trying to have both in the same crate could cause problems if they ever had incompatible dependencies.

bjgill:
Thanks for your prompt answers.

It looks if you're still busy making code changes. Please let me know when you've finished and I'll take a closer look.

bcourtine:
You can review it. It works pretty well (I used the reqwest generator with success on several projects).

There are still some bugs, but I can't find an easy way to correct them, and won't have enough time to work on it soon. For now, I correct theses remaining problems post generation when they occur.

FYI, here are the problems I spotted that are not corrected yet:

* If a query parameter has type `Vec<i32>`, generated code does not compile:

```rust
        let query_string = {
            let mut query = ::url::form_urlencoded::Serializer::new(String::new());
            query.append_pair("vec_i32_param", &vec_i32_param.join(",").to_string());
            query.finish()
        };
```
It must be corrected by something like:

```rust
        let query_string = {
            let mut query = ::url::form_urlencoded::Serializer::new(String::new());
            let vec_str_param: Vec<String> = vec_i32_param.iter().map(|p| p.to_string()).collect();
            query.append_pair("vec_i32_param", &vec_str_param.join(",").to_string());
            query.finish()
        };
```

* Header management has several problems (marked as TODO in `RustClientCodegen.java`) .

With `reqwest`, I did not find a simple way to create a custom raw header (like with `hyper`), and had to introduce a structure for each header (cf. `configuration.rs` file). All possibles headers are collected in java `postProcessOperationsWithModels` method. Theses headers are used without transformation. So, corresponding generated code does not respect Rust case convention. Worse, if a header key contains a char like '-', generated code does not compile.

bcourtine:
Another thought about headers. In `Cargo.toml`, for `reqwest` generation, I would like to remove explicit `hyper` dependency to rely on transitive dependency management.

I had to reintroduce the dependency because of headers. Implementing `reqwest::header::Header` requires to use directly `hyper::error::Error`. It is the only required code where `reqwest` does not obfuscate `hyper` usage.

wing328:
>             let vec_str_param: Vec<String> = vec_i32_param.iter().map(|p| p.to_string()).collect();

What about using `{{#isInteger}}` and/or `{{#isLong}}` tag to add the above line for integer and/or long?

wing328:
Feel free to wait for @bjgill option. I was just throwing ideas...

If it's a list, you will need to do the following instead:
```
{{#isListContainer}}
{{#isInteger}}
...
{{/isInteger}}
{{/isListContainer}}
```

Please use `-DdebugOperations` or `-DdebugModels` to get a full list of mustache tags.

bcourtine:
@wing328 Thanks for the ideas: your proposal sounds good. I just want @bjgill opinion on the best Rust syntax to achieve correction (there may be a more effective way than transforming each integer into a string before concatenating…).

bjgill:
I don't think any of the Rust generators handle query parameters very cleanly. For lists, you proposal seems sensible. I wonder whether we want to special-case integers, though. Will we not want to do the same thing for all non-string types?

Indeed, I believe that `String::to_string` is a no-op. Thus, could you always include `.iter().map(|p| p.to_string()).collect()` if `{{#isListContainer}}`?

bcourtine:
Sounds a good idea. I do that and let you review the whole PR ;)

samples/client/petstore/rust-reqwest/src/apis/configuration.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(Debug, Clone)]
+pub struct HeaderApiKey(String);
+
+impl Header for HeaderApiKey {
+    fn header_name() -> &'static str {
+        "api_key"
+    }
+
+    fn parse_header(raw: &Raw) -> Result<Self, hyper::error::Error> {
+        Ok(HeaderApiKey(std::str::from_utf8(raw.one().unwrap())?.to_string()))
```
bjgill: I would recommend using `expect` instead of `unwrap` in all but the most trivial cases. I had to go and look up `Raw::one` to see why the `unwrap` here was safe.

samples/client/petstore/rust-reqwest/src/apis/configuration.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    fn fmt_header(&self, f: &mut Formatter) -> Result<(), std::fmt::Error> {
+        f.fmt_line(&self.0)
+    }
```
bjgill: Could you implement `Display` instead of this?

samples/client/petstore/rust-reqwest/src/apis/mod.rs
```diff
@@ -0,0 +1,32 @@
+use reqwest;
+use serde_json;
+
+#[derive(Debug)]
+pub enum Error {
+    Reqwest(reqwest::Error),
+    Serde(serde_json::Error),
+}
```
bjgill: It might be nice to use `error_chain` or `fail` instead of creating a custom error framework here. Non-essential, though, especially if this is what the hyper library already does.

samples/client/petstore/rust-reqwest/src/apis/configuration.rs
```diff
@@ -0,0 +1,67 @@
...
+            client: reqwest::Client::new(),
+            basic_auth: None,
+            oauth_access_token: None,
+            api_key: None,
+        }
+    }
+
+    pub fn header_configured_api_key(&self) -> HeaderApiKey {
+        HeaderApiKey(self.api_key.clone().expect("API key not configured"))
+    }
```
bjgill: Why do we need this function? I'm a little nervous that the `expect` makes this somewhat unsafe to call.

samples/client/petstore/rust-reqwest/src/apis/configuration.rs
```diff
@@ -0,0 +1,67 @@
...
+impl Configuration {
+    pub fn new() -> Configuration {
+        Configuration {
+            base_path: "http://petstore.swagger.io/v2".to_owned(),
+            user_agent: Some("Swagger-Codegen/1.0.0/rust".to_owned()),
+            client: reqwest::Client::new(),
+            basic_auth: None,
+            oauth_access_token: None,
+            api_key: None,
+        }
```
bjgill: Clippy would tend to complain at this function - see https://rust-lang-nursery.github.io/rust-clippy/current/index.html#new_without_default

samples/client/petstore/rust-reqwest/src/apis/pet_api.rs
```diff
@@ -0,0 +1,349 @@
...
+        let client = &configuration.client;
+
+
+
+        let query_string = {
+            let mut query = ::url::form_urlencoded::Serializer::new(String::new());
+
+            query.finish()
+        };
+        let uri_str = format!("{}/pet?{}", configuration.base_path, query_string);
```
bjgill: It seems as if when there are no query parameters we construct and provide an empty query string. I'm not sure it's wrong, but it seems a little unusual.

samples/client/petstore/rust-reqwest/src/apis/pet_api.rs
```diff
@@ -0,0 +1,349 @@
...
+
+
+        // send request
+        let req = req_builder.build()?;
+
+        client.execute(req)?.error_for_status()?;
+        Ok(())
+    }
+
+    fn upload_file(&self, pet_id: i64, additional_metadata: &str, file: ::models::File) -> Result<::models::ApiResponse, Error> {
```
bjgill: Should we be doing something with `additional_metadata` and `file`?

samples/client/petstore/rust-reqwest/src/models/api_response.rs
```diff
@@ -0,0 +1,90 @@
+/* 
+ * OpenAPI Petstore
+ *
+ * This is a sample server Petstore server. For this sample, you can use the api key `special-key` to test the authorization filters.
```
bjgill: Is this correct?

samples/client/petstore/rust-reqwest/src/models/api_response.rs
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * OpenAPI spec version: 1.0.0
+ * 
+ * Generated by: https://openapi-generator.tech
+ */
+
+/// ApiResponse : Describes the result of uploading an image resource
```
bjgill: What is this doc comment attached to?

samples/client/petstore/rust-reqwest/src/models/api_response.rs
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  pub fn set_code(&mut self, code: i32) {
+    self.code = Some(code);
+  }
+
+  pub fn with_code(mut self, code: i32) -> ApiResponse {
+    self.code = Some(code);
+    self
+  }
```
bjgill: What is the purpose of all these getters and setters? Why not have the fields public?

samples/client/petstore/rust-reqwest/src/models/api_response.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl ApiResponse {
+  /// Describes the result of uploading an image resource
+  pub fn new() -> ApiResponse {
+    ApiResponse {
+      code: None,
+      _type: None,
+      message: None
+    }
```
bjgill: Can you not implement default instead?

modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
```diff
@@ -56,7 +56,7 @@ impl {{classname}} for {{classname}}Client {
         let query_string = {
             let mut query = ::url::form_urlencoded::Serializer::new(String::new());
 {{#queryParams}}
-            query.append_pair("{{baseName}}", &{{paramName}}{{#isListContainer}}.join(","){{/isListContainer}}.to_string());
+            query.append_pair("{{baseName}}", &{{paramName}}{{#isListContainer}}.iter().map(|p| p.to_string()).collect::<Vec<String>>().join(","){{/isListContainer}}.to_string());
```
bjgill: Nice.

For bonus points, you could use `into_iter` instead of `iter`. As I understand it, `iter` produces references to `p`, and `to_string` will therefore clone `p`. `into_iter` would give you an owned `p`, and thus `to_string` would not clone..

bin/rust-reqwest-petstore.sh
```diff
@@ -0,0 +1,32 @@
...
+if [ ! -f "$executable" ]
+then
+  mvn -B clean package
+fi
+
+# if you've executed sbt assembly previously it will use that instead.
+export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
+ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust-reqwest -DpackageName=petstore_client --library=reqwest $@"
+
+java ${JAVA_OPTS} -jar ${executable} ${ags}
```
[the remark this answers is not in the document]
bcourtine: I don't have a particular reason. I will merge both scripts.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
```diff
@@ -353,9 +368,33 @@ public String toOperationId(String operationId) {
...
+
+            if (operation.authMethods != null) {
+                for (CodegenSecurity authMethod : operation.authMethods) {
+                    if (authMethod.isApiKey && authMethod.isKeyInHeader) {
+                        additionalProperties.put("authHeaderKey", authMethod.keyParamName);
+                    }
+                }
+            }
+
+            // TODO Manage Rust var codestyle (snake case) and compile problems (headers with special chars, like '-').
```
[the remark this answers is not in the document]
bcourtine: It is possible, but complicated. It requires to introduce an object to maintain both native and underscored versions. Native version is required for the `reqwest::header::Header` implementation (header name).

samples/client/petstore/rust-reqwest/README.md
```diff
@@ -0,0 +1,97 @@
...
+## api_key
+- **Type**: API key 
+
+Example
+```
+	auth := context.WithValue(context.TODO(), sw.ContextAPIKey, sw.APIKey{
+		Key: "APIKEY",
+		Prefix: "Bearer", // Omit if not necessary.
+	})
+    r, err := client.Service.Operation(auth, args)
```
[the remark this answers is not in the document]
bcourtine: 👍. I focused on technical code and did not work on documentation.

samples/client/petstore/rust-reqwest/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+                        <id>bundle-test</id>
+                        <phase>integration-test</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>cargo</executable>
+                            <arguments>
+                                <argument>build</argument>
+                                <argument>--release</argument>
```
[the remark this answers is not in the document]
bcourtine: `check` is probably enough. I used a release build copying the existing script in `hyper` samples folder.  Both can be corrected.

samples/client/petstore/rust-reqwest/src/apis/configuration.rs
```diff
@@ -0,0 +1,67 @@
+/* 
+ * OpenAPI Petstore
+ *
+ * This is a sample server Petstore server. For this sample, you can use the api key `special-key` to test the authorization filters.
+ *
+ * OpenAPI spec version: 1.0.0
+ * 
+ * Generated by: https://openapi-generator.tech
+ */
```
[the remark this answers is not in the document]
bcourtine: It is not intentional. I probably used by mistake a generated file as source for modifications.

modules/openapi-generator/src/main/resources/rust/Cargo.mustache
```diff
@@ -4,14 +4,21 @@ version = "{{{packageVersion}}}"
...
 serde_yaml = "0.7"
-serde_json = "1.0"
 base64 = "~0.7.0"
-futures = "0.1.16"
-hyper = "0.11.6"
-url = "1.5"
+futures = "0.1.23"
+{{/hyper}}
+{{#reqwest}}
+reqwest = "~0.8"
```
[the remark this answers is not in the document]
bcourtine: Only because when I started development, 0.9.x was not released. ;) I forgot to check new versions.

samples/client/petstore/rust-reqwest/src/apis/pet_api.rs
bcourtine: Yes. I noticed it too. It comes from the forked `hyper` code. I did not take time to handle this better for now. It is something I correct manualy post-generation. But there is probably something to do.

samples/client/petstore/rust-reqwest/src/apis/pet_api.rs
bcourtine: `File` management is a remaining TODO forked from `hyper` code.

samples/client/petstore/rust-reqwest/src/models/api_response.rs
bcourtine: No. Copy-paste error.

samples/client/petstore/rust-reqwest/src/models/api_response.rs
bcourtine: Copied from original code. Don't know the reason...

But I imagine the purpose is to get a "fluent API", with method calls chained: `model.with_code("code").with_name("name")...`. It is a common pattern in other languages like `java` to keep fields private and to provide this access methods.

bcourtine:
Thanks for the complete review. I will correct this. But since I am quite busy, corrections may not be available before some days.

bcourtine:
I just looked at reqwest 0.9 release notes and 😭 :

> The RequestBuilder has simple methods to set headers using strings, which can work in most places.

Most custom header code added in configuration template can be dropped with this feature.

samples/client/petstore/rust-reqwest/src/models/api_response.rs
bcourtine: I note for later, but I won't correct it in this branch: it would be a breaking change in generated API since models are shared between hyper and reqwest versions.

bcourtine: I note this for later, but I won't correct it in this branch: it would be a breaking change in generated API since models are shared between hyper and reqwest versions.

bcourtine:
@bjgill Most of the points you spotted have been corrected. You can review this new version.

I did not correct model problems, because it would be a breaking change in generated API (since models are used by both `hyper` and `reqwest` versions), especially replacing accessors by public fields.

samples/client/petstore/rust-reqwest/src/models/api_response.rs
bcourtine: After a closer look, it is not as simple. When there is a required var in a model, default constructor `new` has required vars as parameters. So, `Default` trait cannot be implemented in this case.

bjgill: Good point - https://github.com/nrc/derive-new might be what you want, then.

samples/client/petstore/rust-reqwest/src/apis/mod.rs
bjgill: See https://github.com/rust-lang-nursery/failure or https://rust-lang-nursery.github.io/failure/.

In any case, what you've got is fine. Just something you might want to look at for future reference.

bin/rust-petstore.sh
```diff
@@ -26,7 +26,12 @@ then
...
-export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust -DpackageName=petstore_client $@"
+export JAVA_OPTS="${JAVA_OPTS} -Xmx1024M -DloggerPath=conf/log4j.properties"
 
+# Generate client using hyper lib.
+ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust -DpackageName=petstore_client --library=hyper $@"
 java ${JAVA_OPTS} -jar ${executable} ${ags}
+
+# Generate client using reqwest lib.
+ags="generate -t modules/openapi-generator/src/main/resources/rust -i modules/openapi-generator/src/test/resources/2_0/petstore.yaml -g rust -o samples/client/petstore/rust-reqwest -DpackageName=petstore_client --library=reqwest $@"
```
wing328: Usually, we create another script (e.g. `bin/rust-petstore-reqwest.sh`) for a different HTTP library. (You will find a lot of `./bin/java-petstore-{library}.sh` scripts for different HTTP libraries supported by the Java client generator)

I suggest we do the same for Rust reqwest instead of putting both into one singel script.

bcourtine:
@wing328 Done.

Poke @bjgill who suggested merging theses scripts ;)

bjgill:
I'm happy for this to be merged (modulo CI) - sorry for suggesting something that you then had to undo.

bcourtine:
@bjgill I was just trolling. Merging samples scripts (and undoing it) was trivial.

bjgill:
@wing328 - anything else? Otherwise, I'll merge this tomorrow.

wing328:
Thank you @bcourtine again for the contribution and @bjgill for the review.

I've sent out a tweet (https://twitter.com/oas_generator/status/1055863197687664640) to help promote the enhancement.

Have a nice weekend.

wing328:
@bcourtine thanks for the PR, which is included in the v3.3.2 release: https://twitter.com/oas_generator/status/1057649626101112832

Issue #641, opened 2018-07-25
PHP - pass array(key-value pair) as an argument instead of comma separated individual arguments 


rajanmodi:
##### Description

Functions generated automatically in SDK should support array(key-value pair) as an argument.

Suppose there is search function with 66 arguments. 
search($this->arg1 = null, $this->arg2 = null, ...., $this->arg66 = null)

When we need to make a call to this search function ,we need to provide all the arguments
  $apiInstance->search($this->arg1 = 5, $this->arg2 = 10, $this->arg3 = 20,.....,$this->arg66 = 1)

What if I only want to pass only two arguments 
 $apiInstance->search($this->arg1 = 5,$this->arg31 = 95).
now this will assign 5 to arg1 and 95 to arg2. I need to assign value of 95 to arg31

I think solution will be generating function argument as an array with key-value pair for such cases . 

Instead of keeping comma separated arguments, array should be there..

wing328:
@rajanmodi thanks for the suggestion.

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko

ybelenko:
@rajanmodi Hmm... Give us an example? I can't find any function with more than one argument in current PHPClient samples. We're talking about models, right?
```php
/**
 * Constructor
 *
 * @param mixed[] $data Associated array of property values
 *                      initializing the model
 */
public function __construct(array $data = null)
{
    $this->container['bill_id'] = isset($data['bill_id']) ? $data['bill_id'] : null;
    $this->container['bill_date'] = isset($data['bill_date']) ? $data['bill_date'] : null;
    $this->container['charges'] = isset($data['charges']) ? $data['charges'] : null;
    $this->container['charge_desc'] = isset($data['charge_desc']) ? $data['charge_desc'] : null;
}
```

wing328:
@ybelenko I think @rajanmodi is referring to method arguments, e.g. [updatePetWithForm](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore.yaml#L176-L210) with 3 parameters only and let's imagine it has 20 additional optional parameters.

ybelenko:
@wing328 If so, I think that more common practice is to create endpoints with just few parameters, or with one complex JSON object parameter.
I like current "full list of parameters" appouch more and user always can rewrite mustache template a bit.

[...]

PR #1269, opened 2018-10-19
Add group parameters support to PHP templates

wing328:
### PR checklist

### Description of the PR

For https://github.com/OpenAPITools/openapi-generator/issues/641

Still WIP:
- async method needs to be updated
- docstring needs to be updated
- fix test cases

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko for comment/feedback


ybelenko:
@wing328 How it's realated to server PHP generators? Server side api controllers usually looks like:
```php
public function testEndpointParameters($request, $response, $args)
```
I would add an option to change models constructor from classic arguments list to associative array, but we can provide both.

wing328:
> How it's realated to server PHP generators? Server side api controllers usually looks like:

The enhancement is for PHP client at the moment and you've raised a good point about model constructors (in clients or servers) accepting associative array instead.

wing328:
OK. I'll use `array_key_exists`

For default value, I suggest we don't include that in the client as the server (backedn) should ensure the default value ( I think Python, Ruby, etc are following this approach to not set default value in the client side)

wing328:
@ackintosh I've updated the PR based on your feedback. Please have another look when you've time.

ackintosh:
> For default value, I suggest we don't include that in the client as the server (backend) should enforce the default value

Oh, I get it. That makes sense. 💡

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/1337

PR #1286, opened 2018-10-22
Allow package version to be passed on CLI (Rust fix)

MF1-MS:
This MR allows package version to be specified in Rust in the generate argument list, with the argument `-DpackageVersion=<package_version>`. If this argument is present then the version in the resulting Cargo.toml file will be the passed value. If this argument is not present then the version in the OpenAPI definition file will be used, as per current behavior.


wing328:
@MF1-MS thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

PR #1288, opened 2018-10-22
[elm] Add flag to prefix custom type variants

### LINKED ISSUE #309: [elm-client] Enums can have the same values (not allowed)
Having a definition with multiple enums with the same value, will produce types with the exact same constructor in elm, which is not allowed.

```json
    "metadata": {
      "type": "object",
      "properties": {
        "length": {
          "type": "string",
          "enum": ["short", "long"]
        },
        "size": {
          "type": "string",
          "enum": ["short", "medium"]
        }
      },
    }
```
Generated elm code:

```elm
type Length
    = Short
    | Long

type Size
    = Short
    | Medium
```

Should be:

```elm
type Length
    = LengthShort
    | LengthLong

type Size
    = SizeShort
    | SizeMedium
```

OpenAPI spec version: 0.4.0
Generator version 3.0.1

trenneman:
### PR checklist

### Description of the PR

Fixes #309 adding support for an additional CLI option: `--additional-properties prefixCustomTypeVariants=true`


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -61,13 +61,15 @@
     private ElmVersion elmVersion = ElmVersion.ELM_019;
 
     private static final String ELM_VERSION = "elmVersion";
+    private static final String PREFIX_CUSTOM_TYPE_VARIANTS = "prefixCustomTypeVariants";
```
wing328: @trenneman Since the option is Elm specified, what about naming it as "elmPrefixCustomTypeVariants"?

trenneman: Good point, I'll update the PR.

Issue #1287, opened 2018-10-22
[Golang][client] client.decode error is not handled 


grokify:
##### Description

When StatusCode < 300,  `a.client.decode` returns an error but it is not handled in the code which returns a `nil` error. There is a comment that mentions some kind of handling that is necessary but the handling doesn't exist.

https://github.com/OpenAPITools/openapi-generator/blob/8507916d569a8614286708d45a3cbae20e4da6f2/modules/openapi-generator/src/main/resources/go/api.mustache#L276-L282

However, the error is not handled and the next thing that happens is that error is returned as a nil at the end of the function:

https://github.com/OpenAPITools/openapi-generator/blob/8507916d569a8614286708d45a3cbae20e4da6f2/modules/openapi-generator/src/main/resources/go/api.mustache#L307

##### openapi-generator version

master: 8507916d569a8614286708d45a3cbae20e4da6f2

##### OpenAPI declaration file content or url

n/a

##### Command line used for generation

n/a

##### Steps to reproduce

Make an API call that generates a decode error, e.g. if model schema doesn't return what is returned by the API.

##### Related issues/PRs

n/a

##### Suggest a fix/enhancement

Implement `GenericOpenAPIError` similar to when StatusCode >= 300 with `GenericOpenAPIError`.

https://github.com/OpenAPITools/openapi-generator/blob/8507916d569a8614286708d45a3cbae20e4da6f2/modules/openapi-generator/src/main/resources/go/api.mustache#L285-L298

PR #1291, opened 2018-10-22
[Golang][client] fix response decode error handling

grokify:
### PR checklist

@antihax @bvwells @kemokemo 

### Description of the PR

Fix https://github.com/OpenAPITools/openapi-generator/issues/1287

When encountering a `client.decode` error, this will return the error message in a `GenericOpenAPIError` with the response body included. Previously, the error would be dropped silently.

### DIFF
--- a/docs/generators/lua.md
+++ b/docs/generators/lua.md
@@ -2,7 +2,7 @@
 CONFIG OPTIONS for lua
 
 	packageName
-	    Lua package name (convention: lowercase). (Default: openapir-client)
+	    Lua package name (convention: single word). (Default: openapiclient)
 
 	packageVersion
 	    Lua package version. (Default: 1.0.0-1)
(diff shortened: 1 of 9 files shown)


grokify:
Core fix: 649e617c9f13986012b873fe511b6132bb1aa89b
Easier readability: 0416cb38b31ac6175288f969d66db73393f76bd4

The readability fox moves  `if localVarHttpResponse.StatusCode < 300` success handling code below error handling code in `if localVarHttpResponse.StatusCode >= 300` and removes the `< 300` else clause since it is not needed.

This:

* removes an extra if statement
* reduces the locations where a success is returned from 2 to 1
* removes returning a success earlier in the code to follow standard approach of returning errors early and success at the end

grokify:
I've done some thinking about this some more. Now that the decode error is being handled the test is failing due to my read of the RFC 3339 standard.

If we merge this and maintain the test without fixing the server, then all builds will start failing.

Another option to merge this PR is to temporarily comment out `TestPlaceOrder` until the test server response issue can be resolved.

grokify:
In discussion with @wing328, it was confirmed that the Petstore test server has this known issue which was previously reported in the former issue tracker.

We agreed to disable `TestPlaceOrder` test for now and we've done similar things in other areas, such as disabling the `flake8` tests for the Python client.

In d2c300641411e51a8e14eb1d13beab696a9a526a, I've disabled the `TestPlaceOrder` test for now, specifically when it encounters a the `+0000` UTC offset error. Any other errors would still be reported.

wing328:
@grokify thanks for the PR, which is included in the v3.3.2 release: https://twitter.com/oas_generator/status/1057649626101112832

Issue #1217, opened 2018-10-10
[Java] Generated Java APIs often have too many method parameters


Kiran-Sivakumar:
##### Description

APIs often have a lot of optional parameters. This does not cause a problem for SDKs generated in languages that inherently support methods with optional parameters. However, in Java, arguments must be explicitly specified for all parameters, even the ones marked as "optional" in the specification. This can make OpenAPI-generated Java SDKs incredibly cumbersome to use.

##### openapi-generator version

3.3.0

##### Suggest a fix/enhancement

Here is a method generated by the current openapi-generator version, along with a call to the method:

```java
/**
 * Create a new media collection
 *
 * @param requiredParam1 (required)
 * @param optionalParam1 (optional)
 * @param optionalParam2 (optional)
 * @param optionalParam3 (optional)
 * @param optionalParam4 (optional)
 * @param optionalParam5 (optional)
 * @param optionalParam6 (optional)
 * @return Collection
*/
public Collection createCollection(
    String requiredParam1,
    String optionalParam1,
    String optionalParam2,
    String optionalParam3,
    String optionalParam4,
    String optionalParam5,
    String optionalParam6
) {
    // ...
}

// ...

Collection collection = 
    api.createCollection("VALUE_A", "VALUE_B", null, "VALUE_C", null, null, null);
```
Users of the SDK are required to pass in some null value for optional parameters they do not use.

A much better solution would be to generate Java APIs that employ the builder pattern or something similar. This would allow for the following, much nicer code.

```java
Collection collection = api.createCollection("VALUE_A")
        .optionalParameter1("VALUE_B")
        .optionalParameter3("VALUE_C")
        .execute();
```

The above code is similar to the code employed by the Google API Client Library (https://developers.google.com/api-client-library/java/). APIs for endpoints that only have required parameters (no optionals) need not be changed. They can be generated in the current way, although changing them to use the above style as well would consolidate everything nicely.

Kiran-Sivakumar:
Another solution would be to adopt an approach similar to the one used in the AWS and Splunk SDKs. 

```java
Collection collection = api.createCollection(
        new CreateCollectionArgs("VALUE_A")
                    .withOptionalParameter1("VALUE_B")
                    .withOptionalParameter3("VALUE_C")
);
```

This may be preferred over my first proposal.

Kiran-Sivakumar:
[...]

PR #1341, opened 2018-10-30
[Java][okhttp-gson] Add option to employ builders for API requests

Kiran-Sivakumar:
### PR checklist

### Description of the PR

fix https://github.com/OpenAPITools/openapi-generator/issues/1217
This PR (filed against branch 3.4.x) adds the option to generate APIs of the form shown in the proposed solution in the above issue. This makes generated okhttp-gson clients much nicer to use, especially in cases involving unused optional parameters. The hope is to implement a similar solution for all Java generators in the future, so that generated Java clients aren't frustrating to use.

Changes:
- JavaClientCodegen.java modified to add a new CLI option "useBuildersForApiRequests" (default: false)
- okhttp-gson/api.mustache
    - Optionally generate API-request builders
    - Change original methods from "public" to "private" if builders are generated, since user will want to use the public builder methods instead (the builder methods simply wrap around the original methods that take all parameters as input)
    - Generate comments
- okhttp-gson/api_test.mustache added to generate tests that use the builders if the "useBuildersForApiRequests" option is enabled
- bin/java-petstore-okhttp-gson-builders.sh added to generate petstore samples with API-request builders (added one for Windows as well)
- samples/client/petstore/java/okhttp-gson-requestBuilders/: new petstore samples generated with the above script
- docs/generators/java.md updated with new config option

I also ran the original shell scripts under `./bin/` to ensure that absolutely nothing changes in the existing petstore samples when the "useBuildersForApiRequests" option is disabled.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
```diff
@@ -63,6 +63,7 @@
     public static final String FEIGN_VERSION = "feignVersion";
     public static final String PARCELABLE_MODEL = "parcelableModel";
     public static final String USE_RUNTIME_EXCEPTION = "useRuntimeException";
+    public static final String USE_BUILDERS_FOR_API_REQUESTS = "useBuildersForApiRequests";
```
cbornet: Too much options :cry:

cbornet:
As always, I'm reluctant to have too many options. So I would prefer to always generate the builder, keep the constructors `public` and have a way to encourage users to move to the builder style (eg. setting the constructor `@Deprecated` and adding a comment that future version will set it as private. Then in a minor release or 2, we can completely switch to `private`.

Kiran-Sivakumar:
@cbornet Thanks for your feedback. I agree that too many options can be undesirable. Perhaps the option can be removed in the next major release (to only generate the builder style)? Otherwise, I can incorporate your proposed changes.

wing328:
@Kiran-Sivakumar @cbornet FYI. next major version 4.x (with breaking changes) will be released before year-end so we can clean up the options as part of the 4.x release later.

Kiran-Sivakumar:
Ok, so should I keep it as an option for this PR, changing it to be true by default, and file another PR against 4.x in a few weeks that removes the option entirely?

wing328:
@Kiran-Sivakumar I've other ideas. Let me review and get back to you by Thursday.

wing328:
@Kiran-Sivakumar instead of using `useBuildersForApiRequests` in the templates, what about using `x-group-parameters` vendor extensions like what I did for PHP in https://github.com/OpenAPITools/openapi-generator/pull/1337/files#diff-076573ccfdac2ab5af4dc1813e5ccee3R92?

This will avoid another option and make it possible to enable group parameters per operation.

Kiran-Sivakumar:
I see. What if you wanted to enable builders for all operations? Is there an easy way to apply group parameters to all of them?

wing328:
@Kiran-Sivakumar similar to what we did for `consumes`, `produces` (global vs local), if `x-group-parameters` is defined in the top level of the spec, we will add x-group-parameters to all operations in `postProcessOperationsAndModel` (only if x-group-parameters is not defined in the operation level).

Does it sound good to you?

Kiran-Sivakumar:
Oh great. Yes, this sounds good to me. I'll incorporate your feedback.

wing328:
@Kiran-Sivakumar I've filed https://github.com/OpenAPITools/openapi-generator/pull/1405 to add top level x-group-parameters support.

Kiran-Sivakumar:
@cbornet @wing328 I have incorporated your feedback.

wing328:
@Kiran-Sivakumar thanks for the PR, which has been included in the v3.3.4 release: https://twitter.com/oas_generator/status/1068772409795207168

PR #1352, opened 2018-10-31
Quickfix for noncompilation on maven build

raghuraman1:
This is a PR that solves the maven build causing compilation issue for spring boot


raghuraman1:
will figure out how to solve the problem in a better mannaer

PR #1387, opened 2018-11-06
[python] Avoid creating unused ThreadPools

wing328:
Instead, create ApiClient.pool on first request for .pool property.

avoids spawning n-cpus threads (the default for ThreadPool) at instantiation of every ApiClient

### PR checklist

### Description of the PR

As discussed with @minrk, here is same [PR](https://github.com/swagger-api/swagger-codegen/pull/8061) by cherry-picking the commits.

cc @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10)


modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -53,14 +55,15 @@ class ApiClient(object):
         'datetime': datetime.datetime,
         'object': object,
     }
+    _pool = None
 
     def __init__(self, configuration=None, header_name=None, header_value=None,
-                 cookie=None):
+                 cookie=None, pool_threads=1):
```
tomplus: It looks like a breaking change. I suggest leaving `=None` and still using default number of threads. It shouldn't be a problem because the pool will be created on the first request.

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -70,8 +73,19 @@ class ApiClient(object):
...
-        self.pool.close()
-        self.pool.join()
+        if self._pool:
+            self._pool.close()
+            self._pool.join()
+            self._pool = None
+
+    @property
+    def pool(self):
+        """Create thread pool on first request
```
tomplus: :+1:

modules/openapi-generator/src/main/resources/python/api_client.mustache
wing328: Thanks for the feedback.  I'll set `pool_thread=None`

cc @minrk

minrk: FWIW, this is not a breaking change, though it is a change in the default *performance* of concurrent async requests from the same client object. Tying concurrent IO to CPU count doesn't make a lot of sense to me, especially in Python. I'd argue that a fixed default pool size (if not 1, maybe 4 or 10) would be more logical than spawning 64 threads on a big machine, and would result in more consistent behavior.

tomplus: @minrk I agree with you. This default may not work well on a huge server but it's default, it may be optimal in general, this calculation may be changed in the future releases of Python. Someone can base on this behavior and it'd be a breaking change for him/her.

wing328: What about using `pool_thread=None` for current master to make it backward-compatible and using `pool_threads=1` in 4.0.x (a major release with breaking changes scheduled to be released next month)?

tomplus: @wing328 Sounds good to me.

PR #1404, opened 2018-11-08
[rust-server] Always derive Debug

bjgill:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @frol @farcaller 

### Description of the PR

Now that we're using a byte array, we can guarantee that deriving Debug will always work.

Issue #1424, opened 2018-11-13
[SWIFT4] Dates Are Inconsistently Encoded


james-rantmedia:
##### Description

The generated code for GET requests that have URL Query Parameters that are dates currently encodes the date to JSON and inserts it as a query parameter.

The problem with this is that it is possible for the encoded date to contain a `+` character, which is a valid character in a query parameter, but not as an encoded space, as intended.

My problem example:

I have a date with a timezone component, like `2018-11-13T13:20:14.421+01:00`. Adding this date to a URL's query results in: `https://example.com?date=2018-11-13T13:20:14.421+01:00`. The `+` is not encoded by `URLQueryItem` because it is a valid character.

However, when the server receives the request and decodes the query parameters, the result is `2018-11-13T13:20:14.421 01:00`. The + was decoded to a space. This is unintentional, and the server responds with HTTP 400 Bad Request.

##### openapi-generator version

3.3.2

##### Steps to reproduce

- Generate code for Swift4 from an OpenAPI Spec that has a GET request with URL query parameters, one of which is a Date.
- Call the generated request method passing a date that has a positive timezone adjustment, eg `2018-11-13T13:20:14.421+01:00`

##### Related issues/PRs

I will submit a PR to address this issue and update this issue.

##### Suggest a fix/enhancement

Modify the template for API classes to include a URL encoding step for dates added as query parameters. Eg:

```
{{#queryParams}}
{{> _param}}{{#isDateTime}}.addingPercentEncoding(withAllowedCharacters: CharacterSet(charactersIn: "1234567890T.-")){{/isDateTime}}{{#hasMore}}, {{/hasMore}}
{{/queryParams}}
```

james-rantmedia:
After doing some thinking I don't think the fix I suggested in the original issue is the correct way to solve this problem.

It attempts to fix a symptom, and not the cause. I think according to the spec, dates should always be represented as UTC (with the Z suffix), and so there shouldn't ever be a positive timezone adjustment.

I think the proper approach would be to change the date formatter used in Swift4 templates to be the ISO8601DateFormatter with [.withFullDate, .withFullTime, .withFractionalSeconds] as options. using this formatter will result in encoded dates following the expected format, regardless. An example:

let date = Date(timeIntervalSince1970: 0)

// Using ISO8601 with specified options as suggested above
let formatter = ISO8601DateFormatter()
formatter.formatOptions = [.withFullDate, .withFullTime, .withFractionalSeconds]
[...]

PR #1425, opened 2018-11-13
Fixes #1404 URL encode dates for query params

james-rantmedia:
This change URL encodes a date string before it is added as a query parameter to prevent a possible `+` character being mistaken for an encoded space as opposed to the timezone adjustment.

### PR checklist

### Description of the PR

This PR addresses issue #1424 by URL encoded date strings before adding them as URL query parameters, to prevent the `+` in the timezone adjustment from being mistaken as an encoded space.

cc @jgavris (2017/07) @ehyche (2017/08) @Edubits (2017/09) @jaz-ah (2017/09) @d-date (2018/03) for review


james-rantmedia:
I just noticed an issue with this fix just after I submitted the PR 😑, I'll work to resolve it and update this PR.

d-date:
@james-rantmedia 
Thank you for your help!
It's reasonable changes, but could you run `./bin/swift4-petstore-all.sh` as written in PR checklist then commit changes in samples? We cannot understand with just only template changes if the change is correct and don't have any regressions .

Additionally, please write unit test with Xcode so that you can understand the change is correct.

james-rantmedia:
Will do :+1:

james-rantmedia:
Apologies for wasting time, but I'm going to abandon this PR. I've been doing some thinking and I don't think this is the correct way to approach the fix. This PR attempts to fix a symptom, and not the cause. I think according to the spec, dates should always be represented as UTC (with the Z suffix), and so there shouldn't ever be a positive timezone adjustment.

I think the proper approach would be to change the date formatter used in Swift4 templates to be the `ISO8601DateFormatter` with `[.withFullDate, .withFullTime, .withFractionalSeconds]1 as options. using this formatter will result in encoded dates following the expected format, regardless. An example:

```swift
let date = Date(timeIntervalSince1970: 0)

// Using ISO8601 with specified options as suggested above
let formatter = ISO8601DateFormatter()
formatter.formatOptions = [.withFullDate, .withFullTime, .withFractionalSeconds]
let string = formatter.string(from: date) // "1970-01-01T00:00:00.000Z"

// Using the dateformatter code from the current version of the template
let fmt = DateFormatter()
fmt.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
fmt.locale = Locale(identifier: "en_US_POSIX")
let string2 = fmt.string(from: date) // "1970-01-01T01:00:00.000+01:00"
```

You can see that the current templates formatter produces a date with a positive timezone offset (+01:00), whereas the ISO8601 formatter represents it in UTC.

I think I will update the date formatter in the template and create a new PR. Again, apologies for the time wasting.

d-date:
@james-rantmedia Thank you for your PR. We’re waiting for your next PR:)

PR #1479, opened 2018-11-17
Fix NPE for Scala Gatling generator

wing328:
### PR checklist

### Description of the PR

Fix the following NPE:
```
[main] INFO  o.o.c.languages.AbstractScalaCodegen - Environment variable SCALA_POST_PROCESS_FILE not defined so the Scala code may not be properly formatted. To define it, try 'export SCALA_POST_PROCESS_FILE=/usr/local/bin/scalafmt' (Linux/Mac)
[main] INFO  o.o.c.languages.AbstractScalaCodegen - NOTE: To enable file post-processing, 'enablePostProcessFile' must be set to `true` (--enable-post-process-file for CLI).
Exception in thread "main" java.lang.NullPointerException
	at org.openapitools.codegen.languages.ScalaGatlingCodegen.preprocessOpenAPI(ScalaGatlingCodegen.java:242)
	at org.openapitools.codegen.DefaultGenerator.configureGeneratorProperties(DefaultGenerator.java:183)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:854)
	at org.openapitools.codegen.cmd.Generate.run(Generate.java:349)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:62)
```

cc @clasnake (2017/07) @jimschubert (2017/09) @shijinkui (2018/01) @ramzimaalej (2018/03)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaGatlingCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaGatlingCodegen.java
@@ -239,6 +239,11 @@ public class ScalaGatlingCodegen extends AbstractScalaCodegen implements Codegen
                 continue;
             }
             for (Operation operation : path.readOperations()) {
+
+                if (operation.getExtensions() == null) {
+                    operation.setExtensions(new HashMap());
+                }
+
                 if (!operation.getExtensions().keySet().contains("x-gatling-path")) {
                     if (pathname.contains("{")) {
                         String gatlingPath = pathname.replaceAll("\\{", "\\$\\{");


jimschubert:
@wing328 why would extensions be null here? Would it make more sense to ensure this is set and non-null in the type itself?

wing328:
@jimschubert totally agreed with you but the class for `operation` is in another package (io.swagger.core?), which is not something we can change in this project.

I also agree with you that it shouldn't be null in the first place and should be properly initialized with a new Hashmap for the extension when the Operation object is constructed but unfortunately it's not how it's written at the moment.

Issue #893, opened 2018-08-24
Invalid enum var names are generated by multi-byte characters


osjupiter:
##### Description

Hi. I have been using openapi-generator with definitions which contains multi-byte characters.
When I wrote enum definitions  which names contain only non-word characters(RegExp \W), invalid code generated. 
(All of enum names become "_".)

Are double-byte characters not allowed for enum names in OAS ?
If it is valid, I hope these characters will be supported.

##### generated code

java
```java
@JsonAdapter(PetType.Adapter.class)
public enum PetType {
  _("猫"), // cat   --compile error
  _("犬"); // dog
```
go
```golang
type PetType string
// List of PetType
const (
	 PetType = "猫" // compile error
	 PetType = "犬"
)
```

##### openapi-generator version
3.2.3-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
swagger: "2.0"
info:
  title: "example"
  version: "1.0"
paths:
  /pet:
    post:
      parameters:
        - in: body
          name: data
          schema:
            type: object
            properties:
              animal:
                $ref: "#/definitions/PetType"
      responses:
        "200":
          description: ok
definitions:
  PetType:
    type: string
    enum:
      - 猫
      - 犬
    x-enum-names:
      - Cat
      - Dog
```

##### Command line used for generation
```cmd
java -jar openapi-generator-cli-3.2.3-20180824.092221-7.jar generate  -i sample.yaml -g java -o dest
```
##### Steps to reproduce
just generated

##### Related issues/PRs
I could not find it.

##### Suggest a fix/enhancement

I always create a customized config class and add `x-enum-names` extension by overriding  `postProcessModelsEnum` and `updateCodegenPropertyEnum` like [this](https://github.com/osjupiter/swagger-codegen-gradle-sample/blob/master/mytemplate/logic/MySwaggerLogic.groovy).

I think it is not good to positively use multi-byte characters as an api parameter.  
So it is enough to be able to use it by extension.

Thanks.

wing328:
> Are double-byte characters not allowed for enum names in OAS ?

Double-byte characters are allowed for num names in OAS v2, v3.

> I always create a customized config class and add x-enum-names extension by overriding postProcessModelsEnum and updateCodegenPropertyEnum like this.

@osjupiter thanks for the details in reporting the issue. I agree with you the existing solution we'd in place won't meet your requirement and I like your idea of using extensions to control the naming of the enum parameter name.

[...]

PR #1486, opened 2018-11-18
Add support for 'x-enum-values' extension as used by Bungie

Yogarine:
### PR checklist

- [√] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [√] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [√] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [√] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This PR adds support for the 'x-enum-values' extensions as proposed by the Bungie.net API spec. See: [https://github.com/Bungie-net/api](https://github.com/Bungie-net/api#extension-properties-on-openapi-specs-or-how-to-generate-much-cooler-clients-for-the-bnet-api-if-you-want-to-take-the-time-to-do-so)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -3962,11 +3962,28 @@ public class DefaultCodegen implements CodegenConfig {
     }
 
     private void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions) {
-        if (vendorExtensions != null && vendorExtensions.containsKey("x-enum-varnames")) {
+        if (null == vendorExtensions) {
+            return;
+        }
+
+        if (vendorExtensions.containsKey("x-enum-varnames")) {
             List<String> alias = (List<String>) vendorExtensions.get("x-enum-varnames");
             int size = Math.min(enumVars.size(), alias.size());
             for (int i = 0; i < size; i++) {
                 enumVars.get(i).put("name", alias.get(i));
+            }
+        }
+
+        if (vendorExtensions.containsKey("x-enum-values")) {
+            List<Map<String, String>> enumValues = (List<Map<String, String>>) vendorExtensions.get("x-enum-values");
+
+            for (Map<String, String> enumValue : enumValues) {
+                for (Map<String, Object> enumVar : enumVars) {
+                    if (enumValue.get("numericValue").equals(enumVar.get("value"))) {
+                        enumVar.put("name", enumValue.get("identifier").toUpperCase());
+                        break;
+                    }
+                }
             }
         }
     }


jmini:
We have introduced `x-enum-varnames` to control the name associated with the enum value: https://github.com/OpenAPITools/openapi-generator/issues/893

In my opinion this extension make more sense, because the `enum` property needs to conform the type defined by the schema. The additional extension `x-enum-varnames` is about providing a String for the corresponding value.

```
  weather:
    type: object
    properties:
      type:
        type: integer
        format: int32
        enum:
          - 1
          - 2
          - 3
        x-enum-varnames:
          - Sunny
          - Cloudy
          - Rainy
```

Yogarine:
> We have introduced `x-enum-varnames` to control the name associated with the enum value: #893
> 
> In my opinion this extension make more sense, because the `enum` property needs to conform the type defined by the schema. The additional extension `x-enum-varnames` is about providing a String for the corresponding value.
> 
> ```yaml
>   weather:
>     type: object
>     properties:
>       type:
>         type: integer
>         format: int32
>         enum:
>           - 1
>           - 2
>           - 3
>         x-enum-varnames:
>           - Sunny
>           - Cloudy
>           - Rainy
> ```

@jmini Both are extensions that solve the same problem.

Unfortunately Bungie will probably never implement the `x-enum-varnames` extension in their OpenAPI spec. Since OpenAPI Generator is designed for consumers of OpenAPI specs rather than the writers of those specs, I think it is good to support additional custom vendor extensions, even if they are not officially supported. Especially considering it's a small change.

Yogarine:
cc @OpenAPITools/generator-core-team as the change impacts DefaultGenerator.java

@wing328 Could you also take a look-see?

jmini:
@Yogarine thank you a lot, I will try to have a look.

---

As discussed earlier: using the regular `enum` and the 2 extensions `x-enum-varnames` and `x-enum-descriptions` (https://github.com/OpenAPITools/openapi-generator/issues/1693) is the recommend way because if you do not understand the extension, then the Spec still make sense.

I agree with your comment that we can also support this `x-enum-values`.

jmini:
What I do not get: 

Is the idea that both `enum` and `x-enum-values` are set in the Schema?
Is this `numericValue` only for schema of `type: integer`? How is this support to work with String type for example?

Yogarine:
> What I do not get:
> 
> Is the idea that both `enum` and `x-enum-values` are set in the Schema?
> Is this `numericValue` only for schema of `type: integer`? How is this support to work with String type for example?

@jmini You would always need to declare `enum` in the Schema, to remain compatible with the OpenAPI spec.

The `x-enum-values` extension only serves to give additional information about the enum values. This extension is only intended to be used for numeric arrays.

This is how @vthornheart-bng explains it in the [Bungie.net API README](https://github.com/Bungie-net/api#extension-properties-on-openapi-specs-or-how-to-generate-much-cooler-clients-for-the-bnet-api-if-you-want-to-take-the-time-to-do-so):
> **x-enum-values** = I really hate that the OpenAPI spec doesn't have a way to tell you both the identifier and the numeric value of an enum, much less documentation for enum values, so I added this extension as a way to return that info. It'll give you way more than the basically useless base "enum" property.

jmini:
I think there is some consent around the fact that `enum` as defined in OpenAPI v3 is not enough: What Bungie.net API has proposed is close too what is described in https://github.com/OAI/OpenAPI-Specification/issues/681#issuecomment-220334075. Unless that you need to define the value twice:

```json
"Weather":{
    "type":"object",
    "properties":{
        "type":{
            "type":"integer",
            "format":"int32",
            "enum":[
                1,
                2,
                3
            ],
            "x-enum-values":[
                {
                    "numericValue":"1",
                    "identifier":"Sunny",
                    "description":"Blue sky"
                },
                {
                    "numericValue":"2",
                    "identifier":"Cloudy",
                    "description":"Slightly overcast"
                },
                {
                    "numericValue":"3",
                    "identifier":"Rainy",
                    "description":"Take an umbrella with you"
                }
            ]
        }
    }
}
```

---

When we have solved #1693 we will also have support for descriptions:

```json
      "Weather" : {
        "type" : "object",
        "properties" : {
          "type" : {
            "type" : "integer",
            "format" : "int32",
            "enum" : [ 1, 2, 3 ],
            "x-enum-descriptions" : [ "Blue sky", "Slightly overcast", "Take an umbrella with you" ],
            "x-enum-varnames" : [ "Sunny", "Cloudy", "Rainy" ]
          }
        }
      }
```

With the advantage that you do not not have to redeclare that enum values twice (in the `enum` property for OpenAPI compliancy and in the `x-enum-values` as defined by Bungie.net)

---

Of course as you have suggested earlier, we can decide to support both approaches.
This is a decision that the @OpenAPITools/generator-core-team can discuss.

My opinion is that supporting both constructs is OK with following additions for this PR:

* support for descriptions (probably #1752 need to be merged first)
* handling the case where a value is present in `x-enum-values` and not in `enum` (maybe log a warning)
* add some unit tests to catch regressions

What do you think?

Yogarine:
@jmini

> What do you think?

Sounds like a good plan.

jimschubert:
> handling the case where a value is present in x-enum-values and not in enum (maybe log a warning)

I think this should throw an error on spec validation by default. We allow for skipping validation, in which case I think a warning should be logged. But, we have no way to know whether the user considers the x-enum-values or the enum array as the source of truth.  Most likely, an imbalance between the two is a programming or spec authoring error.

FearlessHyena:
Would really like to have this too. Any plans when this will get released?

Issue #1480, opened 2018-11-17
NullPointerException when generating Bungie.net OpenAPI spec


Yogarine:
##### Description

When trying to generate clients for the Bungie.net OpenAPI spec, I get a NullPointerException:
```
Exception in thread "main" java.lang.NullPointerException
	at org.openapitools.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:939)
	at org.openapitools.codegen.DefaultGenerator.processPaths(DefaultGenerator.java:910)
	at org.openapitools.codegen.DefaultGenerator.generateApis(DefaultGenerator.java:488)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:868)
	at org.openapitools.codegen.cmd.Generate.run(Generate.java:349)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:62)
```
(I'll add a full `--verbose` log in a comment.)

##### openapi-generator version

3.3.4-SNAPSHOT (master @ c656194f22f9624ee861abe57964758ae7e7bd39)

##### OpenAPI declaration file content or url

[https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json](https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json)

##### Command line used for generation

```bash
java -jar openapi-generator-cli.jar generate
      --generator-name php
      --config         /local/openapi-config.php.json
      --input-spec     https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json
      --output         /local/bungie-sdk-php
      -DapiTests=false
      -DmodelTests=false
      --verbose
```

openapi-config.php.json:
```json
{
  "packageName": "bungie-sdk-php",
  "composerVendorName": "yogarine",
  "composerProjectName": "bungie-sdk-php",
  "artifactVersion": "2.3.2",
  "invokerPackage" : "Bungie",
  "modelPackage": "Model",
  "apiPackage": "API",
  "variableNamingConvention": "camelCase",
  "srcBasePath": "src",
  "gitUserId": "Yogarine",
  "gitRepoId": "bungie-sdk-php",
  "validateSpec": false
}
```

##### Steps to reproduce

Build with the above command.

Yogarine:
Found the issue. There is a tag at the end with an empty 'name' field:
```json
{
"name": "",
"description": "core"
}
```

Should we just want to ignore the empty name?

PR #1487, opened 2018-11-18
Always use auth methods if they are required by headers

Yogarine:
### PR checklist

- [√] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [√] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [√] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [√] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Always use a security scheme as auth method if the related header implies that it is always required.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -21,10 +21,8 @@ import com.samskivert.mustache.Mustache;
 import com.samskivert.mustache.Template;
 
 import io.swagger.v3.core.util.Json;
-import io.swagger.v3.oas.models.OpenAPI;
-import io.swagger.v3.oas.models.Operation;
-import io.swagger.v3.oas.models.PathItem;
-import io.swagger.v3.oas.models.Paths;
+import io.swagger.v3.oas.models.*;
+import io.swagger.v3.oas.models.headers.Header;
 import io.swagger.v3.oas.models.info.Contact;
 import io.swagger.v3.oas.models.info.Info;
 import io.swagger.v3.oas.models.info.License;
@@ -936,7 +934,8 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
                 for (String tagName : tagNames) {
                     boolean foundTag = false;
                     for (Tag tag : swaggerTags) {
-                        if (tag.getName().equals(tagName)) {
+                        String name = tag.getName();
+                        if (null != name && name.equals(tagName)) {
                             tags.add(tag);
                             foundTag = true;
                             break;
@@ -1136,18 +1135,33 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
     }
 
     private Map<String, SecurityScheme> getAuthMethods(List<SecurityRequirement> securities, Map<String, SecurityScheme> securitySchemes) {
-        if (securities == null || (securitySchemes == null || securitySchemes.isEmpty())) {
+        if (securitySchemes == null || securitySchemes.isEmpty()) {
             return null;
         }
+
         final Map<String, SecurityScheme> authMethods = new HashMap<>();
-        for (SecurityRequirement requirement : securities) {
-            for (String key : requirement.keySet()) {
-                SecurityScheme securityScheme = securitySchemes.get(key);
-                if (securityScheme != null) {
-                    authMethods.put(key, securityScheme);
+
+        Components components       = openAPI.getComponents();
+        Map<String, Header> headers = null != components ? components.getHeaders() : null;
+
+        for (String key : securitySchemes.keySet()) {
+            SecurityScheme securityScheme = securitySchemes.get(key);
+            Header header                 = null != headers ? headers.get(securityScheme.getName()) : null;
+
+            if (null != header && header.getRequired()) {
+                // If the security scheme is required in the headers, always add it.
+                authMethods.put(key, securityScheme);
+            } else if (null != securities) {
+                // Otherwise, check if it is set in the securities.
+                for (SecurityRequirement requirement : securities) {
+                    if (requirement.get(key) != null) {
+                        authMethods.put(key, securityScheme);
+                        break;
+                    }
                 }
             }
         }
+
         return authMethods;
     }
 }


Yogarine:
I unintentionally also included the fix for #1480 in this, but it's a very small cleanup change anyway so I'll just leave it in there.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -1136,18 +1135,37 @@ private static String generateParameterId(Parameter parameter) {
...
-            for (String key : requirement.keySet()) {
-                SecurityScheme securityScheme = securitySchemes.get(key);
-                if (securityScheme != null) {
-                    authMethods.put(key, securityScheme);
+
+        Components components       = openAPI.getComponents();
+        Map<String, Header> headers = (null != components) ? components.getHeaders() : null;
+
+        for (String key : securitySchemes.keySet()) {
+            SecurityScheme securityScheme = securitySchemes.get(key);
```
wing328: @Yogarine What IDE did you use? Seems like it's aligning based on the "="

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -936,7 +934,8 @@ private void processOperation(String resourcePath, String httpMethod, Operation
                 for (String tagName : tagNames) {
                     boolean foundTag = false;
                     for (Tag tag : swaggerTags) {
-                        if (tag.getName().equals(tagName)) {
+                        String name = tag.getName();
+                        if (null != name && name.equals(tagName)) {
```
wing328: Minor suggestion: use `name != null` instead to make it more consistent with the rest of the code base.

wing328:
> Always use a security scheme as auth method if the related header implies that it is always required.

Can you give a minimal example when the related header "implies" it's required?

The Bungie API spec contains ~35k lines, which is not the best spec to illustrate the fix IMO.

wing328:
cc @OpenAPITools/generator-core-team as the change impacts DefaultGenerator.java

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
Yogarine: IntelliJ IDEA. This is a personal preference though. I find it more readable when consecutive assignments are aligned.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
Yogarine: I prefer starting with the value to prevent accidental ‘name = null’ typos which are hard to debug, but if you prefer I’ll switch them around for consistency.

Yogarine:
> > Always use a security scheme as auth method if the related header implies that it is always required.
> 
> Can you give a minimal example when the related header "implies" it's required?
> 
> The Bungie API spec contains ~35k lines, which is not the best spec to illustrate the fix IMO.

From the Bungie spec:
```json
"components": {
    "headers": {
        "X-API-Key": {
            "description": "When you have registered an Application at https://www.bungie.net/en/Application, you will receive an API key.  You should pass it in via this header with every request.",
            "required": true,
            "allowEmptyValue": false,
            "schema": {
                "type": "string"
            }
        }
    },
    "securitySchemes": {
        "apiKey": {
            "type": "apiKey",
            "description": "Every request requires an API key.  To get an API key, register a new application at https://www.bungie.net/en/Application.",
            "name": "X-API-Key",
            "in": "header"
        },
```

Here you can see the X-API-Key is required. However, the `apiKey` security isn't included in any of the `paths`.

ackintosh:
I think it is better to update the spec instead of adding something features to generator. The operation objects defined in the Bungie spec should contain the parameter(Header Object) and the security definition(Security Requirement Object) if needed.  👀

Yogarine:
> I think it is better to update the spec instead of adding something features to generator. The operation objects defined in the Bungie spec should contain the parameter(Header Object) and the security definition(Security Requirement Object) if needed. 👀

Yeah, they should. The problem is I doubt they will ever do that.

Since the OpenAPI Generator is a tool focused on the consumers of a spec, and not the writers of a spec, it should try to be as compatible as possible. And if the headers imply that a header is required for every requests, then it really helps if the generator provides for that.

All in all I believe this is an elegant solution that won't break anything.

Yogarine:
@wing328 I made changes according to the feedback you gave.

@ackintosh What are you're thoughts? I think there is nothing wrong with offering better compatibility when the spec fails to explicitly declare the authentication in all operations when it's implied by the global headers.

ackintosh:
Hmm... personally, I can't agree with that. For me, the spec doesn't seems implying that the `X-API-Key` header is required.

Code generation is based on a spec, but the spec doesn't have both of Header Object and Security Requirement Object. 👀

Yogarine:
> Hmm... personally, I can't agree with that. For me, the spec doesn't seems implying that the `X-API-Key` header is required.
> 
> Code generation is based on a spec, but the spec doesn't have both of Header Object and Security Requirement Object. 👀

Let's say you have this in your `/components/headers`:
```json
        "headers": {
            "X-API-Key": {
                "description": "When you have registered an Application at https://www.bungie.net/en/Application, you will receive an API key.  You should pass it in via this header with every request.",
                "required": true,
                "allowEmptyValue": false,
                "schema": {
                    "type": "string"
                }
            }
        }
```
_(taken from the bungie spec)_

It has `"required": true`, which implies it's always required.

Unfortunately the OpenAPI spec is not really clear what should be done if a header is declared as `required`. However my interpretation (and clearly also Bungie's) is that the header should be provided in all requests.

Yogarine:
> Hmm... personally, I can't agree with that. For me, the spec doesn't seems implying that the `X-API-Key` header is required.
> 
> Code generation is based on a spec, but the spec doesn't have both of Header Object and Security Requirement Object. 👀

@ackintosh Let's say you have this in your `/components/headers`:

```json
        "headers": {
            "X-API-Key": {
                "description": "When you have registered an Application at https://www.bungie.net/en/Application, you will receive an API key.  You should pass it in via this header with every request.",
                "required": true,
                "allowEmptyValue": false,
                "schema": {
                    "type": "string"
                }
            }
        }
```
_(taken from the Bungie spec)_

It has `"required": true`, which implies it's always required.

Unfortunately the OpenAPI spec is not really clear what should be done if a header is declared as `required`. However my interpretation (and clearly also Bungie's) is that the header should be provided in all requests.

Dayde:
Any update on this subject ?

jimschubert:
I agree with @ackintosh that there's nothing in this spec example that says the auth method should be used for each request (other than a comment in the [current bungie spec](https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json) on the securitySchemes object for apiKey). `"required": true` means that the header is required to be included during transmission, and the `"allowEmptyValue": false` just means it needs to have something in the header.

If we begin implying security that isn't defined in the `security` object on a path, this would work according to Bungie's definition, but it could potentially be a security risk for other similarly valid but semantically incorrect definitions. Because adding this non-standard behavior could be a security risk for others and a liability for the project, I think we wouldn't be able to merge this unless it was behind an opt-in flag which very clearly warns about the dangers of implying security in generated code.

Alternatively, I'm working toward improving upon our generator workflow and extensibility by the 5.0.0 release. My idea is to convert the input specification to outputs for code via transformations, which would allow us to load user-defined transformations for scenarios exactly like this. That work is under [project 5](https://github.com/openapitools/openapi-generator/projects/5).

PR #1469, opened 2018-11-16
Upgrade haskell-servant generator to latest LTS

f-f:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Problem: the current `haskell-servant` client is based on a set of libraries (`lts` package set) from one year ago, based on the `8.0.x` release of GHC. It is typical for library writers in the Haskell community to maintain compatibility with the latest 3 major versions of the compiler. By this standard (since `8.2`, `8.4` and `8.6` are out) we are generating an outdated client. Moreover, the `servant` library itself went through several breaking versions in the meanwhile, so the code generated by this generator is not really compatible with a project started today.

This PR upgrades to the latest `lts` ([12.18 for GHC 8.4](https://www.stackage.org/lts-12.18)) and `servant` version, and adds several new features:
- optionals are now `Maybe`s
- `UUID`, `UTCTime` and `Day` are now represented with the correct types
- the configuration for the client now contains only one parameter, the `url`
- add `ToSchema` instances to generated types, so they can be re-exposed for free in the spec of the service using the client
- switch to the `tls-http-manager`, so we can connect to https urls
- add Nix support in the `stack.yaml`
- add a nicer API to call the endpoints

Note: all of this is quite breaking, so I opened the PR against the `4.x` branch

/cc @wing328

PR #1499, opened 2018-11-20
[Haskell][servant] fix "list" type

wing328:
### PR checklist

### Description of the PR

cc @adamConnerSax 

Also cc @jonschoning (haskell-http-client author), @algas (haskell-servant author)

Related issue: https://github.com/haskell-servant/servant/issues/1084


f-f:
Most of these things are already being fixed in https://github.com/OpenAPITools/openapi-generator/pull/1469, so I'd propose we wait for that to be merged and see what we can further improve afterwards

adamConnerSax:
I'm happy with that.  I didn't realize there were fixes already being worked on.  Thanks!

wing328:
@adamConnerSax can you please also test #1469 when you've time?

f-f:
@adamConnerSax I think it makes sense that the generated code doesn't work if the spec you're feeding to it is not valid?

Re: changing the name of the backend: I think it makes sense to have this one default to `haskell`, as it's the simplest one to integrate (I did try working with the `haskell-http-client`, and it requires much more work

f-f:
@adamConnerSax: oh I see, I'll look into this

adamConnerSax:
@f-f Another thing, though one I must be misunderstanding:
In your version, the naming of fields works like `endpointNameFieldName :: FieldType` and then the JSON instances strip off the leading `EndPointName` so they become `FieldName`.  But in the json you get from the server, that will be `field_name` and so those instances won't work in the sense of automatically decoding correctly?
But that must be a thing you've tested already since it's the basic function of those types.  So where is that handled? 
In the version I was working on, the field names were "endpointName_field_name" and the json instances stripped off the "endpointName_" part so the instances matched and decoded correctly.

f-f:
@adamConnerSax you're referring to [this change here in this PR](https://github.com/OpenAPITools/openapi-generator/pull/1499/files#diff-547378a537a613220d3c055afc56f5c2L577)?
I don't know, we didn't encounter any problem with the current logic there (as in, I didn't change about that logic in my PR): I just added a `uncapitalize` function so that the fields are never capitalized (as only types can be capitalized).
But it looks like you're right, as I fail to see how we can preserve a bijection in the props names if we `camelize` that.

Btw the failure you're encountering [here](https://github.com/OpenAPITools/openapi-generator/pull/1499#issuecomment-440400218) was due to me adding generic `ToSchema` istances, which are now not added in case you're having a type which doesnt support that (like e.g. `Data.Aeson.Value`). Fixed in this commit: fe942ceabe

Would you like to add tests to this generator so that we can catch all these corner cases automatically?

f-f:
I think the last thing left to fix (but not blocking I think) would be the slow compilation time. Basically the issue is that we just dump all the types is a single file, which gets huge, so GHC struggles.
The solution for that would be to have a single file per type, in which we import all the types it depends on, and then in the current `Types.hs` we'd just reexport everything.

jonschoning:
> import all the types it depends on
I had wanted to do this, but I have seen examples in the wild in which there are circular dependencies in the models, which makes the imports method difficult if not impossible.  
Of course one could also just say those clients will not be supported (but for my generator i have a goal of maximizing support)

f-f:
> i.e. module-per-type ?

Yes, sorry

> I had wanted to do this, but I have seen examples in the wild in which there are circular dependencies in the models, which makes the imports method difficult if not impossible.
> afaict, although the generators in other langs make separate files, they still dump them all into the same module and aren't calculating imports.

Oh right, this would break down for circular imports. A solution for that would be to detect cycles and put together in the same module all types which form a cycle. Can be done, but it's some work.
I wondered multiple times if we have something generic for that in openapi-generator? I mean, I cannot be the first one who thought about topological sorting and cycle detection for schemas?

A middle ground in our case could be to define all types in a `Types.Core` module, and then reexporting all of them from `Types`, where we'd derive all the (now orphan) generic instances (i.e. `ToJSON`, `FromJSON`, `ToSchema`).
Another solution would be to actually generate the code from the instances, instead of generically deriving them, as that's way less work for GHC (though GHC 8.6/8.8 should improve the situation on that, IIRC I saw some optimization for that going in).

wing328:
Closing this one as it's no longer needed.

Issue #1501, opened 2018-11-20
Skip maven code generation if source file is unchanged


rasmusfaber:
##### Description

When using the Maven task the generator always runs, even if the source file is unchanged.

It would be nice to skip the generation, if there have been no changes.

##### openapi-generator version

3.3.3 / 3.3.4-SNAPSHOT

##### Steps to reproduce

Run maven compile twice in a row, notice that the generator always run, even if there have been no changes.

##### Suggest a fix/enhancement

It might be difficult to track the state, including the maven-options, but if we just compare the timestamp of the source file and the output file, it should be close enough, as long as the skip is not per default

Philzen:
This was implemented a long time ago in 4.x:

```xml
<configuration>
    <skipIfSpecIsUnchanged>true</skipIfSpecIsUnchanged>
</configuration>
```

PR #1397, opened 2018-11-07
Option to overwrite only changed files

wing328:
### PR checklist

- [ ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Cherry-picks comments from #1391 

> Implements feature request #1081 : added --minimal-update option to only overwrite output files that have changed, so as to reduce triggered build steps.

Credits go to @wheezil 

Also cc @OpenAPITools/generator-core-team

Issue #1081, opened 2018-09-21
Feature request: Support overwrite-only-changed


wheezil:
##### Description

When using code-generators in an incremental build, it is time-consuming to have the generator touch all output files and cascade into build dependencies, even if nothing has changed.  While one can usualy make build systems only generate code when the input specification has changed, it is not always possible.  Even then, it would be nice to only update the changed files, so that e.g. changing one field in a module doesn't cause everything to rebuild.

##### openapi-generator version

3.3.0

##### OpenAPI declaration file content or url

N/A

##### Command line used for generation

Currently using a custom generator of our own development.  However I believe the problem/solution to cut across all generators.

##### Steps to reproduce

Run any code-generation.
Note output file timestamps.
Run it again.
Note output file timestamps have all changed.

##### Related issues/PRs

None that I know of.

##### Suggest a fix/enhancement

Add a command-line option to turn on this new behavior.  When this option is specified, the generator will write a temporary file, compare its contents to the existing file, and if they differ, overwrite the existing file.  It is possible to use hash-based conditions, but that probably adds complexity without any concrete benefit.

dennisschagt:
It looks like this has been implemented in https://github.com/OpenAPITools/openapi-generator/pull/2451

PR #1502, opened 2018-11-20
Add option to skip maven execution if source is older than output

rasmusfaber:
### PR checklist

- [X ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [N/A] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [N/A ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Implementation of  #1501

### DIFF
--- a/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
+++ b/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
@@ -56,6 +56,10 @@ import org.sonatype.plexus.build.incremental.DefaultBuildContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Charsets;
+import com.google.common.hash.Hashing;
+import com.google.common.io.Files;
+
 /**
  * Goal which generates client/server code from a OpenAPI json/yaml definition.
  */
@@ -318,6 +322,12 @@ public class CodeGenMojo extends AbstractMojo {
     private Boolean skip;
 
     /**
+     * Skip the execution if the source file is older than the output folder.
+     */
+    @Parameter(name = "skipIfSpecIsUnchanged", property = "codegen.skipIfSpecIsUnchanged", required = false, defaultValue = "false")
+    private Boolean skipIfSpecIsUnchanged;
+
+    /**
      * Add the output directory to the project as a source root, so that the generated java types
      * are compiled and included in the project artifact.
      */
@@ -363,6 +373,21 @@ public class CodeGenMojo extends AbstractMojo {
                                         "Code generation is skipped in delta-build because source-json was not modified.");
                                 return;
                             }
+                        }
+                    }
+                }
+            }
+
+            if (skipIfSpecIsUnchanged) {
+                if (inputSpecFile.exists()) {
+                    File storedInputSpecHashFile = getHashFile(inputSpecFile);
+                    if(storedInputSpecHashFile.exists()) {
+                        String inputSpecHash = Files.asByteSource(inputSpecFile).hash(Hashing.sha256()).toString();
+                        String storedInputSpecHash = Files.asCharSource(storedInputSpecHashFile, Charsets.UTF_8).read();
+                        if (inputSpecHash.equals(storedInputSpecHash)) {
+                            getLog().info(
+                                    "Code generation is skipped because input was unchanged");
+                            return;
                         }
                     }
                 }
@@ -606,6 +631,17 @@ public class CodeGenMojo extends AbstractMojo {
             if (buildContext != null) {
                 buildContext.refresh(new File(getCompileSourceRoot()));
             }
+
+            // Store a checksum of the input spec
+            File storedInputSpecHashFile = getHashFile(inputSpecFile);
+            String inputSpecHash = Files.asByteSource(inputSpecFile).hash(Hashing.sha256()).toString();
+
+            if (storedInputSpecHashFile.getParent() != null && !new File(storedInputSpecHashFile.getParent()).exists()) {
+                File parent = new File(storedInputSpecHashFile.getParent());
+                parent.mkdirs();
+            }
+            Files.asCharSink(storedInputSpecHashFile, Charsets.UTF_8).write(inputSpecHash);
+
         } catch (Exception e) {
             // Maven logs exceptions thrown by plugins only if invoked with -e
             // I find it annoying to jump through hoops to get basic diagnostic information,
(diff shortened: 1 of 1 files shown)


modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
```diff
@@ -368,6 +374,16 @@ public void execute() throws MojoExecutionException {
                 }
             }
 
+            if (skipIfSourceIsOlderThanOutput) {
+                if (inputSpecFile.exists() && output.exists()) {
+                    if (inputSpecFile.lastModified() <= output.lastModified()) {
```
jimschubert: `inputSpecFile` here would reference an OpenAPI document and `output` refers to the output directory.

Is the intention of `skipIfSourceIsOlderThanOutput` to only regenerate if someone has modified the spec document since the last regeneration? If this is the case, I would suggest renaming the property to avoid confusion… something like `skipIfSpecIsUnchanged` might be clearer to those reading the properties list. Even after reading the issue and scanning the PR's changes, I thought this was doing something similar to `skipOvewrite`.

Although… my suggested `skipIfSpecIsUnchanged` might suggest to users that we're storing the atime/mtime of the spec (which we could definitely persisted to `{output}/.openapi-generator/SOME_FILENAME`).

rasmusfaber:
Thanks for the detailed comments.

I have changed the implementation to generate a SHA-256 checksum of the spec and store it in the .openapi-generator folder under the name <specfile>.sha256 and then use that to compare.

The generation of the checksum happens in the CodeGenMojo-class, so it only runs when using Maven. Another option would be to have DefaultGenerator generate it, but that would have higher impact.

wing328:
LGTM (but no expert in the Maven plug-in either)

wheezil:
Greetings!  @wing328  pointed out that we have similar PRs addressing similar problems, so I'd like to have a little discussion to see if we should choose one approach or both.  My PR is https://github.com/OpenAPITools/openapi-generator/pull/1397/files and issue https://github.com/OpenAPITools/openapi-generator/issues/1081.
They are related in the sense that they are trying to solve the same problem, which is not updating the output files unless the spec has changed.
I believe that my approach is better at least for some cases (of course I would!), because 

- It is insensitive to the vagaries of file modification times, and so the example given by @jimschubert -- files are manually touched -- doesn't matter at all. It simply generates the code, but as output files are about to be written, it shunts them to a temp file first, compares the temp file to any existing file, and then replaces the existing file (if any) with the temp file only if there are changes. 
- Works on the command-line in addition to maven, so code-generation in other build environments also benefits.

The thing I don't understand about the difference between my approach and this approach is how it may or may not affect downstream build steps. If the maven plugin change avoids firing the code-generation step at all, this may result in fewer ripples down the dependency tree. Unfortunately I don't understand maven plugins and can't judge.

One other thing to note: my approach will of course always generate new output files if the output is cleaned first.  It basically has no relation to the clean lifecycle.  Because of that, build-without-clean may result in lingering output files whose specification has since disappeared -- I don't check for *removal* of an output file, but perhaps I should?

@rasmusfaber your thoughts appreciated.

jimschubert:
@wheezil I don't think your contributions are doing the same thing.

This PR speeds builds by not wasting cycles if your spec is identical to the spec that previously generated your code. This means that once you've generated code, you assume it shouldn't regenerate unless your spec's hash changes. Maven builds are notoriously slow, which is why I don't personally use it and therefore have little experience with it. This PR helps there.

Your PR is only changing files that are not identical to the contents that would be generated. While that  could be useful in many cases, it would cause issues for example where someone has modified templates to output template generation time in a file header/comment. If the spec hasn't changed for 20 builds, there's not really a reason to run codegen 20 times just to update that header.

I do think the work in this PR could be moved from the Maven plugin into Codegen core, so it'll work the same across all plugins/entrypoints. But I don't think that'll have an effect on your PR. The difference between "should I generate at all" and "should I replace this generated file" are significant enough to keep separate.

wheezil:
@jimschubert Thanks for comment.  I see what you mean that they are addressing separate concerns.

wheezil:
Is this check based purely on file times, and if so does it chase referenced files as well?

wheezil:
@jimschubert regarding performance of the build, is this time spent running the generator, or is it time spent compiling the resulting code?  I have not yet used a large API so I don't know what to expect.
Is it faster to use the maven plugin or to launch the command-line?

jimschubert:
@wheezil I'm confused by your comment about file time or contents (I can't tell if you're referring to spec or generated code). The work in this PR compares the sha hash of a spec document, and if that matches the hash stored in the existing generated code, it skips executing the generator at all. It has nothing to do with the generated files, other than the metadata file storing the hash. When you ask about referenced files, it's something I'm considering to improve upon when I move into core... most likely parsing into the OpenAPI structure, then hashing against those contents. This would mean rather than file modifications, it would regenerate on changes in our abstraction across versions (which often results in fixes or changes in generated code). This would probably need an option to allow file-based or object-based hashing for those who don't want to consider parsing differences to be spec differences.

Your question about Maven is hard to answer. Maven is slow when you have many dependencies, many projects, and other factors. It doesn't have built in caching like Gradle/SBT or other JVM targeting build systems. There are options out there for speeding up builds, but they can be difficult to setup. Maven has parallel build as an option, although I think it's fairly recent. I don't know if it's better to use the exec plugin because, as I mentioned I'm not that familiar with Maven (this being the only project I touch that uses it, and I don't do much of the build management). I do know that I've had issues trying to get the exec plugin to work in the past when I was trying to integrate the gradle plugin into the build. I would personally stick with the Maven plugin, since it has cleanly into the architecture. Or, use our gradle plugin ;)

wheezil:
@jimschubert thanks for clarifying that it compares the contents (hash) of the spec file.  I was confused and thought it was using modification times.

Regarding maven/gradle, I will stick with maven, since that's our team standard.  Thanks again.

jimschubert:
@wheezil I understand the confusion. The original implementation used file modification time compared to output directory modification time.

wing328:
@rasmusfaber thanks for the PR, which has been included in the v3.3.4 release: https://twitter.com/oas_generator/status/1068772409795207168

PR #1512, opened 2018-11-21
[haskell-http-client] fix typename generation bug for array params; update deps

jonschoning:
- fix build for example-app + tests-integration in Samples (due to change in Category)

- upgrade stack to 12.18

- fix typename generation bug for array params


jonschoning:
fix `amendment_indicator` typename bug tested with with `http://api.open.fec.gov/swagger`
via cmdline
```
java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i output/fec.gov.json -l  haskell-http-client -o samples/client/openfec/haskell-http-client --skip-validate-spec
```
re https://github.com/OpenAPITools/openapi-generator/pull/1499

PR #1514, opened 2018-11-21
[elixir] Typespec - allow null on optional struct-attributes

mrmstn:
### PR checklist

### Description of the PR

This PR allows null-values for struct-attributes which are not marked as required. This is very important since the compiler would mark this as a error since a null value in a struct would break the typespec contract. 

With this small typespec fix, non-required and nullable attributes will be markes as possible nil values for the given model

image


wing328:
@mrmstn thanks for the PR. Shall we also use the `nullable` mustache tag in the Elixir client? (we can definitely do it in another PR instead)

Ref: 
- https://github.com/OpenAPITools/openapi-generator/pull/1059/files#diff-0909ca3c214075716264f5878b273ee6R46
- https://github.com/OpenAPITools/openapi-generator/pull/1073/files#diff-37c4b71d084f33d24c1b198a37922680R67

mrmstn:
I think this would be a good idea - I kind of missed the existance of `nullable` all alonge :see_no_evil:

mrmstn:
I also added the nullable type to the musache file.

Based on the following swagger 2.0 defintion, this PR will result to the following results
```yml
definitions:
  Entity:
    type: object
    properties:
      requiredDate:
        type: string
        format: date
        x-nullable: false
      notRequiredDate:
        type: string
        format: date
        x-nullable: false
      nullableRequiredDate:
        type: string
        format: date
        x-nullable: true
      nullableNotRequiredDate:
        type: string
        format: date
        required: false
        x-nullable: true
    required:
      - requiredDate
      - nullableRequiredDate
```

```elixir
defmodule SwaggerPetstore.Model.Entity do
  @moduledoc """
  
  """

  @derive [Poison.Encoder]
  defstruct [
    :"requiredDate",
    :"notRequiredDate",
    :"nullableRequiredDate",
    :"nullableNotRequiredDate"
  ]

  @type t :: %__MODULE__{
    :"requiredDate" => Date.t,
    :"notRequiredDate" => Date.t | nil,
    :"nullableRequiredDate" => Date.t | nil,
    :"nullableNotRequiredDate" => Date.t | nil
  }
end
```

As you can see, even if you didn't allow a non required attribute to be null, the typespec still will marks this as possible null. This is given to the fact that elixir will initiate all attributes in a struct as null and therefore a "optional" attribute in a struct can only be null and not "not present"

PR #229, opened 2018-06-05
[Swift 4] add compatibility for unwrapRequired config option

joeboyscout04:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the correct branch: Default: `master`.
- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.
@jgavris, @ehyche, @Edubits, @jaz-ah, @d-date 

### Description of the PR

I added compatibility for unwrapRequired config option for Swift 4 to fix #228 .

Issue #228, opened 2018-06-05
[Swift 4]  unwrapRequired option doesn't do anything


joeboyscout04:
##### Description

It looks like the unwrapRequired config option doesn't do anything in Swift 4. It is there in Swift 3 generator but not in Swift 4. It's apparent by looking at the mustache templates for Swift 4 that that option doesn't exist anywhere in the templates.

##### openapi-generator version

3.0.0

##### Steps to reproduce

Readily reproducible in any project using this option and swift 4.

##### Related issues/PRs

Coming soon

PR #1527, opened 2018-11-22
[Swift] Fix unwrapRequired behavior so it actually works

joeboyscout04:
### PR checklist

@clasnake @jimschubert  @shijinkui  @ramzimaalej 

### Description of the PR
Previously the unwrapRequired config option didn't exist for swift4 generator (see #228), but the previous PR (#229) didn't fully fix the issue due to some faulty logic and a missing default value.  That has now been corrected so this option should work again.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift4Codegen.java
```diff
@@ -59,7 +59,7 @@
     protected static final String LIBRARY_RX_SWIFT = "RxSwift";
     protected static final String[] RESPONSE_LIBRARIES = {LIBRARY_PROMISE_KIT, LIBRARY_RX_SWIFT};
     protected String projectName = "OpenAPIClient";
-    protected boolean unwrapRequired;
+    protected boolean unwrapRequired = true;
```
wing328: @joeboyscout04 thanks for the fix. Since the default was "false", we'll need to target this in 4.0.x (upcoming major release)

If you change it back to "false" as the default, we can merge it into master (which only allows backward-compatible changes)

wing328:
@joeboyscout04 thanks for the PR. If I understand correctly, is the `unwrapRequired` option same as or very simliar to the `nullable` feature in OAS3?

We've start adding nulalble support to the generators:
- https://github.com/OpenAPITools/openapi-generator/pull/1073
- https://github.com/OpenAPITools/openapi-generator/pull/1059 

so maybe we can use `nullable` attribute in the spec to replace the option "unwrapRequired"

 In OAS2, one can use `x-nullable` vendor extension instead.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift4Codegen.java
joeboyscout04: Unless I'm misunderstanding what this is supposed to do, it is already unwrapping the required parameters by default.  I think this option never actually did anything.

wing328: I mean unwrapRequired defaults to false at the moment. Changing the default setting of `unwrapRequired` to "true" will impact users who do not need this feature. 

(I didn't say your PR does not fix the issue and I agree with you the option may have been broken as it's not switched on by default)

fl034:
@wing328 I really hope that this gets into 4.0 or a 3.4 bugfix release. Having models with all properties as optional (nullable) is a big misuse of the programming language feature.

I currently have multiple projects running on API client code generated by generator of this pull request...

wing328:
@joeboyscout04 we definitely want to include your fix.

I'll try to update your PR to resolve the merge conflicts later today or tomorrow.

cc @jgavris (2017/07) @ehyche (2017/08) @Edubits (2017/09) @jaz-ah (2017/09) @d-date (2018/03)

wing328:
I've filed https://github.com/OpenAPITools/openapi-generator/pull/3783 to use OAS v3 `nullable` (or x-nullable in OAS v2) instead.

wing328:
The fix is available in the newly-added swift5 generator so closing this one instead.

PR #1538, opened 2018-11-23
[elm] Allow setting base path for each request

### LINKED ISSUE #1290: [elm] Add possibility to override basePath
##### Description

Generating the client for elm takes the current url as the basePath and gives no possibility to override it.

This is absolutely OK when an external API is used. But generating the client for a project which is shipped together with the API has always localhost in it. This should be customizable with the current URL under which the project is served.

v3.3.1

Not a specific one.

generate -i http://localhost/swagger/v1/swagger.json -g elm -o /generated

##### Steps to reproduce

1. Have a swagger.json file
2. Generate the elm client with the command above
3. Open one of the Request files
4. See the basePath is localhost and cannot be overriden or passed in.

trenneman:
### PR checklist

### Description of the PR

Allow setting your own `basePath` for each request instead of using the one from your OpenAPI spec.

Fixes #1290. @wing328 can you review?

### DIFF
--- a/docs/generators/elm.md
+++ b/docs/generators/elm.md
@@ -9,4 +9,7 @@ CONFIG OPTIONS for elm
 	elmPrefixCustomTypeVariants
 	    Prefix custom type variants (Default: false)
 
+	elmEnableCustomBasePaths
+	    Enable setting the base path for each request (Default: false)
+
 Back to the [generators list](README.md)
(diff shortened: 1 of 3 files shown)


wing328:
@trenneman you may want to add the feature without introducing another option as other clients (e.g. Ruby, C#, PHP, etc) allow setting the base path (e.g. change it to localhost:8080 for development) by default

trenneman:
@wing328 Elm is a purely functional language. It has no classes, dependency injection or global state and also no typeclasses (it is designed to be very explicit about what you want). I therefore think passing it as a parameter (after adding a flag) makes the most sense.

wing328:
@trenneman thanks for the explanation. Let's go with your enhancement.

pecigonzalo:
@eriktim Im not sure I understand how this is intended to be used, how do we pass the new base path?

eriktim:
@pecigonzalo if you generate the code using the flag `--additional-properties=elmEnableCustomBasePaths=true` each request function takes an additional parameter `baseUrl : String`.

pecigonzalo:
Thanks a lot for the explanation, that works!

Issue #1517, opened 2018-11-21
Python client code improperly escaping regex strings


Raznic:
##### Description

I've got a web server running Django + REST Framework that I am trying to generate Python client code for. One of my Serializers defines a RegexField that includes `\d` as a part of the regular expression. The generator appears to be escaping these characters in the generated client code.

Serializer regex: `r'^\d$'`
Actual Python client code: `re.search(r'^\\d$', value)`
Expected Python client code: `re.search(r'^\d$', value)`

##### openapi-generator version

I've pulled the latest from `master` and the version shows as `3.3.4-SNAPSHOT`.

##### Command line used for generation

```bash
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g python -i <development Swagger URL> -o python
```

##### Steps to reproduce

1. Define a Serializer class with a RegexField in your Django + REST Framework app.
```python
from rest_framework.serializers import ModelSerializer, RegexField
from .model import MyModel
class MyModelSeralizer(ModelSerializer):
    value = RegexField(regex=r'^\d$')

    class Meta:
        model = MyModel
        fields = '__all__'
```
2. Generate Python client code.

3. Try to create a model.
```python
import openapi_client

config = openapi_client.Configuration()
client = openapi_client.ApiClient(client)
api_instance = openapi_client.MyModelApi(client)

data = openapi_client.MyModel(value='9')
api_instance.my_model_create(data)
```
##### Related issues/PRs

#1392 

##### Suggest a fix/enhancement

My Java is pretty rusty, but I would assume the fix for the Python client would be similar to the one for the Ruby client from the related issue/PR.

Raznic:
After a little more testing, it appears the problem is not just limited to `\d`. Any backslash character in the regex string is getting escaped.

PR #1539, opened 2018-11-24
Fix escaped regex values in Python client (#1517).

Raznic:
### PR checklist

### Description of the PR

Fixes issue #1517.

This PR corrects an issue with the PythonClientCodegen class improperly generating regular expression strings. It overrides the toRegularExpression() method to not escape backslashes.

CC: @tomplus @taxpon @frol @mbohlool @cbornet @kenjones-cisco


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -633,6 +633,23 @@ public String toDefaultValue(Schema p) {
...
 
+    @Override
+    public String toRegularExpression(String pattern) {
+        if (StringUtils.isEmpty(pattern)) {
+            return pattern;
+        }
+
+        String regexString = pattern;
+        if (!regexString.startsWith("/")) {
+            regexString = "/" + regexString;
```
tomplus: Can't we add a trailing slash too?

```
regexString = "/" + regexString + "/";
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -633,6 +633,23 @@ public String toDefaultValue(Schema p) {
...
+        }
+        if (StringUtils.countMatches(regexString, '/') == 1) {
```
tomplus: What if `regexString` contains `/`, for instance `^\d{4}\/\d{2}$`? It adds the slash at the beginning but the trailing slash won't be added.

Raznic: This is a little tricky. In your example, we would definitely want to add a `/` at the beginning and end of the `regexString`. According to the [OpenAPI Spec 3.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties), the `pattern` property should be a valid [ECMA 262 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5). This allows for the regular expression to define flags after the final `/`. So for a regex such as `^hello.*/i` we would only want to add the `/` at the beginning (i.e. `/^hello.*/i`). 

I couldn't find anything in the ECMA script about the available flags, but I would assume they are the same as those defined in [Mozilla's developer docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags_2). If that's the case, maybe we add a check to see if the characters after the final `/` are all valid regex flags. If they are, we don't append a `/`, otherwise we do. Thoughts?

ggershoni: Later in the ECMA 262 doc it explains that RegExp Objects are modeled on regular expressions in Perl 5 (https://www.ecma-international.org/ecma-262/5.1/#sec-15.10) and the Perl 5 doc talks about modifiers i.e. `/pattern/i` (https://perldoc.perl.org/perlre.html#Modifiers)

ggershoni: @Raznic in [PR 1521](https://github.com/OpenAPITools/openapi-generator/pull/1521/files) @meganemura replaced the code I put in the `AbstractRubyCodegen.java` with the method `addRegularExpressionDelimiter` from `DefaultCodegen.java`.  See [code](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L3994). If you start your pattern with a forward slash it leaves it alone and if you didn't start it with a forward slash it escapes any it finds in the pattern and then surrounds the text with forward slashes.

Raznic: @ggershoni thanks for the links! I did try using `addRegularExpressionDelimiter` from `DefaultCodegen.java` initially. However, I found this method escapes any `/` in the pattern, which still results in an incorrect regex string. I see that the PR you referenced removes a few test cases where regex patterns have either a leading or trailing `/`, but not both. If we don't have to worry about those cases, then I'm okay with just checking for the leading `/` and encapsulating the pattern with `/` if necessary.

tomplus: Thanks. I wasn't aware that modifiers can be added here. It's more complicated because Python `re` module doesn't support it, it uses [flags](https://docs.python.org/3/library/re.html#module-contents) instead.

case-insensitive pattern matching - perl version:
```perl
print 'A' =~ m/a/i;    # output: 1
```

python:
```python
print(re.match(r'A', 'a', re.I))  # output:<_sre.SRE_Match object at 0x7fbb97b6d6b0>
```

We have to map Perl's modifiers to Python's flags. But for time being we can raise an exception if modifiers are used. Escaped slash (`\/`) should be treated as a normal char.

WDYT?

Raznic: I like the idea of raising an error for modifiers for the time being, though it would be prudent to capture that as a feature idea in a separate issue. And I agree with the idea of treating escaped forwarded slashes as normal characters. Time permitting, I'll try to push some new code up a bit later tonight. Thanks again for all the feedback!

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientCodegenTest.java
```diff
@@ -54,4 +59,37 @@ public void testAdditionalPropertiesPutForConfigValues() throws Exception {
...
+            expectedExceptions = {IllegalArgumentException.class},
+            expectedExceptionsMessageRegExp = "Python client does not currently support regular expression modifiers."
+    )
+    public void testRegularExpressionModifiersOpenAPISchemaVersion3() {
+        final OpenAPI openAPI = new OpenAPIParser()
+                .readLocation("src/test/resources/3_0/issue_1517.yaml", null, new ParseOptions()).getOpenAPI();
+        final PythonClientCodegen codegen = new PythonClientCodegen();
+        final String path = "/modifiers";
+        final Operation p = openAPI.getPaths().get(path).getGet();
+        final CodegenOperation op = codegen.fromOperation(path, "get", p, openAPI.getComponents().getSchemas());
```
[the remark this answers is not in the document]
Raznic: Yup, it does. If you look at the `@Test` annotation it declares that the test is expected to throw an IllegalArgumentException with the error message "Python client does not currently support regular expression modifiers."  If you comment those two lines out and try running the test, it will fail with the aforementioned exception/message.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -379,6 +385,19 @@ public String toApiDocFilename(String name) {
         return toApiName(name);
     }
 
+    @Override
+    public String addRegularExpressionDelimiter(String pattern) {
+        if (StringUtils.isEmpty(pattern)) {
+            return pattern;
+        }
+
+        if (!pattern.matches("^/.*")) {
```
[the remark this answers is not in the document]
Raznic: Hmm, I tried removing the `.*` and re-running my unit test, but I'm getting a couple of failures. I think it does need the `.*`, otherwise it will check to see if the pattern literally matches the string `"/"` and nothing else. Having the `.*` there says match any string that starts with `/` and has any number of characters afterwards.

Raznic:
So it looks like a few of the integration tests are now failing because of the new IllegalArgumentException for unsupported modifiers. What are everyone's thoughts on just ignoring any modifiers (i.e. truncate the string after the final `/`) and logging a warning message instead?

Raznic:
Actually, it looks like the Python client is already ignoring any modifiers. It's grabbing everything between the first and last `/`, so any modifiers after the final `/` are dropped. 

https://github.com/Raznic/openapi-generator/blob/14055be4e51204a35f78f0ce29d60bad32ce0c13/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java#L330

ggershoni:
the `/i` modifier is quite useful as it allows you to ignore case when matching.  But I found [here](https://www.regular-expressions.info/modifiers.html) that one can also trigger some modifiers in the regex pattern itself:
```
>>> print(re.match(r'(?i)Pattern', 'PATTERn'))
<re.Match object; span=(0, 7), match='PATTERn'>
```

FYI Ruby:
```
irb(main):003:0> /(?i)Pattern/.match('PATTERn')
=> #<MatchData "PATTERn">
irb(main):004:0> /Pattern/i.match('PATTERn')
=> #<MatchData "PATTERn">
```

So spitting out a warning sounds fair.  Though considering that some people using OpenAPI Generator will not be able to modify the API spec they are trying to build an SDK for we probably do need to capture handling of modifies (as per spec) as a feature idea.

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientCodegenTest.java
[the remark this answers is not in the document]
tomplus: I see it now, thanks.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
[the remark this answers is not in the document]
tomplus: Sorry, you are right.

tomplus:
@Raznic logging warnings sounds good to me, thanks

@ggershoni nice tip with "embedded" modifiers :)

Raznic:
@tomplus I did a little more digging, and it looks like PythonClientCodegen.java does currently support regex modifiers. I ran `./bin/python-petstore-all.sh` and noted that  at the very least the `re.IGNORECASE` modifier was showing up in some of the generated files.

tomplus:
@Raznic Shame on me :) indeed there are a map to translate modifiers to flags.

LGTM, could you confirm that regenerating python samples doesn't introduce any other changes?

Raznic:
@tomplus no worries, took me awhile to realize it as well :smile:. I re-ran the `bin/python-petstore-all.sh` and `bin/security/python-petstore.sh` and confirmed there were no errors. However, Git shows changes to a number of Python files under the samples folder afterwards. It looks like the generated files haven't been committed in awhile. If you'd like, I can commit these with the PR. Otherwise I'm happy with where things are at for this feature.

wing328:
If no further feedback, I'll merge this PR on the coming weekend.

wing328:
If no further feedback, I'll merge this PR on the coming weekend.

wing328:
@Raznic thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

Raznic:
Thanks @wing328, happy new year to you as well! :smile:

PR #1543, opened 2018-11-26
[csharp] Fix #1531 with non-nullable enums on operation parameters not compiling

### LINKED ISSUE #1531: [csharp] Generated enumerations as query parameters not compiling
##### Description

When specifying query parameters as enumerations, the autogenerated code uses the enumeration type, instead of making it `Nullable`. This C# code fails compile when those query parameters are assigned `null` because C# enumerations are not nullable (they are not objects, as Java enumerations).

The expected behavior is to generate nullable enumerations: `EnumClass? varName`.

This is reproduced with openapi-generator v3.3.3 (current master).

```yaml
openapi: 3.0.0
info:
  description: "This spec is for testing the wrong generation of enumerations in query parameters"
  version: 1.0.0
  title: Enumerations test case
paths:
  /enum_test/:
    get:
      tags:
        - Enums
      summary: Get with enumeration query params.
      description: Get with enumeration query params.
      operationId: getEnumTest
      parameters:
      - name: testEnum
        in: query
        schema:
          $ref: "#/components/schemas/EnumClass"
      responses:
        '200':
components:
  schemas:
    EnumClass:
      type: string
      enum:
      - Value1
      - Value2
      - Value3
```

java -jar openapi-generator-cli-3.3.3.jar generate -g csharp -i test.yml -o test

##### Steps to reproduce

Run the openapi-generator using the `csharp` generator and the attached OpenApi specification. The generated C# code will not compile because enumerations in C# are not nullable.

Enumerations should be generated nullable instead.

rubms:
### PR checklist

C.C. @mandrean (2017/08) @jimschubert (2017/09)

### Description of the PR
Fixed issue #1531 which caused enums used as non-required operation parameters to generate a C# code that does not compile. The reason for the error was that these enums were not generated as `Nullable` but they where being assigned `null` and compared to `null`.

I have fixed the error by:

* Adding a piece of code in the `csharp` `api.mustache` template, that adds a `?` after parameter data types when the parameter is non-required and non-nullable.
* Correctly marking the `isEnum` and `isNullable` properties of operation parameters in the post-process performed in `AbstractCsharpCodegen`.


wing328:
@rubms as discussed, we'll focus on the refactored C# client moving forward and will add better nullable support (which may undo some of the changes in this request)

PR #1552, opened 2018-11-27
[spring] Fixed #1551: Invalid default values for arrays and maps on API method declarations

### LINKED ISSUE #1551: [spring] Invalid default values for arrays and maps on API method declarations
##### Description

When parameters of type array are added to an operation, the generated `Spring` code includes an invalid `defaultValue` in the Spring MVC parameter annotations. For instance, an array of strings query parameter would generate a Spring code like the following:
```java
@Valid @RequestParam(value = "stringArrayParam", required = false, default="new ArrayList<>()") List<String> stringArrayParam
```

That generated `defaultValue` is invalid: the Spring MVC application will throw an error when a request is performed with no value specified for the parameter.

Reproduced with openapi-generator v3.3.3 and previous versions.

```yaml
openapi: 3.0.0
info:
  description: "Spec for testing the default value of array parameters"
  version: 1.0.0
  title: Array parameters test case.
paths:
  /pets:
    get:
      operationId: getPets
      summary: Pets.
      description: "Get pets."
      parameters:
      - name: stringArrayParam
        in: query
        description: "Test array query parameter"
        required: false
        schema:
          type: array
          items:
            type: string
      tags:
      - Pets
      responses:
        200:
          description: Successful request.
components:
  schemas:
    ExampleObject:
      type: object
      properties:
        arrayProperty:
          type: array
          items:
            type: string
```

`java -jar openapi-generator-cli-3.3.3.jar generate -g spring -i test.yml -o test`

Additionally, the configuration property `library` is set to `spring-mvc`:
```
{
	"library": "spring-mvc"
}
```

##### Steps to reproduce

1. Generate a Spring-MVC server application from the previous specification. 
2. Run the application.
3. Throw a GET request to the `/pets` path (`getPets` operation) leaving the `stringArrayParam` query parameter blank or unspecified.

The server will throw the following error: 
```Failed to convert value of type 'java.lang.String' to required type 'java.util.List'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [@io.swagger.annotations.ApiParam @javax.validation.Valid @org.springframework.web.bind.annotation.RequestParam com.rfranco.ics.inputdataapi.autogenerated.dtos.GameType] for value 'new ArrayList<>()'; nested exception is java.lang.IllegalArgumentException: Unexpected value 'new ArrayList<>()'
```

##### Related issues/PRs

This is related to #890, though not the same issue.

The problem is that there are 2 different cases with respect to default values:

* The default value specified by the user, that should be included as default value in the Spring MVC API method annotations.
* The default variable initializer, which may be affected by the default value specified by the user but differs in data-types like arrays or maps.

However the Spring code generator uses the same value (the default variable initializer, in my opinion) for both: initializing variables and specifying Spring MVC annotations.

The Spring MVC code generator must use a different "default value" for both cases.

rubms:
### PR checklist

C.C.: @bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01)

### Description of the PR

Fixed issue #1551 that caused the generation of invalid default values in Spring MVC annotations for array and map parameters.

In Spring there are 2 different cases with respect to default values:

* The default value specified by the user, that should be included as default value in the Spring MVC API method annotations.
* The default variable initializer (e.g. for model properties), which may be affected by the default value specified by the user but differs in data-types like arrays or maps.

However the Spring code generator uses the same value (the default variable initializer, in my opinion) for both: initializing variables and specifying Spring MVC annotations.

This PR includes an overwrite of `fromParameter` for the SpringCodegen class in which the default value for parameters is taken from the one specified by the player. The default value for other types of properties (model properties, for instance) is not modified.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -778,4 +778,26 @@ public void setPerformBeanValidation(boolean performBeanValidation) {
...
+    public String getSpecifiedDefaultValue(Schema p) {
+        if (ModelUtils.isArraySchema(p)) {
+            return null;
+        } else if (ModelUtils.isMapSchema(p)) {
+            return null;
+        }
+
+        return super.toDefaultValue(p);
+    }
+
```
wing328: @rubms thanks for the fix. What about overriding `toDefaultValue` instead?

Ref: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2816

rubms: That's what I first tried, but it breaks the generated code: the default initialization of variables gets broken, causing the generated code not to compile. For instance, overriding `toDefaultValue`, if no default value is specified in the spec for an array it will be initialized to nothing (instead of new ArrayList<>()), generating a non-compiling code:
```java
List<String> listProperty = ;
```

This is why I overrode only the behavior for operation parameters, and not `toDefaultValue`.

wing328: Thanks for the explanation. I think I know what issue you encountered.

I just fix a similar issue in another Java-related generator. Let me file a PR and have you to review my approach before deciding on which one to use.

wing328: I've filed https://github.com/OpenAPITools/openapi-generator/pull/1725. Please review when you've time.

Moving forward, we may need to introduce toParameterDefaultValue and toPropertyDefaultValue for better control of default value in different scenarios.

rubms: Sorry it took so long, I was on holidays. I agree, we need to introduce toParameterDefaultValue and toPropertyDefaultValue  to better control both scenarios. I have reviewed the #1725 PR and it looks good to me!

wing328:
Closed via #1725

wing328:
@rubms thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

PR #1560, opened 2018-11-27
[Python3] Fix #1445: Safe decoding of response content with try-except

### LINKED ISSUE #1445: [Python3] Binary decode fails
##### Description

I'm using openapitools/openapi-generator-cli:v3.3.1 docker image to generate client from my swagger.yml

One of endpoints returns binary data and generated client fails with `UnicodeDecodeError`.

The problem is [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/rest.mustache#L213). I don't need response to be decoded to string as it's binary file.

My workaround is to pass `_preload_content=False` to the api call and handle urllib response object by myself.

I can suggest to simply wrap `.decode("utf-8")` in try-except.

POD666:
### PR checklist

### Description of the PR

https://github.com/OpenAPITools/openapi-generator/issues/1445


wing328:
cc @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10)

tomplus:
I'm not sure if this approach is safe. Maybe we can pass binary data if we detect that raw content is returned (based on returned headers or a model).

Definitely integral tests for it are needed. We have a use case with uploading file but downloading is not tested.

wing328:
What about using `Content-Type` in the HTTP response header to determine whether the data (body) is binary or not?

POD666:
I guess the better option is to keep response data in bytes as long as possible.

Miiike89:
Quick qusetion. Will this Fix be in the upcoming 4.0.0 Release?
In the current beta release it is not included.

spacether:
Does this code path know anything about/use the mime type under the consumes key in the swagger spec? Those mime types can include utf-8 etc definition.

wing328:
Updated the Python Petstore samples. Let's see if the tests pass.

spacether:
@POD666 thank you for this PR.
I am working on cleaning up old python PRs.
Are you still interested in merging this?
Can you resolve the merge conflicts?

PR #1564, opened 2018-11-28
Kotlin collection type

absimas:
### PR checklist

### Description of the PR
Swagger spec file contains only array types thus it is always the preferred choice when converting to kotlin. Allow specifying which collection type to convert to.

@jimschubert @dr4ke616

### DIFF
--- a/docs/generators/kotlin.md
+++ b/docs/generators/kotlin.md
@@ -28,4 +28,9 @@ CONFIG OPTIONS for kotlin
 	        java8 - Java 8 native JSR310
 	        threetenbp - Threetenbp
 
+	collectionType
+	    Option. Collection type to use
+	        array - kotlin.Array
+	        list - kotlin.collections.List
+
 Back to the [generators list](README.md)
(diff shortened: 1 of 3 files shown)


absimas:
I don't quite understand your comment on the default.

Current default is array and this code doesn't change that, thus this is not a breaking change.

absimas:
Oh unless you meant the default for `list` which indeed was `kotlin.collections.List`.

jimschubert:
@absimas yes, sorry for the confusion. For a non-breaking change we would want array and list to remain the same default type until the next major release.

I haven't tested if this PR still works with type mappings (another way users can modify these defaults), so I (or someone else) need to test that before merge.

wing328:
Tested with both `array` and `list`, and the output compiles without issues.

wing328:
@absimas thanks for the PR, which has been included in the v3.3.4 release: https://twitter.com/oas_generator/status/1068772409795207168

PR #1596, opened 2018-12-02
Feature/jaguar generator improvements

jaumard:
Update dart deps
Add default timeout on http requests
Add nullableField option to send or not null field on the payload


wing328:
Thanks for the PR. Please update the PR with a description to explain the change and a better title/subject to describe the PR.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartJaguarClientCodegen.java
```diff
@@ -43,11 +45,15 @@
...
 
+    private boolean nullableFields = true;
+
     public DartJaguarClientCodegen() {
         super();
         browserClient = false;
         outputFolder = "generated-code/dart-jaguar";
         embeddedTemplateDir = templateDir = "dart-jaguar";
+
+        cliOptions.add(new CliOption(NULLABLE_FIELDS, "Is the null fields should be in the JSON payload"));
```
wing328: @jaumard instead of adding another option, what about leveraging the `nullable` attribute (OAS v3) instead?

(in OAS v2, `x-nullable` is used instead)

jaumard: It's not the same ;) my nullableField is for dart generation, it's a global flag for the Json serializer

modules/openapi-generator/src/main/resources/dart-jaguar/class.mustache
```diff
@@ -2,12 +2,12 @@ part '{{classFilename}}.jser.dart';
 
 class {{classname}} {
   {{#vars}}{{#description}} /* {{{description}}} */{{/description}}
-  @Alias('{{{baseName}}}')
+  @Alias('{{baseName}}')
```
wing328: Minor suggestion: using `{{{baseName}}}` to avoid escaping HTML special characters.

jaumard: Sure :) my bad ^^

wing328:
@jaumard thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

PR #1617, opened 2018-12-05
[core] Fix NPE for endpoints without responses

tomekc:
When generating Java client, generator was crashing with cryptic NPE. After debugging, it turned out, that one of the enpoints did not have 'responses' field defined, 
and there was no clue, what operation it was. The code looked like was supposed to deal with that situation, but by simple mistake it moved on and caused NPE.

Additional logging helped isolate the problem.

BTW: IntelliJ IDEA IDE suggested that change, so it is easy to spot it using static code analysis.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -43,15 +43,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


jmini:
Even if not having responses defined is not a valid OpenAPI spec (it will be reported as such by editors like https://editor.swagger.io/) I think that it make sense to support this.

jmini:
> By the way, the validator could issue a warning in such case?

Yes. Good idea!
I am not sure if we have the notion of warnings yet (to be verify), but it could also be use-full for this https://github.com/OpenAPITools/openapi-generator/pull/997

jmini:
@tomekc I have merged your change. Thank you a lot!

For your idea with warnings, can you open a separated issue for that?

wing328:
@tomekc thanks for the contribution, which has been included in the 4.0.0-beta release: https://twitter.com/oas_generator/status/1079727020374806529.

PR #1628, opened 2018-12-06
[C] Support for authentication methods

zhemant:
### PR checklist

### Description of the PR

Adding support for generating code related to authentication methods (basic, oauth and api-key).

Added error printing in case of the server is not available(curl perform failed).

Make apiClient object reusable(url, authentication are not freed and have to be freed manually)  (dataReceived in apiclient will still be freed. As it is not necessay outside the API function).

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
@@ -385,7 +385,13 @@ void apiClient_invoke(apiClient_t    *apiClient,
         if(res == CURLE_OK) {
             curl_easy_getinfo(handle, CURLINFO_RESPONSE_CODE, &apiClient->response_code);
         } else {
-            fprintf(stderr, "curl_easy_perform() failed: %s\n",
+            char *url,*ip,*scheme;
+            long port;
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
```diff
@@ -12,21 +12,29 @@ apiClient_t *apiClient_create() {
...
+
     return apiClient;
 }
 
 void apiClient_free(apiClient_t *apiClient) {
     if(apiClient->dataReceived) {
         free(apiClient->dataReceived);
     }
-    free(apiClient);
+    //free(apiClient);
```
wing328: @zhemant maybe simply remove this line instead of commenting it out?

zhemant: I modified the code in new commits, now in API I free the dataReceived and leave apiClient object as it is for the user to free. So If someone doesn't want to reuse apiClient, instead of calling free function for multiple parameters, only 1 function will do it.

zhemant:
@wing328 Any more modifications are required in this PR?

wing328:
Not from me. Just restarted the CI job. Let's see how it goes.

PR #1664, opened 2018-12-11
[WIP] Add support for multiple inheritance

wing328:
### PR checklist

### Description of the PR

- add support for multiple inheritances
- bug fixes for `allVars`, `vars`
- decommission `supportsInheritance`

Still WIP


cbornet:
Didn't we already have `supportsMixin` for language that support "multiple inheritance" ?

wing328:
@cbornet yup, only JS client generator uses that.

My current approach is to keep the "supportsMixin" and add new code, mustache tags to support mutiple inheritnance for other langauges as part of the code cleanup/refactoring.

cbornet:
I mean, can't we reuse this option for other languages ?

wing328:
Very good question. Ideally we should extend or reuse `supportsMixin` for other languages (in addition to JS) but when I read the following code: 

https://github.com/OpenAPITools/openapi-generator/blob/158760ae7251e8ec34781d1548c8923c803299e2/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1652-L1655

I couldn't understand why only the first model is consider in mixing (multiple inheritance). I don't want to modify it lest it break the JS API client. That's why I introduce a new option instead and add the `allParents` should a language needs multiple inheritance/mixins. We can later migrate JS to use supportsMultipleInheritance to migrate risks in breaking existing usage.

cbornet:
Actually with your modifications `supportsMixins` doesn't have any effect anymore so it could just be removed...

cbornet:
Also I prefer "mixin" over "multiple inheritance" as it better describe what is done. Multiple inheritance is bad and should be avoided at all cost while mixins are ok.

wing328:
> Also I prefer "mixin" over "multiple inheritance" as it better describe what is done. Multiple inheritance is bad and should be avoided at all cost while mixins are ok.

OK. I'll keep the `supportsMixins` name.

cbornet:
:+1:  I wonder what's the effect of the modifications on the JS codegen ? I guess the mixin/multiple-inheritance shall be activated for this one.

cbornet:
Also I don't think we have multiple inheritance in samples

wing328:
@cbornet  yup, I added several spec to test allOf, multiple inheritance, etc (e.g. allOfMultiParent.yaml). We'll add those to the fake petstore spec when more generators are ready for these new features. We'll test with these new spec individually for the time being.

Issue #766, opened 2018-08-07
[Slim] Injection into host or basePath breaks server


ybelenko:
##### Description

Injection into `host` and `basePath` breaks Slim server.

##### openapi-generator version

3.2.1-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
host: petstore.swagger.io */ ' " =end -- \r\n \n \r
basePath: /v2 */ ' " =end -- \r\n \n \r
tags:
  - name: fake
    description: Everything about your Pets */ ' " =end -- \r\n \n \r
    externalDocs:
      description: Find out more */ ' " =end -- \r\n \n \r
      url: 'https://openapis.org'
schemes:
  - http */ ' " =end -- \r\n \n \r
```

##### Command line used for generation

```shell
$ bin/security/php-slim-server-petstore.sh
```

##### Steps to reproduce

```shell
$ bin/security/php-slim-server-petstore.sh
```
then
```shell
$ php -l samples/server/petstore-security-test/php-slim/lib/SlimRouter.php

Parse error: syntax error, unexpected '\' (T_NS_SEPARATOR), expecting ',' or ')' in samples/
server/petstore-security-test/php-slim/lib/SlimRouter.php on line 71

Errors parsing samples/server/petstore-security-test/php-slim/lib/SlimRouter.php
```

##### Suggest a fix/enhancement

Any solution could work:
* Escape quotes in `{{{basePathWithoutHost}}}` and `{{{path}}}`
* Strip out quotes from `{{{basePathWithoutHost}}}` and `{{{path}}}`
* Throw warning when `{{{basePathWithoutHost}}}` and `{{{path}}}` contains invalid chars

ybelenko:
After small investigation, I've found that double quote `"` is escaped in both variables, but single quote `'` isn't escaped.

Of course, I can change my mustache template to use double quotes:
```django
"{{{basePathWithoutHost}}}{{{path}}}", {{classname}}::class . ":{{operationId}}"
```
or even use encoded mustache variables:
```
'{{basePathWithoutHost}}{{path}}', {{classname}}::class . ':{{operationId}}'
```
But it's a half-measure because user can write own template and encounter the same issue over and over again.

I think best solution is to escape single quote within `basePathWithoutHost` and `path` properties somewhere in `AbstractPhpCodegen`.

PR #1687, opened 2018-12-16
[Slim] Encode path to support non-latin characters

ybelenko:
### PR checklist

### Description of the PR

Closes: #766
This PR should be reviewed carefully, because I didn't find a way to fix it without DefaultGenerator modification.

As it turns out `{{basePathWithoutHost}}` construction doesn't solve issue when path contains non-latin characters. For instance, Slim router cannot find related API controller when path looks like `/статьи/авторы`, because escaped mustache variable output is `/статьи/авторы` without any difference, while HTTP clients and web browsers encodes non-latin characters by default.  When I use this link in Chrome I see in server logs that url has been automatically encoded into `/%D1%81%D1%82%D0%B0%D1%82%D1%8C%D0%B8/%D0%B0%D0%B2%D1%82%D0%BE%D1%80%D1%8B`. The same behaviour with Insomnia app. With new method `encodePath` Slim router works as expected.

I've also added few tests to check `encodePath` method output of `PhpSlimServerCodegen`.

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @renepardon

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -537,9 +537,9 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
                     }
                 });
                 Map<String, Object> operation = processOperations(config, tag, ops, allModels);
-
+                URL url = URLPathUtils.getServerURL(openAPI);
                 operation.put("basePath", basePath);
-                operation.put("basePathWithoutHost", basePathWithoutHost);
+                operation.put("basePathWithoutHost", config.encodePath(url.getPath()).replaceAll("/$", ""));
                 operation.put("contextPath", contextPath);
                 operation.put("baseName", tag);
                 operation.put("apiPackage", config.apiPackage());
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConfig.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConfig.java
@@ -81,6 +81,8 @@ public interface CodegenConfig {
 
     String escapeTextWhileAllowingNewLines(String text);
 
+    String encodePath(String text);
+
     String escapeUnsafeCharacters(String input);
 
     String escapeReservedWord(String name);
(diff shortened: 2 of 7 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -533,9 +533,9 @@ public int compare(CodegenOperation one, CodegenOperation another) {
                     }
                 });
                 Map<String, Object> operation = processOperations(config, tag, ops, allModels);
-
+                URL url = URLPathUtils.getServerURL(openAPI);
                 operation.put("basePath", basePath);
-                operation.put("basePathWithoutHost", basePathWithoutHost);
+                operation.put("basePathWithoutHost", config.encodePath(url.getPath()).replaceAll("/$", ""));
```
wing328: @ybelenko why do we need to replace `/$` with empty string?

wing328: Thanks for the explanation

ybelenko: @wing328 Maybe I can change `basePathWithoutHost` somewhere else? In some post process functions in Slim generator? Or maybe we should add `encodedBasepathWithoutHost` property? I think that this problem with non-latin characters exists in other generators as well.

wing328: >  I think that this problem with non-latin characters exists in other generators as well.

I think so but I didn't have time to test and confirm.

wing328: My take is to start with your fix for Slim generator and later we port this fix to the default codegen or other generators if users report similar issues.

ybelenko: Yeah, I agree, it's not so important now. 

Important part is `basePathWithoutHost` value. Whether we should override it or put it in separated prop. Do you know how to override it inside `PhpSlimServerCodegen` codegen class?

What `contextPath` property stands for?

PR #1700, opened 2018-12-18
fix: force to decode as utf-8 when header contains application/json to avoid text garbling.

cimadai:
### PR checklist

  - dart is not exists.
  - @ircecho, @swipesight, @jaumard

### Description of the PR

The original processing is using `response.body` to deserialize as json in `api.mustache`.
However, this is decoded by latin1 if the header contains only `application/json` instead of `application/json; charset=utf-8`.

Because of this behavior, if the response body is encoded UTF-8 but the headers doesn't contain charset, the body will garbling.

cf: https://github.com/dart-lang/http/issues/175

Since playframework 2.6 returns `Content-Type: application/json` without `charset=utf-8`, I changed this parsing algolithm.

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart/api_helper.mustache
+++ b/modules/openapi-generator/src/main/resources/dart/api_helper.mustache
@@ -50,3 +50,15 @@ String parameterToString(dynamic value) {
     return value.toString();
   }
 }
+
+/// Returns the decoded body by utf-8 if application/json with the given headers.
+/// Else, returns the decoded body by default algorithm of dart:http.
+/// Because avoid to text garbling when header only contains "application/json" without "; charset=utf-8".
+String _decodeBodyBytes(Response response) {
+  var contentType = response.headers['content-type'];
+  if (contentType != null && contentType.contains("application/json")) {
+    return utf8.decode(response.bodyBytes);
+  } else {
+    return response.body;
+  }
+}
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/resources/dart/api_helper.mustache
```diff
@@ -50,3 +50,15 @@ String parameterToString(dynamic value) {
     return value.toString();
   }
 }
+
+/// Returns the decoded body by utf-8 if application/json with the given headers.
+/// Else, returns the decoded body by default algorithm of dart:http.
+/// Because avoid to text garbling when header only contains "application/json" without "; charset=utf-8".
+String _decodeBodyBytes(Response response) {
+  var contentType = response.headers['content-type'];
+  if (contentType != null && contentType.contains("application/json")) {
```
wing328: @cimadai later we may want to further improve it to detect other JSON mime type, eg. https://github.com/OpenAPITools/openapi-generator/blob/1db105b6f9dc3058b6b8c17183d26719857e371a/samples/client/petstore/csharp-refactor/OpenAPIClient/src/Org.OpenAPITools/Client/ClientUtils.cs#L265 in C# to detect the JSON mime type using a regular expression

wing328:
LGTM. Thanks for the fix 👍

PR #1711, opened 2018-12-19
Various fixes, enhancements to `csharp-refactor` client

wing328:
### PR checklist

### Description of the PR

- update dependencies to the latest version
- bug fixes to content-type, accept header selection
- add test cases

NOTE: file upload is still broken and will be addressed in another PR instead.

cc @mandrean (2017/08) @jimschubert (2017/09)


samples/client/petstore/csharp-refactor/OpenAPIClient/src/Org.OpenAPITools/Client/ApiClient.cs
```diff
@@ -36,8 +41,15 @@ internal class CustomJsonCodec : RestSharp.Serializers.ISerializer, RestSharp.De
         private string _contentType = "application/json";
         private readonly JsonSerializerSettings _serializerSettings = new JsonSerializerSettings
         {
-            // Swagger generated types generally hide default constructors.
-            ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor
+            // OpenAPI generated types generally hide default constructors.
+            ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor,
+                ContractResolver = new DefaultContractResolver
+                {
+                    NamingStrategy = new CamelCaseNamingStrategy()
```
jimschubert: We'll want to create an issue to support naming strategy here.

samples/client/petstore/csharp-refactor/OpenAPIClient/src/Org.OpenAPITools/Client/ClientUtils.cs
```diff
@@ -192,6 +192,61 @@ public static byte[] ReadAsBytes(Stream inputStream)
...
+        ///    application/json; charset=UTF8
+        ///    APPLICATION/JSON
+        ///    application/vnd.company+json
+        /// </summary>
+        /// <param name="mime">MIME</param>
+        /// <returns>Returns True if MIME type is json.</returns>
+        public static bool IsJsonMime(String mime)
+        {
+            var jsonRegex = new Regex("(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$");
+            return mime != null && (jsonRegex.IsMatch(mime) || mime.Equals("application/json-patch+json"));
```
jimschubert: This could be optimized by returning earlier using `String.IsNullOrWhitespace(mime)` before unnecessarily constructing the regex object… could also cache the regex object as a static variable to reduce its effect on heap:

```
public static Regex jsonRegex = new Regex("(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$");

public static bool IsJsonMime(String mime)
{
    if (String.IsNullOrWhiteSpace(mime)) return false;
    return jsonRegex.IsMatch(mime) || mime.Equals("application/json-patch+json");
}
```

samples/client/petstore/csharp-refactor/OpenAPIClient/src/Org.OpenAPITools/Client/ClientUtils.cs
```diff
@@ -192,6 +192,61 @@ public static byte[] ReadAsBytes(Stream inputStream)
...
+        /// <param name="contentTypes">The Content-Type array to select from.</param>
+        /// <returns>The Content-Type header to use.</returns>
+        public static String SelectHeaderContentType(String[] contentTypes)
+        {
+            if (contentTypes.Length == 0)
+                return "application/json";
+
+            foreach (var contentType in contentTypes)
+            {
+                if (IsJsonMime(contentType.ToLower()))
```
jimschubert: I think this might be incorrect… this would return true for `APPLICATION/JSON; CHARSET=UTF-8`, then apply that as a header when it's not going to be a useful or relevant header. Although HTTP spec defines header names (e.g. 'Content-Type') as case insensitive, the values are case sensitive.

wing328:
@jimschubert thanks for the feedback. As discussed we'll address that in follow-up PRs

jimschubert:
@wing328 I opened #1723. I didn't see a clean solution to the serialization strategy, so I'll revisit that one later.

PR #1712, opened 2018-12-19
[lua] Allow connection on port != 80 or 443

gfeun:
### PR checklist

### Description of the PR

In the current implementation a client connection port is determined by the connection scheme (http or https) and is therefore port 80 or 443
This commit allows to specify a host like "localhost:8080" when creating the client
The specified port is then used when initiating http(s) requests


gfeun:
@daurnimator

modules/openapi-generator/src/main/resources/lua/api.mustache
```diff
@@ -24,8 +24,10 @@ local function new_{{classname}}(host, basePath, schemes)
 		schemes_map[v] = v
 	end
 	local default_scheme = schemes_map.https or schemes_map.http
+	local port = http_util.split_authority(host, default_scheme)
```
daurnimator: Is the 'host' that is passed into this function expected to be an authority? If so the argument should be renamed

Also, `http_util.split_authority` returns the host *and* the port, the usage here is incorrect.

gfeun:
Thanks for your review.
Fixed usage by assigning host return value.

The host is not an authority.
I used split_authority because of this comment in lua-http source code : `-- Splits a :authority header (same as Host) into host and port`
I am not aware and can't find a "split_host" function in the library.

I have taken a look at other language generator templates and it seems there is no consensus around parameter usage for http client configuration. 
For the go language, port can be specified in the "BasePath" parameter along scheme and host : 
`BasePath: "http://petstore.swagger.io:80/v2`

In python only a "host" field can be specified and the url before request is built with : 
`url = self.configuration.host + resource_path`
So the port has to be appended to the host parameter.

For both languages, configuration is passed as a dict to the new_api : 
```python
# Python
config = petstore_api.Configuration()
config.host = 'http://localhost/'
client = petstore_api.ApiClient(config)
```

```golang
// Go
cfg := sw.NewConfiguration()
cfg.AddDefaultHeader("testheader", "testvalue")
cfg.Host = "testhost"
client = sw.NewAPIClient(cfg)
```

So I am not sure what to do. I think the simplest and backward compatible change is the one i made, to treat the host variable like "host:port", split it and and pass host and port down to the client request which will give us this usage: 
```lua
pet_api = petstore.new_pet_api("localhost:8443", "/", {"https"})
```

daurnimator:
> The host is not an authority.

Yes it is? An authority (according to http2 spec) here is a host + a port; which is what you have.

gfeun:
Ok, I didn't know Authority header supersedes the Host header in HTTP2.

gfeun:
Can this be merged @daurnimator ?

PR #1729, opened 2018-12-21
Add an option to generate the alias (map, array) as model

wing328:
### PR checklist

### Description of the PR


wing328:
> The CodegenOperation has returnBaseType = 'Map' but returnType = 'List<A>', which won't compile because class A is no longer generated. I didn't see any code below to rectify this issue.

If I understand correctly, this issue already exists in the current master so please open an issue to track it instead.

The goal of this PR is to restore the old behaviour to generate alias (map, array) as model. and I believe your use case is to set `generateAliasAsModel` to true. If the issue mentioned above still occurs, please kindly mention in the new issue created.

demonfiddler:
Fair enough - the problem with the current behaviour can be seen as a separate issue, which I'll log in due course. In the meantime I'd be happy for the PR to proceed. Many thanks for the rapid turnaround William.

demonfiddler:
Looks like there's a problem. I just rebased onto upstream/option_generate_alias_model and now I'm getting an NPE in my code because CodegenModel.additionalPropertiesType is no longer being set (to AttributeValue) for the alias type (AuditEvent extends Map<String, AttributeValue>).

demonfiddler:
Thanks. With the suggested addAdditionPropertiesToCodeGenModel() override, generating from the CLI with the --generate-alias-as-model option now seems to work, and my AuditEvent (and other alias classes) are generated and referenced correctly in the API's CXF implementation class. My only question is: why is it necessary for the generator sub-class to do this in an override? For backwards compatibility shouldn't it be done automatically by DefaultCodegen? 

**However**, the \<generateAliasAsModel>true\</generateAliasAsModel> Maven plug-in configuration option has no effect. I tried it in both execution/configuration and execution/configuration/configOptions - under the debugger I can see that additionalProperties does not contain the key "generateAliasAsModel" at DefaultCodegen line 213 (```if (additionalProperties.containsKey(CodegenConstants.GENERATE_ALIAS_AS_MODEL)) { etc...```).

Issue #1748, opened 2018-12-23
[REQ] v2 swagger spec, some defaults unset in models


spacether:
### Is your feature request related to a problem? Please describe.

When defining default values in the swagger spec, they are only used as defaults in the python client models if the type of the variable is a string or boolean.

## Describe the solution you'd like

Please include those default arguments in the python models.
Please also include type checking validation of the input so if the user assigns an incorrectly typed value to a field, the model will throw an error.
Note: I will work on this feature.

## Describe alternatives you've considered

Manually assign values all the time. This is possible but not preferred. If we define a default in the spec, it should be used.

## Additional context

This appears to be caused by java code from the swagger-api organization. I have opened issues with them at:
swagger-api/swagger-core#3065
swagger-api/swagger-parser#971

I will work on adding this feature prior to working on: https://github.com/OpenAPITools/openapi-generator/issues/1715

spacether:
This appears to be caused by java code from the swagger-api organization. I have opened an issue with them at:
https://github.com/swagger-api/swagger-core/issues/3065

PR #1776, opened 2018-12-29
[python-client] Add model default values

spacether:
### PR checklist

- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/python-petstore.sh` and `./bin/security/python-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
  - @taxpon  
  - @frol  
  - @mbohlool  
  - @cbornet  
  - @kenjones-cisco  
  - @tomplus   
  - @wing328 

### Description of the PR

This PR lets python client users set model defaults on integers, numbers, and arrays when they are not currently able to in a swagger2.0 spec. The new code loads model variable examples into the model variable default value if the default is not already set and an example exists for the variable.
Fixes this issue: https://github.com/OpenAPITools/openapi-generator/issues/1748

Reasons to do this:
The current swagger-parser is not including default values that are float, integer, or array when processing a swagger2.0 spec. So in a swagger 2.0 spec, there is no way to set defaults on:
- integers
- numbers
- arrays

This PR adds in fallback code which uses and assigns a spec example value as the default value if one exists in the spec.
So it uses examples to set defaults on the types
- string
- boolean
- integers
- numbers
- arrays

### Tests
- test_type_holder_example.py added to verify that examples are loaded into defaults
- test_type_holder_default.py added to verify what default-only definition does and does not work
- samples updated to pass CircleCi tests (173 files)


spacether:
Do I need to make updates so the appveyor + cicleci tests will pass?
Appveyer
Did this test fail because it was unable to connect to a remote server and download .nupkg files?
That failure appears to be unrelated to my fix. 
Can someone kick off this test again?
I don't see a way I can re-run it.

CircleCI
It looks like these tests failed because the clients in non-python languages were updated because I updated the swagger2.0 main example file.
Possible solution: separate my models into a new spec, and add python only-client generation in ./bin/python-petstore.sh which will use a new separate yaml file which includes my new models.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -624,6 +624,14 @@ public String toDefaultValue(Schema p) {
                 else
                     return "True";
             }
+            // include fallback to example, default defined as server only
+            // example is not defined as server only
+            if (p.getExample() != null) {
+                if (p.getExample().toString().equalsIgnoreCase("false"))
```
wing328: What about using `Boolean.valueOf` instead of using `equalsIgnoreCase`?

spacether: Boolean.valueOf(string) will always return True or False, 
so Boolean.valueOf("5") would return False. 
When I pass in "5" as a boolean I should return null. Code updated use equalsIgnoreCase to check for "true" and "false" otherwise return null.

wing328: > When I pass in "5" as a boolean the code should return null

If someone puts down 5 as the example value and assume it will be treated false in boolean, it's the user's issue (not something we need to handle) and I don't think it should return null (clearly there's a value defined)

We've seen spec using 1 or 0 to as example value for boolean type (not entirely correct as they should be using "true" or "false" instead) but still we want to able to handle these cases correctly as some languages (e..g Perl) treat 1 as true.

spacether: Coercing it to a boolean gives the user back some boolean which has worth. I will change it to Boolean.valueOf. This behavior may surprise python users though because they will be using a python client with java casting of string truthiness.
"1" -> True in Python, false in Java
"0" -> True in Python, false in Java

wing328:
@spacether the change looks good to me. Thanks for your contribution.

spacether:
Good to know. Thank you you @wing328 sorry for the breaking change.

wing328:
@spacether no worries. In the templates, you can do something similar to the following to have the default value fall back to example value (not something I recommend but ok if there's a valid use case)
```
{{{defaultValue}}}{{^defaultValue}}{{{example}}}{{/defaultValue}}
```

wing328:
FYI. Filed https://github.com/OpenAPITools/openapi-generator/pull/2213

PR #1789, opened 2019-01-01
[gradle-plugin] Array model change

jimschubert:
### PR checklist

### Description of the PR

We've changed how top-level models defined as an array work. The change was not recent.

I found that CI wasn't running the checks on uploadArchive (so the failing test wasn't caught). This PR fixes the failing test and ensures we run `check` whenever we run `uploadArchives` (if tests fail, archive won't be uploaded).


wing328:
> We've changed how top-level models defined as an array work. The change was not recent.

You can restore the behavior using `--generate-alias-as-model` (default: false) (https://github.com/OpenAPITools/openapi-generator/pull/1729)

> I found that CI wasn't running the checks on uploadArchive (so the failing test wasn't caught). This PR fixes the failing test and ensures we run check whenever we run uploadArchives (if tests fail, archive won't be uploaded).

Shall we run the check everytime the CI runs instead of running via `uploadArchives` only?

jimschubert:
@wing328

> Shall we run the check everytime the CI runs instead of running via uploadArchives only?

I don't understand the question. In this PR, the tests have been added as a dependency to the `uploadArchives` task so tests will run when that task is called. If we were to add this elsewhere in a major/upcoming branch, it would be redundant.

Were you asking about explicitly adding it somewhere else?

wing328:
Sorry that I misunderstood. I thought the task will only be run for stable releases but it will be called when deploying snapshot version as well.

PR #1793, opened 2019-01-02
[Dart][Jaguar] and proto serialization support 

jaumard:
Here is my current work to allow the generator of jaguar dart to support either json or proto serialization. 
I've added a `serializarion` param to choose json or proto. 
I proto models should be generated as proto files and not dart files. For now I'm blocked because I'm trying to do this: 
```
message {{classname}} {
    {{#vars}}
    {{{datatype}}} {{name}} = {{@index}};
    {{/vars}}
}
```
Apparently the {{@index}} is not supported :/ and return empty. But in proto definition you need to specify the index of each field like: 
```
message Person {
  string name = 1;
  int32 id = 2;
  string email = 3;

}
```


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartJaguarClientCodegen.java
```diff
@@ -152,6 +194,19 @@ public void processOpts() {
                     modelImports.add(underscore(modelImport));
                 }
             }
+
+            int index = 1;
+            for (CodegenProperty p : cm.vars) {
+                p.vendorExtensions.put("x-index", index);
+                String protoType = protoTypeMapping.get(p.baseType);
+                if(p.isListContainer) {
+                    String innerType = protoTypeMapping.get(p.mostInnerItems.baseType);
```
jaumard: this is not working yet, as here we have the dart type and not the openapi type :(

jaumard: This is now ok !

jaumard:
It will need a real test case to really know if it works but this should be quite good. !

jaumard:
@wing328 the WIP label doesn't go away automatically ^^

jaumard:
@wing328 can we merge this ? :) don't want to keep rebase in it forever ^^

wing328:
@jaumard sure. Can you resolve the merge conflicts? Let me know if you need any help.

PR #1794, opened 2019-01-02
set isEnum when model is enum

### LINKED ISSUE #1780: [REQ][General] Set boolean flags in case of enum in non-primitive type

In case when enum class is present, places, where this variable is used, is defined as a nonprimitive type and remaining flags are set to false. In low-level language like C, enum is needed to be handed in a special way, not like the nonprimitive type used normally for other classes.

Setting isEnum flag in case the class/model is enum will help in this. We already have isModel to handle nonprimitive type which is class. Setting this flag will give more control over using and defining the variables

zhemant:
### PR checklist

### Description of the PR

fix #1780 , Along with setting isModel to true, set isEnum to true when a model is generated for enum. This will allow better handling for enum variables when enum is of nonPrimitiveType.
 
@wing328 @OpenAPITools/generator-core-team

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2181,12 +2181,12 @@ public class DefaultCodegen implements CodegenConfig {
 
             // set is enum to true when a model is enum and type is non-primitive.
             if (StringUtils.isNotBlank(p.get$ref()) && globalSchemas != null) {
-               Schema ref = globalSchemas.get(ModelUtils.getSimpleRef(p.get$ref()));
-               if (ref != null) {
-                   if (ref.getEnum() != null) {
-                       property.isEnum = true;
-                   }
-               }
+                Schema ref = globalSchemas.get(ModelUtils.getSimpleRef(p.get$ref()));
+                if (ref != null) {
+                    if (ref.getEnum() != null) {
+                        property.isEnum = true;
+                    }
+                }
             }
 
             setNonArrayMapProperty(property, type);
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2180,7 +2180,7 @@ public class DefaultCodegen implements CodegenConfig {
             // --END of revision
 
             // set is enum to true when a model is enum and type is non-primitive.
-            if (!StringUtils.isBlank(p.get$ref())) {
+            if (StringUtils.isNotBlank(p.get$ref()) && globalSchemas != null) {
                Schema ref = globalSchemas.get(ModelUtils.getSimpleRef(p.get$ref()));
                if (ref != null) {
                    if (ref.getEnum() != null) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2178,6 +2178,17 @@ public class DefaultCodegen implements CodegenConfig {
             //    property.baseType = getSimpleRef(p.get$ref());
             //}
             // --END of revision
+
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2177,6 +2177,17 @@ public CodegenProperty fromProperty(String name, Schema p) {
             //    property.baseType = getSimpleRef(p.get$ref());
             //}
             // --END of revision
+
+            // set is enum to true when a model is enum and type is non-primitive.
+            if (StringUtils.isNotBlank(p.get$ref()) && globalSchemas != null) {
+               Schema ref = globalSchemas.get(ModelUtils.getSimpleRef(p.get$ref()));
+               if (ref != null) {
+                   if (ref.getEnum() != null) {
```
wing328: @zhemant What about the following instead?
```
if (ref != null && ref.getEnum() != null) {
...
}
```

zhemant: Yes, this is better. I will make these changes and push again.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2177,6 +2177,15 @@ public CodegenProperty fromProperty(String name, Schema p) {
             //    property.baseType = getSimpleRef(p.get$ref());
             //}
             // --END of revision
+
+            // set is enum to true when a model is enum and type is non-primitive.
+            if (StringUtils.isNotBlank(p.get$ref()) && globalSchemas != null) {
+               Schema ref = globalSchemas.get(ModelUtils.getSimpleRef(p.get$ref()));
```
wing328: What about using `ModelUtils.unaliasSchema(globalSchemas, p)` instead?

wing328:
Looks like this will cause a lot of changes to other generators (e.g. C#) as shown in the CircleCI logs.

We need to consider other ways to address the problem you're facing.

zhemant:
Can we override the fromProperty method in CLibCurlClient.java and there we set isEnum tag? I have an idea of it, I am already sending openAPI for projectName so may be I can use it to set a globalvariable there and use it during fromProperty

wing328:
When you've time in the coming week, please ping me via Gitter and we can discuss other solutions in details.

wing328:
Closing this one for the time being.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jmini: I have analyzed the same code: `ModelUtils.unaliasSchema(globalSchemas, p)` is returning the reference-schema to indicate that a outer-enum must be created (which is correct). 

You could use `ModelUtils.getReferencedSchema(this.openAPI, p)` to get the referenced schema regardless of any other consideration.

jmini:
Please have a look at PR #2001 I have filed to fix the default value case where an outer-enum is created.

```yaml
components:
  schemas:
    TestObject:
      type: object
      properties:
        myEnum:
          $ref: '#/components/schemas/MyEnum'
    MyEnum:
      type: string
      default: FIRSTVALUE
      enum:
      - FIRSTVALUE
      - SECONDVALUE
```

For the java generators `property.isEnum` for `myEnum` can not be set to `true`, because this is the trigger to indicate that an inner enum needs to be generated. I think it is the case for other languages.

My PR sets `allowableValues` in case of outer enum (enum created as separated type).

I did not read the details behind your problem, but I have the felling that the topics are connected.

Feel free to continue the discussion.

zhemant:
I had a look at PR #2001, as I haven't dealt with the default value in an enum, but sure it will be helpful in case default value needs to be set. 

I want to differentiate between enum and non-primitive datatype but we only have isModel flag set to true rest all flags are false. so it was creating an issue. hence I thought of setting isEnum flag but didn't know it would interfere with other generators. 

By looking at your code, I got a few hints on how I can handle it in CLibCurl.java so it doesn't interfere with other generators.

PR #1821, opened 2019-01-05
Deprecated initialCaps

wing328:
### PR checklist

### Description of the PR

Replacing it with `camelize` and will completely remove it in 4.x release

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1505,14 +1505,15 @@ public class DefaultCodegen implements CodegenConfig {
     }
 
     /**
-     * Capitalize the string
+     * Capitalize the string. Please use org.openapitools.codegen.utils.StringUtils.camelize instead as this method will be deprecated.
      *
      * @param name string to be capitalized
      * @return capitalized string
+     * @deprecated
      */
     @SuppressWarnings("static-method")
     public String initialCaps(String name) {
-        return StringUtils.capitalize(name);
+        return camelize(name);
     }
 
     /**


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1505,14 +1505,15 @@ public String snakeCase(String name) {
...
+     * Capitalize the string. Please use org.openapitools.codegen.utils.StringUtils.camelize instead as this method will be deprecated.
      *
      * @param name string to be capitalized
      * @return capitalized string
+     * @deprecated
      */
     @SuppressWarnings("static-method")
     public String initialCaps(String name) {
-        return StringUtils.capitalize(name);
+        return camelize(name);
```
osharav: camelize removes underscores which may cause issues with the InlineModelResolver and other places where underscores are being introduced. in the cpp rest sdk client generator this causes the class names to be different than the file names (some method calls have used initialCaps and some do not) which causes import errors.
I will open a dedicated issue for this.
Regards,
Omer.

osharav:
#2003 cloud be related

wing328:
@osharav thanks for reviewing the change. I could be wrong but I don't think #2002 is related as the user has confirmed 3.4.4 (which does not include this change) also has the issue.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
wing328: Yes, please open a separate issue with a (minimal) spec to reproduce the issue.

osharav:
True, but I have another example with a generated inline response which shows the behavior - I'll attach it to the issue I need to open.

Issue #1812, opened 2019-01-04
[REQ] Python client, use given values for required enums of length 1


spacether:
### Is your feature request related to a problem? Please describe.

If a user describes a required parameter with an enum of length one, they are defining what the parameter's value should be. Please load in that value into any endpoint or models that use it.

## Describe the solution you'd like

Please set the default values for those parameters in the model initialization and the endpoint method, so they can be omitted when instantiating models or making endpoint calls.

## Describe alternatives you've considered

One could manually set things like constant headers every time or assigning these required values in al instances of a model but the above solution is easier on our users.

spacether:
Closed because it was added in python-experimental

PR #1824, opened 2019-01-05
Python client, endpoint hits use parameter defaults

spacether:
### PR checklist

@taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10)

### Description of the PR

This makes it so endpoint hits use the defaults defined in the spec for header/query/form.. parameters.
If one explicitly passes in the parameters, they will overwrite the default values.
One test added in samples/client/petstore/python/test/test_fake_api.py to verify that defaults are used and can be overwritten

This PR will close out this issue: https://github.com/OpenAPITools/openapi-generator/issues/1812


wing328:
@spacether thanks for the PR. For other clients, we usually don't include the default value in the parameter as the server should take care of the default value if it's not provided in the client's request.

spacether:
> @spacether thanks for the PR. For other clients, we usually don't include the default value in the parameter as the server should take care of the default value if it's not provided in the client's request.

@wing328 I understand. The code path already included getDefault to set the parameter default value. After thinking about this more, I determined that what I want to do is load any required constants into the parameter default value.

So now the code does the following when setting a parameter default:
- look for enum of length 1, if present use it as the default
- if there is no enum and there is a default already set (server side default) use it

These defaults are then used as the defaults in the route method and the model initialization method.
I will need to make more updates to get tests to pass.

spacether:
Finishing this is taking a bit more time than I expected. I will re-open it when I am done.

Issue #1848, opened 2019-01-08
[BUG] [Typescript] [Axios] invalid syntax produced when using headers


sruehl:
##### Description

using headers produces invalid code.
`localVarHeaderParameter['{{baseName}}'] = {{paramName}}.join(COLLECTION_FORMATS["{{collectionFormat}}"]));` should be `localVarHeaderParameter['{{baseName}}'] = {{paramName}}.join(COLLECTION_FORMATS["{{collectionFormat}}"]);`
look for the last closing right parenthesis ')'

[typescript-axios/api.mustache#L175](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-axios/api.mustache#L175)

[typescript-axios/api.mustache#L194](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-axios/api.mustache#L194)

##### openapi-generator version

master HEAD

##### OpenAPI declaration file content or url

https://gist.github.com/sruehl/033bb26fc3de97533210d4aab269a55b

##### Command line used for generation

none

##### Steps to reproduce

use a header of type array

##### Related issues/PRs

#1864

##### Suggest a fix

remove the redundant parenthesis ')'

sruehl:
#1864 was merged

PR #1864, opened 2019-01-09
[BUG-1848] Remove redundant right parenthesis

sruehl:
### PR checklist

### Description of the PR

the api.mustache for typescript-axios contains a syntax error when using arrays in header.

#1848


modules/openapi-generator/src/main/resources/typescript-axios/api.mustache
```diff
@@ -191,7 +191,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
                 })
             {{/isCollectionFormatMulti}}
             {{^isCollectionFormatMulti}}
-                    localVarFormParams.set('{{baseName}}', {{paramName}}.join(COLLECTION_FORMATS["{{collectionFormat}}"]));
+                    localVarFormParams.set('{{baseName}}', {{paramName}}.join(COLLECTION_FORMATS["{{collectionFormat}}"]);
```
macjohnny: I think this should be reverted, now one closing bracket `)` is missing

sruehl: oh right thanks

sruehl: pushed via --force thanks for the hint @macjohnny

sruehl:
@macjohnny I left out samples/changes on petstore.yaml intentionally and linked an example in #1848 as I don't wanted to adjust the file for every protocol. Maybe a file with all possibilities would be good for the future so we have a validation of all features in all languages. Another unrelated bug here is that we can define complex types in headers and the won't get stringifyed (But thats a completely different issue).

Issue #1832, opened 2019-01-07
[CI] Python Tests Failing in TravisCI


spacether:
##### Description

CITesting in travis appears to be broken due to some kind of versioning error.
Per here: https://travis-ci.org/OpenAPITools/openapi-generator/builds
Master branch tests are failing

##### openapi-generator version

master

##### OpenAPI declaration file content or url

N/A

##### Command line used for generation

N/A

##### Steps to reproduce

Re-run the master branch TravisCI tests

##### Related issues/PRs

N/A

##### Suggest a fix

When TravisCI is testing the python3 python-asyncio client We see the error:
```
self = <pytest_cov.plugin.CovPlugin object at 0x2ad617571a20>
item = <TestCaseFunction testAdditionalPropertiesClass>

    @compat.hookwrapper
    def pytest_runtest_call(self, item):
>       if (item.get_marker('no_cover')
                or 'no_cover' in getattr(item, 'fixturenames', ())):
E               AttributeError: 'TestCaseFunction' object has no attribute 'get_marker'

.tox/py3/lib/python3.6/site-packages/pytest_cov/plugin.py:289: AttributeError
```
This is inside the pytest-cov module.
Looking at the versions, we see that the below versions are installed on the test system:
pytest-cov==2.6.0
pytest==4.1.0

Per the pytest change log, https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst, pytest4.1.0 removes get_marker and replaces it with get_closest_marker 

This issue was fixed in pytest-cov 6 hours ago per:
https://github.com/pytest-dev/pytest-cov/commit/04f02b559873e1c122fbc669df73824f3419d046#diff-3b81811be824969e5c6bcd3ea919e346
Per the pytest-cov change log here:
https://github.com/pytest-dev/pytest-cov/blob/master/CHANGELOG.rst

We need pytest-cov 2.6.1 to be installed in our TravisCI test environment
or we need to peg pytest and pytest-cov to earlier versions.

spacether:
Per the latest TracisCI builds, TravisCI is now downloading pytest==4.1.0 and pytest-cov==2.6.1 so this error has gone away. However, using an earlier version of pytest-cov (which our requirements.txt allows) would still result in this failure.
Suggest pinning to newer versions of pytest and pytest-cov to prevent this error in the future. These versions are:
pytest>= 3.6.0
pytest-cov>=2.6.1

spacether:
Issue was resolved

Issue #1807, opened 2019-01-03
[BUG] swagger 2.0 spec is missing an xml example with all xml fields


spacether:
##### Description

The example swagger 2.0 spec is here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
It includes some xml examples, but they only include the name and wrapped xml fields
Please add an example which includes
- name
- attribute
- wrapped
- prefix

fields so it can be used for xml server + client testing.

##### openapi-generator version

master branch

##### Command line used for generation

N/A

##### Steps to reproduce

N/A

##### Suggest a fix

Add an example model and a route that uses it.

spacether:
Closed because item was added to our sample spec

Issue #13, opened 2018-05-13
[java rest-assured] default is a reserved keyword


jmini:
##### Description

When the java generator (with restassured library) creates the project without any tags defined, a `DefaultApi` class is created.

In `ApiClient` the method to create this API is:
```
    public DefaultApi default() {
        return DefaultApi.default(config.baseReqSpec.get());
    }
```
This does not compile, because "default" is a reserved keyword in java.

The method (in this case only) should be named `defaultApi()`.

##### openapi-generator version

3.0.0-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.1
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/'
paths:
  /ping:
    get:
      operationId: pingGet
      responses:
        '201':
          description: OK
```

##### Work around

Set a tag on the operations in order to change the `default` value to something else.

jmini:
An other example:
```yaml
openapi: 3.0.1
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/'
paths:
  /return/ping:
    get:
      tags:
        - return
      operationId: pingGet
      responses:
        '201':
          description: OK
components:
  schemas: {}
```

Produces also compile error:

```
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[43,21] <identifier> expected
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[43,28] illegal start of type
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[43,29] <identifier> expected
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[43,30] ';' expected
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[44,9] illegal start of type
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[44,25] ';' expected
[ERROR] testIssue13-oapigen-rest-assured-java/src/main/java/org/openapitools/client/ApiClient.java:[44,55] <identifier> expected
[...]

PR #1883, opened 2019-01-11
[Spec] adds XmlItem model and route for xml testing

spacether:
### PR checklist

@wing328
@jimschubert
@cbornet
@jaz-ah
@ackintosh
@JFCote
@jmini

### Description of the PR

1. This PR adds the model XmlItem and a route that uses it to the swagger 2.0 sample spec.
  It also includes 6 xml mime types that are produced + consumed in that route.
  Downstream clients can use this model and that route for their xml tests.
2. It adds an Api suffix on the java rest-assured client to pass CircleCi tests
3. It pegs the versions of pytest and pytest-cov to fix and pass TravisCI tests

The examples currently in the spec do not include all xml attributes. The XmlItem model includes all xml attributes:
- name
- attribute
- wrapped
- prefix

### Issue Closure
When merged this PR will close these issues:
- [BUG] swagger 2.0 spec is missing an xml example with all xml fields https://github.com/OpenAPITools/openapi-generator/issues/1807
- [java rest-assured] default is a reserved keyword https://github.com/OpenAPITools/openapi-generator/issues/13
- [CI] Python Tests Failing in TravisCI https://github.com/OpenAPITools/openapi-generator/issues/1832


modules/openapi-generator/src/main/resources/Java/libraries/rest-assured/ApiClient.mustache
```diff
@@ -33,8 +33,8 @@ public class ApiClient {
 
      {{#apiInfo}}
      {{#apis}}
-    public {{classname}} {{classVarName}}() {
-        return {{classname}}.{{classVarName}}(config.baseReqSpec.get());
+    public {{classname}} {{classVarName}}Api() {
+        return {{classname}}.{{classVarName}}Api(config.baseReqSpec.get());
```
wing328: @spacether We will need to flag this change as a breaking change without fallback

cc @viclovsky any concern with this change and the change in the line below: https://github.com/OpenAPITools/openapi-generator/pull/1883/files#diff-06d453e322d6a8c6104519bc6df24fb1R40 ?

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -69,7 +69,7 @@ class {{classname}}(object):
 {{/isNullable}}
 {{^isNullable}}
         if {{name}} is not None:
-            self.{{name}} = {{name}}
+            self.{{name}} = {{name}}  # noqa: E501
```
wing328: Instead of adding `# noqa: E501`, what about using post file processing hook to format the python code (assuming the code formatter can handle long line nicely)?

spacether: If we were to change it to multiline we would need to do something like.
self.var_name = (
  var_value)
Which is an unusual way to define a variable, though valid.
Looking at the examples, I think the only line which is too long is: xml_itemp.py:L188:
`            self.prefix_namespace_wrapped_array = prefix_namespace_wrapped_array  # noqa: E501`
So I will just rename this variable to fix it.

@wing328 can you point me to an example of a post processing hook file? I am interested in using that functionality in a future PR. For example I would want to fix def method lines and array assignment lines to take up multiple lines if they are over 80 characters long.

wing328: You will see the following message when running the Python generator:
```
[main] INFO  o.o.c.languages.PythonClientCodegen - Environment variable PYTHON_POST_PROCESS_FILE not defined so the Python code may not be properly formatted. To define it, try 'export PYTHON_POST_PROCESS_FILE="/usr/local/bin/yapf -i"' (Linux/Mac)
[main] INFO  o.o.c.languages.PythonClientCodegen - NOTE: To enable file post-processing, 'enablePostProcessFile' must be set to `true` (--enable-post-process-file for CLI).
```

Please give it a try and let me know if you need any help.

modules/openapi-generator/src/main/resources/Java/libraries/rest-assured/ApiClient.mustache
wing328: Suggestion: undo the changes in the rest-assured templates so that we can merge this PR and make those breaking changes in another PR instead (from the title of this PR, it's hard to tell the API class name been renamed)

modules/openapi-generator/src/main/resources/python/model.mustache
spacether: Thank you for that info @wing328. I will use it in a future PR. For this one it doesn't look needed.

modules/openapi-generator/src/main/resources/Java/libraries/rest-assured/ApiClient.mustache
spacether: Removed the rest-assured fix for issue #13 per your suggestion

viclovsky: > Removed the rest-assured fix for issue #13 per your suggestion

Thanks. It would be great to keep backward compatibility. I suggested how to fix this issue here https://github.com/OpenAPITools/openapi-generator/issues/13#issuecomment-390635959

PR #1890, opened 2019-01-11
Rust reqwest improve

### LINKED ISSUE #525: [Rust] client codegen ignores all fields that are of type "formData"
##### Description

I think the title for this one is fairly self-descriptive.

I intend to fix this one myself fairly soon.

bcourtine:
### PR checklist

/cc @frol @farcaller @bjgill
--

### Description of the PR

This PR contains 3 improvements of the rust client generator with Reqwest library:

* Support form parameters (resolves #525).
* Generate query params "the reqwest way":  (no more trailing "?" in URL when there are no parameters).
* String path parameters are URL encoded (manage special chars, such as '@' in an email username used as path param, for example).


bcourtine:
> FYI. The change in body parameter name is due to an enhancement to preserve body parameter name defined in OAS v2 spec

Thanks. I wondered about that when I regenerated the samples. But since the code still compiled, I didn't investigate further.

wing328:
If no question/feedback from anyone, I'll merge this PR tomorrow.

modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
```diff
@@ -90,6 +86,19 @@ impl {{{classname}}} for {{{classname}}}Client {
         {{/authMethods}}
         {{/hasAuthMethods}}
 
+        {{#hasFormParams}}
+        let mut form_params = HashMap::new();
+        {{#formParams}}
+        {{#isFile}}
+        form_params.insert("{{{baseName}}}".to_string(), unimplemented!());
```
bjgill: I've given this a quick skim. As a thought, when do we want to fail if a user does attempt to create a generator with a file form parameter? This `unimplemented` would mean that we would fail at run-time when that endpoint was called.

Might it be worth using `compile_error!` (https://doc.rust-lang.org/std/macro.compile_error.html) so that the user discovers this at compile time instead?

bjgill: `compile_error!("File form parameters not supported by the rust generator");`, perhaps?

bcourtine: I ported the Hyper implementation (cf #525). We can switch to `compile_error!` macro, but we probably should do the same in `hyper` library templates to be consistent.

bcourtine: I tested `compile_error!` macro and I am questioning…
Since petstore example contains a file form param, with this macro, generated sample code does not compile anymore: the project cannot be used anymore to check syntax error regressions, for example.

So, I suggest keeping `unimplemented!`.

bcourtine:
Looking for @bjgill comment, I kept `unimplemented!` macro (cf. discussion), but added support for `multipart` operations (including file parts).

bcourtine:
@bjgill @wing328 Any news for this PR?

modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
```diff
@@ -49,20 +48,17 @@ impl {{{classname}}} for {{{classname}}}Client {
...
-{{/isKeyInQuery}}{{/isApiKey}}{{/authMethods}}{{/hasAuthMethods}}
+        {{/isKeyInQuery}}{{/isApiKey}}{{/authMethods}}{{/hasAuthMethods}}
 
         if let Some(ref user_agent) = configuration.user_agent {
             req_builder = req_builder.header(reqwest::header::USER_AGENT, user_agent.clone());
         }
 
-        {{#hasHeaderParams}}
-        {{#headerParams}}
+        {{#hasHeaderParams}}{{#headerParams}}
```
bjgill: A minor point - if you leave the mustache tags on their own lines, you'll avoid adding extra blank lines.

bcourtine:
@bjgill Oh, Thanks for the advice. I am not familiar with mustache… I used several tags by line specially because I thought it could reduce blank lines. I will fix that next time I have an evolution to implement.

Issue #1922, opened 2019-01-15
C# client response headers invalid parsing 


samoretc:
This issue was reported in swagger-codegen [here](https://github.com/swagger-api/swagger-codegen/issues/6389).

The values in the headers dictionary in the ApiResponse class are not getting properly set since the type of the header values in the IHTTP headers interface is an `IEnumerable<string>`. This means that the x.Value.ToString() from the following code snippet evaluates to  System.Collections.Generic.List`1[System.String]

`public ApiResponse<Object> v1UserActivitiesDateDateJsonGetWithHttpInfo (string date) { ... return new ApiResponse<Object>(localVarStatusCode, localVarResponse.Headers.ToDictionary(x => x.Key, x => x.Value.ToString()), null); }`

Ideally, I think the type of the headers field in the ApiResponse class should be changed from an `IDictionary<string, string>` to an `IDictionary<string, IEnumerable<string>>` so that callers know that there can be multiple values per header. Would it be acceptable to make this fix or not since it would break existing API users? 

Otherwise, we could take only the first header value or create a string that separates the different header values somehow like @omrat suggested. 

`public ApiResponse<Object> v1UserActivitiesDateDateJsonGetWithHttpInfo (string date) { ... return new ApiResponse<Object>(localVarStatusCode, localVarResponse.Headers.ToDictionary(x => x.Key, x => String.Join(",", x.Value.ToArray())), null); }`

wing328:
@samoretc thanks for reporting the issue. May I know if you've tried the `csharp-refactor` generator to see if it has the same issue?

samoretc:
I just tried, and it does have the same issue.

jimschubert:
@samoretc do you have a minimal test server that this can be reproduced on?

Regarding csharp-refactor, I've just tested this out with a valid multiple-header response:

```
$ wget -S --spider http://localhost:3000/pet/1
Spider mode enabled. Check if remote file exists.
--2019-01-17 14:48:30--  http://localhost:3000/pet/1
Resolving localhost (localhost)... ::1, 127.0.0.1
Connecting to localhost (localhost)|::1|:3000... connected.
HTTP request sent, awaiting response...
  HTTP/1.1 200 OK
  X-Powered-By: Express
  Example: Header1
  Example: Header2
  Example2: Header2
  Content-Type: application/json; charset=utf-8
  Content-Length: 198
[...]

PR #1935, opened 2019-01-17
Fix response headers in CSharp

gbrown-ce:
### PR checklist

### Description of the PR

Currently all response headers are parsed out as "System.String[]" since its doing a ToString on an IEnumerable.  We should be doing a Join on the strings in the list to make a comma delimited list of values.

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp/api.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp/api.mustache
@@ -297,12 +297,12 @@ namespace {{packageName}}.{{apiPackage}}
 
             {{#returnType}}
             return new ApiResponse<{{{returnType}}}>(localVarStatusCode,
-                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => x.Value.ToString()),
+                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => string.Join(",", x.Value)),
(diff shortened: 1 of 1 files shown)


gbrown-ce:
@mandrean @jimschubert This addresses https://github.com/OpenAPITools/openapi-generator/issues/1922

jimschubert:
@gbrown-ce Thanks for the PR.

Did you test this against single and multiple headers? If I remember correctly, RestSharp's `Headers` property can be either a list of strings or an individual string, which is why we often have type checks against `IList` and `ICollection` rather than `IEnumerable`.

If you're able to share the testing steps, would you mind sharing? If not, I can spin up minimal test API to evaluate as I also have a fix in place locally for the `csharp-refactor` generator.

gbrown-ce:
@jimschubert in this case the object is always IEnumerable<string> that I'm doing the Join on.  ToString is never going to be the right answer there.  I don't have a simple test scenario, but I have seen it work with a single header (I don't have a scenario with multiples at the moment)

While we're at it.  We tried switching to csharp-refactor, and we couldn't get it to build properly as a .NET Standard library.  Is there a way to do that?

jimschubert:
@gbrown-ce for the csharp-refactor, how are you generating and trying to build it? Also, which [.NET Standard Version](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)? If the target framework isn't compatible with your .NET Standard target… a simple workaround would be to add the csproj to `.openapi-generator-ignore` and make a change to the `targetFramework` version in the project file. This way, regeneration shouldn't affect your version changes.

I plan to do a major overhaul in the `csharp-refactor` code as there are some inconsistencies in the `csharp` generator (like `v5.0` target framework which doesn't exist is a flag for `netstandard1.3`).

If you'd like, we can also take the discussion to a separate issue for tracking.

PR #1937, opened 2019-01-17
[REQ][Asp.Net Core] Overrightable csproj for controllers and models

aaronclong:
The current Asp.Net Core server generator doesn’t allow for iterative server stub generation. Users are forced to generate a full solution, and despite the template modification and ignore, it still doesn’t provide a truly modular and automated solution for iterative development. Often additive changes to APIs require both clients and servers to be updated, but the current Asp.Net approach doesn’t emphasize this. We are left with modifying generated source code and manually copy/pasting changes in an error prone process.

It would be far more practical to simple generate a csproj with the models and controllers that another project would then reference. The actual core business logic could be IOCed via an interface into the generated controllers. It becomes the developer’s responsibility to actually implement the interface, which the generated code would invoke.

It would look similar to this:

```
public interface IPetControllerLogic 
{
	IList<string> GetPetList();
}

[Route("…")]
public class PetController
{
    public PetController(IPetControllerLogic controllerLogic)
    {		….
    }


    [HttpGet]
    [ProducesResponseType(typeof(IList<string>), 200)]
    [ProducesResponseType(500)]
    public IList<string> GetPets()
    {
        return this.contollerLogic.GetPetList();
    }
}

```
It would provide greater convenience because we could regenerate our API with spec changes and avoid overwriting business logic. The code generation process would be dramatically simplified. 

The downside to this approach would be some caveats in Asp.Net Core’s design. Namely MVC pulls in all controllers in referenced dlls. The consumer of this dll would automatically opt into all the controllers, and there are many cases where this functionally wouldn’t be ideal. In that case, there would need to be a config toggle to make the controller abstract. In doing so, the consumer would need to subclass it for it to be instantiated. It also affords the consumer the ability to provide extra attributes if needed.

I would like to propose a new Asp.net Core option or a separate command to do this.


modules/openapi-generator/src/main/resources/aspnetcorelib/2.1/tupleReturn.mustache
```diff
@@ -0,0 +1,8 @@
+var (statusCode, result) = await _impl.{{operationId}}({{#allParams}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}});
```
jimschubert: Why the use of tuples and an extractor here? Can this be rolled into a simple object, to ease evolution of the generated code?

jimschubert:
@aaronclong thanks for the contribution. I have some questions, forgive me if they seem out of the blue.

Your pull request's title has `[REQ]` in it, which reads as a feature request. Your implementation introduces a fully new generator, including your proposed new option for marking a controller as `abstract`.

I'm wondering a few things:
* Can the `aspnetcore` generator just be updated to output as a standalone server or as a lib via option? This would reduce maintenance overhead.
* Can `OVERRIDE_CONTROLLERS` be renamed? I have a feeling that the concept of making controller classes non-concrete is a generic enough thing which can be reused elsewhere (NancyFX, for instance)
* Your comments seem to imply that you're looking for service generation of interfaces which can be extended by the user and IoC'd into generated controllers.  I don't see that in this PR. Am I misunderstanding what you were saying, or is there more to come?  I do see the interface implementation generated within the controller class, but this looks as though you've created a controller decorator around a controller implementation abstraction rather than an injectable domain service.

I've been meaning to making the changes in the last bullet point to the `aspnetcore` generator for a while, but I've moved into my first new home a few months ago and that has been eating up all of my free time. I'm slowly getting back some of that free time as I'm more settled in, so I hope to get to this in the next month.

A-Joshi:
Guys - can we collaborate? I have made somewhat similar modifications to generate an abstract class as well as some more to create a library etc.

jimschubert:
@A-Joshi do you have a branch with your work that you can share?

modules/openapi-generator/src/main/resources/aspnetcorelib/2.1/tupleReturn.mustache
aaronclong: The main reason was that it was simple and native to .Net. 

While we can roll it into an object, Tuples do the same thing with less code. However, can also see value in an object being returned.

aaronclong:
@jimschubert thanks for getting back to me! I totally understand moving is terrible.

I think setting it as an option would probably be a smart move. I made it a separate command because the templates are fairly different for the lib use case. Also it was a little bit easier to code it as a standalone for someone not super familiar with the code base. However, I am willing to refactor it into the current command, but I think it would be a good idea to keep the templates separated if possible.

I am not married to any names. I am just generally bad at naming variables, classes, etc. If you have any ideas for better ones, I would really appreciate it. 

@A-Joshi I am defiantly down to collaborate if you are. I'll read your issue and get back to you.

A-Joshi:
@aaronclong I have the code working to generate a library etc. and (I think a pattern to allow roundtrips) waiting for some internal approvals to submit a PR. Send me an email if.

jimschubert:
@aaronclong to keep templates separate under the existing generator, we can create a folder in the existing template's directory and conditionally load those templates based on a client option.

I wouldn't recommend separating as a "library" option, because I think that could be confusing.

aaronclong:
@A-Joshi I just sent you and email.

If the approval process is too much or long, I could attempt to redo my merge request with @jimschubert comments.

A-Joshi:
@aaronclong:

Started the process should have the approvals early next week - might try to get the code out this week itself.

Roundtrip - be able to regenerate the full library when the API changes without having to ignore files etc. Essentially keep the generated code as a library that is used in the implementation. 

The concept I used is as follows:
* Generate a full abstract class and models for the UI. I would have preferred an interface but as I mentioned earlier you cannot decorate an interface with ROute and other annotations.
* Package the generated code into a versioned nuget package.
* The implementing server will extend the abstract controller class and implement the methods. The Server development only has to meet the abstract class implementation but does not touch the generated code. 
* If the API is change in a breaking way the server will automatically fal compilation since the implementing class compilatin will fail.
* If the server code is changed as long as the abstract class implementation remains intact there is no impact.

To accomplish this I made a few (relatively small) changes:
* Added the ability to control the generated code with generatewwwroot (true/false), and a buildTarget(program/library), and more fine grained options - classmodifier (abstract, partial or none), operationModifier(abstract, virtual, none), generateBody(true/false), 
* WIth that the template file changes to the existing templates were very small - - to use the properties above.
* Added a new template to generate a nuspec file if the program target is library and skip building the Program.cs, Startup.cs Dockerfile etc.

Things I want to do next is to add an option to specify if the generation shuld be sync or async (Task<> wrapped).

If you can send me an email directly I can share the changes (I shared them with Jim) before the PR.

wing328:
> If you can send me an email directly I can share the changes (I shared them with Jim) before the PR.

What about we merge this PR into a branch of openapi-generator (e.g. aspnetcore-enhancement) so that more people can contribute with PRs (to make collaboration easier)?

Issue #1944, opened 2019-01-19
[BUG] Ruby generator creates invalid Ruby code for integer enums


UkuLoskit:
##### Description
Ruby generator creates invalid Ruby code for integer enums, the produced enum does take into account Ruby syntactical rules for identifiers which only allow identifiers starting with underscore or a non-integer.

##### openapi-generator version

The issue exists in both the latest master (4.0.0) and 3.3.4.

##### OpenAPI declaration file content or url

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: keyword test
  license:
    name: MIT
paths:
  /products:
    get:
      parameters:
      - in: query
        name: color
        required: true
        schema:
          $ref: '#/components/schemas/ValidNumber'
      responses:
        '200':
          description: OK
components:
  schemas:
     ValidNumber:
      type: integer
      enum:
        - 1
        - 2
        - 3
        - 4
        - 5

```

##### Command line used for generation

`java -jar openapi-generator-cli.jar generate -g ruby -i sample.yaml`

##### Steps to reproduce
1. Generate the code from the sample
2. The produced code for the model is the following (`lib/openapi_client/models/valid_number.rb`) 
```ruby

=begin
#keyword test

#No description provided (generated by Openapi Generator https://github.com/openapitools/openapi-generator)

OpenAPI spec version: 1.0.0

Generated by: https://openapi-generator.tech
OpenAPI Generator version: 3.3.4

=end

require 'date'

module OpenapiClient
  class ValidNumber

    1 = 1.freeze
    2 = 2.freeze
    3 = 3.freeze
    4 = 4.freeze
    5 = 5.freeze

    # Builds the enum from string
    # @param [String] The enum value in the form of the string
    # @return [String] The enum value
    def build_from_hash(value)
      constantValues = ValidNumber.constants.select { |c| ValidNumber::const_get(c) == value }
      raise "Invalid ENUM value #{value} for class #ValidNumber" if constantValues.empty?
      value
    end
  end
end
```
3. The *actual* enum declaration is not valid Ruby, as Ruby identifiers cannot start with numbers.
4. The *expected* enum should have a character or underscore prefixed (should be `N1 = 1.freeze`)

##### Related issues/PRs

I was not able to find any related PRs.

##### Suggest a fix
I was able to find the cause of the issue, and will create a PR.

wing328:
Closed via #1945

PR #1945, opened 2019-01-19
Fix Ruby invalid enum generation for numeric enums

UkuLoskit:
### Description of the PR
Attempts to fix issue https://github.com/OpenAPITools/openapi-generator/issues/1944

After these changes the code generator adds the `N` prefix to the enum names as expected.
@cliffano @zlx


wing328:
> @ackintosh thanks for the quick feedback. I've addressed the comment. Should I merge my change into a single commit?

We'll squash the commits into one when merging the PR

Issue #1927, opened 2019-01-16
[BUG][jaxrs-jersey] Generating @DefaultValue("null") for query parameter with referenced enum is wrong


jmini:
##### Description

Generatated `@DefaultValue("null")` in the `Api` file is wrong.

With the provided example, in `SomeApi`:

```java
    @GET
    @Path("/change")
    
    @Produces({ "application/json" })
    @io.swagger.annotations.ApiOperation(value = "", notes = "", response = String.class, tags={ "Some", })
    @io.swagger.annotations.ApiResponses(value = { 
        @io.swagger.annotations.ApiResponse(code = 200, message = "Ok", response = String.class) })
    public Response sayGet_1(@ApiParam(value = "", defaultValue="null") @DefaultValue("null") @QueryParam("color") BasicColor color
,@Context SecurityContext securityContext)
    throws NotFoundException {
        return delegate.sayGet_1(color,securityContext);
    }
```

Should be `@DefaultValue("RED")` in order for the server to start.

##### openapi-generator version

There was no issue with `3.2.0`. Reproducible with `3.3.4` and `4.0.0-SNAPSHOT`

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.1
info:
  title: Default value test
  version: '1.0'
servers:
  - url: 'http://localhost:8088/'
paths:
  /ping:
    get:
      tags:
        - Some
      operationId: pingGet
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: string
  /say:
    get:
      tags:
        - Some
      operationId: sayGet
      parameters:
        - name: message
          in: query
          required: false
          schema:
            type: string
            default: hello
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: string
  /reply:
    get:
      tags:
        - Some
      operationId: replyGet
      parameters:
        - name: something
          in: query
          required: false
          schema:
            type: string
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: string
  /select:
    get:
      tags:
        - Some
      operationId: selectGet
      parameters:
        - name: dayOfTheWeek
          in: query
          required: false
          schema:
            $ref: "#/components/schemas/Day"
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: string
  /change:
    get:
      tags:
        - Some
      operationId: sayGet
      parameters:
        - name: color
          in: query
          required: false
          schema:
[...]

PR #1953, opened 2019-01-21
[core] cleanup: remove OpenAPI parameter from CodegenConfig methods

jmini:
### PR checklist

### Description of the PR

~In order to solve #1927 I will need to access the OpenAPI instance in `DefaultCodegen#toDefaultValue(..)` in order to be able to see if the schema contains a `$ref` and if this is the case, to see if the referenced schema has a default value or not.~

~The main change of this PR is to have:~

* ~`public String toDefaultValue(Schema p, OpenAPI openAPI)`~
* ~Instead of `public String toDefaultValue(Schema p)`~

~To fulfill this requirement the `OpenAPI` is added as parameter to some methods of `DefaultCodegen`.~

~Some of those methods also had `Map<String, Schema> allDefinitions` which corresponds to `openAPI.getComponents().getSchemas()`. This parameter was also removed from the methods, and the `openAPI` instance is used to access to all schemas.~

_Based on feedback the initial intention of this PR was changed._

Historically, to access the `OpenAPI` instance, or all schemas of the OpenAPI instance, parameters were used in different methods. Example:

* `CodegenConfig#fromOperation(String, String, Operation, Map<String, Schema>, OpenAPI)`
* `CodegenConfig#fromModel(String, Schema, Map<String, Schema>)`
* `DefaultCodegen#fromResponse(OpenAPI, String, ApiResponse)`
* `DefaultCodegen#DefaultCodegen.fromRequestBody(RequestBody, Map<String, Schema>, Set<String>, String)`
* `DefaultCodegen#createDiscriminator(String, Schema, Map<String, Schema>)`
* ...

Then @wing328 noticed that access to this information is requested in a lot of places (for unaliasing).
So he introduced a new approach: having two members `globalOpenAPI` and `globalSchemas`.

Having two ways for doing the same task is not consistent. It is harder to understand, to maintain and can lead to errors (for example all the unit tests were not using `globalOpenAPI`, meaning that they were not reproducing real cases, because when the generator is used for real, then the `globalOpenAPI` member is set).

This PR cleans the situation by removing the original approach completely. There is no `openAPI` or `allDefinitions` parameter anymore in the methods. Instead they use the `globalOpenAPI` member which is renamed `openAPI`.

In addition `globalSchemas` was only a convenient and null-pointer safe access to `openAPI.getComponents().getSchemas()`. It is also removed.

This change is mentioned in the migration guide "From 3.x to 4.0.0"

In my opinion the PR improves the readability of the code and reduces the potential coding mistakes.


jmini:
Input @wing328 

> have you considered using
> https://github.com/OpenAPITools/openapi-generator/blob/2e3d04ce0bd19cd7003b2cd804e2e92e61b6d738/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L143-L144
> ?

This is a very good feedback.

After a short discussion on Gitter we agreed that:

* We will rename `this.globalOpenAPI` to `this.openAPI` (and the setter accordingly)
* We will use only this way to access to the openAPI instance, meaning that all methods `fromModel(..)`, `fromRequestBody(..)` will no longer have `openAPI` has parameter.

This way we are consistent and have only one pattern in place.

PR #2016, opened 2019-01-29
[python] [bug] Fixed handling of dotted module names

svenpanne:
### PR checklist

- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. (Irrelevant point, Travis CI checked the PR in my fork.)

### Description of the PR

Previously, if you used a packageName of the form `foo.bar.baz`, half of the
generated files of the python generator went into a subdirectory
`foo/bar/baz` (correct), the other half went into a subdirectory
`foo.bar.baz` (incorrect).

@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
@@ -203,10 +203,10 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
 
         if (generateSourceCodeOnly) {
             // tests in <package>/test
-            testFolder = packageName + File.separatorChar + testFolder;
+            testFolder = packagePath() + File.separatorChar + testFolder;
             // api/model docs in <package>/docs
-            apiDocPath = packageName + File.separatorChar + apiDocPath;
-            modelDocPath = packageName + File.separatorChar + modelDocPath;
+            apiDocPath = packagePath() + File.separatorChar + apiDocPath;
+            modelDocPath = packagePath() + File.separatorChar + modelDocPath;
         }
         // make api and model doc path available in mustache template
         additionalProperties.put("apiDocPath", apiDocPath);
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -201,12 +201,14 @@ public void processOpts() {
         additionalProperties.put(CodegenConstants.PACKAGE_NAME, packageName);
         additionalProperties.put(CodegenConstants.PACKAGE_VERSION, packageVersion);
 
+        // NOTE: The null case can happen during PythonClientOptionsTest.checkOptionsProcessing().
+        String packagePath = packageName == null ? "" : packageName.replace('.', File.separatorChar);
```
rienafairefr: Would be cleaner to have packageName != null during the unit test `PythonClientOptionsTest.checkOptionsProcessing` ?

svenpanne: Yes, that would be cleaner, but due to the slightly obscure mocking Kung Fu in the test I was unable to come up with something better.

I propose to land this change as it is (perhaps add a TODO), the current module name handling is totally unusable in non-trivial existing Python projects. Alternatively, somebody could give me a concrete hint how to fix the test... :wink:

rienafairefr:
Tested on a real spec & project, works fine :+1:

wing328:
@svenpanne thanks for the PR. What about using the `packagePath()` function approach in https://github.com/OpenAPITools/openapi-generator/pull/2041 ?

Later we'll create a Python abstract class for Python client and server generators so keeping things a bit more consistent will make the refactoring easier.

svenpanne:
> @svenpanne thanks for the PR. What about using the `packagePath()` function approach in #2041 ? [...]

Good idea, done. Note that I've fixed and improved the corresponding options test a bit, so the funny null check is gone. `jmockit` is a strange beast if you don't use it on a daily basis... :confused:

Issue #2011, opened 2019-01-28
[BUG][PHP] oneOf produces type `OneOfModelAModelB` in Model $openAPITypes


sakari:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?script-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

PHP model is generated with `OneOfModelAModelB` as type for `$openAPITypes` for

```
foo: 
  oneOf:
    - $ref: 'modelA'
    - $ref: 'modelB'
```
##### openapi-generator version

master at 046db19a8590f3557ae1e04be75792bbeb39cb69

##### Suggest a fix

I think that it would be possible to create a type string like `oneOf[ModeA, ModelB]` in AbstractPhpCodegen.getTypeDeclaration and then deal with that in deserialiser like is done with `map[K, V]` types. Ie match the type string to `oneOf[`there and try to deserialize all the enclosed options until a valid deserialization is found.

If this seems like a legit approach I might be willing to do it as I need some reasonable way to handle this. Or if this has been considered before but the plan lacks implementation that would be something I could do also.

wing328:
@sakari thanks for offering help. Sure we can work together on that. Do you mind pinging me via https://gitter.im (ID: wing328) so that we can come up with a plan to add the support of oneOf (and potentially anyOf)?

jeffski:
I'm running in to this problem also. For now the only way around it is for me to manually update the generated code (not a long term solution). In my situation if the generator could do this for me that would be great, not sure if this is the correct approach globally though.

Here is the work around that I put in place in my generated SDK: https://github.com/shotstack/shotstack-sdk-php/pull/2/commits/37770909704f162c359d4b241deceff33604091e

I added a parent type (Asset) and then have 3 classes (my oneOf types) extend the parent type (TitleAsset, ImageAsset and VideoAsset). In the Clip type I replaced the `OneOfTitleAssetImageAssetVideoAsset` type with my new class `\Shotstack\Client\Model\Asset`.

ksvirkou-hubspot:
Hi everyone 
We have the same issue 
generator version is 5.1.0

[...]

PR #2021, opened 2019-01-30
[PHP] Provide oneOf types in php code generation

sakari:
(this is fairly work in progress but I would like to get feedback if this is at all sensible approach to this before going much further) 

Fixes https://github.com/OpenAPITools/openapi-generator/issues/2011

For Models openAPITypes map the idea is to encode oneOf values as string
`oneOf[type1, .., typeN]` in the property dataType template variable

For Models that consist of a oneOf block like

```
model:
  oneOf:
    - $ref: modelA
    - type: array
      items: string

```

We generate a new template variable oneOfTypes containing the list of
options which in turn get used to generate a new method `Model::oneOf`
which return a list

```
['\fully\qualified\name\modelA', 'string[]']
```

@jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko @renepardon @wing328

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -244,10 +244,34 @@ class ObjectSerializer
             $data = is_string($data) ? json_decode($data) : $data;
             $subClass = substr($class, 0, -2);
             $values = [];
+            if (!is_array($data)) {
+                throw new \InvalidArgumentException('expected array');
+            }
             foreach ($data as $key => $value) {
                 $values[] = self::deserialize($value, $subClass, null);
             }
             return $values;
+        } elseif (substr($class, 0, strlen('oneOf[')) === 'oneOf[') {
+          $inner = substr($class, strlen('oneOf['), -1);
+          $options = explode(',', $inner);
+          $deserialized = [];
+          $errors = [];
+          foreach ($options as $option) {
+            try {
+              $deserialized[] = ['option' => trim($option), 'value' => self::deserialize($data, trim($option), $httpHeaders)];
+            } catch (\InvalidArgumentException $e) {
+              $errors[] = $e->getMessage();
+            }
+          }
+          if (count($deserialized) > 1) {
+            throw new \InvalidArgumentException(
+              'too many matching options for oneOf: '. implode(', ', array_map(function($option) { return $option['option']; }, $deserialized))
+            );
+          }
+          if (count($deserialized) < 1) {
+            throw new \InvalidArgumentException('no values matched oneOf:\n    '. implode('\n    ', $errors));
+          }
+          return $deserialized[0]['value'];
         } elseif ($class === 'object') {
             settype($data, 'array');
             return $data;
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -244,10 +244,34 @@ class ObjectSerializer
...
             }
             return $values;
+        } elseif (substr($class, 0, strlen('oneOf[')) === 'oneOf[') {
+          $inner = substr($class, strlen('oneOf['), -1);
+          $options = explode(',', $inner);
+          $deserialized = [];
+          $errors = [];
+          foreach ($options as $option) {
+            try {
+              $deserialized[] = ['option' => trim($option), 'value' => self::deserialize($data, trim($option), $httpHeaders)];
```
sakari: the deserializer needs to start to throw invalidargumentexceptions to help pick the right model to in the end.

modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -301,17 +325,26 @@ class ObjectSerializer
...
-                if (!isset($propertySetter) || !isset($data->{$instance::attributeMap()[$property]})) {
-                    continue;
+            $reverseAttributes = array_flip($instance::attributeMap());
+            if (!is_object($data)) {
+                return null;
+            }
+            if (!empty($instance::oneOf())) {
+                return self::deserialize($data, "oneOf[". implode(', ', $instance->onefOf())  ."]", $httpHeaders);
+            }
+            foreach(get_object_vars($data) as $key => $propertyValue) {
```
sakari: we need to go through all the data fields and and see if those are settable to the current model instead of looking at the setters so that we can cull the oneof options. This cause us to also require a mechanism for setting additional properties

modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -301,17 +325,26 @@ class ObjectSerializer
...
-                $propertySetter = $instance::setters()[$property];
-
-                if (!isset($propertySetter) || !isset($data->{$instance::attributeMap()[$property]})) {
-                    continue;
+            $reverseAttributes = array_flip($instance::attributeMap());
+            if (!is_object($data)) {
+                return null;
+            }
+            if (!empty($instance::oneOf())) {
+                return self::deserialize($data, "oneOf[". implode(', ', $instance->onefOf())  ."]", $httpHeaders);
```
sakari: convert a model consiting of a oneof block to plain type string. This of course does not deal with cases where the model would contain additional stuff.

wing328:
Instead of introducing `oneOfTypes`, what about storing the data type in `oneOf` instead?

sakari:
sure, I was just unsure if that is already used or will be used for something else

sakari:
@wing328 there is also a bunch of changes related validation and deserialization that I had to ram to get the deserializer corretly distinguish between the options in https://github.com/smartlyio/openapi-generator/tree/deep-validation.

If you think that this approach has hope of getting merged to upstream I'll clean that and cherry-pick those to this branch

ackintosh:
@sakari Thanks for the PR!

I'm interested in [the approach](https://github.com/OpenAPITools/openapi-generator/pull/2021#issuecomment-462508910) but I prefer to keep PR diffs as small as possible. Is it possible to address the approach as day-2 requirement in other PR?

PR #2034, opened 2019-01-31
[Python] handle nullable parameters with None added to allowed_values

### LINKED ISSUE #1997: [BUG] nullable enum property
#### Bug Report Checklist

- [X] What's the version of OpenAPI Generator used?   3.3.0

##### Description

If encountered this using the python but I feel like this is the case in a lot of cases
I encountered an API that was publishing a 2.0 spec, but I'm getting resultst from the API that are out-of-spec and not possible to be parsed by a generate openapi-generator client, mostly sending null on non-nullable properties. But the issue is for enum values, I'm not sure how a client should validate:

a small spec that is apparently valid according to 3.0.2 
```
openapi: 3.0.1
paths:
  /test:
    summary: test
info:
  description: test
  version: 1.0.0
  title: test
components:
  schemas:
    Type:
      properties:
        prop:
          nullable: true
          enum: [A, B, C]
          type: string
```
if the enum is nullable, then getting a `null` as a result data from an API call should be valid and we should get a property typed like `Enum? in C#, or a `None` value in Python, and that should be valid. But the code in the case of an enum lists the allowable values excluding null, even if the enum property is specified nullable

##### Suggest a fix

- explicitely adding None in the `allowed_values` list in Python seems to at least make the client accept what the API sends, not sure if that's the full extent that a fix would take

rienafairefr:
### PR checklist

@wing328 @taxpon @frol @mbohlool @cbornet @kenjones-cisco

### Description of the PR

I think this fixes #1997, if a parameter is nullable, then `None` should be an allowed value.

I ran `./bin/saples/python-petstore-all.sh' it created a bunch of changes that don't seem related, not sure why.

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/model.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model.mustache
@@ -107,7 +107,7 @@ class {{classname}}(object):
 {{/isNullable}}
 {{#isEnum}}
 {{#isContainer}}
-        allowed_values = [{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
+        allowed_values = [{{#isNullable}}None,{{/isNullable}}{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
 {{#isListContainer}}
         if not set({{{name}}}).issubset(set(allowed_values)):
             raise ValueError(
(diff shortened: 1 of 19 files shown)


modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -108,6 +108,9 @@ class {{classname}}(object):
 {{#isEnum}}
 {{#isContainer}}
         allowed_values = [{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
```
spacether: Why not instead, in each allowed_values definitions add something like?
allowed_values = [{{#isNullable}}None, {{/isNullable}}{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501

spacether:
> I ran `./bin/saples/python-petstore-all.sh' it created a bunch of changes that don't seem related, not sure why.

It created a bunch of changes because it runs:
./bin/python-asyncio-petstore.sh
./bin/python-petstore.sh
./bin/python-tornado-petstore.sh

modules/openapi-generator/src/main/resources/python/model.mustache
rienafairefr: Yeah, would probably be cleaner. I wasn't sure, feared to add too much commas or something :-)

wing328:
> I ran `./bin/samples/python-petstore-all.sh' it created a bunch of changes that don't seem related, not sure why.

We only added python scripts to the ensure up-to-date script recenlty so don't worry about those additional changes. Please commit those changes and we'll review accordingly.

rienafairefr:
@spacether @wing328 PR updated, thanks for the review guys

PR #2029, opened 2019-01-31
[PYTHON] Issue #2028 - additionalProperties support

ivan-gomes:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Fixes #2028. Adapted from accepted PR https://github.com/swagger-api/swagger-codegen/pull/8414. 

@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess

Issue #2028, opened 2019-01-31
[BUG] [PYTHON] additionalProperties not supported


ivan-gomes:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Generated Python clients do not have the ability to access the additional properties of models with `additionalProperties` defined. The only ones that can be accessed are those that are explicitly defined in the model and the rest are ignored. In contrast, the generated Java models extend HashMap when `additionalProperties` are defined and they can get accessed by Map#get(String).

##### openapi-generator version

4.0.0

##### OpenAPI declaration file content or url

```yaml
  Element:
    type: "object"
    required:
    - "id"
    properties:
      id:
        type: "string"
      name:
        type: "string"
    additionalProperties:
      type: "object"
```
Full specification: https://github.com/Open-MBEE/mms/blob/develop/mms-ent/repo-amp/src/main/amp/web/mms/mms.swagger.yaml

##### Command line used for generation

```shell
java -jar /path/to/openapi-generator-cli.jar generate -i /path/to/mms.swagger.yaml -l python -o ~/tmp
```

##### Steps to reproduce

Inspecting the generated `Element.py` file after executing the command above with the specification above. There is also a usage of `additionalProperties` in the Petstore API that exhibits the same behavior.

##### Related issues/PRs

Clone of issue made and fixed in `swagger-api/swagger-codegen`: https://github.com/swagger-api/swagger-codegen/issues/8075

##### Suggest a fix

I believe that the additional properties can be stored in a dict, analogously to how they are handled in Java.

spacether:
This feature was added in https://github.com/OpenAPITools/openapi-generator/pull/4154

j--wong:
Hi, 

I just wanna double check, so to use the `additionalProperties` support, do I need to use python experimental generator `-g python-experimental` instead of python `-g python`?

Thanks,
Josh

spacether:
Yes, that's right

j--wong:
[...]

Issue #623, opened 2018-07-23
[Python] Inheritance doesn't work 


emreakca:
##### Description

The generated Python SubClass doesn't inherited from the BaseClass, when allOf is used.
```yaml
SubClass:
  allOf:
    - $ref: '#/components/schemas/BaseClass'
```

The generated class shows
```python
class SubClass(object):
```
instead of
```python
class SubClass(BaseClass):
```
This issue was already posted (#453), but it's not solved nor answered. 
##### openapi-generator version

3.1.1

##### Command line used for generation

```
java -jar ~/openapi-generator-cli-3.1.1.jar generate -g python -i openapi.json -o generated/python-client
```
##### Suggestion
Either do the inheritance correctly (as shown at the beginning), with the right constructors (```__init__```) 
or add the Properties of the BaseClass to the SubClass, as in swagger 2.

macjohnny:
@emreakca @poli-granot this could be fixed by applying the changes in https://github.com/swagger-api/swagger-codegen/commit/843c638c2a86926f201d006daeb2e8c5028d6cfb to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/model.mustache#L13 and other relevant files. would you like to file a PR for this?

spacether:
I am working on a PR which will fix this issue. The problem is that the mustache files iterate through #vars. #vars does not include the variables inherited from the allOf class. However, the requiredVars and optionalVars DO include those needed properties. My fix will iterate over the requiredVars and optionalVars.

spacether:
Inheritance is working in python-experimental
Composed schema inheritance is implemented by storing instances of "ancestor" allOf models rather than using inherited classes.
This was done because multiple allOf models can contain the same property with differing enum and validation constraints on it.
Python class based inheritance generally assumes that a property has its definition/validation in one source of truth class location.

Issue #453, opened 2018-07-04
[Python] Generated model for component with "allOf" has no attributes


ghost:
##### Description

Given below is a snippet of the input YAML file I am using. I have several components in the original file that are all have the same structure as "AddEmailContactRequest". All of these components' classes are generated without any attributes, although the YAML file is read correctly.
Here is the relevant snippet of the generated code for "AddEmailContactRequest":

    `class AddEmailContactRequest(object):
        openapi_types = {
        }
        attribute_map = {
        }

        def __init__(self):
            self.discriminator = None`

While what it **should** generate is the below code (I manually fixed it because, well, I need to keep working :) ):

    `class AddEmailContactRequest(object):
        openapi_types = {
            'type': 'str',
            'attributes': 'EmailContact'
        }
        attribute_map = {
            'type': 'type',
            'attributes': 'attributes'
        }
    
        def __init__(self, type=None, attributes=None):  # noqa: E501
            self._type = None
            self._attributes = None
            self.discriminator = None

            self.type = type
            self.attributes = attributes`

NOTE - I am also getting several warnings as follows during generation:
[main] WARN  o.o.codegen.DefaultCodegen - Unknown type found in the schema: object

##### openapi-generator version

I'm building from branch **4.0.x**, but have encountered this problem with tag **v3.0.3** as well.

##### OpenAPI declaration file content or url

```yaml
openapi: "3.0.1"
servers:
  - url: 'http://www.myserver.com/api'
info:
  version: '1.0'
  title: MYSERVER


paths:
  '/email-contacts':
    post:
      summary: addEmailContact
      tags:
        - sometag
      operationId: addEmailContact
      responses:
        '200':
          description: ''
          content:
            application/json:
              schema:
                type: object
                properties:
                  data:              
                    $ref: '#/components/schemas/AddEmailContactResponse'
      security: []
      x-unitTests: []
      x-operation-settings:
        CollectParameters: false
        AllowDynamicQueryParameters: false
        AllowDynamicFormParameters: false
        IsMultiContentStreaming: false
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                data:            
                  $ref: '#/components/schemas/AddEmailContactRequest'
        required: true


components:
  schemas:
    AddEmailContactRequest:
      allOf:
        - $ref: '#/components/schemas/JsonAPIPostRequestWrapper'
      title: addEmailContactRequest
      type: object
      properties:
        attributes:
[...]

PR #2049, opened 2019-02-04
[Python] add additionalProperties support, Option2

spacether:
### PR checklist

### Description of the PR

- Adds support for additionalProperties to the Python generator
- Fixes inheritance in python when AllOf is used
- Adds additionalProperties codegenParameter to the model that the mustache files have access to in case we want to use the schema for examples in the future
- Adds the ability to access and assign values to a model instance using brackets. model_instance['var_name'] = var_val
- Simplifies model instance variable storage by storing all values in a data dict. This will make it easier to send Nullable value=None items, because items in the data store are correct
- Switches parameter and discriminator typing to use actual python classes
- Adds type checking with TypeError/ValueError/KeyError exceptions which tell where in the data the problem is (data is in the exception string and exception.path_to_item)
  - Model instances can be instantiated with type checking with the  _check_type variable
  - Endpoint hits can check type before sending out data with the _check_type variable
  - Deserialization automatically checks type on all received data

### Issues that will be closed if this is merged:
- Inheritance not working: https://github.com/OpenAPITools/openapi-generator/issues/623
- When model only has allOf, it is blank: https://github.com/OpenAPITools/openapi-generator/issues/453 
- Fixes additionalproperties: https://github.com/OpenAPITools/openapi-generator/issues/2028 

### New Changes
- Model variable types, discriminators values, and response types are now all classes
- api_client call_api method keyword argument changed from response_type to response_types_mixed
- Mode instantiation:
  - required arguments with no enums of length one are positional
  - required arguments WITH enums of length one are set as keyword arguments with defaults = the enum value
  - option arguments are keyword argument **kwargs
  - This should be completely backwards compatible if people are using all keyword arguments which still works

### Breaking Changes:
- Received date and date-time data that is not parsable will now raise an ApiValueError which can be captured with ValueError, Exception, ApiValueError, or OpenApiException. Date/date-time variable will not fall back to string values. Instead one must set the type of that variable in the spec equal to {} which allows that variable to hold any of the openapi types. An example of this can be seen at: samples/client/petstore/python/tests/test_deserialization.py::test_deserialize_str_to_date_failure

### Note:
This is an alternative solution to this PR: https://github.com/OpenAPITools/openapi-generator/pull/2029


spacether:
So I'm not sure why the Shippable test failed. The log doesn't list verbose output. It's last lines are
```
./bin/utils/test-fake-petstore-for-all.sh
# START SCRIPT: ./bin/utils/test-fake-petstore-for-all.sh
[OAS 2.0] Executed ada successfully!
[OAS 3.0] Executed ada successfully!
[OAS 2.0] Executed ada-server successfully!
[OAS 3.0] Executed ada-server successfully!
[OAS 2.0] Executed android successfully!
[OAS 3.0] Executed android successfully!
[OAS 2.0] Executed apache2 successfully!
[OAS 3.0] Executed apache2 successfully!
[OAS 2.0] Executed apex successfully!
[OAS 3.0] Executed apex successfully!
[OAS 2.0] Executed aspnetcore successfully!
[OAS 3.0] Executed aspnetcore successfully!
[OAS 2.0] Executed bash successfully!
[OAS 3.0] Executed bash successfully!
[OAS 2.0] Executed c successfully!
[OAS 3.0] Executed c successfully!
[OAS 2.0] Executed clojure successfully!
[OAS 3.0] Executed clojure successfully!
[OAS 2.0] Executed cwiki successfully!
[OAS 3.0] Executed cwiki successfully!
[OAS 2.0] Executed cpp-qt5-client successfully!
[OAS 3.0] Executed cpp-qt5-client successfully!
[OAS 2.0] Executed cpp-qt5-qhttpengine-server successfully!
[OAS 3.0] Executed cpp-qt5-qhttpengine-server successfully!
[OAS 2.0] Executed cpp-pistache-server successfully!
[OAS 3.0] Executed cpp-pistache-server successfully!
[OAS 2.0] Executed cpp-restbed-server successfully!
[OAS 3.0] Executed cpp-restbed-server successfully!
[OAS 2.0] Executed cpp-restsdk successfully!
[OAS 3.0] Executed cpp-restsdk successfully!
ERROR: Failed to run cpp-tizen
java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -g cpp-tizen -o /tmp/openapi-generator-test-fake-petstore/2.0/cpp-tizen
```

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -33,48 +33,104 @@ class {{classname}}(object):
...
+{{/additionalProperties}}
+{{^additionalProperties}}
+            raise KeyError("{0} has no key '{1}'".format(
+                type(self).__name__, name))
+{{/additionalProperties}}
+
+        passed_type = self.recursive_type(value)
+        if type(name) != str:
+            raise TypeError('Variable name must be type string and %s was not' % name)
+        elif passed_type != required_type and check_type:
```
ivan-gomes: Thank you very much for creating this pull request! I'm the author of https://github.com/OpenAPITools/openapi-generator/issues/2028 and https://github.com/OpenAPITools/openapi-generator/pull/2029 and I think the solution you have here is cleaner and more robust. 

Just one minor issue I've noticed so far. I tested it with my use case of
```javascript
additionalProperties:
            type: "object"
```
which caused this error to be raised when doing `model["foo"] = "bar"`with `ValueError: Variable value must be type object but you passed in str`. It looks like the type checking isn't taking into account subclasses. I would really appreciate if you could address this use case. I would be happy to test for you as well.

ivan-gomes:
> Thank you very much for creating this pull request! I'm the author of #2028 and #2029 and I think the solution you have here is cleaner and more robust.
> 
> Just one minor issue I've noticed so far. I tested it with my use case of
> 
> ```yaml
> additionalProperties:
>     type: "object"
> ```
> 
> which caused this error to be raised when doing `model["foo"] = "bar"`with `ValueError: Variable value must be type object but you passed in str`. It looks like the type checking isn't taking into account subclasses. I would really appreciate if you could address this use case. I would be happy to test for you as well.

Copying my code comment here as it didn't add it to the conversation. Thanks again!

modules/openapi-generator/src/main/resources/python/model.mustache
spacether: Hi Ivan. You're welcome. Thank you for yours. We both had the same idea to add a dict to store the values, this pull request just implemented it as a general case. Thanks for the heads up about that failing case. To fix it, I will add more samples. For the example you gave the code was right to raise an error, because you should have passed in:
model["foo"] = {"some_var": "bar"}
But the error message was wrong in the type that it was demanding. It should have wanted a dict rather than an object. We collapse complex schema types down elsewhere, so I will find that code and use it.

ivan-gomes: The intent of additionalProperties with type object is that any value, regardless of type, is accepted. Upon review of the OpenAPI spec I should be using `additionalProperties: true` to represent that intent (source: https://swagger.io/docs/specification/data-models/dictionaries/). I will try this case out for us.

ivan-gomes: When `additionalProperties: true` is used, the codegen treats it as if `additionalProperties` is not enabled. I suspect that the fact that its type isn't defined causes it to be ignored.

spacether: Ivan, what does inspecting the values in the openapi spec show when we use:
- additionalProperties: true
- additionalProperties: {}
If we see it in those results, we can handle it in the java code.
One can inspect the values in the openapi spec with the below line:
```
java -DdebugOpenAPI -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -t modules/openapi-generator/src/main/resources/python -i modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -g python -o samples/client/petstore/python -DpackageName=petstore_api
```

ivan-gomes: Modifying `additionalProperties` in `AdditionalPropertiesClass` in `modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml` to `additionalProperties: {}` results in 
```javascript
      "AdditionalPropertiesClass" : {
        "type" : "object",
        "properties" : {
          "map_property" : {
            "type" : "object",
            "additionalProperties" : {
              "type" : "string"
            }
          },
          "map_of_map_property" : {
            "type" : "object",
            "additionalProperties" : {
              "type" : "object",
              "additionalProperties" : {
                "type" : "string"
              }
            }
          }
        },
        "additionalProperties" : {
          "type" : "object"
        }
```
Note `"type" : "object"`.

Using `additionalProperties: true` results in
```javascript
      "AdditionalPropertiesClass" : {
        "type" : "object",
        "properties" : {
          "map_property" : {
            "type" : "object",
            "additionalProperties" : {
              "type" : "string"
            }
          },
          "map_of_map_property" : {
            "type" : "object",
            "additionalProperties" : {
              "type" : "object",
              "additionalProperties" : {
                "type" : "string"
              }
            }
          }
        }
      }
```
Note missing `additionalProperties`.

For my specific case, I am fine with using `additionalProperties: {}`, but that results in the same problematic behavior above where `object` is being treated as expected `dict`.

spacether: Ivan I am figuring out how to add this empty object -> all types.
This is also complicated by the fact that any type can have nullable set to true, which would mean that we could accept types like (string) (string OR None) or all the way up to (string, boolean, integer, float, list, dict, None).

For now I will leave nullable out because it is a 3.0 spec feature.
We still will have cases were "str|bool|int|float|list|dict" would be the type or you could have that nested as the innermost value in a nested list or dict schema, like:
dict(str, dict(str, str|bool|int|float|list|dict))

spacether: @wing328 @ivan-gomes I am still working on this. Please refrain from merging until I have updated this PR. My update will include the type=object additionalProperties feature.

ivan-gomes: Roger. Please let me know if there is anything I can do to help.

spacether: @ivan-gomes will do.
Still to do is:
- Move model utils to the model `__init__` file
  - recursive_type
  - valid_type
  - get_types_remainder
  - to_dict
- Add additionalproperties map dict so variable names can be transformed for serialization/deserialization
- Ensure serialization works w/ tests
- Ensure deserialization works w/ tests
- Have serialization use model.to_dict

spacether: @wing328 @ivan-gomes I am nearing completion on this task.
To get this feature to work I added typing of all parameters and discriminators using real python classes. Soon I will get this PR submitted.

spacether:
@wing328 how do you feel about the size of this PR?
It may be better to break it up into pieces. They could be:

1. python merge non-object models into endpoint codegenparameters and model codegenproperties (this is needed for type checking in PR 2 below) https://github.com/OpenAPITools/openapi-generator/issues/1991
2. python change types from strings to classes
3. python add aditionalproperties where type == {} allows all types

What do you think?

ivan-gomes:
Ran it against the API spec that caused me to create the issue, and it works great! Thanks so much for your effort.

My humble opinion is that the changes involved are quite interconnected and trying to retroactively decouple them into separate PRs would be significant work with little practical purpose.

spacether:
Per a discussion with @wing328 this PR is a bit big. I will break it up to smaller pieces for ease of review.

tomghyselinck:
Hi @spacether,

I have made some additional changes to make things work for (multi-level) inheritance:
https://github.com/tomghyselinck/openapi-generator/tree/python_additionalprops_option2

I use this in combination with #2170 to also make the `disciminator.mapping` work as expected.
I also made a small change on that one: See https://github.com/rienafairefr/openapi-generator/pull/2 (based on https://github.com/tomghyselinck/openapi-generator/tree/python-discriminator-map)

With best regards,
Tom.

spacether:
Thanks for doing that Tom. Can you explain how you can use vars rather than allvars in the model mustache file for allof models?
Do you mind if we incorporate your updates in a PR in or along with the above 5 smaller suggested PRs.

tomghyselinck:
Hi @spacether,

Thank you for your quick response.

#### Regarding the `allVars`:

A language-specific implementation of `DefaultCodegen` can set the `supportsInheritance`, (and eventually `supportsMultipleInheritance`) to `true` when it supports inheritance in the generated models.
In that case the properties from "_parent_" schemas are only added to the `allVars` (and not to the `vars`).

The `PythonClientCodegen` did set `supportsInheritance` to `true` so it needed to use the `allVars`.
IMHO this is not the intent of the `supportsInheritance` vs. `allVars` use.

Please note that I also discovered an issue in the use of `allVars`: `DefaultCodegen.fromModel` does not call `postProcessModelProperty` on the `allVars`.
This especially causes issues when `postProcessModelProperty` adds additional item to for example `CodeGroperty.vendorExtensions`.
In particular we had an issue where our regex pattern was not set (`vendorExtensions.x-regex` and `vendorExtensions.x-modifiers`)

I have not yet created a BUG report for the latter issue.

#### Incorpration

Please go ahead and incorporating my changes in the way that fits best for you 😉.

tomghyselinck:
Hi @spacether,

I just want to let you know that there is still a problem with deserializing `enum` models.
I have been working on it, but I did not find a proper solution yet.

I have a quick and dirty workaround:
https://github.com/tomghyselinck/openapi-generator/commit/40658caf5425997715d13a53adb3f63302f316a2

There are two alternative workarounds (or solutions?) in this commit:
1. Change the `dataType` to `(MyEnumType, str)` instead of `(MyEnumType, )`
1. Change the `dataType` to `(str, )` instead of `(MyEnumType, )` (_commented out_)

Both alternative changes trigger a warning `WARN  o.o.codegen.DefaultCodegen - allOf with multiple schemas defined. Using only the first one: <some_type>. To fully utilize allOf, please use $ref instead of inline schema definition`.
I have no clue why yet...

With best regards,
Tom.

spacether:
Thanks @tomghyselinck
Yes, I know about that. My above comment where we fold non object models into their referenced models and moving enum allowable values to the class level (the first time that I have mentioned this) should fix that.

tomghyselinck:
Thank you for the explanation.

(_I intended to ask if #1991 was the solution for this, but my comment was already in the air_ 🙂)

spacether:
Break out PR-statuses:

1. python merge non-object models into endpoint codegenparameters and model codegenproperties,
move enum allowable values into a dict at the class level
2. [SUBMITTED, tests pass] python create exceptions module where all exceptions will inherit from OpenAPiException https://github.com/OpenAPITools/openapi-generator/pull/2393
3. python change types from strings to classes + add type checking
4. [SUBMITTED, tests pass] add additionalProperties examples
5. python add aditionalproperties feature and aditionalproperties where type == {} which accepts all types https://github.com/OpenAPITools/openapi-generator/pull/2405

mverderese:
Hey @spacether! This is exactly the functionality I'm looking for in the codegen tools for the python client. I have a model defined with a couple of rigid properties, and then want to be able to support additional properties of a certain type. If that's not possible, then aliasing a `Dict[str, <type>]` as a Model would be great. I'm a little lost as to what the status of this PR is, so any clarity would be greatly appreciated! Also if you need any help getting this over the finish-line, let me know!

spacether:
Hey @mvrderese this PR is being broken up into smaller pieces so it is easier to review. I would welcome some help. If you want to collaborate please shoot me an email through the link at the bottom of my page here https://spacether.github.io/

ivan-gomes:
Hi @spacether. Can we get an update on where we are on this feature? It addresses the last blocker I have on migrating a project to openapi-generator.

spacether:
> Hi @spacether. Can we get an update on where we are on this feature? It addresses the last blocker I have on migrating a project to openapi-generator.

Hey @ivan-gomes I am still working on this.
The remaining tasks are:
- merge non-object models into models. I am working on it here: https://github.com/spacether/openapi-generator/tree/issue_1991_folds_non_object_models_into_models
  - STATUS: I am nearly ready to submit this and should open a PR by end of week at the latest
- convert data types to classes (this is a simpler PR than the non-object models one)
- add additionalproperties feature (this is a simpler PR than the non-object models one)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -402,18 +466,62 @@ public String modelTestFileFolder() {
...
+        if (typeMapping.containsKey(oasType)) {
+            return typeMapping.get(oasType);
+        }
+        return oasType;
+    }
+
+    public String getTypeString(Schema p, String prefix, String suffix) {
+        // this is used to set dataType, which defines a python tuple of classes
+        String typeSuffix = suffix;
+        if (suffix == ")") {
```
wing328: Please use `")".equals(suffix)` instead

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -402,18 +466,62 @@ public String modelTestFileFolder() {
...
+        } else if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return getSchemaType(p) + "[" + getTypeDeclaration(inner) + "]";
-        } else if (ModelUtils.isMapSchema(p)) {
-            Schema inner = ModelUtils.getAdditionalProperties(p);
+            return prefix + "[" + getTypeString(inner, "(", ")") + "]" + typeSuffix;
+        }
+        String baseType = getSimpleTypeDeclaration(p);
+        if (baseType == "file") {
```
wing328: Please use `"file".equals(baseType)` instead

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -402,18 +466,62 @@ public String modelTestFileFolder() {
...
-            return getSchemaType(p) + "(str, " + getTypeDeclaration(inner) + ")";
+    @Override
+    public String getTypeDeclaration(Schema p) {
+        // this is used to set dataType, which defines a python tuple of classes
+        return getTypeString(p, "", "");
+    }
+
+    @Override
+    public String toInstantiationType(Schema property) {
+        if (property instanceof ArraySchema || property instanceof MapSchema || property.getAdditionalProperties() != null) {
```
wing328: Please use ModelUtils.isArraySchema instead of `property instanceof ArraySchema`. Same for the map schema check using instanceof at the moment.

spacether:
Unable to get necessary reviews and approvals on this PR and feature break out PRs.

spacether:
This is something that I really wanted to complete.
Feedbackwwas that my original PR was tok big so I broke it into pueces. 
Feedback was then that I should put it in a new generator which I did. 
My last PR was open for a month ready for merge with pings asking for review. This feature branch represents 100+ hours of investment of my time. But if it is not reviewed and approved there is nothing I can do. It makes me feel bad when I have made such an investment (100+ hours of dev time) and the code is ignored for a month. The last commitment was a review by Friday and the review was not done.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
spacether: This code only exists to handle v2 file schemas, replacing this with ModelUtils.isFileSchema(p) in a new PR

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
spacether: Updated per feedback here https://github.com/spacether/openapi-generator/tree/issue_1991_uses_class_types

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
spacether: Implemented this feedback here https://github.com/spacether/openapi-generator/tree/issue_1991_uses_class_types

spacether:
The code in this PR was merged into master in the new generator python-experimental

PR #2059, opened 2019-02-05
Add `controller.service_arguments` tag to controllers to make them public

### LINKED ISSUE #1002: [Symfony] Make controllers explicitly public

ossinkine:
### PR checklist

### Description of the PR

Fixes #1002 
Continues #1003

### DIFF
--- a/modules/openapi-generator/src/main/resources/php-symfony/services.mustache
+++ b/modules/openapi-generator/src/main/resources/php-symfony/services.mustache
@@ -28,6 +28,7 @@ services:
          - [setSerializer, ['@{{bundleAlias}}.service.serializer']]
          - [setValidator,  ['@{{bundleAlias}}.service.validator']]
          - [setApiServer,  ['@{{bundleAlias}}.api.api_server']]
+        tags: ['controller.service_arguments']
 
 {{/operations}}
 {{/apis}}
--- a/samples/server/petstore/php-symfony/SymfonyBundle-php/Resources/config/services.yml
+++ b/samples/server/petstore/php-symfony/SymfonyBundle-php/Resources/config/services.yml
@@ -25,6 +25,7 @@ services:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]
          - [setApiServer,  ['@open_apiserver.api.api_server']]
+        tags: ['controller.service_arguments']
 
     open_apiserver.controller.store:
         class: OpenAPI\Server\Controller\StoreController
@@ -32,6 +33,7 @@ services:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]
          - [setApiServer,  ['@open_apiserver.api.api_server']]
+        tags: ['controller.service_arguments']
 
     open_apiserver.controller.user:
         class: OpenAPI\Server\Controller\UserController
@@ -39,4 +41,5 @@ services:
          - [setSerializer, ['@open_apiserver.service.serializer']]
          - [setValidator,  ['@open_apiserver.service.validator']]
          - [setApiServer,  ['@open_apiserver.api.api_server']]
+        tags: ['controller.service_arguments']
 


dkarlovi:
Note, this obsoletes #1003 since it solves the problem in a cleaner way.

wing328:
@dkarlovi thanks for reviewing the change.

cc @Addvilz (the author of #1003) as well

modules/openapi-generator/src/main/resources/php-symfony/composer.mustache
```diff
@@ -25,7 +25,7 @@
         "ext-mbstring": "*",
         "symfony/validator": "*",
         "jms/serializer-bundle": "*",
-        "symfony/framework-bundle": "^2.3|^3.0|^4.1"
```
Addvilz: Instead of removing 2.3 I would bump it to 2.7. It should work as well, and there is a ton of 2.x projects out there still. Plus 2.7 is still receiving security updates. No need to break compat as long as it costs nothing.

Addvilz: Instead of removing 2.3 I would bump it to 2.7. It should work as well, and there is a ton of 2.x projects out there still. Plus 2.7 is still receiving security updates. No need to break compat as long as it costs nothing.

dkarlovi:
@Addvilz supporting a version of Symfony which is no longer supported by Symfony themselves is stretching this project too thin IMO. Of course, it would be best to support every version imaginable, but it's just not possible in most cases. We should only support officially supported versions, not EOL'd ones.

Addvilz:
@dkarlovi Symfony 2.7 IS officially supported till May 2019. 2.8 till November 2019. 

My argument is, supporting these versions costs absolutely nothing (e.g. there is no overhead on the project) AND they are still officially supported. Clearly no need to support 2.3 which is long extinct, but a bump to 2.7 would be advisable.

dkarlovi:
@Addvilz 2.7 / 2.8 is in security maintenance only so **only critical security issues** will get any attention there, they no longer receive bugfixes, let alone new features. I'm not sure we can call that "supported".

The tag we're adding came out with Symfony 3.3 (which is also EOL) so we cannot really support 2.x with this patch at all.

Addvilz:
@dkarlovi it's not about what we call supported, it's what Symfony Release Process defines as supported. And security fix period falls under LTS support for versions I mentioned, and they are therefore supported. Users rely on these LTS support releases for that exact reason. I am aware of several large scale projects still in production on 2.8 specifically because 2.8 is LTS. 

Versions prior to 3.3 are not affected by this issue at all and compatibility is also not affected. Adding a tag will have no effect on everything prior 3.3 as controllers were public by default and this issue simply did not exist.

Again, support for these versions costs nothing at all to OpenApi, so I don't see any reason to not support them while possible. There is no requirement for active support here and nor requirement to maintain backwards compatibility. I don't see semantics as a very good reason to break other peoples code if not doing so has literally no side effects.

dkarlovi:
> Users rely on these LTS support releases for that exact reason. I am aware of several large scale projects still in production on 2.8 specifically because 2.8 is LTS.

Well, that's fair, they just use the older version of the generator?

> Versions prior to 3.3 are not affected by this issue at all and compatibility is also not affected. Adding a tag will have no effect on everything prior 3.3 as controllers were public by default and this issue simply did not exist.

That's true. :+1: 

> Again, support for these versions costs nothing at all to OpenApi, so I don't see any reason to not support them while possible.

The cost is any change, current or future, must consider all the supported versions. Supporting anything but the most recent versions always comes with a cost, which becomes higher the more elaborate your support is.

Having had experience with supporting 3 separate versions of Symfony in a quite used package where we wanted to introduce static code analysis: it's a nightmare.

I'd say a project core member must decide here what they wish to support. @wing328?

dkarlovi:
Thinking about this more, I agree with @Addvilz: this PR is not a place to do this change. Since it doesn't impact users of 2.x, we can keep it and just add the tag. Reducing the number of supported versions is a different problem which doesn't need to be solved here for this to be merged.

@ossinkine would you mind removing the composer changes?

ossinkine:
I've removed composer changes

wing328:
> I'd say a project core member must decide here what they wish to support. @wing328?

What about following Swift approach to create generator based on versions? e.g.

`swift` => swift 2.x (deprecated)
`swift3` => swift 3.x
`swift4` => swift 4.x

so we can keep the existing `php-symfony` generators for 2.x and 3.x. For 4.x, we create `php-symfony4` instead similar to what [Addvilz](/Addvilz) suggested

wing328:

PR #2072, opened 2019-02-06
Add "servers" support to the operation, path in PHP API client

wing328:
### PR checklist

### Description of the PR

An example of the change can be found in https://github.com/OpenAPITools/openapi-generator/pull/2072/commits/64bf0d48c0b51de9e4e5a2efde49332d8f8bb60e#diff-15582b49a62d7a5ecac56414cfe6407fR332

cc @jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09), @ybelenko (2018/07), @renepardon (2018/12)

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/api.mustache
+++ b/modules/openapi-generator/src/main/resources/php/api.mustache
@@ -55,7 +55,7 @@ use {{invokerPackage}}\ObjectSerializer;
     protected $headerSelector;
 
     /**
-     * @var Host index
+     * @var int Host index
      */
     protected $hostIndex;
 
@@ -63,7 +63,7 @@ use {{invokerPackage}}\ObjectSerializer;
      * @param ClientInterface $client
      * @param Configuration   $config
      * @param HeaderSelector  $selector
-     * @param int             $host_index
+     * @param int             $host_index (Optional) host index to select the list of hosts if defined in the OpenAPI spec
      */
     public function __construct(
         ClientInterface $client = null,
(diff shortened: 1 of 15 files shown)


modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -54,19 +54,47 @@ use {{invokerPackage}}\ObjectSerializer;
      */
     protected $headerSelector;
 
+    /**
+     * @var Host index
```
ackintosh: A 1st keyword of `@var` identifies the type of the property. so I think it would be `@var int Host index` .

samples/client/petstore/php/OpenAPIClient-php/lib/Api/AnotherFakeApi.php
```diff
@@ -64,19 +64,47 @@ class AnotherFakeApi
...
      * @param Configuration   $config
      * @param HeaderSelector  $selector
+     * @param int             $host_index
      */
     public function __construct(
         ClientInterface $client = null,
         Configuration $config = null,
-        HeaderSelector $selector = null
+        HeaderSelector $selector = null,
+        $host_index = 0
```
ackintosh: How about the `$host_index` generates only if `servers` exists in pathItems?

I just wondering that the `$host_index` is defined but unused. 👀

wing328: Yes, very good suggestion. My intention is to make the constructor consistent across different API classes.

What about updating the doc string to explain it cleary that it's optional and only needed when the endpoint has multiple servers/hosts defined?

modules/openapi-generator/src/main/resources/php/api.mustache
wing328: 👍  I'll update it in the next push

samples/client/petstore/php/OpenAPIClient-php/lib/Api/AnotherFakeApi.php
ackintosh: I agree with that. 👍

wing328: Pushed some enhancements based on your feedback.

PR #2107, opened 2019-02-10
Don't escape [ and ] in query param keys

BastiOfBerlin:
rails' default naming for arrays in query parameters is name[]=...&name[]=...
It doesn't recognize the escaped forms %5B + %5D but fails silently

### PR checklist

### Description of the PR

Adapting the solution [here](https://stackoverflow.com/a/48307962) to prevent HttpParameterCodec from encoding '[' and ']' in query param keys.

This is important for using Rails as a backend as the square brackets are added to the parameter names as per default.

I suppose this change doesn't interfere with most people's APIs as square brackets are not common suffixes other to arrays. Encoding them on the other hand is not necessary for URIs and users could still do this if required for their specific project.

However, if this still considered to be a breaking change, I suppose introducing an additional parameter such as 'rails-compat-query-array-params'.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-angular/encoder.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/encoder.mustache
@@ -14,7 +14,9 @@
 export class CustomHttpUrlEncodingCodec extends HttpUrlEncodingCodec {
     encodeKey(k: string): string {
         k = super.encodeKey(k);
-        return k.replace(/\+/gi, '%2B');
+        return k.replace(/\+/gi, '%2B')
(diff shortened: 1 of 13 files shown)


wing328:
@BastiOfBerlin I don't think it's a breaking change but it's fine to include breaking changes in the upcoming 4.0.0 release.

wing328:
I'll update the samples in a separate commit.

macjohnny:
@BastiOfBerlin @wing328 I think we should revert this because square brackets are reserved characters, see https://stackoverflow.com/a/39294315 for a good explanation.
PHP e.g. accepts both, unencoded and encoded square brackets as array indicator

PR #2112, opened 2019-02-11
handle composed schemas in InlineModelResolver

rienafairefr:
### PR checklist

### Description of the PR

`InlineModelResolver` wasn't handling the composedschema case correctly or even at all, with this PR it can transform inline models like this:

```    
ComposedModel:
  allOf:
  - $ref: '#/components/schemas/Model'
  - type: object
    properties:
      prop:
       type: string
```
into
```    
ComposedModel:
  allOf:
  - $ref: '#/components/schemas/Model'
  - $ref: '#/components/schemas/ComposedModel_allOf'
ComposedModel_allOf:
  type: object
    properties:
      prop:
        type: string
```
this might help for handling correctly allOf/anyOf/oneOf, even in the case of the presence of inline model


ackintosh:
@rienafairefr Could you please merge master to resolve the conflict?

wing328:
UPDATE: add new files via https://github.com/OpenAPITools/openapi-generator/pull/2112/commits/e29cf137be18cf03aa24b8d1de750a997e502409

cvgaviao:
Could someone explain me the benefits of creating a new synthetic schema into the openapi model when using a no-multiple-inheritance languages, as java? 
Maybe I'm wrong, but it will just create two practically equal classes with different names and practically will doubled the size of the api jar.

chameleon82:
@richardwhiuk  Is any way to not generate `allOf` models as those models not used in Api and there is no benefit to have it in most projects?

wing328:
There's a bug in this implementation when "type: object" is not specified in the inline schema definition and I've filed https://github.com/OpenAPITools/openapi-generator/pull/5992 to fix it. The local test result is good.

wing328:
To skip those model generation of inline schema in allOf, please add those models to .openapi-generator-ignore so that these files won't be generated.

wing328:
@rienafairefr first of all thanks for the PR but we've feedbacks from our users that they do prefer the old behavior to not generate the inline model separately.

I'll file a PR to roll back the change. Let me know if you've any concerns, especially regarding your use case.

PR #2121, opened 2019-02-11
[Python] handle oneOf/allOf/anyOf composed schemas in the Python client

rienafairefr:
### PR checklist

@wing328 @ci

### Description of the PR

I saw that `allOf`/`oneOf`/`anyOf` was not yet fully handled in general, and in particular for my use case, in the Python client codegen. This PR is a tentative to handle deserialization of models that use oneOf/anyOf/allOf, by making use of the `allOf`/`oneOf`/`anyOf` fields in `CodegenModel`

I've added a `composed_hierarchy` to the model, which is specified when the underlying openapi model uses oneOf/allOf/anyOf, if there is no discriminator field. When receiving some data, to deserialize it to the correct class, we try to match the incoming data to all possible subclasse
- oneOf: we raise an exception if 0 or more than one of them match
- allOf: we raise an exception if not all of them match
- anyOf: we raise an exception if 0 of them match

with oneOf/anyOf, the instance that is returned will be of the correct subclass, and with allOf, we return an instance of the current model class.

The case of a more advanced schema (when more than one of  `allOf`/`oneOf`/`anyOf` is used) is **not** handled. 
Caveat: cases of inline models might not be handled properly in some or all of the cases.


tomghyselinck:
Hi @rienafairefr,

Thank you very much for this PR!

I have been doing some tests on your code.

It seems to work on basic situations, but I still have some troubles with special/extended cases.

The final class implementation still seems to miss some attributes of the base class when using (multi-level) inheritance.
In the example below `MyImpl` will miss the properties `implType`, `sharedData` and `moreSharedData` from `BaseClass` resp. `SubClass`.

Also when referencing `SubClass` in the OpenAPI yaml file as operation `response` and `request` body it won't be able to properly discriminate. At least one reason is that the required property of the base class (`implType` from `BaseClass` in our example) is missing. In particular for the `request` body, it will be the server implementation which will fail because the `implType` property is not given to the server.

### Python server comparison

Please note that the Python server (`python-flask`) implementation defines similar inheritance on `allOf`/`oneOf`/`anyOf` which seems to solve many of these issues.

This implementation _also_ avoids issues with `oneOf` of which the items has `required` fields (it assigns `self._some_attribute` instead of `self.some_attribute` (the latter fails because it's a `property` which does some checks on the value which is being set).

It there a possibility to sync the client implementation with the server implementation regarding model definitions?
@wing328: I read that the Python server and client implementations will converge some time? Is that correct? I personally believe that would be a very good approach.

### Another final note

The "_child_" (or "_final_") implementations in the Python models (either client or server) could give a proper default value to the `discriminator` property. (when one of their parent(s) has a `discriminator`)
   For example:
   * `SubClass.__init__(implType='SubClass', ...)` instead of `SubClass.__init__(implType=None, ...)`
   * `MyImpl.__init__(implType='MyImpl', ...)` instead of `MyImpl.__init__(implType=None, ...)`

This would be very convenient for users of the API so they won't need to provide that field (_which is mostly the same as the class name_). Especially when they need to provide a new item to a Python function (i.e. REST operation)

### Example (_stripped very much_):
```yaml
BaseClass:
  discriminator:
    propertyName: implType
  properties:
    implType:
      type: string
    sharedData:
      type: string
SubClass:
  discriminator:
    propertyName: implType
  allOf:
  - $ref: BaseClass
  - properties:
      sharedData:
        type: string
MyImpl:
  allOf:
  - $ref: SubClass
  - properties:
      implData:
        type: string
```

With best regards,
Tom.

rienafairefr:
Thanks for the review @tomghyselinck 
You mentioned the python-flask implementation but I couldn't find where the allOf/oneOf/anyOf are handled there. I'd gladly use the python-flask models implementation and sync up, for sure :+1:
I sent my inheritance-using spec to `python-flask` and couldn't find a place where a deserialization from json would give us a "child" concrete type instead of a parent type, maybe I just didn't look at the right places ^^

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -640,6 +642,60 @@ class ApiClient(object):
 
         instance = klass(**kwargs)
```
tomghyselinck: Hi @rienafairefr,

I did some further investigation on the issues regarding (multi-level) inheritance.
I hope the following helps you:

1. First of all with your PR we should also apply the following changes on `PythonClientCodegen.java`.
   I suppose this should be correct since the Python client code does not support real inheritance at Python level, but it "_copies_" the _attributes_ to the child classes.

```diff
diff --git a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
index aa6e8ad015..6fc0066264 100644
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
@@ -57,7 +57,7 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
         // at the moment
         importMapping.clear();

-        supportsInheritance = true;
+        //supportsInheritance = true;
         modelPackage = "models";
         apiPackage = "api";
         outputFolder = "generated-code" + File.separatorChar + "python";
```
2. Secondly, when the "_base_" classes (when using `allOf`) or the "_selection_" classes (when using `oneOf`/`anyOf`) have a `discriminator`, we should try that one first. If that one fails, we can use your "_implicit_" hierarchy code.

   The current code especially fails when:

   1. we have a `responseBody` which returns a `BaseClass`
   1. with an object structure as below

   In that case the `if hierarchy['allOf']:` will return an instance of the `BaseClass` instead of an instance of `SomeChild`.

```yaml
schemas:
  RemoteId:
    properties:
      id:
        type: int
  BaseClass:
    discriminator:
      propertyName: implType
    allOf:
    - $ref: RemoteId
    - properties:
        implType:
          type: string
  SomeChild:
    allOf:
    - $ref: BaseClass
    - properties:
        childData:
          type: string
```

(_I would liked to make the suggestion on multiple lines (till the end of the file), but I did not succeed in it, sorry for that_ 😕)
```suggestion
        instance = klass(**kwargs)

        if hasattr(instance, 'get_real_child_model'):
            klass_name = instance.get_real_child_model(data)
            if klass_name:
                instance = self.__deserialize(data, klass_name)

        if hasattr(instance, 'composed_hierarchy'):
            hierarchy = instance.composed_hierarchy
            if sum([1 if v else 0 for v in hierarchy.values()]) > 1:
                # ignore implicit AllOf
                return instance
            def error_msg(composed):
                return "Failed to parse `{0}` as {1} ({2} {3})" \
                    .format(data, klass,
                            composed,
                            ' '.join(hierarchy[composed]))
            if hierarchy['allOf']:
                matches = []
                for sub_klass in hierarchy['allOf']:
                    try:
                        if sub_klass in self.__deserializer_track:
                            # seen it before, break the recursion
                            matches.append(None)
                        else:
                            matches.append(self.__deserialize(data, sub_klass))
                    except:  # noqa: E722
                        pass
                if len(matches) == len(hierarchy['allOf']):
                    self.__deserializer_track = []
                    return instance
                # not all matched -> error
                raise rest.ApiException(status=0, reason=error_msg('allOf'))
            if hierarchy['anyOf']:
                for sub_klass in hierarchy['anyOf']:
                    try:
                        # if at least one of the sub-class matches
                        # terminate with an instance of this class
                        return self.__deserialize(data, sub_klass)
                    except:  # noqa: E722
                        pass
                # none matched -> error
                raise rest.ApiException(status=0, reason=error_msg('anyOf'))
            if hierarchy['oneOf']:
                matches = []
                for sub_klass in hierarchy['oneOf']:
                    try:
                        matches.append(self.__deserialize(data, sub_klass))
                    except:  # noqa: E722
                        pass
                if len(matches) == 1:
                    return matches[0]
                # none matched, or more than one matched -> error
                raise rest.ApiException(status=0, reason=error_msg('oneOf'))

        return instance
```

tomghyselinck:
Hi @rienafairefr,

Are you aware of #2049 which should also fix the support for `allOf`?
Does this also fix your needs for `oneOf`` / `anyOf`?

With best regards,
Tom.

rienafairefr:
Hi @tomghyselinck, I looked at #2049 but I don't think the cases of anyOf/oneOf are handled inthere :-|

modules/openapi-generator/src/main/resources/python/api_client.mustache
rienafairefr: Thanks for the comment, yeah I guess this goes to show that the idea of handling the allOf/oneOf/anyOf at a single level inside a model is bound to fail, because some information about how to deserialize a given file resides not only in the model's, but also in the model's children and parent, we have to go deeper somehow, but not too deeper otherwise we become stuck in recursion. Not sure how to proceed

spacether:
@rienafairefr can you describe what your code does for the case of the Dog class?
It looks like it attempts to instantiate all child classes and returns an exception if the instantiation fails, is that right?
How do we ensure that the instance of Dog will have the variables defined in the Dog class and the child Animal class? We could accomplish this by using allVars.
@tomghyselinck do you have an alternative suggestion to using allVars?
One solution would be to make a class Dog which inherited from allOf classes.
Not sure how one would handle intersecting classes, like two classes that both include the property red. That solution would probably not work for more complex cases.

### How about the below different proposal?
Validation of explicit properties is handled in our new composed class, just like normal models that we define today.
So for the MixedPropertiesAndAdditionalPropertiesClass class, it will validate and define the below properties in the class:
- uuid
- datatype
- map
However, it also needs properties added from child classes, specifically the Animal class.
Some of those could be required, some of those could be optional.
And if a property's value is changed, it should pass the required validation in the child class.
What if we store the child instances in the composed model?
So we would store an instance of Animal class in the instance of MixedPropertiesAndAdditionalPropertiesClass.
Any serialization of MixedPropertiesAndAdditionalPropertiesClass would create the dict using the properties in the current class:
- uuid
- datatype
- map
And merge in the dict representations of the child classes too.
This method also lets use preserve child class validation.
If a property in MixedPropertiesAndAdditionalPropertiesClass is changed, change the value in the corresponding child instances so we preserve their validation.

spacether:
One more use case that I'm not sure it this PR covers:
If we input prop_name='some value' into a model and prop_name does not exist in the composed class nor its child classes are we raising an exception? Where?

modules/openapi-generator/src/main/resources/python/api_client.mustache
spacether: What information is stored in the model parent?

spacether:
@rienafairefr the feature looks like it's almost there. Can you add tests to this PR that demonstrate instantiation and property access of Dog or other composed schemas?

rienafairefr:
@spacether Json schemas encompass more stuff than just inheritance like in the oop world, so even though it's tempting, we can't really do only that.

for allOf, yeah, we could have instances of the 'child' classes, and delegate to those. For anyOf or oneOf it would be possible too, but a bit more useless because anyOf or oneOf are instance of only one of the sub-classes. 
In the current situation, properties from the composed child schemas are dumped in the same place, maybe there is something to do here. We'd have to kinda keep track of where each model property comes from. 

There is already a deserialization of Dog instances (e.g. test_deserialize_dict_str_dog ), but it's using the disciminator thing. The standard petstore spec doesn't have anyOf or oneOf. I'll add that to the `3_0/petstore-with-fake-endpoints-models-for-testing.yaml` spec and add a unit test in the openapi3/petstore python client

spacether:
@rienafairefr thank you for explaining how oop inheritance won't work here.

Doesn't anyOf allow for one or more child schema matches? That's what this source describes: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
If dog only returns an animal animal instance without also setting breed on the dog instance, isn't it incorrectly building the dog instance?

tomghyselinck:
> @tomghyselinck do you have an alternative suggestion to using allVars?
> One solution would be to make a class Dog which inherited from allOf classes.
> Not sure how one would handle intersecting classes, like two classes that both include the property red. The inheritance solution may not work for more complex composed cases.

Hi @spacether,

Sorry for the late reply.

What do you mean with alternative to using `allVars`? AFAIK this PR uses `vars` and not `allVars`?

spacether:
> > @tomghyselinck do you have an alternative suggestion to using allVars?
> > One solution would be to make a class Dog which inherited from allOf classes.
> > Not sure how one would handle intersecting classes, like two classes that both include the property red. The inheritance solution may not work for more complex composed cases.
> 
> Hi @spacether,
> 
> Sorry for the late reply.
> 
> What do you mean with alternative to using `allVars`? AFAIK this PR uses `vars` and not `allVars`?

Hi @rienafairefr and @tomghyselinck I want to make sure that this implementation covers the case that I described above for accessing all properties on a model instance that has composed schema.
If we switch to allVars when using allOf I think all of those properties will be accessible, but using allVars is not the only way to do this. Let's look at a current example, how it is currently working and what the desired behavior is in my opinion.

Looking at Dog and Cat:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1266
This uses allOf
```
  Dog:
    allOf:
      - $ref: '#/definitions/Animal'
      - type: object
        properties:
          breed:
            type: string
  Cat:
    allOf:
      - $ref: '#/definitions/Animal'
      - type: object
        properties:
          declawed:
            type: boolean
  Animal:
    type: object
    discriminator: className
    required:
      - className
    properties:
      className:
        type: string
      color:
        type: string
        default: 'red'
```
A Dog instance should have the properties:
- breed (optional)
- className (required)
- color (optional)
A Cat instance should have the properties:
- declawed (optional)
- className (required)
- color (optional)
Only className is required per allOf:
https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
The allOf composed schema is part of Swagger 2.0 spec.
Looking at our tests, I don't see any tests for instance properties on Dog and Cat instances.
Can we add them?
- https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/test/test_cat.py (missing any tests)
  - 
We do test Dog 
- here https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/tests/test_deserialization.py
But the test is not checking to see that breed, className, and color are present

What do you think? Does what I've described make sense?

spacether:
oneOf/anyOf/allOf composed models have been added in the python-experimental generator in this diff: https://github.com/OpenAPITools/openapi-generator/pull/4446
That generator inherrits much of its code from the python generator and should be a drop in replacement for devs using the python generator.

rienafairefr:
Thanks @spacether I'll test out he python-experimental generator, this PR can probably be closed

spacether:
@rienafairefr thank you for this PR.
I am working on cleaning up old python PRs.
What would you like to do with this one; work on integrating it or closing it?

rienafairefr:
I think the `python-experimental` generator handled that issue, so I recommend closing this PR

PR #2128, opened 2019-02-12
Revert "Don't escape [ and ] in query param keys"

macjohnny:
Reverts OpenAPITools/openapi-generator#2107


macjohnny:
cc @BastiOfBerlin @wing328 @TiFu @JohannesHoppe

macjohnny:
see also https://stackoverflow.com/a/39294315 and https://bugzilla.mozilla.org/show_bug.cgi?id=473822

Issue #855, opened 2018-08-21
[DefaultGenerator] NullPointerException on code generation


hl191:
##### Description

Generating Code for yaml produces Nullpoint exception

```
Exception: null
    at org.openapitools.codegen.DefaultGenerator.processOperation (DefaultGenerator.java:941)
    at org.openapitools.codegen.DefaultGenerator.processPaths (DefaultGenerator.java:835)
    at org.openapitools.codegen.DefaultGenerator.generateApis (DefaultGenerator.java:462)
    at org.openapitools.codegen.DefaultGenerator.generate (DefaultGenerator.java:793)
    at org.openapitools.codegen.plugin.CodeGenMojo.execute (CodeGenMojo.java:567)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:956)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:290)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:194)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
[...]

PR #2133, opened 2019-02-12
[core] Fix NullPointer when schema is not set in header

jmini:
### PR checklist

- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @OpenAPITools/generator-core-team 

### Description of the PR

Fix a null pointer reported in https://github.com/OpenAPITools/openapi-generator/issues/855#issuecomment-462730420


jmini:
The result of the PR is that we have following imports:

```java
import io.swagger.jaxrs.*;
import io.swagger.jaxrs.PATCH;

import javax.ws.rs.*;
```

I didn't know that this is allowed.

I think I would prefer a solution where we remove the 2 lines `import io.swagger.jaxrs...;` and use the fully qualified name of the annotation like `@io.swagger.jaxrs.PATCH` in the generated code.

What do you think?

PR #1367, opened 2018-11-04
[typescript-fetch] Add support for File (application/octet-stream) response types

spallister:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.
@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10)

### Description of the PR

This is a copy of a similar PR I submitted to the swagger-codgen repo [here](https://github.com/swagger-api/swagger-codegen/pull/8410).  A while back, @wing328 reached out to let me know the majority of the community moved over here, where I might want to consider re-submitting my PR.

The purpose of this PR is to return `application/octet-stream` response types as `File`s, instead of processing them as JSON.  In the PR I had submitted to the swagger-codegen repo, I was returning such response types as `Blob`s, but soon after had a need to obtain the file's name (i.e. which `File` supports, but `Blob` does not).

Note that this PR includes a change for `DefaultGenerator.java`, as I needed to add a property to the mustache template model to determine whether the API contained at least one 'File Response'-type Operation.  In particular, this was needed so that I could conditionally include a `responseToFile` helper method within `api.mustache` as necessary (i.e. our TS linter would otherwise throw an unused method warning for applications where no 'File Response'-type Operations were in use).

It looks like there is a TypeScript-fetch rewrite taking place over at #569, but figured I'd submit this in case it can be merged in the interim, or incorporated into that rewrite.

Fixes #531

Issue #440, opened 2018-07-03
[Python] Client code is wrong when an endpoint accepts multiple MediaTypes


raubel:
##### Description

My OpenApi 3.0 specification contains an endpoint for which the response media type can be `application/octet-stream` or, if the query cannot be completed, `application/json`.
The generated Python client code builds a query with a MediaType being the concatenation of both: `application/octet-streamapplication/json`

##### openapi-generator version

3.0.3

##### OpenAPI declaration file content or url

Below is my (simplified) OpenApi specification:
```yaml
openapi: 3.0.0

servers:
- url: http://localhost/myapi

info:
  version: "1.0"
  title: My API

security:
- basicAuth: []

paths:

  /download:
    get:
      operationId: download
      responses:
        200:
          description: OK
          content:
            application/octet-stream:
              schema:
                type: string
                format: binary
        500:
          description: Error
          content:
            application/json:
              schema:
                type: string
```

##### Command line used for generation

I call the generator directly from my Java application:
```java
	CodegenConfigurator configurator = new CodegenConfigurator();
	configurator.setInputSpec("spec.yml");
	configurator.setGeneratorName("python");
	configurator.setOutputDir("out");
	new DefaultGenerator().opts(configurator.toClientOptInput()).generate();
```

##### Python test file

Running this script with the generated `openapi_client`
```python
import unittest
import openapi_client

configuration = openapi_client.Configuration()
configuration.debug = True

# create an instance of the API class
api_instance = openapi_client.DefaultApi(openapi_client.ApiClient(configuration))

class TestModels(unittest.TestCase):
    def test_download(self):
        api_instance.download()

if __name__ == '__main__':
    unittest.main()
```

results in the following output:
```
2018-07-02 15:17:38,841 DEBUG Starting new HTTP connection (1): localhost
send: 'GET /myapi HTTP/1.1\r\nHost: localhost:10080\r\nAccept-Encoding: identity\r\nContent-Type: application/json\r\nAccept: application/octet-streamapplication/json\r\nAuthorization: Basic c2VtYWRtaW46c2VtYWRtaW4=\r\nUser-Agent: OpenAPI-Generator/1.0.0/python\r\n\r\n'
reply: 'HTTP/1.1 406 Not Acceptable\r\n'
...
```

ackintosh:
@raubel Thanks for reporting the issue!

[...]

Issue #206, opened 2018-06-02
[Python] Skip utf8 decoding for specific Content-Types in py3


g-bon:
In rest.py if PY3 is used and _preload_content is True, the response content is always decoded as utf8. This doesn't make sense for non-text files, e.g. an api that returns a file.
 
```
if six.PY3:
    r.data = r.data.decode('utf8')
```

https://github.com/OpenAPITools/openapi-generator/blob/b5942624d6714f38c0bde9ae9c7400df1d48b1ed/modules/openapi-generator/src/main/resources/python/rest.mustache#L213

I welcome any suggestion on how to tackle this but the two solution that come to mind are:
- Check on the content_type and make a whitelist/blacklist of content types that should be decoded to utf8.

A simple example would be something like 

```
if six.PY3 and headers.get('Content-Type') not in ['application/octet-stream']:
    r.data = r.data.decode('utf8')
```

Where obviously the list of Content-Types to "skip" would be a meaningful one and not only octet-stream.

- Another, harder approach would be to infer the content type from the actual data but feels fragile and overkill to me. Positive side would be that it would likely do the right thing even if the wrong or no content-type is specified in the response headers.

Happy to work on this when I have some confirmation the problem is here / feedback on the right approach.

gierschv:
Same issue here (FlatIO/api-client-python#1), I believe a whitelist of mime-types could be helpful here.

gierschv:
No, I didn't have time to work on this, I found this issue by looking into another bug.

Another idea for the whitelist: we could have `application/octet-stream` whitelisted by default and add a vendor param in the spec to add additional mime types (e.g. `x-disable-decoding-content-types: [...]`). In my use case, I have at least MIDI/MP3/WAV content types that can be returned by some endpoints.

rienafairefr:
I've pushed a PR that I think addresses this issue, #2134

spacether:
Another solution would be to pass the spec required response_type through and if it is of type file, skip the decoding. 

My preference is for a whitelist or blacklist of Content-Types though.

Just a heads up, we should also handle the Content-Disposition header as that indicates a file to be downloaded. per: https://github.com/OpenAPITools/openapi-generator/blob/b5942624d6714f38c0bde9ae9c7400df1d48b1ed/modules/openapi-generator/src/main/resources/python/api_client.mustache#L528 in api_client.mustache def __deserialize_file it is used to extract the filename to save.

spacether:
[...]

PR #2134, opened 2019-02-12
[core][Python] rest client: handle multiple response types

rienafairefr:
### PR checklist

### Description of the PR

In OAS 3 it's possible to return multiple content-type, and for now the rest clients in the different languages don't support that, e.g. problems or related PRs: #206 #1367 #440
I've added returnType `returnType` field to the CodegenResponse class, this way we can get a mapping between the response code and the expected return type of that response, and I've implemented that logic in the Python client, through a `response_types` parameter dictionary. The `Accept` header is modified to include all the cases that the client expects, not just `application/json`.
The REST response is UTF-8 decoded if needed (not decoded in case of a `file` type), and deserialized according to the response type expected for the HTTP status code that is encountered.

Tested working on a API endpoint that returns a binary file on HTTP 200, and can also send a JSON with an error message for HTTP 401/403/500.


modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -160,20 +163,24 @@ class ApiClient(object):
             # use server/host defined in path or operation instead
             url = _host + resource_path
 
+        _decode_utf8 = {k: v != 'file' for k,v in six.iteritems(response_types)}
```
tomghyselinck: `... for k,v in ...` causes build error on Travis CI, it should be:
```suggestion
        _decode_utf8 = {k: v != 'file' for k, v in six.iteritems(response_types)}
```

rienafairefr: Thanks

spacether:
Hi @rienafairefr, can you fix the Travis-CI errors? It looks like your 
modules/openapi-generator/src/main/resources/python/asyncio/rest.mustache is missing `import six` at the top.

spacether:
@rienafairefr why not remove the response_type variable and instead pass your response_types dict only?
You are passing the same data in both locations. What do other people think? @wing328

rienafairefr:
@spacether thanks I pushed the missing `import six`.
I think you're right about removing response_type, I'm pushing that as well, hopefully CI will not blow up ^^

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2395,6 +2395,14 @@ public CodegenOperation fromOperation(String path,
             if (methodResponse != null) {
                 Schema responseSchema = ModelUtils.unaliasSchema(this.openAPI, ModelUtils.getSchemaFromResponse(methodResponse));
 
+                for (CodegenResponse response : op.responses){
+                    Schema thisResponseSchema = (Schema) response.schema;
+                    if (thisResponseSchema != null) {
+                        CodegenProperty cm = fromProperty("response", thisResponseSchema);
+                        response.dataType = cm.dataType;
```
spacether: @rienafairefr why are you setting the response's dataType equal to the CodegenProperty dataType here? What is incorrect about how fromResponse is setting response.dataType?

modules/openapi-generator/src/main/resources/python/api.mustache
```diff
@@ -210,6 +210,13 @@ class {{classname}}(object):
         # Authentication setting
         auth_settings = [{{#authMethods}}'{{name}}'{{#hasMore}}, {{/hasMore}}{{/authMethods}}]  # noqa: E501
 
+        # multiple potential response types
+        response_types = {
+        {{#responses}}
+            {{code}}: {{#dataType}}'{{dataType}}'{{/dataType}}{{^dataType}}None{{/dataType}}{{#hasMore}},{{/hasMore}}
```
spacether: Can we use baseType here instead of dataType?
It looks like dataType is used for primitive types. It looks like baseType is used for model names per here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2678
Then the baseType is used to import the model name here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2382

spacether:
@rienafairefr what's the status of this PR?
The work that you have done here is great and would be very useful to the end users + clients. 
if you resolve the merge conflicts and the open questions I can give this a review now that I am on the review committee.

spacether:
@rienafairefr thank you for this PR.
I am working on cleaning up old Python PRs.
This feature looks very helpful as it lets the users ingest different models depending upon the response code sent back.
@rienafairefr what's the status of this PR?
Are you interested in updating and merging it or would you rather it be closed?

rienafairefr:
I had need for that at some point but not anymore so I guess the PR staled. The feature is still useful IMO, I'll try to look at it

spacether:
This PR has been closed due to inactivity

PR #2136, opened 2019-02-12
[rust-server] Changed query parameters to be url encoded

Herrera93:
### PR checklist

### Description of the PR

Based on this [issue](https://github.com/servo/rust-url/issues/416) the `utf8_percent_encode` doesn't percent encode `+` instead treats it as a space which is may cause a problem if the query parameters contain one. For example if we wanted to use a DateTime range in query parameters:

`{base_path}{path}?start=2019-02-12T00:00:00+00:00&end=2019-02-13T15:00:00+00:00`

These parameters may be decoded as:

```
start = 2019-02-12T00:00:00 00:00
end = 2019-02-13T15:00:00 00:00
```

To solve this I changed query parameters to be form-urlencoded using `url::form_urlencoded` based on what's done in [`rust` client generator](https://github.com/OpenAPITools/openapi-generator/blob/ea08106c809c17a660c9180210e12ae02c2d5914/modules/openapi-generator/src/main/resources/rust/request.rs#L106).

### DIFF
--- a/modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
+++ b/modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
@@ -25,6 +25,7 @@ use std::path::Path;
 use std::sync::Arc;
 use std::str;
 use std::str::FromStr;
+use std::string::ToString;
 
 use mimetypes;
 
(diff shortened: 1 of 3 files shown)


kanekv:
@bjgill can you take a look, please?

modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
```diff
@@ -246,18 +247,23 @@ impl<F, C> Api<C> for Client<F> where
...
+        let mut query_string = self::url::form_urlencoded::Serializer::new("".to_owned());
 {{#queryParams}}{{#-first}}
         // Query parameters
-{{/-first}}{{#required}}        let query_{{{paramName}}} = format!("{{{baseName}}}={{=<% %>=}}{<% paramName %>}<%={{ }}=%>&", {{{paramName}}}=param_{{{paramName}}}{{#isListContainer}}.join(","){{/isListContainer}}{{^isListContainer}}.to_string(){{/isListContainer}});
-{{/required}}{{^required}}        let query_{{{paramName}}} = param_{{{paramName}}}.map_or_else(String::new, |query| format!("{{{baseName}}}={{=<% %>=}}{<% paramName %>}<%={{ }}=%>&", {{{paramName}}}=query{{#isListContainer}}.join(","){{/isListContainer}}{{^isListContainer}}.to_string(){{/isListContainer}}));
+{{/-first}}{{#required}}        query_string.append_pair("{{{baseName}}}", &param_{{{paramName}}}{{#isListContainer}}.join(","){{/isListContainer}}{{^isListContainer}}.to_string(){{/isListContainer}});
+{{/required}}{{^required}}        if let Some({{{paramName}}}) = param_{{{paramName}}} {
```
bjgill: This appears to be a change in behaviour. Previously, if we would treat `None` the same as `""`. For example - `foo=bar&baz=&...`.

The new behaviour is to omit optional parameters with a value of `None`. For example - `foo=bar&...`.

Is this intentional? We could probably simplify some of the logic below if we reverted to the old behaviour. I've not used form parameters myself, so am happy to keep the new behaviour if it's more usual.

modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
```diff
@@ -246,18 +247,23 @@ impl<F, C> Api<C> for Client<F> where
     C: Has<XSpanIdString> {{#hasAuthMethods}}+ Has<Option<AuthData>>{{/hasAuthMethods}}{
 {{#apiInfo}}{{#apis}}{{#operations}}{{#operation}}
     fn {{#vendorExtensions}}{{{operation_id}}}{{/vendorExtensions}}(&self{{#allParams}}, param_{{{paramName}}}: {{^required}}Option<{{/required}}{{#isListContainer}}&{{/isListContainer}}{{{dataType}}}{{^required}}>{{/required}}{{/allParams}}, context: &C) -> Box<Future<Item={{{operationId}}}Response, Error=ApiError>> {
+        let mut query_string = self::url::form_urlencoded::Serializer::new("".to_owned());
 {{#queryParams}}{{#-first}}
         // Query parameters
```
bjgill: Could we perhaps move the `let mut query_string = ...` line inside the `{{#queryParams}}{{#-first}}` tags? That would allow us to remove the checks below for `query_string_str != ""` in the case where we don't have any query strings.

modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
```diff
@@ -246,18 +247,23 @@ impl<F, C> Api<C> for Client<F> where
...
-            "{}{{{basePathWithoutHost}}}{{path}}{{#queryParams}}{{#-first}}?{{/-first}}{{=<% %>=}}{<% paramName %>}<%={{ }}=%>{{/queryParams}}",
-            self.base_path{{#pathParams}}, {{{baseName}}}=utf8_percent_encode(&param_{{{paramName}}}.to_string(), ID_ENCODE_SET){{/pathParams}}{{#queryParams}},
-            {{{paramName}}}=utf8_percent_encode(&query_{{{paramName}}}, QUERY_ENCODE_SET){{/queryParams}}
+        let mut uri = format!(
+            "{}{{{basePathWithoutHost}}}{{path}}",
+            self.base_path{{#pathParams}}, {{{baseName}}}=utf8_percent_encode(&param_{{{paramName}}}.to_string(), ID_ENCODE_SET){{/pathParams}}
         );
-
+        let query_string_str = query_string.finish();
+        if query_string_str != "" {
```
bjgill: For future reference, you might be interested in the [String::is_empty method](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.is_empty-1)

modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache
Herrera93: @bjgill From what I've seen optional query parameters are usually handled in a present/not present way instead of empty/non-empty value.

But either way is valid based on [URI spec](https://tools.ietf.org/html/rfc3986#section-3.4), do you think we should change it to empty/non-empty value ?

kanekv: @bjgill I'd vote for removing it. If None is treated as "", there is no semantics to omit the value. If people want to have empty strings they can explicitly provide Some("").

bjgill: An excellent point, @kanekv. The new behaviour does seem most sensible, then.

Herrera93:
@bjgill Thanks for the feedback, I push some changes where I moved the `let mut query_string..` to the `queryParams` block and removed the check if there aren't any query parameters

Herrera93:
@bjgill There seems to be an error in one of the builds, based on the logs it looks like a `Kotlin` problem:

```
* What went wrong:
Could not determine the dependencies of task ':compileKotlin'.
> Could not resolve all files for configuration ':kotlinCompilerClasspath'.
   > Could not download kotlin-compiler-embeddable.jar (org.jetbrains.kotlin:kotlin-compiler-embeddable:1.2.61)
      > Could not get resource 'https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-compiler-embeddable/1.2.61/kotlin-compiler-embeddable-1.2.61.jar'.
         > Connection reset
```

Can you take a look ?

bjgill:
That looks unrelated. `master` also appears to be suffering similar sorts of errors at the moment.

I can see a way to re-run the build. Maybe try pushing a no-op amended commit to force a re-build?

bjgill:
Hurrah! I'll give it another day in case any of the rest of the technical committee want to give this a review, but will otherwise merge tomorrow.

PR #2159, opened 2019-02-14
[rust-server] Added client documentation to rust-server

Herrera93:
### PR checklist

### Description of the PR

I added documentation for the client generated by `rust-server`. This documentation is based on [`rust` client docs](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/rust) with some modifications to fit the use of variables. 

I modified `RustServerCodegen` to include `String` as a primitive type and determine if a model property is a primitive, this way it won't show a reference to file if the property is primitive.

Some modifications that need to be done are:
 - Update the auth example
 - Find a way to determine if a nested container is primitive i.e. [`HashMap<String, HashMap<String, String>>`](
https://github.com/joylabs/openapi-generator/blob/a6ccc6bc4c17c81891d2fefe8ba55c19e3304fa6/samples/server/petstore/rust-server/output/petstore-with-fake-endpoints-models-for-testing/docs/AdditionalPropertiesClass.md)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -783,7 +812,7 @@ public void postProcessParameter(CodegenParameter parameter) {
 
             String name = "models::" + getTypeDeclaration(parameter.dataType);
             parameter.dataType = name;
-            parameter.baseType = name;
+            // parameter.baseType = name;
```
bjgill: If this is unused, why not delete?

modules/openapi-generator/src/main/resources/rust-server/README.mustache
```diff
@@ -110,3 +110,75 @@ Now replace the implementations in `src/server.rs` with your own code as require
...
+{{#isApiKey}}- **Type**: API key 
+
+Example
+```
+	auth := context.WithValue(context.TODO(), sw.ContextAPIKey, sw.APIKey{
+		Key: "APIKEY",
+		Prefix: "Bearer", // Omit if not necessary.
+	})
+    r, err := client.Service.Operation(auth, args)
+```
```
bjgill: This looks like go to me. Can we just delete it?

Herrera93:
@bjgill Mainly There are doc comments but they only contain operation and model descriptions but do not contain parameters or properties descriptions.

I wanted to create centralized documentation for the client that can be explored in an interactive way so if there is a possibility that the user wants to expose the client.

bjgill:
Ah, right. I see. In general, my preference would be to improve the doc comments rather than adding this parallel documentation. However, you've done this useful work, so I'll not let the perfect stand in the way of the good.

I'll give this a couple of days for other reviews, and then 🚢

wing328:
@Herrera93 thanks for the PR 👍 

@bjgill thanks for the review.

Merging this into master to avoid too many outstanding PRs in this repo.

PR #2161, opened 2019-02-15
Use oauth token for basic bearer auth in Rust.

boxdot:
### PR checklist

### Description of the PR

Use `req_builder.bearer_auth` instead of `basic_auth` when `BearerAuth` security schema is specified.

@frol, @farcaller, @bjgill

### DIFF
--- a/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
+++ b/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
@@ -83,17 +83,17 @@ impl {{{classname}}} for {{{classname}}}Client {
             req_builder = req_builder.basic_auth(auth_conf.0.to_owned(), auth_conf.1.to_owned());
         };
         {{/isBasicBearer}}
+        {{#isBasicBearer}}
+        if let Some(ref token) = configuration.bearer_access_token {
+            req_builder = req_builder.bearer_auth(token.to_owned());
+        };
+        {{/isBasicBearer}}
         {{/isBasic}}
         {{#isOAuth}}
         if let Some(ref token) = configuration.oauth_access_token {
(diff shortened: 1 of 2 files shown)


bjgill:
Thanks for the PR.

I don't know much about authentication, so assume that you've verified that this works. 

I think the CI failure is spurious, so have restarted the build. Assuming it goes green, I'll give it a couple of days for other reviews, and then :shipit:

modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
```diff
@@ -78,15 +78,22 @@ impl {{{classname}}} for {{{classname}}}Client {
...
             req_builder = req_builder.basic_auth(auth_conf.0.to_owned(), auth_conf.1.to_owned());
         };
+        {{/isBasicBearer}}
         {{/isBasic}}
         {{#isOAuth}}
         if let Some(ref token) = configuration.oauth_access_token {
             req_builder = req_builder.bearer_auth(token.to_owned());
         };
         {{/isOAuth}}
+        {{#isBasicBearer}}
```
wing328: @boxdot thanks for the PR. I would suggest put this new block inside the {{#isBasic}} ... {{/isBasic}} block. What do you think?

And what about creating bearer_access_token instead or reusing oauth_access_token?

Some other clients are using `access_token` so it can be shared with OAuth, Bearer authentications.

gferon: (I'm working with @boxdot)

I think having a single `access_token` also works, but I guess by adding `bearer_access_token` (see last commit in PR), we avoid breaking existing clients.

wing328:
FYI. CircelCI error has been fixed in master.

PR #2162, opened 2019-02-15
[R] various improvements and bug fixes

wing328:
### PR checklist

### Description of the PR

- use optional parameters
- fix header parameters
- rename callAPI to CallAPI to conform to the style guide
- fix API key authentication

### DIFF
--- a/modules/openapi-generator/src/main/resources/r/api.mustache
+++ b/modules/openapi-generator/src/main/resources/r/api.mustache
@@ -32,27 +32,25 @@
       }
     },
     {{#operation}}
-    {{operationId}} = function({{#allParams}}{{paramName}}, {{/allParams}}...){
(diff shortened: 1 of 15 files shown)


modules/openapi-generator/src/main/resources/r/api.mustache
```diff
@@ -93,13 +93,13 @@
       {{#isApiKey}}
```
Ramanth: in the lines above at line 91, the usage of jsonlite::base64_enc() is not working when the authentication string is longer.
The issue is that the above method  is encoding but adding a extra '\n' in the output of the encoded string.

Please refer to a similar issue in the below link:
[jsonline issue](https://github.com/jeroen/jsonlite/issues/220)

Please make a fix or any workaround. Thanks in advance

wing328: Thanks. Will use other base64 encoding function instead. Can you please open an issue to track this instead?

wing328: @Ramanth I've filed https://github.com/OpenAPITools/openapi-generator/pull/2171 to fix the issue. Can you please check to see if that works for you?

wing328: #2171 merged into r_refactor. Please give it a try.

PR #2166, opened 2019-02-15
[kotlin][client] bytearray conversion

alexnavratil:
### PR checklist

@jimschubert , @dr4ke616 

### Description of the PR

Moshi doesn't support parsing a base64 encoded string (as defined for fields with type: string, format: byte) to a kotlin.ByteArray. The simplest solution to overcome this limitation was to convert those fields to kotlin.String instead of kotlin.ByteArray. Another solution would be to write a custom type adapter for moshi to support string to ByteArray conversions. I have chosen the first method as this was the simplest for now. If there's something against it, I have already prepared a custom type adapter ;)

PS: There might be problems with converting fields of type number to BigDecimal as well as this isn't supported by moshi as far as I know. For this case, a custom type adapter is the only solution. If you think, it would be good to add this also in this PR, I would implement it.


jimschubert:
@alexnavratil great catch!

If you have the type adapter, can we use that instead? The issue with changing the default mapping is that although it works, it's not _technically_ correct. A savvy user might notice this, and remap via the `type-mappings` option back to a byte array. With a type adapter, we'd be safe in both situations and any other type-mapping changes would be on the user to ensure there's a valid adapter.

alexnavratil:
good point, I have added the ByteArrayAdapter as promised.

maxammann:
Isn't this wrong? I think he openapi spec specifies that the string must be base64. Also encoding just a random bytearray as UTF-8 is very dangerous.

PR #2173, opened 2019-02-18
Add support for java.time.Instant for java generators

stijnkoopal:
### PR checklist

- [x] Ran ./bin/spring-mvc-petstore-j8-instant.sh` 

### Description of the PR

We are using `java.time.Instant` as Date(Time) type for all our backend services. Currently we achieve this by using a custom `dateLibrary` and custom `typeMappings` and `importMappings`. This is suboptimal however.

`java.time.Instant` is a valid type for code generation if you are using UTC timestamps only. Hence, this PR.


wing328:

stijnkoopal:
Fair point, fixed!

wing328:
cc @bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01)

cbornet:
It's really easy to switch from an OffsetDateTime to an Instant. Do we really need an option for this ? Also note that Jackson will only accept "Z" format for Instant which doesn't comply with RFC3339

stijnkoopal:
Well, we use Instant throughout our service-2-service APIs since most of your business logic, data storage, and data exchange should be in UTC. As you say, it is easy to use Instant by the following configuration:
```
configOptions = [
        dateLibrary: "custom",
        java8: true,
        jsr310: true
    ]
    typeMappings = [
        "DateTime": "Instant"
    ]
    importMappings = [
        Instant: "java.time.Instant"
    ]`
```

However then we need to have this in most of our services, hence I thought I would add it to the source.

cbornet:
Have you tried using OffsetDateTime directly ? It represents a time in UTC timezone with a ZoneOffset and it's really easy to convert it to/from an Instant.
```java
OffsetDateTime dt = OffsetDateTime.now();
Instant instant = dt.toInstant();
dt = instant.atOffset(ZoneOffset.UTC);
```

tomdevroomen:
I've seen the PR and I have to agree with @stijnkoopal having a config value for using java.time.Instant would be beneficial.
Wether OffsetDateTime and Instant conversion would be easy isn't the goal of this PR, it would add extra code to a codebase to do the conversion.
With this config option this wouldn't be needed.

cbornet:
You can already use typeMappings and importMappings if you really want Instant. Have you tried it ?

stijnkoopal:
Yeah, that was the thing we were trying to avoid. We'll choice the way of configuration.

PR #2202, opened 2019-02-20
#2195 add enum value getter for spring generator

### LINKED ISSUE #2195: [BUG][JAVA-Spring] Enum without value getter
##### Description

Generated enum class is missing getter for it's value. Is there any option with which I can turn on getter generation? Method toString is sufficient for String enums, but if I want numbered values it's annoying to convert values between different types.

3.3.4 openapi-generator-maven-plugin

```yaml
components:
  schemas:
    QueueStatus:
      type: integer
      enum:
        - 1
        - 10
        - 100
        - 1000
      x-enum-varnames:
        - EDITABLE
        - DELETABLE
        - LOCK
        - SPECIAL
```

`mvn clean install`

##### Generated output class
```java
public enum QueueStatus {
  
  EDITABLE(1),
  
  DELETABLE(10),
  
  LOCK(100),
  
  SPECIAL(1000);

  private Integer value;

  QueueStatus(Integer value) {
    this.value = value;
  }

  @Override
  @JsonValue
  public String toString() {
    return String.valueOf(value);
  }

  @JsonCreator
  public static QueueStatus fromValue(String text) {
    for (QueueStatus b : QueueStatus.values()) {
      if (String.valueOf(b.value).equals(text)) {
        return b;
      }
    }
    throw new IllegalArgumentException("Unexpected value '" + text + "'");
  }
}
```

bilak:
fixes #2195


wing328:
@bilak thanks for the PR but it seems like it removes a lot of files. Is that expected?

bilak:
@wing328 actually not, I just wanted to follow the contribution guide and this happened. If you can guide me through the process I can fix it.

wing328:
@bilak thanks for the explanation. Given that there are merge conflicts, do you mind filing a new one based on the latest master and do not update the samples? I'll update the sample later as part of the test. What do you think?

bilak:
@wing328 definitelly can do that, just have a question. My intention was to modify this files [enumClass.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/enumClass.mustache) and [enumOuterClass.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/enumOuterClass.mustache). After usage in my local project I realised that it would be very helpful if the signature of method `fromValue` changes from `fromValue(String value)` to `fromValue({{{dataType}}} value)` but this is not backward compatible. This is helpful for jackson library at it knows how to create the object. What do you think about that? Do I have to change the signature or would it be ok to add another method with signature as I described?

Thanks

wing328:
> After usage in my local project I realised that it would be very helpful if the signature of method fromValue changes from fromValue(String value) to fromValue({{{dataType}}} value) but this is not backward compatible. This is helpful for jackson library at it knows how to create the object. What do you think about that?

It's ok to have breaking changes for the upcoming 4.0.0 release but I would recommend creating a separate PR just for that as smaller PRs are easier to merge. In other words, a new PR based on latest master for this PR and another PR for fromValue.

bilak:
closing in favour of #2346 and #2347

PR #2226, opened 2019-02-24
[Ruby] various improvements

wing328:
### PR checklist

### Description of the PR

- add to_s method in ApiError class
- fix initialize in model (baseName vs name)
- validate the input parameter (attributes) in model's initialize
- add code sample for model 
- add auto-generated Travis CI config file


autopp:
@wing328 I'll review it. Please wait for a while.

wing328:
@autopp please take your time and thanks for reviewing the change.

modules/openapi-generator/src/main/resources/ruby-client/api_error.mustache
```diff
@@ -26,5 +26,10 @@ module {{moduleName}}
         super arg
       end
     end
+
+    # Override to_s to print variables
+    def to_s
```
autopp: It seems good to return information on response!
However, if left as it is, the error message passed in the form below will be hidden 🤔 
```ruby
ApiError.new('some reason')
ApiError.new(message: 'some reason', code: 400)
```
I'll think about how it should be...

autopp: In any case, if you change to_s, can you also fix the corresponding test?
(This causes CircleCI to fail)
```
Failures:

  1) Petstore::ApiClient#initialize should save the message if one is given
     Failure/Error: expect(err.message).to eq("Hello")

       expected: "Hello"
            got: "{:code=>nil, :response_headers=>nil, :response_body=>nil}"

       (compared using ==)
     # ./spec/custom/api_error_spec.rb:7:in `block (3 levels) in <top (required)>'

  2) Petstore::ApiClient#initialize should save the hash as message if no message is given
     Failure/Error: expect(err.message).to eq("{:code=>500, :response_body=>\"server error\"}")

       expected: "{:code=>500, :response_body=>\"server error\"}"
            got: "{:code=>500, :response_headers=>nil, :response_body=>\"server error\"}"

       (compared using ==)
     # ./spec/custom/api_error_spec.rb:12:in `block (3 levels) in <top (required)>'
```

wing328:
@autopp thanks for the review. I've incorporated your feedback and updated the PR accordingly. Please have another look when you've time.

modules/openapi-generator/src/main/resources/ruby-client/api_error.mustache
```diff
@@ -26,5 +26,24 @@ module {{moduleName}}
         super arg
       end
     end
+
+    # Override to_s to display a friendly error message
+    #def to_s
+    #  message
+    #end
+
+    def message
```
autopp: @wing328 Hmm...
In Ruby `Exception#message` returns the same result as `Exception#to_s` by default, so it may be confusing for the user to change the definition of message only for this class.
https://ruby-doc.org/core-2.6.1/Exception.html#method-i-message

Instead of overwriting `to_s` and `message`, how about changing the definition of `initialize`?

E.g.

```ruby
    def initialize(arg = nil)
      super
      if arg.is_a? Hash
        arg.each do |k, v|
          instance_variable_set "@#{k}", v
        end
      end
    end
```

```
[1] pry(main)> require 'petstore/api_error'
=> true
[2] pry(main)> Petstore::ApiError.new().to_s
=> "Petstore::ApiError"
[3] pry(main)> Petstore::ApiError.new("Hello").to_s
=> "Hello"
[4] pry(main)> Petstore::ApiError.new(code: 400, response_headers: {}, response_body: 'Error').to_s
=> "{:code=>400, :response_headers=>{}, :response_body=>\"Error\"}"
[5] pry(main)> Petstore::ApiError.new(message: "Hello", code: 400, response_headers: {}, response_body: 'Error').to_s
=> "{:message=>\"Hello\", :code=>400, :response_headers=>{}, :response_body=>\"Error\"}"
```

wing328: @autopp Thanks for the suggestion. Let me share a bit more background information. I got this idea from Twilio Ruby client: https://github.com/twilio/twilio-ruby/blob/08f1ba2f042eeefb29f19f7dd13add40c4117ac9/lib/twilio-ruby/framework/error.rb#L33 (this is not to say we must follow this approach)

And the enhancement I want to make is to allow the users to easily find out the error message (users may not even notice ApiError's parent class is StandardError and has a message field. That's why I overwrite the to_s method to show a friendly message with all the available information. 

I thought about doing everything in `initialize` too but not sure if there are use cases in which some of the fields can/will be updated later. That's why I overwrite the "message" to return all useful information in ApiError (https://stackoverflow.com/a/32691187/677735) when users need access to "message" (in other words, I think the "message" should be constructed based on all latest attributes/values in ApiError)

autopp: @wing328 Thanks for the explanation, I was satisfied. I think that it is good with the current code 👍 
Could you cancel the comment out of `to_s`?

wing328: @autopp thanks for the reminder. I completely forgot.

Issue #788, opened 2018-08-11
[Code Analysis] Report by sonarcloud.io


wing328:
##### Description

The code analysis report by https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&resolved=false&types=BUG uncovers some bugs and vulnerabilities. 

##### openapi-generator version

Latest master 

##### Suggest a fix/enhancement

@ymohdriz is very kind to offer help to fix the bugs in the report. If anyone wants to help out (e.g. fixing the vulnerabilities), please reply to let us know.

jmini:
Does this mean that someone registered us at sonarcloud? We have issue #33 pending since month...

I think that we can also integrate "sonarcloud.io" with GitHub PRs to get a report of the changes introduced by a PR. We should also investigate this.

PR #2229, opened 2019-02-25
fix resources management

ramzimaalej:
Fix the following issues: 

https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&open=AWUi1WNYg-pHJSBwrBAz&resolved=false&types=BUG

https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&open=AWUi1WPRg-pHJSBwrBKT&resolved=false&types=BUG

https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&open=AWUi1WPRg-pHJSBwrBKU&resolved=false&types=BUG

https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&open=AWUi1WPRg-pHJSBwrBMR&resolved=false&types=BUG

@wing328


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1251,23 +1251,7 @@ public String toExampleValue(Schema schema) {
...
-        } else if (ModelUtils.isIntegerSchema(schema)) {
-            return "null";
-        } else if (ModelUtils.isStringSchema(schema)) {
-            return "null";
-        } else if (ModelUtils.isObjectSchema(schema)) {
-            return "null";
-        } else {
-            return "null";
-        }
+        return "null";
```
wing328: The code block aims to illustrate what type to check when computing the example value.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1282,23 +1266,7 @@ public String toDefaultValue(Schema schema) {
...
```
wing328: Same as above: https://github.com/OpenAPITools/openapi-generator/pull/2229/files#r261599400

wing328:
cc @OpenAPITools/generator-core-team  as well

wing328:
For https://github.com/OpenAPITools/openapi-generator/issues/788

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
ramzimaalej: @wing328 I don't get your point, the previous code block returns null in all cases. Can you please elaborate more?

wing328: My point is that for someone new to the code base, they may have no clue what type to check and what function to leverage (e.g. ModelUtils.isIneger) to check the type.

The code is put there for onboarding new users.

wing328:
Minor suggestion: next time you may want to create a new branch for the change.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jimschubert: Maybe it can be reverted and a note can be added that the conditionals exist as examples for new contributors?

ramzimaalej: I would put the conditionals as comment to enlighten new contributors and just leave the code return null. I personally prefer to have cleaner code and cover the unusual case in a comment

wing328:
Sure.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jimschubert: I often agree with this approach. However, if any of the methods are renamed, comments are often missed and this can be confusing.

From another approach, new and existing contributors might not easily reason about what a single "null" result would mean here, while the explicit cases provide clear intention.

I think commenting on either way would be acceptable. We can always change later if people have questions.

PR #2257, opened 2019-02-27
Send "true" or "false" for boolean params.

### LINKED ISSUE #2204: [BUG] PHP sends boolean params as "1" or "" instead of "true" or "false"
#### Bug Report Checklist

##### Description

When params are of type "boolean", they should be sent as "true" or "false", but the PHP generator sends them as "1" or "". But it correctly documents the parameters as `@param  bool`. So if you send boolean parameters, it ends up doing something like `(string) true`, which evaluates to `1`.

4.0.0-SNAPSHOT

https://gist.github.com/jacobweber/e83874fbf1c0ee6ede79100a7507e11f

`java -jar openapi-generator.jar generate -i <url> -g php -o /path/to/output`

##### Steps to reproduce

1. Build the above spec.
2. Run `composer install` from the output dir.
3. Run the "Getting Started" example code in README.md:
```
$admin = True; // bool | Include admin users?
try {
    $result = $apiInstance->usersGet($admin);
    print_r($result);
} catch (Exception $e) {
    echo 'Exception when calling DefaultApi->usersGet: ', $e->getMessage(), PHP_EOL;
}
```
4. It calls `GET http://localhost/api/users?admin=1` instead of `GET http://localhost/api/users?admin=true`.

##### Related issues/PRs

N/A

##### Suggest a fix

Detect booleans and convert to "true" or "false".

jacobweber:
### PR checklist

- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) ❤️, @ybelenko (2018/07), @renepardon (2018/12)

### Description of the PR

If you pass a PHP boolean value as a parameter in the query/path/form, this will serialize it as "true" or "false" instead of "1" or "".

Fixes #2204.

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -165,8 +165,9 @@ class ObjectSerializer
      * Take value and turn it into a string suitable for inclusion in
      * the parameter. If it's a string, pass through unchanged
      * If it's a datetime object, format it in ISO8601
+     * If it's a boolean, convert it to "true" or "false".
      *
-     * @param string|\DateTime $value the value of the parameter
+     * @param string|bool|\DateTime $value the value of the parameter
      *
      * @return string the header string
      */
@@ -174,6 +175,8 @@ class ObjectSerializer
     {
         if ($value instanceof \DateTime) { // datetime in ISO8601 format
             return $value->format(\DateTime::ATOM);
+        } else if (is_bool($value)) {
+            return $value ? 'true' : 'false';
         } else {
             return $value;
         }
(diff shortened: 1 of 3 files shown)


wing328:
@jacobweber thanks for the fix, which looks good to me 👍

etremblay:
Hi, this pr fixes boolean value encoding in `ObjectSerializer::toString()`, but in current master, this function is not used to encode query string parameters.

https://github.com/OpenAPITools/openapi-generator/blob/27680214fd209add606296e7412df01a06c44c1b/modules/openapi-generator/src/main/resources/php/api.mustache#L481

I think encoding query parameters was the whole purpose of https://github.com/OpenAPITools/openapi-generator/issues/2204

Would you like me to open a proper pull request?

etremblay:
After some investigations, I found that the lost of this fix is a regression from 

https://github.com/OpenAPITools/openapi-generator/pull/3984

PR #2258, opened 2019-02-27
[haskell-http-client] Allow logger selection via cabal flags. Emit Consumes */* for requestBody when not specified

jonschoning:


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/HaskellHttpClientCodegen.java
```diff
@@ -260,7 +260,7 @@ public HaskellHttpClientCodegen() {
 
         cliOptions.add(CliOption.newString(PROP_MODEL_DERIVING, "Additional classes to include in the deriving() clause of Models"));
         cliOptions.add(CliOption.newBoolean(PROP_STRICT_FIELDS, "Add strictness annotations to all model fields").defaultValue((Boolean.TRUE.toString())));
-        cliOptions.add(CliOption.newBoolean(PROP_USE_MONAD_LOGGER, "Use the monad-logger package to provide logging (if false, use the katip logging package)").defaultValue((Boolean.FALSE.toString())));
+        cliOptions.add(CliOption.newBoolean(PROP_USE_KATIP, "Sets the default value for the UseKatip cabal flag. If true, the katip package provides logging instead of monad-logger").defaultValue((Boolean.FALSE.toString())));
```
wing328: @jonschoning if you want to gather feedback from users on this new option, you may want to set it to "true" so as to force everyone to katip it by default.

jonschoning: done

PR #2291, opened 2019-03-04
[R] Changes to deserialize the map container type vars

Ramanth:
Added deserialize support for models with parameter of map continer type.

### PR checklist

### Description of the PR

Added changes to the model mustache file in the toJSONString() method where the handling of the isMapContainer type vars is handled, which is not handled earlier.

in the isMapContainer handled when it is primitive type map and non-primitive type cases.
which is finally returning a list with names and values equivalent to keys and values of the map.

Please review and take required action.

@wing328 , for your review and suggestion.

Thanks,
Ramanth

### DIFF
--- a/modules/openapi-generator/src/main/resources/r/model.mustache
+++ b/modules/openapi-generator/src/main/resources/r/model.mustache
@@ -213,7 +213,7 @@
 			  self$`{{baseName}}` <- vector("list", length = nrow(data.frame))
 			  for (row in 1:nrow(data.frame)) {
 				  {{baseName}}.node <- {{dataType}}$new()
-				  {{baseName}}.node$fromJSON(jsonlite::toJSON(data.frame[row,,drop = TRUE], auto_unbox = TRUE))
+				  {{baseName}}.node$fromJSON(jsonlite::toJSON(data.frame[row,,drop = F], auto_unbox = TRUE))
 				  self$`{{baseName}}`[[row]] <- {{baseName}}.node
 			  }
 			  {{/isPrimitiveType}}
@@ -225,7 +225,7 @@
 			  {{/isPrimitiveType}}
 			  {{^isPrimitiveType}}
 				x <- {{classname}}Object$`{{baseName}}`
-				self$`{{baseName}}` <- sapply(names(x), function(y) {{dataType}}$new()$fromJSONString(jsonlite::toJSON(x[[y]])))  
+				self$`{{baseName}}` <- sapply(names(x), function(y){ Obj <- {{dataType}}$new();Obj$fromJSON(jsonlite::toJSON(x[[y]], auto_unbox = TRUE));Obj})  
 			  {{/isPrimitiveType}}
 		{{/isMapContainer}}  
 	  {{/isContainer}}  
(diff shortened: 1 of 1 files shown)


wing328:
As discussed, please keep the indentation of the mustache tags for easier review.

For map/dictionary in R, what about using https://www.rdocumentation.org/packages/dict/versions/0.10.0 ?

Ramanth:
@wing328  the dict is not yet released to CRAN. user may need to install it from github.
its mentioned the package [hash ](https://cran.r-project.org/web/packages/hash/index.html) is available on github only limitaion it has keys as character or numeric and not vectors (which is additional in dict ).
Should we go with dict or hash?

Ramanth:
@wing328 i tried to fix the indentation as much as i can
Please review and provide your inputs

wing328:
Looks like [hash](https://github.com/decisionpatterns/r-hash) is more mature. Let's use it.

Ramanth:
@wing328  i think we don't need to use hash as well as i see it, since we can use list and for equivalent of dict we can have **named lists**

for dict types of return types we can return them as "list(ModelObject)" as equivalent to  "dict(str, ModelObject)" in python

any thoughts?

modules/openapi-generator/src/main/resources/r/model.mustache
```diff
@@ -212,19 +213,30 @@
...
+      {{/isContainer}}  
+      {{^isContainer}} 
       {{#isPrimitiveType}}
       self$`{{baseName}}` <- {{classname}}Object$`{{baseName}}`
       {{/isPrimitiveType}}
       {{^isPrimitiveType}}
       self$`{{baseName}}` <- {{dataType}}$new()$fromJSON(jsonlite::toJSON({{classname}}Object${{baseName}}, auto_unbox = TRUE))
-      {{/isPrimitiveType}}
-      {{/isListContainer}}
+      {{/isPrimitiveType}}
```
wing328: minor suggestion: remove trailing space

modules/openapi-generator/src/main/resources/r/model.mustache
```diff
@@ -212,19 +213,30 @@
       self$`{{baseName}}` <- vector("list", length = nrow(data.frame))
       for (row in 1:nrow(data.frame)) {
           {{baseName}}.node <- {{dataType}}$new()
-          {{baseName}}.node$fromJSON(jsonlite::toJSON(data.frame[row,,drop = TRUE], auto_unbox = TRUE))
+          {{baseName}}.node$fromJSON(jsonlite::toJSON(data.frame[row,,drop = F], auto_unbox = TRUE))
```
wing328: Do you mind explaining why change drop from TRUE to F (false)?

Ramanth: @wing328 drop = TRUE is actually making a data.frame of one column reducing the dimension and making it as list
and later when it is being converted back to object it is not being converted properly to original object as its dimension is reduced.

For Example: consider this json 
```
[{"id" : 1234}, {"id": 1122}, {"id" : 1024}]
```
when the above json is used with jsonlite with drop = TRUE the key name "id" is lost when it is converted from object to json and back to object.

Im not sure why the drop = TRUE is kept initially?

wing328: If I remember correctly, it only works if I set it to TRUE.

I think we need to add some tests to ensure all previous use cases are covered.

Ramanth:
@wing328 i am very much afraid that the changes we are discussing will not work other than for simpler return types and simpler model deserialization.

As far as i see, we need to introduce a function called "deserialize" which deserializes to objects recursively to handle the nested structures.

in the RClientCodegen i found that we are returning the Array and Map type of containers as simple objects, these needs to be handled.

Ramanth:
@wing328 i am closing this PR as i raised a new one with updated deserialization that handles various scenarios.  This one -> https://github.com/OpenAPITools/openapi-generator/pull/2369

PR #2295, opened 2019-03-04
second part of fixing Sonar issues

ramzimaalej:
@wing328

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1270,12 +1270,15 @@ public class DefaultCodegen implements CodegenConfig {
     }
 
     /**
-     * Return property value depending on property type
+     * Return property value depending on property type.
      * @param schema property type
      * @return property value
      */
     @SuppressWarnings("squid:S3923")
     private String getPropertyDefaultValue(Schema schema) {
+        /**
+         * Although all branches return null, this is left intentionally as examples for new contributors
+         */
         if (ModelUtils.isBooleanSchema(schema)) {
             return "null";
         } else if (ModelUtils.isDateSchema(schema)) {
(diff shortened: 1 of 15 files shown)


wing328:

wing328:
For #788

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1274,8 +1274,8 @@ public String toDefaultValue(Schema schema) {
      * @param schema property type
      * @return property value
      */
+    @SuppressWarnings("squid:S3923")
```
jimschubert: Might want to comment here that the branches are kept for examples, and the repeated `return "null"` is intentional?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractAdaCodegen.java
```diff
@@ -426,10 +426,8 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
         final Map<String, List<String>> scopes = new HashMap<>();
         for (SecurityRequirement requirement : securities) {
             for (String key : requirement.keySet()) {
-                SecurityScheme securityScheme = securitySchemes.get(key);
-                if (securityScheme != null) {
-                    scopes.put(key, requirement.get(key));
-                }
+                Optional.ofNullable(securitySchemes).ifPresent(_securitySchemes -> Optional.ofNullable(securitySchemes.get(key))
```
jimschubert: Is this all to avoid wrapping the deleted code with an `if (securitySchemes != null)` block?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
ramzimaalej: sure thing

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractAdaCodegen.java
ramzimaalej: Thanks for your comment @jimschubert I have refactored the code to better express my intentions. It's just a personal preference to use Optional.ofNullable instead of checking for null.

jimschubert:
Looks good to me. Any @OpenAPITools/generator-core-team have additional comments before merge?

wing328:
I'd another review and the changes look good to me.

wing328:
For #788

PR #2301, opened 2019-03-05
[R] Handling null values for optional params while building the model json object

Ramanth:
### PR checklist

### Description of the PR
**Error when optional parameters are not given for the model object**

When generating the json of a model using the toJSONString() we encounter an error when we dont pass the optional paramameters (which are of non-primitive type)

Below changes handles the listContainer type optional parameters by replacing them with "null". so that it will be ignored by the endpoint.

Please make necessary changes if required.
@wing328 , please look into and take required action.

**Example to test:**
please consider a model which is hanving an example as below
```
{
	"key1" : "value1",
	"key2" : [  {"k1" : "v1"},
                    {"k2" : "v2"} ],
	"key3" : {  "k3" : "v3",
                     "k4" : "v4"
         	      }
}
```

In the above example consider key2 as an optional parameter and of type listContainer (i.e, array). 
Which is throwing an error when we dont provide the value for key2 while generating the model object.
Please review this changes and make required changes to handle such scenarios.

Thanks,
Ramanth.

### DIFF
--- a/modules/openapi-generator/src/main/resources/r/model.mustache
+++ b/modules/openapi-generator/src/main/resources/r/model.mustache
@@ -157,8 +157,8 @@
       }
       {{/vars}}
     },
-    toJSONString = function() {
-      sprintf(
+       toJSONString = function() {
+      outstring = sprintf(
         '{
            {{#vars}}
            "{{baseName}}":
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/resources/r/model.mustache
```diff
@@ -181,24 +181,32 @@
...
+        jsonlite::toJSON(self$`{{baseName}}`$toJSON(), auto_unbox=TRUE)
         {{/isPrimitiveType}}
         {{/isListContainer}}
+	}
+	else{
+	  "null"
+	}{{#hasMore}},{{/hasMore}}
         {{/vars}}
       )
+	  outstring = gsub("[\r\n]| ", "", outstring)
```
wing328: I'm afraid this won't work if the data/value contains line breaks

Ramanth: @wing328 you are right, i can remove this line considering what you mentioned. Thanks

wing328:
As discussed, instead of using `sprintf`, what about using string concatenation instead so that we can check for null before adding anything to the output?

wing328:
https://github.com/OpenAPITools/openapi-generator/pull/2777 addresses this problem for Dart2. We should adopt a similar approach for R

Ramanth:
raised a new [PR](https://github.com/OpenAPITools/openapi-generator/pull/3082) with different approach.

PR #2332, opened 2019-03-08
[typescript-rxjs] drop unneeded function wrapping 

denyo:
### PR checklist

### Description of the PR
This PR removes unneeded wrapping functions within api controllers.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-rxjs/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-rxjs/apis.mustache
@@ -37,7 +37,7 @@ export class {{classname}} extends BaseAPI {
      * {{&summary}}
      {{/summary}}
      */
-    {{nickname}}Raw({{#allParams.0}}requestParameters: {{operationIdCamelCase}}Request{{/allParams.0}}): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}> {
+    private {{nickname}}Raw({{#allParams.0}}requestParameters: {{operationIdCamelCase}}Request{{/allParams.0}}): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}> {
         {{#allParams}}
         {{#required}}
         if (requestParameters.{{paramName}} === null || requestParameters.{{paramName}} === undefined) {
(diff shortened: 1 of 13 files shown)


macjohnny:
@denyo btw. can you remind we why there is this separation `get...()` and `get...Raw()`?

denyo:
@macjohnny I am not really sure. I just took what was already there in `typescript-fetch` which this generator is based on. 

The only thing that I can think of and which happened to me in my current project is that you need the raw xhr response in order to extract a certain header (e.g. for pagination). 
I was able to achieve this by extending the `BaseAPI`'s `protected request()` by `ResponseOpts`:
```
protected request<T>(reqOpts: RequestOpts, resOpts?: ResponseOpts): Observable<T> {
...
    if (resOpts && resOpts.response === 'raw') {
        return res as any as T;
    } else {
        return res.response as T;
    }
...
}
```

In the api controller I then have something like
```
private loginUsingPOSTRaw(requestParameters: LoginUsingPOSTRequest, responseOptions?: ResponseOpts): Observable<AjaxResponse> {
    ...
    return this.request<AjaxResponse>({
        ...
    }, responseOptions);
}

loginUsingPOST(requestParameters: LoginUsingPOSTRequest): Observable<MachineOperatorDto>
loginUsingPOST(requestParameters: LoginUsingPOSTRequest, responseOptions: ResponseOpts): Observable<AjaxResponse>

loginUsingPOST(requestParameters: LoginUsingPOSTRequest, responseOptions?: ResponseOpts): Observable<any> {
    return this.loginUsingPOSTRaw(requestParameters, responseOptions);
}
```

And where ever I call this function with `{ response: 'raw' }` I can then extract a header with `res.xhr.getResponseHeader(SOME_KEY)`.

I did these modifications manually. So they aren't part of the generator as of now.

macjohnny:
ok, but currently there is no difference between the two, right?

denyo:
Yes, right now they have the same signature and the public one is calling the private `RAW` one.

denyo:
It looks like in the `typescript-fetch` client this was used to unwrap the response promise and have the right type since promises don't support generics? If thats the case I can simplify this client.

macjohnny:
> It looks like in the typescript-fetch client this was used to unwrap the response promise and have the right type since promises don't support generics? If thats the case I can simplify this client.

sounds reasonable, with rxjs, the flow is straight-forward and nicely typed

denyo:
There we go.

Issue #2312, opened 2019-03-07
[BUG] haskell-http-client: Arbitrary instance for recursive data structures runs forever


guoshimin:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

@jonschoning  haskell-http-client generates QuickCheck-based JSON roundtrip tests. For recursive data structures, the Arbitrary instance runs forever.

##### openapi-generator version

master

##### OpenAPI declaration file content or url

https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.9/api/openapi-spec/swagger.json

##### Steps to reproduce

Run `stack test` on the generated code. The test will be stuck at `V1beta1CustomResourceDefinition`. The cause is `V1beta1JSONSchemaProps` which is a recursive structure.

jonschoning:
recursive datatypes typically require writing a custom Arbitrary instances to control the size of the tree.
some initial thoughts: 
 1. provide a list of datatypes to the generator for which it should not automatically create arbitrary instances for.  These datatypes can then be prepared in a different non-generated file with the custom logic.
 2. provide a list of datatypes to the generator for which MimeJson tests should be excluded entirely
 3. find a way to automatically generate shrinking trees for recursive types; given the limits of codegen, i'm not sure off the top of my head how to achieve it, but am open to ideas.

guoshimin:
Taking the suggestion from http://www.cse.chalmers.se/~rjmh/QuickCheck/manual_body.html#16, how about:
- generate a `{{typeName}}_sizedGen :: Int -> Gen {{typeName}}` for every record type
- In the implementation, if n is 0, generate `None` for all optional fields. Otherwise, for each optional record field, generate `None` with 50% probability and call sizedGen with `n-1` (or some other decreasing function on `n`) with 50% probability.
- The `Arbitrary` instance for `{{typeName}}` would be `sized {{typeName}}_sizedGen`

jonschoning:
[...]

PR #2343, opened 2019-03-09
[haskell-http-client] use a decreasing size for Gen (Maybe a), to prevent infinite recursion (#2312)

jonschoning:


jonschoning:
c.f. https://github.com/OpenAPITools/openapi-generator/issues/2312

PR #2373, opened 2019-03-12
[WIP] add support to remote files

ramzimaalej:
### PR checklist

- [] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh`, `./bin/security/{LANG}-petstore.sh` and `./bin/openapi3/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.

### Description of the PR

### DIFF
--- a/modules/openapi-generator-maven-plugin/pom.xml
+++ b/modules/openapi-generator-maven-plugin/pom.xml
@@ -46,6 +46,12 @@
             <artifactId>maven-plugin-annotations</artifactId>
             <version>3.4</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/commons-validator/commons-validator -->
+        <dependency>
+            <groupId>commons-validator</groupId>
+            <artifactId>commons-validator</artifactId>
+            <version>1.6</version>
+        </dependency>
         <dependency>
             <groupId>org.openapitools</groupId>
             <artifactId>openapi-generator</artifactId>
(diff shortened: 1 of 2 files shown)


wing328:
@ramzimaalej thanks for the PR.

The `-i` command line argument in the CLI already supports both local files and remote URLs so you may want to reuse the logic there.

Catscratch:
Bump! Please merge it into master as soon as it is working.

ramzimaalej:
using the same code as "-i" command would mean slower execution as it loads and validates the api file, unless we want to undertake a refactoring to make the code look the same. What I did here is to verify if the provided file location is valid, if yes download the file and initiate the processing in the same fashion as a local file

@wing328

jimschubert:
@ramzimaalej it seems to me (please correct me if I'm wrong), that this doesn't handle split remote files and would in fact break this support? Also, renaming to a single local yaml file seems like this would break when the remote file is JSON.

I'm not following the benefit of including the logic here when it already exists in swagger-parser. Could you provide an example of how this adds value in this scenario?

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/3826

Issue #2362, opened 2019-03-11
[REQ][haskell-http-client] Support type-mapping and import-mappings


guoshimin:
## Describe the solution you'd like

Support type-mapping and import-mappings so we can provide custom implementation for selected model, for example, as is done in https://github.com/kubernetes-client/gen/blob/master/openapi/java.xml.

## Additional context

In the OpenAPI spec for Kubernetes, there is an `IntOrString` type that's currently mapped to `object` if there is no custom mapping for it: https://github.com/kubernetes-client/gen/blob/c07ee54dcb0ae70c6ef587dded93b6ea75f354ea/openapi/preprocess_spec.py#L275. It would be more type safe to create (by hand) a Sum type for it.

@jonschoning

jonschoning:
type-mappings:
  - the behavior of type-mappings changed in openapi-generator in the case that the spec defines a primitive model ref for the type according to https://github.com/kubernetes-client/gen/issues/93#issuecomment-457844107 , so if we can to supply a custom `IntOrString`, but we'll need to remove the `intstr.IntOrString` model definition from the spec.

import-mappings:
 - i will look into adding support

jonschoning:
modify preprocessor to remove models:
https://github.com/jonschoning/kubernetes-client-gen/commit/b876d778d5f9190b6e08f305bd82a9e1219cf459

regen with import-mappings and type-mappings: 
https://github.com/jonschoning/kubernetes-client-haskell/commit/063587bdf4e0f841d41bbaea4d928b6b6579c2e2

add CustomTypes and CustomInstances: 
https://github.com/jonschoning/kubernetes-client-haskell/commit/a268948aca9bdc4259d712e5df27058ca99350d0

tests passed:
https://gist.github.com/jonschoning/df46961e0c999902eec07a3f7785822e

guoshimin:
Thanks for the quick fix! I also did the same test and it worked.

PR #2381, opened 2019-03-13
[haskell-http-client] add support for import-mappings

jonschoning:
import mappings take the form of:
```
--import-mappings %DATATYPE%=%MODULE%
```

full example (with --type-mappings):
```
--type-mappings intstr.IntOrString=IntOrString
--import-mappings IntOrString=Kubernetes.CustomTypes
```

The import-mapped module will be an exposed module of the generated library.

The import mappings will be re-exported from ImportMappings, which itself is re-xported from Models.

Model.hs
```
module Kubernetes.Model (module Kubernetes.Model, module Kubernetes.ImportMappings) where
import Kubernetes.ImportMappings
```
ImportMappings.hs:
```
module Kubernetes.ImportMappings (module ImportMappings) where
import Kubernetes.CustomTypes as ImportMappings (IntOrString(..))
```

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/HaskellHttpClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/HaskellHttpClientCodegen.java
@@ -35,6 +35,7 @@ import java.io.File;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


jonschoning:
cc @guoshimin

jonschoning:
ref https://github.com/OpenAPITools/openapi-generator/issues/2362

guoshimin:
We also need to have a module in the test directory to define the Arbitrary instances for the custom types.

PR #2382, opened 2019-03-13
[haskell-http-client] fix bug in test for Maybe A.Value from aa7ea8bdca

jonschoning:


guoshimin:
Thanks for fixing. Just noticed it too.

wing328:
To fix a regression in aa7ea8bdca

PR #2406, opened 2019-03-14
[haskell-http-client] add cli option `customTestInstanceModule`

jonschoning:
setting `customTestInstanceModule` will import typeclasses from the
specified module into tests/Instances.hs, to provide typeclass
instances for types not known by the generator

this property set using `--additional-properties`

example:
```
--additional-properties=customTestInstanceModule=CustomInstances
```


jonschoning:
> We also need to have a module in the test directory to define the Arbitrary instances for the custom types.
https://github.com/OpenAPITools/openapi-generator/pull/2381

c.f https://github.com/OpenAPITools/openapi-generator/issues/2362

cc @guoshimin

PR #2421, opened 2019-03-15
New generator - Scala Play Framework

adigerber:
### PR checklist

### Description of the PR

**This is work in progress!**

This PR introduces a new generator definition - Scala Play Framework.

#### Features
* [x] Models (including JSON formats)
  * [x] Basic properties (primitives)
  * [x] Container properties (map/list)
  * [x] Marking properties as required (by default properties are generated using Scala's `Option`)
  * [x] Objects with `additionalProperties`
  * [x] Definitions declared with `type: array`
  * [x] Date/time properties
  * [x] File properties _(might skip in this PR)_ **Edit: operations support files, models do not**
* [x] API
  * [x] Base traits
  * [x] Stubs
  * [x] Controllers - **Done, but needs to be tested in runtime for correctness**
  * [x] Default module (for DI)
  * [x] Routes
  * [x] Error handler
* [x] Support files
  * [x] Minimal SBT
  * [x] README
  * [x] ~LICENSE~ **removed**
  * [x] Logging
  * [x] Application configuration (including Play configuration)
  * [x] API docs
* [x] Generation flags/additional properties
  * [x] ~Generate~ Skip generation of stubs
  * [x] Generate async API
  * [x] Generate API docs
  * ~Skip routes generation~ **Edit: since routes are essential for Play apps I chose not to implement this.**
  * [x] Set routes file name
  * _Anything else I might've forgotten_
* [ ] Tests

--- 

CC (technical committee): @clasnake, @jimschubert, @shijinkui, @ramzimaalej

##### Migrated from swagger-codegen
**Note** this was originally written for swagger-codegen so there may be some remnants around in the code before this is fully complete.

Some issues that need attention:
* [x] Default values for stubs - check if given schema/property is required
* [x] When generating against `modules/openapi-generator/src/test/resources/3_0/petstore.yaml`, models `InlineObject` and `InlineObject1` are generated but are not used anywhere at the moment **Edit: removed these files by running the generator with `-DskipFormModel=true`**
* [x] Enums are generated inside their parent object (e.g. `Pet.Status`), but there is no context of the parent object in the API operations so `enumName` and `datatypeWithEnum` do not reference the enum (e.g. `Status` instead of `Pet.Status`). Couldn't find a way to inject the parent model name anywhere. **Edit: since java-play-framework does not convert enum params in operations this will not be done here either.**
* [ ] Friendly output of `JsResultException` (https://github.com/OpenAPITools/openapi-generator/pull/2421#issuecomment-476435770)


modules/openapi-generator/src/main/resources/scala-play-framework/simpleParentJsonFormat.mustache
```diff
@@ -0,0 +1,7 @@
+{{! Keep indentation at 4 spaces for proper formatting of the output. }}
```
jimschubert: I've found mustache to be inconsistent with indents in some situations. I don't remember the specifics, but I think indents may be dropped in files at an import depth greater than 1?

To resolve the issue I had created a lambda that you can reuse if the indents become an issue. See the kotlin-server generator. Example usage of the lambda: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/_api_body.mustache

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
```diff
@@ -0,0 +1,264 @@
...
+import java.io.File;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.openapitools.codegen.utils.StringUtils.camelize;
+
+public class ScalaPlayFrameworkServerCodegen extends AbstractScalaCodegen implements CodegenConfig {
+    public static final String TITLE = "title";
+    public static final String IMPL_STUBS = "implStubs";
```
jimschubert: I think this name may be confusing, as it indicates generation of stubbed implementations... but this is a default behavior in most generators that support interfaces.

I'm wondering if this is the same option in the Java Play generator, `controllerOnly`:

> Whether to generate only API interface stubs without the server files.

If so, we may want to align the option name to avoid confusion for those evaluating the Play generation with different languages. This wouldn't be necessary in this PR, though. I just wanted to document the thought.

modules/openapi-generator/src/main/resources/scala-play-framework/LICENSE.mustache
```diff
@@ -0,0 +1,8 @@
+This software is licensed under the Apache 2 license, quoted below.
```
jimschubert: We probably don't want to enforce a LICENSE in the template.

modules/openapi-generator/src/main/resources/scala-play-framework/caseClass.mustache
```diff
@@ -0,0 +1,18 @@
...
+  implicit lazy val {{classVarName}}JsonFormat: Format[{{classname}}] = {{^isExtensibleObject}}Json.format[{{classname}}]{{/isExtensibleObject}}{{#isExtensibleObject}}{
+{{>extensibleObjectJsonFormat}}
+  }{{/isExtensibleObject}}{{#vars}}{{#isEnum}}
+
+  // noinspection TypeAnnotation
+  object {{enumName}} extends Enumeration { {{#allowableValues}}{{#values}}
+    val {{.}} = Value{{/values}}{{/allowableValues}}
+
+    type {{enumName}} = Value
+    implicit lazy val {{enumName}}JsonFormat: Format[Value] = Format(Reads.enumNameReads(this), Writes.enumNameWrites[this.type])
```
jimschubert: May want to import `Format` and other Json related types at the object level here and/or above line 7, otherwise this assumes they exist in the including mustache file.

modules/openapi-generator/src/main/resources/scala-play-framework/simpleParentJsonFormat.mustache
adigerber: Thanks! I'll try experimenting more with the indents and if nothing works I'll give this method a shot.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
adigerber: `implStubs` is similar to `controllerOnly` in that they both affect whether stubs are generated, but `controllerOnly` also affects the interface in java-play-framework, while in this generator I wish to allow generating the trait regardless of the stubs. Perhaps I should make both configurable?

Also in the spirit of keeping things similar, I'll rename `apiFutures` to `supportAsync`.

modules/openapi-generator/src/main/resources/scala-play-framework/caseClass.mustache
adigerber: This is a partial that's only used in `model.mustache` which has this import. I split the model into three main partials (`arrayCaseClass.mustache`, `caseClass.mustache` and `caseObject.mustache`) in an attempt to keep it somewhat cleaner.

wing328:
From the list, looks like many items have been completed. Great work @adigerber !

What about adding the auto-generated README and then we'll release a beta version to collect feedback from the community?

adigerber:
@wing328 Sure, will do later today.

adigerber:
@wing328 I added the README, but I think there's more work to be done before even beta release, specifically - error handling (i.e. converting exceptions to suitable HTTP responses).
I've already started working on that a few days ago but haven't had the time to finish. I'll see if I can get it done within the next couple hours.

adigerber:
I've finished most of the code generation I wanted to do.  
I generated a clean sample (using `bin/scala-play-framework-petstore.sh`), which builds successfully and runs seemingly without problems (tested JSON body, query string and form parameters including file uploads).

One small fix I have planned is to correctly format `JsResultException` - when sending a request to an endpoint which expects a JSON object with required properties, if the requirements aren't met a `JsResultException` is thrown by Play's JSON library. This is caught in the generated error handler, which in turn responds with HTTP 400, but the output is not friendly:

```
JsResultException(errors:List((/photoUrls,List(JsonValidationError(List(error.path.missing),WrappedArray()))), (/name,List(JsonValidationError(List(error.path.missing),WrappedArray())))))
```

---

Apart from tests and the above fix this generator should be **ready** :-]

wing328:
@adigerber thanks. I'll take a look and merge it into master later today if no question from me.

adigerber:
@wing328 Sure. Something is preventing Circle CI from approving the PR though, so I'll see what I can do about it.

CC (technical committee; for final review): @clasnake, @jimschubert, @shijinkui, @ramzimaalej

wing328:
Tested locally and the result looks good.

wing328:
I'll update the doc and make some other minor enhancements in a separate PR

wing328:
https://github.com/OpenAPITools/openapi-generator/pull/2512 merged into master with some minor enhancements made to the Scala play generator.

wing328:
Tweet to promote the new generator: https://twitter.com/oas_generator/status/1110821644774178816

Reddit: https://www.reddit.com/r/scala/comments/b5zb53/generate_scala_play_server_stub_using_openapi/

Show HN: https://news.ycombinator.com/item?id=19497940

PR #2446, opened 2019-03-19
[typescript-axios] Fix JSON serialization of falsy request bodies

### LINKED ISSUE #2445: [BUG] [typescript-axios] empty string body isn't properly JSON serialized
With a typescript-axios generated client, doing a `POST` request which expects some JSON. Passing `""` as the body should serialize as `""` (a JSON empty string), but serializes as an empty object instead.

I won't do the whole `Bug Report Checklist` because I'll provide a fix quickly.

It is very similar to https://github.com/OpenAPITools/openapi-generator/issues/1432

#### Bug Report Checklist

hectorj:
### PR checklist

### Description of the PR

Fix #2445

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -175,7 +175,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
     {{/hasFormParams}}
     {{#bodyParam}}
             const needsSerialization = (<any>"{{dataType}}" !== "string") || localVarRequestOptions.headers['Content-Type'] === 'application/json';
-            localVarRequestOptions.data =  needsSerialization ? JSON.stringify({{paramName}} || {}) : ({{paramName}} || "");
+            localVarRequestOptions.data =  needsSerialization ? JSON.stringify({{paramName}} !== undefined ? {{paramName}} : {}) : ({{paramName}} || "");
     {{/bodyParam}}
 
             return {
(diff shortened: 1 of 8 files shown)


hectorj:
(WIP because I need to do update the Petstore sample)

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -175,7 +175,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
     {{/hasFormParams}}
     {{#bodyParam}}
             const needsSerialization = (<any>"{{dataType}}" !== "string") || localVarRequestOptions.headers['Content-Type'] === 'application/json';
-            localVarRequestOptions.data =  needsSerialization ? JSON.stringify({{paramName}} || {}) : ({{paramName}} || "");
+            localVarRequestOptions.data =  needsSerialization ? JSON.stringify({{paramName}} !== undefined ? {{paramName}} : {}) : ({{paramName}} || "");
```
macjohnny: is it correct to send an empty object when the body is undefined?
what if body is null?

hectorj: `null` is a valid JSON value

For `{}` when it's `undefined`, I have no idea so I kept the current behavior

hectorj: But you're right, an `undefined` body param should either throw an error or give an empty body :thinking:

macjohnny:

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
macjohnny: I think it is ok to treat `undefined` similar to `null` and send an empty object.

macjohnny: the more I think about it, the more I consider the `const needsSerialization = (<any>"{{dataType}}" !== "string")` part the origin of the problem, since it does not check for any other type like boolean, etc.
so the `undefined` check is a good way to send whatever is passed as body as long as it is defined

hectorj:
Not sure how to get the bot to remove the WIP label

macjohnny:
@wing328 can you help remove the WIP label?

PR #2459, opened 2019-03-20
[rust-server] Added API Key auth to rust-server

Herrera93:
### PR checklist

### Description of the PR

Added API Key authentication in rust-server for both in header and query. Still missing in cookie.

For query, since we're not available to modify `hyper` request after creating it, I added it to the `uri` created. To do this I had to move `let query_string = ...` outside of the `queryParams` block which means it'll be present even if there aren't any queryParams or query API Key. I think we can change this by creating an additional property to check for both conditions.

For header it was easier since we can use `header_mut` on request to add any new headers to it. I only added the variable `x-keyVarName` to vendor extensions in order to camel case the header.

One other downside to this is that we may need to refactor it in case of multiple API Keys in headers.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
@@ -732,6 +732,14 @@ public class RustServerCodegen extends DefaultCodegen implements CodegenConfig {
                 }
                 header.nameInCamelCase = toModelName(header.baseName);
             }
+
+            if (op.authMethods != null) {
+                for (CodegenSecurity s : op.authMethods) {
+                    if (s.isApiKey && s.isKeyInHeader) {
+                        s.vendorExtensions.put("x-apiKeyName", toModelName(s.keyParamName));
+                    }
+                }
+            }
         }
 
         return objs;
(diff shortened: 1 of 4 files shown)


Herrera93:
@bjgill Can you check this out ?

wing328:
cc Rust Techical committee: @frol (2017/07) @farcaller (2017/08) @bjgill (2017/12)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -732,6 +732,14 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
                 }
                 header.nameInCamelCase = toModelName(header.baseName);
             }
+
+            if (op.authMethods != null) {
+                for (CodegenSecurity s : op.authMethods) {
+                    if (s.isApiKey && s.isKeyInHeader) {
+                        s.vendorExtensions.put("x-keyVarName", toModelName(s.keyParamName));
```
richardwhiuk: If this is really needed, can you use a better name? e.g. `apiKeyName`?

Herrera93: I'm open to changing the name if you think that fits better.
As for the needing of this variable is so we can prevent problems like this [example](https://swagger.io/docs/specification/authentication/api-keys/) where the header is `X-API-KEY` which is the value of `keyParamName` but we wouldn't be available to use it like this:

```
// compile error
header! { (X-API-KEY, "X-API-KEY") => [String] }
```

This is the solution I came up with but if you have a better one we can use that one.

Herrera93: @richardwhiuk 
Just changed the name from `x-kayVarName` to `x-apiKeyName`

wing328:

PR #2463, opened 2019-03-20
[C] C generator refactored

zhemant:
### PR checklist

### Description of the PR

Updated mustache for model and api to cover as much possible data types
Updated samples according to new mustaches
change data type from file to binary_t to handle binary data (support for OAS3)
Updated cmake mustache

@wing328


samples/client/petstore/c/unit-tests/manual-PetAPI.c
```diff
@@ -115,15 +112,26 @@ int main() {
...
 		                                          EXAMPLE_PET_ID,
 		                                          "dec",
-		                                          file);
+		                                          data);
 
 		api_response_free(respo);
 		fclose(file);
 	} else {
-    apiClient_free(apiClient3);
+    //apiClient_free(apiClient3);
```
wing328: I would suggest removing this line as it's commented out and people have no idea why it's left there.

zhemant: I updated it, as apiClient object has to be freed once the requirement is over. Modified all test samples accordingly.

samples/client/petstore/c/unit-tests/manual-UserAPI.c
```diff
@@ -102,16 +105,19 @@ int main() {
...
 
 	UserAPI_logoutUser(apiClient4);
+    apiClient_free(apiClient4);
 
 
 // delete user test
 	apiClient_t *apiClient5 = apiClient_create();
 
 	UserAPI_deleteUser(apiClient5, "example123");
+    apiClient_free(apiClient5);
```
wing328: Minor suggestion: indent ` apiClient_free` properly with reference to the previous line.

wing328: You may simply want to use the code formatter to format the manually written test files.

zhemant: I have formatted manually written test files with uncrustify with the configuration we have in auto-generated code.

wing328:
@zhemant what about using FILE* instead? 

It seems to me FILE* is pretty common for handling files in C.

zhemant:
Yes, but this new structure is keeping in mind we also need to handle binary data which may not be a file pointer. In OAS3 binary and file are the same. And in curl, we need to pass the binary/file as data(char*/uint8_t) and its size. Hence to make it common for both data type user has to pass a struct of binary_t with data and length of data so it is easy to handle both files and binary while sending.

In current codes for handling FILE* we are doing same, reading data into a struct with char* and length.

wing328:
@zhemant thanks for the explanation, which totally makes sense.

wing328:
## Upgrade Note

Binary and file are now handled with the binary_t pointer

PR #1107, opened 2018-09-25
[KOTLIN][SPRING] Bug fix for optional / default types

dr4ke616:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.3.x`, `4.0.x`. Default: `master`.
- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

CC'ing relevant contributors:
- @jimschubert (2017/09)

### Description of the PR

fixes #1106 

Very minor changes to this PR - just template changes in the kotlin-spring generator.

Issue #1106, opened 2018-09-25
[KOTLIN][SPRING] Better Handling of Optional/Default Types


dr4ke616:
##### Description

Few very minor bugs around how default values and required fields are handled in `requestBody`s and `parameters`. In particular in correct use of nullable types (`?`). 

- Request bodies are not respecting the `required` field when provided. Therefore types are not being correctly set as null-able in controller method signatures.

```yaml
requestBodies:
  MetaInfoBody:
    required: true  # no matter if set to true or false
    content:
      application/json:
        schema:
          $ref: "#/components/schemas/MetaInfo"
```

The above definition yields the below code no matter if `required` is set to `true` of `false`.

```kotlin
bodiesPost(@RequestBody metaInfo: MetaInfo)
```

- Request parameters that have default values, the types remain null-able. For example, a request parameter such as the following, which has required `false` *but* has a default value, still set the types in the `RequestParam` as null-able.

```yaml
components:
  parameters:
    Limit:
      name: "limit"
      in: query
      required: false  # required is false, but has a default
      schema:
        type: "integer"
        minimum: 1
        maximum: 50
        default: 10
```

The above definition would yields:

```kotlin
@RequestParam(value = "limit", required = false, defaultValue = "10") limit: Int?  // <-- still null-able
```

##### openapi-generator version

Latest release - `org.openapitools:openapi-generator:2.3.2`

PR #2487, opened 2019-03-23
[Kotlin][Spring] optional types for required=false without defaultValue

markus-wa:
### PR checklist

cc: @jimschubert (2017/09)  @dr4ke616 (2018/08)

### Description of the PR

This fixes a regression ~~introduced (presumeably) by #1107 / #1106~~

The parameter types should only be non-nullable if they are either required or have a default value. Since `defaultValue` is `"null"` (a string) by default, mustache treats it as a value (not-false or unset) the `{{#defaultValue}}bla{{/defaultValue}}` branch was alway hit, never `{{^defaultValue}}....`.

This change follows a similar pattern of the C# generator which uses the helper property `isNullable` for this instead.

It also fixes `Failed to convert value of type 'java.lang.String' to required type 'java.lang.Integer'; nested exception is java.lang.NumberFormatException: For input string: "null"` if the optional parameter isn't provided in the request by removing the `defaultValue` parameter if it's `"null"`.


markus-wa:
I'm personally not quite happy with this https://github.com/OpenAPITools/openapi-generator/pull/2487/files#diff-333e74d43089b01287cc3b8a0b0f6986R509

but I've put it as ready for review so we can gather feedback on it.

I've noticed that the regression actually seems to be coming from some change that changed the default value of `defaultValue` to `"null"` (string) instead of `null`. The reason I believe this is because the old sample code didn't have the `, defaultValue="null"` here https://github.com/OpenAPITools/openapi-generator/pull/2487/files#diff-f4ba6d632a20f27265f9bcc7c462e79fR59

Unfortunately I can't find the change, the only things I found around that area seem to be from 2016 or earlier.

wing328:
> Unfortunately I can't find the change, the only things I found around that area seem to be from 2016 or earlier.

It could be related to a change I made as I did fix default values in some other generators. My apologies in advance.

Is it correct to say that the issue should be fix by changing the default value of defaultValue to `null` instead of "null" (string)? I think I can file a PR based on similar fix I did for some other generators?

(I'm aware of your workaround in https://github.com/OpenAPITools/openapi-generator/pull/2487/files#diff-333e74d43089b01287cc3b8a0b0f6986R509)

markus-wa:
Yeah changing it to `null` should do the trick. `isNullable` wouldn't be required either (I just left it in as a relic from my previous change c2f7148e7da69ca6fc6d6cc28386955393bf0182).

> I think I can file a PR based on similar fix I did for some other generators?

That would be great, would probably be much cleaner than this here 😄!

wing328:
@markus-wa I've filed https://github.com/OpenAPITools/openapi-generator/pull/2513 to fix the default value. Please review when you've time.

markus-wa:
Closing in favour of  #2513

PR #2489, opened 2019-03-23
[core] add support for model as query params

zhemant:
### PR checklist

### Description of the PR

Add support for models as query params for OAS3.

Sampleyaml : https://pastebin.com/MYk8nmdP

@wing328

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2849,8 +2849,19 @@ public class DefaultCodegen implements CodegenConfig {
             codegenParameter.vendorExtensions.putAll(parameter.getExtensions());
         }
 
-        if (parameter.getSchema() != null) {
-            Schema parameterSchema = ModelUtils.unaliasSchema(this.openAPI, parameter.getSchema());
+        Schema s;
+        if(parameter.getSchema() != null) {
+            s = parameter.getSchema();
+        } else if (parameter.getContent() != null) {
+            Content content = parameter.getContent();
+            MediaType mediaType = content.values().iterator().next();
+            s = mediaType.getSchema();
+        } else {
+            s = null;
+        }
+
+        if (s != null) {
+            Schema parameterSchema = ModelUtils.unaliasSchema(this.openAPI, s);
             if (parameterSchema == null) {
                 LOGGER.warn("warning!  Schema not found for parameter \"" + parameter.getName() + "\", using String");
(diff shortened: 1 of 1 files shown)


wing328:
@zhemant thanks for the PR 👍 

cc @OpenAPITools/generator-core-team for review

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2849,8 +2849,19 @@ public CodegenParameter fromParameter(Parameter parameter, Set<String> imports)
...
         }
 
-        if (parameter.getSchema() != null) {
-            Schema parameterSchema = ModelUtils.unaliasSchema(this.openAPI, parameter.getSchema());
+        Schema s;
+        if(parameter.getSchema() != null) {
+            s = parameter.getSchema();
+        } else if (parameter.getContent() != null) {
+            Content content = parameter.getContent();
+            MediaType mediaType = content.values().iterator().next();
```
wing328: @zhemant thanks for the enhancement. Does it mean only the first medaType will be supported? If that's the case, what about logging a warning to inform the user?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2849,8 +2849,19 @@ public CodegenParameter fromParameter(Parameter parameter, Set<String> imports)
...
+            s = mediaType.getSchema();
+        } else {
+            s = null;
```
wing328: What about logging an error so that the user is informed of something wrong?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
zhemant: Yes, only first mediaType will be supported. I will push another commit with warning logging to inform the user.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
zhemant: Incase of null, there is already logging at https://github.com/OpenAPITools/openapi-generator/blob/b4fcf6f3d60f6ac45854cd109ff2581778541b72/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L3004 about the error hence I did not add any extra logging.

wing328:

wing328:
@zhemant thanks for the PR with the enhancement. Let's go with your implementation to start with.

PR #2504, opened 2019-03-25
[Rust Server] Improve XML support

richardwhiuk:
CC: Rust technical committee - @frol @farcaller @bjgill

### PR checklist

### Description of the PR

This change significantly improves `rust-server`'s support for generating APIs with XML bodies.

- Restore XML namespace support which was lost in OpenAPI upgrade
- Remove non snake case rust warning for xml wrap_in methods
- Add XML rust-server tests
- Fix wrapping XML arrays when a property of another object

As part of this it makes a couple of core changes to Rust Server's code generator

- Run all tests, not just those for OpenAPI 2.0
- Force wrapping (aliasing) for rust-server - this is required because Rust doesn't allow implementing traits on types which the user doesn't own. Note, with wrapping disabled, the current generator creates code which doesn't compile.

The majority of this code was originally written by @bjgil. Additional fixes were contributed by myself, and others at @metaswitch.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -64,6 +67,9 @@
     public RustServerCodegen() {
         super();
 
+        // Force generating models for wrapping types
+        ModelUtils.setGenerateAliasAsModel(true);
```
wing328: @richardwhiuk thanks for the PR. Do you mind explaining why we must enforce this option to true?

Some users simply want to reuse the same array definition through out the spec with aliases.

richardwhiuk: Sorry, this should only apply to the Rust Generator - I need to figure out where to move it to so it's only invoked when the Rust generator is in use.

Currently, the generator assumes that the model will exist when generating the API bindings.

This means that it currently generates compile errors like:

```
error[E0412]: cannot find type `AnotherXmlArray` in module `models`
  --> output/openapi-v3/src/lib.rs:94:52
   |
94 |     fn xml_other_put(&self, string: Option<models::AnotherXmlArray>, context: &C) -> Box<Future<Item=XmlOtherPutResponse, Error=ApiError>>;
   |                                                    ^^^^^^^^^^^^^^^ did you mean `AnotherXmlInner`?
```

(for non aliasing, that should be `Option<Vec<....>>`.)

Worse, this is a backwards compatibility break - previously the generator did generate aliasing - #1729 changes this so the default is not to generate aliasing, meaning that the 4.x branch generates Rust code that doesn't compiler.

Longer term we can look at enhancing the Rust Server code generator to support generating code which compiles when aliasing is disabled (though unit structs in the manner that the Rust generator generates are fairly idiomatic, and allow for additional methods to be declared on them).

wing328: > Sorry, this should only apply to the Rust Generator - I need to figure out where to move it to so it's only invoked when the Rust generator is in use.

Currently, it's already the case, which means only the Rust server generator will have this option enabled.

Thanks for the detailed explanation. We can certainly support generating array/map as model later on.

wing328:
The CircleCI failure is related to outdated samples. I'll update those after merging this PR into master.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
wing328: > Currently, it's already the case, which means only the Rust server generator will have this option enabled.

I know what you mean now. I'll need to uncomment this line and show a warning message for the time being.

wing328: Filed https://github.com/OpenAPITools/openapi-generator/pull/2714

richardwhiuk: Yeah, I think the constructor must get run at startup, which means the other generators pick this up.

I'm slightly surprised we've defaulted `generateAliasAsModel` to false - this seems like the reverse of the default in 3.x?

wing328: 4.x is a branch with breaking changes (with or without fallback) and the main reason is that most generators do not support alias (array/map) as model so that's we set `generateAliasAsModel` to false as the default.

PR #2535, opened 2019-03-28
dart2: do not send null fields in the json request

ayushin:
### PR checklist

- [X ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.

### Description of the PR

Do not encode properties which are "null" in the model object in json and omit them instead.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1430,7 +1430,7 @@ public class DefaultCodegen implements CodegenConfig {
         } else if (ModelUtils.isDateTimeSchema(schema)) {
             return "DateTime";
         } else if (ModelUtils.isNumberSchema(schema)) {
-            if (schema.getFormat() == null) { // no format defined
+            if (schema.getFormat() == null || schema.getFormat().equals("decimal")) { // no format defined
                 return "number";
             } else if (ModelUtils.isFloatSchema(schema)) {
                 return SchemaTypeUtil.FLOAT_FORMAT;
--- a/modules/openapi-generator/src/main/resources/dart2/class.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/class.mustache
@@ -16,9 +16,7 @@ class {{classname}} {
   {{classname}}.fromJson(Map<String, dynamic> json) {
     if (json == null) return;
   {{#vars}}
-    if (json['{{baseName}}'] == null) {
-      {{name}} = null;
-    } else {
+    if (json['{{baseName}}'] != null) {
   {{#isDateTime}}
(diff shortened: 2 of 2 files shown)


wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1430,7 +1430,7 @@ private static String getPrimitiveType(Schema schema) {
         } else if (ModelUtils.isDateTimeSchema(schema)) {
             return "DateTime";
         } else if (ModelUtils.isNumberSchema(schema)) {
-            if (schema.getFormat() == null) { // no format defined
+            if (schema.getFormat() == null || schema.getFormat().equals("decimal")) { // no format defined
```
wing328: I suggest not to do this in this PR as this changing the defaul odegen will impact all other generators.

My recommendation is to discuss this further on how we support more formats.

modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -75,6 +73,8 @@ class {{classname}} {
       {{/isDateTime}}
     {{/vars}}
     };
+    json.removeWhere((key, value) => value == null);
```
wing328: In the future, we want to support nullable so this will not work well with nullable property.

Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-20

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
ayushin: yes, actually this slipped in by mistake

modules/openapi-generator/src/main/resources/dart2/class.mustache
ayushin: yes, I agree. so I guess this should be an option, like -DomitNullFields?

wing328: Are you free for a quick chat this week? Please ping me via https://gitter.im (ID: wing328) when you've time.

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/2777

PR #2546, opened 2019-03-29
Deprecate initialCaps and replace with camelize

wing328:
### PR checklist

### Description of the PR


jimschubert:
Can I propose that we separate TitleCase and camelCase from `camelize` into two separate methods? I assume `initialCaps` existed because many don't realize that `camelize(value)` is title cased and `camelize(value, boolean=false)` is camelCase.

wing328:
>  I assume initialCaps existed because many don't realize that camelize(value) is title cased and 

Based on what I recall, initialCaps was already in the code base before I added camelize and underscore a while go. The implementations of initialCaps does not fully meet our requirement as it only make the first letter upper case and that's why we want to replace it with `camelize`

> Can I propose that we separate TitleCase and camelCase from camelize into two separate methods?

Good idea. What about introducing the following methods and mark `camelize`, `underscore` as private?

- toSnakeCase
- toTitleCase
- toCamelCase
- toKebabCase

jimschubert:
@wing328 I think the four methods you've listed would be very clear, and easily discoverable by a good IDE.

I remember looking for `titleCase` and `kebabCase` when I first started contributing, a few years ago and didn't see these options. After a short while, I realized the usage of the current methods… but I think the above would be easier for new contributors.

👍

wing328:
@jimschubert 👌 I'll do that in a separate PR.

Issue #2516, opened 2019-03-26
[BUG] Generating Java Client produces NULLUNIQUENAME.java classes


ncroma:
##### Description

By generating a Java Client using the yaml file posted below, I get a NULLUNIQUENAME.java as the model class for the LoginApi.java method's request-body parameter - `login(List<NULLUNIQUENAME> NULLUNIQUENAME)` - instead of a InlineObject.class.

##### openapi-generator version

I'm using the docker command posted bellow, I would assume it is using the last generator version.

##### OpenAPI declaration file content or url

```
openapi: 3.0.0
info:
  version: 1.0.0
  title: My Api
tags:
  - name: login
paths:
  /api/login:
    post:
      operationId: login
      summary: Logs in
      tags:
        - login
      requestBody:
        content:
          application/json:
            schema:
              type: array
              items:
                type: object
                properties:
                  prop:
                    type: string
servers:
  - url: http://localhost:3000/
```

##### Command line used for generation

I'm using this docker command:
`docker run --rm -v %cd%/api:/output -v %cd%:/local openapitools/openapi-generator-cli generate -i /local/oas-doc.yaml -g java -o /output`

Zomzog:
@wing328 can we close this issue?

PR #2569, opened 2019-04-01
Change default inline object name

Zomzog:
### PR checklist

### Description of the PR

Default name changed for more usable one. For #2516

@wing328
@jimschubert
@cbornet
@ackintosh
@jmini

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -409,7 +409,7 @@ public class InlineModelResolver {
     private String uniqueName(String key) {
         if (key == null) {
             key = "InlineObject";
-            LOGGER.warn("Found an inline schema without the `title` attribute. Default the model name to InlineObject instead. To have better control of the model naming, define the model separately so that it can be reused through out the spec.");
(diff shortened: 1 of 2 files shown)


wing328:
@Zomzog Thanks for the PR. I agree InlineObject is better than what we've right now.

May I suggest we also revise the warning message as follows?
```
LOGGER.warn("Found an inline schema without the `title` attribute. Default the model name to InlineObject instead. To have better control of the model naming, define the model separately so that it can be reused through out the spec.");
```

Zomzog:
Good idea, I've updated the warning message.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -408,8 +408,8 @@ private void addGenerated(String name, Schema model) {
 
     private String uniqueName(String key) {
         if (key == null) {
-            key = "NULL_UNIQUE_NAME";
-            LOGGER.warn("null key found. Default to NULL_UNIQUE_NAME");
+            key = "InlineObject";
+            LOGGER.warn("Found an inline schema without the `title` attribute. Default the model name to InlineObject instead. To have better control of the model naming, define the model separately so that it can be reused through out the spec.");
```
wing328: Sorry please make one more minor change
`through out` => `throughout` (one word)

Zomzog: changed

wing328:
The CI failures is not caused by this change.

PR #2578, opened 2019-04-02
[typescript-axios] Fix Bearer authentication

### LINKED ISSUE #1446: [typescript-axios] Generating bearerAuth create Basic auth instead
##### Description

Attempting to generate from an openapi definition with bearerAuth marked as the security schema. When it generates using axios-typescript it generates Basic auth security instead? It does generate a comment referencing Bearer auth but no code to add the header.

"@harmowatch/openapi-generator-cli": "0.0.3-3.3.2"

```yaml
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

This generates
```typescript
            // authentication bearerAuth required
            // http basic authentication required
            if (
                configuration &&
                (configuration.username || configuration.password)
            ) {
                localVarHeaderParameter["Authorization"] =
                    "Basic " +
                    btoa(configuration.username + ":" + configuration.password);
            }
```

Generating through npm.
npm=6.2.0
node=v10.9.0
"generate": "openapi-generator generate -i api/myapi.yaml -g typescript-axios -o generated-sources/myapi"

##### Steps to reproduce

Running generate commmand.

hectorj:
### Description of the PR

[typescript-axios] Fix Bearer authentication

It was generated the same as Basic authentication. Now we handle both cases differently.

Fix #1446

### PR checklist

### Technical Committee

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -74,12 +74,18 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             }
             {{/isKeyInQuery}}
             {{/isApiKey}}
-            {{#isBasic}}
+            {{#isBasicBasic}}
             // http basic authentication required
             if (configuration && (configuration.username || configuration.password)) {
                 localVarHeaderParameter["Authorization"] = "Basic " + btoa(configuration.username + ":" + configuration.password);
             }
-            {{/isBasic}}
+            {{/isBasicBasic}}
+            {{#isBasicBearer}}
+            // http bearer authentication required
+            if (configuration && configuration.password) {
+                localVarHeaderParameter["Authorization"] = "Bearer " + btoa(configuration.password);
+            }
+            {{/isBasicBearer}}
             {{#isOAuth}}
             // oauth required
             if (configuration && configuration.accessToken) {
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -82,8 +82,11 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             {{/isBasicBasic}}
             {{#isBasicBearer}}
             // http bearer authentication required
-            if (configuration && configuration.password) {
-                localVarHeaderParameter["Authorization"] = "Bearer " + btoa(configuration.password);
+            if (configuration && configuration.accessToken) {
+                const accessToken = typeof configuration.accessToken === 'function'
+                    ? configuration.accessToken()
+                    : configuration.accessToken;
+                localVarHeaderParameter["Authorization"] = "Bearer " + accessToken;
             }
             {{/isBasicBearer}}
             {{#isOAuth}}


hectorj:
There does not seem to be any tests covering this yet, but I'm not sure where I could put it.

Templates seem to be tested via the petstore samples, but it does not contain Bearer authentication. Is it ok to add it to the petstore (which means re-generating all the samples) ? Are there some other samples ? I'm open to suggestions for this.

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -74,12 +74,18 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
...
             // http basic authentication required
             if (configuration && (configuration.username || configuration.password)) {
                 localVarHeaderParameter["Authorization"] = "Basic " + btoa(configuration.username + ":" + configuration.password);
             }
-            {{/isBasic}}
+            {{/isBasicBasic}}
+            {{#isBasicBearer}}
+            // http bearer authentication required
+            if (configuration && configuration.password) {
+                localVarHeaderParameter["Authorization"] = "Bearer " + btoa(configuration.password);
```
macjohnny: wouldn't it be better to use the `configuration.accessToken`?
https://github.com/OpenAPITools/openapi-generator/blob/0889b8ec1349d6f912ad7dfa221d3980f1801dba/modules/openapi-generator/src/main/resources/typescript-axios/configuration.mustache#L34-L40

this is also used e.g. in the angular client:
https://github.com/OpenAPITools/openapi-generator/blob/0889b8ec1349d6f912ad7dfa221d3980f1801dba/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L213-L220

hectorj: You're right, will fix

hectorj: fixed in https://github.com/OpenAPITools/openapi-generator/pull/2578/commits/dbc340061bd57fd2bb5f6442d7421d2ded561653

thanks

macjohnny:
> There does not seem to be any tests covering this yet, but I'm not sure where I could put it.
> 
> Templates seem to be tested via the petstore samples, but it does not contain Bearer authentication. Is it ok to add it to the petstore (which means re-generating all the samples) ? Are there some other samples ? I'm open to suggestions for this.

you could check if there is a separate petstore yaml file (e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml) that covers the case you want to test and use that to create separate samples, e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/bin/typescript-angular-v7-petstore-provided-in-root.sh#L30

hectorj:
Some samples seem to use the `2_0` petstore while some others use the `3_0`. I don't think I saw any language using both. How is it chosen ?

Should we replace the typescript-axios samples with the `3_0` version or add it as a new sample ?
If we add it as new, do we want to redo all of them (with the various options), or just the default one ?

(thank you for your help BTW, I'm still discovering the internals of the project :slightly_smiling_face: )

macjohnny:
good questions. to be honest: I don't know how to answer them.

maybe @wing328 can give us a hint?

wing328:
> Some samples seem to use the 2_0 petstore while some others use the 3_0. I don't think I saw any language using both. How is it chosen ?

PHP, Ruby and a few more have both Petstore samples (under different folders) for Petstore OAS v2,  v3 spec. These generators have better support for new features in OAS v3 (e.g. nullable, allOf, etc)

If the TS generators are supporting these new features, we should definitely generate the OAS v3 Petstore samples and create some integration tests accordingly.

Issue #2577, opened 2019-04-02
[BUG] Maven plugin: skipValidateSpec option


adigerber:
##### Description

* `skipValidateSpec` affects the wrong option - `skipOverwrite`.
* Why even have `skipValidateSpec` if there's already `validateSpec` which supposedly does the same thing?

##### openapi-generator version

Master branch - 9c7d407

##### Suggest a fix

Remove `skipValidateSpec` completely in favor of `validateSpec`, or change it to call `setValidateSpec` instead of `setSkipOverwrite`.

jimschubert:
the `setSkipOverwrite` issue was fixed via #2583

> Why even have skipValidateSpec if there's already validateSpec which supposedly does the same thing?

`skipValidateSpec` is more desirable here because it aligns configuration across CLI/Gradle/Maven, where we are forced to the truthy option by the CLI's option parser framework (we can't easily or cleanly pass `validateSpec=false`). The option also implicitly conveys that we validate the spec by default.

jimschubert:
see #2614

adigerber:
Thanks for the fixes @jimschubert and sorry for the late response. Closing this.

PR #2583, opened 2019-04-02
Remove SkipValidSpec option

### LINKED ISSUE #2319: [BUG] Gradle plugin, value of validateSpec is overridden by skipValidateSpec
##### Description
Please see:
https://github.com/OpenAPITools/openapi-generator/blob/b302c16f74fdbf8d6034b924e0906ad60b4d5d29/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt#L528

```kotlin
            skipValidateSpec.ifNotEmpty { value ->
                configurator.setValidateSpec(value)
            }
```
as a result validateSpec is set as value of skipValidateSpec, as by default skipValidateSpec == false it means that validateSpec will be assigned false, and build will not fail even in case of errors in provided swagger / openApi specification.

4.0.0-beta2

./gradlew openApiGenerate

##### Steps to reproduce
Run openApiGenerate for invalid swagger/openAPI specification, it should continue even after reporting that errors where found:
There were issues with the specification, but validation has been explicitly disabled.

##### Suggest a fix
Please rewrite it as:
```kotlin
            skipValidateSpec.ifNotEmpty { value ->
                configurator.setSkipValidateSpec(value)
            }
```

Zomzog:
### PR checklist

### Description of the PR

It looks like `skipValidSpec` try to do the same thing than `validateSpec` so I've removed it because I think it can create wired behavior if both are set. 

But maybe I've missed something and I can do another solution.

Must fix #2319 and #2577

@wing328
@jimschubert
@cbornet
@ackintosh
@jmini

### DIFF
--- a/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt
+++ b/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt
@@ -525,7 +525,7 @@ open class GenerateTask : DefaultTask() {
             }
 
             skipValidateSpec.ifNotEmpty { value ->
-                configurator.setValidateSpec(value)
+                configurator.setValidateSpec(!value)
             }
 
             generateAliasAsModel.ifNotEmpty { value ->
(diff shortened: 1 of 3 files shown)


jimschubert:
`skipValidateSpec` is required in CLI as I don't believe there's a way to pass `false` to a Boolean option.

The others (Gradle/Maven) had this property for similar configuration across the entry points. In Gradle, it seems the issue could be resolved by modifying the condition where the property is set.

It seems this PR completely removed the CLI option for skipping.

Zomzog:
OK I will try it but in this case does the solution is to remove `validateSpec`?

Zomzog:
I've changed the PR to fix `skipValidateSpec`. 
I think the question is still open, can `validateSpec` be deprecated or removed?

jimschubert:
> I think the question is still open, can validateSpec be deprecated or removed?

I think the maven plugin could be cleaned up to avoid having both, and for consistency it should follow the `skipValidateSpec`.

jimschubert:
@Zomzog I opened #2614 to address the remaining issue in the maven plugin.

PR #2611, opened 2019-04-05
[PetStore Spec] Fetch petstore.swagger.io in HTTPS

karismann:
### PR checklist

### Description of the PR

petstore sample spec should be fetched in HTTPS by default
furthermore, actual specs can't be tested with "try it out" option in swagger editor util cause of insecure request , fixed by this PR

swaggereditor-errorHTTP


wing328:
@karismann thanks for the PR but the CIs are testing against a Petstore server running locally (127.0.0.1) which doesn't have any SSL cert setup so I would suggest keep using http instead of https

(host tables are updated to map petstore.swagger.io to 127.0.0.1)

karismann:
thanks for that explanation @wing328

PR #2629, opened 2019-04-09
[aspnetcore] Support async tasks and some code cleanups

A-Joshi:
### PR checklist

### Description of the PR

Support for async and Task for both library and program
Support for Controller and Model namespaces.
Small fixes and code generation clean up

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
@@ -261,6 +261,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
         setOperationModifier();
         setModelClassModifier();
         setUseSwashbuckle();
+        setOperationIsAsync();
 
         // CHeck for class modifier if not present set the default value.
         additionalProperties.put(PROJECT_SDK, projectSdk);
@@ -475,6 +476,17 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
             useSwashbuckle = convertPropertyToBooleanAndWriteBack(USE_SWASHBUCKLE);
         } else {
             additionalProperties.put(USE_SWASHBUCKLE, useSwashbuckle);
+        }
+    }
+
+    private void setOperationIsAsync() {
+        if (isLibrary) {
+            operationIsAsync = false;
+            additionalProperties.put(OPERATION_IS_ASYNC, operationIsAsync);
+        } else  if (additionalProperties.containsKey(OPERATION_IS_ASYNC)) {
+            operationIsAsync = convertPropertyToBooleanAndWriteBack(OPERATION_IS_ASYNC);
+        } else {
+            additionalProperties.put(OPERATION_IS_ASYNC, operationIsAsync);
         }
     }
 }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
@@ -409,8 +409,6 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
             operationModifier.setOptValue(classModifier.getOptValue());
             additionalProperties.put(OPERATION_MODIFIER, operationModifier.getOptValue());
             LOGGER.warn("classModifier is " + classModifier.getOptValue() + " so forcing operatonModifier to " + operationModifier.getOptValue());
-        //} else {
-            //setCliOption(operationModifier);
         }
     }
 
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
@@ -274,7 +274,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
         }
 
         if (!additionalProperties.containsKey(CodegenConstants.MODEL_PACKAGE)) {
-            modelPackage = packageName + ".Controllers";
+            modelPackage = packageName + ".Models";
             additionalProperties.put(CodegenConstants.MODEL_PACKAGE, modelPackage);
         }
 
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
@@ -39,13 +39,17 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
     public static final String ASPNET_CORE_VERSION = "aspnetCoreVersion";
     public static final String CLASS_MODIFIER = "classModifier";
     public static final String OPERATION_MODIFIER = "operationModifier";
+    public static final String OPERATION_IS_ASYNC = "operationIsAsync";
+    public static final String OPERATION_RESULT_TASK = "operationResultTask";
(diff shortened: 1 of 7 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -362,8 +409,8 @@ private void setClassModifier() {
             operationModifier.setOptValue(classModifier.getOptValue());
             additionalProperties.put(OPERATION_MODIFIER, operationModifier.getOptValue());
             LOGGER.warn("classModifier is " + classModifier.getOptValue() + " so forcing operatonModifier to " + operationModifier.getOptValue());
-        } else {
-            setCliOption(operationModifier);
+        //} else {
+            //setCliOption(operationModifier);
```
wing328: @A-Joshi minor suggestion: what about completely remove these 2 lines of commented code?

A-Joshi: Removed.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -342,15 +381,23 @@ public String getNullableType(Schema p, String type) {
...
     private void setCliOption(CliOption cliOption) throws IllegalArgumentException {
         if (additionalProperties.containsKey(cliOption.getOpt())) {
-            cliOption.setOptValue(additionalProperties.get(cliOption.getOpt()).toString());
-            if (cliOption.getOptValue() == null) {
-                cliOption.setOptValue(cliOption.getDefault());
-                throw new IllegalArgumentException(cliOption.getOpt() + ": Invalid value '" + additionalProperties.get(cliOption.getOpt()).toString() + "'" +
-                        ". " + cliOption.getDescription());
+            // TODO Hack - not sure why the empty strings become boolean.
+            Object obj = additionalProperties.get(cliOption.getOpt());
+            if (cliOption.getType() != SchemaTypeUtil.BOOLEAN_TYPE) {
```
wing328: @A-Joshi What about using `SchemaTypeUtil.BOOLEAN_TYPE.equals(cliOption.getType())` instead?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -130,6 +140,14 @@ public AspNetCoreServerCodegen() {
...
                 sourceFolder);
 
+        addOption(CodegenConstants.API_PACKAGE,
+                CodegenConstants.API_PACKAGE_DESC,
+                apiPackage);
+
+        addOption(CodegenConstants.MODEL_PACKAGE,
+                CodegenConstants.MODEL_PACKAGE_DESC,
+                modelPackage);
+
```
wing328: @A-Joshi If I remember it correctly, these are global options so we shouldn't need to add it to the generator manually?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -407,4 +465,26 @@ private void setAspnetCoreVersion(String packageFolder) {
         }
         additionalProperties.put(COMPATIBILITY_VERSION, compatibilityVersion);
     }
+
+    private void setUseSwashbuckle() {
+        if (isLibrary) {
+            useSwashbuckle = false;
+        }
```
jimschubert: Looks like this requires an else block.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -342,15 +381,23 @@ public String getNullableType(Schema p, String type) {
 
     private void setCliOption(CliOption cliOption) throws IllegalArgumentException {
         if (additionalProperties.containsKey(cliOption.getOpt())) {
-            cliOption.setOptValue(additionalProperties.get(cliOption.getOpt()).toString());
-            if (cliOption.getOptValue() == null) {
-                cliOption.setOptValue(cliOption.getDefault());
-                throw new IllegalArgumentException(cliOption.getOpt() + ": Invalid value '" + additionalProperties.get(cliOption.getOpt()).toString() + "'" +
-                        ". " + cliOption.getDescription());
+            // TODO Hack - not sure why the empty strings become boolean.
+            Object obj = additionalProperties.get(cliOption.getOpt());
```
jimschubert: I'm not sure how I feel about a generator setting a CLI option. There shouldn't be a need to sync options to CLI options, because it means those options can change after the point that they're expected to be read and used to configure the generator. 

What logical flow did you find a need to keep these on sync? Ideally, only the CLI should be there owner of seeing these values (to remove confusion described above).

jimschubert:
Sorry I hadn't seen this PR, been pretty busy the past couple of weeks. Aside from a handful of comments, this looks good. It seems almost identical to the previous work but with some cleanup to cli switches.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Did not realize that - removed the additional code.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Its really a hack - the generator seems to change empty strings to the boolean false -its actually to fix that. I will open an issue to see why that happens but fixing it will require care in case other generators depend on this conversion.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Added an else and also a warning to indicate that useSwashbuckle default is different when building a library.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Good point - changed.

jimschubert:
Thanks for the update. Looks like all my comments were addressed. @wing328 mind taking a quick look and merging?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -182,6 +196,21 @@ public AspNetCoreServerCodegen() {
                 "Generates method body.",
                 generateBody);
 
+        addSwitch(OPERATION_IS_ASYNC,
+                "Set methods to async or sync.",
+                operationIsAsync);
```
wing328: @A-Joshi What about reusing `asyncServer` option in Nancyfx generator?
```
	asyncServer
	    Set to true to enable the generation of async routes/endpoints. (Default: false)
```
Ref: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CSharpNancyFXServerCodegen.java#L98

A-Joshi: I thought about it but it is the operation (method) which is async not necessarily the server. In the future I can see this extended to have a list of operations which are async while leaving others sync etc.

wing328: 👌

PR #2658, opened 2019-04-14
Fix ServiceLoader initialization

adigerber:
### PR checklist

### Description of the PR

This PR fixes the service loader initializations for `CodegenConfig` so that the correct class loader is used, regardless of the thread's context class loader.

This essentially fixes an issue that occurs when calling `CodegenConfigurator.toClientOptInput` from a thread that doesn't use the same class loader which is used to load `openapi-generator` module (e.g. build tool plugins), and as a result, generator names cannot be specified.

There's still an issue with code generation itself but that is related to the `swagger-parser` module; I'll file another PR there to fix that as well.


jmini:
Thank you a lot for this PR!

PR #1850, opened 2019-01-08
Something in the dependencies changed. This switch is no longer needed.

Drezil:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Implemented fix for https://github.com/OpenAPITools/openapi-generator/issues/1849

PR #2676, opened 2019-04-16
[haskell-servant] Fix parsing of special characters

f-f:
### PR checklist

### Description of the PR

This reverts the change from #1850, as that patch broke the generator for our APIs.

In fact, after updating to upstream our service wanted to parse the following JSON:

```
[
  {
    "subscriptionStartDate": "2016-07-22",
    "subscriptionEndDate": "2016-07-22",
  }
]
```

..as this, which doesn't look correct:

```
[
  {
    "subscription*tDate": "2016-07-22",
    "subscriptionEndDate": "2016-07-22",
  }
]
```

The change was motivated as "a dependency upstream changed", but the only dependency involved here is `text`, and the signature of `Data.Text.replace` didn't change since the last time I updated this piece of code (version was [`1.2.3.0`](https://hackage.haskell.org/package/text-1.2.3.0), now we're at `1.2.3.1`)

Also it looks like the change started to ignore the quote (`'`) before the symbols to replace, which is purposefully there, so that the special characters are substituted only for name mangling, and not in every case as it happens here.
So this makes me guess that this was just a misunderstanding of the meaning of that code, so I added a comment on top of the function to explain better what's going on in there.

/cc @wing328 @jonschoning @algas @Drezil


Drezil:
This is very interesting. We use this in our company to generate stubs and had exactly the same problem, which motivated the change - and pushed the patch that resolved our problems.

We work against an api that has "Percentage" as a field and that generated json as "%age" - similar to your "Start" <-> "*t"..

I will try your version later today if i got some time to better see where the problem is actually located.

f-f:
> We work against an api that has "Percentage" as a field and that generated json as "%age" - similar to your "Start" <-> "*t"..

@Drezil this sounds like a bug, because only a field called `'Percentage` (note the quote there) should be translated to `"%age"`
However from the code here in this PR I don't see how this replacement without the quote can happen, so I'm definitely interested in a replication

wing328:
Given that there's no further question, I've merged this fix into master.

Thanks for the fix by @f-f

f-f:
Thank you! 😊

Drezil:
@f-f  After updating to the latest official version i still have issues with this. Maybe we can figure it out together.
At least i now have to make time to fix this for real.
I made a minimal example: https://github.com/Drezil/openapi_haskell_minimal_example

swagger definition is: https://github.com/Drezil/openapi_haskell_minimal_example/blob/master/swagger.yaml

```bash
$ > curl -X POST http://localhost:8080/foo -d '{"fooStartDate":"foo","bar_StartDate":"bar"}' -H "accept: application/json" -H "Content-Type: application/json"
Error in $: key "barUnderscoreStartDate" not present⏎
$ > curl -X POST http://localhost:8080/foo -d '{"fooStartDate":"foo","barUnderscoreStartDate":"bar"}' -H "accept: application/json" -H "Content-Type: application/jso
n"
{"result":["bar"],"classUnderscorenames":["foo"]}⏎
```
Connecting with the Haskell-client to the Haskell-server works fine, as both use "Underscore" instead of the "_" defined in the swagger-api. But with that swagger-ui or other language server/clients still break.

Your problem with `Start` & `*t` is still fixed in this version, but the generated/consumed code for the dual problem i tried to solve is still present.
I have to program against an external api using underscores in field-names.

I also tried to dig into Aeson to look for answers. I don't understand how https://github.com/f-f/openapi-generator/blob/064e299caa6279ec5230ee897e4fadb25f23009c/modules/openapi-generator/src/main/resources/haskell-servant/Types.mustache#L84-L87 could have any effect, because it is basically a `String -> String`-Conversion, converting things like "class_name" into "classUnderscoreName" (the haskell-type generated) and vice-versa.
The change i did let me parse "class_name" correctly, but generated "*t" from a field "Start" - which caused your problem.

I don't see a solution how we could generate fields like "my*" while also being able to generate "myStar" - at least not with what we have now.

I think the solution is in some incantation of the Aeson-generic-derivation.
Maybe we need http://hackage.haskell.org/package/aeson-1.4.3.0/docs/Data-Aeson-Types.html#v:fieldLabelModifier for parsing & http://hackage.haskell.org/package/aeson-1.4.3.0/docs/Data-Aeson-Types.html#v:constructorTagModifier for generating? Or vice-versa?

Or we need to generate much more complicated code with explicit replacement for each case defined?

f-f:
@Drezil sorry for delay, I just came back from vacation so now I have some time to look at your repro. I'll get back to you as soon as possible

f-f:
@Drezil I realized I never got back to you on this, sorry for that. Unfortunately I'm not involved anymore with projects using this so I won't have time to continue working on it for now

wing328:
@f-f that's ok. We truly appreciate your contributions to this project

@Drezil please open a new issue to track https://github.com/OpenAPITools/openapi-generator/pull/2676#issuecomment-499435065 if you've not done so.

PR #2710, opened 2019-04-21
Link query parameter to model object

### LINKED ISSUE #2655: [BUG] Query Parameter Serialization (Spring)
##### Description

Given a query param complex object.

using v4.0.0-SNAPSHOT: parameter class is generated but is not linked in the rest-controller. Also to works correctly is necessary remove @RequestParam annotation when style=form & explode: true 

4.0.0beta2

```yaml
parameters:
        - in: query
          name: pageQuery
          schema:
            type: object
            properties:
              offset:
                type: integer
                format: int32
              limit:
                type: integer
                format: int32
```
 -->

```
java -jar openapi-generator-cli-4.0.0-SNAPSHOT.jar generate -i openapi.yaml -g spring -o test --additional-properties=library=spring-boot,beanValidations=true,reactive=true
```

##### Steps to reproduce

Create a open api specification with a query param with complex object as described above.

##### Suggest a fix

Using version 4.0.0 version should link generated pojo class in rest-controller instead of "object" also remove @RequestParam annotation to works correctly.

Zomzog:
### PR checklist

### Description of the PR

Must fix #2655

I have changed the code in defaultCodgen. Maybe it's too low level and must be done in abstractJava? 

Another solution is to create another `fromProperty` for `fromParameter` with more parameters and force `complexType` with the $ref. But I'm afraid of losing all language-specific overriding.

@wing328 (2015/07) 
@jimschubert (2016/05) 
@cbornet (2016/05)
@ackintosh (2018/02) 
@jmini (2018/04)


jimschubert:
I've evaluated this locally, and it appears to _almost_ fix the linked issue.

Generating with the following from this branch:

```
openapi-generator generate -g spring \
  -i modules/openapi-generator/src/test/resources/3_0/objectQueryParam.yaml \
  -o .bak/2655 \
  --additional-properties=library=spring-boot,beanValidations=true,reactive=true \
  --skip-validate-spec
```

And modifying the output in .bak/2655 for PonyApi.java to remove the `@RequestParam` as requested in the linked issue:

```
    @ApiOperation(value = "", nickname = "list", notes = "", tags={  })
    @ApiResponses(value = {  })
    @RequestMapping(value = "/pony",
        method = RequestMethod.GET)
    default Mono<ResponseEntity<PageQuery>> list(@ApiParam(value = "", defaultValue = "null") @Valid PageQuery pageQuery, ServerWebExchange exchange) {
        exchange.getResponse().setStatusCode(HttpStatus.OK);
        return Mono.just(new ResponseEntity(pageQuery, HttpStatus.OK));
    }
```

starting via `mvn spring-boot:run` and curling:

```
curl -X GET \
  'http://localhost:8080/pony?offset=1&limit=2' \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache'
```

I have the input parameters echoed as seems to be expected by the linked issue: `{"offset":1,"limit":2}`.

Should the `@RequestParam` modification be done as part of this PR or as a separate one?

Zomzog:
Thx for the review, I will try to fix the `@RequestParam` in this PR.

Zomzog:
@jimschubert 
I've fixed the @RequestParam, but there are still issues with typescript generator. I'm not sure if I must fix them on typescript generator or limit my modification to java generator.

wing328:
@Zomzog Can you please resolve the merge conflicts when you've time?

wing328:
Looks like the change results in additional incorrect import in TS generators:
```
--- a/samples/client/petstore/typescript-angular-v2/npm/api/pet.service.ts
+++ b/samples/client/petstore/typescript-angular-v2/npm/api/pet.service.ts
@@ -21,6 +21,8 @@ import { Observable }                                        from 'rxjs/Observab
 import '../rxjs-operators';
 
 import { ApiResponse } from '../model/apiResponse';
+import { Array&lt;&#39;available&#39; | &#39;pending&#39; | &#39;sold&#39;&gt; } from '../model/arrayavailablePendingSold';
+import { Array&lt;string&gt; } from '../model/arraystring';
 import { Pet } from '../model/pet';
```

My guess is that it's due to https://github.com/OpenAPITools/openapi-generator/pull/2710/files#diff-7cb46fa53f89a458a7b7cb201d2214a8R3059

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3023,6 +3024,7 @@ public CodegenParameter fromParameter(Parameter parameter, Set<String> imports)
             String parameterDataType = this.getParameterDataType(parameter, parameterSchema);
             if (parameterDataType != null) {
                 codegenParameter.dataType = parameterDataType;
+                imports.add(parameterDataType);
```
wing328: Should the import below (line 3059) already do the job?

Zomzog: I've changed my solution for using line 3059. It looks better.

jimschubert:
I think this looks great, thank you!

mikhail-putilov:
Hey guys, looks like this PR broke generation of my `api.yml`:
Previously I had:
```
parameters:
  - name: after_last_modified_date
    in: query
    description: >
      timestamp in ISO...
    schema:
      type: string
      format: date-time
      example: "2019-06-17T15:53:00+01:00"
```
And it led to the generation of the following:
```
@ApiParam(value = "timestamp in ISO ...") @Valid @RequestParam(value = "after_last_modified_date", required = false) OffsetDateTime afterLastModifiedDate
```

And after this PR now the generated code lacks `@RequestParam`:
```
@ApiParam(value = "timestamp in ISO ...") @Valid OffsetDateTime afterLastModifiedDate
```

mikhail-putilov:
I believe `SpringCodeGen` have no notion of `HandlerMethodArgumentResolver`. That creates all problems with adding such arguments like `Pageable pageable` and `UriComponentsBuilder uriComponentsBuilder`. As well as there is no way to generate `throws URISyntaxException` which occurs quite naturally when one wants to return `ResponseEntity.created(...)` from a controller.

Ofc there is always a way to bypass all small problems. But gosh, that's probably the main problem why I can't just introduce API-first approach with generation of code out of OpenAPI spec for my current company. This is a big blocker for API-first approach in current spring boot development.

mikhail-putilov:
Sorry for being cloudy. That's just devastating.

wing328:
@mikhail-putilov please use an older stable version (e.g. 4.0.0) for the time being.

PR #2741, opened 2019-04-25
[Elixir] update to 1.6 version

karismann:
### PR checklist

@mrmstn

### Description of the PR

- update to elixir 1.6 / tesla 1.2
- add integration-test on pet Api
- add missing 200 responses in petstore-with-fake-endpoints-models-for-testing.yaml
- update sample and add in ensureuptodate shell


wing328:
cc @mrmstn (2018/12)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElixirClientCodegen.java
```diff
@@ -50,10 +50,10 @@
     // This is the name of elixir project name;
     protected static final String defaultPackageName = "openapi_client";
 
-    String supportedElixirVersion = "1.4";
+    String supportedElixirVersion = "1.8";
```
mrmstn: tesla requires `1.4`, poison requires `1.6` - I would suggest we just stick to 1.6 instead of 1.8 to allow higher backwards compatibility since 1.8 isn't really a requirement

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElixirClientCodegen.java
```diff
@@ -50,10 +50,10 @@
...
     List<String> extraApplications = Arrays.asList(":logger");
     List<String> deps = Arrays.asList(
-            "{:tesla, \"~> 1.0.0\"}",
+            "{:tesla, \"~> 1.2.1\"}",
             "{:poison, \"~> 3.0.0\"}"
```
mrmstn:
```suggestion
             "{:poison, \"~> 3.0\"}"
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElixirClientCodegen.java
```diff
@@ -50,10 +50,10 @@
     // This is the name of elixir project name;
     protected static final String defaultPackageName = "openapi_client";
 
-    String supportedElixirVersion = "1.4";
+    String supportedElixirVersion = "1.8";
     List<String> extraApplications = Arrays.asList(":logger");
     List<String> deps = Arrays.asList(
-            "{:tesla, \"~> 1.0.0\"}",
+            "{:tesla, \"~> 1.2.1\"}",
```
mrmstn:
```suggestion
            "{:tesla, \"~> 1.2\"}",
```

karismann:
> Thank you very much for your PR, I marked some things that I would like to change to allow the generated client to be "open" or in other words break less dependencies

thanks for your review . i made the changes you suggested

wing328:
What about making `supportedElixirVersion` configurable via an option (and default to 1.6) ?

mrmstn:
@wing328 I can't quite see the use for this type of enforcement, but definitely worth thinking about. This also won't do any harm if we could provide such a config parameter

wing328:
@mrmstn ok, we will consider adding such option if there are more demands for it

PR #2781, opened 2019-05-01
[C++][Pistache]Fix import of model header

### LINKED ISSUE #2024: [BUG] [C++][Pistache-server] Object defined inside the body content does not compile
##### Description
When the YAML file describes the use of an object inside the body content (without any content type like application/json), the generator produces a model file called InlineObject.h (and cpp) but to use it, the generator set the wrong include file name :

```
#include "Inline_object.h"
```
instead of

```
#include "InlineObject.h"
```

It seems that there is a confusion between the file name and the class name

```
class  Inline_object
    : public ModelBase
{
    ...
} ;
```

Current master 4.0.0-SNAPSHOT

```yaml
openapi: 3.0.0
info:
  version: 1.0.0
  title: Check generation of pistache
  description: Internal ref filename is check_object_immediately_in_the_content.yaml

servers:
  - url: http://localhost:8080

paths:
  /sm-contexts:
    post:
      summary:  Create SM Context
      tags:
        - SM contexts collection
      operationId: PostSmContexts
      requestBody:
        description: representation of the SM context to be created in the SMF
        required: true
        content:
          multipart/related:
            schema:
              type: object
              properties: # Request parts
                dataMessage:
                  type: string

      responses:
        '200':
          description: Everythings gonna be alright
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Content"
        default:
          description: unexpected error


components:
  schemas:
    Content:
      type: string

```

Generate :
```
openapi-generator-cli.sh generate -i ./openapi.yaml -g cpp-pistache-server -c ./config.json -o .
```

Compile :
```
g++ -c  -I./api -I./model -I./impl  -Wall -g -std=c++11 -o obj/api/SMContextsCollectionApi.o api/SMContextsCollectionApi.cpp
```

##### Steps to reproduce

Generate & compile

##### Related issues/PRs

N/A

etherealjoy:
### PR checklist

- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.

### Description of the PR
Fixes #2024
- Fixes import of model header due to rework (similar to #2168, #2750, #2744) 
- sanitizing the name is done in `AbstractCppCodegen` already so not needed twice

@stkrwork @ravinikam @MartinDelille @fvarose

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppPistacheServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppPistacheServerCodegen.java
@@ -285,7 +285,7 @@ public class CppPistacheServerCodegen extends AbstractCppCodegen {
 
     @Override
     public String toApiFilename(String name) {
-        return  modelNamePrefix + toModelName(name) + "Api";
+        return toApiName(name);
     }
 
     /**
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppPistacheServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppPistacheServerCodegen.java
@@ -264,7 +264,7 @@ public class CppPistacheServerCodegen extends AbstractCppCodegen {
 
     @Override
     public String toModelFilename(String name) {
-        return camelize(toModelName(name));
+        return toModelName(name);
     }
 
     @Override
@@ -285,7 +285,7 @@ public class CppPistacheServerCodegen extends AbstractCppCodegen {
 
     @Override
     public String toApiFilename(String name) {
-        return  modelNamePrefix + camelize(name) + "Api";
+        return  modelNamePrefix + toModelName(name) + "Api";
     }
 
     /**


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppPistacheServerCodegen.java
```diff
@@ -285,7 +285,7 @@ public String apiFilename(String templateName, String tag) {
 
     @Override
     public String toApiFilename(String name) {
-        return  modelNamePrefix + camelize(name) + "Api";
+        return  modelNamePrefix + toModelName(name) + "Api";
```
stkrwork: We use toModelName in toApiFilename? Wouldn't a toApiName make more sense and make the code more readable?

etherealjoy: You are right, it is the correct way. the super class `AbstractCppCodegen` does it correctly already
```C++
    public String toApiName(String type)
```

Updated

etherealjoy:
@stkrwork 
I need to fix the same thing for `cpprest` should I do it here?

stkrwork:
if you are already looking into it, can you also check it in restbed generator?

etherealjoy:
OK I will check it also

stkrwork:
thanks

Issue #2702, opened 2019-04-19
[BUG] '/' prepended to paths that don't start with '/'


dkliban:
##### Description

I am currently trying to switch from swagger-codegen-cli to openapi-generator-cli, but I ran into a problem with the generated Python client code. I do not experience the same behavior when generating client code using swagger-codegen-cli. 

`/` is prepended to paths that don't start with a `/`. This is a problem because my OpenAPI schema includes paths that consist of a single path parameter. e.g.: `{artifact_href}`. In the generated code this is turned into `/{artifact_href}`. Since the path parameter being passed in looks like `/pulp/api/v3/artifacts/1234/`, the client makes a request to `http://localhost//pulp/api/v3/artifacts/1234/` and the server is unable to service the request. Notice the `//` in the URL.

This problem is not unique to the Python generator. I am seeing the exact same behavior for Ruby also. 
 
##### openapi-generator version

The latest docker image of openapi-generator-cli.

##### OpenAPI declaration file content or url

https://docs.pulpproject.org/en/3.0/nightly/api.json

##### Command line used for generation

    docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate \
        -i /local/api.json \
        -g python \
        -o /local/pulpcore-client \
        -DpackageName=pulpcore.client.pulpcore \
        -DprojectName=pulpcore-client \
        -DpackageVersion=3.0.0rc1 \
        --skip-validate-spec

##### Steps to reproduce

Generate a Python client with the schema from above. Observe that the client that is generated has 

```
        return self.api_client.call_api(
            '/{artifact_href}', 'GET',
```
while the path for this operation in the OpenAPI schema was simply `{artifact_href}`. I would expect the generator to not modify the path value.

jmini:
This was introduced with https://github.com/OpenAPITools/openapi-generator/pull/1034.

Not having a leading `/` is not conform to the OpenAPI Spec:

> A relative path to an individual endpoint. The field name MUST begin with a slash.
source https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#pathsObject
This is why we decided to add this when missing.

But you are bringing an interesting case...
And for OpenAPI-Generator an interesting question: what should we do with invalid spec? Try to fix it or not...

A question from my side:
Why is your parameter `"/pulp/api/v3/artifacts/1234/"` and not `"pulp/api/v3/artifacts/1234/"`?

jimschubert:
[...]

PR #1034, opened 2018-09-14
[all] Add leading slash in path if missing

jmini:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.3.x`, `4.0.x`. Default: `master`.
- [x] Copied the technical committee: @OpenAPITools/generator-core-team  @OpenAPITools/openapi-generator-collaborators (change in core)

### Description of the PR

From https://github.com/OpenAPITools/openapi-generator/pull/967#issuecomment-418602467 and #997.

This PR adds the leading `/` if it is missing in the path definition.

---

More details:

This spec is not valid:

```yaml
openapi: '3.0.1'
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/api'
paths:
  ping/me:
    get:
      operationId: pingGet
      responses:
        '201':
          description: OK
```

Because the path `ping/me` do not conform to this rule:

> The field name MUST begin with a slash.

See the [OpenAPI Specification 3.0.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#patterned-fields).

Some editors like [Swagger online editor](https://editor.swagger.io/) or the [KaiZen-Editor](https://github.com/RepreZen/KaiZen-OpenApi-Editor/) are reporting errors, but Swagger-Parser does not.

Some templates uses `{{{basePathWithoutHost}}}{{{path}}}`. 

* Before this PR it will generate `http://localhost:8000/apiping/me` (incorrect) 
* After this PR it will generate `http://localhost:8000/api/ping/me` (correct)

PR #2783, opened 2019-05-01
[all] Adds strict spec option

jimschubert:
Introduces an option to allow user customization of strict specification
behaviors. For instance, OpenAPI 3.x requires a path object name to be
prefixed with '/' so we append any missing '/', but this may not be
desirable to some users or generators. In this commit, this fix specifically is
the only modification affected.

### PR checklist

### Description of the PR

Adds an option which allows users to disable any strict specification behaviors applied during generation. For example, #1034 introduces a modification which prefixes all paths missing a beginning slash to better fit "MUST" definitions in the OpenAPI Specification 3.x. Ideally, this behavior would be rolled up into the parser (i.e. if we've validated via the parser, we shouldn't need to fix the spec document).

Such a switch is necessary because, as discussed in #2702, users may not want the input document to be modified to match strict behaviors against the specification.

Use of this option to disable strict behaviors, however, does mean that one or more generators may not work as expected. For example, if the url part of a document is: `http://localhost/api` and the path is `resource/{id}`, this may result in a generated URL such as `http://localhost/apiresource/{id}`.

Note that there seems to be a discrepancy or error in the OpenAPI 3.0.x specification. A server url object default is a server with relative url `/`. The url requirement is defined as:

> REQUIRED. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the OpenAPI document is being served. Variable substitutions will be made when a variable is named in {brackets}.

But, URLs with and without trailing `/` are both valid.

The field name for a paths object is defined as:

> A relative path to an individual endpoint. The field name MUST begin with a forward slash (/). The path is appended (no relative URL resolution) to the expanded URL from the Server Object's url field in order to construct the full URL. Path templating is allowed. When matching URLs, concrete (non-templated) paths would be matched before their templated counterparts. Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical. In case of ambiguous matching, it's up to the tooling to decide which one to use.

So, strictly following the spec here, it seems that when `server.url = /` and `paths["/api"]` exist, the desired outcome (no relative URL resolution, appending path to URL) is `//api`, which obviously isn't correct as this means the protocol-relative hostname `api`.

This PR does not attempt to resolve any concerns with the above scenario; it is only provided as an example of why we follow strict behaviors by default and need to sometimes modify what a user provides (even when it successfully validates against the spec).

/cc @OpenAPITools/generator-core-team 

## TODO

Once merged, the option will need to be applied to the gradle plugin. It seems to always resolve against maven with greater precedence over mavenLocal, and I'm hesitant to shift repository settings around and introduce any issues with the build pipeline.

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -216,7 +216,7 @@ public class Generate implements Runnable {
 
     @Option(name = {"--strict-spec"},
             title = "true/false strict behavior",
-            description = "'MUST' and 'SHALL' wording in OpenAPI spec is strictly adhered to, e.g. no fixes will be applied to documents which pass validation but don't follow the spec.",
+            description = "'MUST' and 'SHALL' wording in OpenAPI spec is strictly adhered to. e.g. when false, no fixes will be applied to documents which pass validation but don't follow the spec.",
             arity = 1)
     private Boolean strictSpecBehavior;
 
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -214,6 +214,12 @@ public class Generate implements Runnable {
             description = "Skips the default behavior of validating an input specification.")
     private Boolean skipValidateSpec;
 
+    @Option(name = {"--strict-spec"},
+            title = "true/false strict behavior",
+            description = "'MUST' and 'SHALL' wording in OpenAPI spec is strictly adhered to, e.g. no fixes will be applied to documents which pass validation but don't follow the spec.",
+            arity = 1)
+    private Boolean strictSpecBehavior;
+
     @Option(name = {"--log-to-stderr"},
             title = "Log to STDERR",
             description = "write all log messages (not just errors) to STDOUT."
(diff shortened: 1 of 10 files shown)


modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
```diff
@@ -214,6 +214,12 @@
             description = "Skips the default behavior of validating an input specification.")
     private Boolean skipValidateSpec;
 
+    @Option(name = {"--strict-spec"},
+            title = "true/false strict behavior",
+            description = "'MUST' and 'SHALL' wording in OpenAPI spec is strictly adhered to, e.g. no fixes will be applied to documents which pass validation but don't follow the spec.",
```
dkliban: s/, e.g. no/. When false, no/

modules/openapi-generator-maven-plugin/README.md
```diff
@@ -57,6 +57,7 @@ mvn clean compile
 - `logToStderr` - write all log messages (not just errors) to STDOUT
 - `enablePostProcessFile` - enable file post-processing hook
 - `skipValidateSpec` - Whether or not to skip validating the input spec prior to generation. By default, invalid specifications will result in an error.
+- `strictSpec` - Whether or not to treat an input document strictly against the spec. 'MUST' and 'SHALL' wording in OpenAPI spec is strictly adhered to, e.g. no fixes will be applied to documents which pass validation but don't follow the spec.
```
dkliban: s/, e.g. no/. When false, no/

jimschubert:
Thanks for the review, it made me realize I had missed documenting the CLI option in a couple places.

jimschubert:
Forgot to assign from `CodegenConfigurator` to `CodegenConfig`. Opened #2814 with a fix for this.

PR #2799, opened 2019-05-02
Update DefaultCodeGen to allow additional primitive types

steco:
If a string field is specified with a format which is also defined using --typeMappings, it will be treated as a primitive type

For example, if you run the following command:

`java -jar openapi-generator-cli.jar generate [options] --type-mappings dateorcutlabel=CutLabel`

The generated code will assume there is a class `CutLabel` and use it for any parameter which has a specific similiar to this:

> {
>   "name": "effectiveAt",
>   "in": "query",
>   "description": "Optional. The effective date of the data",
>   "required": false,
>   "type": "string",
>   "format": "**dateorcutlabel**"
> }

### PR checklist

### Description of the PR

### DIFF
--- a/bin/windows/android-petstore-httpclient.bat
+++ b/bin/windows/android-petstore-httpclient.bat
@@ -5,6 +5,6 @@ If Not Exist %executable% (
 )
 
 REM set JAVA_OPTS=%JAVA_OPTS% -Xmx1024M -DloggerPath=conf/log4j.properties
-set ags=generate -i modules\openapi-generator\src\test\resources\2_0\petstore.yaml -g android -o samples\client\petstore\android\httpclient-Dlibrary=httpclient
+set ags=generate -i modules\openapi-generator\src\test\resources\2_0\petstore.yaml -g android -o samples\client\petstore\android\httpclient -Dlibrary=httpclient
 
 java %JAVA_OPTS% -jar %executable% %ags%
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1411,7 +1411,7 @@ public class DefaultCodegen implements CodegenConfig {
      * @param schema
      * @return type
      */
-    private static String getPrimitiveType(Schema schema) {
+    private String getPrimitiveType(Schema schema) {
         if (schema == null) {
             throw new RuntimeException("schema cannot be null in getPrimitiveType");
         } else if (ModelUtils.isStringSchema(schema) && "number".equals(schema.getFormat())) {
@@ -1452,6 +1452,13 @@ public class DefaultCodegen implements CodegenConfig {
         } else if (ModelUtils.isUUIDSchema(schema)) {
             return "UUID";
         } else if (ModelUtils.isStringSchema(schema)) {
+            if(typeMapping.containsKey(schema.getFormat())) {
+                // If the format matches a typeMapping (supplied with the --typeMappings flag)
+                // then treat the format as a primitive type.
+                // This allows the typeMapping flag to add a new custom type which can then
+                // be used in the format field.   
+                return schema.getFormat();
+            }
             return "string";
         } else if (ModelUtils.isFreeFormObject(schema)) {
             return "object";
@@ -3548,7 +3555,7 @@ public class DefaultCodegen implements CodegenConfig {
      * @param schemas The complete set of model definitions (schemas).
      * @return A mapping from model name to type alias
      */
-    static Map<String, String> getAllAliases(Map<String, Schema> schemas) {
+    Map<String, String> getAllAliases(Map<String, Schema> schemas) {
         if (schemas == null || schemas.isEmpty()) {
             return new HashMap<>();
         }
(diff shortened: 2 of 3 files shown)


wing328:

steco:
Fixed - thanks William

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1452,6 +1452,10 @@ private static String getPrimitiveType(Schema schema) {
         } else if (ModelUtils.isUUIDSchema(schema)) {
             return "UUID";
         } else if (ModelUtils.isStringSchema(schema)) {
+            if(typeMapping.containsKey(schema.getFormat()))
+            {
```
wing328: Can this open curly bracket go to line 1455 instead to conform to Java code style?

bin/windows/android-petstore-httpclient.bat
```diff
@@ -1,10 +1,7 @@
 set executable=.\modules\openapi-generator-cli\target\openapi-generator-cli.jar
 
-If Not Exist %executable% (
-  mvn clean package
-)
```
wing328: Why remove these? Are these lines not working as expected (to build the JAR if not present)?

wing328:
cc @openapitools-bot as the change covers default codegen

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
steco: Fixed and updated

bin/windows/android-petstore-httpclient.bat
steco: These lines don't work on my Windows machine (I need to run mvnw and it is not on my path anyway) so I removed them, reinstated for people who have a more complete setup.

wing328: Very good point. May be we should use `mvnw` instead of `mvn` to make it easier for devleopers using Windows machines to build the project.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1452,6 +1452,13 @@ private static String getPrimitiveType(Schema schema) {
         } else if (ModelUtils.isUUIDSchema(schema)) {
             return "UUID";
         } else if (ModelUtils.isStringSchema(schema)) {
+            if(typeMapping.containsKey(schema.getFormat())) {
+                // If the format matches a typeMapping (supplied with the --typeMappings flag)
+                // then treat the format as a primitive type.
+                // This allows the typeMapping flag to add a new custom type which can then
+                // be used in the format field.   
+                return schema.getFormat();
```
wing328: @steco Previously there's a special case handled in the code above: https://github.com/OpenAPITools/openapi-generator/pull/2799/files#diff-7cb46fa53f89a458a7b7cb201d2214a8R1417 (type: string, format: number).

If we just return format, would it be an issue if `schema.getFormat()` happens to be one of the existing OAS type (e.g. number)?

Should we return something like `string_number` instead to make it distinguishable with the existing types? Would that work in your case?

steco: I don't see the special case for type: string, format: number.  I see the following specific formats for strings - "date", "date-time", "password", "byte", "binary", "uuid", email" (in SchemaTypeUtil).

If someone has a swagger.json file which contains fields which are also primitive types, ie also in typeMapping (eg with type:string, format:number) then this change will mean that the field would change from being a string to being a double in the generated code.  However, I think this would be unlikely.

Using string_number would work for me, but I think this adds unnecessary complexity and makes a custom field different to the existing special cases (listed above).

jimschubert: format could be anything. I don't think we should try to cover edge cases, as I think that would over-complicate things.

sigpwned:
I have a similar requirement and independently generated what is (essentially) the same PR as above. It worked for my Java application as well. I would love to see this get merged and released soon! My team will likely have to maintain a patched version until then.

Thank you @steco for issuing the PR, and to the broader OpenAPI team for making such a great toolchain around OpenAPI!

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
sigpwned: I realize I'm not a contributor or member here, so I may not fully grasp everything in play here. However, I am a user who also needs this feature, and I implemented a very similar patch independently, so here's my opinion, for whatever it's worth.

@wing328's concern above appears to be that the edited code no longer handles a special case: string / number. However, that case still appears to be handled on (what is currently) line 1417:

    // Line 1417
    } else if (ModelUtils.isStringSchema(schema) && "number".equals(schema.getFormat())) {
        // special handle of type: string, format: number
        return "BigDecimal";
    } else if (ModelUtils.isByteArraySchema(schema)) {
    
Because that code is still there, it looks (to me, at least) like we can be confident it still handles the string/number case properly. Even better, that code didn't change and all edits to the function come after that test, which increases our confidence further.

I ran just this code change through the build and all tests pass, which is further evidence that this particular change does The Right Thing. (Note: In retrospect, I only ran surefire tests.)

With this change in, I have also confirmed that I can use type mappings to generate references to custom types instead of String for specific string/format combinations only.

wing328:
@steco as discussed over Gitter, the edge case rarely happens and if it does, we will instead use "string_{format}" as a bug fix (non-breaking change) to handle the edge case correctly.

PR #2807, opened 2019-05-03
[Elm] Bugfix .encode in modelTypeDiscriminator

andys8:
The function is called `encode`. This fixed the call.

See https://github.com/OpenAPITools/openapi-generator/blob/dfa7e616c0f311f3ebf58c04287f200c8e491f08/modules/openapi-generator/src/main/resources/elm/modelTypeArray.mustache#L11

I have encountered this issue, and this is probably an easy fix.
@eriktim Can you have a short look?


eriktim:
Thanks, @andys8!

PR #2813, opened 2019-05-03
[Golang][client] Make it possible to generate go.mod and go.sum files

### LINKED ISSUE #2796: [REQ][Golang][client] Add go.mod to go client templates

I need to use the autogenerated go client as a dependency of another go module, which is much easier when it's a module as well.

I'd like the go client templates to also contain go.mod/go.sum files that would be rendered (not sure if this should be by default, but it probably wouldn't hurt). To do that, the go config file would also need to contain the full module name, e.g. "github.com/user/mymodule".

I could maintain these files by hand outside of openapi-generator functionality.

## Additional context

This seems to be a simple enough modification that I would probably be able to send a PR myself, but first I'd like to hear if it's wanted/acceptable.

bkabrda:
### PR checklist

@antihax @bvwells @grokify @kemokemo

### Description of the PR

Fixes #2796.
I didn't regenerate samples, since this is only triggered when moduleName is specified (either in config or on commandline). If you wish, I can definitely add this option to the scripts that generate samples and amend this PR.

I do have one question - it seems that the files under docs/generators are autogenerated, but I failed to find a script that generates them. Am I just looking wrong or are these handwritten? Thanks!

### DIFF
--- a/modules/openapi-generator/src/main/resources/go/go.mod.mustache
+++ b/modules/openapi-generator/src/main/resources/go/go.mod.mustache
@@ -1,0 +1,6 @@
+module github.com/{{gitUserId}}/{{gitRepoId}}
+
+require (
+    github.com/antihax/optional v0.0.0-20180406194304-ca021399b1a6
+    golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a
+)
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientCodegen.java
@@ -60,7 +60,6 @@ public class GoClientCodegen extends AbstractGoCodegen {
                 CodegenConstants.PREPEND_FORM_OR_BODY_PARAMETERS,
                 CodegenConstants.PREPEND_FORM_OR_BODY_PARAMETERS_DESC)
                 .defaultValue(Boolean.FALSE.toString()));
-
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


etherealjoy:
I think we need this also for go server.

bkabrda:
So after a Gitter discussion with @wing328 I made this use GIT_USER_ID and GIT_REPO_ID and generate go.mod and go.sum by default.

modules/openapi-generator/src/main/resources/go/go.mod.mustache
```diff
@@ -0,0 +1,6 @@
+module github.com/{{gitUserId}}/{{gitRepoId}}
```
bvwells: This all seems to be very specific to modules in GitHub repos. Personally I would think about making the complete module path as a input variable to the generator...

bvwells:
Hi @bkabrda, missed the thread on gitter, but this change makes the module definition very specific to GitHub repos. Might want to rethink the use of GIT_USER_ID and GIT_REPO_ID as the input to the generator. Perhaps specify the module path as a single path to the generator...

modules/openapi-generator/src/main/resources/go/go.mod.mustache
wing328: Yup, we talked about this as well. No one has asked for supporting other git hosting services so far (which is a bit surprising to me). We can introduce another option "gitHost" if there's a need to support Gitlab, BitBucket or other Git hosting services.

bvwells: Okay, fair enough. For me, it feels a bit clunky to have to specify these two (and perhaps in future three)  pieces of information separately on the command line.

wing328: Yup, agreed some users prefer just one single option for this particular setting. gitUserID and gitRepoID are global options used by other generators as well (e.g. PHP) so another benefit is consistency among generators. Let's go with the current approach and see if more Go users have an opinion on this.

PR #2804, opened 2019-05-03
Add warning about .net 2.0 framework reached EOL

wing328:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Add warning about .net 2.0 framework reached EOL

PR #2786, opened 2019-05-01
[csharp-dotnet2] REMOVED

jimschubert:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Removes C# 2.0 support. See #116

Per [Microsoft Product Lifecycle](https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=.NET%20Framework&Filter=FilterNO), support for .NET Framework 2.0 ended in 2011:

> Support for .NET Framework 2.0 ended on July 12, 2011. .NET 3.5 SP1 is the only supported service pack level after this date. We strongly encourage customers to migrate to .NET Framework 3.5 SP1. For more information, please visit the .NET Framework Support Lifecycle Policy FAQ https://support.microsoft.com/help/17455

cc @OpenAPITools/openapi-generator-core-team @mandrean (2017/08), @jimschubert (2017/09)

Issue #845, opened 2018-08-18
Create a "core" artifact


jimschubert:
##### Description

There's currently no clearly defined "interface" between templates, extension points, or code generation.

If we extract model definitions and interfaces to a "core" artifact, this would reduce noise for those who want to extend and/or contribute to the project.

This would ideally be done after other work in the [Separation of Concerns](https://github.com/OpenAPITools/openapi-generator/projects/5) project in this repo.

##### Related PRs

* #503
* #837
* #840 
* #841 
* #842 (maybe) 
* #843 
* #844 

##### Suggest a fix/enhancement

From #503:

![image](https://user-images.githubusercontent.com/109659/44302995-2ec9b300-a303-11e8-87a6-c9f7238ab009.png)

In the above diagram, we'd include any models (codegen, options, template) or interfaces which may be consumed by users for creating custom generators or plugins. **NOTE** Template Definition above refers to a strongly-typed model that would replace the current `Map<String, Object>` (not the resources defining templates); work for this is already started in #837.

jmini:
I am not sure how I can add an idea to the list, but I think we should remove using `System.setProperty(..)` from our code.

It is ok to read them (requested for backward compatibility), but setting the values is a bad practice.
At the end they are some inputs options for the `DefaultGenerator`

Related issue: I have described how System properties are set by `CodegenConfigurator.setSystemProperties()` here: https://github.com/OpenAPITools/openapi-generator/issues/811#issuecomment-415313725

Should I open a new issue for this task?

jimschubert:
@jmini yeah I think that should be another card. Work from this card will be a breaking change, but removing setProperty should be done as soon as possible, imho.

I would append that setProperty is used in the Gradle plugin, I think it's fine in this case. It shouldn't exist in the generator packages, though, because these can be consumed by other artifacts or tools.

jmini:
Removal of `System.setProperty(..)` is now tracked in https://github.com/OpenAPITools/openapi-generator/issues/1597

jimschubert:
Template Engine component is done via #6357

jimschubert:
"Global Options" and "Generation Options" components are completed via #2946. I'd updated these terms to "Workflow Settings" (rather than Global Options) and "Generator Settings" rather than "Generation Options" to better differentiate between the touch points.

Issue #840, opened 2018-08-18
Create reusable feature metadata for all generators


jimschubert:
##### Description

As discussed in #503, this would be a starting point for creating a compatibility matrix. We need some way for generators to define all feature sets supported by the generator.

We need to take into consideration that a base (abstract) type used across multiple generators may implement more features than a subtype intends to support. This means we need to be able to mutate the feature metadata from derived types.

##### Related issues/PRs

* #503

##### Suggest a fix/enhancement

I have a prototype in which I created a bitmask type in Kotlin, and apply feature sets as properties.

The feature sets I created are:

<details>
<summary>Click to expand Bitmask/Feature snippet</summary>

```kotlin
class BitMask(val value: Long)
interface Flags  {
    val bit: Long

    fun toBitMask(): BitMask = BitMask(bit)
}

// Flags extensions
infix fun Flags.and(other: Long): BitMask = BitMask(bit and other)
infix fun <T: Flags> Flags.or(other: T): BitMask = BitMask(bit or other.bit)
infix operator fun Flags.plus(other: Flags): BitMask = BitMask(bit or other.bit)
inline fun <reified T> enabledValues(mask: BitMask) : List<T> where T : Enum<T>, T : Flags {
    return enumValues<T>().filter {
        mask hasFlag it
    }
}

infix fun BitMask.or(other: Flags): BitMask = BitMask(value or other.bit)
infix operator fun BitMask.plus(other: BitMask): BitMask = BitMask(value or other.value)
infix operator fun BitMask.plus(other: Flags): BitMask = BitMask(value or other.bit)
infix fun <T: Flags> BitMask.hasFlag(which: T): Boolean {
    // an Undefined flag is a special case.
    if(value == 0L || (value > 0L && which.bit == 0L)) return false

    return value and which.bit == which.bit
}
infix fun <T: Flags> BitMask.unset(which: T): BitMask = BitMask(value xor which.bit)

// Features
enum class ClientModificationFeature(override val bit: Long) : Flags {
    Undefined(0),
    BasePath(1 shl 0),
    Authorizations(1 shl 1),
    UserAgent(1 shl 2);
}
enum class DataTypeFeature(override val bit: Long) : Flags {
    Undefined(0),
    Int32(1 shl 0),
    Int64(1 shl 1),
    Float(1 shl 2),
    Double(1 shl 3),
    Decimal(1 shl 4),
    String(1 shl 5),
    Byte(1 shl 6),
[...]

PR #2816, opened 2019-05-04
[feat] Intro GeneratorMetadata (stability index)

jimschubert:
### PR checklist

### Description of the PR
GeneratorMetadata offers an immutable object created via Builder pattern
which allows generators to explicitly define their stability (stable,
beta, experimental, deprecated) as well as a message to be shown during
generation.

This is a step toward:

* Fleshing out the "Core" artifact (#845)
* Providing a place to encapsulate feature-oriented metadata (#840)
* Providing a means to communicate end of life scheduling (#116)

This new structure, specifically the Stability property, allows us to
offer future enhancements such as allowing users to filter down to only
"Stable" generators via CLI, and eventually any compat table (see #503).

See [project 5](https://github.com/OpenAPITools/openapi-generator/projects/5) for other potentially related work.

---

This is an addition to what was introduced for the C# 2.0 generator in #2804. I had opened #2786, not realizing that the generator hadn't been deprecated as discussed in #116. After consideration, I decided to have something typed which would explicitly define the generator as deprecated in code.

Example:

```
$ ./bin/csharp-dotnet2-petstore.sh
# START SCRIPT: ./bin/csharp-dotnet2-petstore.sh
[main] INFO  o.o.codegen.DefaultGenerator - OpenAPI Generator: csharp-dotnet2 (client)
[main] WARN  o.o.codegen.DefaultGenerator - Generator 'csharp-dotnet2' is considered deprecated.
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one
[main] WARN  o.o.c.l.CSharpDotNet2ClientCodegen - Per Microsoft Product Lifecycle (https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=.NET%20Framework&Filter=FilterNO), support for .NET Framework 2.0 ended in 2011 so there may be security issues using the auto-generated C# 2.0 source code.
```


jimschubert:
Question:

This creates some redundancy in the "list" CLI command, where some generators previously encoded the stability into the name of the generator:

```
    - scala-httpclient-deprecated (deprecated)
    - swift2-deprecated (deprecated)
    - swift3-deprecated (deprecated)
```

I think this is fine for now, since machine parsing should be done via the `list --short` option. It seems to me that we might want an `UNMAINTAINED` option to differentiate between something which is deprecated (i.e. "this will be removed at some point") versus something which is either low priority or being kept with no intention to remove.  If that's not the case, we may want to consider explicitly defining end of life for the above mentioned deprecated generators.

jimschubert:
for visibility @OpenAPITools/openapi-generator-collaborators

modules/openapi-generator-core/src/main/java/org/openapitools/codegen/meta/Stability.java
```diff
@@ -0,0 +1,37 @@
+package org.openapitools.codegen.meta;
+
+/**
+ * Represents the "stability index" of a generator or feature, based on the stability indexes defined in the node.js ecosystem.
+ */
+public enum Stability {
```
jimschubert: NOTE: We could apply `Stability` to other things as well, such as the templating engine adapters.

PR #2818, opened 2019-05-04
Better handling of form data

wing328:
### PR checklist

### Description of the PR

- better form data detection
- better handling of content-type (show a warning message if content-type cannot be detected)
- minor code format

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1450,7 +1450,8 @@ public class DefaultCodegen implements CodegenConfig {
             } else if (ModelUtils.isDoubleSchema(schema)) {
                 return SchemaTypeUtil.DOUBLE_FORMAT;
             } else {
-                LOGGER.warn("Unknown `format` detected for " + schema.getName() + ": " + schema.getFormat());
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4194,7 +4196,8 @@ public void writePropertyBack(String propertyKey, boolean value) {
 
     protected String getContentType(RequestBody requestBody) {
         if (requestBody == null || requestBody.getContent() == null || requestBody.getContent().isEmpty()) {
-            return null;
+            LOGGER.warn("Cannot determine the content type. Default to UNKNOWN_CONTENT_TYPE.");
+            return "UNKNOWN_CONTENT_TYPE";
```
jimschubert: Can we fallback to a valid mime type such as `application/octet-stream`?

wing328: Yup, we can definitely do that but the reason why I did this is to make it consistent with similar unknown cases in the code (e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L3131) so that users can easily grep "UNKNOWN" to identify which operations/models need to be fixed.

If later users have feedback on having a better default (e.g. application/octet-stream), we can definitely make the switch.

Issue #2350, opened 2019-03-10
[REQ] New ASP.NET Core 3.0 server generator


wing328:
To start with, we'll need to create the following:

- new AspNetCore3ServerCodegen.java based on AspNetCoreServerCodegen.java
- new resource folder `modules/openapi-generator/src/main/resources/aspnetcore3` based on `modules/openapi-generator/src/main/resources/aspnetcore` (2.0 or 2.1)
- add batch file to generate Petstore samples under `samples/server/petstore/aspnetcore3`

cc @SeanFarrow

SeanFarrow:
Let me get a vm with .net core 3 on sorted in the next few days.
I’ll then take a look at this, so if you could assign it to me that would be appreciated.

A-Joshi:
@SeanFarrow 
Have been slowly adding in library and other support for the aspnetcore generator particularly for 2.2 and happy to collaborate for 3.0.

etherealjoy:
I am interested in ASP.NET Core 3.0 and I have lots of manual things that I do with the 2.1 generator and I would really like a clean sheet .NET Core 3 including clients and using Microsoft's new high performance json parser

A-Joshi:
Sure - no objections about that - just curious what do you have to do
manually? Moving to 3.0 would be a good idea - though I suspect it will
take a little time after it is releaseed before one can move to it.

wing328:
Discussions related to authentication support: https://github.com/OpenAPITools/openapi-generator/issues/2352

A-Joshi:
Based on what I seen so far [Migrate from ASP.NET Core 2.2 to 3.0 Preview 2](https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio)  it might not be that difficult to simply extend the current current generator for one more version.

etherealjoy:
I did these
- update TFM
- Change The Configure services
- Add Json.NET manually
- Adapt Configure services and Configure due to change in hosting environment and Json.NET
- Update package versions

Additionally I was wondering if we use Json.NET or use MS internal Json package
Nice to have would be the DI mechanism to inject the API handlers instead of manually editing the HomeController. This is tedious everytime we generate code due to API update

A-Joshi:
@etherealjoy it might be simpler to adopt the approach of using the generator to generate an abstract class and have the server implement it. I am waiting for #2629 to be merged and will be able to provide more concrete examples.

[...]

PR #2824, opened 2019-05-05
[aspnetcore]: Preliminary support for ASP.NET 3.0 Core preview 5

A-Joshi:
### PR checklist

### Description of the PR

Preliminary support for ASP.NET Core 3.0 (preview 4). See issue #2350

No support for the new style routing or new Json library (still uses Newtonsoft).

### DIFF
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -27,8 +27,8 @@ sidebar_label: aspnetcore
 |isLibrary|Is the build a library| |false|
 |useFrameworkReference|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
 |useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
-|newtonsoftVersion|Version for Microsoft.AspNetCore.Mvc.NewtonsoftJson for ASP.NET Core 3.0+| |3.0.0-preview4-19216-03|
-|useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
+|newtonsoftVersion|Version for Microsoft.AspNetCore.Mvc.NewtonsoftJson for ASP.NET Core 3.0+| |3.0.0-preview5-19227-01|
+|useDefaultRouting|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
 |buildTarget|Target to build an application or library| |program|
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -27,7 +27,7 @@ sidebar_label: aspnetcore
 |isLibrary|Is the build a library| |false|
 |isFramework|Are we using packagereference or framework| |false|
 |useNewtonsoft|Uses the Newtonsoft JSN library.| |true|
-|useEndpointRoutng|Uses the newend point routing JSN library.| |false|
+|useDefaultRoutng|Uses the newend point routing JSN library.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
 |buildTarget|Target to build an application or library| |program|
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -26,7 +26,7 @@ sidebar_label: aspnetcore
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
 |isFramework|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
-|useNewtonsoft|Uses the Newtonsoft JSN library.| |true|
+|useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
 |useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -27,6 +27,7 @@ sidebar_label: aspnetcore
 |isLibrary|Is the build a library| |false|
 |useFrameworkReference|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
 |useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
+|newtonsoftVersion|Version for Microsoft.AspNetCore.Mvc.NewtonsoftJson for ASP.NET Core 3.0+| |3.0.0-preview4-19216-03|
 |useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -18,12 +18,16 @@ sidebar_label: aspnetcore
 |sourceFolder|source folder for generated code| |src|
 |compatibilityVersion|ASP.Net Core CompatibilityVersion| |Version_2_1|
 |aspnetCoreVersion|ASP.NET Core version: 2.2 (default), 2.1, 2.0 (deprecated)| |2.2|
+|swashbuckleVersion|Swashbucke version: 3.0.0 (default), 4.0.0| |3.0.0|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
+|isFramework|Are we using packagereference or framework| |false|
+|useNewtonsoft|Uses the Newtonsoft JSN library.| |true|
+|useEndpointRoutng|Uses the newend point routing JSN library.| |false|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
 |buildTarget|Target to build an application or library| |program|
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -17,15 +17,15 @@ sidebar_label: aspnetcore
 |packageGuid|The GUID that will be associated with the C# project| |null|
 |sourceFolder|source folder for generated code| |src|
 |compatibilityVersion|ASP.Net Core CompatibilityVersion| |Version_2_2|
-|aspnetCoreVersion|ASP.NET Core version: 3.0 (preview4 only), 2.2 (default), 2.1, 2.0 (deprecated)| |2.2|
-|swashbuckleVersion|Swashbucke version: 3.0.0 (default), 4.0.0| |3.0.0|
+|aspnetCoreVersion|ASP.NET Core version: 3.0 (preview4 only), 2.2, 2.1, 2.0 (deprecated)| |2.2|
+|swashbuckleVersion|Swashbucke version: 3.0.0, 4.0.0| |3.0.0|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
-|isFramework|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
+|useFrameworkReference|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
 |useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
 |useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -16,8 +16,8 @@ sidebar_label: aspnetcore
 |packageVersion|C# package version.| |1.0.0|
 |packageGuid|The GUID that will be associated with the C# project| |null|
 |sourceFolder|source folder for generated code| |src|
-|compatibilityVersion|ASP.Net Core CompatibilityVersion| |Version_2_1|
-|aspnetCoreVersion|ASP.NET Core version: 2.2 (default), 2.1, 2.0 (deprecated)| |2.2|
+|compatibilityVersion|ASP.Net Core CompatibilityVersion| |Version_2_2|
+|aspnetCoreVersion|ASP.NET Core version: 3.0 (preview4 only), 2.2 (default), 2.1, 2.0 (deprecated)| |2.2|
 |swashbuckleVersion|Swashbucke version: 3.0.0 (default), 4.0.0| |3.0.0|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
@@ -25,9 +25,9 @@ sidebar_label: aspnetcore
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
-|isFramework|Are we using packagereference or framework| |false|
+|isFramework|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
 |useNewtonsoft|Uses the Newtonsoft JSN library.| |true|
-|useDefaultRoutng|Uses the newend point routing JSN library.| |true|
+|useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
 |classModifier|Class Modifier can be empty, abstract| ||
 |operationModifier|Operation Modifier can be virtual, abstract or partial| |virtual|
 |buildTarget|Target to build an application or library| |program|
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
@@ -54,7 +54,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
     public static final String IS_LIBRARY = "isLibrary";
     public static final String USE_FRAMEWORK_REFERENCE = "useFrameworkReference";
     public static final String USE_NEWTONSOFT = "useNewtonsoft";
-    public static final String USE_DEFAULT_ROUTING = "useDefaultRoutng";
+    public static final String USE_DEFAULT_ROUTING = "useDefaultRouting";
     public static final String NEWTONSOFT_VERSION = "newtonsoftVersion";
 
     private String packageGuid = "{" + randomUUID().toString().toUpperCase(Locale.ROOT) + "}";
@@ -80,7 +80,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
     private boolean isLibrary = false;
     private boolean useFrameworkReference = false;
     private boolean useNewtonsoft = true;
-    private boolean useDefaultRoutng = true;
+    private boolean useDefaultRouting = true;
     private String newtonsoftVersion = "3.0.0-preview5-19227-01";
 
     public AspNetCoreServerCodegen() {
@@ -162,6 +162,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
 
         swashbuckleVersion.addEnum("3.0.0", "Swashbuckle 3.0.0");
         swashbuckleVersion.addEnum("4.0.0", "Swashbuckle 4.0.0");
+        swashbuckleVersion.addEnum("5.0.0", "Swashbuckle 5.0.0");
         swashbuckleVersion.setDefault("3.0.0");
         swashbuckleVersion.setOptValue(swashbuckleVersion.getDefault());
         addOption(swashbuckleVersion.getOpt(), swashbuckleVersion.getDescription(), swashbuckleVersion.getOptValue());
@@ -206,7 +207,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
 
         addSwitch(USE_DEFAULT_ROUTING,
                 "Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.",
-                useDefaultRoutng);
+                useDefaultRouting);
 
         classModifier.addEnum("", "Keep class default with no modifier");
         classModifier.addEnum("abstract", "Make class abstract");
@@ -558,13 +559,13 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
     private void setUseEndpointRouting() {
         if (aspnetCoreVersion.getOptValue().startsWith("3.")) {
             LOGGER.warn("ASP.NET core version is " + aspnetCoreVersion.getOptValue() + " so switching to old style endpoint routing.");
-            useDefaultRoutng = false;
-            additionalProperties.put(USE_DEFAULT_ROUTING, useDefaultRoutng);
+            useDefaultRouting = false;
+            additionalProperties.put(USE_DEFAULT_ROUTING, useDefaultRouting);
         } else {
             if (additionalProperties.containsKey(USE_DEFAULT_ROUTING)) {
-                useDefaultRoutng = convertPropertyToBooleanAndWriteBack(USE_DEFAULT_ROUTING);
+                useDefaultRouting = convertPropertyToBooleanAndWriteBack(USE_DEFAULT_ROUTING);
             } else {
-                additionalProperties.put(USE_DEFAULT_ROUTING, useDefaultRoutng);
+                additionalProperties.put(USE_DEFAULT_ROUTING, useDefaultRouting);
             }
         }
     }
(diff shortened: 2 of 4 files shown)


docs/generators/aspnetcore.md
```diff
@@ -16,14 +16,18 @@ sidebar_label: aspnetcore
...
+|aspnetCoreVersion|ASP.NET Core version: 3.0 (preview4 only), 2.2 (default), 2.1, 2.0 (deprecated)| |2.2|
+|swashbuckleVersion|Swashbucke version: 3.0.0 (default), 4.0.0| |3.0.0|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
+|isFramework|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
+|useNewtonsoft|Uses the Newtonsoft JSN library.| |true|
```
etherealjoy: Typo JSON

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -51,6 +52,9 @@
     public static final String SDK_LIB = "Microsoft.NET.Sdk";
     public static final String COMPATIBILITY_VERSION = "compatibilityVersion";
     public static final String IS_LIBRARY = "isLibrary";
+    public static final String IS_FRAMEWORK = "isFramework";
+    public static final String USE_NEWtONSOFT = "useNewtonsoft";
```
etherealjoy: typo NEWTONSOFT

etherealjoy:
I thought the signature got changed here
```C#
public void Configure(IApplicationBuilder app,  IWebHostEnvironment env)
```

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
```diff
@@ -42,11 +42,11 @@ namespace {{packageName}}
         {
             // Add framework services.
             services
-                .AddMvc()
+                .AddMvc({{^useEndpointRouting}}opts => opts.EnableEndpointRouting = false{{/useEndpointRouting}})
```
etherealjoy: I am wondering if for a web api it would make more sense to just use MvcCore

etherealjoy:
Does it makes sense to use static files when swashbuckle is not used? In pure WebApi scenarios?

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
A-Joshi: It probably might - for now the option to turn off the Swashbuckle support (useSwashbuckle) is there - eventually the default could be flipped. It might be a breaking change for people aleady using it so avoided doing that. 

I have tried to keep the old behaviour (particularly for 2.2 and earlier) unchanged as much as possible.

A-Joshi:
> I thought the signature got changed here
> 
> ```cs
> public void Configure(IApplicationBuilder app,  IWebHostEnvironment env)
> ```

Yes - but the old style is still available with a deprecation warning:
"Startup.cs(67,56): warning CS0618: 'IHostingEnvironment' is obsolete: 'This type is obsolete and will be removed in a future version. The recommended alternative is Microsoft.AspNetCore.Hosting.IWebHostEnvironment.' [/Src/contract-first/contracts/aspnetcore-server-task-async/target/contracts/src/Contracts/Contracts.csproj]"

There are a couple more such warnings that will have to be cleaned up (aside from support for new routing and json libraries) before it is fully ready.

docs/generators/aspnetcore.md
A-Joshi: Good catch -fixed

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Good catch - fixed

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
A-Joshi: Variable name was also wrong  should have been useDefaultRouting

etherealjoy:
@all
If it breaks current 2.2 version does not it make sense then to use a different version instead for the 3.0?
This would make it easier to maintain and also use the new features that come with the the 3.0 version

A-Joshi:
> @ALL
> If it breaks current 2.2 version does not it make sense then to use a different version instead for the 3.0?
> This would make it easier to maintain and also use the new features that come with the 3.0 version

Does not break 2.2 - did not want to send the PR till the break was fixed.

jimschubert:
Once #2816 is merged, can we mark this generator as "experimental"?

docs/generators/aspnetcore.md
```diff
@@ -16,14 +16,18 @@ sidebar_label: aspnetcore
...
+|swashbuckleVersion|Swashbucke version: 3.0.0 (default), 4.0.0| |3.0.0|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
+|isFramework|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
+|useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
+|useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
```
jimschubert: If "default routing" isn't supported in 3.0, I would think that this shouldn't be user-configurable until it's supported.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -60,18 +64,22 @@
     private boolean useSwashbuckle = true;
     protected int serverPort = 8080;
     protected String serverHost = "0.0.0.0";
-    protected CliOption aspnetCoreVersion = new CliOption(ASPNET_CORE_VERSION, "ASP.NET Core version: 2.2 (default), 2.1, 2.0 (deprecated)");
+    protected CliOption swashbuckleVersion = new CliOption(SWASHBUCKLE_VERSION, "Swashbucke version: 3.0.0 (default), 4.0.0");
     ; // default to 2.1
+    protected CliOption aspnetCoreVersion = new CliOption(ASPNET_CORE_VERSION, "ASP.NET Core version: 3.0 (preview4 only), 2.2 (default), 2.1, 2.0 (deprecated)");
```
jimschubert: We should remove default version definition from being embedded in the help text. This will already be displayed via `config-help`:

```
	aspnetCoreVersion
	    ASP.NET Core version: 3.0 (preview4 only), 2.2 (default), 2.1, 2.0 (deprecated) (Default: 2.2)
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -60,18 +64,22 @@
     private boolean useSwashbuckle = true;
     protected int serverPort = 8080;
     protected String serverHost = "0.0.0.0";
-    protected CliOption aspnetCoreVersion = new CliOption(ASPNET_CORE_VERSION, "ASP.NET Core version: 2.2 (default), 2.1, 2.0 (deprecated)");
+    protected CliOption swashbuckleVersion = new CliOption(SWASHBUCKLE_VERSION, "Swashbucke version: 3.0.0 (default), 4.0.0");
```
jimschubert: We should remove default version definition from being embedded in the help text. This will already be displayed via config-help:

```
	swashbuckleVersion
	    Swashbucke version: 3.0.0 (default), 4.0.0 (Default: 3.0.0)
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -445,16 +476,16 @@ private void setBuildTarget() {
...
-            embeddedTemplateDir = templateDir = "aspnetcore/2.0";
-            supportingFiles.add(new SupportingFile("web.config", packageFolder, "web.config"));
-            LOGGER.info("ASP.NET core version: 2.0");
-            compatibilityVersion = null;
-        } else {
-            // default, do nothing
-            LOGGER.info("ASP.NET core version: " + aspnetCoreVersion.getOptValue());
-            compatibilityVersion = "Version_" + aspnetCoreVersion.getOptValue().replace(".", "_");
-        }
+            if ("2.0".equals(aspnetCoreVersion.getOptValue())) {
```
jimschubert: bad indenting

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -482,4 +513,61 @@ private void setOperationIsAsync() {
             additionalProperties.put(OPERATION_IS_ASYNC, operationIsAsync);
         }
     }
+
+    private void setIsFramework() {
+        if (aspnetCoreVersion.getOptValue().startsWith("3.")) {
+            // default, do nothing
```
jimschubert: this block is setting a field and has a side effect of assigning a key/value in `additionalProperties`. Saying "do nothing" could be confusing. Similarly, "3.x" isn't the default defined in this type and could be confusing. Should probably remove comment since the log message below provides runtime and in-code explanation.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -482,4 +513,61 @@ private void setOperationIsAsync() {
...
+            LOGGER.warn("ASP.NET core version is " + aspnetCoreVersion.getOptValue() + " so changing  to use frameworkReference instead of packageReference ");
```
jimschubert: This sets `isFramework` which may indirectly affect `frameworkReference`, but the log message should describe what's being done by the generator rather than the template. For example, someone could create a custom template, and could find it confusing if they've hard-coded frameworkReference or something else, and yet the generator says it's "changing to use frameworkReference".

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -482,4 +513,61 @@ private void setOperationIsAsync() {
...
+                additionalProperties.put(IS_FRAMEWORK, isFramework);
+            }
+        }
+
+    }
+
+    private void setUseNewtonsoft() {
+        if (aspnetCoreVersion.getOptValue().startsWith("2.")) {
+            LOGGER.warn("ASP.NET core version is " + aspnetCoreVersion.getOptValue() + " so staying on default json library.");
+            useNewtonsoft = false;
```
jimschubert: I feel like there's either a typo here or on the conditional a couple lines above. Is it intentional to disallow JSON.net in aspnetcore 2.x? If so, this may be considered a breaking change and should probably be done separately from this PR to generate discussion.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -482,4 +513,61 @@ private void setOperationIsAsync() {
...
+            if (additionalProperties.containsKey(USE_DEFAULT_ROUTING)) {
+                useDefaultRoutng = convertPropertyToBooleanAndWriteBack(USE_DEFAULT_ROUTING);
+            } else {
+                additionalProperties.put(USE_DEFAULT_ROUTING, useDefaultRoutng);
+            }
+        }
+    }
+
+    private void setSwashbuckleVersion() {
+        setCliOption(swashbuckleVersion);
```
jimschubert: Long comment, but minor (can be addressed later).

Only the CLI should be setting CLI options. We should read options once (in `processOpts`), then have nothing to do with CLIOption types. That is, we should load into fields and work from those fields directly. I'm concerned that making this method side-effecting as it is and assuming options parsing within the generator, things can become out of sync. 

For example, a user can pass `--additional-properties=swashbuckleVersion=3.0.5". Although this isn't defined in the enum of available versions, this is a valid use case that we allow for properties defined in `additionalProperties`; this is why you often see the guard that we only set the value if it doesn't already exist.  `CodegenConfigurator` also allows for JSON deserialization into the `CodegenConfigurator` type (to load these values from a config), which sorta bypasses both the cli options and additional properties (see [CodegenConfigurator](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java#L634)).

In all of these cases, this method will reset the user-provided value to whatever is originally hardcoded as the swashbuckleVersion default within the generator.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
```diff
@@ -42,11 +42,11 @@ namespace {{packageName}}
         {
             // Add framework services.
             services
-                .AddMvc()
+                .AddMvc({{^useDefaultRoutng}}opts => opts.EnableEndpointRouting = false{{/useDefaultRoutng}})
                 {{#compatibilityVersion}}
                 .SetCompatibilityVersion(CompatibilityVersion.{{compatibilityVersion}})
                 {{/compatibilityVersion}}
-                .AddJsonOptions(opts =>
+                .{{#useNewtonsoft}}AddNewtonsoftJson{{/useNewtonsoft}}{{^useNewtonsoft}}AddJsonOptions{{/useNewtonsoft}}(opts =>
```
jimschubert: Variable `useNewtonsoft` will also need to be evaluated in `controller.mustache` and `model.mustache`. Currently, the artifact won't be loaded, but controllers and models are still tightly coupled to json serialize/deserialize using the library.

docs/generators/aspnetcore.md
A-Joshi: I have put it in prep for the future and also because the support for 2.1 needed the configuration.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: The change is not a breaking change:

This one is a little tricky - the MVC configuration in 2.x and 3.x use AddJsonOptions but mean different things: the default JSON in 2.x is implicitly Newtonsoft, while in 3.x and later it will be the Microsoft one. 

I can change it use something like useDefaultJson or jsonLibrary. I don't like the former since the default in different cases means different things. Maybe explicitly defining the jsonLibrary as Newtonsoft or Microsoft but that makes the mustache file complex - since mustache does not have conditionals based on values.

What are your thoughts there?

jimschubert: My main concern is that for 2.x, when `useNewtonSoft` is false, this results in the NewtonSoft library not being referenced, but is still used on controllers and models. From a maintainability perspective (both built-in templates and custom user templates), I find that very confusing.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Removed the default in the docstring.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Fixed.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: Changed the variable to useFrameworkReference to make things explicitly clear. Its one of the breaking differences between 2.x and 3.x - for the framework the reference has to be by FrameworkReference. 

I think if people hard code stuff that contradicts the setting there might be little we can do about it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
A-Joshi: The other choice here is to check the version of Swashbuckle is valid for the ASP.NET Core chosen and  throw an error if the version is not valid. 

In this particular case I think you can use v 4.0.0 with version 2.x and 3.x but you cannot use v3.0.0 with 3.x and later.

What are you thoughts here - error or warn and correct - not sure what the general policy for this is.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
A-Joshi: see the earlier comment about setting the value of useNewtonsoft.

A-Joshi:
I am beginning to think that instead of automatically correcting the value if invalid, then just throw an error. It will simplify some of the logic.  

Thoughts?

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
jimschubert: I'm not sure I follow.

To put the concern a different way, as a user if I set `useNewtonSoft=false`, and the controller/model generation still includes NewtonSoft imports, I'd consider that a bug. Maybe the variable name is what's misleading here.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
jimschubert: I think my preference would be to set a default if one doesn't exist, and warn if it's something we don't expect will work. We'll probably want to be clear that anything below 4.0.0 _may_ not work, but trust that the user understands the versioning.  For example, someone might have a fork of Swashbuckle which does work, and they could reference and add the file with the reference to `.openapi-generator-ignore`. In that case, a warn+correct would be unexpected if the value was explicitly provided.

etherealjoy:
> The PR does not address supporting the new features of 3.0 - it simply ensure that you can use 3.0 in a backward compatible way.

Here is what I also would like to hear the opinion. Because if we use the new features of 3.0 we will have to use a new template otherwise maintaining backwards compatibility will be too much. 
Besides it also make contribution from more people easier to manage due to lesser chance of breaking 2.x

wing328:
What about merging this PR into master and then ask the community for feedback?

etherealjoy:
This is not going to work yet. Routing changes are missing. 
I think we can stick with JSON.NET instead of Utf8JsonReader/Writer. Utf8JsonReader is still lacking many features. 
Additionally I believe ASP.NET Core 3.0 brings so many new things that it makes more sense to have it as a separate generator. e.g use ControllerWithView and Razor at the same time instead of Swashbucke with Mvc in the past.

etherealjoy:
@wing328 @A-Joshi 
I would like to give to give feedback after 3 weeks of using ASP.NET Core 3.0 Preview 5.
- `Utf8JsonReader` is not feature complete so we are better off for now using `JSON.NET`
- API Controllers and Controller with views use the Utf8JsonReader so incompatibility with net core client models. So we should continue to use ASP.NetCore MVC package
- We can add Razor to to MVC so not an issue there.
- We can continue to use the same generator because the diff from here mainly is the Configure.

New changes needed 
- In `Configure`, endpoint routing has to be activated
- `JSON.NET` has to be added a an external dependency in `.csproj`
- JSON camel case NamingStartegy has to use instead of deprecated option.
- docker alpine runtime is available
- rest is same :)

A-Joshi:
Thanks - will take a look this week. Was out of action for some personal
reasons.

A-Joshi:
Could you give sample of what the changes you needed to do were?

> @wing328 @A-Joshi
> New changes needed
> 
>     * In `Configure`, endpoint routing has to be activated
What did you have to change? parameter value to be set?
> 
>     * `JSON.NET` has to be added a an external dependency in `.csproj`
What was the dependency you needed to add? I think its already there.
> 
>     * JSON camel case NamingStartegy has to use instead of deprecated option.
Not sure what the change needed was?
> 
>     * docker alpine runtime is available
Can look at this for the next round.
> 
>     * rest is same :)


Can we get the PR added in ? It does not break 2.x and it will be easier to work against the master once this is merged in.

etherealjoy:
I did these

> In Configure, endpoint routing has to be activated

In `Configure`, shown here without `swashbuckle`
```
public void Configure(IApplicationBuilder app,  IWebHostEnvironment env)
{
	app.UseHttpsRedirection();
	app.UseRouting();
	app.UseEndpoints(endpoints =>
	{
		endpoints.MapControllers();
	});	
	if (env.IsDevelopment())
	{
		app.UseDeveloperExceptionPage();
	}
	else
	{
		app.UseHsts();
	}
}
```


> JSON.NET has to be added a an external dependency in .csproj

In `.csproj` file
```
<PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" 
          Version="3.0.0-preview5-19227-01" />
```

> JSON camel case `NamingStrategy` has to use instead of deprecated option.


```
using Newtonsoft.Json.Converters;
using Newtonsoft.Json.Serialization;


public void ConfigureServices(IServiceCollection services)
{
	// Add framework services.
	services
		.AddMvc()
		.SetCompatibilityVersion(CompatibilityVersion.Version_3_0)
		.AddNewtonsoftJson(opts =>
			{
				opts.SerializerSettings.ContractResolver = 
						new CamelCasePropertyNamesContractResolver();
				opts.SerializerSettings.Converters.Add(new StringEnumConverter
				{
					NamingStrategy = new CamelCaseNamingStrategy()
				});
			}
		);

}
```

> docker alpine runtime is available

In `Dockerfile`
`FROM mcr.microsoft.com/dotnet/core/aspnet:3.0-alpine AS runtime`

A-Joshi:
@etherealjoy:
Added the parts you found above except the docker image - have to give it a little thought whether we should make the whole base docker image name configurable or pass in the parameter like alpine.

Can you test the new push please?

@wing328, @jimschubert  - Assuming @etherealjoy  is good then I suggest we merge it in - there are no changes for the version 2.1/2.2 while this will add preliminary support for 3.0 preview 5 in compatibility mode.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
```diff
@@ -42,16 +42,16 @@ namespace {{packageName}}
...
                 {
                     opts.SerializerSettings.ContractResolver = new CamelCasePropertyNamesContractResolver();
                     opts.SerializerSettings.Converters.Add(new StringEnumConverter
                     {
-                        CamelCaseText = true
```
etherealjoy: this option is needed for older 2.2/2.1 framework version.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
etherealjoy: For `2.x` `NewtonSoft` is used by the framework by default itself because it is part of the fw, but MS removed it from the framework starting from `3.x` and manual addition of package is needed. 
Which in this case the user really has no choice to not use NewtonSoft in `2.x`
Since we would need to use `NewtonSoft` for model compatibility reasons it does not makes sense to allow the user to set this flag because if the fw is > `3.x` we will need to add it

etherealjoy:
I will try this out and give feedback.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
etherealjoy: Indeed, since the models are generated with `Netwonsoft` in mind it is required that in `3.x` it is included via the package dependency and for `2.x` is there as part of core fw.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
A-Joshi: It is generated for older frameworks - the code generated is conditional based on 2.1/2.2 or 3.0+
Eventually we might want to make the CamelCase choice configurable in case of people wanting to control that.

jimschubert:
Regarding the recommendation of having 3.0 be a separate generator…

Rather than conditionally assigning locations for files between versions (e.g. `2.2/filename.mustache` and `3.0/filename.mustache`), why not just use the core generation's `library` feature? These are different versions of the same library, but there's no reason we can't use that option to reference library versions. This allows us to put common code under `resources/aspnetcore` and any version-specific overrides under their respective library folders. This would be a breaking change for anyone who has extended templates, so such a change would have to be done in release 4.1.0. This is just a thought about managing the concern of maintainability as Microsoft makes breaking changes between versions. I think it would be much easier to manage these as libraries rather than completely separate generators.

A-Joshi:
Sounds like a good idea - for now though I would like to keep the
current PR focused on the current changes and then add new changes/ideas
later via different PR.

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
```diff
@@ -109,7 +109,13 @@ namespace {{packageName}}
 
                     //TODO: Or alternatively use the original Swagger contract that's included in the static files
                     // c.SwaggerEndpoint("/openapi-original.json", "{{#appName}}{{{appName}}}{{/appName}}{{^appName}}{{packageName}}{{/appName}} Original");
-                }){{/useSwashbuckle}};
+                }){{/useSwashbuckle}};{{^useDefaultRoutng}}
+            	app.UseHttpsRedirection();
```
etherealjoy: This is already done at the beginning of the method.

etherealjoy:
I tried this and it works using these options.
```shell
java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i \ 
/home/dev/api/swagger.yaml -o /home/dev/gen --additional-properties  \
aspnetCoreVersion=3.0,useNewtonsoft=true,useDefaultRoutng=false,useSwashbuckle=false \ 
-g aspnetcore
```
However `SwashBuckle` on `3.0` needs version `5.0`.
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1088
So Swashbuckle needs and update for `3.0`

docs/generators/aspnetcore.md
```diff
@@ -16,14 +16,19 @@ sidebar_label: aspnetcore
...
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |returnICollection|Return ICollection&lt;T&gt; instead of the concrete type.| |false|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
 |isLibrary|Is the build a library| |false|
+|useFrameworkReference|Use frameworkReference for ASP.NET Core 3.0+ and  PackageReference  ASP.NET Core 2.2 or earlier.| |false|
+|useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
+|newtonsoftVersion|Version for Microsoft.AspNetCore.Mvc.NewtonsoftJson for ASP.NET Core 3.0+| |3.0.0-preview4-19216-03|
+|useDefaultRoutng|Use default routing for the  ASP.NET Core version. For 3.0 turn off default because it is not yet supported.| |true|
```
etherealjoy: typo `useDefaultRoutng`

A-Joshi: Good catch changed the name to useDefaultRouting

modules/openapi-generator/src/main/resources/aspnetcore/2.1/Startup.mustache
A-Joshi: Removed the duplicate

A-Joshi:
You can change the swashbuckle version using the parameter "swashbuckleVersion=4.0.0" I have added 5.0.0 as a possible value but have not tested with that.

A-Joshi:
@etherealjoy are you OK with it now?
@jimschubert @wing328  can we merge this one?

etherealjoy:
Ok from my side.

etherealjoy:
Update to swashbuckle 5 is needed to use it with 3.0 but we can do this later.

wing328:
Let's merge this and gather feedback from the community.

PR #2861, opened 2019-05-10
[java/resttemplate] Generate valid code if no Authentication implementations present

jonfreedman:
Take the logic used to decide which instances to add to the authentications map and re-use to not import classes or offer non-functional util methods

@bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger @karismann @Zomzog

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/resttemplate/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/resttemplate/ApiClient.mustache
@@ -65,13 +65,11 @@ import java.util.Map.Entry;
 import java.util.TimeZone;
 
 import {{invokerPackage}}.auth.Authentication;
-import {{invokerPackage}}.auth.HttpBasicAuth;
-import {{invokerPackage}}.auth.HttpBearerAuth;
-import {{invokerPackage}}.auth.ApiKeyAuth;
-{{#hasOAuthMethods}}
-import {{invokerPackage}}.auth.OAuth;
-{{/hasOAuthMethods}}
-
+{{#authMethods}}{{#isBasic}}{{#isBasicBasic}}
+import {{invokerPackage}}.auth.HttpBasicAuth;{{/isBasicBasic}}{{^isBasicBasic}}
+import {{invokerPackage}}.auth.HttpBearerAuth;{{/isBasicBasic}}{{/isBasic}}{{#isApiKey}}
+import {{invokerPackage}}.auth.ApiKeyAuth;{{/isApiKey}}{{#isOAuth}}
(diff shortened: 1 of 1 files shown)


wing328:
@jonfreedman thanks for the PR. Look like the change will results in duplicated import `org.openapitools.client.auth.ApiKeyAuth`:
```diff
--- a/samples/client/petstore/java/resttemplate-withXml/src/main/java/org/openapitools/client/ApiClient.java
+++ b/samples/client/petstore/java/resttemplate-withXml/src/main/java/org/openapitools/client/ApiClient.java
@@ -59,12 +59,12 @@ import java.util.Map.Entry;
 import java.util.TimeZone;
 
 import org.openapitools.client.auth.Authentication;
-import org.openapitools.client.auth.HttpBasicAuth;
-import org.openapitools.client.auth.HttpBearerAuth;
+
+import org.openapitools.client.auth.ApiKeyAuth;
 import org.openapitools.client.auth.ApiKeyAuth;
+import org.openapitools.client.auth.HttpBasicAuth;
 import org.openapitools.client.auth.OAuth;
```

jonfreedman:
Can you please check the population of the authentications map, if there's a duplicate import I guess there must be a duplicate put too?
If that's the case do you think it's better to add more variables like hasOAuthMethods instead?

wing328:
Yes, we have `{{#hasOAuthMethods}} ... {{/hasOAuthMethods}}`, e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/pom.mustache#L221

jonfreedman:
Yep I saw that one, but would also need hasBasicAuthMethods, hasTokenAuthMethods etc. which I assume don't exist yet?

wing328:
> but would also need hasBasicAuthMethods, hasTokenAuthMethods etc. which I assume don't exist yet?

Can you please add hasBasicAuthMethods, hasTokenAuthMethods, etc as part of this PR?

jonfreedman:
I can give it a go - can you point me in the right direction?

wing328:
https://github.com/OpenAPITools/openapi-generator/blob/9323cad06ba7a42dee3219f5349f516b201860fa/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java#L1276 is a good starting point.

And then add similar checks in https://github.com/OpenAPITools/openapi-generator/blob/9323cad06ba7a42dee3219f5349f516b201860fa/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java#L842

wing328:
Closing this as there's no update. Please open a new one when you've time.

Thanks again for the PR.

PR #2879, opened 2019-05-13
[Groovy] improve code parsing body/form params

karismann:
### PR checklist

### Description of the PR

- improve code parsing body/form params
- add removeOption for CliOptions in DefaultCodeGen class 
- remove overrided method identical to parent (AbstractJavaCodeGen and AbstractApexCodeGen)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -1456,6 +1456,17 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
         return booleanGetterPrefix + getterAndSetterCapitalize(name);
     }
 
+    @Override
+    public String sanitizeTag(String tag) {
+        tag = camelize(underscore(sanitizeName(tag)));
+
+        // tag starts with numbers
+        if (tag.matches("^\\d.*")) {
+            tag = "Class" + tag;
+        }
+        return tag;
+    }
+
     /**
      * Camelize the method name of the getter and setter
      *
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -1060,7 +1060,7 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
         String defaultContentType = "application/json";
         Set<String> producesInfo = getProducesInfo(openAPI, operation);
         if (producesInfo != null && !producesInfo.isEmpty()) {
-            ArrayList<String> produces = new ArrayList<String>(producesInfo);
+            ArrayList<String> produces = new ArrayList<>(producesInfo);
             StringBuilder sb = new StringBuilder();
             for (String produce : produces) {
                 if (defaultContentType.equalsIgnoreCase(produce)) {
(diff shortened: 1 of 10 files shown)


modules/openapi-generator/src/main/resources/Groovy/api.mustache
```diff
@@ -40,44 +40,20 @@ class {{classname}} {
         {{/headerParams}}
 
         {{#bodyParam}}
-        {{^consumes}}
-        contentType = 'application/json';
-        {{/consumes}}
-        {{#consumes.0}}
-        contentType = '{{{mediaType}}}';
-        {{/consumes.0}}
+        {{^consumes}}contentType = 'application/json';{{/consumes}}
```
wing328: You can do the following instead to avoid leading spaces:
```
{{^consumes}}
contentType = 'application/json';
{{/consumes}}
```

modules/openapi-generator/src/main/resources/Groovy/api.mustache
```diff
@@ -40,44 +40,20 @@ class {{classname}} {
...
+        {{#consumes.0}}contentType = '{{{mediaType}}}';{{/consumes.0}}
```
wing328: Same as above.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
```diff
@@ -80,13 +81,6 @@ public JavaJAXRSSpecServerCodegen() {
 
         super.embeddedTemplateDir = templateDir = JAXRS_TEMPLATE_DIRECTORY_NAME + File.separator + "spec";
 
-        for (int i = 0; i < cliOptions.size(); i++) {
-            if (CodegenConstants.LIBRARY.equals(cliOptions.get(i).getOpt())) {
-                cliOptions.remove(i);
-                break;
-            }
-        }
-
```
wing328: Why these are removed? I don't even recall the reason why we've this block of code.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -1444,42 +1445,17 @@ public String toRegularExpression(String pattern) {
...
-    public String sanitizeTag(String tag) {
-        tag = camelize(underscore(sanitizeName(tag)));
-
-        // tag starts with numbers
-        if (tag.matches("^\\d.*")) {
-            tag = "Class" + tag;
-        }
-        return tag;
-    }
-
```
wing328: Would removing this cause issue with tag starting with numbers?

karismann: because parent method (DefautCodeGen class) is actually doing exactly the same things so this overriding is useless

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
karismann: it's not removed but replaced by a generic method (removeOption in line 62) to be used in others methods

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
wing328: Understood. If I remember correctly, the reason is that we don't want generators to use the default defined in the DefaultCodegen class as changing the default usually results in a lot of changes to at least a few generators. That's why we prefer to override the methods in the language generator class.

It's perfectly fine to remove it in this PR but later you may see us adding it back and slightly revising the logic to address certain edge cases.

karismann: i understand the logic, thanks for the explanation

karismann: anyway i will remove this change with your other review with another commit

Issue #2022, opened 2019-01-30
[BUG][Python][Python-Aiohttp] util._deserialize throws AttributeError in Python 3.7


bejohu:
##### Description

```
Traceback (most recent call last):
  File "*******/venv/lib64/python3.7/site-packages/aiohttp/web_protocol.py", line 418, in start
    resp = await task
  File "*******/venv/lib64/python3.7/site-packages/aiohttp/web_app.py", line 458, in _handle
    resp = await handler(request)
  File "*******/venv/lib64/python3.7/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "*******/venv/lib64/python3.7/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "*******/venv/lib64/python3.7/site-packages/connexion/apis/aiohttp_api.py", line 32, in oauth_problem_middleware
    response = yield from handler(request)
  File "*******/venv/lib64/python3.7/site-packages/connexion/decorators/coroutine_wrappers.py", line 23, in wrapper
    connexion_response = yield from connexion_response
  File "*******/venv/lib64/python3.7/site-packages/connexion/apis/aiohttp_api.py", line 198, in get_response
    response = yield from response
  File "*******/openapi_server/controllers/search_controller.py", line 22, in search
    search_results = SearchResults.from_dict(data)
  File "*******/openapi_server/models/search_results.py", line 56, in from_dict
    return util.deserialize_model(dikt, cls)
  File "*******/openapi_server/util.py", line 106, in deserialize_model
    setattr(instance, attr, _deserialize(value, attr_type))
  File "*******/openapi_server/util.py", line 29, in _deserialize
    elif type(klass) == typing.GenericMeta:
AttributeError: module 'typing' has no attribute 'GenericMeta'
```

* `typing.GenericMeta` is [not available](https://github.com/python/typing/issues/468) in Python 3.7
* Relying on `__extra__` is [not recommended](https://github.com/python/typing/issues/136#issuecomment-138174590)

##### openapi-generator version

4.0.0-SNAPSHOT

##### Suggested fix

In `_deserialize` in [util.mustache](modules/openapi-generator/src/main/resources/python-aiohttp/util.mustache) use [typing_inspect](https://pypi.org/project/typing-inspect/), package which [might get included](https://github.com/python/typing/issues/570) in [typing](https://docs.python.org/3/library/typing.html) anyway.

[...]

Issue #1866, opened 2019-01-09
[REQ] Python-Flask - Python 3.7 util._deserialization


nathan5280:
### Is your feature request related to a problem? Please describe.
Unable to deserialize model objects when using Python 3.7.

When running with Python 3.7 openapi_server.util._deserialize fails because typing.GenericMeta was removed from 3.7.

## Describe the solution you'd like

I'm not sure on where we are at with support for 3.7, but I worked on this fix and figured I should pass it along for when we get ready to support 3.7.  Other than this I haven't found other issues.  

### Python 3.6
```python
def _deserialize(data, klass):
    """Deserializes dict, list, str into an object.

    :param data: dict, list or str.
    :param klass: class literal, or string of class name.

    :return: object.
    """
    from cdst_places_app.openapi_server.models.base_model_ import Model

    if data is None:
        return None

    if klass in six.integer_types or klass in (float, str, bool):
        return _deserialize_primitive(data, klass)
    elif klass == object:
        return _deserialize_object(data)
    elif klass == datetime.date:
        return deserialize_date(data)
    elif klass == datetime.datetime:
        return deserialize_datetime(data)
    elif type(klass) == typing.GenericMeta:
        if klass.__extra__ == list:
            return _deserialize_list(data, klass.__args__[0])
        if klass.__extra__ == dict:
            return _deserialize_dict(data, klass.__args__[1])
    else:
        return deserialize_model(data, klass)
```
### Python 3.7
```python
import typing_inspect


def _deserialize(data, klass):
    """Deserializes dict, list, str into an object.

    :param data: dict, list or str.
    :param klass: class literal, or string of class name.

    :return: object.
    """
    from cdst_places_app.openapi_server.models.base_model_ import Model

    if data is None:
        return None

    if klass in six.integer_types or klass in (float, str, bool):
        return _deserialize_primitive(data, klass)
    elif klass == object:
        return _deserialize_object(data)
    elif klass == datetime.date:
        return deserialize_date(data)
    elif klass == datetime.datetime:
        return deserialize_datetime(data)
    elif typing_inspect.is_generic_type(klass):
        if typing_inspect.get_origin(klass) == list:
            return _deserialize_list(data, klass.__args__[0])
        if typing_inspect.get_origin(klass) == dict:
            return _deserialize_dict(data, klass.__args__[1])
    else:
        return deserialize_model(data, klass)
```
I'm not sure if the dict piece
```python
[...]

PR #2884, opened 2019-05-13
[python-server] Support python 3.5+ for all server-generators

GuillaumeSmaha:
### PR checklist

- [x] Copied the [technical committee] @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07)  @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess  (2019/01)

### Description of the PR
This change allows support for python 3.5+ on all python-server-generators.
There are 2 issues requiring this fix:
- https://github.com/OpenAPITools/openapi-generator/issues/1866
- https://github.com/OpenAPITools/openapi-generator/issues/2022

PS: After running ./run/{LANG}-petstore.sh, I got change on the version and unrelated stuff like `-        - write:pets`. Should I revert this change ?


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
```diff
@@ -201,6 +201,7 @@ public void processOpts() {
         }
         supportingFiles.add(new SupportingFile("__main__.mustache", packagePath(), "__main__.py"));
         supportingFiles.add(new SupportingFile("util.mustache", packagePath(), "util.py"));
+        supportingFiles.add(new SupportingFile("typing_patch.mustache", packagePath(), "typing_patch.py"));
```
wing328: Suggestion: what about naming it as `typing_utils.py` instead?

Patch sounds like the code is more like a workaround and not production ready

GuillaumeSmaha: Yes, it's is better with `typing_utils.py`

wing328:
@GuillaumeSmaha when you've time, can we have a quick chat via https://gitter.im (ID: wing328) as I've a question about this PR?

Issue #411, opened 2018-06-28
CI: Rework publish pipeline for "openapi-generator-gradle-plugin"


jmini:
The automated deploy of the `openapi-generator-gradle-plugin` is not working.

With `3.0.1` and `3.0.2` there was a TravisCI timeout in the job.
With `3.0.3` the TravisCI Job is OK.

Log: https://api.travis-ci.org/v3/job/397359374/log.txt

Maven deploy:

Source:
https://github.com/OpenAPITools/openapi-generator/blob/9b909df543b8c541a17aa5a545473e6e511b140d/.travis.yml#L108-L109

Log Extract: end of the maven deploy step:
```
Finished mvn clean deploy for master
```

Gradle Uplaoad:

Source:
https://github.com/OpenAPITools/openapi-generator/blob/9b909df543b8c541a17aa5a545473e6e511b140d/.travis.yml#L112-L113

Log Extract for the `uploadArchives` task:
```
~/build/OpenAPITools/openapi-generator ~/build/OpenAPITools/openapi-generator

Welcome to Gradle 4.7!

Here are the highlights of this release:
 - Incremental annotation processing
 - JDK 10 support
 - Grouped non-interactive console logs
 - Failed tests are re-run first for quicker feedback

For more details see https://docs.gradle.org/4.7/release-notes.html

> Task :compileKotlin UP-TO-DATE
> Task :compileJava NO-SOURCE
> Task :pluginDescriptors UP-TO-DATE
> Task :processResources UP-TO-DATE
> Task :classes UP-TO-DATE
> Task :jar UP-TO-DATE
> Task :javadoc NO-SOURCE
> Task :javadocJar UP-TO-DATE
> Task :sourcesJar UP-TO-DATE
> Task :signArchives
> Task :uploadArchives

BUILD SUCCESSFUL in 3m 42s
8 actionable tasks: 2 executed, 6 up-to-date
Finished ./gradlew uploadArchives
```

The artefacts were uploaded to Nexus. But a look in the Staging repository show multiple staging repo:

https://oss.sonatype.org/#stagingRepositories

<img width="617" alt="Nexus" src="https://user-images.githubusercontent.com/1222165/42016283-48f95e18-7aab-11e8-95d2-2f9128956d25.png">

Content of the different repos:

* `openapitools-1031` contains only:
[...]

PR #2886, opened 2019-05-14
[gradle] Reworking publishing pipeline

jimschubert:
### PR checklist

### Description of the PR

TravisCI proxies separate external requests with different orginating IP
addresses, while Sonatype associates artifacts for auto-generated
repositories by IP address. This leads to many gradle deploys from CI
resulting in "split" staging repositories with no way to combine in
Sonatype Nexus.

This introduces a workflow which should resolve this issue on the next
revision release. Specifically, nexus-publish-plugin is included to
create singular staging repositories from TravisCI and
gradle-nexus-staging-plugin is included to auto-release and promote
this repository.

NOTE:
We need to publish via `publishPluginMavenPublicationToNexusRepository`,
because `publishToNexus` will publish _all_ publish-related tasks,
including the one intended only for the Gradle Plugin Portal.

Tested in standalone open source Nexus Repo Manager, which doesn't
support the staging plugin. So, only SNAPSHOT workflow has been
validated locally.

The `enableFeaturePreview('STABLE_PUBLISHING')` option in settings.gradle is a warning presented by the new nexus-publish-plugin:

> As part of making the publishing plugins stable, the 'deferred configurable' behavior of the 'publishing {}' block has been deprecated. In Gradle 5.0 the 'enableFeaturePreview('STABLE_PUBLISHING')' flag will be removed and the new behavior will become the default. Please add 'enableFeaturePreview('STABLE_PUBLISHING')' to your settings file and do a test run by publishing to a local repository. If all artifacts are published as expected, there is nothing else to do. If the published artifacts change unexpectedly, please see the migration guide for more details: https://docs.gradle.org/4.10.2/userguide/publishing_maven.html#publishing_maven:deferred_configuration.

This PR is an attempt at resolving #411

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ after_success:
         echo "Finished ./gradlew publishPluginMavenPublicationToNexusRepository closeAndReleaseRepository";
         popd;
       fi;
-      if [ -n $TRAVIS_TAG ] && [[ "$TRAVIS_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+      if [ -n $TRAVIS_TAG ] && [[ "$TRAVIS_TAG" =~ ^[v][0-9]+\.[0-9]+\.[0-9]+$ ]]; then
         echo "Publishing the gradle plugin to Gradle Portal on tag $TRAVIS_TAG (only)";
         pushd .;
         cd modules/openapi-generator-gradle-plugin;
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,8 +136,8 @@ after_success:
         echo "Finished mvn clean deploy for $TRAVIS_BRANCH";
         pushd .;
         cd modules/openapi-generator-gradle-plugin;
-        ./gradlew -Psigning.keyId="$SIGNING_KEY" -Psigning.password="$SIGNING_PASSPHRASE" -Psigning.secretKeyRingFile="${TRAVIS_BUILD_DIR}/sec.gpg" -PossrhUsername="${SONATYPE_USERNAME}" -PossrhPassword="${SONATYPE_PASSWORD}" publishPluginMavenPublicationToNexusRepository --no-daemon;
-        echo "Finished ./gradlew publishPluginMavenPublicationToNexusRepository";
+        ./gradlew -Psigning.keyId="$SIGNING_KEY" -Psigning.password="$SIGNING_PASSPHRASE" -Psigning.secretKeyRingFile="${TRAVIS_BUILD_DIR}/sec.gpg" -PossrhUsername="${SONATYPE_USERNAME}" -PossrhPassword="${SONATYPE_PASSWORD}" publishPluginMavenPublicationToNexusRepository closeRepository --no-daemon;
+        echo "Finished ./gradlew publishPluginMavenPublicationToNexusRepository closeRepository";
         popd;
       elif ([[ "$TRAVIS_BRANCH" =~ ^[0-9]+\.[0-9]+\.x$ ]]) ; then
         mvn clean deploy --settings CI/settings.xml;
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -136,8 +136,8 @@ after_success:
         echo "Finished mvn clean deploy for $TRAVIS_BRANCH";
         pushd .;
         cd modules/openapi-generator-gradle-plugin;
-        ./gradlew -Psigning.keyId="$SIGNING_KEY" -Psigning.password="$SIGNING_PASSPHRASE" -Psigning.secretKeyRingFile="${TRAVIS_BUILD_DIR}/sec.gpg" -PossrhUsername="${SONATYPE_USERNAME}" -PossrhPassword="${SONATYPE_PASSWORD}" uploadArchives --no-daemon;
-        echo "Finished ./gradlew uploadArchives";
```
jimschubert: If the 4.0.1 release fails, we may consider reverting to the `uploadArchives` and `maven` publishing plugin.

.travis.yml
```diff
@@ -136,8 +136,8 @@ after_success:
...
+        ./gradlew -Psigning.keyId="$SIGNING_KEY" -Psigning.password="$SIGNING_PASSPHRASE" -Psigning.secretKeyRingFile="${TRAVIS_BUILD_DIR}/sec.gpg" -PossrhUsername="${SONATYPE_USERNAME}" -PossrhPassword="${SONATYPE_PASSWORD}" publishPluginMavenPublicationToNexusRepository --no-daemon;
+        echo "Finished ./gradlew publishPluginMavenPublicationToNexusRepository";
```
jmini: We need to see how this goes, but I guess that calling the task `closeAndReleaseRepository` would also be necessary. (the maven nexus plugin is doing something like this)

jmini:
Because of the change `maven` to `maven-publish` I have updated the helper-pom to use the task `publishToMavenLocal`, see 6b0c3fa

jmini:
Now it is failing because of this error:

> Cannot perform signing task ':signMavenJavaPublication' because it has no configured signatory

I think there is a way to say that signing is only for release version:
https://docs.gradle.org/current/userguide/signing_plugin.html#sec:conditional_signing

jimschubert:
@jmini thanks for also looking into this. Seems I'll need to work out the release signing for maven-publish. I'll try and get to that on the next day or so.

.travis.yml
jimschubert: This is actually the SNAPSHOT path (master), but your comment made me realize that I had completely missed updating the `uploadArchives` in the following conditional block which is used for releases. I believe the releases branch is the only one that requires closing.

jimschubert:
In digging a little more, it seems we will need to call `closeRepository` on master (for snapshots) and  `closeAndReleaseRepository` on versioned branches (for releases).

Here's the task graph of the target task:

```
\--- :publishPluginMavenPublicationToNexusRepository
     +--- :generatePomFileForPluginMavenPublication
     +--- :initializeNexusStagingRepository
     +--- :jar
     |    +--- :classes
     |    |    +--- :compileJava
     |    |    |    \--- :compileKotlin
     |    |    \--- :processResources
     |    |         \--- :pluginDescriptors
     |    +--- :compileKotlin
     |    \--- :inspectClassesForKotlinIC
     |         +--- :classes
     |         |    +--- :compileJava
     |         |    |    \--- :compileKotlin
     |         |    \--- :processResources
     |         |         \--- :pluginDescriptors
     |         \--- :compileKotlin
     \--- :signPluginMavenPublication
          +--- :generatePomFileForPluginMavenPublication
          \--- :jar
               +--- :classes
               |    +--- :compileJava
               |    |    \--- :compileKotlin
               |    \--- :processResources
               |         \--- :pluginDescriptors
               +--- :compileKotlin
               \--- :inspectClassesForKotlinIC
                    +--- :classes
                    |    +--- :compileJava
                    |    |    \--- :compileKotlin
                    |    \--- :processResources
                    |         \--- :pluginDescriptors
                    \--- :compileKotlin
```

`initializeNexusStagingRepository` will create the repository with our defined staging id.  iirc from discussion with an engineer at Sonatype, these repositories are cleaned after 24 hours anyway, but I wouldn't want an open "snapshot" repository to become mixed in anyway with a release repository on the same day.

PR #2388, opened 2019-03-13
PR for ability to easily debug newly created codegen classes.

jason-henriksen:
### PR checklist

- [ X ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ X ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh`, `./bin/security/{LANG}-petstore.sh` and `./bin/openapi3/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ X ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [ N/A ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Adds a unit test to the meta command generated code.  Running that unit test under the debugger causes the cli code to be run under the debugger which makes the debug experience of working on a new codegen library much easier to work with and enhance.

PR #2896, opened 2019-05-15
[meta] remove cli dependency

jmini:
### PR checklist

### Description of the PR
Introduced with https://github.com/OpenAPITools/openapi-generator/pull/2388, newly generated generator created with CLI command:

```
openapi-generator meta -n myClientCodegen -t DOCUMENTATION -p com.my.company.example -o out
```

has a dependency on: `org.openapitools:openapi-generator-cli`

See the corresponding pom in the generated samples:

https://github.com/OpenAPITools/openapi-generator/blob/cbe39ba881c42a3d70a36397c2ee504e9dfb5ca9/samples/meta-codegen/lib/pom.xml#L116-L120

---
Input @jimschubert on Gitter

> maybe we should revert OpenAPITools/openapi-generator#2388 and provide clearer instructions on how to debug manually?

---

The `openapi-generator-cli` dependency is wrong, because the generators (normally in the `org.openapitools:openapi-generator` project) do not have to know the CLI project `org.openapitools:openapi-generator-cli`.
They can be used within the `openapi-generator-maven-plugin`, the `openapi-generator-gradle-plugin` or directly. In all those cases it is without the CLI.

As described in the [readme](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/meta-codegen/lib/README.md) of the generated plugin, the way to add the generator to the classpath of the CLI module is like this:
```
java -cp /path/to/openapi-generator-cli.jar:/path/to/your.jar org.openapitools.codegen.OpenAPIGenerator generate -g myClientCodegen -i /path/to/openapi.yaml -o ./test
```

---
The intention of the PR was to be able to create a test case: `DebugCodegenLauncher`

https://github.com/OpenAPITools/openapi-generator/blob/cbe39ba881c42a3d70a36397c2ee504e9dfb5ca9/samples/meta-codegen/lib/src/test/java/org/openapitools/codegen/debug/DebugCodegenLauncher.java#L6-L16

This is a great idea, but I think it would be much easier to create a `MyclientcodegenGeneratorTest` class that call the `MyclientcodegenGenerator` without the `openapi-generator-cli` dependency.


jmini:
After this change, it is possible to run `./bin/meta-codegen.sh` again, without having to download anything from the internet or without to have something installed in the local `.m2` cache.

samples/meta-codegen/lib/src/test/java/com/my/company/codegen/MyclientcodegenGeneratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * 1) Set a break point in MyclientcodegenGenerator.java in the postProcessOperationsWithModels() method.
+ * 2) To launch this test in Eclipse: right-click | Debug As | JUnit Test
+ *
+ */
+public class MyclientcodegenGeneratorTest {
+
+  @Test
+  public void launchCodeGenerator() {
+    // use this test to launch you code generator in the debugger.
+    // this allows you to easily set break points in MyclientcodegenGenerator.
```
jimschubert: Maybe we should add a comment with a link to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java so users can easily find sample logic?

Another option might be if we add a "helper" model in the main artifact for encapsulating options into an ADT and applying then to CodegenConfigurator, rather than having duplicate procedural logic between cli/maven/gradle. We could then use the same in these tests.

jmini: I have added a comment.

I agree that some refactoring could be useful, but I do not have time to look at this right now.

jimschubert: I'm working on a refactor of CodegenConfigurator similar to what I've suggested above.

Issue #2739, opened 2019-04-25
[BUG] [typescript-fetch] (v4.0.0) Generated models no longer produce valid to/from JSON functions - can't compile


someone1:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I've been using a nightly snapshot from early December (shortly after my last PR) to generate my typescript client. I decided to try out the latest beta and am happy with some of the changes/fixes (great job)! However, there seems to have been a change with how models are computed/generated, models that "extend" another model (e.g. allOf with a $ref) no longer seem to be generated as such. Additionally, the change I [previously put in](https://github.com/OpenAPITools/openapi-generator/commit/1e023f3b6e4d655d6e553bb359ffb128f5a20382#diff-0af874e4e56fd12fd8e0a4e97f4646f6L34) to use `allVars` vs `vars` seems to be breaking things now whereas before it was necessary.

##### openapi-generator version

I'm testing via the latest version shipped with the npm package: 4.0.0-beta3
The old version that was working as expected:: 4.0.0-SNAPSHOT (20181206.114535-69)

##### OpenAPI declaration file content or url

```yaml
swagger: '2.0'
info:
  title: 'Simple Test'

paths:
  '/get/Simple':
    get:
      summary: Retrieve List of Simple
      operationId: 'get.Simple'
      responses:
        200:
          description: OK
          schema:
            $ref: '#/definitions/ModelFull'

definitions:
  ModelPartial:
    type: object
    required:
      - id
    properties:
      id:
        type: string
      name:
        type: string
  ModelFull:
    allOf:
      - $ref: '#/definitions/ModelPartial'
      - properties:
          creationDate:
            type: string
            format: date-time
            readOnly: true
```

Generated Code (notice the missing inherited fields including the required field `id` in `ModelFullFromJSON`):
```ts
/**
 * 
 * @export
 * @interface ModelFull
 */
export interface ModelFull {
    /**
     * 
     * @type {string}
[...]

PR #1545, opened 2018-11-26
[Typescript][Fetch] Fix bugs with new typescript-fetch generator

someone1:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR
Fixes multiple issue with the update typescript-fetch generator

- fix #1535 - always initialize the `queryParameters` variable
- fix #1546 - use parameter type `Date` when a string is formatted as `date` or `date-time`
- fix #1547 - use allVars template variable for serializing/deserializing models. Pay attention to readOnly parameters.
- fix #1561 - omit time component when serializing parameters with format: date
- Update tsconfig.json to include `"moduleResolution" : "node",` option
- fix #1565 - add enums found in api operations

@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10)

PR #2899, opened 2019-05-15
[Typescript][Fetch] switch to vars from allVars

someone1:
Signed-off-by: Prateek Malhotra <someone1@gmail.com>

### PR checklist

### Description of the PR

Attempt to address #2739 - In a [previous PR,](https://github.com/OpenAPITools/openapi-generator/pull/1545) I switched to `allVars` in the typescript model templates as `vars` only included the defined properties for that model, not any properties from any (at that time) inherited classes. Since then, the inheritance/composition models changed internally and `vars` includes all properties for the model as the model composition is now flattened and no longer inherits from any base class.

This gets the code in a compilable state but there are some worries here:

- Why is `allVars` blank?
- There are "*AllOf.ts" models generated that seem erroneous - these shouldn't be generated.
- Can we get better testing in place? I suggest using a yaml spec that's more comprehensive and possibly has examples from previous issues found (I think this exists somewhere already in this project, I just can't find it again). I also suggest that for the typescript tests that `tsc` (or even `npm install`) is run on the generated code to ensure it compiles. This combination would have found these regressions during development.

CC @wing328

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache
@@ -31,7 +31,7 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
 export function {{classname}}FromJSON(json: any): {{classname}} {
     {{#hasVars}}
     return {
-        {{#allVars}}
+        {{#vars}}
         {{#isPrimitiveType}}
         {{#isDate}}
         '{{name}}': {{^required}}!exists(json, '{{baseName}}') ? undefined : {{/required}}new Date(json['{{baseName}}']),
@@ -58,7 +58,7 @@ export function {{classname}}FromJSON(json: any): {{classname}} {
         {{/isFreeFormObject}}
         {{/isContainer}}
         {{/isPrimitiveType}}
-        {{/allVars}}
+        {{/vars}}
     };
     {{/hasVars}}
     {{^hasVars}}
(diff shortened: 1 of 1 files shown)


wing328:
> Why is allVars blank?

Right. I think it's a bug that needs to be addressed.

someone1:
Regarding testing - I see that the `petstore-with-fake-endpoints-models-for-testing.yaml` is run as part of `test-fake-petstore-for-all.sh`. This file has a lot of good example but does not have the example I am opening a bug report for here.

Does anyone object to these changes?

- Add the `allOf` composition example that this PR is trying to address (e.g. composition _without_ the discriminator)
- Have some sort of post-process for compiling the typescript generated code after running `test-fake-petstore-for-all.sh` - does anyone have a suggestion on how to do this? I see that this is only referenced in the Shippable CI config - would it make sense to add a script file and attach it at the end there? Currently, the result it generates can't compile for what seems like a number of different bugs that should probably be addressed.

macjohnny:
@wing328 is the `allVars` empty thing still an issue?

wing328:
Yes, I do not have an ETA so I'll merge this PR into master for the time being.

nolanlocke:
> There are "*AllOf.ts" models generated that seem erroneous - these shouldn't be generated.

Any word on this? This doesn't seem right to me either and was curious if this was intended or another issue needs to be created (if one doesn't exist)

macjohnny:
@nolanlocke I think this is unintended and should be fixed, too.

wing328:
UPDATE: allVars has been fixed with https://github.com/OpenAPITools/openapi-generator/pull/3616

PR #2900, opened 2019-05-15
[typescript-fetch] Fix uploading files

Teyras:
The current master converts anything that is not FormData to JSON. This behavior causes operations that accept a file to send requests where the payload is an empty JSON object.

This PR adds another branch that lets Blobs pass without modification.


samples/client/petstore/typescript-fetch/builds/default/apis/PetApi.ts
```diff
@@ -23,7 +23,7 @@ import {
 } from '../models';
 
 export interface AddPetRequest {
-    body: Pet;
+    pet: Pet;
```
wing328: Look like you're using an old version. please pull the latest master and merge into your branch and then update the samples again.

wing328:
please copy the TS technical committee as stated in the PR template.

samples/client/petstore/typescript-fetch/builds/with-interfaces/runtime.ts
```diff
@@ -209,17 +209,9 @@ export function querystring(params: HTTPQuery, prefix: string = ''): string {
             }
             return `${encodeURIComponent(fullKey)}=${encodeURIComponent(String(value))}`;
         })
-        .filter(part => part.length > 0)
         .join('&');
 }
 
-export function mapValues(data: any, fn: (item: any) => any) {
```
macjohnny: this should not be removed. did you generate the samples with the most recent master merged into your branch?

Teyras: I'm completely stumped now. I tried merging upstream/master into my branch before your review. After that, I pulled upstream master into my master and merged it into my branch. This got me back to where I was before @wing328 reviewed my changes.

macjohnny: @Teyras are you sure you re-generated the samples *after* merging `origin/master`?

samples/client/petstore/typescript-fetch/builds/es6-target/runtime.ts
```diff
@@ -209,17 +209,9 @@ export function querystring(params: HTTPQuery, prefix: string = ''): string {
             }
             return `${encodeURIComponent(fullKey)}=${encodeURIComponent(String(value))}`;
         })
-        .filter(part => part.length > 0)
         .join('&');
 }
 
-export function mapValues(data: any, fn: (item: any) => any) {
```
macjohnny: please run `mvn clean` before re-generating the samples.
the `mapValues` function actually is available in your version of 
https://github.com/OpenAPITools/openapi-generator/blob/73ac4482ea97e7210aefceda22aee82885612143/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L207-L212
but it is removed in the generated samples, which indicates that the version you use for generating the samples is not the one you want to merge...

Teyras: Wow. That seems to have un-removed `mapValues`, but the `body` parameter was renamed back to `pet` again. My code seems to be haunted.

macjohnny: the `body` to `pet` renaming could be due to general changes, not necessarily related to your code.

Teyras:
@macjohnny is there anything to do before this can be merged?

macjohnny:
the ci tests detected a discrepancy between the generated samples.
https://circleci.com/gh/OpenAPITools/openapi-generator/6662#tests/containers/2

please ensure that 
- you merge the most recent master into your branch
- then run `mvn clean`
- then re-generate the samples
- make sure to commit all changed files

macjohnny:
I think the errors in https://travis-ci.org/OpenAPITools/openapi-generator/builds/534779079?utm_source=github_status&utm_medium=notification will also be gone after that

wing328:
@macjohnny is right

@Teyras let me know if you need help merging the latest master of the official repo into your branch.

macjohnny:
@Teyras the typescript-fetch tests fail because the tests need to be adapted due to changes of the body-parameter name (this is not related to your PR):

https://github.com/OpenAPITools/openapi-generator/blob/bf7ee5ec4cd93fcd075643c276281bd5e9fbc00c/samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts#L19
should be
```typescript
return api.addPet({ pet: fixture }).then(() => {
```
and
https://github.com/OpenAPITools/openapi-generator/blob/bf7ee5ec4cd93fcd075643c276281bd5e9fbc00c/samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts#L32
should be 
```typescript
return api.updatePet({ pet: result }).then(() => {
```

could you please push those changes?

macjohnny:
the tests fail due to
```
 1) PetApi without custom request options should add and delete Pet:
     ReferenceError: Blob is not defined
      at PetApi.BaseAPI.createFetchParams (/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/typescript-fetch/builds/with-npm-version/dist/runtime.js:175:81)
```

macjohnny:
@Teyras any idea how to fix it?

Teyras:
@macjohnny nope, Blob is a part of the File API so I would expect it to exist, just like FormData exists. What environment is used to run the tests?

Teyras:
@macjohnny I think the problem is that the tests run in node, which doesn't have the Blob API. We could install a polyfill (https://www.npmjs.com/package/blob), but we'd have to find a way to make it available without importing it.

macjohnny:
@Teyras I think adding https://www.npmjs.com/package/blob to the tests is a good solution. However, this also indicates that this PR introduces a breaking change in the `typescript-fetch` generator, because others might use it in node as well, missing the `Blob` API by default.

Teyras:
@macjohnny Okay... The only backwards-compatible solution I can think of is either setting `Blob` to something harmless in `runtime.ts` or making sure it is defined before using `instanceof`.

macjohnny:
@Teyras how about adding
```typescript
declare var Blob: any;
```
this will be removed by the typescript compiler and avoids throwing an error if it is not a `Blob`.
In order for your fix to work also in node environments, one would have to install e.g. https://www.npmjs.com/package/blob

if it is a `Blob` in a node environment, it will throw an error `Uncaught ReferenceError: Blob is not defined` due to the `instanceof Blob`, but it didn't work correctly in that case anyway, so I think this would be ok to publish.

Teyras:
@macjohnny I think that the argument that we should avoid BC breaks still holds - the client was fine in node environments, at least until you tried to upload a Blob to an API (which you couldn't anyway). 

If we are to support node, we should probably also find a way to upload Buffer instances...

macjohnny:
@Teyras If I am not mistaken the `declare var Blob: any;` approach should avoid a breaking change.

Teyras:
@macjohnny Won't it crash on node during the instanceof call because Blob is undefined?

macjohnny:
> @macjohnny Won't it crash on node during the instanceof call because Blob is undefined?
you are right. but this could be avoided by adding
`if (typeof Blob === 'undefined') const Blob = function(){}`
then the `instanceof Blob` should just return false

Teyras:
@macjohnny that looks like a sound solution. However, I think we should also add support for `Buffer`, since `node-fetch` supports it in request bodies. What do you think?

macjohnny:
adding support for `Buffer` sounds good

samples/client/petstore/typescript-fetch/builds/es6-target/apis/PetApi.ts
```diff
@@ -208,16 +207,12 @@ export class PetApi extends runtime.BaseAPI {
...
-            queryParameters['maxCount'] = requestParameters.maxCount;
-        }
-
         const headerParameters: runtime.HTTPHeaders = {};
 
         if (this.configuration && this.configuration.accessToken) {
             // oauth required
             if (typeof this.configuration.accessToken === 'function') {
-                headerParameters["Authorization"] = this.configuration.accessToken("petstore_auth", ["read:pets"]);
+                headerParameters["Authorization"] = this.configuration.accessToken("petstore_auth", ["write:pets", "read:pets"]);
```
macjohnny: why does this change?

wing328: It would be due to a fix in oauth scope to propagate these values probably.

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -3,6 +3,9 @@
 
 export const BASE_PATH = "{{{basePath}}}".replace(/\/+$/, "");
 
+const Blob = Blob !== undefined ? Blob : function() {};
```
macjohnny: [CI](https://travis-ci.org/OpenAPITools/openapi-generator/builds/536245806?utm_source=github_status&utm_medium=notification) fails due to `runtime.ts(17,14): error TS2448: Block-scoped variable 'Blob' used before its declaration.`
In https://www.typescriptlang.org/play/index.html you can try
```typescript
const Blob = Blob !== undefined ? Blob : function () { };

function isBl(a: any) {
    return a instanceof Blob;
}
```
which does not work.
However, this works:
```typescript
if (typeof Blob === 'undefined') {
    const Blob = function () { };
}

function isBl(a: any) {
    return a instanceof Blob;
}
```

Teyras: Hey, seeing `isBl` gave me an idea - we could make a function that checks if the argument is a Blob and also returns false when Blob is undefined. I think it would be cleaner than conditionally defining consts.

macjohnny: sounds good

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -3,6 +3,9 @@
 
 export const BASE_PATH = "{{{basePath}}}".replace(/\/+$/, "");
 
+const isBlob = (value: any) => Blob !== undefined && value instanceof Blob;
```
macjohnny:
```suggestion
const isBlob = (value: any) => typeof Blob !== 'undefined' && value instanceof Blob;
```
since Blob is not necessarily a declared variable

macjohnny: see e.g. https://stackoverflow.com/a/4725697

macjohnny:
the samples seem to have been generated with a different version. it is strange that in each commit e.g. the `body: Pet` and `pet: Pet` change back and forth. @Teyras are you using the same very command each time you generate the samples? are you sure you always run `mvn clean` before generating the samples?

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -3,6 +3,9 @@
 
 export const BASE_PATH = "{{{basePath}}}".replace(/\/+$/, "");
 
+const isBlob = (value: any) => typeof Blob !== 'undefined' && value instanceof Blob;
+const isBuffer = (value: any) => typeof Buffer !== 'undefined' && value instanceof Buffer;
```
macjohnny: maybe it is easier to only support Blob, since `Buffer` does not exist in the `dom` typings https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts

Teyras:
@macjohnny It is possible I occassionally swap `/bin/openapi3/typescript-fetch-petstore-all.sh` and `/bin/typescript-fetch-petstore-all.sh`

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
Teyras: It should be enough to install `@types/node`, which seems pretty sensible.

macjohnny: but I imagine a user wouldn't want to install `@types/node` for a web project, since the node API won't be available in the web...

Teyras: I don't think a user would care much... create-react-app for typescript installs @types/node too.

Teyras:
@macjohnny could we merge this if I add @types/node or do you think that removing Buffer support is better? Also, does the order of sample generation command I mentioned in the previous command matter?

macjohnny:
@Teyras I think removing buffer support is better for the time being, as it makes this PR 100% backwards compatible. 
Concerning the sample generation, I would use `bin/typescript-fetch-petstore-all.sh`, as this is the script that also the CI uses to check whether the samples are up to date (see https://github.com/OpenAPITools/openapi-generator/blob/f93e75b078e3e5b759a434aee8c72e38115baaa7/bin/utils/ensure-up-to-date#L42)

samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts
```diff
@@ -16,7 +16,7 @@ describe('PetApi', () => {
             });
 
             it('should add and delete Pet', () => {
-                return api.addPet({ body: fixture }).then(() => {
+                return api.addPet({ pet: fixture }).then(() => {
```
macjohnny:
```suggestion
                return api.addPet({ body: fixture }).then(() => {
```

samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts
```diff
@@ -29,7 +29,7 @@ describe('PetApi', () => {
             it('should update Pet by ID', () => {
                 return api.getPetById({ petId: fixture.id }).then((result: Pet) => {
                     result.name = 'newname';
-                    return api.updatePet({ body: result }).then(() => {
+                    return api.updatePet({ pet: result }).then(() => {
```
macjohnny:
```suggestion
                    return api.updatePet({ body: result }).then(() => {
```

macjohnny:
@wing328 can you re-start the circle-ci tests?

wing328:
All tests passed 👍

wing328:
@Teyras thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

Issue #273, opened 2018-06-10
[C++] [Restbed] Restbed Generator templates need clean up and refactoring


stkrwork:
##### Description

Restbed Generator needs to be reviewed and refactored.

##### Suggest a fix/enhancement

review current code, and refactor if necessary

muttleyxd:
I'm interested in restbed generator, what I don't like with current implementation is, that it requires to modify generated code, there's no way to inject your own handlers, as restbed does not allow to set REST method handler few times.
So if you set handler two times (logically overwriting old one) only first will be called (and that one is taken by generated code).
Basically at this moment this generated code is either useless (because we need to call restbed manually) or cant be regenerated.

My idea is to add
`std::function<int(...)> callback_;`
to generated classes, this way it would call the callback and expect status_code reply from it.
So if you regenerate auto code your code won't get overwritten.

I've got this working, any interest in PR?

stkrwork:
@muttleyxd Please open a PR, so we can take a look

f18m:
Hi @stkrwork , Hi @muttleyxd ,

sorry to go over this "old" issue... I just tried using the restbed server-side generator with both the OpenAPI generator CLI v4.3.1 and v5.0.0-beta2 and I stumped over the same problem described by @muttleyxd : the automatically-generated restbed::Service-derived class is publishing in its constructor the restbed::Resource... 
That means there's no chance to set the GET handler or whatever other handler exists on the restbed::Resource... 

how am I supposed to use the generated C++ code?
Do you have any hint? Or any working example?

@muttleyxd : did you manage to open a  PR ? I could help on that perhaps...

Thanks

muttleyxd:
Well, I did open a PR, where you can set the callback through the method

https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/cpp-restbed/api/PetApi.h#L65

but now I see it was a bad idea. A good refactor would be to do it like in Pistache generator (PetApi would be class with virtual methods and user would have to override these in their handler code).

f18m:
> Well, I did open a PR, where you can set the callback through the method
> 
> https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/cpp-restbed/api/PetApi.h#L65
> 
> but now I see it was a bad idea. A good refactor would be to do it like in Pistache generator (PetApi would be class with virtual methods and user would have to override these in their handler code).

[...]

PR #2911, opened 2019-05-16
[C++][Restbed] Add handler callback methods

muttleyxd:
### PR checklist

Added second commit with updated samples (not 100% sure what I should do with them), are there any tests for these? I didn't find any.
@ravinikam @stkrwork  @fvarose  @etherealjoy  @MartinDelille 

### Description of the PR

Related issue: #273 
What I don't like about current generator: you need to modify generated code, there isn't a way for injecting your own handler. librestbed does not allow to replace already existing handler, which means you either modify generated code or don't use it at all. I modified it a little, to allow setting a callback, which responds with std::pair<int (HTTP status code), std::string (response)>

I'm not an expert on Mustache, so I would be really grateful for some tips where I could optimize/refactor it.


etherealjoy:
A generic comment here, Callback is also an OpenApi 3 keyword, won't this be confusing?

modules/openapi-generator/src/main/resources/cpp-restbed-server/api-header.mustache
```diff
@@ -49,6 +50,31 @@ public:
     {{#vendorExtensions.x-codegen-otherMethods}}
     void {{httpMethod}}_method_handler(const std::shared_ptr<restbed::Session> session);
     {{/vendorExtensions.x-codegen-otherMethods}}
+
+	void set_callback_{{httpMethod}}(
```
etherealjoy: If there are multiple endpoints with the same tag in the API returning and/or accepting similar parameters this will cause  redefinition of the methods

muttleyxd:
> A generic comment here, Callback is also an OpenApi 3 keyword, won't this be confusing?

Definitely a wrong name then (I didn't know this feature before). Should I change 'callback' to 'handler'?

etherealjoy:
I think handler would be a better name

modules/openapi-generator/src/main/resources/cpp-restbed-server/api-header.mustache
muttleyxd: Could you provide me an example when this could happen?
I tried something like this:

```
paths:
  '/method/{param1}':
    parameters:
      - name: param1
        in: path
        type: string
        required: true
        description: Some string parameter
    post:
      operationId: methodParam1Post
      tags:
        - sometag
    (...)
  '/method/{param2}':
    parameters:
      - name: param2
        in: path
        type: integer
        required: true
        description: Some other parameter
    post:
      operationId: methodParam2Post
      tags:
        - sometag
    (...)
```

but I got this:
```
class  SometagApiMethodParam1Resource: public restbed::Resource
{
public:
	SometagApiMethodParam1Resource();
    virtual ~SometagApiMethodParam1Resource();
    void POST_method_handler(const std::shared_ptr<restbed::Session> session);

	void set_handler_POST(
		std::function<std::pair<int, std::string>(
			std::string const &
		)> handler
	);

private:
	std::function<std::pair<int, std::string>(
		std::string const &
	)> handler_POST_;
};

class  SometagApiMethodParam2Resource: public restbed::Resource
{
public:
	SometagApiMethodParam2Resource();
    virtual ~SometagApiMethodParam2Resource();
    void POST_method_handler(const std::shared_ptr<restbed::Session> session);

	void set_handler_POST(
		std::function<std::pair<int, std::string>(
			int32_t const &
		)> handler
	);


private:
	std::function<std::pair<int, std::string>(
		int32_t const &
	)> handler_POST_;

};
```

muttleyxd:
> I think handler would be a better name

Done

modules/openapi-generator/src/main/resources/cpp-restbed-server/api-header.mustache
etherealjoy: I see you are right, I was thinking in the terms of the other framework. It looks fine then.

wing328:
@muttleyxd thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

engpedrorafael:
I appreciate and agree that this is a very important feature, though I don't see how to make use of the **set_handler_** GET/POST/DELETE without changing the generated code.
The methods **set_handler_*** are generated as member functions of the ***Resource** classes which can only be called through an object instantiated from the respective class.
From the generated code, we can see that at the constructor of the API class is where the ***Resources** are created as local objects and passed to the Base Class using the method publish:

```
PetApi::PetApi() {
	std::shared_ptr<PetApiPetResource> spPetApiPetResource = std::make_shared<PetApiPetResource>();
	this->publish(spPetApiPetResource);	
	...
```
I don't see a way of accessing the ***Resource** objects pubished after the creation of the object, in order to call the respective **set_handler_*** to set my custom handlers, and this could not have been done before since we cannot call it without an object.

Can you please explain, maybe with a small example, how to make use of this in order to set custom handlers

muttleyxd:
Yeah, I think I'd probably need to change generated constructor or add another method for setting handlers.
Maybe virtual dispatch would be good here, what do you think about it? So one would have to do
```c++
class Derived : public PetApi
{
public:
  <type> PetApiPetResource_GET(...) override;
};
```

For now I usually create new PetApi class and and use the resources myself, something like that
```c++
class MyPetApi
{
public:
  MyPetApi() {
    std::shared_ptr<PetApiPetResource> spPetApiPetResource = std::make_shared<PetApiPetResource>();
    spPetApiPetResource.set_handler_PUT([&](auto const& body) { return handlePetPut(body); })
	this->publish(spPetApiPetResource);
  }

private:
  std::pair<int, std::string> handlePetPut(std::string const& body) {
    // Do something here
    return {200, "Success"};
  }
};

PR #2921, opened 2019-05-17
To fix conflict params name 'url'

vellengs:
Sometime we have api params named url that will conflict with imported module url named url.

### PR checklist

### Description of the PR

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -3,7 +3,7 @@
 /// <reference path="{{apiRelativeToRoot}}custom.d.ts" />
 {{>licenseInfo}}
 
-import * as url from 'url';
+import * as $url from 'url';
 import globalAxios, { AxiosPromise, AxiosInstance } from 'axios';
 import { Configuration } from '{{apiRelativeToRoot}}configuration';
 import { BASE_PATH, COLLECTION_FORMATS, RequestArgs, BaseAPI, RequiredError } from '{{apiRelativeToRoot}}base';
@@ -44,7 +44,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
     {{/allParams}}
             const localVarPath = `{{{path}}}`{{#pathParams}}
                 .replace(`{${"{{baseName}}"}}`, encodeURIComponent(String({{paramName}}))){{/pathParams}};
-            const localVarUrlObj = url.parse(localVarPath, true);
+            const localVarUrlObj = $url.parse(localVarPath, true);
             let baseOptions;
             if (configuration) {
                 baseOptions = configuration.baseOptions;
@@ -188,7 +188,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
     {{/bodyParam}}
 
             return {
-                url: url.format(localVarUrlObj),
+                url: $url.format(localVarUrlObj),
                 options: localVarRequestOptions,
             };
         },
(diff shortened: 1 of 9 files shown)


wing328:
@vellengs thanks for the PR. Can you please also run `./bin/typescript-axios-petstore-all.sh` to run to update the samples as well?

wing328:

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -3,7 +3,7 @@
 /// <reference path="{{apiRelativeToRoot}}custom.d.ts" />
 {{>licenseInfo}}
 
-import * as url from 'url';
+import * as $url from 'url';
```
macjohnny: instead of renaming the parameter, I think it is better to add `url` to the reserved keywords. See e.g. https://github.com/OpenAPITools/openapi-generator/blob/1b1c13db0fa923a059b234da1d92dd98c72d9aaa/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java#L262-L263

wing328: @macjohnny yup, that's one way to do it.

My little suggestion is to prefix it (like what's done in the PR) in this case as "url" may be a popular choice for naming variables/classes.

Reserved keywords are usually language specified such as return in PHP that can must be prefixed/suffixed.

macjohnny: but someone might also use `url$` as a parameter name. I think it is a good idea to add all global symbols to the reserved keywords to avoid any naming conflicts.

wing328: I think the dollar sign and other symbols will be removed in toVarName or toParamName as part of the sanitization.

what about using the "localVar" prefix approach?

localVar may not sound right in this case so what about "globalImportUrl"?

vellengs: "globalImportUrl" seems fine.

wing328: @vellengs please go ahead with this approach to see if it works for you (I think it should).

vellengs: @wing328  can you build the shell 
./bin/typescript-axios-petstore-all.sh.
I have promblem with Marven.

wing328: Can you use maven wrapper instead?
```
./mvnw clean install
```

vellengs: @naelrashdeen I changed two part of  'api.mustache' and 'apiInner.mustache'
why the build samples file still have  
```
import * as url from 'url';
```

wing328:
I've updated the samples. Let's see how the CI tests go.

wing328:
@vellengs thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

PierreREN:
Is 'url' mentioned above [https://www.npmjs.com/package/url](url) as there seems no other url package in npm library? And if so, when building my project using "npm run build" which actually refers to "vue-cli-service build", such error occurs.
`Could not find a declaration file for module 'url'. 'XXX/node_modules/url/url.js' implicitly has an 'any' type. Try `npm install @types/url` if it exists or add a new declaration (.d.ts) file containing `declare module 'url';``
How do you solve such a problem? @wing328

PR #2933, opened 2019-05-19
[C#]Remove null checks for C# value types

etherealjoy:
### PR checklist

### Description of the PR

Remove null checks on C# value types which throws a lot of warnings.

Null checks on enums which is a value type are already avoided. Combining simple types and enum as value types and avoiding null checks.
See here.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/value-types

@wing328 @mandrean @jimschubert


etherealjoy:
I will regenerate code if it looks OK

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -454,6 +461,19 @@ private void postProcessEnumRefs(final Map<String, Object> models) {
                         var.isPrimitiveType = true;
                     }
                 }
+                for (CodegenProperty var : model.vars) {
```
etherealjoy: Question:
why is this line (some lines above) not covering all `enums`
```
for (CodegenProperty var : model.allVars) 
```

wing328: It should. That could be a bug with allVars.

I'll look into it in the coming week.

etherealjoy:
I think I need to add extra condition if it is a `valueType` but not `nullable`, however I see parsing of the nullable field is not done.

wing328:
You mean something like `{{#isNullable}} ... {{/isNulalble}}`?

etherealjoy:
Exactly. So if the value is nullable it will not be e.g. `int` but `int?`. But in this case if the `dataType` is used for postprocess it is not anymore a `valueType`
I will try out some specs with `nullable` fields.

wing328:
I think nullable is supported in the C# generators already: https://github.com/OpenAPITools/openapi-generator/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+nullable+csharp

 but you will need to use `x-nullable` in OAS v2 instead which does not support nullable natively.

etherealjoy:
Thank you. I see that the simple type is already changed to nullable type. Then this will be covered.
I will make the test with this.

etherealjoy:
Ok, I verified it. the nullable types are having null checks. only value Types are skipped.

wing328:
Can you please update the C# netcore sample as well?
```

	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/docs/EnumTest.md
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/ApiResponse.cs
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/Cat.cs
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/CatAllOf.cs
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/Category.cs
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/EnumTest.cs
	modified:   samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Model/FormatTest.cs
```

PR #2937, opened 2019-05-19
[Apex] Set namedCredential file name from the property

### LINKED ISSUE #2904: [BUG] Apex named credential property does not work
##### Description

When generating an apex client, the namedCredential property does not override the name of the named credential. In code, it looks like this property is ignored, and the 

4.0.0

```
openapi-generator generate -i https://directplus.documentation.dnb.com/assets/APIs/searchCriteria.json -g apex --additional-properties namedCredential=DnB_Direct
```

##### Suggest a fix

It looks like the name for the file is pulled from `openAPI.getInfo().getTitle()` rather than reading from the namedCredential property. Should be an easy fix to read the property instead.

https://github.com/OpenAPITools/openapi-generator/blob/046db19a8590f3557ae1e04be75792bbeb39cb69/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ApexClientCodegen.java#L164-L171

loganm:
### PR checklist

### Description of the PR

@asnelling It seems like you're the person to review this one.

This fixes #2904

The namedCredential property is intended to override the namedCredentials name, but it wasn't being used. Instead, the title from the openapi spec file was used in all cases, and the namedCredential property wasn't used anywhere.

If a namedCredential property is provided, it will be used instead of the default name. For backwards compatibility, the spec title is still used for the named credential by default.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ApexClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ApexClientCodegen.java
@@ -21,6 +21,7 @@ import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.info.Info;
 import io.swagger.v3.oas.models.media.ArraySchema;
 import io.swagger.v3.oas.models.media.Schema;
+import org.apache.commons.lang3.StringUtils;
 import org.openapitools.codegen.CliOption;
 import org.openapitools.codegen.CodegenProperty;
 import org.openapitools.codegen.SupportingFile;
@@ -46,7 +47,7 @@ public class ApexClientCodegen extends AbstractApexCodegen {
     private String classPrefix = "OAS";
     private String apiVersion = "42.0";
     private String buildMethod = "sfdx";
-    private String namedCredential = classPrefix;
+    private String namedCredential;
     private String srcPath = "force-app/main/default/";
     private String sfdxConfigPath = "config/";
     private HashMap<String, Object> primitiveDefaults = new HashMap<String, Object>();
@@ -161,8 +162,10 @@ public class ApexClientCodegen extends AbstractApexCodegen {
 
     @Override
     public void preprocessOpenAPI(OpenAPI openAPI) {
-        Info info = openAPI.getInfo();
-        String calloutLabel = info.getTitle();
+        String calloutLabel = openAPI.getInfo().getTitle();
+        if (StringUtils.isNotBlank(namedCredential)) {
+            calloutLabel = namedCredential;
+        }
         additionalProperties.put("calloutLabel", calloutLabel);
         String sanitized = sanitizeName(calloutLabel);
         additionalProperties.put("calloutName", sanitized);


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ApexClientCodegen.java
```diff
@@ -161,8 +161,10 @@ public void processOpts() {
 
     @Override
     public void preprocessOpenAPI(OpenAPI openAPI) {
-        Info info = openAPI.getInfo();
-        String calloutLabel = info.getTitle();
+        String calloutLabel = openAPI.getInfo().getTitle();
+        if (namedCredential != null && !namedCredential.trim().isEmpty()) {
```
wing328: What about using `StringUtils.isNotEmpty` instead?

loganm: @wing328 Very good idea. I'm not a Java developer by trade, so I'm a bit gun shy with libraries.

wing328:
cc @muenzpraeger

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ApexClientCodegen.java
```diff
@@ -46,7 +47,7 @@
     private String classPrefix = "OAS";
     private String apiVersion = "42.0";
     private String buildMethod = "sfdx";
-    private String namedCredential = classPrefix;
+    private String namedCredential;
```
wing328: @loganm What about keeping the default as `classPrefix`?

loganm: The current behaviour of the generator is to use the spec title for the namedCredential. For example, I'm using a Dun and Bradstreet spec with the title 'Search_for_a_Company_or_Contact'.

The default classPrefix is 'OAS' which doesn't mean much. The spec title is more informative about purpose of the named credential.

Now that the namedCredential var is being referenced, leaving the default value as classPrefix here would also change the default behaviour of the generator. It could be a breaking change for anyone using it presently.


Do you think it's possible to get at the spec title as the default here instead of the class prefix?

wing328: > Now that the namedCredential var is being referenced, leaving the default value as classPrefix here would also change the default behavior of the generator. It could be a breaking change for anyone using it presently.

Yes, we're on the same page to make this change backward-compatible.

Let me review again as I thought the default value is no longer the same.

wing328: I just updated the petstore samples and I can see the issue you've fixed in this PR 👍

wing328:
I've updated the Apex samples and add apex-petstore.sh to ensure-up-to-date script via 5ad318bfae

wing328:
@loganm thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

Issue #2706, opened 2019-04-20
System properties do not make it down to dependencies


jdomizio:
#### Bug Report Checklist

- [n/a] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Option to set system property `-Dio.swagger.v3.parser.util.RemoteUrl.trustAll=true` in `openapi-generator-cli` does not actually set a system property that can be read by `System.getProperty()` call in `io.swagger.v3.parser.util.RemoteUrl`

##### openapi-generator version

v4.0.0-beta3

##### OpenAPI declaration file content or url

n/a

##### Command line used for generation

```sh
docker run --rm -v ${PWD}:/local \
    openapitools/openapi-generator-cli:v4.0.0-beta3 generate \
    -Dio.swagger.v3.parser.util.RemoteUrl.trustAll=true \
    -i https://xxx.dev.yyy.local/api/v1/swagger.json \
    -g csharp-netcore \
    -o /local/xxx
```

Unfortunately the url I am testing against is on a closed network, and I don't have an example that is internet-accessible.

The actual output I am seeing is here
(https://gist.github.com/jdomizio/a21471b8a1b65b8b5904aad2e9276b44)

the expected output is the normal output for code generation.

##### Steps to reproduce

Run the above command against a swagger document hosted under https with a self-signed certificate.  After pulling the v4.0.0-beta3 tag and modifying `org.openapitools.codegen.config.CodegenConfigurator.setSystemProperties` 

to

```java
    private void setSystemProperties() {
        for (Map.Entry<String, String> entry : systemProperties.entrySet()) {
            System.setProperty(entry.getKey(), entry.getValue());
            GeneratorProperties.setProperty(entry.getKey(), entry.getValue());
        }
    }
```

the expected system property gets set and I can successfully run the cli tool against self-signed endpoints again.

##### Related issues/PRs

(https://github.com/OpenAPITools/openapi-generator/issues/845#issuecomment-415316114)

##### Suggest a fix

[...]

PR #2946, opened 2019-05-20
[core] GeneratorSettings, WorkflowSettings, and cleanup in CodegenConfigurator

jimschubert:
### PR checklist

### Description of the PR

This is general enhancements as part of the [Separation of Concerns](https://github.com/OpenAPITools/openapi-generator/projects/5) project, and is work toward moving things to a Core package (#845). In doing so, my goal is to separate conceptual responsibilities of our internal implementations.

This PR leaves `CodegenConfigurator` as an "entrypoint" to configuring internals. There's still some work to be done to simplify that interface so we may remove redundant code from CLI, Maven Plugin, Gradle Plugin, and openapi-generator-online (which duplicates much of `CodegenConfigurator` in `Generator.java`).

A potentially minor breaking change, I've removed all getters from `CodegenConfigurator`. The caller already knows the properties being set on this type, and any evaluation should be done after calling `toClientOptInput` or the new `toContext`.

This PR introduces two new types: `GeneratorSettings` and `WorkflowSettings`. These are intentionally immutable, which will aid in testing once they are applied to a generator instance (generator authors can evaluate input settings against modified settings to validate any transformations done in the generator). These types are separate so we can clearly differentiate between those settings defining the _what_ (generator outputs) versus the _how_ (the workflow taken to generate outputs). 

As part of creating these new Settings types and moving to make `CodegenConfigurator` more SOLID, I've also introduced a `DynamicSettings` type which handles the deserialization of those settings which were previously considered `CodegenConfigurator`.  Now, `CodegenConfigurator` is simply orchestrating how we configure everything in preparation for generation.  `DynamicSettings` doesn't live in the Core module because it relies on Jackson JSON, and there's technically no need for this to live in a user-consumable module (users can just set additional properties directly); loading the config from file is application-specific.

While focusing on this settings cleanup, I wanted to rename `GeneratorProperties` to avoid confusion with `GeneratorSettings`.  The intention is that `GeneratorSettings` are those settings which are applied to a generator (an instance of `CodegenConfig`). `GeneratorProperties`, on the other hand, is a way to access System properties and modify them on a thread local to avoid race conditions during CI or other workflows… so I've opted to rename this to `GlobalSettings`. `GlobalSettings` will therefore be anything that can be considered outside the responsibility of a generator or its workflow (e.g. `debugModels` could be set as a System Property and used by either). We recently had a report of `-D` options not making it down to swagger-parser (see #2706), and this is because the CLI has a custom argument of `-D` which can easily be confused as the Java property `-D`. In this PR, I've added a deprecation warning with instructions to stop using the `-D` argument to `generate` and move this option to System properties. If we find that something like `-DdebugModels` or other supported "system properties" currently cause conflicts elsewhere, we may want to add a new `global-settings` option. I didn't do this in this PR because it seemed like a little too much for a single change.


jimschubert:
cc @OpenAPITools/generator-core-team I've filed this to hit master since the changes shouldn't be "public api" (CodegenConfigurator getters and `GeneratorProperties` rename). Everything in this PR should have a pretty clear workaround.

jimschubert:
@OpenAPITools/openapi-generator-collaborators 

I'm looking for feedback about whether this should target master or the 4.1.0 branch.

Working through all usages of `-D` in scripts, I've found that they were being used as a shortcut for `--additional-properties`. I don't think this was ever an intended use case. I still think we'd need to get rid of the `-D` argument to the generate command because it's easily confused with the `-D` argument to java for setting system properties.  I'd considered adding a `-P` short name for `--additional-properties`, which I think would more clearly indicate that these are properties rather than defined system properties.

Any thoughts one way or another?

wing328:
> I'm looking for feedback about whether this should target master or the 4.1.0 branch.

My take is 4.1.0 branch as it's a breaking change with fallbacks

wing328:
>  I'd considered adding a -P short name for --additional-properties

Good idea for a short name. What about `-p` (lower case) instead as all the other short names are lower case too?

jimschubert:
I've updated the target to the 4.1.x branch and added the `-p` option as a shortcut for `--additional-properties`. I've also regenerated samples.

jimschubert:
I've submitted a final commit, there were a couple of objc scripts which passed on my machine but failed on Shippable with the same syntax (? This is probably an IFS expansion issue).

Builds are passing now. Sorry for the tons of file changes.

wing328:
## Upgrade Note

`-D` to pass generator options has been deprecated. Please use `--additional-properties` instead, for example, from:
```
ags="$ags -DprojectName=Petstore --model-package Samples.Petstore"
```
to
```
ags="$ags --additional-properties projectName=Petstore --model-package Samples.Petstore"
```

PR #2948, opened 2019-05-20
Added support for patterns on model properties

ingvirafn:
ASPNET Core uses the attribute RegularExpression to validate the model state. This is supported in OpenAPI spec with the 'pattern' property.
This PR adds the reg-ex attribute to the model codegen.

### DIFF
--- a/modules/openapi-generator/src/main/resources/aspnetcore/2.1/model.mustache
+++ b/modules/openapi-generator/src/main/resources/aspnetcore/2.1/model.mustache
@@ -26,6 +26,9 @@ namespace {{modelPackage}}
         {{#required}}
         [Required]
         {{/required}}
+        {{#pattern}}
+        [RegularExpression("{{{pattern}}}")]
+        {{/pattern}}
         [DataMember(Name="{{baseName}}", EmitDefaultValue={{#isNullable}}true{{/isNullable}}{{^isNullable}}false{{/isNullable}})]
         {{#isEnum}}
         public {{{datatypeWithEnum}}}{{#isEnum}}{{^isContainer}}?{{/isContainer}}{{/isEnum}} {{name}} { get; set; }


jimschubert:
Thanks for the contribution!

I've tested with a petstore simple spec and a pattern added to the `Pet` model's name:

```
"pattern": "/[^\\S\\r\\n]/"
```

Generating code on this branch creates the expected attribute:

```
/// <summary>
/// Gets or Sets Name
/// </summary>
[Required]
[RegularExpression("/[^\\S\\r\\n]/")]
[DataMember(Name="name", EmitDefaultValue=false)]
public string Name { get; set; }
```

Putting an invalid name via swagger ui:

```
{
  "id": 0,
  "category": {
    "id": 0,
    "name": "string"
  },
  "name": "this is a name",
  "photoUrls": [
    "string"
  ],
  "tags": [
    {
      "id": 0,
      "name": "string"
    }
  ],
  "status": "available"
}
```

It responds with the error object:
```
{
  "errors": {
    "name": [
      "The field Name must match the regular expression '/[^\\S\\r\\n]/'."
    ]
  },
  "title": "One or more validation errors occurred.",
  "status": 400,
  "traceId": "0HLN0HUH2OUIH:00000005"
}
```

I do wonder if here are any concerns with this response object, since it exposes the regex value to the client. Do you know if it's possible to make the error message generic?

wing328:
> I do wonder if here are any concerns with this response object, since it exposes the regex value to the client. Do you know if it's possible to make the error message generic?

To me, it's ok as the error message should be informative so not a bad idea for the client to know why the validation fails.

ingvirafn:
I agree with William, if somebody has access to the service he should be
able to see how to consume it. And this is the same information in the
specification so I can't see why this should be a secret.


It's possible specify custom error message if that is needed:

[RegularExpression(@"^[a-zA-Z''-'\s]{1,40}$", ErrorMessage = "Characters
are not allowed.")]


See here for details:
https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.regularexpressionattribute?view=netframework-4.8

wing328:
Let's go with this implementation and see if we've more feedback from the users on this.

wing328:
@ingvirafn thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

PR #2963, opened 2019-05-21
Resolves #2962 - Add properties config to Maven parameters

### LINKED ISSUE #2962: [REQ] Expose all top-level config options as properties

Some of the options in the CodeGenMojo are exposed such that you can pass the property on the Maven command-line using `-Dopenapi.generator.maven.plugin.<property>`, but not all. I propose that ALL of those properties should be exposed and that the documentation should be updated to reflect the property name as well.

Modify the CodeGenMojo to include property exposure in all @Parameter annotations

None that I can think of

## Additional context

I can submit a PR for this if it is helpful.

InfoSec812:
### Description of the PR

Exposes the CodeGenMojo configuration settings as java properties to that they can be modified using `-D` arguments

Resolves #2962


modules/openapi-generator-maven-plugin/README.md
```diff
@@ -101,7 +106,7 @@ Specifying a custom generator is a bit different. It doesn't support the classpa
             </goals>
             <configuration>
                 <inputSpec>${project.basedir}/src/main/resources/yaml/yamlfilename.yaml</inputSpec>
-                <!-- language file, like e.g. JavaJaxRSCodegen -->
+                <!- language file, like e.g. JavaJaxRSCodegen -->
```
wing328: @InfoSec812 the dash seems to be accidentally removed?

InfoSec812: Whoops! Good catch. Must have accidentally done that while doing a find/replace to update the docs. Ill fix it now.

wing328:
@InfoSec812 thanks for the PR. When you've time, I wonder if you can resolve the merge conflicts and then I'll merge this PR into master.

InfoSec812:
> @InfoSec812 thanks for the PR. When you've time, I wonder if you can resolve the merge conflicts and then I'll merge this PR into master.

Resolved and I included an exposed property for the `engine` parameter

rvowles:
There appears to be a bug in this PR. I just build master and am using it, and its exploding with this field as it does not match its named parameter. Maven is trying to give it a default value by setting the field "strictSpec" and the field is called "strictSpecBehaviour".  @InfoSec812  @wing328

wing328:
@rvowles please open an issue to track it and we'll look into it.

PR #2973, opened 2019-05-22
[C++] [Qt5] Set default base path and change constructor to const ref

etherealjoy:
### PR checklist

### Description of the PR

Changes
- Change constructor signature to const reference for `QString`
- Add initializer for `basepath`, `host`
- moved common code to super

`cc:`
@ravinikam @stkrwork @MartinDelille


samples/client/petstore/cpp-qt5/client/OAIPetApi.cpp
```diff
@@ -91,7 +91,7 @@ OAIPetApi::deletePet(const qint64& pet_id, const QString& api_key) {
     OAIHttpRequestInput input(fullPath, "DELETE");
 
     if (api_key != nullptr) {
-        input.headers.insert("api_key", "api_key");
+        input.headers.insert("api_key", api_key);
```
stkrwork: this change is not part of this PR. Were the samples not regenerated in the PR that performed this change?

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-header.mustache
```diff
@@ -18,7 +18,7 @@ class {{classname}}: public QObject {
 
 public:
     {{classname}}();
-    {{classname}}(QString host, QString basePath);
+    {{classname}}(const QString& host, const QString& basePath);
```
stkrwork:
```suggestion
    {{classname}}(const QString& host, const QString& basePath = "{{{basePathWithoutHost}}}");
```

samples/client/petstore/cpp-qt5/client/OAIPetApi.cpp
etherealjoy: No it was not updated. But this is correct.

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-header.mustache
etherealjoy: Yes good idea.

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-body.mustache
```diff
@@ -9,15 +9,15 @@
 namespace {{this}} {
 {{/cppNamespaceDeclarations}}
 
-{{classname}}::{{classname}}() {
+{{classname}}::{{classname}}() : basePath("{{{basePathWithoutHost}}}"){
```
stkrwork: Whats the default value of host?

etherealjoy: Yes, I need to add this.
I thought of making these members private but it might break some things.

stkrwork: Should the user really be able to change the base path and host after initializing the api?

stkrwork: So private members would be better i think

stkrwork: ok. then lets add a setter

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-header.mustache
etherealjoy: I had to revert this when I provide default values, because then it looks like the default constructor with no parameters.

modules/openapi-generator/src/main/resources/cpp-pistache-server/main-api-server.mustache
```diff
@@ -53,7 +53,7 @@ int main() {
     std::vector<int> sigs{SIGQUIT, SIGINT, SIGTERM, SIGHUP};
     setUpUnixSignals(sigs);
 #endif
-    Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port({{serverPort}}));
+    Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port({{#serverPort}}{{serverPort}}{{/serverPort}}{{^serverPort}}8080{{/serverPort}}));
```
stkrwork: if you change stuff for pistache as well here, then you have to regenerate the samples for pistache as well

etherealjoy: I did, there is no change. 
The method produces default 8080 `serverPort` value, now it does not so that we don't need to append port for clients.
Check the removal of code from the pistache java file.

etherealjoy: I also regenerated Qt5 server, which was not updated before.

samples/server/petstore/cpp-pistache/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-4.0.0-SNAPSHOT
\ No newline at end of file
+4.0.1-SNAPSHOT
```
etherealjoy: for pistahce only this changed

wing328:
@etherealjoy thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

PR #2981, opened 2019-05-23
[Kotlin client] Improve JSON parser

### LINKED ISSUE #1271: [BUG][KOTLIN] No JsonAdapter for LocalDateTime in Moshi serializer
##### Description

`date-time` fields can't be parsed from a generated Kotlin client with default configuration. It fails with this exception:
`java.lang.IllegalArgumentException: Platform class java.time.LocalDateTime annotated [] requires explicit JsonAdapter to be registered`

The reason for it is that in [Serializer.kt.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-client/infrastructure/Serializer.kt.mustache) Moshi only gets a JsonAdapter for the `java.util.Date` type but not for `java.time.LocalDateTime` (or `org.threeten.bp.LocalDateTime` when ThreeTen is used).

v3.3.1

`generate -g kotlin -i https://petstore.swagger.io/v2/swagger.yaml -o swaggerClientTest`

##### Steps to reproduce

It can be reproduced with the petstore swagger spec.

1. Generate Kotlin client with the command above.
2. Write a simple test like this:
```kotlin
@Test
fun getOrderById() {
    val order = StoreApi().getOrderById(1)
    assertNotNull(order)
}
```
3. It will fail with the respective exception.

A workaround is to implement a `JsonAdapter` for the respective type and add it to Moshi in the generated `Serializer` class. Maybe this can be integrated in the generator.

### LINKED ISSUE #1647: [BUG][KOTLIN] Wrong import for fields with "format: date"
##### Description

Models that have a field with `type: string` and `format: date` are generated with a wrong `import` when using threetenbp as the date library.

3.3.4

Every OpenAPI spec with at least one field like this will trigger the bug:

```yaml
created:
    type: string
    format: date
```

Gradle plugin was used with the following config:

```groovy
openApiGenerate {
    generatorName = "kotlin"
    inputSpec = "$rootDir/openApiClient/OpenAPI.yml".toString()
    outputDir = "$rootDir/openApiClient".toString()
    configOptions = [
            dateLibrary   : "threetenbp",
            collectionType: "list"
    ]
}
```

##### Steps to reproduce

1. Use OpenAPI spec with at least one model with a `format: date` field
1. Generate the client
1. There will be `Unresolved reference: org` compiler errors in the model files

### LINKED ISSUE #2063: [bug][Kotlin] Default typeMapping for 'date' should be LocalDate
#### Bug Report Checklist

##### Description

When a spec specifies a property as : 

```
"type": "string",
"format": "date",
```

the Kotlin generator generates the corresponding field as a `LocalDateTime` and later, the parsing fails when Jackson is used (Gson too) saying it cannot find a converter for this type.

As I understand the spec, `"format":"date"` should result in a `LocalDate` and `"format":"date-time"` in a `LocalDateTime`

I tried version 3.3.4, but I read the code of [master](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java) and the problem is still present.

##### Steps to reproduce

Use as an input any spec that has a property configured as : 

```
"type": "string",
"format": "date",
```

I think there is only one line to change at https://github.com/OpenAPITools/openapi-generator/blob/0b66fa5c8214e2907a326fa8684cb2adcfa438a6/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java#L175

it should be

```
 typeMapping.put("Date", "java.time.LocalDate");
```

I can provide a PR if you think it is really a bug.

Cheers.

karismann:
### PR checklist

@jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @Zomzog (2019/04)

### Description of the PR

- add UUID, LocalDateTime, LocalDate adapter for JSON serialize/deserialize (fix #1271)
- map date to LocalDate type (fix #2063)
- fix import in model for threetenbp date (fix #1647)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
@@ -166,14 +166,14 @@ public abstract class AbstractKotlinCodegen extends DefaultCodegen implements Co
         typeMapping.put("ByteArray", "kotlin.ByteArray");
         typeMapping.put("number", "java.math.BigDecimal");
         typeMapping.put("date-time", "java.time.LocalDateTime");
-        typeMapping.put("date", "java.time.LocalDateTime");
+        typeMapping.put("date", "java.time.LocalDate");
         typeMapping.put("file", "java.io.File");
         typeMapping.put("array", "kotlin.Array");
         typeMapping.put("list", "kotlin.collections.List");
         typeMapping.put("map", "kotlin.collections.Map");
         typeMapping.put("object", "kotlin.Any");
         typeMapping.put("binary", "kotlin.Array<kotlin.Byte>");
-        typeMapping.put("Date", "java.time.LocalDateTime");
+        typeMapping.put("Date", "java.time.LocalDate");
         typeMapping.put("DateTime", "java.time.LocalDateTime");
 
         instantiationTypes.put("array", "kotlin.arrayOf");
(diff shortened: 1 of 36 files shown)


jimschubert:
@karismann this looks good. Is it ready to merge?

karismann:
@jimschubert  yes, thanks for the review!

PR #2988, opened 2019-05-24
improve Kotlin code quality

kroegerama:
Some minor improvements to the kotlin-client template. Use `val` instead of `var` where possible. Use `.apply{}` combined with the builder pattern. Use `map[key] = value` instead of `map.put(key, value)`.

@jimschubert @dr4ke616 @karismann @Zomzog

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/infrastructure/ApiClient.kt.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/infrastructure/ApiClient.kt.mustache
@@ -24,7 +24,7 @@ open class ApiClient(val baseUrl: String) {
         protected const val XmlMediaType = "application/xml"
 
         @JvmStatic
-        val client by lazy {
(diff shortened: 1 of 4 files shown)


jimschubert:
@kroegerama  This looks good, thanks for the improvement!

Can you regenerate the kotlin samples to allow CircleCI to pass?

Those scripts are (and require that you've done `mvn install` beforehand):

```
bin/kotlin-client-petstore.sh
bin/kotlin-client-string.sh
bin/kotlin-client-threetenbp.sh
```

wing328:
@kroegerama thanks for the PR. For upcoming PRs, I would suggest creating a new branch as per Git best practice.

About Kotlin code format, you can also use post-processing file hook to format the Kotlin file using your favorite code formatter. Let me know if you need more information on that.

wing328:
@kroegerama thanks again for your contribution, which has been included in the v4.0.1 release (https://twitter.com/oas_generator/status/1135534738658062336)

PR #3020, opened 2019-05-28
[TS][Inversify] Rename map to GlobalImportOperators

wing328:
### PR checklist

### Description of the PR

Rename map to GlobalImportOperators so that `map` doesn't need to be a reserved word.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
@@ -8,7 +8,7 @@ import { Observable } from "rxjs/Observable";
 import { Observable } from "rxjs";
 {{/useRxJS6}}
 
-import { map } from "rxjs/operators";
+import { GlobalImportOperators } from "rxjs/operators";
 import IHttpClient from "../IHttpClient";
 import { inject, injectable } from "inversify";
 import { IAPIConfiguration } from "../IAPIConfiguration";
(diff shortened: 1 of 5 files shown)


wing328:
> this does not seem correct to me. I expect that this does not work

@macjohnny How can we make it work? :)

What's the issue?

(I've not tested it locally)

macjohnny:
@wing328 I don't understand the reason why `map` can't be a reserved keyword.

modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
```diff
@@ -8,7 +8,7 @@ import { Observable } from "rxjs/Observable";
 import { Observable } from "rxjs";
 {{/useRxJS6}}
 
-import { map } from "rxjs/operators";
+import { GlobalImportOperators } from "rxjs/operators";
```
macjohnny: `"rxjs/operators"` does not export a `GlobalImportOperators` symbol

wing328: Ah I got it....

wing328: so we need something like
```
import * as GlobalImportOperators from "rxjs/operators";
```

and then call it by 
```
GlobalImportOperators.map
```
(or something like that) ?

macjohnny:
@wing328 I vote against renaming imported symbols and for making them reserved keywords, since I consider it a design flaw to simply insert arbitrary (user-defined) variable names into a code template. When simply renaming imports, there will always be a naming conflict, although someties less likely.

See also 
https://github.com/OpenAPITools/openapi-generator/pull/2921

wing328:
> When simply renaming imports, there will always be a naming conflict, although sometimes less likely.

Agreed the naming conflicts can still occur but should be less likely.

I'll close this PR as no one has complained about naming conflicts.

We'll revisit this when there are more demands from the users to name their variables or models as "map" or other reserved words.

PR #2983, opened 2019-05-23
Use golang's provided method names (gin)

kkirsche:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. If contributing template-only or documentation-only changes which will change sample output, be sure to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) first.
- [ ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.1.x`, `5.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR
This commit modifies the gin template for the router to leverage the http constants for method types as defined by RFC 7231 section 4.3. These are documented on:

https://golang.org/pkg/net/http/#pkg-constants

This removes the need for the `strings` dependency and does not require any new dependencies, as `net/http` is already imported.

PR #3028, opened 2019-05-29
refactor(golang): Use http provided constants for http methods

mcristina422:
### PR checklist

@antihax @bvwells @grokify @kemokemo

### Description of the PR

This modifies the golang client to use the http constants for method types as defined by RFC 7231 section 4.3. These are documented on:

https://golang.org/pkg/net/http/#pkg-constants

This does not require any new dependencies, as net/http is already imported.

Based on: https://github.com/OpenAPITools/openapi-generator/pull/2983


wing328:
CI reports the following errors:
```
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[ERROR] Error fetching link: /home/circleci/OpenAPITools/openapi-generator/modules/openapi-generator/home/circleci/OpenAPITools/openapi-generator/modules/openapi-generator-cli/target/apidocs. Ignored it.
# _/home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/go/go-petstore
go-petstore/api_another_fake.go:17:2: imported and not used: "strings"
go-petstore/api_fake.go:17:2: imported and not used: "strings"
go-petstore/api_fake_classname_tags123.go:17:2: imported and not used: "strings"
FAIL	_/home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/go [build failed]
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (go-test) on project Goswagger: Command execution failed. Process exited with an error: 2 (Exit value: 2) -> [Help 1]
[ERROR] 
```

Looks like the `strings` import needs to be removed?

mcristina422:
Oh, the edge case comes into play when the `api` file gets rendered with a path that includes a path parameter. Since sometimes this exists (or doesn't exist) for multiple operations, the logic of including this is not trivial. 

Also `fmt` should probably also be moved into that block in 
3deaa80 since `fmt` and `strings` both are only used in that case. 

This is my first real interaction with mustache and this edge case seems more complex to handle than I initially thought. 

Maybe the simplest solution is to always import it and hush the compiler with 
```
// Linger please
var (
	_ context.Context
        _ strings.Replacer
)
```

Thoughts?

Also what does this do exactly? Maybe I'm misunderstanding something where this could be used
```
{{#imports}}	"{{import}}"
{{/imports}}
```

wing328:
The `imports` is for importing models (e.g. Pet) defined in the endpoint/operation.

wing328:
> This is my first real interaction with mustache and this edge case seems more complex to handle than I initially thought.

Yup, that's another edge case that I've not yet encountered with other generators.

As a workaround, what about using "strings" somehow in the code so as to hardcode the import?

mcristina422:
Ah, the import logic is handled here https://github.com/OpenAPITools/openapi-generator/blob/26146707994ae2e8a78e4a52cf1c8e3eaf23bba8/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java#L364-L371 so just adding `strings` to that was enough. No workarounds needed

wing328:
Thanks again for the PR, which has been included in the v4.0.2 release: https://twitter.com/oas_generator/status/1141610197766426626

PR #3047, opened 2019-05-31
Update version to 4.0.2-SNAPSHOT

jmini:
### PR checklist

### Description of the PR


appveyor.yml
```diff
@@ -50,6 +49,8 @@ build_script:
   - msbuild samples\client\petstore\csharp\OpenAPIClientNetStandard\Org.OpenAPITools.sln /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
   # install openapi-generator locally
   - mvn clean install --quiet
+  # run the locally installed openapi-generator-gradle-plugin
+  - gradle -PopenApiGeneratorVersion=4.0.2-SNAPSHOT -b modules\openapi-generator-gradle-plugin\samples\local-spec\build.gradle buildGoSdk --info
```
jimschubert: I think `-PopenApiGeneratorVersion=4.0.2-SNAPSHOT` can be removed after changes to fix `release_version_update.sh` and `release_version_update_docs.sh`

modules/openapi-generator-maven-plugin/examples/multi-module/java-client/pom.xml
```diff
@@ -18,7 +18,7 @@
             <plugin>
                 <groupId>org.openapitools</groupId>
                 <artifactId>openapi-generator-maven-plugin</artifactId>
-                <version>4.0.0</version>
```
jimschubert: This file was previously not being updated, which is why there's a jump from 4.0.0 to 4.0.2-SNAPSHOT in this PR.

wing328: Updated the release checklist to include this file: https://github.com/OpenAPITools/openapi-generator/wiki/Release-Checklist

jmini: It really depends what we would like to do. 

I like when documentation and usage examples (maven & gradle) use the last released version. But we can do it an other way in this project.

If we use the current SNAPSHOT version, at least we should add this at the top of the `pom`:
```xml
  <pluginRepositories>
    <pluginRepository>
      <id>sonatype-snapshots</id>
      <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
    </pluginRepository>
  </pluginRepositories>
```
in order for the users to be able to try the lib without having to compile the plugin locally them-self.
(and add of course the equivalent for the gradle examples)

jimschubert: I've added the `pluginRepository` config to the maven plugin's examples which are referenced in the `release_version_update_docs.sh` script:

```
 java-client.xml
 multi-module/java-client/pom.xml
 non-java-invalid-spec.xml
 non-java.xml
```

I think it would make sense for documentation to always reference a major version, but I think samples in the repository should probably reference the current version of the repository. That is, maven examples are the same as everything under `./samples`, since they're examples of how the maven plugin works. I would personally think it was a bug if if I found an example in another repository that didn't track the same version of the repository.

I don't feel too strongly about this, though.

PR #3077, opened 2019-06-03
[typescript-rxjs] refactor to arrow functions and short hand object creation

denyo:
### PR checklist

### Description of the PR

* refactor to arrow functions where possible
* abstract error throwing null checks into a function `throwIfRequired`
* remove unused parameter in `RequiredError` (this is a breaking change if people are using it outside the generated code)
* refactor `headers` and `query` objects to be created inline to be less verbose and skip redundant null checks

This all results in a noticeable code savings. 
In my current react project with CRA2 (9 controllers and 14 endpoints) I get a build with 4368 characters less which is about 4%, although file size after gzip it is only 230 Bytes which is about 1%.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
@@ -132,7 +132,8 @@ export class BaseAPI {
         Object.assign(Object.create(Object.getPrototypeOf(this)), this)
 }
 
-class RequiredError extends Error {
+// export for not being a breaking change
+export class RequiredError extends Error {
     name: 'RequiredError' = 'RequiredError';
 }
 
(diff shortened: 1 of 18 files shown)


modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -175,17 +164,24 @@ export interface RequestOpts {
...
+        const value = params[key];
+        if (value instanceof Array) {
+            return value.map((val) => `${encodeURIComponent(key)}=${encodeURIComponent(String(val))}`)
+                .join('&');
+        }
+        return `${encodeURIComponent(key)}=${encodeURIComponent(String(value))}`;
+    })
+    .join('&');
+
+// alias fallback for not being a breaking change
```
macjohnny: nice

denyo:
I missed the condition to add properties to the `headers` and `query` objects. It's there now with the spread operator.

PR #3078, opened 2019-06-03
[C++] [Qt5] Add API timeout handling

### LINKED ISSUE #3064: [REQ][Qt5] Optional TimeOut
### Is your feature request related to a problem? 
Yes
### Please describe.
I think Qt5 doesn't manage timeout problems by default. This means that if there is a timeout because of a firewall or an unresponsive server, the connection won't die since there is no mechanism to solve this situation.

The usual solution is to set a Qtimer with setsingleshot=true and a timeout, then connect this qtimer timeout() signal to the QNetworkReply abort() slot. If the response happens before the timeout, then the timer has to be stopped. I think you can do this by connecting the finished signal of QNetworkReply to stop slot of the timer, or by calling stop manually when processing the response.

I thought of maybe making the constructor of the generated client depend on an optional timeout with a default value, for example MyConstructor(int timeout = 30). Another option is to set the optional timeout before making a request by using a simple setter.

I've also read about a complementary solution for a similar problem that consists in to combine the use of a qtimer with downloadProgress to check if a download has stopped or is slower than a threshold then abort the connection.

None comes to my mind

## Additional context

etherealjoy:
### PR checklist

### Description of the PR

Fixes #3064
- Add timeout specifier for api calls
- Default behaviour is unchanged, by default no timeout like before
- backwards compatible

`cc:`
@stkrwork @ravinikam @MartinDelille 

`cc:`
@luStopai

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
@@ -45,16 +45,27 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
     : QObject(parent), manager(nullptr)
 {
     qsrand(QDateTime::currentDateTime().toTime_t());
-
+    timeout = 0;    
+    timer = new QTimer();
     manager = new QNetworkAccessManager(this);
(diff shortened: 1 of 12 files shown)


etherealjoy:
@stkrwork 
Thank you for the review

PR #3080, opened 2019-06-03
[scala-play-server] Fix generated enums named after a reserved word

adigerber:
### PR checklist

### Description of the PR

This is a bugfix for the Scala-Play language.  
When an enum property's name is a reserved word in Scala, it is escaped with backticks (in `AbstractScalaCodegen`) so the generated code can compile. However, generating code with backticks is a bit tricky - the name must start and end with a backtick else the code doesn't compile.

In Scala-Play language, when enum properties are generated, the enum definition is contained inside the companion object of the model. The name of the enum is camelized (as per `toEnumName` method). To be able to serialize the enums, a corresponding JSON format must be introduced as well, named `<EnumName>JsonFormat`. If the enum name contains backticks (as is the case with a reserved word), the JSON format will contain backticks that are not properly terminated.

For example, given the property name `type`, it is escaped to ``` `type` ```, and the enum name is camelized to ``` `Type` ```, and the resulting JSON format is named ``` `Type`JsonFormat ``` which produces invalid Scala code.

To fix this, I changed `toEnumName` to strip the backticks before camelizing the property's name. As of now there are no reserved words in Scala that start with an uppercase; using predefined type names (such as `Int`, `Array`, `Any` and so on) as the enum name does not affect the compilation (i.e. these are valid names that can compile).

### Technical Committee
@clasnake, @jimschubert, @shijinkui, @ramzimaalej

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
@@ -358,7 +358,15 @@ public class ScalaPlayFrameworkServerCodegen extends AbstractScalaCodegen implem
 
     @Override
     public String toEnumName(CodegenProperty property) {
-        return camelize(property.name);
+        return camelizeStripReservedEscape(property.name);
+    }
+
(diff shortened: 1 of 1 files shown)


jimschubert:
Thanks for this improvement!

PR #3091, opened 2019-06-04
[C++][Qt5] Add Q_DECLARE_METATYPE to the generated models and remove ref in signals

### LINKED ISSUE #3063: [REQ][Qt5] Make response objects queueable
### Is your feature request related to a problem? 
No, it's an improvement
### Please describe.
I think it would be great if we could run requests in a QThread, but right now it's not possible because the returned objects cannot be queueable (signals&slots) from one thread to a different thread.
This is in order to be able to run requests in background without freezing the UI.

I think if the response objects where pointers to qobjects instead of regular objects it should work. Notice that they should be registered with qRegisterMetaType() ( https://doc.qt.io/qt-5/qt.html#ConnectionType-enum )

Right now the alternative is to manually create a proxy class that makes use of the generated client to make the request, receive the response and then convert the response objects to some pointers to custom qobjects, which are queueable. So the purpose of this suggestion is to avoid this workaround

## Additional context

etherealjoy:
### PR checklist

### Description of the PR

Fixes  #3063

Changes:
- Add `Q_DECLARE_METATYPE` to enable type to be known to  `QMetaType`
- Remove pass by reference from signals due to possibility of descoping

`cc`:
@stkrwork @ravinikam @MartinDelille 

`cc`:
@luStopai


wing328:
@etherealjoy when you've time, can you please ping me via https://gitter.im (ID: wing328) for a quick chat?

PR #3039, opened 2019-05-31
Ruby client: escape path parameters

ccouzens:
Path parameters should be escaped when encoded into the path.

In the path '/pet/{petId}' let's pretend petId is a string instead of a
number.

If the user uses "Bobby" as the petId then they correctly get the path
'/pet/Bobby'.
But if they put 'Bobby/Tables' as the petId then they used to get the
path '/pet/Bobby/Tables' which will be interpreted by the server as a
different route.

Using CGI::Escape they now get '/pet/Bobby%2FTables' which is correct.

Ruby technical committee:
@cliffano (2017/07) @zlx (2017/09) @autopp (2019/02)

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. If contributing template-only or documentation-only changes which will change sample output, be sure to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) first.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.1.x`, `5.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

See top

PR #3093, opened 2019-06-04
Ruby: Avoid double escaping path items

ccouzens:
`URI.encode` is obsolete. `CGI.escape`, `URI.encode_www_form` or
`URI.encode_www_form_component` are recommended instead.
https://ruby-doc.org/stdlib-2.6/libdoc/uri/rdoc/URI/Escape.html#method-i-escape

URI.encode has different behaviour to CGI.escape:

```ruby
URI.encode('hello/world?test%string')
=> "hello/world?test%25string"
CGI.escape('hello/world?test%string')
=> "hello%2Fworld%3Ftest%25string"
```

I recently raised pull request #3039
201cbdce29cc6cdbbbe9efcb1afb250a05bc2ffd

That pull request escapes path items at insertion.

Before either pull request, the path item 'hello?world' would go into
the URL as 'hello?world'. That behaviour was insecure as if an attacker
could control the path item value, they could change the URL the
application connected to.

After #3039 'hello?world' would go in as 'hello%253Fworld'. This was
safer than before, but it's still not correct.
If I'd realised at the time, I would have made it correct at the time.

What this pull request does is make it go in as 'hello%35world', which
is correct.

ApiClient::build_request_url was URI.encoding the whole path.
This wasn't protecting against all undesirable characters in the path
items, but was escaping % characters a 2nd time which was unhelpful.

I have additionally removed URI.encode from Configuration::base_url as I
can't see any benefit it could be bringing.
There is no justification for it in the commit where it was originally
added: 47c8597d36a9bc0983ba5c40e2489bb094f9f076

CC: ruby technical committee @cliffano @zlx @autopp

### PR checklist

### Description of the PR

see above

### DIFF
--- a/modules/openapi-generator/src/main/resources/ruby-client/api.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api.mustache
@@ -2,7 +2,6 @@
 {{> api_info}}
 =end
 
-require 'uri'
 require 'cgi'
 
 module {{moduleName}}
--- a/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
@@ -7,7 +7,6 @@ require 'json'
 require 'logger'
 require 'tempfile'
 require 'typhoeus'
-require 'uri'
 
 module {{moduleName}}
   class ApiClient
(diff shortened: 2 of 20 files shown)


ccouzens:
Thank you @autopp 

> BTW, as a confirmation just in case, is 'hello%35world' in the description of PR a mistake of 'hello%3Fworld'?

Yes, I should have put 'hello%3Fworld'

autopp:
@ccouzens The PR is merged. Thanks 👍

PR #3112, opened 2019-06-06
Prefix request interfaces with classname of api

bobvanderlinden:
### PR checklist

- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @leonyu

### Description of the PR

At the moment the `typescript-fetch` generator outputs interfaces for each operation. For example: `OperationARequest`. This request interface can reside in multiple API classes when the specification has duplicate operation ids. This results in name collisions when the typescript library tries to export all types from the API files in a single file.

The duplicate operationIds are the result of OpenAPI specification generators that derive the specification from code. Usually they generate tags from classes and operationId from method names.

This PR will prefix the classname of the API to each of the `{{operationId}}Request` interfaces. That way such collisions will happen to a lesser degree.

This replaces the PR filed against master (#3111).

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -15,7 +15,7 @@ import {
 {{#operations}}
 {{#operation}}
 {{#allParams.0}}
-export interface {{operationIdCamelCase}}Request {
+export interface {{classname}}{{operationIdCamelCase}}Request {
     {{#allParams}}
     {{paramName}}{{^required}}?{{/required}}: {{#isEnum}}{{{datatypeWithEnum}}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}};
     {{/allParams}}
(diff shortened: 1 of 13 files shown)


macjohnny:

macjohnny:
maybe we should add an option, then we could avoid a breaking change. what do you think?

bobvanderlinden:
I have reconsidered this PR.

It's nice that it will support duplicate operationIds, like the older generator did, but duplicate operationIds go against the OpenAPI specification.

On the other hand, many OpenAPI generators that base the specification on controller classes and action methods _do_ output duplicate operationIds. They essentially output invalid OpenAPI specifications. For this reason you _do_ want to support duplicate operationIds.

That makes changes like this PR a bit strange. I'm on the fence if this needs to be merged. I do think many people will run into this issue, because they generate openapi from an existing server implementation.

Making this an option is a possibility, but it only adds to the complexity for something that doesn't have huge impact (it only changes the names of Request types).

macjohnny:

macjohnny:
I am still in favor of an option, in order to avoid breaking changes.

TiFu:
I am against adding an option (and merging this PR). It just adds complexity on our end for something which should be fixed upstream in the generator creating the OpenAPI spec from code. 

IMO the oai-generator should alert the user to this issue and abort the code generation process - after all the spec is clearly invalid.

bobvanderlinden:
@TiFu I agree. It would be nice if there were some guidelines on the mapping between class/method with tag/operationId. I'm not aware of any and it seems like that is what is causing these invalid specifications to pop up after generating them from code.

Do you know of any?

wing328:
@bobvanderlinden you will need to customize the toOperationId method so as to have complete control of the method name

Issue #522, opened 2018-07-09
[Go] Optional values should be pointers


langston-barrett:
##### Description

If a field of a struct isn't specified as `required`, then it should be a pointer to that value so that one can differentiate between unset by the user or set to the "zero value". See the following Go playground: https://play.golang.org/p/85r_JbH5XVr

See also e.g. this [Stackoverflow answer](https://stackoverflow.com/questions/38486564/golang-unmarshal-marshal-json-with-int-set-to-0-does-not-seem-to-work).

##### openapi-generator version

Docker container

##### OpenAPI declaration file content or url
```
---
openapi: 3.0.0
info:
  title: unset
  version: 0.1.0

paths:
  "/foo":
    get:
      summary: foo
      responses:
        '200':
          description: OK
          content:
            application/xml:
              schema:
                "$ref": "#/components/schemas/Bar"

components:
  schemas:
    Bar:
      properties:
        X:
          type: integer
        Y:
          type: integer
      required:
        - X
```

grokify:
It should be. The template is here:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/go/model.mustache

You can do the following to try this out:

```
$ git clone https://github.com/OpenAPITools/openapi-generator
$ cd openapi-generator
$ git checkout -b go/client/fix/optional-as-pointer
$ vi modules/openapi-generator/src/main/resources/go/model.mustache
$ mvn clean package
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar help
```
Use the resulting `openapi-generator-cli.jar` to rebuild your client and see if it works as expected.

If so, then do the following to send a PR with the following:

```go
$ bin/go-petstore.sh
$ cd samples/client/petstore/go
$ mvn integration-test
$ cd ../../../..
$ mvn package
$ git add *
$ git commit -m 'use pointers for optional properties'
$ bin/util/ensure-up-to-date
```

If `ensure-up-to-date` creates new files or tells you to update your code, do it and run it again. Rinse and repeat until it says your working tree is clean.

Then push to your fork and send a PR.

siddharthist:
@grokify Wow, thank you for the thorough instructions! I hope you have this saved as a "quick reply" :smile: I'd like to solve this, I'll report back here (within a week) if I'm able to.

siddharthist:
[...]

PR #3113, opened 2019-06-06
[Go][client] make structure members pointers, provide custom marshalling

bkabrda:
### PR checklist

### Description of the PR

@antihax (2017/11) @bvwells (2017/12) @grokify (2018/07) @kemokemo (2018/09

This is my shot at fixing #522. It's still WIP, but I wanted to gather some initial comments to see if it makes sense to take it all the way to the end.

Basically, this PR makes all the members of generated structures pointers and it adds getters/setters for them. Additionally, it provides custom marshalling to check that required/nullable attributes are set properly in JSON when marshalling. Todo list of things that need to be done before this could be considered good to go:

* [x] Not sure about this, but I think I should change the basic types for struct members in Java code (typeMappings).
  * It actually turns out to be much cleaner to do this purely on the template level, so that's what I did.
* [x] This will need update of documentations as well.
  * I updated to `model_doc.mustache` to reflect the changes.
* [x] I need to take a closer look at enums and see if something needs to be adapted for them to work ok.
  * This seems to not require any changes for enums.
* [x] One of the lines (for getting the "zero" value of a type) is duplicated in two places; it'd be ideal to hide it behind some sort of macro. Quick google search didn't reveal if mustache provides something like this, I need to take a look again.
  * Turns out that the easiest way to do this is to create an uninitialized var of that specific type and return that - no macro needed.
* [x] Not sure if getters should return empty instance or `nil` for structure members that point to other structures. I guess `nil` would be more in line with what's going on here generally, but not sure - I'm trying to mimic `gen-accessors.go` referenced in [1], which returns empty structures, but I'm not 100 % sure why that's a good thing to do.
  * I think this is ok as long as it's documented, which it is now.
* [x] I don't think custom unmarshalling needs to be created, but not sure.
  * This would probably only be required if we wanted to add `Get{{name}}ExplicitNull` method to nullable struct members. I don't think we need that right now. If we ever need that, we can add it without introducing and API breakage.
* [x] Fix tests

Any comments would be appreciated.
Thanks!

[1] https://github.com/OpenAPITools/openapi-generator/issues/522#issuecomment-499014015


bkabrda:
All right, so I think this is in a pretty good shape to get reviewed right now. I've regenerated the v2 code and fixed all tests but one - `TestFindPetsByTag`. It seems to me like the testing server is broken right now (returning 500 when I run the test locally) and my change isn't really touching anything related to that test, so I *think* I didn't break it.

I haven't included the changes from running `./bin/openapi3/go-petstore.sh`, since it seems this hasn't been run for a while and would include other changes as well - probably a separate PR would be good to address that (I'm happy to do that once this is merged).

I'll continue looking into this and will be adding further changes as additional commits in case someone starts reviewing, so that it's clear what's been reviewed and what hasn't.

antihax:
This looks like a good start. I would be concerned with collisions in the name space down the line but so far it looks like that is avoided.

bkabrda:
Allright, I added documentation for the included utility functions. From my POV everything seems to be fine, so getting a full review with list of things to fix/improve would be great.

@antihax yeah, I think the naming collisions are avoided (at least I can't think of any collisions happening right now).

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientCodegen.java
```diff
@@ -122,6 +127,32 @@ public void processOpts() {
...
+        List<Map<String, String>> imports = (List<Map<String, String>>) objs.get("imports");
+
+        boolean addedErrorsImport = false;
+        List<Map<String, Object>> models = (List<Map<String, Object>>) objs.get("models");
+        for (Map<String, Object> m : models) {
+            Object v = m.get("model");
+            if (v instanceof CodegenModel) {
+                CodegenModel model = (CodegenModel) v;
+                if (!model.isEnum) {
+                    imports.add(createMapping("import", "encoding/json"));
```
jimschubert: This and the import below have a potential to break in the case that someone invoke with `--import-mappings=import=encoding/json,import=error` (or however the mappings are done for golang). See CodegenConfigurator line 540.

modules/openapi-generator/src/main/resources/go/README.mustache
```diff
@@ -113,6 +113,22 @@ r, err := client.Service.Operation(auth, args)
 {{/isOAuth}}
 {{/authMethods}}
 
+## Documentation for Utility Methods
+
+Due to the fact that model structure members are all pointers, this package contains
+a number of utility functions to easily obtain pointers to values of basic types.
+Each of these functions takes a value of the given basic type and returns a pointer to it:
+
+* `Bool`
```
jimschubert: I like this API decision, but I wonder if it would be better to prefix these with something to avoid conflict with other potentially imported types or models. Something like `PtrBool` or similar?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientCodegen.java
bkabrda: Ok, thanks, I'll take a look.

modules/openapi-generator/src/main/resources/go/README.mustache
bkabrda: Sounds like a good idea, I'll do that.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientCodegen.java
bkabrda: But the same argument would apply to current `AbstractGoCodegen.postProcessModels` which adds some imports too, right? So it's a generic issue that's already there AFAICS. (And I think the best course of action would be to deduplicate the imports list after processing is done, but I think that's out of scope of this PR).

wing328:
@bkabrda looks like a rebase has failed and as a result, this PR contains commits not authored by you. Can you please create a new PR by cherrying picking the commits authored by you?

wing328:
It also seems to me it's a breaking change without fallback but I'll review again once a new PR is submitted

bkabrda:
Yeah, I'll do a new PR. This functionality does contain backwards incompatible changes, which is why I was advised by Jim Schubert to rebase it against 4.1.x. Let's discuss at the new PR, I'll open it shortly and link it here.

bkabrda:
New PR opened: #3241

PR #3130, opened 2019-06-09
Ruby: Add option for old path parameter behaviour

### LINKED ISSUE #3119: [BUG] [Ruby] client escapes all path parameters
#### Bug Report Checklist

##### Description

Our application's OpenAPI schema contains path parameters that contains '/'. The latest generated Ruby client stopped working for all the operations that contain a path parameter with '/' in it. This is a regression for our use case of openapi-generator. 

I am using the latest openapi-generator-cli docker image. This is a regression since v4.0.0 tag.

https://docs.pulpproject.org/en/3.0/nightly/api.json

```bash
docker run -u $(id -u) --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate \
    -i /local/api.json \
    -g ruby \
    -o /local/$1-client \
    -DgemName=$1_client \
    -DgemLicense="GPLv2" \
    -DgemVersion=0.1.0rc2 \
    --skip-validate-spec \
    --strict-spec=false
```

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/pull/3039

##### Suggest a fix

@ccouzens, could you make the path parameter escaping optional?

ccouzens:
The OpenAPI standard is to percent encode special characters such as
slashes, hashes and question marks when they appear in path parameters.

The Ruby client generator was buggy. It did not encode these characters.
So a few days ago I fixed it.
https://github.com/OpenAPITools/openapi-generator/pull/3039

Unfortunately, some projects rely on the incorrect behaviour.
In particular, they need slashes not to be escaped.

For example:
https://github.com/OpenAPITools/openapi-generator/issues/3119

I've added an additional property to the ruby generator to undo the
escaping of slashes in path parameters. This will enable those projects
to continue to make use of the Ruby client code generator.
`--additional-properties compatibilitySlashesInPathParameters`

I'd suggest the long term plan for projects that need this behaviour
should be to work with the OpenAPI Specification writers to support
marking certain path parameters as allowing slashes.

### PR checklist

@cliffano @zlx @autopp 

### Description of the PR

Fix #3119
See above

### DIFF
--- a/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
@@ -254,6 +254,9 @@ module {{moduleName}}
 
     def build_request_url(path)
       # Add leading and trailing slashes to path
+      {{#compatibilitySlashesInPathParameters}}
+      path = path.gsub('%2F', '/')
+      {{/compatibilitySlashesInPathParameters}}
       path = "/#{path}".gsub(/\/+/, '/')
       @config.base_url + path
     end
--- a/docs/generators/ruby-client.md
+++ b/docs/generators/ruby-client.md
@@ -1,0 +1,10 @@
+
+---
+id: generator-opts-client-ruby-client
+title: Config Options for ruby client
+sidebar_label: ruby-client
+---
+
+| Option | Description | Values | Default |
+| ------ | ----------- | ------ | ------- |
+|compatibilitySlashesInPathParameters|Don't encode slashes in path parameters. Not compatible with the OpenAPI standard. Designed to support projects which relied on the non-standard behaviour before it was fixed.| |false|
--- a/modules/openapi-generator/src/main/resources/ruby-client/api_client_spec.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api_client_spec.mustache
@@ -40,6 +40,25 @@ describe {{moduleName}}::ApiClient do
           expect({{moduleName}}::Configuration.default.base_path).to eq('')
         end
       end
+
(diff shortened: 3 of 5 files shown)


modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
```diff
@@ -255,6 +255,9 @@ module {{moduleName}}
     def build_request_url(path)
       # Add leading and trailing slashes to path
       path = "/#{path}".gsub(/\/+/, '/')
+      {{#compatibilitySlashesInPathParameters}}
+      path = path.gsub('%2F', '/')
```
dkliban: This needs to happen before `path = "/#{path}".gsub(/\/+/, '/')`, otherwise the URL starts with // when the path already started with /.

dkliban:
@ccouzens I test this PR and it does not solve the problem all the way. The URL that is generated by the api_client::build_request_url now contains and extra '/' on the front. The server always returns a 404. This behavior is very similar to https://github.com/OpenAPITools/openapi-generator/issues/2702.  I think changing the order of substitution should fix the problem.

modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
ccouzens: Thank you.

I've amended the commit. The order is changed and `/%2F` going to `/` is now in the test cases.

ccouzens:
@dkliban I've swapped the order like you suggested. That should fix the double slash problem

ccouzens:
I'm closing this because #3204 solved the issue instead.

PR #3146, opened 2019-06-12
[Ruby][Client] Add "send" to ruby reserved word list

martynav:
Ruby clients call `self.send(:"#{key}=", value)`.  This breaks if a response entity exposes a 'send' attribute.

This change adds ruby's "#send" to the reserved word list for the ruby generator.


wing328:
cc Ruby Tech committee: @cliffano (2017/07) @zlx (2017/09) @autopp (2019/02)

autopp:
I'm sorry that the review will be late. I'll review it.

PR #2494, opened 2019-03-24
[go] Generate inline enum models and prefix all enums

fantavlik:
### PR checklist

- [ ✓ ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ✓ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ✓ ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. *Default: `master`*.
- [ ✓ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

CC: @antihax @bvwells @grokify @kemokemo 

This aims to address the following issues:
https://github.com/OpenAPITools/openapi-generator/issues/535
(also see related closed PR: https://github.com/OpenAPITools/openapi-generator/pull/542)
https://github.com/OpenAPITools/openapi-generator/issues/2050
https://github.com/OpenAPITools/openapi-generator/issues/2360

### Description of the PR

There are four main aspects:
1. Enums that are defined inline were not being generated (see https://github.com/OpenAPITools/openapi-generator/issues/2360 for an example), this PR follows similar logic to the typescript-fetch generation template to remedy this: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache#L92-L109
2. Enums names often result in namespace collisions within the package being generated (see https://github.com/OpenAPITools/openapi-generator/issues/535), prefixes have been added to help prevent this for most cases.
3. Numeric enum values were improperly being quoted, now only strings are.
4. Numeric enum names were not properly escaping plus/minus signs and decimal points e.g. to differentiate between `-1` and `1` or `1.1` and `11`

PR #2897, opened 2019-05-15
[core] Support case where composed schema is single type definition

fantavlik:
### Description of the PR

Fix for https://github.com/OpenAPITools/openapi-generator/issues/2164

The general fix is to handle the case where a composed schema is a single element defining a type.

For example:
```yaml
allOf/anyOf/oneOf:
  type: string
```

^ this case is not really model composition but rather defining the type of the model which is equivalent to `type: <>`

Input:
```yaml
components:
  schemas:
    RatType:
      anyOf:
      - type: string
      enum:
        - NR
        - EUTRA
        - WLAN
        - VIRTUAL
      type: string

```

Output for Go model_rat_type.go:
```go
package openapi
type RatType string

// List of RatType
const (
	NR RatType = "NR"
	EUTRA RatType = "EUTRA"
	WLAN RatType = "WLAN"
	VIRTUAL RatType = "VIRTUAL"
)
```

@wing328 can you review this if you get a moment as it is a core change? I've tested Go and regenerated the sample files and there are no changes for Go at least, not sure about other languages.

cc: @mroland91 and @goyalnikhil

Issue #2518, opened 2019-03-26
[BUG] [PYTHON] Naming conflict between named models and auto-named inline objects


ivan-gomes:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When there is a defined model, ex: `Content`, that has a property of type object (defined inline), ex: `body`, the codegen creates a model for it, ex. `ContentBody`. However, if there is another defined model by the same name, ex. `ContentBody`, this results in a naming conflict where the defined model is ignored/overwritten. See example spec for clarification.

##### openapi-generator version

3.3.4

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
  title: Test
  version: 1.0.0
paths: {}
components:
  schemas:
    Content:
      type: object
      properties:
        body:
          type: object
          properties:
            wrong:
              $ref: '#/components/schemas/ContentBody'
    ContentBody:
      type: object
      properties:
        right:
          type: string
```

##### Command line used for generation

`java -jar openapi-generator-cli-3.3.4.jar generate -g python -DpackageName=openapi_client -DprojectName=openapi-python-client -DpackageVersion=1.0.0 -DhideGenerationTimestamp=true -i test-openapi.yaml -o /var/tmp/openapi-client`

##### Steps to reproduce

1. Save example spec as `test-openapi.yaml`
2. Run command above to generate Python client
3. Observe under `openapi_client/models` that there are only two models, `Content` and `ContentBody`, where `ContentBody` has property `wrong` (property of the inlined object) but not `right`

##### Related issues/PRs

None

##### Suggest a fix

Desired functionality would be that naming conflicts would be detected and the auto-generated name would be made unique, potentially using an auto-incrementing suffix. In this particular example, there should be a total of three models: Content (as defined), ContentBody (as defined), and ContentBody1 (auto-named from inline object).

PR #3162, opened 2019-06-17
[core][go] Generate inline models that were not previously generated

fantavlik:
#### Core Changes
* Refactor InlineModelResolver.java to recursively identify inline schemas that require their own model definitions.
Examples of inline schemas that were not being generated that need their own models: enums, objects with defined properties, and composed schemas.
Examples of inline locations that are now being checked recursively for inline models: within object properties, within object additionalProperties, within definitions of composed schemas (similar approach to https://github.com/OpenAPITools/openapi-generator/pull/2112), within definitions of "not", and within array items.
* Add more contextual naming rather than InlineObject, schemas defined inline within an operation section will get prefixes with the operation id, schemas within an array items section will get the parent model name + `Items` suffix, etc. Schemas with "title" fields will still use the title if defined.
#### Go changes (affecting all Go-based generators as changes were made within AbstractGoCodegen.java)
* Prefix all enum values with model name, this is needed to avoid collisions within the package
* Fix issue where numerical enums were quoted
* Fix issue where numerical enums would begin with underscore

This is a followup from https://github.com/OpenAPITools/openapi-generator/pull/2494 - I attempted to fix this issue using Go templates at first but realized that a core change could fix these kinds of issues across all languages by refactoring the InlineModelResolver.java.

This also takes the support added by https://github.com/OpenAPITools/openapi-generator/pull/2112 and extends that implementation to be recursive ~(I changed the suffixes slightly to be `AllOf`, rather than `_allOf` but otherwise the support should be the same)~ the suffixes now match `AllOf`, `OneOf`, `AnyOf` since this seems like the better convention.

Models like this which have many inline schemas that should be resolved into their own separate schemas/models were losing information (e.g. declaring enums as simple strings or declaring arrays of models arrays of simple objects).

Example:
```yaml
Cat:
  description: Complex model with many inline models defined within it.
  type: object
  properties:
    myFood:
      oneOf: # inline oneOf
        - type: object
          title: Food
          properties:
            ingredients:
              type: array
              items:
                type: string
            pounds:
              type: integer
          required:
            - fooProp
    myHabitat:
      allOf: # inline allOf
        - type: object
          title: HabitatGeography
          properties:
            lat:
              type: number
              format: float
            long:
              type: number
              format: float
            continent:
              type: string
              enum:
                - Africa
                - Antarctica
                - Asia
                - Europe
                - North America
                - Oceania
                - South America
          required: [lat, long]
        - type: object
          description: Weather of the habitat defined here
          properties:
            rainfallInches:
              type: number
            averageTemperatureCelsius:
              type: number
    myTaxonomy:
      anyOf: # inline anyOf
        - description: Referenced schema
          type: object
          title: Species
          properties:
            name:
              type: string
            genus:
              type: string
            karyotype:
              type: string
        - description: Referenced schema
          type: object
          title: Order
          properties:
            name:
              type: string
            class:
              type: string
    cuteness: # inline enum
      type: integer
      enum: [1,3,5]
    kittens:
      type: array
      items: # inline array where the items are inline models
        allOf:
    preferences:
      type: object
      properties:
        favoriteToy: # inline object with properties defined
          type: string
      additionalProperties: # inline object with additionalProperties containing inline models
        title: Metadata
        type: string
        enum: [hidden,createdOn,createdBy,modifiedOn,modifiedBy]
```

Sample output before these changes (Go):
```go
// Complex model with many inline models defined within it.
type Cat struct {
	MyFood OneOfobject `json:"myFood,omitempty"`
	MyHabitat map[string]interface{} `json:"myHabitat,omitempty"`
	MyTaxonomy AnyOfobjectobject `json:"myTaxonomy,omitempty"`
	Cuteness int32 `json:"cuteness,omitempty"`
	Kittens []Cat `json:"kittens,omitempty"`
	Preferences map[string]string `json:"preferences,omitempty"`
}
...
// OneOfobject does not exist - not generated
// AnyOfobjectobject does not exist - not generated
```

New output (Go):
```go
// Complex model with many inline models defined within it.
type Cat struct {
	MyFood CatMyFood `json:"myFood,omitempty"`
	MyHabitat CatMyHabitat `json:"myHabitat,omitempty"`
	MyTaxonomy CatMyTaxonomy `json:"myTaxonomy,omitempty"`
	Cuteness CatCuteness `json:"cuteness,omitempty"`
	Kittens []CatKittensItems `json:"kittens,omitempty"`
	Preferences CatPreferences `json:"preferences,omitempty"`
}
...
type CatMyFood struct {
	Ingredients []string `json:"ingredients,omitempty"`
	Pounds int32 `json:"pounds,omitempty"`
}
...
type CatMyHabitat struct {
	Lat float32 `json:"lat"`
	Long float32 `json:"long"`
	Continent CatMyHabitatAllOfContinent `json:"continent,omitempty"`
	RainfallInches float32 `json:"rainfallInches,omitempty"`
	AverageTemperatureCelsius float32 `json:"averageTemperatureCelsius,omitempty"`
}
...
type CatMyHabitatAllOfContinent string

// List of CatMyHabitatAllOfContinent
const (
	CAT_MY_HABITAT_ALL_OF_CONTINENT_AFRICA CatMyHabitatAllOfContinent = "Africa"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_ANTARCTICA CatMyHabitatAllOfContinent = "Antarctica"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_ASIA CatMyHabitatAllOfContinent = "Asia"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_EUROPE CatMyHabitatAllOfContinent = "Europe"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_NORTH_AMERICA CatMyHabitatAllOfContinent = "North America"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_OCEANIA CatMyHabitatAllOfContinent = "Oceania"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_SOUTH_AMERICA CatMyHabitatAllOfContinent = "South America"
)
...
type CatMyTaxonomy struct {
	Name string `json:"name,omitempty"`
	Genus string `json:"genus,omitempty"`
	Karyotype string `json:"karyotype,omitempty"`
	Class string `json:"class,omitempty"`
}
...
type CatCuteness int32

// List of CatCuteness
const (
	CAT_CUTENESS_1 CatCuteness = 1
	CAT_CUTENESS_3 CatCuteness = 3
	CAT_CUTENESS_5 CatCuteness = 5
)
...
type CatKittensItems struct {
	MyFood CatMyFood `json:"myFood,omitempty"`
	MyHabitat CatMyHabitat `json:"myHabitat,omitempty"`
	MyTaxonomy CatMyTaxonomy `json:"myTaxonomy,omitempty"`
	Cuteness CatCuteness `json:"cuteness,omitempty"`
	Kittens []CatKittensItems `json:"kittens,omitempty"`
	Preferences CatPreferences `json:"preferences,omitempty"`
}
...
type CatPreferences struct {
	FavoriteToy string `json:"favoriteToy,omitempty"`
}
...
// Note: Go doesn't currently support complex additionalProperties, but this model is still generated:
type CatPreferencesMetadata string

// List of CatPreferencesMetadata
const (
	CAT_PREFERENCES_METADATA_HIDDEN CatPreferencesMetadata = "hidden"
	CAT_PREFERENCES_METADATA_CREATED_ON CatPreferencesMetadata = "createdOn"
	CAT_PREFERENCES_METADATA_CREATED_BY CatPreferencesMetadata = "createdBy"
	CAT_PREFERENCES_METADATA_MODIFIED_ON CatPreferencesMetadata = "modifiedOn"
	CAT_PREFERENCES_METADATA_MODIFIED_BY CatPreferencesMetadata = "modifiedBy"
)
```

I've only regenerated samples for Go so far but pretty much every language will have new models defined that were missing before.


fantavlik:
Issues that I have not tested but I believe are addressed in this PR: https://github.com/OpenAPITools/openapi-generator/issues/2518
https://github.com/OpenAPITools/openapi-generator/issues/2360
https://github.com/OpenAPITools/openapi-generator/issues/2495

fantavlik:
cc @wing328 for core changes and @antihax (2017/11) @bvwells (2017/12) @grokify (2018/07) @kemokemo (2018/09) for go-specific changes

fantavlik:
Would love to get approval of this approach before I start fixing tests (which I am happy to do) cc: @OpenAPITools/generator-core-team

grokify:
@fantavlik 

Can you explain more how this handles more than one schema? In the example, `anyOf` is the only example with more than one schema, but it would be useful all the examples to have more than one schema.

In the `myTaxonomy` example you provided, it's unclear how the `CatMyTaxonomy` struct is generated from the specification as the resulting struct as a `Name` property which isn't in the example.  It may help if we had example schema definitions for `Species` and `Order`.

Spec:

```yaml
Cat:
  type: object
  properties:
    myTaxonomy:
      anyOf: # inline anyOf
        - $ref: '#/components/schemas/Species'
        - $ref: '#/components/schemas/Order'
```

```go
type CatMyTaxonomy struct {
	Name string `json:"name"`
}
```

Here are some example schema definitions for `Species` and `Order` for discussion

* Normally, structs would be generated for `Species` and `Order`. Would these still be generated?
* How would these differing properties, such as `Species.genus` and `Order.class` be implemented in the `CatMyTaxonomy` struct?

```yaml
Species:
  description: Referenced schema
  type: object
  properties:
    name:
      type: string
    genus:
      type: string
    karyotype:
      type: string
Order:
  description: Referenced schema
  type: object
  properties:
    name:
      type: string
    class:
      type: string
```

fantavlik:
Good call @grokify - the example I gave didn't do a good job of demonstrating  the change, let me try and rework it to be fully inline.

For your specific question about `CatMyTaxonomy` the `myTaxonomy` field has been pulled into its own schema under `#/components/schemas` as if it had originally been defined like:
```yaml
components:
  schemas:
    CatMyTaxonomy:
      anyOf: # inline anyOf
        - $ref: '#/components/schemas/Species'
        - $ref: '#/components/schemas/Order'
    Cat:
      type: object
        properties:
          myTaxonomy:
            - $ref: '#/components/schemas/CatMyTaxonomy'
```
So I'm not changing anything about how `anyOf` are generated, just making sure that models defined inline (in this case within the `myTaxonomy` property of `Cat`) are generated the same as if they were defined in the schemas section (as with the `CatMyTaxonomy` snippet above).

The missing referenced schemas are:
```
    Food:
      type: object
      properties:
        ingredients:
          type: array
          items:
            type: string
        pounds:
          type: integer
      required:
        - fooProp
    Habitat:
      type: object
      properties:
        lat:
          type: string
        long:
          type: string
      required: [lat, long]
    Species:
      type: object
      properties:
        name:
          type: string
      required: [name]
    Order:
      type: object
      properties:
        name:
          type: string
      required: [name]
```

> Normally, structs would be generated for Species and Order. Would these still be generated?

Yes indeed, the key to the change I've made is that the `anyOf` model for the Cat.myTaxonomy taxonomy is being generated whereas before these changes a reference to a missing `AnyOfSpeciesOrder` model was being generated.

> How would these differing properties, such as Species.genus and Order.class be implemented in the CatMyTaxonomy struct?

So just updating the example to use these for Species and Order:
```yaml
Species:
  description: Referenced schema
  type: object
  properties:
    name:
      type: string
    genus:
      type: string
    karyotype:
      type: string
Order:
  description: Referenced schema
  type: object
  properties:
    name:
      type: string
    class:
      type: string
```

Yields the following output which is consistent with how the go generator already does anyOf (a struct that is the union of all properties):

`model_cat_my_taxonomy.go`:
```go
type CatMyTaxonomy struct {
	Name string `json:"name,omitempty"`
	Genus string `json:"genus,omitempty"`
	Karyotype string `json:"karyotype,omitempty"`
	Class string `json:"class,omitempty"`
}
```
while still outputting the `Species` and `Order` models:
`model_species.go`:
```go
// Referenced schema
type Species struct {
	Name string `json:"name,omitempty"`
	Genus string `json:"genus,omitempty"`
	Karyotype string `json:"karyotype,omitempty"`
}
```
`model_order.go`:
```go
// Referenced schema
type Order struct {
	Name string `json:"name,omitempty"`
	Class string `json:"class,omitempty"`
}
```

fantavlik:
@grokify I updated the example in the PR description to have multiple entries in the anyOf and allOf and also added more nesting of models to show how many are currently missed by the existing InlineModelResolver.java implementation

grokify:
Thanks for the update. Much clearer.

Seems like the following should be namespaced to prevent collisions.

Other than that it looks reasonable to me.

```go
// Note: Go doesn't currently support complex additionalProperties, but this model is still generated:
type Metadata string

// List of Metadata
const (
	METADATA_HIDDEN Metadata = "hidden"
	METADATA_CREATED_ON Metadata = "createdOn"
	METADATA_CREATED_BY Metadata = "createdBy"
	METADATA_MODIFIED_ON Metadata = "modifiedOn"
	METADATA_MODIFIED_BY Metadata = "modifiedBy"
)
```

Perhaps something like:

```go
// List of Metadata
const (
	CAT_PREFERENCES_METADATA_HIDDEN Metadata = "hidden"
	CAT_PREFERENCES_METADATA_CREATED_ON Metadata = "createdOn"
	CAT_PREFERENCES_METADATA_CREATED_BY Metadata = "createdBy"
	CAT_PREFERENCES_METADATA_MODIFIED_ON Metadata = "modifiedOn"
	CAT_PREFERENCES_METADATA_MODIFIED_BY Metadata = "modifiedBy"
)
```

fantavlik:
Great feedback, @grokify 

I have updated the namespacing such that this is now generated:
```go
type CatPreferencesMetadata string

// List of CatPreferencesMetadata
const (
	CAT_PREFERENCES_METADATA_HIDDEN CatPreferencesMetadata = "hidden"
	CAT_PREFERENCES_METADATA_CREATED_ON CatPreferencesMetadata = "createdOn"
	CAT_PREFERENCES_METADATA_CREATED_BY CatPreferencesMetadata = "createdBy"
	CAT_PREFERENCES_METADATA_MODIFIED_ON CatPreferencesMetadata = "modifiedOn"
	CAT_PREFERENCES_METADATA_MODIFIED_BY CatPreferencesMetadata = "modifiedBy"
)
```

also updated the InlineModelResolverTest.java to use the more illustrative example that you suggested (the one in this PR description above)

Let me know if there's anything else needed.

jimschubert:
I think the work in this PR may resolve some issues around Enum handling. The biggest issue I see is that it results in a breaking issue for consumers of existing clients for every client that supports enums. Take, for instance, this one change of a csharp client generated against this branch:

```diff
diff --git a/samples/client/petstore/csharp/OpenAPIClientNetStandard/src/Org.OpenAPITools/Model/EnumArrays.cs b/samples/client/petstore/csharp/OpenAPIClientNetStandard/src/Org.OpenAPITools/Model/EnumArrays.cs
index 4c0e96fd21..0f588412bb 100644
--- a/samples/client/petstore/csharp/OpenAPIClientNetStandard/src/Org.OpenAPITools/Model/EnumArrays.cs
+++ b/samples/client/petstore/csharp/OpenAPIClientNetStandard/src/Org.OpenAPITools/Model/EnumArrays.cs
@@ -28,69 +28,28 @@ namespace Org.OpenAPITools.Model
     [DataContract]
     public partial class EnumArrays :  IEquatable<EnumArrays>
     {
-        /// <summary>
-        /// Defines JustSymbol
-        /// </summary>
-        [JsonConverter(typeof(StringEnumConverter))]
-        public enum JustSymbolEnum
-        {
-            /// <summary>
-            /// Enum GreaterThanOrEqualTo for value: >=
-            /// </summary>
-            [EnumMember(Value = ">=")]
-            GreaterThanOrEqualTo = 1,
-
-            /// <summary>
-            /// Enum Dollar for value: $
-            /// </summary>
-            [EnumMember(Value = "$")]
-            Dollar = 2
-
-        }
-
         /// <summary>
         /// Gets or Sets JustSymbol
         /// </summary>
         [DataMember(Name="just_symbol", EmitDefaultValue=false)]
-        public JustSymbolEnum? JustSymbol { get; set; }
-        /// <summary>
-        /// Defines ArrayEnum
-        /// </summary>
-        [JsonConverter(typeof(StringEnumConverter))]
-        public enum ArrayEnumEnum
-        {
-            /// <summary>
-            /// Enum Fish for value: fish
-            /// </summary>
-            [EnumMember(Value = "fish")]
-            Fish = 1,
-
-            /// <summary>
-            /// Enum Crab for value: crab
-            /// </summary>
-            [EnumMember(Value = "crab")]
-            Crab = 2
-
-        }
-
-
-        /// <summary>
-        /// Gets or Sets ArrayEnum
-        /// </summary>
-        [DataMember(Name="array_enum", EmitDefaultValue=false)]
-        public List<ArrayEnumEnum> ArrayEnum { get; set; }
+        public EnumArraysJustSymbol? JustSymbol { get; set; }
         /// <summary>
         /// Initializes a new instance of the <see cref="EnumArrays" /> class.
         /// </summary>
         /// <param name="justSymbol">justSymbol.</param>
         /// <param name="arrayEnum">arrayEnum.</param>
-        public EnumArrays(JustSymbolEnum? justSymbol = default(JustSymbolEnum?), List<ArrayEnumEnum> arrayEnum = default(List<ArrayEnumEnum>))
+        public EnumArrays(EnumArraysJustSymbol justSymbol = default(EnumArraysJustSymbol), List<EnumArraysArrayEnumItems> arrayEnum = default(List<EnumArraysArrayEnumItems>))
         {
             this.JustSymbol = justSymbol;
             this.ArrayEnum = arrayEnum;
         }


+        /// <summary>
+        /// Gets or Sets ArrayEnum
+        /// </summary>
+        [DataMember(Name="array_enum", EmitDefaultValue=false)]
+        public List<EnumArraysArrayEnumItems> ArrayEnum { get; set; }

         /// <summary>
         /// Returns the string presentation of the object
```

Enums defined as inline are no longer inlined (see `JustSymbolEnum`). An object holding an array of enums don't make a lot of sense to me. For example, the spec document used in the C# generators has this object:

```
  EnumArrays:
    type: object
    properties:
      just_symbol:
        type: string
        enum:
          - ">="
          - "$"
      array_enum:
        type: array
        items:
          type: string
          enum:
            - fish
            - crab
```

As both an author and consumer of this document, I'd expect an object called `EnumArrays` which has two properties, one which is a normal enum and one which is an array of enums, where both of these enums are defined as nested types on `EnumArrays`. This is how the code previously generated. On this branch, these inlined enums are promoted to first-class models with odd prefix/suffixing of the names. I think for some languages this may make sense, but for other like C# it doesn't make sense and seems to reduce the quality of the generated code.

That said, I'm wondering how hard would it be to have this enum promotion logic as opt-in behind a generation flag?  Or, could it be added as another generation option (on CodegenConfig) which generators must override in order to opt-in to the model shuffling?

I believe @jmini has done a lot of work on the inline model resolver, and may have additional comments.

fantavlik:
@jimschubert I see what you mean, including the inline enum within the C# model files was intentional.

What I've observed is that many (most?) languages don't have separate templating support for inline enums, they work for enums that have been promoted to the `#/components/schemas` section and generate nothing for inline enums.

I think it would be pretty simple to add a flag (and perhaps adjust the default behavior for C# and Typescript and other languages that already support inline enums) such that enums are not promoted to their own schema under `#/components/schemas`, something like `--resolveInlineEnums=false` is that something that would address this issue for you?

zzzuzik:
was this PR abandon?

kurtmc:
Is there a plan to merge this? Currently the generated go code is quite broken on v4.2.2 if you use anyOf/oneOf.

fantavlik:
I'm still around to address feedback - sorry this has been lingering so long. @jimschubert if I were to add a generation flag e.g. `--resolveInlineEnums=false` that was `false` by default for languages like C# that support that in templating (avoiding breaking changes) would that resolve your concerns? I think it's doable.

jimschubert:
@fantavlik I believe what you're asking is what I suggested above. Although, I'd suggest defaulting to false always, and having users opt in. I think the only time it would be true is for a generator where it just doesn't make sense in the language for this option to be false (as you've proposed for go).

To put it another way, as in my example in my previous comment, we should try to maintain user expectations about code structure by matching the spec document as closely as possible. That should be the default behavior, and any deviation would be an exclusion preferably with a warning log so the user knows that we've adjusted from expectations. See LOGGER.warn calls in ModelUtils for example https://github.com/OpenAPITools/openapi-generator/blob/2d24d42e652e1abeb979a89d50235c383c12bf44/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L799.

fantavlik:
New Codegen option in CodegenConfig: `resolveInlineEnums` - `false` by default for all generators other than Go.

Sample WARN output as you suggested @jimschubert when this option is true and an inline enum is promoted:
```
$ java -jar openapi-generator-cli.jar generate -g go -i modules/openapi-generator/sr
c/test/resources/3_0/inline_model_resolver.yaml
...
[main] WARN  o.o.codegen.InlineModelResolver - Model CatMyHabitatAllOfContinent promoted to its own schema due to resolveInlineEnums=true
[main] WARN  o.o.codegen.InlineModelResolver - Model CatCuteness promoted to its own schema due to resolveInlineEnums=true
[main] WARN  o.o.codegen.InlineModelResolver - Model CatPreferencesMetadata promoted to its own schema due to resolveInlineEnums=true
```
I also fixed some behavior such that schemas with titles will use that title regardless of how nested they are within another model which seemed more consistent with previous behavior.

@jimschubert let me know if you have any other feedback.

JSarvesh:
Hi Is this PR merged in  master as I am facing the same issue with openapi-generator 4.2.3 where the anyof, allof enums are generated as empty structs.

AnthonySendra:
Hello,

Any update on this? I'm stuck with this issue with last version 4.3.0.
Thanks!

jimschubert:
@fantavlik sorry I never saw your fixes or more recent comments. I've been a little MIA since the start of the pandemic as well, and it's taking me forever to get caught back up. I apologize. If that ever happens again, please feel free to ping the core team to finish up any review or ping me directly on slack.

I was trying to evaluate this PR, and worked on getting a master merged squared up but I had to fix a couple tests and modify the suffixes applied in your branch from title cased (`AllOf`) to hungarian notation (`_allOf`) to match master and avoid breaking changes. I then have two NPE in DefaultCodegenTest for tests:

* testUseOneOfInterfaces
* testFormParameterHasDefaultValue

And a tests failure in the same for:

* testGetSchemaTypeWithComposedSchemaWithOneOf

I didn't see where any merge conflict resolutions I had made would have caused this. I was wondering if you could take a look? https://github.com/jimschubert/openapi-generator/tree/fantavlik-inline-resolver

fantavlik:
@jimschubert thanks so much for taking a look - I'm in a similar MIA situation but I hope to find some nights/weekends to work through the issues you've spotted as soon as I'm able.

fantapop:
@fantavlik I checked out your branch to see if would get me any closer to fixing #5202 . At first glance it appears that this is what's needed. When I run the generation with 
`-DdebugSupportingFiles` I can see all the data required for processing the inline types which aren't being generated right.  The full generation of the types is, however, failing with a `NullPointerException` so it never finishes.  I'm not sure if its related to something that was already discussed above or not.  The exception is here:
```
  Exception: null
    at org.openapitools.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:1078)
    at org.openapitools.codegen.DefaultGenerator.processPaths(DefaultGenerator.java:969)
    at org.openapitools.codegen.DefaultGenerator.generateApis(DefaultGenerator.java:528)
    at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:922)
    at org.openapitools.codegen.cmd.Generate.run(Generate.java:396)
    at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:60)
Caused by: java.lang.NullPointerException
    at org.openapitools.codegen.DefaultGenerator.cloneOAuthFlow(DefaultGenerator.java:1265)
    at org.openapitools.codegen.DefaultGenerator.getAuthMethods(DefaultGenerator.java:1232)
    at org.openapitools.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:1054)
    ... 5 more
```

Give me some time tomorrow evening and I'll try to put up a minimal spec to reproduce this issue.

fantavlik:
Awesome thanks for that @fantapop -- I plan to spend some time in the next week to work through any other issues with this PR so good timing.

Also nice username, who knew there were >1 `fant*` users in the Seattle area working on OpenAPI? ;)

fantapop:
Haha, yes! I noticed your username as well.  I'm currently bunkering down in
Florida with my mother in law but we'll need to be returning to Seattle at some
point when things start opening up again.  Thank you for your continued work on
this.

I dug a bit more, the NullPointerException that I was experiencing is because
I'm using a security scope without declaring it inside the top level security
definition. I didn't realize the framework we're using to generate our openapi
spec  is not adding this.

After declaring the scope the generation is finishing. I don't, however, see
any of the additional models being created which map to the inline types.  Will
I need to augment the template in some way to include the new models in model
creation?  I was expecting them to be added to the list of models that is
getting iterated over at the top level.

Another thing that I noticed is that in my example an enum which is part of an
anyOf is not listed in generated data type.

I've attempted to include a minimal spec here for you to use to see these
issues.  Let me know if you have any problems with it.

```
---
components:
  examples: {}
  headers: {}
  parameters: {}
  requestBodies: {}
  responses: {}
  schemas:
    Email:
      type: string
      pattern: "^.*?@.*?$"
    Uuid:
      type: string
      pattern: "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-4[0-9a-fA-F]{3}-[89ABab][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$"
    PhoneNumber:
      type: string
      pattern: "^\\+[0-9]+$"
    User:
      properties:
        id:
          "$ref": "#/components/schemas/Uuid"
        email:
          "$ref": "#/components/schemas/Email"
        firstName:
          type: string
        lastName:
          type: string
        phoneNumber:
          "$ref": "#/components/schemas/PhoneNumber"
      required:
      - id
      type: object
      additionalProperties: true
    LookupIdentifier:
      oneOf:
      - "$ref": "#/components/schemas/Uuid"
      - "$ref": "#/components/schemas/Email"
      - "$ref": "#/components/schemas/PhoneNumber"
      - type: string
        enum:
        - me
        nullable: false
openapi: 3.0.0
paths:
  "/users/v1/{id}":
    get:
      operationId: getUserById
      responses:
        '200':
          content:
            application/json:
              schema:
                "$ref": "#/components/schemas/User"
          description: Ok
      parameters:
      - in: path
        name: id
        required: true
        schema:
          "$ref": "#/components/schemas/LookupIdentifier"
```

fantavlik:
OK @jimschubert I think I've fixed the three tests you called out and pushed your master merge into this PR. The test failures were all related to references to inline models that had been promoted to their own `#/components/schemas` models which is expected. Let me know what else I can do to get this to be merge-ready.

fantavlik:
Just merged master into here and resolved conflicts as well, mostly within InlineModelResolver.java -- I resolved by using this PR's version for that file but I'll look at that more closely now in case other changes need to be incorporated.

fantavlik:
I'm pretty sure I clobbered some of these changes: https://github.com/OpenAPITools/openapi-generator/pull/5290 though I don't see any test failures. I might need some help folding those changes into this branch.

fantavlik:
Hey @fantapop I just took a look at your spec. Re: `Email`, `Uuid`, and `PhoneNumber` since they are all strings they are not treated as inline models since they are going to be represented as strings regardless of their differing `pattern`s. That was the case before this PR and I'm not attempting to change that behavior here.

The inline enum in `LookupIdentifier` however definitely should be generating a separate model so I'm looking into that now, thanks for the heads up.

fantavlik:
Ah, so the inline enum is properly being split out as `model_lookup_identifier_one_of.go` (see contents below) but the Go generation code doesn't know how to handle a oneOf containing a string and a string enum (it should probably just default to a string). Again sorry but I'm not trying to solve that this PR, I'm just trying to ensure that any inline models are resolved into $refs so that they're generated separately. 
```go
/*
 */

package openapi
// LookupIdentifierOneOf the model 'LookupIdentifierOneOf'
type LookupIdentifierOneOf string

// List of LookupIdentifier_oneOf
const (
	LOOKUP_IDENTIFIER_ONE_OF_ME LookupIdentifierOneOf = "me"
)
```

fantapop:
I went back to look again after you merged master in and I realized that I was looking at the wrong code.  I mistakenly was looking at master and not at your actual branch.  I retested with the inline-resolver branch.  Things seem to be working now with the exception of the issues you've mentioned above.   Thanks for continuing to push on this!

fantavlik:
@jimschubert I believe I've addressed the tests you called out and everything is passing locally, I think the CI issues are possibly intermittent but I don't have permission to restart those jobs. I know you have other commitments so take a look at your convenience.

fantapop:
Hey @fantavlik I think I've identified an additional issue.   I have a member of a schema object that looks like this:

```json
"country":{
    "nullable": true,
    "allOf":[
        {
            "$ref":"#/components/schemas/Country"
        }
    ]
},
```

Country is an Enum, something like this:

```json
"Country":{
    "description":"Country",
    "enum":[
        "US",
        "AF",
        ....
    ],
    "type": "string"
}
```

On your branch here its generating a new type to encapsulate the `allOf` which is expected. The interface ends up looking something like this in the typescript-fetch template:
```typescript
export interface UpdateParams {
    country?: UpdateParamsCountry | null;
    // ...other stuff
```
And then inside UpdateParamsCountry:
```typescript
export interface UpdateParamsCountry extends Country { // extends?
    // empty because allOf array isn't being handled
}
```

partially, this type is just wrong because typescript-fetch isn't supporting that anyOf array which is now being generated in your branch.  However, the part that seems wrong in the core here is that the type `UpdateParamsCountry` has the parent and parentSchema field set to `Country`. This doesn't happen when there are greater than one elements in the `allOf`.  It also doesn't happen for `oneOf` and `anyOf` at all.

The typescript-fetch template here is broken because the value of parent is an enum and interfaces can't extend enums.  Here is the template generating that code.

```mustache
export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
{{#additionalPropertiesType}}
    [key: string]: {{{additionalPropertiesType}}}{{#hasVars}} | any{{/hasVars}};
{{/additionalPropertiesType}}
{{#vars}}
    /**
     * {{#lambda.indented_star_4}}{{{unescapedDescription}}}{{/lambda.indented_star_4}}
     * @type {{=<% %>=}}{<%&datatype%>}<%={{ }}=%>
     * @memberof {{classname}}
     */
    {{#isReadOnly}}readonly {{/isReadOnly}}{{name}}{{^required}}?{{/required}}: {{#isEnum}}{{{datatypeWithEnum}}}{{/isEnum}}{{^isEnum}}{{{datatype}}}{{#isNullable}} | null{{/isNullable}}{{/isEnum}};
{{/vars}}
}
```

And here is the debug log containing the values that the template is using:

```
{
    "importPath" : "models.UpdateParamsCountry",
    "hasImports" : true,
    "model" : {
      "parent" : "Country",
      "parentSchema" : "Country",
      "interfaces" : [ "Country" ],
      "anyOf" : [ ],
      "oneOf" : [ ],
      "allOf" : [ "Country" ],
      "name" : "UpdateParamsCountry",
      "classname" : "UpdateParamsCountry",
      "classVarName" : "updateParamsCountry",
      "modelJson" : "{\n  \"nullable\" : true,\n  \"allOf\" : [ {\n    \"$ref\" : \"#/components/schemas/Country\"\n  } ]\n}",
      "classFilename" : "UpdateParamsCountry",
      "isAlias" : false,
      "isString" : false,
      "isInteger" : false,
      "isLong" : false,
      "isNumber" : false,
      "isNumeric" : false,
      "isFloat" : false,
      "isDouble" : false,
      "vars" : [ ],
      "allVars" : [ ],
      "requiredVars" : [ ],
      "optionalVars" : [ ],
      "readOnlyVars" : [ ],
      "readWriteVars" : [ ],
      "parentVars" : [ ],
      "mandatory" : [ ],
      "allMandatory" : [ ],
      "imports" : [ "Country" ],
      "hasVars" : false,
      "emptyVars" : true,
      "hasMoreModels" : true,
      "hasEnums" : false,
      "isEnum" : false,
      "isNullable" : false,
      "hasRequired" : false,
      "hasOptional" : false,
      "isArrayModel" : false,
      "hasChildren" : false,
      "isMapModel" : false,
      "isDeprecated" : false,
      "hasOnlyReadOnly" : true,
      "vendorExtensions" : { },
      "uniqueItems" : false,
      "exclusiveMinimum" : false,
      "exclusiveMaximum" : false
    }
  }
```

On a related note, it would be nice to have a optimization where if the `allOf` `anyOf` or `oneOf` only include a single type in their arrays, that the extra level of indirection is removed.

jimschubert:
@fantavlik hey, I've tried a few times to merge master into this branch so we could get it merged, but the changes are pretty intense. I'll try to take another look at it again over the weekend.

fantavlik:
Thanks @jimschubert let me know if I can help!

jimschubert:
@fantavlik I've spent a few more hours trying to get this rebased with current master, and I'm having no luck (probably have spent 12-15 hours total trying to sync this PR with current master changes). Any time I try rebasing, it gets unwieldy due to the reorganization of methods either in your branch or in master. I've also tried merging up master, resetting InlineModelResolver to just your changes and then incorporating any changes made on master since the last master integration, but whenever I do this it doesn't seem to match with your expectations in this branch as it breaks about 30 tests.

I'm really interested in getting the `gatherInlineModels` work into 5.0 so we can have a more fully synchronized object structure up front.

I know it's a lot to ask, but do you think you could sync master and incorporate the composed schema flattening stuff from master into your work?

fantavlik:
@jimschubert I really appreciate you trying to get this in for v5 -- let me spend some time this weekend. It should hopefully be a bit easier for me to resolve these conflicts and tests since I have the context.

fantavlik:
@jimschubert I believe I've resolved the merge conflicts, regenerated samples, and fixed the tests within `modules/openapi-generator/src/test/...`. I think there are still some failures in CI, I can't easily tell what's going on but if there are additional tests failing I can take an attempt at fixing those if you don't mind pointing me to the failures.

jimschubert:
Rebuilding this branch after the latest changes, I'm seeing that `discriminator` is being evaluated at the lowest level of a composed schema rather than the highest as it was previously. I'm wondering if @spacether or @sebastien-rosset might be able to take a look and offer an opinion about how to fix the issue on the branch?

We have `modules/openapi-generator/src/test/resources/3_0/composed-schemas.yaml` defined as (irrelevant parts removed):

```
openapi: 3.0.1
info:
  version: 1.0.0
  title: Example
  license:
    name: MIT
servers:
  - url: http://api.example.xyz/v1
paths:
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                # This field will not match to any type.
                - description: Any kind of pet
              discriminator:
                propertyName: pet_type
      responses:
        '200':
          description: Updated

components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
    Dog:
      allOf:
        # This field will not match to any type.
        - description: Dog information
        - type: object
          properties:
            bark:
              type: boolean
            breed:
              type: string
              enum: [Dingo, Husky, Retriever, Shepherd]
    Cat:
      allOf:
        - type: object
          properties:
            hunts:
              type: boolean
            age:
              type: integer
```

On master, the discriminator is evaluated on the `oneOf` and ignores the `allOf` composition on Cat and Dog. On this branch, it's evaluated at the `allOf` level.

The specification states:

> When using the discriminator, inline schemas will not be considered.

But this is apparently limited only to the schema _at the same node_ where the discriminator is defined as the example in this typescript generator is very similar to that of the spec. The spec defines `discriminator` on Pet and has inline schemas in Dog and Cat, so the exception thrown by our DefaultCodegen doesn't seem correct for anything other than discriminator at the same level as allOf. The error is thrown on line 2816 of DefaultCodegen in this PR.

image

sebastien-rosset:
> Rebuilding this branch after the latest changes, I'm seeing that `discriminator` is being evaluated at the lowest level of a composed schema rather than the highest as it was previously. I'm wondering if @spacether or @sebastien-rosset might be able to take a look and offer an opinion about how to fix the issue on the branch?
> 

Just to understand the example, in the full yaml document, is there only one `discriminator` keyword? Or in the above example you have removed some lines to simplify the example? Also, presumably in the actual model, `pet_type` is defined as a property of `Pet`, right?

sebastien-rosset:
> 
> The specification states:
> 
> > When using the discriminator, inline schemas will not be considered.
> 
> But this is apparently limited only to the schema _at the same node_ where the discriminator is defined.

This sentence from the spec is very ambiguous. Do you have a clear interpretation of what it means? Do you have a pointer to an issue/document that clarifies its meaning?

sebastien-rosset:
@fantavlik , @jimschubert this PR seems to introduce changes that break backward compatibility. For example:

> Prefix all enum values with model name, this is needed to avoid collisions within the package

This would be highly problematic because existing clients that have written code make assumption about how the client code has been generated. IMO, it would be better if there was a CLI option to control the prefix such that it does not break backward compatibility for existing clients.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -236,6 +236,8 @@ apiTemplateFiles are for API outputs only (controllers/handlers).
     // flag to indicate whether enum value prefixes are removed
     protected boolean removeEnumValuePrefix = true;
 
+    // set to true to promote inline enums to their own separate models
+    protected boolean resolveInlineEnums = false;
```
spacether, line 240: We also have the ModelUtils.unaliasScema that decides whether or not a model is made.
How about naming this resolveInlineSchemasAsModels.
Doesn't it apply to more than just enums?

jimschubert:
@sebastien-rosset 

> Just to understand the example, in the full yaml document, is there only one discriminator keyword? Or in the above example you have removed some lines to simplify the example? Also, presumably in the actual model, pet_type is defined as a property of Pet, right?

I'd removed only schemas and operations that had nothing to do with the example. `pet_type` is presented exactly as it is in the example spec used by the Typescript axios sample.

Regarding the ambiguous line from the spec, I don't have a clear disambiguation link. Discriminator is a known source of ambiguity in the spec, and I think the spec authors know and acknowledge that.

I agree with your evaluation of the prefixing. We should probably have an option to complement the model and API prefix and suffix options, then update the 14 existing locations with the Enum suffix to use the option.

sebastien-rosset:
> @sebastien-rosset
> 
> > Just to understand the example, in the full yaml document, is there only one discriminator keyword? Or in the above example you have removed some lines to simplify the example? Also, presumably in the actual model, pet_type is defined as a property of Pet, right?
> 
> I'd removed only schemas and operations that had nothing to do with the example. `pet_type` is presented exactly as it is in the example spec used by the Typescript axios sample.
> 
> Regarding the ambiguous line from the spec, I don't have a clear disambiguation link. Discriminator is a known source of ambiguity in the spec, and I think the spec authors know and acknowledge that.
> 
> I agree with your evaluation of the prefixing. We should probably have an option to complement the model and API prefix and suffix options, then update the 14 existing locations with the Enum suffix to use the option.

I'm surprised we don't generate a validation failure. Shouldn't the `pet_type` property be declared explicitly as shown below? This is the first time I am seeing a discriminator without the property being defined explicitly. I need to check what that does to the codegen logic.
```yaml
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
         pet_type:
           type: string
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
fantavlik: The specific request from previous feedback was to treat enums differently here than other models, this flag actually only applies to enums not other models as the name implies. See: https://github.com/OpenAPITools/openapi-generator/pull/3162#issuecomment-520193930

fantavlik:
> I agree with your evaluation of the prefixing. We should probably have an option to complement the model and API prefix and suffix options, then update the 14 existing locations with the Enum suffix to use the option.

I'll get going on this, it's most important for Go because public variables names are shared throughout the entire package so collision is much more of a problem than, say, namespacing within a single file

> In the PR description, it would be good to clarify which part of "go" is impacted. There are multiple go generators, go server, go client, go experimental client...

Changes were made in AbstractGoCodegen.java so all go generation is impacted, I'll clarify that in the PR description.

> I'm surprised we don't generate a validation failure with this OAI doc. Shouldn't the pet_type property be declared explicitly as shown below? This is the first time I am seeing a discriminator without the property being defined explicitly.

Agree with this, I wouldn't expect the "deep" property within a nested oneOf / allOf to be valid based on the behavior with `required` (although I realize that `required` is part of JSON Schema whereas `discriminator` is not...)

fantavlik:
@jimschubert I added a config option `legacyGoEnumNaming` which if set to true uses the original enum prefixing before this PR, so when running 
```
$ java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g go -o testgo1 --additional-properties=legacyGoEnumNaming=true -i ../openapi-generator/src/test/resources/3_0/inline_model_resolver.yaml --skip-validate-spec
```
You will now see
```go
// from model_cal_my_habitat_all_of_continent.go

// CatMyHabitatAllOfContinent the model 'CatMyHabitatAllOfContinent'
type CatMyHabitatAllOfContinent string

// List of CatMyHabitatAllOfContinent
const (
	AFRICA CatMyHabitatAllOfContinent = "Africa"
	ANTARCTICA CatMyHabitatAllOfContinent = "Antarctica"
	ASIA CatMyHabitatAllOfContinent = "Asia"
	EUROPE CatMyHabitatAllOfContinent = "Europe"
	NORTH_AMERICA CatMyHabitatAllOfContinent = "North America"
	OCEANIA CatMyHabitatAllOfContinent = "Oceania"
	SOUTH_AMERICA CatMyHabitatAllOfContinent = "South America"
)
```
Rather than the new default:
```go
// from model_cal_my_habitat_all_of_continent.go

// CatMyHabitatAllOfContinent the model 'CatMyHabitatAllOfContinent'
type CatMyHabitatAllOfContinent string

// List of CatMyHabitatAllOfContinent
const (
	CAT_MY_HABITAT_ALL_OF_CONTINENT_AFRICA CatMyHabitatAllOfContinent = "Africa"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_ANTARCTICA CatMyHabitatAllOfContinent = "Antarctica"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_ASIA CatMyHabitatAllOfContinent = "Asia"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_EUROPE CatMyHabitatAllOfContinent = "Europe"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_NORTH_AMERICA CatMyHabitatAllOfContinent = "North America"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_OCEANIA CatMyHabitatAllOfContinent = "Oceania"
	CAT_MY_HABITAT_ALL_OF_CONTINENT_SOUTH_AMERICA CatMyHabitatAllOfContinent = "South America"
)
```

I hope that addresses the comments re: prefixes

fantavlik:
Also all-caps naming (even for `const`s) is definitely an anti-pattern in Go but that's for another PR 😉

jimschubert:
@fantavlik I think `legacyGoEnumNaming` is a good approach for now.

I was trying to evaluate your branch against latest master locally, but I'm seeing some failures in `SharedTypeScriptTest.java`. I'd have to spend some more time looking into this as I would guess that's why CI on this PR are all failing.

Maybe someone else on @OpenAPITools/generator-core-team could also re-review the PR, and maybe help look at the test failure?

docs/generators/go.md
```diff
@@ -8,6 +8,7 @@ sidebar_label: go
 |enumClassPrefix|Prefix enum with class name| |false|
 |hideGenerationTimestamp|Hides the generation timestamp when files are generated.| |true|
 |isGoSubmodule|whether the generated Go module is a submodule| |false|
+|legacyGoEnumNaming|whether to use legacy enum prefixes and suffixes which are shorter but subject to namespace collisions| |false|
```
wing328, line 11: @fantavlik thanks for the PR. How's this option different from the`enumClassPrefix` option?

fantavlik: Ah, good catch I think `enumClassPrefix` must have come along while this PR was in flight. I will remove `legacyGoEnumNaming` and rework my logic to obey the existing `enumClassPrefix`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -104,95 +105,220 @@ private void flattenPaths(OpenAPI openAPI) {
...
+            return;
+        } else if (schema.getAdditionalProperties() != null) {
+            // If non-object type is specified but also additionalProperties
+            LOGGER.error("Illegal schema found with non-object type combined with" +
+                    " additionalProperties, no additionalProperties should be defined:\n " +
+                    schema.toString());
+            return;
+        }
+        // Check array items
+        if (schema instanceof ArraySchema) {
```
spacether, line 201: Arraymodels are generated in some generators. Are we covering that use case?

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -104,95 +105,220 @@ private void flattenPaths(OpenAPI openAPI) {
...
+     * Return true if a model should be generated e.g. object with properties,
+     * enum, oneOf, allOf, anyOf, etc.
+     *
+     * @param schema target schema
      */
-    private void flattenRequestBody(OpenAPI openAPI, String pathname, Operation operation) {
-        RequestBody requestBody = operation.getRequestBody();
-        if (requestBody == null) {
-            return;
+    private boolean isModelNeeded(Schema schema) {
```
spacether, line 116: Would is be simpler to make a ref to the current schema, unalias the ref schema, and return True if the unaliased schema still has a ref in it (meaning that a model would be made?
For example:
```
Schema refSchema = new Schema();
refSchema.set$ref("#/components/schemas/"+name);
Schema unaliasedSchema = config.unaliasSchema(refSchema, config.importMapping());
if (unaliasedSchema.get$ref() != null) {
    return true;
}
return false;
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
fantavlik: The main affect of this PR for Array models would be to promote consistency such that the same models get generated regardless of whether they appear via inline or as a $ref
e.g. making sure this inline spec
```yaml
        kittens:
          type: array
          items: # inline array where the items are enums
            type: string
            enum:
              - Kitten1
              - Kitten2
              - Kitten3
```
generates the same as
```yaml
        kittens:
          type: array
          items:
            $ref: '#/components/schemas/Cat'
```

In general the ArrayModels should be handled the same after this PR as before it unless there was particular logic to handle complex inline array models such as the above (I have seen languages like Typescript handle inline enum generation using complex template logic for example that should no longer be needed).

If you have specific languages in mind that you're concerned about for backwards compatibility I'm happy to take a look. The most likely impact would be a change to the name of the inline model to be more specific e.g. `OASInlineObject1` -> `OASUploadFileBody` or completely ungenerated models now being generated.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
fantavlik: There is [an instance](https://github.com/OpenAPITools/openapi-generator/blob/5bd338f8ee3f22983d66b9f5fbca6699987043e5/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java#L152) where we don't actually want to unalias the schema but just need to check whether a complex model exists so I'd like to re-use that logic and also I'm not a fan of functions that do unintuitive/unexpected things like in this case checking whether a model is needed AND unaliasing if so. It would save 4 lines of code but doesn't make it "simpler" in my opinion.

docs/generators/go.md
fantavlik: Done, I pulled the use of `enumClassPrefix` from templates and into `AbstractGoCodegen.java` here: https://github.com/OpenAPITools/openapi-generator/pull/3162/commits/f9d8093992049b7c021a8be62c588b388e795720

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
spacether: Thank you for pointing out that use case.
In that case you could call unaliasedSchema directly because the source schema already has a ref. Then if the unaliased schema had a ref in it, we know that the model would be produced.

> we don't actually want to unalias the schema but just need to check whether a complex model exists

I agree. But we can use the results of the unaliased schema to tell if a model will be made.
A comment could explain why we are using unaliasSchema to tell if a model is going to be made, making this a more intuitive usage.
My suggestion was not to just replace one block with my pseudocode above, it was to replace the entire body of isModelNeeded with that block. This would reduce the size of the method by 16 lines. It's you call though.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -104,95 +105,220 @@ private void flattenPaths(OpenAPI openAPI) {
...
+            // any to catch OpenAPI violations
+            if (isModelNeeded(schema) || "object".equals(schema.getType()) ||
+                    schema.getProperties() != null || schema.getAdditionalProperties() != null ||
+                    schema instanceof ComposedSchema) {
+                LOGGER.error("Illegal schema found with $ref combined with other properties," +
+                        " no properties should be defined alongside a $ref:\n " + schema.toString());
+            }
+            return;
+        }
+        if (schema.getType() == null || "object".equals(schema.getType())) {
```
spacether, line 160: Why are we handling null type in here?
Isn't null type an AnyType schema?
Or is this detecting composed schemas with properties?
How about adding comments describing these use cases here?

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -104,95 +105,220 @@ private void flattenPaths(OpenAPI openAPI) {
...
+                    // Recurse to create $refs for inner models
+                    gatherInlineModels(inner, modelPrefix + "OneOf");
+                    if (isModelNeeded(inner)) {
+                        Schema refSchema = this.makeSchemaResolve(modelPrefix, "OneOf", inner);
+                        newOneOf.add(refSchema); // replace with ref
+                    } else {
+                        newOneOf.add(inner);
+                    }
+                }
+                m.setOneOf(newOneOf);
```
spacether, line 260: This looks good. If you wanted you could shorten this by:
- making a list of allOf, anyOf, and oneOf values.
Loop though the list of lists and handle each case accordingly.
Collapse this similar code from being in 3 places to one place.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
spacether: Thank you for explaining this

fantavlik:
Still facing generation issues with `typescript-axios-composed-schemas.yaml`

After some experimenting I think I've proven that there is something wrong with the logic, it's throwing this error 
```
Exception in thread "main" java.lang.RuntimeException: Could not process model 'InlineBody'.Please make sure that your schema is correct!
        at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:475)
        at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:851)
Caused by: java.lang.RuntimeException: Invalid inline schema defined in allOf in 'InlineBody1File'. Per the OpenApi spec, for this case when a composed schema defines a discriminator, the allOf schemas must use $ref. Change this inline definition to a $ref definition
        at org.openapitools.codegen.DefaultCodegen.getAllOfDescendants(DefaultCodegen.java:2816)
        at org.openapitools.codegen.DefaultCodegen.createDiscriminator(DefaultCodegen.java:2873)
        at org.openapitools.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:2234)
        at org.openapitools.codegen.DefaultGenerator.processModels(DefaultGenerator.java:1128)
        at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:470)
        ... 4 more
```
With the following spec:
```yaml
openapi: 3.0.1
info:
  version: 1.0.0
  title: Example
  license:
    name: MIT
servers:
  - url: http://api.example.xyz/v1
paths:
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                # This field will not match to any type.
                - description: Any kind of pet
              discriminator:
                propertyName: pet_type
      responses:
        '200':
          description: Updated

  /file:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                file:
                  allOf:
                    - type: file
                    # This field will not match to any type.
                    - description: The file to upload
      responses:
        '200':
          description: File uploaded

components:
  schemas:
    Pet:
      type: object
      properties:
        pet_type:
          type: string
          enum: [Cat, Dog]
      required:
        - pet_type
    Dog:
      allOf:
        - type: object
          properties:
            bark:
              type: boolean
            breed:
              type: string
              enum: [Dingo, Husky, Retriever, Shepherd]
    Cat:
      allOf:
        - type: object
          properties:
            hunts:
              type: boolean
            age:
              type: integer
    PetByAge:
      type: object
      properties:
        age:
          type: integer
        nickname:
          type: string
      required:
        - age

    PetByType:
      type: object
      properties:
        pet_type:
          type: string
          enum: [Cat, Dog]
        hunts:
          type: boolean
      required:
        - pet_type
```
If I remove the `file/` path the spec generates without any errors, which doesn't make sense since the file/ endpoint doesn't reference any discriminator models.

fantavlik:
I think I'm quite close now, but running into https://github.com/OpenAPITools/openapi-generator/issues/3870 in the python-aiohttp sample tests now. I'll try to workaround that.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
fantavlik: Good point, added this comment let me know if that's not clear:
```java
        // Check object models / any type models / composed models for properties,
        // if the schema has a type defined that is not "object" it should not define
        // any properties
```

fantavlik:
Well I've slogged through python and php test issues unrelated to my changes and now running into ruby issues. I need to stop for now, the thing about making changes that produce loads of new models that should have been generated before is that it uncovers a bunch of previously unseen issues in a bunch of languages.

fantavlik:
I fixed the issues with Ruby samples, the samples were just out of date and referencing deleted files.

fantavlik:
Moving on to Go samples failures next. I feel like I'm close? But I also feel like every day an eagle swoops down and eats my liver.

wing328:
@fantavlik thanks for leading the effort to improve the inline model resolver. How urgent do you need this change to be included in the next release v5.0.1 (we'll release v5.0.0 within hours)?

If it's not urgent, may I suggest putting this PR on hold for the time being while the core team performs another review in early or mid Jan?

fantavlik:
> @fantavlik thanks for leading the effort to improve the inline model resolver. How urgent do you need this change to be included in the next release v5.0.1 (we'll release v5.0.0 within hours)?
> 
> If it's not urgent, may I suggest putting this PR on hold for the time being while the core team performs another review in early or mid Jan?

I have been dragging my feet on this but I finally have some more time now and moving forward, I am happy to wait to January.

wing328:
👌 thanks again for the PR. Happy holidays!

fantavlik:
Same to you William! Looking forward to finishing this up in the new year.

kurtmc:
@fantavlik Thank you for working on this, I have been following this PR for a while and I am keen to see it get merged.

Please let me know if there is anything I can do to help. Happy to test out your branch against a bunch of openapi specs that are used in production.

richardwhiuk:
Same, keen to see this landed - Rust doesn't support anonymous, or inline enums. Let me know if you need any help getting this landed.

lesam:
If/when this is landed, will there be a flag to turn off strict validation of enum values for languages where they are not currently validated? Otherwise this seems like a potentially breaking change.

longquanzheng:
Hi, is there any updates on this PR? thanks for the hard work, this is really important for me.

codyolsen:
It looks like this issue has been a can kicked down the road for the last 4 years. @fantavlik you've put a lot of work into this but it looks like the tests are still a blocking issue. Are you still actively invested in this PR? 

It looks like from the history on this issue the main problem has been fragmented test failures with the different languages. Would it help if we could split the PR into the different language slices so we can piece meal merge the support in for each language and let contributors with those language needs/expertise help with the pertinent slices? Perhaps that would help avoid letting perfection become the enemy of progress?

fantavlik:
Hey all, glad there's is still interest in this and sorry I abandoned this 3 years ago - I'm in a better place now personally to try to get this merged but want to make sure @wing328 and any other maintainers still support moving forward with this before I dig in again.

@codyolsen I had hoped to originally do what you suggest and break the PR apart by language but InlineModelResolver.java affects the inputs to all language templates. I guess one option to leave the existing InlineModelResolver.java code in place feature flag the refactored resolver. @wing328 let me know if that's something you are open to.

wing328:
Hi all, we've merged a few PRs to improve inline schema handling: https://github.com/OpenAPITools/openapi-generator/pulls?q=is%3Apr+inline+schema+is%3Aclosed+author%3Awing328

Doc: https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md#inline-schema-naming

Thanks @fantavlik for this PR as a staring point 🙏 

We've still haven't added an option to generate all enum as models. Hopefully it will be something we can support in 7.x or later.

longquanzheng:
Hi all, 
I actually worked around the issue by not using inline enums later so it’s not blocking me any more . But still very appreciated you all trying to improve it.

https://github.com/OpenAPITools/openapi-generator/pull/2897#issuecomment-1358839510

codyolsen:
@wing328 That's great there has been some progress on inline schema handling! Is there a way that we can merge in what @fantavlik has done on this CR to compliment the changes?

codyolsen:
@longquanzheng That works if you have control of the model, but if it's a 3rd party schema you can't control then you are out of luck.

wing328:
> We still haven't added an option to generate all enum as models. Hopefully it will be something we can support in 7.x or later.

that's done in latest master (upcoming v7.0.0 release)

wing328:
closing this one as pretty much all enhancements are added to the master

thanks again for this PR as a starting point.

Issue #2923, opened 2019-05-17
[BUG][ASPNETCORE] isEnum behaves unexpected since 4.0.0-beta3


mboogerd:
##### Description
I am using a custom mustache template to generate an alternative controller whose operations use _nullable_ enums for _optional_ enum query parameters. This has been working fine until 4.0.0 beta2, but has started breaking from beta3 (and is still broken in the 4.0.0 release). As far as I have narrowed down the issue, the `{{isEnum}}` expression no longer evaluates to `true`. This happens at least when used in the context of a query parameter, but still works fine for models. From the code differences between the two releases I could not find out why it has stopped working for (query) parameters

As a result of the above, my custom mustache template `dataTypeNullable`:
`{{&dataType}}{{#isEnum}}{{^required}}?{{/required}}{{/isEnum}}`

no longer functions as desired when used in the context of generating operation parameters for my controllers. If this isn't a bug, it would be much appreciated if someone could point me in the direction of a fix!

##### openapi-generator version

OK <= 4.0.0-beta2
NOK > 4.0.0-beta2

##### Minimal OpenAPI declaration to reproduce

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: "Example API"
servers:
  - url: http://example/
paths:
  /:
    get:
      parameters:
        - in: query
          name: param
          schema:
            $ref: "#/components/schemas/SomeEnum"
          required: false
      operationId: exampleOperation
      responses:
        default:
          description: "not implemented error"
components:
  schemas:
    SomeEnum:
      type: string
      enum:
        - item1
        - item2

##### Command line used for generation

We have been using the generator as part of a multi-stage docker build, as follows (although I am pretty sure this doesn't matter).
```
FROM openapitools/openapi-generator-cli:v4.0.0-beta2 AS generator

WORKDIR /local
COPY . .

# we know that this file docker-entrypoint.sh is there from the original openapi docker image's Dockerfile
RUN /usr/local/bin/docker-entrypoint.sh generate \
    -DpackageName=TestService \
    -DdebugOperations=true \
    -i test-service.yaml \
    -g aspnetcore \
    -o out/
```

##### Steps to reproduce

- Run the 4.0.0-beta2 generator with `debugOperations` set to true and capture the debug log
- Run the 4.0.0-beta3 generator with `debugOperations` set to true and capture the debug log
[...]

Issue #2645, opened 2019-04-11
[BUG] Property isEnum false for referenced ENUM schemas


DonDi94:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

When generating the code of the Erlang server, I have noticed that the ENUM check would not work on one parameter. By reducing the file I found out that the generator wasn't putting the enum values in the generated code, not because `allowableValues` was null but because `isEnum` was set on false for referenced ENUM parameters.

##### openapi-generator version

4.0.0, commit b426bab

##### OpenAPI declaration file content or url

```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Enums Issue
paths: 
  /test:    
    post:
      operationId: TestEnum
      parameters:
        - name: paramA
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/EnumString'
        - name: paramB
          in: query
          required: true
          schema:
            type: string
            enum:
              - first
              - second
              - third
      responses:
        DefaultResponse:
          description: unexpected error
          content:       
            application/json:
              schema:
                type: object
                properties:
                  message:
                    type: string
components:
  schemas:
    EnumString:
      type: string
      enum:
        - first
        - second
        - third
```

##### Command line used for generation

java -jar openapi-generator-cli.jar -i enums.yaml -g erlang-server -o server/

##### Steps to reproduce

- Create the file enums.yaml
- Generate the code with the command
- Open server/src/openapi_api.erl, look at function `request_param_info`(line 52)
- Compare the param infos of paramA and paramB

##### Suggest a fix

In `DefaultCodegen`, `fromProperty`, adding `property.isEnum= true;` at line 2068 after assigning `allowableValues` seems to fix the issue.

```java
        //Referenced enum case:
        if (referencedSchema.getEnum() != null && !referencedSchema.getEnum().isEmpty()) {
            List<Object> _enum = referencedSchema.getEnum();

            Map<String, Object> allowableValues = new HashMap<String, Object>();
            allowableValues.put("values", _enum);
            if (allowableValues.size() > 0) {
                property.allowableValues = allowableValues;
                property.isEnum = true;
            }
        }
```

[...]

PR #3186, opened 2019-06-20
Fixed type assignment for referenced enums

DonDi94:
Change-Id: Ic6e87bfdfa58a4a8f6f1f3aac043c57260ed873c

### PR checklist

### Description of the PR

Referencing to issue #2923 

The change is in DefaultCodegen, so I haven't ran any language-specific script.

The referenced enum schema was missing both the type checking and the assignment of the variable `isEnum`. I moved the typechecking of the property to an external function which I then use for both the properties `p` and `referencedSchema`.

I have been using this fix for a personal project and it has been working well so far. I also tested the case in the issue request and it is solved. The library builds correctly with `mvn clean install`, but sadly I couldn't test through `mvn verify -Psamples` as it stops on language specific tests because I am missing the needed tools and running it in a protected environment.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1995,6 +1995,125 @@ public class DefaultCodegen implements CodegenConfig {
         }
 
         String type = getSchemaType(p);
+        setPropertyType(property, p, name);
+
+        //Inline enum case:
+        if (p.getEnum() != null && !p.getEnum().isEmpty()) {
+            List<Object> _enum = p.getEnum();
+            property._enum = new ArrayList<String>();
+            for (Object i : _enum) {
+                property._enum.add(String.valueOf(i));
+            }
+            property.isEnum = true;
+
+            Map<String, Object> allowableValues = new HashMap<String, Object>();
+            allowableValues.put("values", _enum);
+            if (allowableValues.size() > 0) {
+                property.allowableValues = allowableValues;
+            }
+        }
+
+        Schema referencedSchema = ModelUtils.getReferencedSchema(this.openAPI, p);
+        setPropertyType(property, referencedSchema, name);
+
+        //Referenced enum case:
+        if (referencedSchema.getEnum() != null && !referencedSchema.getEnum().isEmpty()) {
+            List<Object> _enum = referencedSchema.getEnum();
+            property._enum = new ArrayList<String>();
+            for (Object i : _enum) {
+                property._enum.add(String.valueOf(i));
+            }
+            property.isEnum = true;
+
+            Map<String, Object> allowableValues = new HashMap<String, Object>();
+            allowableValues.put("values", _enum);
+            if (allowableValues.size() > 0) {
+                property.allowableValues = allowableValues;
+            }
+        }
+
+        if (referencedSchema.getNullable() != null) {
+            property.isNullable = referencedSchema.getNullable();
+        }
+
+        property.dataType = getTypeDeclaration(p);
+        property.dataFormat = p.getFormat();
+        property.baseType = getSchemaType(p);
+
+        // this can cause issues for clients which don't support enums
+        if (property.isEnum) {
+            property.datatypeWithEnum = toEnumName(property);
+            property.enumName = toEnumName(property);
+        } else {
+            property.datatypeWithEnum = property.dataType;
+        }
+
+        if (ModelUtils.isArraySchema(p)) {
+            property.isContainer = true;
+            property.isListContainer = true;
+            property.containerType = "array";
+            property.baseType = getSchemaType(p);
+            if (p.getXml() != null) {
+                property.isXmlWrapped = p.getXml().getWrapped() == null ? false : p.getXml().getWrapped();
+                property.xmlPrefix = p.getXml().getPrefix();
+                property.xmlNamespace = p.getXml().getNamespace();
+                property.xmlName = p.getXml().getName();
+            }
+
+            // handle inner property
+            property.maxItems = p.getMaxItems();
+            property.minItems = p.getMinItems();
+            String itemName = null;
+            if (p.getExtensions() != null && p.getExtensions().get("x-item-name") != null) {
+                itemName = p.getExtensions().get("x-item-name").toString();
+            }
+            if (itemName == null) {
+                itemName = property.name;
+            }
+            Schema innerSchema = ModelUtils.unaliasSchema(this.openAPI, ((ArraySchema) p).getItems());
+            if (innerSchema == null) {
+                LOGGER.error("Undefined array inner type for `{}`. Default to String.", p.getName());
+                innerSchema = new StringSchema().description("//TODO automatically added by openapi-generator due to undefined type");
+                ((ArraySchema) p).setItems(innerSchema);
+            }
+            CodegenProperty cp = fromProperty(itemName, innerSchema);
+            updatePropertyForArray(property, cp);
+        } else if (ModelUtils.isMapSchema(p)) {
+            property.isContainer = true;
+            property.isMapContainer = true;
+            property.containerType = "map";
+            property.baseType = getSchemaType(p);
+            property.minItems = p.getMinProperties();
+            property.maxItems = p.getMaxProperties();
+
+            // handle inner property
+            Schema innerSchema = ModelUtils.unaliasSchema(this.openAPI, ModelUtils.getAdditionalProperties(p));
+            if (innerSchema == null) {
+                LOGGER.error("Undefined map inner type for `{}`. Default to String.", p.getName());
+                innerSchema = new StringSchema().description("//TODO automatically added by openapi-generator due to undefined type");
+                p.setAdditionalProperties(innerSchema);
+            }
+            CodegenProperty cp = fromProperty("inner", innerSchema);
+            updatePropertyForMap(property, cp);
+        } else if (ModelUtils.isFreeFormObject(p)) {
+            property.isFreeFormObject = true;
+            property.baseType = getSchemaType(p);
+        } else { // model
+            // TODO revise the logic below
+            //if (StringUtils.isNotBlank(p.get$ref())) {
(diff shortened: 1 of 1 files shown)


wing328:

wing328:

DonDi94:
Thank you very much for pointing that out, I didn't know about that. As I was pushing with my personal account I didn't think about the commit author... I wonder if it's even possible to remove the previous commits with the wrong author as they still seem to be reachable.

DonDi94:
@wing328 I don't know why the previous push didn't pass the tests as I just changed the author... Anyway now everything seems to be in place for merge, the commit is authored with my private account.

jmini:
I did try this on my personal project I have to ensure QA of OpenAPI generator.
One of the spec I am using is: [treepots.yaml](https://github.com/jmini/openapi-experiments/blob/master/openapi-generator-utils/treepots/treepots.yaml)

When this PR is applied, the java code generated for this model is changed:
```yaml
    ObjWithEnums:
      type: object
      properties:
        IProp:
            $ref: "#/components/schemas/IntEnum"
        LProp:
            $ref: "#/components/schemas/LongEnum"
        SProp:
            $ref: "#/components/schemas/StringEnum"
        IProp2:
            $ref: "#/components/schemas/Int2Enum"
        LProp2:
            $ref: "#/components/schemas/Long2Enum"
        SProp2:
            $ref: "#/components/schemas/String2Enum"
        Custom:
$ref: "#/components/schemas/Enum2WithCustomName"
```

Instead of using the external generated enum class, an internal one is created.
This is not the idea.


---

I did not do a detailed analysis, but I have the feeling that this PR transform all the
"I am referencing an external enum" to an "let's create an inline enum for that property".

Those are 2 differents concepts.
The Java templates generate proper code in both cases.

---

I did not had a look at the original issue, and from previous work on the enum topic, I can say that the code is not straight forward (probably grown historically).

We can do some cleanup in this area and have something better.

`{{isEnum}}` can not be set to true if your are referencing an Schema that is an enum.
Or if you do so, you need to change all the generators that are relying on this to create inner-enum classes (I am speaking only from the java-generators which I know the best).

DonDi94:
@jmini I agree with your comment, I didn't think about that. There is some weird behavior with your example that I didn't encounter with my tests. I've been using mainly Erlang generation, which is in not great conditions, and that was relying on the "isEnum" property for adding the conditions. I guess I will change it to use allowableValues as flag for enum parameters.

I can't see a use for the "datatypeWithEnum" property. Is that what you refer to as "creating a new inline object"?

DonDi94:
@jmini I think there is an inconsistency with what you said and how things are handled. While I was fixing my local version of the generator I found out this:

```yaml
openapi: 3.0.0
info:
    version: "1.0"
    title: Test
paths:
  /:
    get:
      parameters:
        - name: paramA
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/NotEnum'
        - name: paramB
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/Enum'
      responses:
        default:
          description: "Error"
components:
  schemas:
    NotEnum:
      type: string
    Enum:
      type: string
      enum: [a, b, c]
```

If you generate and look at the log (I use erlang, where the string type is mapped to binary), we have that `paramA` has `binary` for `dataType`, while `paramB` has `Enum` ([GIST, line  16](https://gist.github.com/DonDi94/0527efda3fa9a7767805f5ed398a68e8)). From what I understood from your comment, the referencing enum parameter shouldn't get the `_enum` listing or set `isEnum = true` because that would be like using the referenced schema to create an additional inline schema. With the same reasoning, I would expect `paramA` to not be assigned as `string` (`binary`) but instead it should just reference the `NotEnum` schema and have it as 'dataType`. At the moment, the parameter model created for `paramA` has no reference to `NotEnum`, but instead it is as if it was an inline-defined parameter.

This creates a problem because if we change the definition of `NotEnum` to have a `oneOf` it will be discarded.

```yaml
components:
  schemas:
    NotEnum:
      type: string
      oneOf:
        - pattern: '^[a-z ]*'
        - pattern: '^[A-Z ]*'
    Enum:
      type: string
      enum: [a, b, c]
```

LOG for `paramA` with the spec change: https://gist.github.com/DonDi94/b0a057e6218afdece6ba762b3974dc56

If `NotEnum` is of type `object`, it gets referenced correctly in `paramA`'s `dataType`.

jmini:
I did not want to merge this. Reverting...

jmini:
@DonDi94 sorry about this, we need to continue this discussion. I have clicked on the wrong button

jmini:
5b3e9eb72f6127c1d1be923054d0f83c56fdfc2e pushed to master

jmini:
@DonDi94:
> I think there is an inconsistency with what you said and how things are handled. 

Can you please be more specific? 

----

Yes it is possible that there are problem in the transformation OpenAPI Spec -> CodegenXxxx layer. Not everything is consistent and sometimes the code quality or the initial intention are hard to understand.
Sometimes the problem is also located in the step template step (that are consuming the CodegenXxxx instances).

This is how this project lives. Step by step it is improved to support more cases correctly.

----

I did not look at `oneOf` in details, but I have the feeling that it is not completely supported yet.
`oneOf` for a not-object type is saying: "I would like to apply one or the other validation", and not "lets create an object for that"

----

I went back an forth with this "Should `isEnum` be true when the referenced object is an enum" (I think there are also discussion about this in the Swagger-Codegen project).

From an OpenAPI semantic perspective:

`(Spec 1)`
```
components:
  schemas:
    SomeObject:
      type: object
      Properties:
        id:
          type: integer
        someProp:
          $ref: '#/components/schemas/SomeEnum'
    SomeEnum:
      type: string
      enum: [a, b, c]
```

And `(Spec 2)`:
```
components:
  schemas:
    SomeObject:
      type: object
      Properties:
        id:
          type: string
        someProp:
          type: string
          enum: [a, b, c]
```

Are exactly the same.

This fact would give your theory a point.

But in reality, and the way some generator (I am speaking only for the generators generating java), the code produced by the two examples I have given previously is not the same.

The `(Spec 1)` form forces the generator to produce an independent enum class in the model classes (sometime called "external enum"). This is useful for cases like this:

`(Spec 3)` Example:
```
components:
  schemas:
    MyObject:
      type: object
      Properties:
        id:
          type: integer
        someProp:
          $ref: '#/components/schemas/SomeEnum'
    SomeOtherObject:
      type: object
      Properties:
        firstName:
          type: string
        lastName:
          type: string
        initState:
          $ref: '#/components/schemas/SomeEnum'
    SomeEnum:
      type: string
      enum: [a, b, c]
```

Having in this case the same Enum object shared by `MyObject` and `SomeOtherObject` is really interesting, at least in Java (I can not speak for other languages).

In the client case, when you have a first endpoint that returns a `MyObject` instance and that you need the `someProp` value to pass it to `initState` property in a `SomeOtherObject` instance that you send to a second endpoint.

Good or bad, your change make that two different enum classes would created with the `(Spec 3)` example: `MyObject.SomePropEnum` and `SomeOtherObject.InitStateEnum`.

This is a major change for the user of the generated project.

IMO the current behaviour is correct (creating only `SomeEnum` enum class and using this in both `MyObject.someProp` and `SomeOtherObject.initState`) and there should be a possibility to keep it like this.

I am ready to continue the discussion to find a solution that will work for you too.

_PS: and I am really sorry about the wrong-button click when I have hit merge instead of comment this morning._

DonDi94:
> @DonDi94 sorry about this, we need to continue this discussion. I have clicked on the wrong button

Oh no, I saw the branch being merged, I thought that you revised it and I was so happy for a moment :'D

I keep this as the example to reference to
```yaml
openapi: 3.0.0
info:
    version: "1.0"
    title: Test
paths:
  /:
    get:
      parameters:
        - name: paramA
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/NotEnum'
        - name: paramB
          in: query
          required: true
          schema:
            $ref: '#/components/schemas/Enum'
      responses:
        default:
          description: "Error"
components:
  schemas:
    NotEnum:
      type: string
      oneOf:
        - pattern: "^[a-z]*$"
        - pattern: "^[0-9]*$"
    Enum:
      type: string
      enum: [a, b, c]
```

I did manage to write a fix for the Erlang server generator that worked in theory, but when testing I had two issues:

- The `NotEnum` problem that I described before, which I believe it's inconsistent with what you said about Enum referenced schemas (not to criticize, but to understand and/or to potentially fix). 
To specify: what I would expect from what you said about the Enums is that `paramA`'s processed model would refer to `NotEnum` as `dataType`, in the same way that `paramB`'s model currently refers to `Enum`. In reality, `paramA` will be treated as a string parameter itself rather than a parameter with a referenced string schema. This becomes a problem in the case that there is a `oneOf` or any other composed schema: the information about the `oneOf` is lost as it's still not well supported, and there is no reference left about the referenced schema `NotEnum` that could be used to validate the value through an external schema validator.

- In Erlang, request bodies or objects in general are validated through Jesse, a JSON schema validator configured with the parsed JSON schema that the generator produces. Jesse uses the schema names to find the structure and validate it, but it requires values to be parsed. This becomes a problem: with my PR, referenced ENUMs are treated like other simple parameters and so they are validated correctly, but 1) there is loss of information in other parameters because of the previous point and 2) it is not a good fix as it breaks the examples you mentioned; I also agree more with your logic on the difference between referenced and not referenced. Without my PR I would use the schema name to validate through Jesse, which is a much more flexible and extensive validation which supports composed schemas. However, the model is missing information about the `allowedValues` base type, which doesn't allow me to parse values beforehand, and thus the validation doesn't work.

I haven't had much time lately, but I thought that it would be good to have a type reference in the allowed values to solve my second point. That would be needed if the referenced schemas like `NotEnum` were treated according to the first point.

I also wanted to look for another schema validator, possibly with an integrated parser so that it tries to parse the input from a binary into an integer if the schema needs the input to be an integer. That would solve the typing issue at least.

jmini:
I did a suggestion for your issue there:
https://github.com/OpenAPITools/openapi-generator/issues/2645

wing328:
@DonDi94 thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

PR #3192, opened 2019-06-21
Perl: Basic bearer auth support

yue9944882:
cc @wing328 

adding support for bearer token auth in the api-client. note that it's reusing the OAUTH access token in the configuration.


wing328:
@yue9944882 Please provide a description for the PR :)

yue9944882:
updated

Issue #3135, opened 2019-06-10
[Announcement][Kotlin][Retrofit][Coroutines] New generator: kotlin-retrofit


kroegerama:
👋 Hello OpenAPI-Team!

I created a new generator called "kotlin-retrofit". Are you guys interested in integrating it into your repository and helping me complete/improve it?

**Generator branch: [kroegerama/kotlin-retrofit](https://github.com/kroegerama/openapi-generator/tree/kotlin-retrofit)**

Example project: [Generated petstore](https://github.com/kroegerama/kotlin-retrofit-generated)

## The main features are:
* generates a ready to use module for Android Studio
* only model classes, api classes and two helper classes are generated
* supports per-route authentication (for now: ApiKey, Basic, Bearer)
* uses kotlin data classes
* uses [moshi-codegen](https://github.com/square/moshi#codegen) to parse/serialize json
* uses [retrofit 2.6](https://square.github.io/retrofit/) with support for suspend functions

The petstore.yaml works just fine right now. Also a few other apis that I tested.

## Todos: <sub><sup>(also see README.mustache)</sup></sub>
* **done (allOf)** ~~add better support for json schema inheritance (allOf, anyOf, etc.)~~
* support for `collectionFormat` (e.g. csv)
* add OAuth functionality
* update proguard/p8 rules for okhttp/moshi
* option to switch coroutine generation off
* add documentation
* add tests

@jimschubert @wing328 @karismann @Zomzog

wing328:
> I created a new generator called "kotlin-retrofit". Are you guys interested in integrating it into your repository and helping me complete/improve it?

Yes, we are definitely interested in adding more generators to benefit the community.

I wonder if you can open a PR to start with and we'll review based on what you've created so far.

NicolaeGP:
@kroegerama check out #5076  Perhaps it would be something simple for you to add since you already have support for coroutines in your generator.

kroegerama:
You may want to try my new project:
https://github.com/kroegerama/openapi-kgen

It's a new generator, written entirely from scratch. A generated example can be found here:
https://github.com/kroegerama/openapi-kgen-generated

djensen47:
First, just wanted to say that this is a wonderful, useful project and I appreciate all the hard work put into it.

Second, my querstion is why did this get closed? Is there any possibility of getting kroegerama/openapi-kgen merged into this project?

PR #3195, opened 2019-06-21
New generator: kotlin-retrofit (lightweight, Android, coroutines)

kroegerama:
See #3135


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java
```diff
@@ -0,0 +1,979 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ * Copyright 2018 SmartBear Software
+ * Copyright 2019 kroegerama
```
wing328: For new generators (or files), we only keep line 2.

We credit individual template creator in https://github.com/OpenAPITools/openapi-generator#template-creator. Please feel free to add your Github account and new generator there instead.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java
```diff
@@ -0,0 +1,979 @@
...
+//        public void execute(Template.Fragment fragment, Writer writer) throws IOException {
+//            String text = fragment.execute();
+//            text = camelize(text);
+//            text = toVarName(text);
+//            if (text.length() > 0) {
+//                text = text.substring(0, 1).toUpperCase(Locale.ROOT) + text.substring(1);
+//            }
+//            writer.write(text);
+//        }
+//    }
```
wing328: If this block of commented code is not used, I would suggest simply remove it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java
```diff
@@ -0,0 +1,979 @@
...
+        modelTemplateFiles.put("model.mustache", ".kt");
+        apiTemplateFiles.put("api.mustache", ".kt");
+        apiPackage = packageName + ".apis";
+        modelPackage = packageName + ".models";
+
+//        additionalProperties.put("enumcase", new EnumCaseLambda());
+//        CamelCaseLambda camelCase = new CamelCaseLambda();
+//        camelCase.generator(this);
+//        camelCase.escapeAsParamName(true);
+//        additionalProperties.put("camelcase", camelCase);
```
wing328: Same feedback as above: https://github.com/OpenAPITools/openapi-generator/pull/3195/files#r296541581

wing328:
@kroegerama thanks for contributing a new generator.

Have you considered incorporating the retrofit support via a `--library` option to the existing `kotlin` client generator?

The current `java` client generator supports many HTTP libraries via the `--library` option.

wing328:
cc Kotlin tech committe: @jimschubert (2017/09) , @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java
```diff
@@ -0,0 +1,958 @@
...
+import org.openapitools.codegen.utils.ModelUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.*;
+
+import static org.openapitools.codegen.utils.StringUtils.*;
+
+public class KotlinRetrofitCodegen extends DefaultCodegen implements CodegenConfig {
```
MatkovIvan: I guess `AbstractKotlinCodegen` should be used here

kroegerama: Nope. AbstractKotlinCodegen has some wrong implementations. See my explanation [here](https://github.com/OpenAPITools/openapi-generator/blob/76ad145f198467ed160e55950abcf6f6596b2372/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java#L189) which is just one error I found in the abstract class.
We could merge the corrections in my Codegen into AbstractKotlinCodegen, though.

MatkovIvan: Fix found error in `AbstractKotlinCodegen` is the right option. Plus, it can be done in separate PR to make contributions smaller -> get reviews easier -> merge faster.

BTW it's a nice contribution, looking forward to it :+1:

wing328: Agreed with @MatkovIvan to fix the bug in AbstractKotlinCodegen instead.

snackycracky:
hey, thanks for the effort. Would be awesome if any kind of authenticationScheme is supported :)

Nekromancer:
Hey. Great idea you have there. Do you have any ETA on release? Maybe i could somehow help push this further?

craigmurray1120:
Would love to see this generator gets merged into master.

Thank you @kroegerama for his awesome work.

modules/openapi-generator/src/main/resources/kotlin-retrofit/api.mustache
```diff
@@ -0,0 +1,58 @@
...
+{{#imports}}import {{import}}
+{{/imports}}
+
+import {{packageName}}.RetrofitHolder
+import retrofit2.Response
+import retrofit2.Retrofit
+import retrofit2.create
+import retrofit2.http.*
+import okhttp3.*
+import retrofit2.http.Headers
```
jimschubert: Is this necessary with the glob import above?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinRetrofitCodegen.java
```diff
@@ -0,0 +1,959 @@
...
+        final String infrastructureFolder = (sourceFolder + File.separator + packageName).replace(".", "/");
+
+        supportingFiles.add(new SupportingFile("AuthInterceptors.kt.mustache", infrastructureFolder, "AuthInterceptors.kt"));
+        supportingFiles.add(new SupportingFile("RetrofitHolder.kt.mustache", infrastructureFolder, "RetrofitHolder.kt"));
+        supportingFiles.add(new SupportingFile("EnumUtils.kt.mustache", infrastructureFolder, "EnumUtils.kt"));
+
+        supportingFiles.add(new SupportingFile("proguard-rules.pro.mustache", "", "proguard-rules.pro"));
+        supportingFiles.add(new SupportingFile("gitignore.mustache", "", ".gitignore"));
+
+        supportingFiles.add(new SupportingFile("AndroidManifest.xml.mustache", "src/main", "AndroidManifest.xml"));
```
jimschubert: Note: we may want to put Android options behind a CLI option. I scanned and didn't see any Android specific code in client files.

jimschubert:
I agree that this would make more sense as a library option to the existing client generator. It seems that one of the major reasons it wasn't was because some keywords (like `data`) are set as reserved but disallowed for variables. Many of these aren't allowed elsewhere (can't recall exactly, but I think as Enums), so I think any special casing logic makes the most sense in the base class to benefit all Kotlin generators.

What might you need from the core team to get this into master? Maybe we could do it as a separate Beta generator and move to a library option later?

wing328:
Here is another PR to add retrofit2 as a library option to the Kotlin client generator: https://github.com/OpenAPITools/openapi-generator/pull/4518

wing328:
https://github.com/OpenAPITools/openapi-generator/pull/4518 has been merged into master.

Please give it a try. Closing this one for the time being.

PR #3202, opened 2019-06-22
[openapi-yaml] Prerequisities for Swagger 2 template.

MichalFoksa:
### PR checklist

### Description of the PR

I am looking for a way to convert from Openapi 3 to Swagger 2 specification. I think `openapi-yaml` generator with custom template is a good starting point. 

In order to start, I needed to modify current generator to:
1. Group operations by resourcePath (they are grouped by tag currently) so that operations with same path are rendered one after the other:
    Example:
```
      /pet:
        post:
          .....
        put:
          .....
      /pet/findByStatus:
        get:
        .....
```

   And not (it might happen when grouped by a tag):
```
      /pet:
        post:
        .....
      /pet/findByStatus:
        get:
        .....
      /pet: 
        put:
        .....
 ```

2. Created a Mustache lambda `OnChangeLambda` to return fragment only when current fragment is different than previous fragment. It will be used to render endpoint path only for a first operation in group.

3. Registered lambdas like lowercase, onchage (above one), uppercase, etc.

Later I am hoping to submit Swagger 2 template a new library.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OpenAPIYamlGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OpenAPIYamlGenerator.java
@@ -18,9 +18,23 @@
 package org.openapitools.codegen.languages;
 
 import org.openapitools.codegen.*;
+import org.openapitools.codegen.templating.mustache.CamelCaseLambda;
+import org.openapitools.codegen.templating.mustache.IndentedLambda;
+import org.openapitools.codegen.templating.mustache.LowercaseLambda;
+import org.openapitools.codegen.templating.mustache.TitlecaseLambda;
+import org.openapitools.codegen.templating.mustache.UppercaseLambda;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableMap;
+import com.samskivert.mustache.Mustache;
+import com.samskivert.mustache.Template;
+import io.swagger.v3.oas.models.Operation;
+
+import java.io.IOException;
+import java.io.Writer;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 
 public class OpenAPIYamlGenerator extends DefaultCodegen implements CodegenConfig {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by MichalFoksa** [comment]
PR re-based to master.

File: modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OpenAPIYamlGenerator.java [hunk 1]
```diff
@@ -61,6 +75,44 @@ public void processOpts() {
         }
         LOGGER.info("Output file [outputFile={}]", outputFile);
         supportingFiles.add(new SupportingFile("openapi.mustache", outputFile));
+
+        addMustacheLambdas(additionalProperties);
+    }
+
+    private void addMustacheLambdas(Map<String, Object> objs) {
```
jmini: Since the same method is used in `AbstractCSharpCodegen`, `AbstractFSharpCodegen` and `KotlinServerCodegen` I think is would make sense to mutualize it move it somewhere (instead of all those copy-pasted methods)

Maybe as protected method in `DefaultCodegen`
Or as public static method in a utility class.

@MichalFoksa What do you think?
@jimschubert any opinion on this?

MichalFoksa:
Will look into it later

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OpenAPIYamlGenerator.java
MichalFoksa: Will look into it later

jimschubert: @jmini I agree. I didn't put it in the DefaultCodegen type because I wasn't sure if would be acceptable globally. If moved there, we'll want to document these as available Mustache lambdas in our template docs.

MichalFoksa: How would you like it? As a new method in CodegenConfig interface and 

1. an empty method in DefaultCodegen which gets called by DefaultGenerator and each generator uses it to register own lambdas. 
    `DefaultCodegen.addMustacheLambdas() {}`


2. Or do you want it to already register some lambdas, for exapmle:
```
    DefaultCodegen.addMustacheLambdas() {
        lambdas = new ImmutableMap.Builder<String, Mustache.Lambda>()
                .put("lowercase", new LowercaseLambda().generator(this))
                .put("uppercase", new UppercaseLambda())
                .put("titlecase", new TitlecaseLambda())
                .put("camelcase", new CamelCaseLambda().generator(this))
                .put("indented", new IndentedLambda())
                .....
                .build();
        objs.put("lambda", lambdas);
    }
```
?

wing328:
The CI failure has been addressed in the master

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OpenAPIYamlGenerator.java
jmini: @MichalFoksa: I do not think that we need the method in the `CodegenConfig` interface. If I understand @jimschubert correctly the idea is: `2.` just register the Lambdas that we have for all generators. And add some explanations about usage in [`templating.md`](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/templating.md) probably as new section "Lambdas" after "Variables" and before "Extensions".

Can you create a follow up PR?

jimschubert: It _might_ need to be on the interface as well, as I believe another generator (C++ maybe) created lambdas with different names. I'll take a look later when I have a chance.

MichalFoksa: I will create a follow up PR. In meantime I have discovered #3364.

MichalFoksa: Here you are the follow up PR #3368

jmini:
Small code review => PR #3384 (I have one error locally, because I check for hidden vars)

Issue #3176, opened 2019-06-19
[BUG] [SPRING-CLOUD] Generate Optionals for Dtos


SirWayne:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description
If you use nullable in yml for a property, the dto/model generated code is not a Optional. Actual output ist the type. I would expect the type is a optional of the declared type

##### openapi-generator version

4.0.1

##### OpenAPI declaration file content or url

If you post the code inline, please wrap it with
```yaml
  FooDto:
    type: "object"
    properties:
      bar:
        type: "string"
        nullable: true
```

##### Command line used for generation

Use maven for the generation

```xml
    <configuration>
        <inputSpec>${project.basedir}/src/main/resources/api.yaml</inputSpec>
        <generatorName>spring</generatorName>
        <configOptions>
            <sourceFolder>src/main/java</sourceFolder>
            <library>spring-cloud</library>
            <useOptional>true</useOptional>
        </configOptions>
    </configuration>
```

PR #3214, opened 2019-06-25
Fix/generate optionals for dtos 3176

SirWayne:
### PR checklist

- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches) Default: `master`.
- [x] Copied the [technical committee]@bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04)to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Fix to generate Optionals for the JavaSpring Language.

Fix: https://github.com/OpenAPITools/openapi-generator/issues/3176


modules/openapi-generator/src/main/resources/JavaSpring/nullableDataType.mustache
```diff
@@ -1 +1 @@
-{{#isNullable}}JsonNullable<{{{datatypeWithEnum}}}>{{/isNullable}}{{^isNullable}}{{{datatypeWithEnum}}}{{/isNullable}}
\ No newline at end of file
+{{#isNullable}}JsonNullable<{{{datatypeWithEnum}}}>{{/isNullable}}{{^isNullable}}{{>optionalDataType}}{{/isNullable}}
```
cbornet: It seems weird to have both JsonNullable and Optional...

SirWayne: I was not sure if the JsonNullable could be removed due to breaking changes.
What do you mean I should remove them?

cbornet: No you can't remove JsonNullable. JsonNullable already has a notion of presence/absence so it's weird to have a double notion for this.
Note that having Optional POJO fields is a bad practice in the first place...

SirWayne: > Note that having Optional POJO fields is a bad practice in the first place...

Sorry no it's not really a bad practice. Nullable objects are a bad practice. Since jackson have a JDK8Module, is no reason why you should not use the Optional instead of own JsonNullable object. The Optional works fine.
Here is a good presentation why Optionals are not a bad practice.
https://slides.codefx.org/expert-java-8/2017-09-18-objektforum-stuttgart/index.html#/

I think nullable collections and the JsonNullable is more a code smell and bad practice

cbornet: Exactly in the [presentation you linked](https://slides.codefx.org/expert-java-8/2017-09-18-objektforum-stuttgart/index.html#/_basic_rules): 

> make everyone setting Optional to null buy a round of drinks or wear a silly hat

And that's exactly what you do by setting your field to Optional in the POJO :smiley:  ! :beer: accepted !

SirWayne: Ok that's a bug, the default value have to be `Optional.empty()`. We have constructor for that job ;). I will fix that. 

But this was not your reason, why the Optional is a bad practice?

SirWayne: Thanks for the hint, fixed it. Now no `Optional` is `null`. Default value is `Optional.empty()`

cbornet: The reasons are all in your presentation (which is really good btw): no serializability, decreased performance due to de-referencing, was designed only as a method return value, etc...
I know there's a lot of bikeshedding over the usage of Optionals so if you want to use them in your POJOs, just do it. Generator options are there for this.

As for JsonNullable, note that it has a different purpose than Optional. It's meant to handle tri-state variables that can happen when doing [JSON Merge Patch](https://tools.ietf.org/html/rfc7386). You need the 3 states: absent, present and null, present and non-null. You can't do that with Optionals.

cbornet: Oh! I see that you want to use Optionals for nullable fields. Don't do that : you can't set an Optional to null so you only have the choice between `empty` and `not-empty` but nullable is about having the choice between `absent`, `null`, and `non-null` so that's different.

SirWayne: Yes i like the presentation too. For that reason i would like to use Optional instead of nullable fields. 
Thanks for your hint, i know the tri-state issue, but i thought (i will double check that) there is a solution with Optional. First i thought `empty` `not-empty` and `empty string`, but this will not work with Integer.

SirWayne: It looks like you really need `null` as state. Then the difference is not so big except that you do not need the extra lib.
https://github.com/FasterXML/jackson-databind/issues/578
https://github.com/FasterXML/jackson-datatype-jdk8/issues/2

SirWayne: You've convinced me. It is really better to use the `JsonNullable`. Optionals and zero cost too much beer ;).

So will close the ticket. Thanks for your support.

One question to the `JsonNullable`. Is there a reason why the class have no method `isUndefined()`?

e.g.

```
public boolean isUndefined(){
return UNDEFINED.equals(this);
}
```

PR #3218, opened 2019-06-25
[Typescript-Fetch] If a query parameter is a date, format it as YYYY-MM-DD

markoudev:
### PR checklist

Tech Committee: @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10)

### Description of the PR

For query parameters that are specified as `type: string` and `format: date`, the URL now contains the entire ISO-string while it should only include the date part in format `YYYY-MM-DD`. This distinction between `date` and `date-time` is already properly made in the models, but not in the client API generation.

In my project I cannot use the generated library for certain API calls that include something like `?date=` in their URL. The remote server may reject the request because it's not in the format `YYYY-MM-DD`.

I noticed that the sample Petstore does not include an example that uses `format: date`. There are ones that use `format: date-time`. I guess that adding such an operation means that all samples must be regenerated and checked, which I thought is something out of the scope of this PR.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -69,7 +69,7 @@ export class {{classname}} extends runtime.BaseAPI {
             {{/isDateTime}}
             {{^isDateTime}}
             {{#isDate}}
-            queryParameters['{{baseName}}'] = (requestParameters.{{paramName}} as any).toISOString();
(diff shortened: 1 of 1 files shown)


macjohnny:
CI failure is not related to this PR

wing328:
@markoudev thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

PR #3220, opened 2019-06-25
Added RxJava2 to java-vertx server codegen

jharmelink:
RxJava2 uses another base package. To support this package in java-vertx server codegen, a `rxVersion2` `boolean` flag/option was added which defaults to `false`.

@phiz71 @wing328

### DIFF
--- a/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
+++ b/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
@@ -245,7 +245,7 @@ public class CodeGenMojo extends AbstractMojo {
      * To treat a document strictly against the spec.
      */
     @Parameter(name = "strictSpec", property = "openapi.generator.maven.plugin.strictSpec", required = false)
-    private Boolean strictSpecBehavior;
+    private Boolean strictSpec;
 
     /**
      * To generate alias (array, map) as model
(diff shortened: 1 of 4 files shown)


modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
```diff
@@ -245,7 +245,7 @@
      * To treat a document strictly against the spec.
      */
     @Parameter(name = "strictSpec", property = "openapi.generator.maven.plugin.strictSpec", required = false)
-    private Boolean strictSpecBehavior;
+    private Boolean strictSpec;
```
wing328: @jharmelink are these changes in this file intentional?

wing328:
@jharmelink from what I understand, this is not a breaking change so you can file the change against the master branch (next patch release will be in a week)

modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
jharmelink: This is actually a bugfix to get the generation working again. Not related to this change.

jharmelink:
I'll create a new PR for master

modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
wing328: The issue should have been fixed in the master. Please have a look and let us know if that's not the case

jharmelink:
Created new PR for master: https://github.com/OpenAPITools/openapi-generator/pull/3240

PR #3234, opened 2019-06-27
[dart-jaguar] Adds option to customise the generated API client name

athornz:
### PR checklist

### Description of the PR

Derives the generated API class name from the pub name, rather than the generic `JaguarApiGen` name.

CC @ircecho @swipesight @jaumard


wing328:
@athornz thanks for the PR. Quick question about the client name. Is that related to the pubName setting somehow?
```
        pubName
            Name in generated pubspec
```

If yes, shall we reuse `pubName` instead?

(disclosure: no expert in Dart)

athornz:
It could make sense to reuse it, but the naming convention for dart packages is snake_case which is likely not what the user would want for the class name.

wing328:
> dart packages is snake_case which is likely not what the user would want for the class name.

We can use camelize, etc string utility function to convert `pubName` into proper format for {{clientName}} in the generator

athornz:
Sure that seems like a good option 👍

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartJaguarClientCodegen.java
```diff
@@ -142,6 +147,13 @@ public void processOpts() {
             additionalProperties.put(PUB_NAME, pubName);
         }
```
wing328: @athornz what about setting clientName here using StringUtils methods?

That way we don't need to introduce another option

athornz: Great suggestion, works really well. I've pushed the changes now.

wing328:
The CircleCI error is not related to this PR.

Since it's a breaking change with fallback, I've retargeted this PR to 4.1.x branch.

wing328:
Looks like the rebase has failed and therefore this PR contains commits not authored by you.

I'll try to cherry-pick your commits and create a new PR.

athornz:
Thanks!

wing328:
@athornz have a nice weekend

wing328:
@athornz thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

PR #3240, opened 2019-06-28
Added RxJava2 to java-vertx server codegen

jharmelink:
RxJava2 uses another base package. To support this package in java-vertx server codegen, a rxVersion2 boolean flag/option was added which defaults to false.

@wing328


wing328:
The CircleCI failure is due to outdated documentation. I'll fix that after merging this into master.

PR #3241, opened 2019-06-28
Go struct members pointers

bkabrda:
### PR checklist

### Description of the PR

@antihax (2017/11) @bvwells (2017/12) @grokify (2018/07) @kemokemo (2018/09)

This is a newly opened PR #3113, since the old one failed rebasing to 4.1.x in the Github UI. See #3113 for previous discussion.


modules/openapi-generator/src/main/resources/go/model.mustache
```diff
@@ -35,9 +35,95 @@ type {{classname}} struct {
 {{#description}}
 	// {{{description}}}
 {{/description}}
-	{{name}} {{#isNullable}}*{{/isNullable}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+	{{name}} *{{{dataType}}} `json:"{{baseName}},omitempty"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
yue9944882: what's the point of adding a field `isExplicitNull` to mark it explicity? as a convention of go, it's a nullable if it's a pointer, otoh, a non-pointer field means it's a non-optional field

modules/openapi-generator/src/main/resources/go/model.mustache
```diff
@@ -35,9 +35,95 @@ type {{classname}} struct {
...
+	if o == nil || o.{{name}} == nil {
+		var ret {{dataType}}
+		return ret
+	}
+	return *o.{{name}}
+}
+
+// Get{{name}}Ok returns a tuple with the {{name}} field if it's non-nil, zero value otherwise
+// and a boolean to check if the value has been set.
+func (o *{{classname}}) Get{{name}}Ok() ({{dataType}}, bool) {
```
yue9944882: the name doesn't really sonuds ok.. in favor of others e.g. `Get{{name}}IfExists`..

modules/openapi-generator/src/main/resources/go/model.mustache
yue9944882: also, it's breaking backward-compatibility. the field shouldn't twiddle between pointer and non-pointer at least

modules/openapi-generator/src/main/resources/go/model.mustache
```diff
@@ -35,9 +35,95 @@ type {{classname}} struct {
...
+// when serializing to JSON (pass true as argument to set this, false to unset)
+// The {{name}} value is set to nil even if false is passed
+func (o *{{classname}}) Set{{name}}ExplicitNull(b bool) {
+	o.{{name}} = nil
+	o.isExplicitNull{{name}} = b
+}
+{{/isNullable}}
+{{/vars}}
+
+func (o {{classname}}) MarshalJSON() ([]byte, error) {
```
yue9944882: 1. this should belong to another pull, discriminating the change..
2. the custom marshal has to be a opt-in extension..

modules/openapi-generator/src/main/resources/go/model.mustache
bkabrda: The point is to be able to have a field that can be sent with `null` as a value. The whole point of this PR is that it's very hard in Go to have fields that allow setting either a zero value (0, false, empty string) or a `null` explicitly. This PR allows for various combinations of all of these features thanks to `isExplicitNull` and the custom marshalling.

modules/openapi-generator/src/main/resources/go/model.mustache
bkabrda: I actually prefer `Get{{name}}Ok`, since it returns two values - the value of `{{name}}` and whether or not retrieving it was ok. With `IfExists`, I would expect the function to return only the value of `{{name}}` if it exists. YMMV.

modules/openapi-generator/src/main/resources/go/model.mustache
bkabrda: Yeah, I've had a discussion about this with @wing328 and we'll probably create another go client generator to accept this change (and possibly other breaking changes) that would otherwise be unacceptable for 4.1.x and would have to wait for 5.0.0 - which would be a loooong time.

modules/openapi-generator/src/main/resources/go/model.mustache
bkabrda: 1. As noted in one of the comments above, the custom marshalling is a core of this change without which the rest wouldn't work reasonably.
2. Why?

modules/openapi-generator/src/main/resources/go/model.mustache
wing328: I notice that {{#isNullable}} ... {{/isNullable}} has been removed. Does it mean model properties with or without nullable set to true are handled the same way?

bkabrda: These are handled explicitly in the custom marshalling method, see https://github.com/OpenAPITools/openapi-generator/pull/3241/files#diff-6403a68f6b1039398715805f9043d0a1R95 and following lines.

wing328:
As discussed, you will resubmit the change for `go-experimental` generator instead so closing this one for the time being.

modules/openapi-generator/src/main/resources/go/model.mustache
jimschubert: @bkabrda For the new go generator, would you mind tagging it with "Experimental" 

https://github.com/OpenAPITools/openapi-generator/blob/6e1c8976fcf0d42904240ff7218f4105b03121f8/modules/openapi-generator-core/src/main/java/org/openapitools/codegen/meta/Stability.java#L19

It would look like this in the new generator's constructor:

```
        generatorMetadata = GeneratorMetadata.newBuilder(generatorMetadata)
                .stability(Stability.EXPERIMENTAL)
               .build();
```

This will allow us to tag/filter generators to clearly present which ones are considered stable to users.

PR #3265, opened 2019-07-02
[online] Honor generator environment variable and system property

jimschubert:
Previously, this pulled from the GENERATOR_HOST system property. This
should have been an environment variable. Fallback is now generator.host
system property.

### PR checklist


jimschubert:
I realized during testing that I targeted the environment variable via `System.getProperty` rather than `System.getenv`. The PR honors both.

cbornet:
Springfox already has a `springfox.documentation.swagger.v2.host` to do that.
Otherwise it can detect the hostname from the request (respecting the x-forwarding-* headers) so forcing it to localhost by default  is not correct.

jimschubert:
@cbornet I'll look at passing this to `springfox.documentation.swagger.v2.host` in the Dockerfile and on the server, and removing `generator.host` as an available property.

We'll probably want to keep `GENERATOR_HOST` to avoid hardcoded `*.swagger.v2.*` property which will most likely change in a newer springfox.

PR #3275, opened 2019-07-03
generate schema classes for optionally provided secondary spec files

### LINKED ISSUE #3257: [BUG] [angular-typescript] Unused models not generated
#### Bug Report Checklist

##### Description

When defining model entities that are not used directly by the resscources these models are not generated with the typescript-angular generator. (see `SpecialPet` in the example below)

This is a problem since our api uses the `allOf` to extend a base class. The ressource formally returns the base class, but depending on inputs, different acutal responses (subclasses) are returned. It is clear that there needs to be some manual casting, but without the subclass at hand we'd need to manually code it.

4.0.2

```yaml
openapi: 3.0.0
info:
  version: "1.0.0"
  title: "Swagger Petstore"

paths:
  /pet/findByStatus:
    get:
      responses:
        200:
          description: "successful operation"
          content:
            application/json:
              schema:
                type: "array"
                items:
                  $ref: '#/components/schemas/SpecialPet'
        400:
          description: "Invalid status value"
  
components:
  schemas:
    Pet:
      type: "object"
      properties:
        id:
          type: "integer"
          format: "int64"
        name:
          type: "string"
          example: "doggie"
    SpecialPet:
      type: "object"
      allOf: 
      - type: object
        properties:
          specialType:
            type: string
```

##### Suggest a fix

There should be an option to allow generation of unused models.

daniel-sc:
### PR checklist

### Description of the PR

This allows to generate classes for (technically) unused schemas in secondary specification files using the new option `--additional-schemas-input-specs` - see #3257.

fixes #3257

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -66,6 +66,10 @@ public class Generate implements Runnable {
             description = "location of the OpenAPI spec, as URL or file (required)")
     private String spec;
 
+    @Option(name = {"--additional-schemas-input-specs"}, title = "additional spec files", required = false,
+            description = "comma separated list of locations of OpenAPI specs, as URL or file for which components should be generated (optional)")
+    private String additionalSchemasInputSpecs;
+
     @Option(name = {"-t", "--template-dir"}, title = "template directory",
             description = "folder containing the template files")
     private String templateDir;
@@ -272,12 +276,15 @@ public class Generate implements Runnable {
         }
 
         if (isNotEmpty(spec)) {
-            if (!spec.matches("^http(s)?://.*") && !new File(spec).exists()) {
-                System.err.println("[error] The spec file is not found: " + spec);
-                System.err.println("[error] Check the path of the OpenAPI spec and try again.");
-                System.exit(1);
-            }
+            checkSpecPath(spec);
             configurator.setInputSpec(spec);
+        }
+
+        if (isNotEmpty(additionalSchemasInputSpecs)) {
+            for (String additionalModelInputSpec : additionalSchemasInputSpecs.split(",")) {
+                checkSpecPath(additionalModelInputSpec);
+                configurator.getAdditionalSchemasInputSpecs().add(additionalModelInputSpec);
+            }
         }
 
         if (isNotEmpty(generatorName)) {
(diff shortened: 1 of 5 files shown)


macjohnny:
Cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10)

macjohnny:
Cc @wing328 @jmini @jimschubert

macjohnny:
@wing328 do the docs need to be updated somewhere else, too?

macjohnny:
ping @wing328

jimschubert:
I think this addition will add a lot of complexity to the generator just to support an edge case. I'd much rather guide users looking for this functionality to rerun generation against each additional spec and generate the models that way.

A main reason for this is that it'll complicate what we consider a "valid" spec. If one of these additional specs don't define a path object, they're technically invalid.

Another concern is naming conflicts. This impression will overwrite any schemas with the same name, and this can be confusing. Additionally, the parsing done by swagger-parser may not result in the expected once graph if you're referencing models across these files (better to stick to `$ref`).

daniel-sc:
Hi @jimschubert,

really appreciate the feedback!

I'll try to argue the other case, as to my experience and projects I've seen, the setup as described in https://github.com/OpenAPITools/openapi-generator/issues/3257 is often used. I agree that this might not follow 100% the intentions of the technical specs, but as many teams use this approach to split up model and paths and also use inheritance like this (at least as long `oneOf`  is not completely supported by some (many?) generators), I'd suppose this is a rather heavily used edge case.

Would it be possible to mark this flag as "beta" so that it can be removed again, if acutal problems show up?

(Of course I'll resolve the merge conflicts, once we're somewhat sure if we want this included..)

jimschubert:
@daniel-sc I don't disagree that there's utility in separating schemas by file. In fact, his is already supported by swagger-parser, and should work as expected. My comment above focuses on targeting multiple, potentially unrelated, spec documents.

As an example, this implementation would allow me to target `petstore.json` plus `petstore-simple.json` and `petstore-invalid.json`... which `Pet` schema would be expected in the outcome? What if two of these are version 2.0 and one is version 3.0.2? Should we do inline model resolution before or after adding supplemental models? There's a non-trivial amount of complexity that's added by trying to extend the behaviors of the specification.

There current master should generate unused models if they're referenced in the spec. Let me try the example in the open issue later to understand what's missing.

daniel-sc:
@jimschubert Did you manage to check the example in the issue?

PR #3287, opened 2019-07-06
[online] Revert GENERATOR_HOST explicit setting

jimschubert:
### PR checklist

### Description of the PR

Removing `GENERATOR_HOST` usage, and moving configuration to server.  We may want to consider removing `GENERATOR_HOST` completely from GenApiServer (if possible), and just using built-in support for retrieving the host/port. I was trying to reduce confusion around this variable in the code that is removed by this PR, but upon discussion it seems the environment variable just adds confusion and is therefore not the appropriate configuration to target.


jimschubert:
@cbornet @wing328

modules/openapi-generator-online/Dockerfile
```diff
@@ -4,9 +4,10 @@ WORKDIR /generator
 
 COPY target/openapi-generator-online.jar /generator/openapi-generator-online.jar
 
-ENV GENERATOR_HOST=http://localhost
+# GENERATOR_HOST can be used to determine the target location of a download link.
+# The default value asumes binding to host via: docker -p 8080:8080 image_name
+ENV GENERATOR_HOST=http://localhost:8080
```
cbornet: I'm not sure I understand. Do we keep GENERATOR_HOST?

jimschubert: @cbornet `GENERATOR_HOST` is still used here:

https://github.com/OpenAPITools/openapi-generator/blob/109808e60d15e97680f274bda8ed854da52b67ea/modules/openapi-generator-online/src/main/java/org/openapitools/codegen/online/service/GenApiService.java#L161

This isn't "correct" and will need to be changed. The previous value here would be incorrect for all invocations unless bound to port 80, which I believe requires super user privileges. This change at least causes the variable to match the exposed port.

Many people try out docker images by binding to random host ports (`-P`), and when doing this… `GENERATOR_HOST` will be useless and the file download link generation won't work.

This PR limits focus to get the proxying host setting working on the server. I can look at the other work later.

jimschubert:
This reverts changes I added in #3263 and #3265.

jimschubert:
CircleCI failure is unrelated. Merging this to let the deployment do its thing, and creating an issue for tracking removal of `GENERATOR_HOST` from GenApiService.

modules/openapi-generator-online/Dockerfile
jimschubert: Created #3288 to address this other concern.

PR #3316, opened 2019-07-09
[typescript-rxjs] runtime and api cleanup

denyo:
### PR checklist

### Description of the PR

This PR simplifies models `RequestArgs` and `ResponseArgs` and makes them an alias for `{ AjaxRequest, AjaxResponse } from 'rxjs/ajax';`. Also parameter names make more sense now and unneeded checks are removed.
For people that have previously applied any middlewares this is breaking due to the model changes but it is quite easy to fix. Docs where also extended.

#### Middleware Before
```
export class AuthInterceptor extends Configuration {
  private static config: AuthInterceptor;

  private constructor() {
    const middleware: Middleware[] = [
      {
        pre: (context: RequestContext): RequestArgs => ({
          ...context,
          options: {
            ...context.options,
            headers: {
              ...context.options.headers,
              Authorization: `Bearer ${getAuthToken()}`
            }
          }
        })
      }
    ];

    super({ middleware });
  }

  public static get Instance() {
    return AuthInterceptor.config || (AuthInterceptor.config = new this());
  }
}

```

#### Middleware After
```
export class AuthInterceptor extends Configuration {
  private static config: AuthInterceptor;

  private constructor() {
    const middleware: Middleware[] = [
      {
        pre: (request: RequestArgs): RequestArgs => ({
          ...request,
          headers: {
            ...request.headers,
            Authorization: `Bearer ${getAuthToken()}`,
          },
        }),
      },
    ];

    super({ middleware });
  }

  public static get Instance() {
    return AuthInterceptor.config || (AuthInterceptor.config = new this());
  }
}
```


modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -184,12 +171,11 @@ export const throwIfRequired = (params: {[key: string]: any}, key: string, nickn
     }
 }
 
-export interface RequestContext extends RequestArgs {}
-export interface ResponseContext extends RequestArgs {
-    response: AjaxResponse;
-}
+// alias for easier importing
+export interface RequestArgs extends AjaxRequest {}
+export interface ResponseArgs extends AjaxResponse {}
```
macjohnny:
```suggestion
export interface RequestContext extends RequestArgs {};
```
for backwards compatibility

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -184,12 +171,11 @@ export const throwIfRequired = (params: {[key: string]: any}, key: string, nickn
     }
 }
 
-export interface RequestContext extends RequestArgs {}
-export interface ResponseContext extends RequestArgs {
```
macjohnny: this isnt backwards compatible, so I would not delete ResponseContext

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -83,46 +83,38 @@ export class BaseAPI {
...
+            map((request) => {
+                this.middleware.filter((item) => item.pre).forEach((mw) => (request = mw.pre!(request)));
+                return request;
             }),
             concatMap((args) =>
-                ajax({ url: args.url, ...args.options }).pipe(
+                ajax(args).pipe(
                     map((response) => {
-                        if (postMiddlewares) {
-                            postMiddlewares.forEach((mw) => (response = mw.post!({ ...params, response })));
```
macjohnny: this isn't backwards-compatible so I would revert it

macjohnny:
cc @akehir

macjohnny:
according to https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches it should be ok to merge this into `4.1.x`, as the breaking-change is easy to fix

macjohnny:
@denyo can you cherry-pick into 4.1.x?

akehir:
> ~@denyo can you cherry-pick into 4.1.x?~
> `4.0.3` is the last patch release before `4.1.0`, so it is fine to merge this into master

@macjohnny,  during the release, the 4.1.0 branch would be merged into master, and future 4.1.x patch releases would be branched of master (unexpectedly containing the breaking change). Merging to master seems to violate the merge direction description of the git branches wiki entry.

How are breaking changes documented in order to help people experiencing the breakage to be able to fix this easily? 

I understand that simplifying the template is a good idea, however, I would be reluctant to introduce unexpected breaking changes (especially since whoever doesn't like the template can easily override it in the local project), if it cannot be communicated properly.

denyo:
I kind of had the same thoughts as @akehir.
On the one hand I am all for simplification and performance, on the other I am not a fan of breaking changes either in patch or minor version bumps.
In general until recently with 1 or 2 bug reports about this particular generator I had the feeling it wasn't really used by anyone so I wasn't too worried. Are there any usage stats about this?
Since it is initially based on `typescript-fetch` some of their technical debt was obvisouly dragged over. So this PR is mainly towards cleaning this up since `rxjs` is providing the ideal interfaces already.

macjohnny:
now that `4.0.3` has been released, the next release (i.e. `4.1.0`) will be taken from master, as the `4.1.x` branch should be merged into master now.
in the next minor release (`4.1.0`), breaking changes with fallback are permitted. I labelled this PR with `Breaking changes with fallback`, so it can be found in the release notes (see e.g. https://github.com/OpenAPITools/openapi-generator/releases/tag/v4.0.0), so the migration steps should be documented here in this PR. @denyo can you do that?

denyo:
@macjohnny migration steps are in the description with before and after? Or what do you mean?

macjohnny:
> migration steps are in the description with before and after? Or what do you mean?

I think the description in the PR should be enough, but maybe it might be worth adding an additional line explaining explicitly which signature changed.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
```diff
@@ -228,4 +296,71 @@ private void addExtraReservedWords() {
...
         this.reservedWords.add("AjaxResponse");
     }
+
+    class ExtendedCodegenOperation extends CodegenOperation {
+        public boolean hasHttpHeaders;
+
+        public ExtendedCodegenOperation(CodegenOperation o) {
+            super();
+
+            // Copy all fields of CodegenOperation
```
TiFu: Without having checked in the code, would it be possible to call `super(o)` instead of copying all the assignments?

denyo: Unfortunately no.
I am no java expert but I think the reason is that there is only the default constructor.
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project openapi-generator: Compilation failure
[ERROR] .../openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java:[304,13] constructor CodegenOperation in class org.openapitools.codegen.CodegenOperation cannot be applied to given types;
[ERROR]   required: no arguments
[ERROR]   found: org.openapitools.codegen.CodegenOperation
[ERROR]   reason: actual and formal argument lists differ in length
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
```diff
@@ -190,22 +201,79 @@ private void addOperationModelImportInfomation(Map<String, Object> operations) {
...
                             .replace(param.enumName, op.operationIdCamelCase + param.enumName);
                 }
             }
         }
 
-        operations.put("hasEnums", hasEnum);
+        operations.put("hasEnums", hasEnums);
+    }
+
+    private void addConditionalImportInformation(Map<String, Object> operations) {
```
macjohnny: could this be moved to the abstract typescript codegen?

denyo: I suppose it could go there so other generators can use it as well. 
Should I move it there?

macjohnny: yes, please do that

denyo: This would require moving `ExtendedCodegenOperation` also into `AbstractTypeScriptClientCodegen` because of the newly added `hasHttpHeaders` field. I have the feeling that doing so would impact the other generators using the basic `CodegenOperation`?

macjohnny: I can't really tell you the impact of that, but maybe we leave it as it is and consider it in the typescript refactoring.

denyo: Ok, great. Then I leave it as it is.

PR #3322, opened 2019-07-09
Do not check status code for default response

### LINKED ISSUE #3321: [BUG][GO] default response not handled for error cases
#### Bug Report Checklist

##### Description

Current golang client does not handle default response data types when the the server returns an HTTP error code (4xx).

master

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    post:
      summary: Create a pet
      operationId: createPets
      tags:
        - pets
      responses:
        '201':
          description: Null response
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
    Error:
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

`java -jar  modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i petstore.yaml -g go -o out`

##### Steps to reproduce

1. Generate client as per command line above
2. Verify that generated code at api_pets.go checks HTTP status code against 0:

```go
	if localVarHttpResponse.StatusCode >= 300 {
		newErr := GenericOpenAPIError{
			body:  localVarBody,
			error: localVarHttpResponse.Status,
		}
		if localVarHttpResponse.StatusCode == 0 {
			// ...               
		}
	}
```
That doesn't work because the status code will never be zero at that point.

##### Related issues/PRs

Couldn't find any.

##### Suggest a fix

Do not generate the check for status code when generating code for the default response.

thiagoarrais:
### PR checklist

### Description of the PR

Fixes #3321 by not checking HTTP status when generating code for a default response.

Reviewers, please check review comments.

cc @antihax (2017/11) @bvwells (2017/12) @grokify (2018/07) @kemokemo (2018/09)


modules/openapi-generator/src/main/resources/go/api.mustache
```diff
@@ -299,7 +299,7 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
 			}
 			newErr.model = v
 			return {{#returnType}}localVarReturnValue, {{/returnType}}localVarHttpResponse, newErr
-		}
+		{{^wildcard}} } {{/wildcard}}
 		{{/dataType}}
 		{{/responses}}
 		return {{#returnType}}localVarReturnValue, {{/returnType}}localVarHttpResponse, newErr
```
thiagoarrais: This gets generated even when there is a default response. We end up with a double return in the final code. But I couldn't find a way to reliably avoid it. Any ideas?

modules/openapi-generator/src/main/resources/go/api.mustache
```diff
@@ -290,7 +290,7 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
 		}
 		{{#responses}}
 		{{#dataType}}
-		if localVarHttpResponse.StatusCode == {{{code}}} {
+		{{^wildcard}}if localVarHttpResponse.StatusCode == {{{code}}} { {{/wildcard}}
```
thiagoarrais: This counts on the fact that the wildcard response will be the last one. My (limited) testing confirmed this, but will that always be the case?

antihax:
Can you regenerate the test client?

modules/openapi-generator/src/main/resources/go/api.mustache
antihax: Ideally this should catch any error code so the error model is returned if one is specified. It may need to catch the default/wildcard in these cases.

If i remember correct, this should be an interface so it could be possible to make this completely generic without specifically checking each code?

thiagoarrais: > Ideally this should catch any error code so the error model is returned if one is specified. It may need to catch the default/wildcard in these cases.

My approach here was supressing the status code check for the default case so that it handles any error code. But that may overcatch if the wildcard response isn't the last one. Hence my original question.

> If i remember correct, this should be an interface so it could be possible to make this completely generic without specifically checking each code?

Sorry, I don't understand. Are you suggesting removing that check completely and just trying to unmarshall the error model based on `datatype`?

thiagoarrais:
Done in 942f65f

modules/openapi-generator/src/main/resources/go/api.mustache
```diff
@@ -299,7 +299,7 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
 			}
 			newErr.model = v
 			return {{#returnType}}localVarReturnValue, {{/returnType}}localVarHttpResponse, newErr
-		}
+		{{^wildcard}} } {{/wildcard}}
```
wing328: Please use the following instead:
```
		{{^wildcard}}
		}
		{{/wildcard}}
```
to avoid extra space/line break.

thiagoarrais:
I'm getting this double return on the generated code:

```go
// ...
	err = a.client.decode(&v, localVarBody, localVarHttpResponse.Header.Get("Content-Type"))
	if err != nil {
		newErr.error = err.Error()
		return localVarReturnValue, localVarHttpResponse, newErr
	}
	newErr.model = v
	return localVarReturnValue, localVarHttpResponse, newErr

	return localVarReturnValue, localVarHttpResponse, newErr
}
```

I'd like to avoid it, but my mustache-fu is weak. Any ideas?

thiagoarrais:
I've rebased this to latest master and ported the change go `go-experimental`. I'll start seeking a solution to [my previous question](https://github.com/OpenAPITools/openapi-generator/pull/3322#issuecomment-510977367). Any blockers here besides that?

wing328:
cc @bkabrda for review as the change targets `go-experimental`

bkabrda:
Hmm, it's weird that this is something that happens only in the `api_default.go` and not the other `api_*.go`. I'm wondering whether it might be a general problem in openapi-generator that occurs for certain responses. I'll try to dig in a bit and see if I find out something.

bkabrda:
So my understanding of the issue is this:

* Any wildcard response (that includes "4XX"-like responses and also "default") have code == 0.
* Wildcard responses can be used alongside of other specific responses.
* It's true that the go generated code won't work for any wildcard responses.

However, I think the solution presented here is not entirely correct. The problem, I think, is that if there was e.g. also a 400 response defined, it could theoretically be listed *after* the default response in the `responses` array in the Java code. That would mean that not adding the `if` clause for wildcards would actually catch a 400 response instead of the more specific 400 response handler (which would have it's code generated after the default handler).

I guess the correct solution is to both:
1. Do what this patch is doing.
2. Make sure the Java code sorts the responses in order from the more specific ones to the less specific ones. E.g. for an endpoint with "default", "4XX" and "400" responses defined, they should be listed in order "400", "4XX", "default".

If the above problem is also in templates for other languages, having the responses sorted would also help all of them.

Does this sound reasonable?

bkabrda:
Minor clarification of the above comment: it seems that the wildcard response numbers such as "4XX" actually produce invalid code (at least for Go), as they're treated as numbers, so there will be literal `4XX` put in the generated code. I think in addition to what I wrote above, we'll also need to mark these as wildcards and set their `code` to 0.

@thiagoarrais do you want to work on the points that I mentioned? I think I could find some time within ~1 week to get these fixed if you can't/don't want to work on them.

thiagoarrais:
> Make sure the Java code sorts the responses in order from the more specific ones to the less specific ones. E.g. for an endpoint with "default", "4XX" and "400" responses defined, they should be listed in order "400", "4XX", "default".

I can work on that, but I haven't looked into the Java code yet. Can you point me into the right direction?

bkabrda:
@thiagoarrais so you'll need to modify `org.openapitools.codegen.DefaultCodegen`. In the method `fromResponse`, you'll have to put `r.code = 0` for the `4XX` (and similar) wildcards - you can find the list of all possible wildcards at [1].

To modify the order of responses, you'll need to do sorting on `op.responses` in the `fromOperation` method (also on `DefaultCodegen`).

I'm somewhat new to openapi-generator myself, but I think this would be all that you need to do to get this done.

[1] https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#patterned-fields-1

thiagoarrais:
Once CI finishes this is done by me. The other wildcards (1xx, 2xx, and so on) needed a little more investigation and were left for a later PR. Please let me know if you need anything else.

bkabrda:
Right, so even though we haven't reached a complete solution yet, we certainly improved the situation here, so 👍 for me to merge it. @thiagoarrais will you be interested in taking this further in a followup PR? I can work on it if you're not interested (I'd like to solve this now to get a complete solution to the problem).

thiagoarrais:
Can't work on that right now, but maybe in a near future. Maybe we can open an issue as a placeholder for whomever finds the time first?

bkabrda:
Once this is merged, I'll open a followup issue and work on it, time permitting.

thiagoarrais:
Found some potential issues with this. Please avoid merging for the time being.

thiagoarrais:
My problem has something to raise in OpenAPI 3 specs. Still digging, but it seems like the sorting code in `DefaultCodeGen` isn't getting run or isn't effective. Any ideas?

thiagoarrais:
OK. This is ready. I had switched the sorting comparation around.

thiagoarrais:
Can't tell if the Travis failure is related to the PR. Any pointers?

thiagoarrais:
@wing328: anything else needed for going forward with this?

modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
```diff
@@ -556,14 +556,14 @@ public void testDefaultResponseShouldBeLast() {
             .addApiResponse(
                 "default", new ApiResponse().description("Default"))
             .addApiResponse(
-                "201", new ApiResponse().description("OK"))
+                "422", new ApiResponse().description("Error"))
```
wing328: @thiagoarrais I notice that you've updated the test to replace 2xx with 4xx status code.

Does it mean the original "201" status code test fail?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2564,6 +2564,11 @@ public CodegenOperation fromOperation(String path,
                     op.isResponseFile = Boolean.TRUE;
                 }
             }
+            op.responses.sort((a, b) -> {
+                int aDefault = "0".equals(a.code) ? 1 : 0;
+                int bDefault = "0".equals(b.code) ? 1 : 0;
+                return aDefault - bDefault;
+            });
```
wing328: Changes in DefaultCodegen

modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
thiagoarrais: No. 422 just seemed more realistic. 201 is a success and more likely to be equated to the default case.

thiagoarrais:
Squashed and rebased. Anything else needed for merging?

thiagoarrais:
Folks. Anything else that needs to be addressed here?

wing328:
@thiagoarrais Let me review by tomorrow ...

wing328:
Sorry for the delay. Please resolve the conflicts and I'll merge after that 🙏

thiagoarrais:
Hey, @wing328, no need to apologize. We all understand that.

Anyway... resolved conflicts and rebased. Should be good to go.

thiagoarrais:
Ran update as requested by [CircleCI](https://circleci.com/gh/OpenAPITools/openapi-generator/9455#tests/containers/2). Regarding the [Travis failure](https://travis-ci.org/OpenAPITools/openapi-generator/builds/596280447), I'm not sure it is related to this PR, since a [similar build already ran successfully](https://travis-ci.org/thiagoarrais/openapi-generator/builds/596280443) for my fork.

wing328:
CircleCI issue not related to this PR. Tests passed locally:
```
=== RUN   TestUpdateUser
--- PASS: TestUpdateUser (0.33s)
=== RUN   TestDeleteUser
--- PASS: TestDeleteUser (0.16s)
PASS
ok  	_/private/var/tmp/openapi-generator/samples/client/petstore/go	7.199s
[INFO] BUILD SUCCESS
[INFO] Total time: 01:20 min
[INFO] Finished at: 2019-10-15T01:28:26+08:00
```

wing328:
@thiagoarrais PR merged into master. Thanks for your contribution.

wing328:
@thiagoarrais  thanks for the PR, which has been included in v4.2.0 release: https://twitter.com/oas_generator/status/1189824932345069569

PR #3323, opened 2019-07-09
[typescript-fetch] Fix access token type

hello-brsd:
### PR checklist

### Description of the PR

There was a typescript error, because only `this.configuration.accessToken` function takes two arguments,` this.configuration.apiKey` doesn't. Moreover, `apiKey` is unnecessary in bearer mode, and it is used a few lines below.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -113,8 +113,8 @@ export class {{classname}} extends runtime.BaseAPI {
         }
         {{/isBasicBasic}}
         {{#isBasicBearer}}
-        if (this.configuration && (this.configuration.accessToken || this.configuration.apiKey)) {
-            const token = this.configuration.accessToken || this.configuration.apiKey;
+        if (this.configuration && this.configuration.accessToken) {
+            const token = this.configuration.accessToken;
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
```diff
@@ -113,8 +113,8 @@ export class {{classname}} extends runtime.BaseAPI {
         }
         {{/isBasicBasic}}
         {{#isBasicBearer}}
-        if (this.configuration && (this.configuration.accessToken || this.configuration.apiKey)) {
-            const token = this.configuration.accessToken || this.configuration.apiKey;
```
macjohnny: why don't we want to use apiKey here?

hello-brsd: I think the `this.configuration.apiKey` is used for` # isApiKey` block (lines 126), not for `# isBasicBearer`.

TiFu: I think that's a fair argument and this PR would make the code easier to read. 

I will mark this as a breaking change because it might affect people relying on `configuration.apiKey` being used as an accessToken.

macjohnny:
circle ci failure is unrelated

wing328:
Breaking changes without fallback should be merged into 5.0.x branch instead. I do not plan to revert this. Let's see if the users have an opinion on this change.

PR #3344, opened 2019-07-11
[JaxRS-Spec] add openApiSpecFileLocation parameter

### LINKED ISSUE #809: [JAXRS-spec] Make path for openapi.yaml configurable
Currently the path of the generated openapi.yaml is always "src/main/openapi" (see [here](https://github.com/OpenAPITools/openapi-generator/blob/7acba8a4d86b3cdadc5707345817513dad11c36f/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java#L130)). This should be configurable, because the microprofile openapi spec, for example, needs the openapi.yaml at the [META-INF](https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#421-location-and-formats) folder.

jmini:
### PR checklist

- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. Java: @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04)

### Description of the PR

Currently the OpenAPI Spec is generated to `src/main/openapi/openapi.yaml`. This is not conform with the [Eclipse MicroProfile OpenAPI](https://github.com/eclipse/microprofile-open-api/) which is the spec that defines how OpenAPI should be used on top of JAX-RS.

This PR introduces a new parameter `openApiSpecFileLocation` to be able to specify where the generated spec should be generated.

* Default value: `src/main/openapi/openapi.yaml` (no change for existing users)
* Possible value `src/main/resources/META-INF/openapi.yaml` to be conform with the MP-OpenAPI spec
* In order to not generate the spec, the value can be set to `null` or empty string `""`

Fixes #809

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
@@ -37,12 +37,14 @@ public class JavaJAXRSSpecServerCodegen extends AbstractJavaJAXRSServerCodegen {
     public static final String GENERATE_POM = "generatePom";
     public static final String USE_SWAGGER_ANNOTATIONS = "useSwaggerAnnotations";
     public static final String JACKSON = "jackson";
+    public static final String OPEN_API_SPEC_FILE_LOCATION = "openApiSpecFileLocation";
 
     private boolean interfaceOnly = false;
     private boolean returnResponse = false;
     private boolean generatePom = true;
     private boolean useSwaggerAnnotations = true;
     private boolean useJackson = false;
+    private String openApiSpecFileLocation = "src/main/openapi/openapi.yaml";
 
     private String primaryResourceName;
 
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
```diff
@@ -134,18 +141,38 @@ public void processOpts() {
                     (sourceFolder + '/' + invokerPackage).replace(".", "/"), "RestApplication.java"));
         }
 
-        supportingFiles.add(new SupportingFile("openapi.mustache",
-                "src/main/openapi",
-                "openapi.yaml")
-        );
+        if(StringUtils.isNotEmpty(openApiSpecFileLocation)) {
+            int index = openApiSpecFileLocation.lastIndexOf('/');
```
wing328: @jmini shall we use `File.separator` instead of `/` so that the option also works in Windows?

jmini: I think even on windows the users will use `openApiSpecFileLocation = some/location/openapi.yaml` as input.
(think of a maven/gradle build that works on both windows and linux/macos)

That will result of creating a file at `output-folder\some\location\openapi.yaml`.

As you can see I have a test:
https://github.com/OpenAPITools/openapi-generator/blob/516163b509e5c15f18e6cdc3a9885291a7b30417/modules/openapi-generator/src/test/java/org/openapitools/codegen/java/jaxrs/JavaJAXRSSpecServerCodegenTest.java#L236-L255
Line 240 the variable is set to `"src/main/resources/META-INF/openapi.yaml"`
Line 255 there is a check that the file is created.

This is already working on windows (appveyor CI is green)

wing328: > I think even on windows the users will use openApiSpecFileLocation = some/location/openapi.yaml as input.

👌 Let's see.

PR #1394, opened 2018-11-07
Add group parameters support in TypeScript Fetch

wing328:
### PR checklist

- [ ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.4.x`, `4.0.x`. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

- prototype to show group parameter support in TS fetch

TODO:
- update all method signature to support x-group-parameters
- update docstring about using group parameters

PR #3351, opened 2019-07-12
Fix #3349 : [typescript-fetch] Proposition to follow spec

### LINKED ISSUE #3349: [BUG] Typescript-fetch generator not respecting the specification
#### Bug Report Checklist

##### Description

There was a big refactoring in the typescript-fetch generator that was introduced with version 4.X.X.  A lot of stuff has been added and it's good work. Unfortunately, there is one big problem. The new generated code is not following the specification which break every software using the previous fetch generator. 

Let me explain with an example:

If you look at the test (which currently passes), you can see that the `addPet` function doesn't receive a `Pet` object as expected. 
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts#L19

Instead, it uses an interface called `AddPetRequest` which doesn't exist at all in the spec.
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-fetch/builds/default/apis/PetApi.ts#L25

You can see in the spec here that it should receive a simple `Pet` object.
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/petstore.yaml#L36

Latest 4.0.3 and master

Using default `petstore.yaml` used for the samples.

`./bin/typescript-fetch-petstore-all.sh`

##### Steps to reproduce

Just generate the code using the `sh` script.

##### Related issues/PRs

N/A

##### Suggest a fix

We need to remove the `Request` interface. The code should respect the spec.

For example, the test should be like this:

```
it('should add and delete Pet', () => {
      return api.addPet(fixture).then(() => {
    });
});
```

instead of this:

```
it('should add and delete Pet', () => {
       return api.addPet({ body: fixture }).then(() => {
   });
});
```

JFCote:
### PR checklist

### Description of the PR

This is to fix #3349. I removed the interface wrapping the real signature of the call. I regenerated the samples + adapted the tests.


modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
```diff
@@ -39,40 +27,32 @@ export class {{classname}} extends runtime.BaseAPI {
...
-
-        {{/required}}
-        {{/allParams}}
+    async {{nickname}}Raw({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}}): Promise<runtime.ApiResponse<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}>> {
         const queryParameters: runtime.HTTPQuery = {};
 
         {{#queryParams}}
         {{#isListContainer}}
-        if (requestParameters.{{paramName}}) {
+        if ({{paramName}}) {
```
macjohnny: this could lead to many naming conflicts with symbols in the scope of the file

JFCote: I'm in the process of proposing a new version of this PR based on @TiFu comment on the issue. Should push this today! :)
Thanks for the review

JFCote:
@TiFu and @macjohnny  New version is ready for review. Let me know if changes are needed! Hope everything is ok :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -24,19 +24,18 @@
...
 import java.util.Map;
 
 public class TypeScriptFetchClientCodegen extends AbstractTypeScriptClientCodegen {
 
     public static final String NPM_REPOSITORY = "npmRepository";
     public static final String WITH_INTERFACES = "withInterfaces";
+    public static final String USE_SINGLE_REQUEST_PARAMETER = "useSingleRequestParameter";
 
     protected String npmRepository = null;
+    protected boolean useSingleRequestParameter = false;
```
macjohnny: default should be `true` in order to avoid a breaking change
```suggestion
    protected boolean useSingleRequestParameter = true;
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -56,6 +55,7 @@ public TypeScriptFetchClientCodegen() {
 
         this.cliOptions.add(new CliOption(NPM_REPOSITORY, "Use this property to set an url your private npmRepo in the package.json"));
         this.cliOptions.add(new CliOption(WITH_INTERFACES, "Setting this property to true will generate interfaces next to the default class implementations.", SchemaTypeUtil.BOOLEAN_TYPE).defaultValue(Boolean.FALSE.toString()));
+        this.cliOptions.add(new CliOption(USE_SINGLE_REQUEST_PARAMETER, "Setting this property to true will generate functions with a single parameter containing all parameters", SchemaTypeUtil.BOOLEAN_TYPE).defaultValue(Boolean.FALSE.toString()));
```
TiFu: Maybe we can phrase this a little bit clearer?

e.g. 
`Setting this property to true will generate functions with a single argument containing all API endpoint parameters instead of one argument per parameter.`

JFCote: Do you want me to change the name of the option?
I see in other thread about this that some people call it "GroupParameter" or "CollectParameter"?
That would require change everywhere in my PR.

JFCote:
@TiFu @macjohnny It should be good. Let me know if it is missing something.

samples/client/petstore/typescript-fetch/builds/with-npm-version/apis/UserApi.ts
```diff
@@ -276,9 +276,9 @@ export class UserApi extends runtime.BaseAPI {
    /**
     * Logs out current logged in user session
     */
-    async logoutUser(): Promise<void> {
+     async logoutUser(): Promise<void> {
```
macjohnny: indentation should be corrected

JFCote:
@wing328 or anyone else know why it doesn't pass in `circleci` but works perfectly on my local computer?

image

macjohnny:
@JFCote is the file permission mode `755` for 
https://github.com/OpenAPITools/openapi-generator/blob/8e2f5b8853c996f4e47f8b76173c3619306fdc26/bin/typescript-fetch-petstore-all.sh
and
https://github.com/OpenAPITools/openapi-generator/blob/8e2f5b8853c996f4e47f8b76173c3619306fdc26/bin/typescript-fetch-petstore-multiple-parameters.sh
?

JFCote:
@macjohnny I had done it but did a rename... It seems the rename reset the permission. I've done a new commit with the 755 permissions. We'll see if it works now!

macjohnny:
@JFCote you need to run https://github.com/OpenAPITools/openapi-generator/blob/master/bin/utils/export_docs_generators.sh

JFCote:
Thanks! Will do that now

docs/generators/fsharp-giraffe-server.md
```diff
@@ -15,7 +15,7 @@ sidebar_label: fsharp-giraffe-server
 |packageName|F# module name (convention: Title.Case).| |OpenAPI|
 |packageVersion|F# package version.| |1.0.0|
 |packageGuid|The GUID that will be associated with the C# project| |null|
-|sourceFolder|source folder for generated code| |OpenAPI/src|
+|sourceFolder|source folder for generated code| |OpenAPI\src|
```
macjohnny: shouldn't this be `OpenAPI/src`

JFCote: Was auto-generated, not my change. I can revert it back in needed

JFCote: Seems to make the test fail... will revert that file

JFCote:
@macjohnny If everything is ok with you, you can merge (or I can do it), whatever works better for you. Thanks!

wing328:
@JFCote nice work. Instead of adding an additional option, what about using `x-group-parameters` instead like what we did for PHP and Java client generator?

Ref: https://github.com/OpenAPITools/openapi-generator/pull/1394

macjohnny:
@wing328 I think it is good to have an additional option instead of the `x-group-parameters`, since this should be the same for all api endpoints generated

wing328:
@macjohnny I agree some users simply want to apply this option to all endpoints but there were use cases before in which users do not want to apply parameter grouping to method with just one parameter. That's why we come up with the approach x-group-parameters to provide flexibility meeting different requirement.

(To enable all endpoints, one can define x-group-parameters at the top level of the spec)

TiFu:
If that is implemented, I'd expect there to be a CLI option which sets ``x-group-parameters` to true for all endpoints which have less than `n` parameters.

JFCote:
@wing328 @macjohnny @TiFu The problem I see with the suggestion is that right now, in version 4.X.X, it is already grouped, **by default**. This was a breaking change compared to version 3.X.X which I was using. Because of this, I created this option so people like me who like to have all their parameters can use the new generator option.

If I understand well the `x-group-parameters`, it would be false by default, unless specified in the spec, which would do yet another breaking change for people already using version 4.X.X.

I will let you guys decide and do a new PR if needed because I'm going on vacation for a couple of weeks and won't be online that much.

Thanks!

PR #816, opened 2018-08-14
Added server variable support

TiFu:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `3.3.x`, `4.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

See #793 

I implemented a solution similar to `authMethods` in supportingFiles JSON (see generated json below).

Spec: https://gist.github.com/TiFu/0934b362f25c66bc2b758f33e15520fc
Generated Supporting Files JSON: https://gist.github.com/TiFu/919a8888837e4e24fd32a05dc8d289f9

Generated under the root object (use `-DdebugSupportingFiles` to print the full mustache template used):

```
 "servers" : [ {
    "url" : "http://petstore.swagger.io:{port}/{basePath}",
    "description" : "The production API server",
    "variables" : [ {
      "name" : "port",
      "defaultValue" : "8443",
      "enumValues" : [ "8443", "443" ]
    }, {
      "name" : "basePath",
      "defaultValue" : "v2"
    } ]
  }, {
    "url" : "http://petstore.swagger.io",
    "description" : "The production API server",
    "variables" : [ ]
  } ],
```
# Question

Do I have to regenerate any samples? This is a non-breaking change because I only added a new property to the JSON (which is passed to mustache) for supportingFiles.

CC @wing328  @jmini

PR #3363, opened 2019-07-15
[core][maven][gradle] User-defined server variable substitutions

jimschubert:
### PR checklist

### Description of the PR

This provides an initial capability for users to define values for server variable substitutions.

Given a spec with a server URL definition such as:

```
{
  "openapi" : "3.0.1",
  "info" : {
    "title" : "OpenAPI Petstore",
    "description" : "This is a sample server Petstore server. For this sample, you can use the api key `special-key` to test the authorization filters.",
    "license" : {
      "name" : "Apache-2.0",
      "url" : "http://www.apache.org/licenses/LICENSE-2.0.html"
    },
    "version" : "1.0.0"
  },
  "servers" : [ {
      "url": "https://{api}.{server}/{basePath}",
      "description": "The production API server",
      "variables": {
        "api": {
          "default": "petstore",
          "description": "This is the specific API to work with."
        },
        "server": {
          "default": "swagger.io"
        },
        "basePath": {
          "default": "v2"
        }
      }
    }],
…
```

The default will be `https://petstore.swagger.io/v2`, but we can now generate with user-declared options. For example:

*defaults* (generates targeting `https://petstore.swagger.io/v2`) :

```
generate -g kotlin -i .bak/openapi/openapi-servers.json -o .bak/server-variables/defaults
```

*user defined values* (generates targeting `https://petstore.local/api`):

```
openapi-generator generate -g kotlin \
    -i .bak/openapi/openapi-servers.json \
    -o .bak/server-variables/ours \
    --server-variables=api=petstore,basePath=api,server=local
```

The results:
```
.bak/server-variables/user/s/m/k/o/o/c/a/UserApi.kt:28:class UserApi(basePath: kotlin.String = "https://petstore.local/api") : ApiClient(basePath) {
.bak/server-variables/user/s/m/k/o/o/c/a/PetApi.kt:29:class PetApi(basePath: kotlin.String = "https://petstore.local/api") : ApiClient(basePath) {
.bak/server-variables/user/s/m/k/o/o/c/a/StoreApi.kt:28:class StoreApi(basePath: kotlin.String = "https://petstore.local/api") : ApiClient(basePath) {
.bak/server-variables/defaults/s/m/k/o/o/c/a/UserApi.kt:28:class UserApi(basePath: kotlin.String = "https://petstore.swagger.io/v2") : ApiClient(basePath) {
.bak/server-variables/defaults/s/m/k/o/o/c/a/PetApi.kt:29:class PetApi(basePath: kotlin.String = "https://petstore.swagger.io/v2") : ApiClient(basePath) {
.bak/server-variables/defaults/s/m/k/o/o/c/a/StoreApi.kt:28:class StoreApi(basePath: kotlin.String = "https://petstore.swagger.io/v2") : ApiClient(basePath) {
```

The design/usage of `URLPathUtils.getServerURL` was a little weird (it offloads generation logic to a static utility). Rather than rewriting this, I've simply passed the overrides to it. I anticipate some URLs being missed due to the static method call, but we can address these as found. I hope to have workflows rewritten by the 5.0 release and will refactor `URLPathUtils.getServerURL` at some point.


jimschubert:
This is a 3.x enhancement (for server variable substitutions). See: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#server-object-example

jmini:
How does you change affect the `enum` case in server variables?

```json
{
  "openapi" : "3.0.1",
  "info" : {
    "title" : "OpenAPI Petstore",
    "description" : "This is a sample server Petstore server. For this sample, you can use the api key `special-key` to test the authorization filters.",
    "license" : {
      "name" : "Apache-2.0",
      "url" : "http://www.apache.org/licenses/LICENSE-2.0.html"
    },
    "version" : "1.0.0"
  },
  "servers" : [ {
      "url": "https://{variant}.example.com/",
      "description": "The production API server",
      "variables": {
        "variant": {
          "default": "v3",
          "enum": ["v1", "v2", "v3"],
          "description": "The version you want to work with"
        }
      }
    }],
    "paths" : {}
}
```

I was thinking that a case like this create 3 server-urls:

* `https://v1.example.com/`
* `https://v2.example.com/`
* `https://v3.example.com/`

But maybe this is not the way it works.

---
Second question related to `enum`:

Is there some validation that the value provided are compatible with what is defined in the OpenAPI spec? According to the `enum` documentation in server-variables:

> An enumeration of string values to be used if the substitution options are from a limited set.

so in my previous example, it is not allowed to set the value to `variant=v4` (or at least it should log a warning)

jimschubert:
@jmini thanks for the review. The enums are allowed values.  Do you think we should warn if the value doesn't exist? I'd like to use the strict spec option for that validation, but I would need to do the rework of the static helpers to do it in a good way. A warning log might be the lowest impact. I can add that later today.

To your enum question, your example would work to create a constrained set of server urls only if there's only enum options, and if we strictly stick to only those values. It's possible that someone may expose a v4 preview/beta version without updating the public doc in your example, and I don't want to be so strict that users wouldn't be able to generate the undocumented version.

jimschubert:
I would add, as well, we currently only pull the first server url and don't do any construction of server sets which involves enums (at least not where I can tell)

jimschubert:
I had some stuff come up. Will try to add that warning in the next day or so.

TiFu:
I added support for server variables in `DefaultCodegen` in #816 by adding the servers & variables to the generated supporting files json. What effect does this PR have on that implementation?

jimschubert:
@tifu looking over your changes, it seems we'll want to figure out how to get the options passed by the user to the templates. We could do that on a separate PR.

This PR allows users to define the values for a variable, so that the baseUrl which is generally passed into an API client targets the desired version of the server URL.

jimschubert:
@TiFu after digging into how I'd pass these values to the template, I began to see your concern. I need to rename `serverVariables()` in this PR to `serverVariableOverrides()` to make it clear these are user values, not spec values.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4953,14 +4958,34 @@ public boolean isDataTypeString(String dataType) {
...
+
+            // Sets the override value for a server variable pattern.
+            // NOTE: OpenAPI Specification doesn't prevent multiple server URLs with variables. If multiple objects have the same
+            //       variables pattern, user overrides will apply to _all_ of these patterns. We may want to consider indexed overrides.
+            if (variableOverrides != null && !variableOverrides.isEmpty()) {
+                String value = variableOverrides.getOrDefault(variableEntry.getKey(), variable.getDefault());
+                codegenServerVariable.value = value;
+
+                if (enums != null && !enums.isEmpty() && !enums.contains(value)) {
+                    LOGGER.warn("Variable override of '{}' is not listed in the enum of allowed values ({}).", value, StringUtils.join(enums, ","));
```
jimschubert: @jmini do you think this will be sufficient?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4953,14 +4958,34 @@ public boolean isDataTypeString(String dataType) {
...
             codegenServerVariable.defaultValue = variable.getDefault();
             codegenServerVariable.description = escapeText(variable.getDescription());
-            codegenServerVariable.enumValues = variable.getEnum();
+            codegenServerVariable.enumValues = enums;
             codegenServerVariable.name = variableEntry.getKey();
+
+            // Sets the override value for a server variable pattern.
+            // NOTE: OpenAPI Specification doesn't prevent multiple server URLs with variables. If multiple objects have the same
+            //       variables pattern, user overrides will apply to _all_ of these patterns. We may want to consider indexed overrides.
+            if (variableOverrides != null && !variableOverrides.isEmpty()) {
```
jimschubert: @TiFu Any objection to this approach? I think it adds as little impact to the templates as possible. For example, if we add support for defining user overrides for multiple templated server urls, we'd change logic here and that would have no effect on existing templates.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jimschubert: Similar log message in `URLPathUtils`. Looks like a refactor of that type would do good to merge with this code, or maybe just use `fromServers` and retrieve the first server URL to match existing logic. Again, that's a good task for another PR.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
TiFu: No, that sounds fine. Sorry for the late response.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jmini: Yes, having a warning is good to me

coder82:
Wow this is exactly what is needed.

jimschubert:
Moved maven implementation to #3609 due to CI issues with resolving the maven plugin when it is a SNAPSHOT.

PR #3403, opened 2019-07-19
Fix NPM build for Typescript-fetch

JFCote:
### PR checklist

### Description of the PR

This PR fixes 2 things:
1) The npm build was "working" but was not usable as a package because there was no output generated (Fix by adding the .npmignore file). Because of this, when you used the package via npm, webpack was always telling you that you need a specific loader for `ts` files. A real package should contains compiled code (.js).
2) I followed a tutorial on how to do a good typescript package so I followed convention of putting "source" code in the "src" folder instead of directly at the root.

I followed this tutorial to do a good npm package: https://medium.com/cameron-nokes/the-30-second-guide-to-publishing-a-typescript-package-to-npm-89d93ff7bccd


modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
```diff
@@ -6,7 +6,7 @@
   "main": "./dist/index.js",
```
macjohnny: is this path still valid?

JFCote: Yes it is.

macjohnny: but how can that be? i would expect all compiled files to be in `dist/src/`

macjohnny:
@JFCote can you please add the `.npmignore` files to the samples?

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
JFCote: Compiled files are in `dist` folder.
Source files are in `src` folder.

JFCote:
@macjohnny Currently in vacation, will check all of this when I'm back! Thanks for the review!

modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache
macjohnny: but the `tsconfig.json` does not explicitly configure `src` as the root directory, so I would expect something like `/src/index.ts` to be compiled into `/dist/src/index.js`

JFCote: I followed a real life example here: https://github.com/ccnokes/dom-event-utils
I guess the `src` for the root directory is "default" value so it doesn't have to be specified.

I've commited files that should fix the circleCI problem. Let me know if there is other problems! Thanks

macjohnny: ok i tested it, I get `dist/index.js`

Bessonov:
@JFCote 
> I followed a tutorial on how to do a good typescript package so I followed convention of putting "source" code in the "src" folder instead of directly at the root.

Although we use `src` folder by default, this is a breaking change for us.

JFCote:
@Bessonov Can you explain a little bit more how this is a breaking change? If you use npm publish, you should'nt have any problem importing the resulting package, even if the structure has changed because webpack take care of everything.

Let us know what is your use case. Thanks

Bessonov:
We don't use it as a whole module, but integrate it in own api module wit some utilities and configuration. Basically, we have follwong structure:

```
- myApi
  - src
    - index.ts <-- exports only relevant and configured parts
    - configuration.ts
    - fetch <-- here we place the generated artifacts
```

After this change the artifacts are not under `src/fetch`, but `src/fetch/src`. Of course we can handle it with some post processing, but well, that's a definition of breaking changes :)

JFCote:
@Bessonov I understand now. The changes was supposed to be mostly for the npm version of the generator but it seems it has created some regression for use case like yours.

@macjohnny What do you suggest we do for this?

rob-deutsch:
@Bessonov what postprocessing did you implement for this?

Like you, I was also using openapi-generator to integrate it into an existing npm package.

Bessonov:
@rob-deutsch because all generators has some flaws, we use a run scripts like (pseudo code):

```bash
#!/bin/bash

# pre processing
rm -rf ...

# do work
docker run generator ....

# post processing
# remove files
for path in ".gitignore" ".openapi-generator-ignore" "tsconfig.json" ".openapi-generator"; do
	rm -rf "${rest_api}/src/fetch/${path}"
done
# new version of generator create resources in src folder
mv ${rest_api}/src/fetch/src/* ${rest_api}/src/fetch
rmdir ${rest_api}/src/fetch/src
```
So, no magic.

rob-deutsch:
Thanks @Bessonov. That's more elegant than my `import * from './service/src';` solution

rob-deutsch:
That being said, I seem to be running into issues...

For some odd reason webpack/tsc is sometimes inserting the following line into the compiled files:

`Object.defineProperty(exports, "__esModule", { value: true });`

This line crashes my browser with:

`runtime.ts:74 Uncaught ReferenceError: exports is not defined`

What's odd is that this line isn't always added. Only sometimes does webpack/tsc choose to add this line.

PR #3405, opened 2019-07-20
Ruby client faraday

dkliban:
### Description of the PR

This patch adds ability to specify which library to use when generating a Ruby Client. The two options are 'typhoeus' and 'faraday'. The implementation with 'typhoeus' was already present and this patch adds support for 'faraday'. 

I have not tested this with the SSL option or token Auth.

I did test that the client can use basic Auth and perform both JSON and multipart/form-encoded requests.


modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
```diff
@@ -0,0 +1,392 @@
...
+
+    def self.default
+      @@default ||= ApiClient.new
+    end
+
+    # Call an API with given options.
+    #
+    # @return [Array<(Object, Integer, Hash)>] an array of 3 elements:
+    #   the data deserialized from response body (could be nil), response status code and response headers.
+    def call_api(http_method, path, opts = {})
```
dkliban: This template is very similar to api_client.mastache. The main difference is the implementation of call_api() and build_request(). Would it make sense to combine the two templates and simply pick the implementation of these methods based on the library selected?

modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
```diff
@@ -0,0 +1,392 @@
...
+      connection = Faraday.new(:url => config.base_url) do |conn|
```
jcarres-mdsol: @dkliban  Thanks so much for this PR!
I wonder if you have considered how to add middlewares here to fully use the power of Faraday. For instance one could add caching or tracing or whatever in here. I think that's one of the best things about moving to Faraday

modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
```diff
@@ -0,0 +1,392 @@
...
+              fail ApiError.new(:code => 0,
+                                :message => response.return_message)
+            else
+              fail ApiError.new(:code => response.status,
+                                :response_headers => response.headers,
+                                :response_body => response.body),
+                   response.reason_phrase
+            end
+          end
+      rescue Faraday::TimeoutError
```
jcarres-mdsol: I'm torn about making all failign cases the same Exception class, it may be difficult to distinguish among errors if that's what the caller wants to do. Also here you probably want to deal with Net::ReadTimeout and Faraday::ConnectionFailed

modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
jcarres-mdsol: This file is ~400 lines.  Keeping both templates in sync when changes happen outside those methods will be a PITA I think. 
So I'd say for long term maintainability, it makes sense to merge them but I'm an outsider of this project, I hope maintainers can chime in.

jimschubert: The generator supports a `library` option, which allows library specific overrides for individual files. Using this may make the logic and template separation clearer. The Java client generator is a good example. Let us know if you'd like to move toward that option.

wing328: Yes, I think there's a library option added by this PR: 

https://github.com/OpenAPITools/openapi-generator/pull/3405/files#diff-23d62a3ed9c41293f930468ed3ee74a8R27

so the other 2 options (Ruby HTTP library names) are not needed

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
```diff
@@ -141,6 +143,18 @@ public RubyClientCodegen() {
         cliOptions.add(new CliOption(CodegenConstants.HIDE_GENERATION_TIMESTAMP, CodegenConstants.HIDE_GENERATION_TIMESTAMP_DESC).
                 defaultValue(Boolean.TRUE.toString()));
 
+        cliOptions.add(CliOption.newBoolean(FARADAY, "Faraday (https://github.com/lostisland/faraday)"));
+        cliOptions.add(CliOption.newBoolean(TYPHOEUS, "Typhoeus (https://github.com/typhoeus/typhoeus)"));
```
wing328: @dkliban I don't think we need line 146, 147 given that we have the library option.

modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
wing328: > This template is very similar to api_client.mastache

I'll see if I can combine the two ...

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
```diff
@@ -141,6 +143,18 @@ public RubyClientCodegen() {
...
+
+        supportedLibraries.put(FARADAY, "Faraday (https://github.com/lostisland/faraday)");
+        supportedLibraries.put(TYPHOEUS, "Typhoeus >= 1.0.1 (https://github.com/typhoeus/typhoeus)");
```
wing328: @dkliban I would mention the "Typhoeus" support is beta, e.g.
```
Typhoeus >= 1.0.1 (https://github.com/typhoeus/typhoeus) (Beta support)
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
```diff
@@ -141,6 +143,18 @@ public RubyClientCodegen() {
...
+
+        CliOption libraryOption = new CliOption(CodegenConstants.LIBRARY, "HTTP library template (sub-template) to use");
+        libraryOption.setEnum(supportedLibraries);
+        // set FARADAY as the default
+        libraryOption.setDefault(FARADAY);
```
wing328: What about using Faraday as the default later (say 4.2.0) after we collect more feedback from the users?

modules/openapi-generator/src/main/resources/ruby-client/faraday_api_client.mustache
dkliban: That makes sense. I'll remove it

> Yes, I think there's a library option added by this PR:
> 
> https://github.com/OpenAPITools/openapi-generator/pull/3405/files#diff-23d62a3ed9c41293f930468ed3ee74a8R27
> 
> so the other 2 options (Ruby HTTP library names) are not needed

Makes sense. I'll remove them.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
dkliban: That would be great.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java
```diff
@@ -141,6 +143,18 @@ public RubyClientCodegen() {
...
+        cliOptions.add(libraryOption);
+        setLibrary(FARADAY);
```
dkliban: should the parameter here be the value of libraryOption instead of FARADAY?

wing328: Yes, please set it to TYPHOEUS

bin/ruby-petstore-faraday.json
```diff
@@ -0,0 +1,5 @@
+{
+  "gemName": "petstore",
```
wing328: I think `typhoeus` is the default so we need to add `"library": "faraday",` here if I'm not mistaken.

wing328: I'll take care of it in another PR.

wing328:
I've filed https://github.com/OpenAPITools/openapi-generator/pull/3520 for various improvements

PR #3437, opened 2019-07-24
[typescript-rxjs]: support reponseType blob

denyo:
### PR checklist

### Description of the PR

The goal is to generate the correct `responseType` which is passed to `rxjs/ajax`. 
By default this is `json` as visible here: <https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/dom/AjaxObservable.ts#L167)> but it can be overriden.

I am not too sure about all the possible types but I found `json`, `text`, `blob` and `arraybuffer` to be supported by `rxjs/ajax` based on <https://github.com/Reactive-Extensions/RxJS-DOM/pull/121/files#diff-d3b897234759c07ed79441f35e56b67fR22>. Perhaps it's also possible to pass `document` and `ms-stream` as described there for completeness: <https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType>.

The one that I personally need for a project is `blob` in order to download an excel file.

The relevant lines for this in the `api.json` are:
```
"/api/export": {  
    "get": {
        "produces": ["*/*"],
        "responses": {
            "200": {
                "description": "OK",
                "schema": {  
                    "type": "string",
                    "format": "byte" // <–– here
                }
            }
        }
    }
}
```

However `AbstractTypeScriptClientCodegen` maps this to a string because of `typeMapping.put("ByteArray", "string");`.
I tried removing this via `this.typeMapping.remove("ByteArray");` in `TypeScriptRxjsClientCodegen` but then I end up with another import being generated for a model `ByteArray` which does not exist.

Here are the changes on the call:
```
-    exportUsingGET = (): Observable<object> => {
-        return this.request<object>({
+    exportUsingGET = (): Observable<ByteArray> => {
+        return this.request<ByteArray>({
             path: '/api/export',
             method: 'GET',
             responseType: 'blob' // <–– that's the one I need
         });
     }
```

In order to solve this I need some input on how to get there.
Or asked differently what's the reason for the mapping of `ByteArray` to `string`?
Or any how to figure out the appropriate `responseType`?

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
@@ -97,6 +97,7 @@ export class BaseAPI {
             method: requestOpts.method,
             headers: requestOpts.headers,
             body: requestOpts.body instanceof FormData ? requestOpts.body : JSON.stringify(requestOpts.body),
+            responseType: requestOpts.responseType || 'json'
         };
     }
 
@@ -149,6 +150,7 @@ export interface RequestOpts {
     headers?: HttpHeaders;
     query?: HttpQuery;
     body?: HttpBody;
+    responseType?: string;
 }
 
 export const encodeURI = (value: any) => encodeURIComponent(String(value))
--- a/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
@@ -150,7 +150,7 @@ export interface RequestOpts {
     headers?: HttpHeaders;
     query?: HttpQuery;
     body?: HttpBody;
-    responseType?: string;
+    responseType?: 'json' | 'blob' | 'arraybuffer' | 'text';
 }
 
 export const encodeURI = (value: any) => encodeURIComponent(String(value))
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
@@ -306,6 +306,7 @@ public class TypeScriptRxjsClientCodegen extends AbstractTypeScriptClientCodegen
             this.hasProduces = o.hasProduces;
             this.hasParams = o.hasParams;
             this.hasOptionalParams = o.hasOptionalParams;
+            this.hasRequiredParams = o.hasRequiredParams;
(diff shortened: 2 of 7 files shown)


modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -149,6 +150,7 @@ export interface RequestOpts {
     headers?: HttpHeaders;
     query?: HttpQuery;
     body?: HttpBody;
+    responseType?: string;
```
macjohnny: can we give a set of possible values here?

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -97,6 +97,7 @@ export class BaseAPI {
             method: requestOpts.method,
             headers: requestOpts.headers,
             body: requestOpts.body instanceof FormData ? requestOpts.body : JSON.stringify(requestOpts.body),
+            responseType: requestOpts.responseType || 'json'
```
macjohnny: See
https://github.com/OpenAPITools/openapi-generator/blob/2c4fdd079207374822269b58b4f7a93378456b3c/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L139
and
https://github.com/OpenAPITools/openapi-generator/blob/2c4fdd079207374822269b58b4f7a93378456b3c/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L339-L341
to have an example how the response type is determined in the angular generator.

denyo: Thanks for the feedback. I will look into this but I have a feeling that these are the missing pieces.

denyo:
I think I got it for a spec like 
```
"/api/export": {  
    "get": {
        "produces": ["*/*"],
        "responses": {
            "200": {
                "description": "OK",
                "schema": {  
                    "type": "file", // <–– crucial 
                }
            }
        }
    }
}
```

the generated code then is
```
exportUsingGET = (): Observable<Blob> => {
        return this.request<Blob>({
            path: '/api/export',
            method: 'GET',
            responseType: 'blob'
        });
    }
```

The generated samples don't show this.

macjohnny:
@denyo if you have tested it and confirm it works, we can merge it.

macjohnny:
@denyo the CI is happy, can we merge it?

denyo:
> @denyo the CI is happy, can we merge it?

Yes.

wing328:
@denyo thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

Issue #1159, opened 2018-10-03
Add a OCaml client generator


wing328:
##### Description

We would like to create an [OCaml](https://ocaml.org/) client generator so that OCaml developers can find it more easily to consume REST APIs.

Similar to how some other generators (e.g. R, Lua, PowerShell) are created, we will use the reverse engineering process to create the OCaml client generator.

- create OCaml petstore samples (implemented at least 2 endpoints to start with: `getPetById`, `addPet`)
- create the templates and eng
- test the auto-generated OCaml code and revise the OCaml generator, templates 

cc @andrenth

##### openapi-generator version

Latest master

##### Suggest a fix/enhancement

If anyone wants to help on this (e.g. testing, creating templates, etc), please reply to let us know.

wing328:
Added via #3446

For issues/feedback, please open an issue for tracking.

PR #3444, opened 2019-07-24
Ocaml client

cgensoul:
#1159 Add OCaml client code generation support

This PR answers #1159 and adds OCaml Client code generation to openapi-generator. The generated OCaml project relies on the dune tool for the build process and uses :
- cohttp as the http client lib using the lwt_unix backend (lwt is a future lib for OCaml, a competing future lib is Async which Cohttp supports as well but this generator only target the lwt implementation for now)
- yojson as the JSon serialization lib
- ppx_deriving_yojson for automatic generation of JSon serializers and deserializers for generated types
- ppx_deriving.std for automatic generation of string convertion functions for generated types

It supports headers params, path params, query params, form params and body params. It currently has no support for authentication, nor file upload.

It generates one model file per OAS definition except for OAS enum definitions for which no model file is generated. Enums are treated globally and are all generated in a enums.ml file that contains one type per distinct enumeration present in the input file, even for enumerations that are not proper definitions but just local restrictions on a string parameter of an operation.

For APIs files, both an implementation (.ml) and an interface (.mli) files are generated.


wing328:
@cgensoul thanks for the PR. We definitely welcome the OCaml client generator. Is there anything we can help with this PR?

cgensoul:
Hi,

The PR is on its way. I just wanted to add a few cleanup commits before
submitting it. I did not squash all my commits into a big one before the
PR, so let me know if it's a problem. It's the first time I am using github
and the PR mechanism so I am not sure of what is the usual way of doing
things in this regard.

Christophe,

wing328:
> I did not squash all my commits into a big one before the PR, so let me know if it's a problem

Not a problem at all. We'll squash it as part of the merge.

wing328:
>  It's the first time I am using github and the PR mechanism so I am not sure of what is the usual way of doing things in this regard.

No problem. If you need help with PR or git in general, please let us know.

cgensoul:
Yes thanks. Well I submitted it, let me know if something needs to be done
differently.

PR #3446, opened 2019-07-24
[OCaml] new client generator

cgensoul:
#1159 Add OCaml client code generation support

This PR answers #1159 and adds OCaml Client code generation to openapi-generator. The generated OCaml project relies on the dune tool for the build process and uses :
- cohttp as the http client lib using the lwt_unix backend (lwt is a future lib for OCaml, a competing future lib is Async which Cohttp supports as well but this generator only target the lwt implementation for now)
- yojson as the JSon serialization lib
- ppx_deriving_yojson for automatic generation of JSon serializers and deserializers for generated types
- ppx_deriving.std for automatic generation of string convertion functions for generated types

It supports headers params, path params, query params, form params and body params. It currently has no support for authentication, nor file upload.

It generates one model file per OAS definition except for OAS enum definitions for which no model file is generated. Enums are treated globally and are all generated in a enums.ml file that contains one type per distinct enumeration present in the input file, even for enumerations that are not proper definitions but just local restrictions on a string parameter of an operation.

Note that generating one model file per OAS definition will work as long as there are no mutually recursive OAS definitions in the input file. Otherwise the generated OCaml code will not compile since it will generate mutually recursive compilation units which is not supported by the language.

For APIs files, both an implementation (.ml) and an interface (.mli) files are generated.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/OCamlClientCodegen.java
```diff
@@ -0,0 +1,703 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ * Copyright 2018 SmartBear Software
```
wing328: @cgensoul please remove line 3 as per https://opensource.stackexchange.com/questions/7300/copyright-notice-in-the-file-header-apache-v2-license/7301#7301

wing328:
Don't worry about the CircleCI failure. I'll take care of it before merge.

cgensoul:
Ok. I removed the smart bear copyright.

I noticed that the pet store sample that is processed during the tests
describes a GET operation with a body. Is this on purpose ? If it is what
is the expected outcome of the generation ? For now in the OCaml case it
generate code that doesn’t compile which seems fine to me since this kind
of operation doesn’t really make sense but I’d like to hear how you
intended generators to respond to this input ?

Le jeu. 25 juil. 2019 à 08:20, William Cheng <notifications@github.com> a
écrit :

wing328:
> I noticed that the pet store sample that is processed during the tests
describes a GET operation with a body

Which operation? What's the operationId?

cgensoul:
It's the testEnumParameters operation which is described like this :

/fake:
   get:
      tags:
        - fake
      summary: To test enum parameters
      description: To test enum parameters
      operationId: testEnumParameters
      consumes:
        - "application/x-www-form-urlencoded"

wing328:
Per https://stackoverflow.com/questions/978061/http-get-with-request-body, GET operation can have a request body and I remember there were discussion on that before. That's why we've added a test case to the spec covering it moving forward.

wing328:
FYI. I've pushed https://github.com/OpenAPITools/openapi-generator/pull/3446/commits/277822cb3b0f9b40f7a8040ebabe6a1ab7875188 to update the doc

cgensoul:
Ya I'd seen that stack overflow post as well after I had asked you how to handle it :) I guess I should at least make the code compilable then, even if the body is not really sent with the request. I'll see what I can do.

Thanks for fixing the doc !

wing328:
>  I guess I should at least make the code compilable then, even if the body is not really sent with the request

We can skip this particular case for the time being so that we can release a beta version for users to try it out and gather feedback. Let me know if you want to proceed this way and I'll share the details.

cgensoul:
I am fine with that. There are many other things to address anyway. For example the return code of the response is completly ignored (and 200 is assumed) for now which will probably be quite annoying for any real use case.

cgensoul:
Finally it was so easy to fix that I did it anyway ;)

wing328:
@cgensoul 👍 so is it ready for review and merge into master?

wing328:
Are there some commands to make sure the OCaml Petstore sample compiles?

An auto-generated README would be nice so that OCaml beginners will find it easy to use the OCaml library.

cgensoul:
The command to compile the lib once all the tools (opam first then dune through opam is the easiest path) and dependencies (through opam as well) are installed is :

`dune build`

at the root of the generated code directory.

I'll add a README pointing to installation instructions of the tools and dependencies next week. There is an example program using the library committed in the repository in the samples/client/petstore/ocaml-client/bin directory. It's called test.ml and has an accompanying dune build file. I'll clean it up and reference it in the README.

Regarding merging to master, it can be merged without any risk since it's all new code and no existing classes have been modified by this PR but I think proper handling of response codes should be added before it's really useful to anyone. So maybe it's best to wait a bit to save some reviewing work ?

cgensoul:
I added handling of response codes ;)

wing328:
Tried running `dune build` locally and I got:
```
➜  ocaml-client git:(cgensoul-ocaml-client) dune build


File "dune", line 6, characters 18-33:
6 |    (libraries str cohttp-lwt-unix lwt yojson ppx_deriving_yojson.runtime)
                      ^^^^^^^^^^^^^^^
Error: Library "cohttp-lwt-unix" not found.
Hint: try: dune external-lib-deps --missing @@default
File "dune", line 7, characters 20-39:
7 |    (preprocess (pps ppx_deriving_yojson ppx_deriving.std))
                        ^^^^^^^^^^^^^^^^^^^
Error: Library "ppx_deriving_yojson" not found.
Hint: try: dune external-lib-deps --missing @@default
```

Do I need to run other commands in order to install those libraries locally?

cgensoul:
Yes indeed. You need to install the dependencies.

opam install ppx_deriving_yojson cohttp ppx_deriving.std cohttp-lwt-unix

That should be enough. There is an opam search command in case you need to
find out the name of a package.

Le dim. 28 juil. 2019 à 05:25, William Cheng <notifications@github.com> a
écrit :

wing328:
Thanks for the tips. It works for me after I run `opam install ppx_deriving_yojson cohttp ppx_deriving cohttp-lwt-unix` (wihtout `.std`)

I would recommend including the opam install command in the readme as well.

cgensoul:
Ya that was my intention.

Le dim. 28 juil. 2019 à 15:40, William Cheng <notifications@github.com> a
écrit :

wing328:
@cgensoul thanks for the new generator. Do you have a Twitter account so that I can tag you in my tweet to promote the new generator?

cgensoul:
Yes my tweeter account is @cgensoul.

Le dim. 28 juil. 2019 à 15:51, William Cheng <notifications@github.com> a
écrit :

wing328:
I've filed https://github.com/OpenAPITools/openapi-generator/pull/3483 for various enhancements.

wing328:
Tweet: https://twitter.com/oas_generator/status/1155676626954276864

Reddit: https://www.reddit.com/r/ocaml/comments/cj5fx5/openapi_generator_now_supports_ocaml_client/

OCaml: https://news.ycombinator.com/item?id=20551269

wing328:
## Note

This generator is still in beta and may subject to breaking changes (without fallbacks) in the future.

wing328:
@cgensoul thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

PR #3447, opened 2019-07-24
Feature/mustache lambda tests

MichalFoksa:
### PR checklist

### Description of the PR
Follow up PR to #3368.

1. In registering common Mustache lambdas, do not fallback to `_lambda` key when `lambda` is already take in additionalProperties. It is possible only if `lambda` key was taken by `DefaultCodegen` itself.
2. Unit tests of common Mustache lambdas.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -240,13 +240,10 @@ public class DefaultCodegen implements CodegenConfig {
         }
 
         if (additionalProperties.containsKey("lambda")) {
-            LOGGER.warn("A property named 'lambda' already exists. Mustache lambdas renamed from 'lambda' to '_lambda'. " +
-                    "You'll likely need to use a custom template, " +
-                    "see https://github.com/OpenAPITools/openapi-generator/blob/master/docs/templating.md. ");
-            additionalProperties.put("_lambda", lambdas);
-        } else {
-            additionalProperties.put("lambda", lambdas);
+            LOGGER.error("A property called 'lambda' already exists in additionalProperties");
+            throw new RuntimeException("A property called 'lambda' already exists in additionalProperties");
         }
+        additionalProperties.put("lambda", lambdas);
     }
 
     // override with any special post-processing for all models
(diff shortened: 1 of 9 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -240,13 +240,10 @@ private void registerMustacheLambdas() {
         }
 
         if (additionalProperties.containsKey("lambda")) {
-            LOGGER.warn("A property named 'lambda' already exists. Mustache lambdas renamed from 'lambda' to '_lambda'. " +
-                    "You'll likely need to use a custom template, " +
```
jimschubert: Has the detailed message moved elsewhere? If not, we should keep the link to templating docs.

MichalFoksa: Hi @jimschubert 

I do not understand the question. 
Condition `if (additionalProperties.containsKey("lambda"))` should never evaluate to true and if then only because something has changed in DefautlCodegen and that should be caught by a first build.

I think the construct bellow made sense when it was in a generator derived from DefaultCodegen, but not now. 
```
 if (additionalProperties.containsKey("lambda")) {
            LOGGER.warn("A property named 'lambda' already exists. Mustache lambdas renamed from 'lambda' to '_lambda'. " +
                    "You'll likely need to use a custom template, " +
                    "see https://github.com/OpenAPITools/openapi-generator/blob/master/docs/templating.md. ");
            additionalProperties.put("_lambda", lambdas);
        } else {
            additionalProperties.put("lambda", lambdas);
        }
``` 

Makes sense?

jimschubert: I see, thanks for the clarification. I incorrectly remembered your earlier change being in DefaultGenerator, and missed that it's actually called in this class constructor.

MichalFoksa:
Re-based PR to current master.

jimschubert:
Sorry for the delay. Thanks for writing these tests and for updating the branch!

MichalFoksa:
@jimschubert No problem, You have been working on 4.1.0.

BTW: Congrats on the release. Now I can replace custom build generator with the 4.1.0 release in all of my repos. I really appreciate it, 

Michal

wing328:
@MichalFoksa thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

MichalFoksa:
:smile: you are welcome. Congrats to release.

Issue #3388, opened 2019-07-18
[BUG] [DefaultCodeGen] Map<String, X> is generated as API requestBody if Model class has additionalProperties=true


balazs-zsoldos:
##### Description

In case a model class has _additionalProperties=true_, in the API a Map is inserted as requestBody parameter instead of the model class.

I

##### openapi-generator version

4.0.3

##### OpenAPI declaration file content or url

I used the following pom.xml to generate the sources: https://gist.github.com/balazs-zsoldos/7c707b098d97bac59a3b411d5d6b0664

For an example, see _GroupResourceAPI.addUserToGroup(...)_. The requestBody parameter here is a Map, but it should be _UpdateUserToGroupBean_.

##### Suggest a fix

I analyzed the problem and I think the behavior could be changed in DefaultCodeGen.fromRequestBody(...) function, more specifically changing [this and the next line](https://github.com/OpenAPITools/openapi-generator/blob/a85bd0f8f9785b6d7f215da56085bafb8c647016/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L4744) to the following code solves the issue:

    if (name != null) {
        CodegenModel codegenModel = fromModel(name, schema);
        codegenParameter.baseType = codegenModel.classname;
        codegenParameter.dataType = getTypeDeclaration(codegenModel.classname);
        imports.add(codegenParameter.baseType);
    } else {
        codegenParameter.dataType = getTypeDeclaration(schema);
        codegenParameter.baseType = getSchemaType(inner);
    }

However, I am not sure how this would affect all generators that already exist. If there is a specific reason why Map is inserted in this case instead of the model type. I would like to send a patch that does not cause any trouble.

Do you think this is a good fix? Or shall I create a new configuration property for the generator like _useModelTypeForRequestBodyParamWhereAdditionalPropertiesIsTrue_ and make it default false? In that case, the current behavior is not changed only if explicitly sets this configuration to _true_.

PR #3449, opened 2019-07-24
#3388 requestbody parameter type fix in case of additionalProperties

balazs-zsoldos:
### Description of the PR

Solution for issue #3388. I have checked and return values or properties of models are generated well, meaning the real model class is used and not a Map in case additionalProperties is true on a model class.

Therefore, this fix generates the model class (if available) as type of the requestBody parameter in all languages.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -4741,10 +4741,22 @@ public class DefaultCodegen implements CodegenConfig {
             codegenParameter.paramName = toParamName(codegenParameter.baseName);
             codegenParameter.items = codegenProperty.items;
             codegenParameter.mostInnerItems = codegenProperty.mostInnerItems;
-            codegenParameter.dataType = getTypeDeclaration(schema);
-            codegenParameter.baseType = getSchemaType(inner);
             codegenParameter.isContainer = Boolean.TRUE;
             codegenParameter.isMapContainer = Boolean.TRUE;
+
+            CodegenModel codegenModel = null;
+            if (name != null) {
+                codegenModel = fromModel(name, schema);
+            }
+
+            if (codegenModel != null) {
+                codegenParameter.baseType = codegenModel.classname;
+                codegenParameter.dataType = getTypeDeclaration(codegenModel.classname);
+                imports.add(codegenParameter.baseType);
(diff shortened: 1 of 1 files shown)


richardwhiuk:
I think https://github.com/OpenAPITools/openapi-generator/issues/426 is related to this - where there's a desire for the generators to cleanly separate implementation and interface logic.

balazs-zsoldos:
@trajano  Sorry for the late reply, I was on holiday. I hope that the PR I sent will be accepted soon and it will be part of the next release.

Until that, my way is adding a openapitools.patch file to the root of my projects and mentioning in the README that one has to compile OpenApiTools first with the patch. This is not a good solution, but I hope this is temporary.

richardwhiuk:
Can you add a test to show how this changes the generated code?

balazs-zsoldos:
> Can you add a test to show how this changes the generated code?

@richardwhiuk I would like to do so. Probably in a couple of days, but I must analyze first, how the other tests are written.

balazs-zsoldos:
@richardwhiuk  Added a unit test. Hope you will like it.

I am not sure if I should add a requestBody parameter type with additionalProperties to the schema that is used for sample code generation as it seems that this feature was hardly tested in the different languages till now. So samples might get syntax errors or unexpected result and I do not speak as many languages as many generators there are to check all :/.

richardwhiuk:
I was expecting an addition to the sample APIs to show how this affects code generation...

balazs-zsoldos:
@richardwhiuk I can do that as well. However, no additionalProperties is used in the sample at the moment. There are a huge amount of generators and if I add a requestBody with additionalParameters, some of the generators might not generate a good code (even worse, syntactically incorrect code). The question is: Shall I add such a requestbody to the sample and the community will check all languages and all generators if the samples are correct? Or is there a way how I can test that at least the generated sample codes compile in every language?

richardwhiuk:
CI should check that the updated samples compile.

balazs-zsoldos:
@richardwhiuk I added a sample function call to petstore-with-fake-endpoints-models-for-testing.yaml. However, if I execute the re-generation under _bin\windows_, not only the relevant file is changed in the generated sample but many of them. Is this on purpose? Shall I commit re-generated samples although almost every file will be changed?

balazs-zsoldos:
@richardwhiuk Could you, please, give some advice? Shall I commit the 2404 changed files to this PR after sample-regeneration?

PR #3450, opened 2019-07-24
Support Multiple API Keys

nmuesch:
### PR checklist

### Description of the PR

Follow up to #3210


nmuesch:
Hey @wing328 👋 I opened this PR as a followup to your review on #3210 I haven't done the updates to the sample files yet but was wondering if you could take a look to see if this is more along the lines of what you were thinking (and to confirm I've opened the PR at the right place) Thanks!

bkabrda:
So just to make sure that I understand this PR correctly, it allows adding multiple keys by removing the assumption that there is just one `apikey` in the context and gets them by names that were specified in the security schemes. Is that correct?

wing328:
@bkabrda yes, I think the goal is to support multiple API keys as the current approach only supports one API key. 

I was expecting the PR to use maps somehow similar to other generators, e.g.:

- C#: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/csharp/OpenAPIClient/src/Org.OpenAPITools/Client/Configuration.cs#L120

- PHP: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/php/OpenAPIClient-php/lib/Configuration.php#L49

 but there could be other approaches to support multiple API key.

bkabrda:
Right, so I think the potential problem with this approach is that the names of the API keys, which are given in the OpenAPI spec, could potentially conflict with other names in the context. These are currently `token`, `basic`, `accesstoken`, which someone *could* use as names for API keys.

So why don't we rather put a map with API keys in the context, which would prevent these collisions? This map would have a reserved keyname there, much like the previously used `ContextAPIKey = contextKey("apikey")`, so something like `ContextAPIKeys = contextKey("apikeys")`, which would be of type `map[string]string`. @nmuesch does that make sense to you?

nmuesch:
Hey, @wing328 and @bkabrda  apologies for the delay on this one. These latest commits use a different approach, using the map idea that you had both mentioned. Is this more in line with what you were thinking? If so I can run the script to generate the examples. Thanks for the reviews!

modules/openapi-generator/src/main/resources/go-experimental/README.mustache
```diff
@@ -61,7 +61,7 @@ Class | Method | HTTP request | Description
 Example
 
 ```golang
-auth := context.WithValue(context.Background(), sw.ContextAPIKey, sw.APIKey{
+auth := context.WithValue(context.Background(), sw.ContextAPIKeys, sw.APIKey{
```
bkabrda: This doesn't look right. It sets the context value to be the APIKey, but in fact it should be the map, right?

nmuesch: Good catch! This was indeed outdated. I couldn't find a way to only show the example once and only if there was an API Key auth present in the list of authMethods, so I followed the approach of the other languages READMEs, which was to remove the example and just list each type of key. 

Ex:
Python - https://github.com/OpenAPITools/openapi-generator/tree/5956569e7a0d423521168f894bd1ac7b56ad35f1/samples/openapi3/client/petstore/python#documentation-for-authorization
Kotlin - https://github.com/OpenAPITools/openapi-generator/tree/5956569e7a0d423521168f894bd1ac7b56ad35f1/samples/openapi3/client/petstore/kotlin#documentation-for-authorization
(There are other examples for this as well)

bkabrda: So how about adding an example to each entry of the list? I do understand the issue with not being able to implement document this really good, but there should be some example of how to put the map in the context.

nmuesch: Talked about this offline a bit. I'm hesitant to leave example code because the example will not work if there are multiple API keys required. (Because each example will setup a context with a different one-element map) I assume thats why the other client README templates don't have an example.

I did though, add a note that says you have to add this key to a map[string]APIKey inside each APIKey section. What do you think?

bkabrda: Yeah, looks much better now 👍

wing328:

If no further feedback from anyone, I'll merge it this weekend.

nmuesch:
Hey, just wanted to follow up and see if there was anything else I could do here to get this merged? Thanks 🙇

modules/openapi-generator/src/main/resources/go-experimental/README.mustache
```diff
@@ -54,19 +54,14 @@ Class | Method | HTTP request | Description
...
-Example
-
-```golang
-auth := context.WithValue(context.Background(), sw.ContextAPIKey, sw.APIKey{
-    Key: "APIKEY",
-    Prefix: "Bearer", // Omit if not necessary.
-})
-r, err := client.Service.Operation(auth, args)
-```
+Note, each API key must be added to a map of `map[string]APIKey` where the key is: {{keyParamName}} and passed in as the auth context for each request.
```
wing328: @nmuesch minor feedback. Is there a way we can keep the sample code?

It would be great to show the code to use the API key (single or multiple)

modules/openapi-generator/src/main/resources/go-experimental/api.mustache
```diff
@@ -249,19 +249,21 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
...
+		if auth, ok := ctx.Value(ContextAPIKeys).(map[string]APIKey); ok {
+			if auth, ok := auth["{{keyParamName}}"]; ok {
+				var key string
+				if auth.Prefix != "" {
+					key = auth.Prefix + " " + auth.Key
+				} else {
+					key = auth.Key
+				}
+				{{#isKeyInHeader}}
+				localVarHeaderParams["{{keyParamName}}"] = key
```
wing328: @nmuesch can we check to see if the key is not empty before adding to localVarHeaderParams or localVarQueryParams?

modules/openapi-generator/src/main/resources/go-experimental/README.mustache
nmuesch: Hey, thanks for the feedback. There was a brief discussion of this here https://github.com/OpenAPITools/openapi-generator/pull/3450#pullrequestreview-277692301 interested in your thoughts. The main reason I changed it was because it made the `example` code incorrect for any apps that have multiple keys. I couldn't figure out a way to show the full multi-key example using the template variables.

modules/openapi-generator/src/main/resources/go-experimental/api.mustache
wing328: The reason is that sometimes the key can either be put in the header or the URL query string so the goal is to skip the API key in header or URL query string if the value is not set. We can address it in another PR later when someone reports the issue.

modules/openapi-generator/src/main/resources/go-experimental/README.mustache
wing328: We can do it in the actual code sample instead and we can do it in another PR instead of this one as I don't want this PR to be opened for too long

wing328:
@nmuesch thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #3466, opened 2019-07-25
fix(typescript-angular): improve small template issues

eseliger:
### PR checklist

### Description of the PR

Fixes several minor issues in the templates for typescript-angular

- `canConsumeForm` is only included in the service if the method is actually needed. Reduces code size and helps with compiling the project where no unused locals cause compile errors
- no summary or no notes for an operation don't cause the jsdoc to start with an empty line
- only render consumes property when it is actually used. Also helpful when compiling the project with no unused locals
- fix formatting of doc block for encoder.ts

Anyone knows of a better solution to fix the consumes rather than having this "or" which needs the code to be duplicated? I didn't want to add another property to the CodegenOperation since it's only used for 1 generator. Maybe this is also acceptable since it'll be refactored soon anyways. Please advise :) 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -333,7 +333,6 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         boolean hasSomeFormParams = false;
         for (CodegenOperation op : ops) {
             if (op.getHasFormParams()) {
-                System.out.println(op);
                 hasSomeFormParams = true;
             }
             if ((boolean) additionalProperties.get("useHttpClient")) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -330,7 +330,12 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         objs.put("apiFilename", getApiFilenameFromClassname(objs.get("classname").toString()));
 
         List<CodegenOperation> ops = (List<CodegenOperation>) objs.get("operation");
+        boolean hasSomeFormParams = false;
         for (CodegenOperation op : ops) {
+            if (op.getHasFormParams()) {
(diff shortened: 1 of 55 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -330,7 +330,12 @@ public void postProcessParameter(CodegenParameter parameter) {
         objs.put("apiFilename", getApiFilenameFromClassname(objs.get("classname").toString()));
 
         List<CodegenOperation> ops = (List<CodegenOperation>) objs.get("operation");
+        boolean hasSomeFormParams = false;
         for (CodegenOperation op : ops) {
+            if (op.getHasFormParams()) {
+                System.out.println(op);
```
macjohnny:
```suggestion
```

eseliger:
> I would rather avoid more template complexity that "only" removes some unused variables.

I understand that adding more variables to the templates is not really desirable, but for this case I see the problem at how the client can be used. If you don't specify "npmName" as an argument, only the TS code is generated, without tsconfig files etc. In order to use it you have to add it as part of your own compilation (which can be desirable to not have to worry about "npm link"-ing packages for local development), but since TS will FAIL compilation with unused variables, as long as the rule is enabled, I think it really makes sense to fix that, but please prove me wrong and show me another good way of solving this :)

macjohnny:
I absolutely understand, the code needs to adhere to the standards of the project if it is part of the compilation. One way to solve it would be to post-process the generated files with a command being executed after the generation, see e.g. https://wesleygrimes.com/angular/2019/02/14/how-to-use-tslint-to-autoremove-all-unused-imports-in-a-typescript-project.html

eseliger:
I managed to adjust most stuff to local codestyle using `prettier --write && tslint --fix` for post-processing, but since unused private members and variables checks have been implemented in tsc itself, there is no fixer in tslint anymore to help with that and it's grown up from a linter issue to a compiler issue 🤔 for linting I would say if people complain, let's leave it since there is no need to do linting on the generator files anyways, just go ahead and exclude them from the linter scope. But for compilation it's difficult to exclude files :D

macjohnny:
ok, I see, this should be fixed in the generated code then.

eseliger:
Thank you for your understanding :)

wing328:
@eseliger thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

Issue #3149, opened 2019-06-13
[BUG] Import Mapping seems to not work with typescript-angular


filipvh:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))
##### Description

I used importMapping in combination with the typescript-angular generator.
Here is a redacted config file:
```json
{
  "npmName": "the-package-name",
  "npmVersion": "1.8.0",
  "ngVersion": "1.7.0",
  "modelFileSuffix": ".type",
  "npmRepository" : "http://domain/is/redacted",
  "snapshot" : false,
  "importMappings": {
        "EnumValue":"some-other-package"
  }
}
```

##### openapi-generator version

4.0.1

##### What happens?

 - The 'EnumValue' is not generated (is missing from model folder) => OK
 - In an other type/interface EnumValue is still refferenced locally => NOK

```js
import { EnumValue } from './enumValue.type';
```

##### What was expected?

 - The 'EnumValue' is not generated (is missing from model folder)
 - In an other type/interface EnumValue refferences the external package

```js
import { EnumValue } from 'some-other-package';
```

##### Command line used for generation

```
generate -i model.yaml -g typescript-angular -c config.json -o out
```

macjohnny:
@filipvh would you like to try and fix this issue?

filipvh:
@macjohnny  I can try. Wanted to make sure that it is an issue and that there isn't a duplicate i couldn't identify. If one would expect it to do what I expect... :-)

scottydawg:
@filipvh - I'm up against this issue as well for the `TypeScriptNodeClientCodegen`. I'm going to look into a fix unless you've already got something going here that might address both.

[...]

PR #3469, opened 2019-07-25
Support importMapping definitions for TypeScriptNodeClientCodegen

scottydawg:
### PR checklist

### Description of the PR

This addresses an issue similar to Issue #3149 but for the TypeScript Node generator. This is a more general issue that could be implemented for all TypeScript generators, but I'm keeping the scope small initially.

The `importMapping` behavior described [here](https://openapi-generator.tech/docs/customization#bringing-your-own-models) and [here](https://openapi-generator.tech/docs/usage#target-external-models) in the usage docs, and supported [by the Maven plugin configuration](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator-maven-plugin#general-configuration-parameters) does not currently work for TypeScript generation.

This change allows a user to define these mappings and produce valid TypeScript import statements when using the TypeScriptNodeClientCodegen.

> Note: Running the shell script `./bin/openapi3/typescript-node-petstore-all.sh` in the **master** branch generates sample files that are different from the files currently in the repo. Running that script against this branch (typescript-import-mappings) produces the same differences. I have not included them here to avoid conflating unrelated changes. 

cc: 
* @TiFu 
* @taxpon 
* @sebastianhaas 
* @kenisteward 
* @Vrolijkx
* @macjohnny
* @nicokoenig
* @topce
* @akehir

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -18,7 +18,6 @@
 package org.openapitools.codegen.languages;
 
 import io.swagger.v3.oas.models.media.Schema;
-import io.swagger.v3.parser.util.SchemaTypeUtil;
 import org.openapitools.codegen.*;
 import org.openapitools.codegen.utils.ModelUtils;
 import org.slf4j.Logger;
@@ -33,6 +32,7 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
     private static final Logger LOGGER = LoggerFactory.getLogger(TypeScriptNodeClientCodegen.class);
 
     public static final String NPM_REPOSITORY = "npmRepository";
+    private static final String DEFAULT_IMPORT_PREFIX = "./";
 
     protected String npmRepository = null;
     protected String apiSuffix = "Api";
@@ -106,7 +106,11 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
 
     @Override
     public String toModelFilename(String name) {
-        return camelize(toModelName(name), true);
+        if (importMapping.containsKey(name)) {
+            return importMapping.get(name);
+        }
+
+        return DEFAULT_IMPORT_PREFIX + camelize(toModelName(name), true);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/resources/typescript-node/model.mustache
```diff
@@ -2,7 +2,7 @@
 {{#models}}
 {{#model}}
 {{#tsImports}}
-import { {{classname}} } from './{{filename}}';
+import { {{classname}} } from '{{filename}}';
```
scottydawg: I removed the `./` from the template and added it to the `toFileName()` method in the generator. Having the prefix here makes it impossible to have a correct import for an external dependency, like:

```typescript
import { MyThing } from '@mynamespace/mypackage/mything';
```

macjohnny:
@scottydawg thanks for the PR. It would be nice if you could port that to the other typescript-generators, too

macjohnny:
there are still some issues, please have a look at https://circleci.com/gh/OpenAPITools/openapi-generator/7864#tests/containers/2

scottydawg:
It's strange; CI is definitely producing different samples, but running `./bin/typescript-node-petstore-all.sh` locally does _not_. I'll try to figure out what's going on.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -96,22 +96,37 @@ public String toApiFilename(String name) {
         if (name.length() == 0) {
             return "default" + apiSuffix;
         }
+        if (importMapping.containsKey(name)) {
+            return importMapping.get(name);
+        }
```
scottydawg: There are 4 methods here that convert a type name into some sort of file path. My first pass at this only updated one; this latest commit updates all of them. 

That this condition is checked in 4 different places is evidence that this could be addressed in a more general way. The `DefaultGenerator` already [checks the importMapping](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java#L1152), but it's geared for imports that don't have file path dependencies (e.g. Java). In the TypeScript implementations, we need to construct the import differently based on whether it is mapped (in which case we use the mapped reference verbatim) or unmapped (in which case we attempt to construct a relative file path).

scottydawg:
@macjohnny - This one should pass CI. I wasn't seeing the failures locally because I hadn't `mvn install`'d locally with my changes. My locally generated samples are now unchanged from what's in the repo.

wing328:
@scottydawg thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

Issue #3435, opened 2019-07-24
[BUG][Java][client] Erroneous support for nullable attributes


bkabrda:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
  - The included `modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml` can be used to reproduce this.
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
  - All versions seem to be affected (I used 4.0.2)
- [x] Have you search for related issues/PRs?
  - The same behaviour applied to Go, reported at [1] and fixed in go-experimental in [2].
- [x] What's the actual output vs expected output?
  - The problem is that for attributes marked as `nullable`, `null` is never sent.
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The problem is very similar to what was described in [1] - if an attribute is marked as `nullable` in the OpenAPI spec, it's not actually possible to send a `null` value in a POST/PUT request for it (using okhttp/gson client, but AFAICS other clients would have these issues as well).

##### openapi-generator version

4.0.2, but the issue seems to always have been there.

##### OpenAPI declaration file content or url

`modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml`

##### Command line used for generation

`openapi-generator -g java -i modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml -o /tmp/petstore/`

##### Steps to reproduce

You just need to generate a Java client for OpenAPI spec that has any nullable attributes and see that it doesn't really send nulls for them in the request JSON objects.

##### Related issues/PRs

[1] https://github.com/OpenAPITools/openapi-generator/issues/522
[2] https://github.com/OpenAPITools/openapi-generator/pull/3371

##### Suggest a fix

My solution for Go from [2], which I believe is correct, considers 4 different states for every attribute:

1. not required, not nullable
2. required, not nullable
3. not required, nullable
4. required, nullable

All these 4 states need to be covered properly in order to be able to send the requests correctly according to the spec. Here's my thinking on how the 4 above scenarios should behave:

[...]

Issue #689, opened 2018-07-30
Improve Java-Client Tests and CI


jmini:
Input @MatanRubin on #513:

> this was a serious bug in a very basic scenario for a very common library (i.e. the generated client fails a simple GET for Java + RestTemplate). Don't we have tests that cover this kind of scenarios for all languages and libraries? Not trying to blame anyone here, just asking out of curiosity and to see maybe I can contribute some tests that will help prevent this kind of issues.

Well the more I am confronted with changes to java-client templates, the more I think we need to do a real setup.

I theory some of the Java-Client have some tests for clients implemented (replacement of the generated stup) that perform test against some server serving the "Petstore sample". During the creation of OpenAPI-Generator, some of them were removed => #247. This reminds me that I should finish restoring them.

But this is not enough... Have a look at the PR I have merge this morning: #646. It is about special chars being wrongly encoded in query-params.
A test case would be ideal for that.

We need a setup where we can control the server for each test.
Using the generated code (XxxxApi) we send a request and we create expectations on what should arrive to the server.
Or in the server we define what should be returned and we see if the client did interpret the answer correctly.

I am wondering which library can be used to do something like this. Easy embedding of a small HTTP-Server. I have started to have a look at https://github.com/arteam/embedded-http-server, I can't imagine that there is not something.
(it is not about mocking the server, but having a simple one where we can perform assertions on request that arrived and were we can control the answer).

Any suggestions?

Then I think that we should generate client for all libs we support. I started something in this direction: https://github.com/jmini/openapi-experiments/tree/master/openapi-generator-utils/icespellmint. The [spec](https://github.com/jmini/openapi-experiments/blob/master/openapi-generator-utils/icespellmint/icespellmint.yaml) used to create the client will be updated to test a lot of cases we have (all HTTP verbs like "GET", "POST", "PUT"..., with or without parameters, with different supported content-type (json, text, xml)).

This way we can control if something is an improvement or creates some regressions.

There is an other interesting issue: #476, I have started to work on this, but again I want to have a great coverage to be sure that I am not breaking anything.

Feedback is welcomed.

wing328:
Previously (a while ago) I suggested using our own server generator to generate a server stub that allows us to add new test cases easily. 

If time permits, I may try to do that with the Rails generator or one of the PHP server generators.

jmini:
@wing328: Thank you for this feedback, but I would like to have something 100% java. Each unit test for the java-client should looks like this:

* Small server started
[...]

PR #3474, opened 2019-07-26
[Java][client] make it possible to send explicit nulls for nullable fields

bkabrda:
# PR checklist

### Description of the PR

This is my attempt to fix https://github.com/OpenAPITools/openapi-generator/issues/3435. Basically it makes sure that explicit `null` values can be sent for nullable fields:

* For a required nullable field, the `null` is always sent when that's the value.
* For an optional (not required) nullable field, I introduced a simple new `Optional` class with `OptionalFilter` that make sure that the `null` is sent if and only if it was set explicitly. AFAICS, the API of any of the generated classes didn't change in a backwards-incompatible way.

Discussion points:
* I added the `Optional` since `java.util.Optional` was added in JDK 8 and the templates still have support for previous Java versions. If desired, I can replace this with a different (third party) implementation pretty easily.
* I didn't run the `./bin/...` generators yet to keep the PR clean and reviewable. I'm planning to run it later.
* I only added this for jackson, as I didn't figure out an easy way to do this for gson. I could try working on a different PR that would do this for gson, but that would take me a lot more time.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -417,6 +417,8 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
         importMapping.put("JsonTypeInfo", "com.fasterxml.jackson.annotation.JsonTypeInfo");
         importMapping.put("JsonCreator", "com.fasterxml.jackson.annotation.JsonCreator");
         importMapping.put("JsonValue", "com.fasterxml.jackson.annotation.JsonValue");
+        importMapping.put("JsonIgnore", "com.fasterxml.jackson.annotation.JsonIgnore");
+        importMapping.put("JsonInclude", "com.fasterxml.jackson.annotation.JsonInclude");
         importMapping.put("SerializedName", "com.google.gson.annotations.SerializedName");
         importMapping.put("TypeAdapter", "com.google.gson.TypeAdapter");
         importMapping.put("JsonAdapter", "com.google.gson.annotations.JsonAdapter");
@@ -800,7 +802,13 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
                 }
             }
             return null;
+        } else if (ModelUtils.isObjectSchema(p)) {
+            if (p.getDefault() != null) {
+                return super.toDefaultValue(p);
+            }
+            return null;
         }
+
         return super.toDefaultValue(p);
     }
 
--- a/modules/openapi-generator/src/main/resources/Java/libraries/google-api-client/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/google-api-client/ApiClient.mustache
@@ -4,6 +4,7 @@ import {{apiPackage}}.*;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import org.openapitools.jackson.nullable.JsonNullableModule;
 {{#joda}}
 import com.fasterxml.jackson.datatype.joda.JodaModule;
 {{/joda}}
@@ -53,6 +54,8 @@ public class ApiClient {
         module.addDeserializer(ZonedDateTime.class, CustomInstantDeserializer.ZONED_DATE_TIME);
         objectMapper.registerModule(module);
         {{/threetenbp}}
+        JsonNullableModule jnm = new JsonNullableModule();
+        objectMapper.registerModule(jnm);
         return objectMapper;
     }
 
--- a/modules/openapi-generator/src/main/resources/Java/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/pojo.mustache
@@ -19,21 +19,6 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
     {{/mostInnerItems}}
     {{/isContainer}}
     {{/isEnum}}
-  {{#jackson}}
-  public static final String JSON_PROPERTY_{{nameInSnakeCase}} = "{{baseName}}";
-  @JsonProperty(JSON_PROPERTY_{{nameInSnakeCase}})
-  {{#withXml}}
-    {{^isContainer}}
-  @JacksonXmlProperty({{#isXmlAttribute}}isAttribute = true, {{/isXmlAttribute}}{{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}localName = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
-    {{/isContainer}}
-    {{#isContainer}}
-      {{#isXmlWrapped}}
-  // items.xmlName={{items.xmlName}}
-  @JacksonXmlElementWrapper(useWrapping = {{isXmlWrapped}}, {{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}localName = "{{#items.xmlName}}{{items.xmlName}}{{/items.xmlName}}{{^items.xmlName}}{{items.baseName}}{{/items.xmlName}}")
-      {{/isXmlWrapped}}
-    {{/isContainer}}
-  {{/withXml}}
-  {{/jackson}}
   {{#withXml}}
   {{#isXmlAttribute}}
   @XmlAttribute(name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
@@ -59,12 +44,25 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
   @SerializedName(SERIALIZED_NAME_{{nameInSnakeCase}})
   {{/gson}}
+  {{#jackson}}
+  public static final String JSON_PROPERTY_{{nameInSnakeCase}} = "{{baseName}}";
+  {{/jackson}}
+  {{#vendorExtensions.isJacksonOptionalNullable}}
+  {{#isContainer}}
+  private JsonNullable<{{{datatypeWithEnum}}}> {{name}} = JsonNullable.<{{{datatypeWithEnum}}}>undefined();
+  {{/isContainer}}
+  {{^isContainer}}
+  private JsonNullable<{{{datatypeWithEnum}}}> {{name}} = JsonNullable.<{{{datatypeWithEnum}}}>{{#defaultValue}}of({{{.}}}){{/defaultValue}}{{^defaultValue}}undefined(){{/defaultValue}};
+  {{/isContainer}}
+  {{/vendorExtensions.isJacksonOptionalNullable}}
+  {{^vendorExtensions.isJacksonOptionalNullable}}
   {{#isContainer}}
   private {{{datatypeWithEnum}}} {{name}}{{#required}}{{#defaultValue}} = {{{.}}}{{/defaultValue}}{{/required}}{{^required}} = null{{/required}};
   {{/isContainer}}
   {{^isContainer}}
   private {{{datatypeWithEnum}}} {{name}}{{#defaultValue}} = {{{.}}}{{/defaultValue}};
   {{/isContainer}}
+  {{/vendorExtensions.isJacksonOptionalNullable}}
 
   {{/vars}}
   {{#parcelableModel}}
@@ -89,14 +87,28 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   {{/gson}}
   {{/parcelableModel}}
   {{#vars}}
+
   {{^isReadOnly}}
   public {{classname}} {{name}}({{{datatypeWithEnum}}} {{name}}) {
-    this.{{name}} = {{name}};
+    {{#vendorExtensions.isJacksonOptionalNullable}}this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{name}});{{/vendorExtensions.isJacksonOptionalNullable}}
+    {{^vendorExtensions.isJacksonOptionalNullable}}this.{{name}} = {{name}};{{/vendorExtensions.isJacksonOptionalNullable}}
     return this;
   }
   {{#isListContainer}}
 
   public {{classname}} add{{nameInCamelCase}}Item({{{items.datatypeWithEnum}}} {{name}}Item) {
+    {{#vendorExtensions.isJacksonOptionalNullable}}
+    if (this.{{name}} == null || !this.{{name}}.isPresent()) {
+      this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{{defaultValue}}});
+    }
+    try {
+      this.{{name}}.get().add({{name}}Item);
+    } catch (java.util.NoSuchElementException e) {
+      // this can never happen, as we make sure above that the value is present
+    }
+    return this;
+    {{/vendorExtensions.isJacksonOptionalNullable}}
+    {{^vendorExtensions.isJacksonOptionalNullable}}
     {{^required}}
     if (this.{{name}} == null) {
       this.{{name}} = {{{defaultValue}}};
@@ -104,11 +116,24 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
     {{/required}}
     this.{{name}}.add({{name}}Item);
     return this;
+    {{/vendorExtensions.isJacksonOptionalNullable}}
   }
   {{/isListContainer}}
   {{#isMapContainer}}
 
   public {{classname}} put{{nameInCamelCase}}Item(String key, {{{items.datatypeWithEnum}}} {{name}}Item) {
+    {{#vendorExtensions.isJacksonOptionalNullable}}
+    if (this.{{name}} == null || !this.{{name}}.isPresent()) {
+      this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{{defaultValue}}});
+    }
+    try {
+      this.{{name}}.get().put(key, {{name}}Item);
+    } catch (java.util.NoSuchElementException e) {
+      // this can never happen, as we make sure above that the value is present
+    }
+    return this;
+    {{/vendorExtensions.isJacksonOptionalNullable}}
+    {{^vendorExtensions.isJacksonOptionalNullable}}
     {{^required}}
     if (this.{{name}} == null) {
       this.{{name}} = {{{defaultValue}}};
@@ -116,6 +141,7 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
     {{/required}}
     this.{{name}}.put(key, {{name}}Item);
     return this;
+    {{/vendorExtensions.isJacksonOptionalNullable}}
   }
   {{/isMapContainer}}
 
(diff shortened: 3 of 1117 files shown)


cbornet:
Please see my comment in #3435 . Wouldn't JsonNullable be a good fit for this ?

bkabrda:
@cbornet I added a commit that removes the custom Optional implementation that I did and instead uses JsonNullable. Thanks for the suggestion, it works like a charm.

Some more things I still need to do:
* I made sure that elements that map to `null` in maps with nullable items are preserved, but I didn't really check what the behavior is for arrays, so I need to see how that works.
* If/when this change seems fine, I'll have to update `JSON.mustache` and `pom.xml` of all the other library templates that use jackson to make sure they include JsonNullable relevant code.

modules/openapi-generator/src/main/resources/Java/pojo.mustache
```diff
@@ -58,7 +44,15 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
   @SerializedName(SERIALIZED_NAME_{{nameInSnakeCase}})
   {{/gson}}
+  {{#jackson}}
+{{> jackson_annotations}}
+  {{/jackson}}
+  {{#vendorExtensions.isJacksonOptionalNullable}}
+  private JsonNullable<{{{datatypeWithEnum}}}> {{name}} = JsonNullable.{{#defaultValue}}of({{{.}}}){{/defaultValue}}{{^defaultValue}}undefined(){{/defaultValue}};
```
cbornet: Is there a need for a vendor extension ? I think it can be wrapped with JsonNullable as soon as the field is "nullable".

bkabrda: This was my line of thinking with the vendor extension:

I needed to have a bunch of "if and else" conditions in the pojo template to determine if the context have `jackson`, `!required` and `nullable` set at the same time. I could theoretically drop the `!required` as condition (even though I like having the simplicity of going with `JsonInclude.Include.ALWAYS` for required fields), but I can't drop the `jackson` as condition.
If I didn't have this as one variable in the mustache rendering context, it'd be hard doing the "else" part. Additionally, since `jackson` is Java specific, I decided to use a "vendor" variable for this, as the other uses of vendor extensions in other codegens seemed to be similar to what I'm doing here.

I'm of course ok with improving this if you have a better idea.

bkabrda:
@cbornet hmm, so I noticed that this isn't really working. The "explicit nulls" (IOW JsonNullable that is defined and set to null) are not serialized as null. Any idea on what I'm doing wrong? I can't seem to make it work.

bkabrda:
Right, so it seems that this works when I use the jackson annotations on the `{{getter}}_JsonNullable` instead of the attribute itself. I'm guessing that this may be caused by the internal implementation of jackson and how it works with private attributes and their getters/setters. I'm going to push an additional commit that fixes this.

bkabrda:
Fixed and seems to work fine. This is now pretty much ready from my side, except I'll still have to edit pom.xml/JSON.java templates in other generators using jackson and also actually regenerate all affected samples.

bkabrda:
As requested, I'm sharing a simple example of what a generated pojo would look like. There are some extraneous whitespaces that I need to get rid of in my changes, but everything seems to work. (Note that none of the samples really illustrate this good and I can't share my real work yet, so this is a made up example).

The biggest changes are on optional-nullable fields and maps that contain null values, so I'm sharing an example with these two:

```java
public class MySchema {
  // mymap has additionalProperties that are nullable
  private Map<String, Long> mymap = new HashMap<String, Long>();

  // end is a nullable, not required Long
  private JsonNullable<Long> end = JsonNullable.undefined();

  public MonitorOptions mymap(Map<String, Long> mymap) {
    
    this.mymap = mymap;
    return this;
  }

  public MonitorOptions putMymapItem(String key, Long mymapItem) {
    if (this.mymap == null) {
      this.mymap = new HashMap<String, Long>();
    }
    this.mymap.put(key, mymapItem);
    return this;
  }


   /**
   * Get mymap
   * @return mymap
  **/
  @javax.annotation.Nullable
  @ApiModelProperty(value = "")

  @JsonProperty("mymap")
  @JsonInclude(content = JsonInclude.Include.ALWAYS, value = JsonInclude.Include.USE_DEFAULTS)


  public Map<String, Long> getMymap() {
    return mymap;
  }

  public void setMymap(Map<String, Long> mymap) {
    this.mymap = mymap;
  }

  public Downtime end(Long end) {
    this.end = JsonNullable.of(end);
    
    return this;
  }

  @JsonProperty("end")
  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)

  public JsonNullable<Long> getEnd_JsonNullable() {
    return end;
  }

  public void setEnd_JsonUndefined() {
    this.end = JsonNullable.undefined();
  }

   /**
   * Get end
   * @return end
  **/
  @javax.annotation.Nullable
  @ApiModelProperty(example = "1412793983", value = "")

  public Long getEnd() {
        return end.orElse(null);
  }

  public void setEnd(Long end) {
    this.end = JsonNullable.of(end);
  }
}
```

bkabrda:
I pushed some additional commits:
* one to remove some of the extraneous workspace
* one to add the JsonNullable integration to the rest of library templates that use jackson
* one that actually regenerates all the samples

bkabrda:
(Sorry for so many force-pushes, I'm just trying to squash all the bugs/typos... I've already figured out that I can run the tests for the regenerated libraries individually on my laptop, so I'll try to fix everything before the next force push which will hopefully also be the last one.)

jmini:
> Sorry for so many force-pushes

No worry. 

If you prefer, you can also push multiple mini-commits to your branch. At the end all of them will be squashed into one single commit on the master branch.

bkabrda:
So finally all tests are passing. Can I get another round of review? Thanks!

CC @cbornet

jimschubert:
I think JsonNullable might be confusing if someone uses `withXml` for an API with mixed JSON and XML endpoints. Would a nullable property in such a scenario serialize correctly for the xml endpoint?

I kinda wish the `JsonNullable` object didn't include the transport protocol in the name. It seems it would be nice to call it a `Nullable`.

bkabrda:
> I think JsonNullable might be confusing if someone uses withXml for an API with mixed JSON and XML endpoints. Would a nullable property in such a scenario serialize correctly for the xml endpoint?

So I agree with you on the naming part, but there's not much I can do about that unless I wanted to radically change the jackson-databind-nullable library.

I have to admit I didn't try to use this with XML endpoints. I was assuming that jackson would automagically do the correct thing, but I do agree that it's something to test for sure.

> I kinda wish the JsonNullable object didn't include the transport protocol in the name. It seems it would be nice to call it a Nullable.

Yeah, I guess that would be better from this perspective, but I honestly consider that to be out of scope of this PR, as it's already complex as it is.

jmini:
Is this feature always included, or is it something the user can opt-in/opt-out?

I see that some code in the template is guarded with `{{#vendorExtensions.isJacksonOptionalNullable}}..{{/vendorExtensions.isJacksonOptionalNullable}}`

But not all of them.

bkabrda:
This feature is always included.

The `isJacksonOptionalNullable` is basically the only reasonable way of expressing the following condition in the mustache template: The serialization library is jackson AND this field is optional AND this field is nullable (to be precise, I also need to be able to express negation of this condition, which would be the hard part in mustache). In the places that this is not used, one (or more) of the clauses in this condition are false or not required and thus this is not necessary everywhere.

Hope this makes it clearer. I'd be happy to add some comments if that would improve the situation.

bkabrda:
Just to make the above statement clearer: this feature is always included for Java clients that use jackson as their parsing library - with the exception of jersey1, which doesn't work with new enough version of jackson (which is required for JsonNullable to work).

jimschubert:
I'm wondering if it might address everyone's comments in the PR to change the vendor extension to a generator option `useJsonNullable`. This way, you can set the default according to the logic you've already applied, but it makes the functionally both discoverable and overridable from a user's perspective.

bkabrda:
@jimschubert yeah, I think that sounds fair. I can do that. FTR, what do you think about making it default tu `true` to turn this feature on by default? From my POV it fixes current deficiencies (as opposed to bringing a completely new feature).

jimschubert:
@bkabrda I think defaulting to true is fine. In minor releases, we allow breaking changes with fallback.  I think without a user facing option, this would break all existing Java clients meaning a code change in a minor release. The option (as I understand the changes) allows a user to fallback to similar code as before.

bkabrda:
Ok folks, I really appreciate all the feedback that I received, so thanks for that. As for adding/not adding the new option, I guess that's your decision and I will implement whatever way you want. Let me try to sum up the pros of both ways:

Pros of having a new option:
- People who want to preserve exactly the same (admittedly broken) behavior that all the Java clients using Jackson had up until now will be able to do so.

Pros of not having a new option:
- Less complexity with not introducing a new option (which would likely be completely removed in a future version anyway).
- Fixing the broken behavior for everyone all the time.

Personally I would prefer to not have a new option, since the existing methods of the generated POJOs stay stable, so there is no breaking change to their API - there are only additions. The only thing left to verify is what the behavior is for XML clients, I'll try to do that today (but no promises, this PR has unfortunately already eaten too much of my time and I have to attend to other things as well).

bkabrda:
@jimschubert @cbornet alternatively, if it seems that there's no agreement on the form in which the feature is accepted, would it make sense to:

* Separate this PR into 2 smaller PRs:
  * A PR that would only do the Java code changes, IOW it would fill in the mustache context with the variables that are necessary to make this work - this change wouldn't hurt anyone, as it wouldn't change anything on the template level
  * A PR that would do the rest of the changes (including adding the actual imports in the Java code)
* We could then keep discussing the second PR until everyone is happy with it

I'd love to do at least that, since that would make it possible for me to not have to build my own openapi-generator - I would only have to use my custom templates, which is much easier to do for my usecase.

Would that make sense to you? I don't mean to bother you too much, but I really need to make this work in a reasonable way for a project I'm working on. Thanks for considering and thanks for keeping the discussion constructive :+1:

bkabrda:
So after some testing, it seems that JsonNullable doesn't serialize correctly to XML right now. This makes me think that we should adopt the approach with 2 PRs that I noted above and implement XML serialization to JsonNullable before the second PR can be merged. I'll open an RFE for JsonNullable in the meanwhile. Does this sound good to everyone?

bkabrda:
FTR, it seems I was wrong about JsonNullable not working with XmlMapper right now - a misconfiguration on my side. I pushed another commit that makes everything work, so from my POV this is 100 % functional in all cases - I'd just love to hear if it's ok to accept or I should go for the 2 PR approach (or something else).

Thanks!

modules/openapi-generator/src/main/resources/Java/libraries/google-api-client/ApiClient.mustache
```diff
@@ -53,6 +54,8 @@ public class ApiClient {
         module.addDeserializer(ZonedDateTime.class, CustomInstantDeserializer.ZONED_DATE_TIME);
         objectMapper.registerModule(module);
         {{/threetenbp}}
+        JsonNullableModule jnm = new JsonNullableModule();
```
jimschubert: I don't think I saw this when I reviewed before. If JsonNullable provides a module which uses Jackson's APIs, then I think it will work the same across formats (json, xml, yaml).  That addresses the question I previously raised.

wing328:
> People who want to preserve exactly the same (admittedly broken) behavior that all the Java clients using Jackson had up until now will be able to do so.

My suggestion is to fix it rather introducing another option to preserve the incorrect behaviour. The user can always use the old stable versions if they need the incorrect behaviour for whatever reasons.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenProperty.java
```diff
@@ -61,6 +61,7 @@
     public boolean isReadOnly;
     public boolean isWriteOnly;
     public boolean isNullable;
+    public boolean isOptionalNullable;
```
wing328: @bkabrda Seems like this is only set or processed in the Java client generator and you're using the vendor extension in the mustache template instead, right?

My take is to remove it and use only the vendor extension

bkabrda: The reason why I put it here is that I have a strong belief that most of the other templates have the same problem that I'm trying to address in this PR for Java. So that made me thing that these other templates will benefit from having access to this variable in the rendering context, as the solution will usually look somewhat similar to what I did here. I do believe that for a large number of languages, we will need to specialcase these not-required-and-nullable fields, so I felt it was appropriate to make this general.

If that explanation is not sufficient, I can definitely make it Java-specific and we can make this general as soon as we find that another language needs it.

wing328: Is `isOptionalNullable` used in the Java client template as part of this PR? I couldn't find it.

What about using `{{^required}}{{#isNullable}} ... {{/isNullable}}{{/required}}` for the time being?

bkabrda: So yes and no. Currently, the `isJacksonOptionalNullable` variable is used in the Java client templates. The reason for that is that I couldn't find an easy way to make this work with gson - so my initial PR used `isOptionalNullable`, but then I moved to using `vendorExtensions.isJacksonOptionalNullable` (which is basically `isOptionalNullable && jackson`).

It is possible to use `{{^required}}{{#isNullable}} ... {{/isNullable}}{{/required}}`, but the problem is that sometimes you need an "else clause" to that (IOW a reverse of that condition), which is impossible to do in mustache AFAICS - you'd need to do explicitly do something else for all the other three combinations of possible values of `required` and `isNullable`.

So if this seems to be a problem for you, I can definitely drop this for now and only use `isJacksonOptionalNullable` and we can reintroduce this whenever it's used in another set of templates to share the functionality.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -800,7 +802,13 @@ public String toDefaultValue(Schema p) {
                 }
             }
             return null;
+        } else if (ModelUtils.isObjectSchema(p)) {
+            if (p.getDefault() != null) {
+                return super.toDefaultValue(p);
+            }
+            return null;
```
wing328: @bkabrda thanks for adding the condition to handle default value for model. How is this going to work? Will the default value of a model become a string? or we need to pass the model property one by one somehow?

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenProperty.java
wing328: I understand your intention to make it easier for all generators to easily support optional nullable field but given that it's not used at the moment, I suggest we remove it from this PR for the time being to keep this PR simple (easier to get it merged)

(btw, when adding a new property to the Codegen Property class, you will also need to update other methods, e.g. https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenProperty.java#L538)

bkabrda: Ok, I'll remove it. 👍

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
bkabrda: This effectively creates a distinction between the default value being `null` (which would be explicit `null` for JsonNullable fields) and no default value - with the previous behaviour, the default value for referenced schemas was always `null`, which doesn't work well with JsonNullable.

This doesn't really change anything, because it will do a diff from `private MyModel foo = null;` to `private MyModel foo;`, which are equivalent. But for JsonNullable, it will correctly do either `foo = new JsonNullable<MyModel>(null)` (when the default is explicit null) or `foo = new JsonNullable<MyModel>()` (when there is no default value, thus meaning that we don't want to send explicit null by default).

Hope this makes it clearer.

wing328: Thanks for the explanation.

bkabrda:
@wing328 thank you for your review! I pushed additional 3 commits:
* One that removes the `isOptionalNullable` field of `CodegenProperty`; it also makes sure that the imports for `JsonNullable` are only added when at least one of the model's fields actually uses that.
* One that fixes initialization of `JsonNullable` objects by using proper `JsonNullable.<CORRECTTYPE>of(...)`.
* One that regenerates the samples after the two above changes (hopefully keeping these separate will make it easier for you to review).

bkabrda:
Hmm, so this fails in travis for `mvn clean compile -f modules/openapi-generator-maven-plugin/examples/java-client.xml` which it didn't do before. I tried adding the nullable module as a dependency of the maven plugin, but that didn't help. I can't reproduce this locally, so I'd appreciate any help with this.

modules/openapi-generator/src/main/resources/Java/libraries/google-api-client/ApiClient.mustache
bkabrda: FTR, I also made a PR for jackson-databind-nullable to document how to use JsonNullable with jackson's XML-serialization: https://github.com/OpenAPITools/jackson-databind-nullable/pull/11/files

wing328:
@bkabrda can you try merging the latest master of openapi-generator (official repo) into your branch?

bkabrda:
@wing328 that doesn't seem to have helped. I guess I need to add jackson-databind-nullable as a dependency somewhere, but I can't seem to figure out where (and the same command still works for me when I run it locally).

jmini:
> I tried adding the nullable module as a dependency of the maven plugin

Of the plugin or of the project itself?
I think you need to add it as dependency of the project.

I think that following command (I did not try on your branch):
```
mvn clean generate-sources -f modules/openapi-generator-maven-plugin/examples/java-client.xml
```
Generates code (under `modules/openapi-generator-maven-plugin/examples/target/generated-sources/`) requiring the jackson-nullable lib.

jmini:
Hint: If I remember well, I think that travis is merging master into the PR branch before running (you might try this to reproduce).

jmini:
Travis is also installing the maven-plugin into the local maven repo.

So when it runs, it is the updated version of the maven plugin (built from your sources) that is executed (and not one coming from maven-central).

bkabrda:
@jmini thanks a lot for this, I can reproduce the error locally now, so hopefully I'll be able to fix it 👍

bkabrda:
Ok, so the tests are passing now and I feel like I've addressed all the comments that were raised. Please let me know if this is ok to be merged or if there's something more that I need to improve on.

wing328:
@bkabrda I wonder if you can help resolve the merge conflicts when you've time. (I've merged @jmini PR before yours)

I'll put other Java PR on hold before merging this one.

bkabrda:
Rebased and pushed - let's wait to see if the tests pass. I actually found out one minor bug while testing this PR and fixed it - this made me realize that there are actually no tests for OpenAPI 3 Java clients. I did as much testing as I could, but this is pretty complex and someone could break it with their PRs. When this is merged, I'd love to write tests, but can't guarantee 100 % that I'll have time to do so, so I'd appreciate if someone could help with this (testing various combinations of nullable/required/readOnly/container fields and how they behave during serialization would be great).

wing328:
>  I actually found out one minor bug while testing this PR and fixed it 

@bkabrda can you please elaborate a bit more on the bug you fixed?

bkabrda:
@wing328 it was related to https://github.com/OpenAPITools/openapi-generator/pull/3615 - previously, I was assuming in `add{{AListField}}Item` and `put{{AMapField}}Item` that the attributes were never null, which got changed by that PR to being initially null. I discovered that during my local testing, but it would be really awesome to have some tests for at least one jackson-based Java client for OpenAPI 3 for various combination of field attributes as I noted in my comment above.

bkabrda:
(Oh, I just realized I might have misunderstood your question. I didn't find a bug that was already in current code of openapi-generator, I found a bug that was result of combining PR #3615 with this PR when I rebased. That's why I said we should have tests for openapi 3 for Java clients.)

wing328:
> That's why I said we should have tests for openapi 3 for Java clients.)

Totally agreed. I added new some tests earlier this week: https://github.com/OpenAPITools/openapi-generator/pull/3640

Please feel free to add more later with a separate PR.

jmini:
> That's why I said we should have tests for openapi 3 for Java clients.

I am going to present something based on [Mock-Server](http://www.mock-server.com/) in #689, feel free to follow this issue and/or to add your comment.

----

I am sorry I could not test this before it was merged, so I am reviewing the impact of this PR now.

I observe `@JsonProperty` being moved from the field to the getters.
=> is that intended?

I observe `@JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)` on the getter. 
=> What is the impact?

@bkabrda I hope you can elaborate on those points

bkabrda:
> I observe @JsonProperty being moved from the field to the getters.

Yes, It's intended. It was required for the JsonNullable attribute (I couldn't get this to work with JsonProperty on the JsonNullable attribute itself for some reason, but I couldn't figure out why). It is fully backwards compatible AFAICS.

> I observe @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS) on the getter.

Yeah, I used this to make sure we have explicit statement on what the behavior is on inclusion of all fields to make sure the behavior is stable. I probably could have assumed that `USE_DEFAULTS` [1] is always the default if nothing is specified, but I thought this was safer. There should be no backwards incompatible effect of this. (But if you don't want it there, it can be removed safely, I think.)

[1] https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonInclude.Include.html#USE_DEFAULTS

jmini:
> @JsonProperty being moved from the field to the getters.

I did some tests with:

* an object containing a boolean property (I tried with the `is` and with `get` prefix)
* an object containing an `_` such that the field name and the serialized string in the annotation are not the same.

Both worked for me.
=> It also makes https://github.com/OpenAPITools/openapi-generator/issues/3573 less important, because it fixes it as side effect.

---

About using all the Json annotations in the model or not, I am not sure.
I am not a big fan of it, especially because in this case, it is not changing anything, so it could be removed.

This Jackson-annotation topic goes also behind this PR: 
There is a PR #3441 to add `@JsonIgnoreProperties` on the class. Somehow redundant with the Mapper Configuration we provide, but I can understand that some users prefers the annotation on the model. This way they can use any mapper instance (without having to change the config). If you have an opinion, fell free to comment.

jmini:
> I only added this for jackson, as I didn't figure out an easy way to do this for gson. I could try working on a different PR that would do this for gson, but that would take me a lot more time.

The problem is that `jackson-databind-nullable` is made for Jackson.
For gson, something similar to [`JsonNullable`](https://github.com/OpenAPITools/jackson-databind-nullable/blob/master/src/main/java/org/openapitools/jackson/nullable/JsonNullable.java) could be used, but the rest (serializer / deserializer / ...) must be written the JSON way.

bkabrda:
> The problem is that jackson-databind-nullable is made for Jackson.
For gson, something similar to JsonNullable could be used, but the rest (serializer / deserializer / ...) must be written the JSON way.

Yeah. I'll admit I'm not familiar with gson too much (and Java in general TBH), so making this PR work just for jackson was complex enough for me that I didn't want to go into it. The way I understand it, it might be possible to make JsonNullable compatible with gson - I think that would be preferrable to using a different nullable implementation for each serialization library.

@jmini if you see any specific followup things to fix, please let me know, I'll be happy to look into them. Thanks!

wing328:
@bkabrda thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

jmini:
The new dependency was not added to the generated `build.gradle`. See PR https://github.com/OpenAPITools/openapi-generator/pull/3793 to fix it.

RockyMM:
Hello all,

From my perspective, as a user of  `useXml` option, this PR produces a regression I described in #4014. Can anyone take a look?

Thanks!

PR #3494, opened 2019-07-29
[Rust Server] Allow more than 32 configured paths

richardwhiuk:
### PR checklist

@bjgill @farcaller @frol 

### Description of the PR

Currently the Rust Server generator uses an array to store the paths available on an API.

If the API has more than 32 configured paths, then the produced API fails to compile with:

```
error[E0277]: `&[&str; 37]` is not an iterator
   --> src/server/mod.rs:113:60
    |
113 |         pub static ref GLOBAL_REGEX_SET: regex::RegexSet = regex::RegexSet::new(&[
    |                                                            ^^^^^^^^^^^^^^^^^^^^ `&[&str; 37]` is not an iterator
    |
    = help: the trait `std::iter::Iterator` is not implemented for `&[&str; 37]`
    = note: required because of the requirements on the impl of `std::iter::IntoIterator` for `&[&str; 37]`
    = note: required by `server::paths::regex::RegexSet::new`
```

This is because Iterator is only implemented for arrays up to 32 in size.

This change avoids that limitation by moving to using a Vec, and adds a test API to the Rust Server tests which has 33 paths, to avoid hitting this limit again.


modules/openapi-generator/src/test/resources/3_0/rust-server/ops-v3.yaml
```diff
@@ -0,0 +1,192 @@
+# Test large number of configured operations
```
bjgill: 😻 the new test!

Would be nice if this had a more descriptive name than `ops-v3`, but not essential.

Issue #1610, opened 2018-12-04
Codegen operation scopes inconsistency


dmolnarqu:
##### Description

Operation auth method scopes in template currently hold all defined scopes for specific security definition.
Swagger 2.* populates the list only with scopes defined for that particular operation.

Example:
If operation1 requires `scope:op1`, but there's also `scope:op2` defined globally, then
```
-- operation1 context
...{{#authMethods}}{{#scopes}}{{scope}}, {{/scopes}}{{/authMethods}}..
```
swagger codegen produces
```
scope:op1, 
```
openapi generator produces
```
scope:op1, scope:op2, 
```

This issue was mentioned before in #1043.

Is this change intended? It's neven been questioned in that issue.
I could find that ada server generator fixes it at higher level (#1044), but shouldn't it be responsibility of base codegen? 
Is there a more direct way for a template to access scopes required only by operation itself?

##### openapi-generator version
3.3.4

PR #1984, opened 2019-01-25
general support to add scopes for bearer auth too

MBcom:
implemented authorize workflow in aspnet core too

### PR checklist

- [x ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x ] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [x ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

this PR solves #1983 and adds role/scope based security support to each route

PR #3495, opened 2019-07-29
[Core, Rust Server, ASP.NET Core] Fix Codegen Operation Scope Consistency

richardwhiuk:
### PR checklist

 - I've run both the Rust Server changes, and the ASP.NET `petstore.sh` scripts.

   - Rust: @frol @farcaller  @bjgill
   - ASP.NET: No technical committee exists.

### Description of the PR

- Filter scopes based on operation - this fixes https://github.com/OpenAPITools/openapi-generator/issues/1610

- Partially revert #1984 to not rely on custom attributes as to whether scopes exist, and instead use  built in Mustache function. - @MBcom you may be interested in the change here.

- Tests are added to Rust Server codegen examples to check that the generated code is correct.

Code is by @mthebridge and @richardwhiuk . Any bugs are mine, any credit is Mark's :smile:.


samples/server/petstore/rust-server/output/openapi-v3/src/server/context.rs
```diff
@@ -83,6 +83,16 @@ impl<T, A, B, C, D> hyper::server::Service for AddContext<T, A>
...
 
+        {
+            use hyper::header::{Authorization as HyperAuth, Basic, Bearer};
+            use std::ops::Deref;
+            if let Some(bearer) = req.headers().get::<HyperAuth<Bearer>>().cloned() {
+                let auth_data = AuthData::Bearer(bearer.deref().clone());
+                let context = context.push(Some(auth_data));
+                let context = context.push(None::<Authorization>);
+                return self.inner.call((req, context));
+            }
```
bjgill: Just want to check that this is intentional. Is it OK to do nothing if we don't have a bearer header? Or am I misreading how this works?

bjgill:
Also - looks as if CI is unhappy.

MBcom:
@richardwhiuk Sorry it took so long to have a look to your pr  
here a sample oas3 definition
```yaml
 delete:
      summary: Remove a sth
      operationId: removeSth
      responses:
        '204':
          description: Operation successfull
        '404':
          description: not found
      security:
          - bearerAuth:
            - sth:delete
```  
that should be translated to the following authorize annotation
```c#
        [Authorize(Roles = "sth:delete")]
```  
but your code returns the following  
```c#
        [Authorize]
```  
so something is going wrong on creating the `Authorize` annotation  
at first sight the template looked right - maybe you need to take a look at your security class??

richardwhiuk:
@MBcom 

I'm slightly confused by your example, as I'm not sure how you were defining the `bearerAuth` security scheme referenced in your example.

If you had something like the following:

```yaml
components:
  securitySchemes:
    bearerAuth:
      scheme: bearer
      bearerFormat: token
      type: http
security:
  - bearerAuth:
    - "sth:access"
```

Then note, that according to https://swagger.io/specification/#securityRequirementObject, the security requirement may not contain a list of scope names. In this case, the code will generate `[Authorize]`, which might not be correct, but I'm not sure what the correct answer is.

If, instead, you were defining a `bearerAuth` as a global level oauth2 level scheme, then this should be fixed by https://github.com/OpenAPITools/openapi-generator/pull/3495/commits/664c71c4f2071ad68c8686739992f6e6caf2df2e

Can you let me know, so I can address any further issues here?

samples/server/petstore/rust-server/output/openapi-v3/src/server/context.rs
richardwhiuk: It's definitely fine here - because this code is API operation agnostic, and the authentication requirements may be different on a per operation basis.

An absent bearer will either be handled by the OpenAPI scope code (rejecting an absent bearer) or by the scope checking code (which will reject the request as it doesn't have sufficient scopes on it).

The former code isn't provided by the auto-generated code, the later is provided by https://github.com/OpenAPITools/openapi-generator/pull/3495/files/b1df259d3824c7de0b27feac684d5f5bd4577d6e#diff-1214066435f929c4d1bc643c36364470R143 where we check whether the Authorization scopes match.

richardwhiuk:
@MBcom  can you comment?

If I don't hear back, I'll assume that https://github.com/OpenAPITools/openapi-generator/commit/664c71c4f2071ad68c8686739992f6e6caf2df2e fixes your issue and go ahead and merge this - but I don't want to break ASP inadvertently...

MBcom:
@richardwhiuk sry for my late answer  

when there is something like this
```yaml
delete:
      summary: Remove a sth
      operationId: removeSth
      responses:
        '204':
          description: Operation successfull
        '404':
          description: not found
      security:
          - bearerAuth: []
```  
you are right - there can be an `[Authorize]` or an `[Authorize(Roles = "")]` annotation   
  
but in my example of https://github.com/OpenAPITools/openapi-generator/pull/3495#issuecomment-529877347 - the specific route must be annotated including the roles/ scopes
  
and yes, a global security annotation is currently missing - we should open a new issue for that

wing328:
@richardwhiuk thanks for the PR, which has been included in the v4.2.2 release: https://twitter.com/oas_generator/status/1201432648544972800

Issue #3260, opened 2019-07-02
[BUG] Inline model resolving failure when the model name contains dot `.`


yue9944882:
running go generator using the following definitions as input:

```yaml
  io.kimmin.InlineFailure:
  #InlineFailure:
    type: object
    properties:
      spec:
        type: object
        properties:
          foo:
            type: integer
```

the generated go codes will be:

```go
package petstore

type IoKimminInlineFailure struct {
	Spec map[string]interface{} `json:"spec,omitempty"` // `map[string]interface{}` should be resolved
}
```

by debugging in my local environment, i happen to find out that everything will be happy when commenting the following code from `ModelUtils#getSimpleRef`:
 
```java
    public static String getSimpleRef(String ref) {
        if (ref.startsWith("#/components/")) {
            ref = ref.substring(ref.lastIndexOf("/") + 1);
        } else if (ref.startsWith("#/definitions/")) {
            ref = ref.substring(ref.lastIndexOf("/") + 1);
        } else {
-            LOGGER.warn("Failed to get the schema name: {}", ref);
-            //throw new RuntimeException("Failed to get the schema: " + ref);
-            return null;

        }

        return ref;
    }
```

then the correct output will be:

```
package petstore

type IoKimminInlineFailure struct {
	Spec map[string]IoKimminInlineFailureSpec `json:"spec,omitempty"` 
}
```

am not sure, if it's the correct fix but seems working for me..

@wing328 ptal

wing328:
As discussed, I think it's due to nested inline schemas defined in the spec.

As a workaround, you can define these models separately, which can be reused throughout the spec.

`getSimpleRef` is used quite heavily so we need to be very careful when changing it.

After your patch, if you run `./bin/utils/ensure-up-to-date`, does any sample get changed?

cc @OpenAPITools/generator-core-team

yue9944882:
note that everything will work fine if the model name doesn't contain `.`

yue9944882:
update: the issue also applies to java generator

yue9944882:
politely bumping the issue

jmini:
Can you provide a complete example to reproduce the error: You can start with [`ping.yaml`](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/ping.yaml) and edit it, so that your problem can be reproduced?

yue9944882:
try run this spec and you'll see the surprise

```yaml
[...]

PR #3498, opened 2019-07-30
Better handling of dot in inline model name

wing328:
### PR checklist

### Description of the PR

Better handling of non-alphanumeric characters in inline model naming.

Fixes https://github.com/OpenAPITools/openapi-generator/issues/3260

Tested locally with the spec provided by the user and the issue no longer exists.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -416,6 +416,9 @@ public class InlineModelResolver {
      */
     private String resolveModelName(String title, String key) {
         if (title == null) {
+            // for auto-generated schema name, replace non-alphanumeric characters with underscore
+            // to avoid bugs with schema look up with inline schema created on the fly
+            // e.g. io.schema.User_name => io_schema_User_name
             return uniqueName(key).replaceAll("[^A-Za-z0-9]", "_");
         } else {
             return uniqueName(title);
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -407,11 +407,18 @@ public class InlineModelResolver {
         }
     }
 
+    /**
+     * Generates a unique model name. Non-alphanumeric characters will be replaced
+     * with underscores
+     *
+     * @param title String title field in the schema if present
+     * @param key String model name
+     */
     private String resolveModelName(String title, String key) {
         if (title == null) {
-            return uniqueName(key);
+            return uniqueName(key).replaceAll("[^A-Za-z0-9]", "_");
         } else {
-            return uniqueName(title);
+            return uniqueName(title).replaceAll("[^A-Za-z0-9]", "_");
         }
     }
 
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -407,11 +407,18 @@ private void fixStringModel(Schema m) {
...
+     * Generates a unique model name. Non-alphanumeric characters will be replaced
+     * with underscores
+     *
+     * @param title String title field in the schema if present
+     * @param key String model name
+     */
     private String resolveModelName(String title, String key) {
         if (title == null) {
-            return uniqueName(key);
+            return uniqueName(key).replaceAll("[^A-Za-z0-9]", "_");
```
macjohnny: are `$` or `_` not allowed to be in the model name?

etherealjoy: I think `$` would mean something else in some scripting languages

macjohnny: @wing328 how about `_`?

wing328: Very good question. The spec doesn't restrict the use of symbols (e..g $, _, .) in model naming.

This is auto-generated model name based on the parent model name and the property name. The issue occurs when the parent model name contains `.`.

Since the model name is auto-generated if `title` attribute is not present, this fix will simply get rid of non-alphanumeric characters to make our life easier.

If the user wants complete control of the model naming, they can still use `title` (https://github.com/OpenAPITools/openapi-generator/pull/3498/commits/35e49b9ef66556d34edba357d8a897caf7932ec5)

jimschubert:
I think this might be getting done too close to spec reading, making the time far too restrictive.

The `title` in JSON Schema is not necessarily intended to be used as a model name (http://json-schema.org/latest/json-schema-validation.html#rfc.section.10.1). This would be equivalent to using summary to generate an `operationId` in a Path object.

It seems this would mean that users can't define model names such as `user_name` directly in the schema without adding the optional (and redundant) `title: user_name`. I know this will break several of my existing specs, and will require that I add support for `title` to custom documentation builders (in Finagle). I imagine it's a similar breaking change for many others.

Also, while not a common use case, it means we disallow any Unicode characters in model names.

I'm wondering if we could add the fix in the CodegenConfig instance instead?

wing328:
> It seems this would mean that users can't define model names such as user_name directly in the schema without adding the optional (and redundant) title: user_name

They can. This change will still auto-generate the model name as "user_name" if the parent model name is "user" and the property name is "name".

To be clear, this is not the ultimate (inline) model name users will see in the auto-generated code. For "user_name", users will instead find the model "UserName" in auto-generated Java client/server-side code.

> The title in JSON Schema is not necessarily intended to be used as a model name (http://json-schema.org/latest/json-schema-validation.html#rfc.section.10.1).

I agree `title` may not be the best attribute to name the model and this PR does *not* change the usage of `title` in the current implementation.

jimschubert:
@wing328 I see. We may want a comment about that behavior in this code. It seems by looking at the code that we're transforming spec document content at a point before generators have a say in the matter, which I think should be avoided.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
richardwhiuk: Why does it matter if this name has a dot in it? Should we instead change `uniqueName` so it doesn't generate a name with a dot in it?

wing328: > Why does it matter if this name has a dot in it? 

It's causing a bug later that the model (created on the fly) cannot be looked up. Again this is auto-generated "schema" name, which will be used to determine the model name using `toModelName`.

This change has no impact on the schema name in the spec provided by the user.

wing328:
@jimschubert added more comments via 38d772be04

PR #3501, opened 2019-07-30
Add lowercase lamda

### LINKED ISSUE #3381: [BUG] [SPRING-CLOUD] Invalid generated ConfigurationProperties annotation
#### Bug Report Checklist

##### Description

The generated Client Configurations class generates invalid properties annotation, when you use a title with uppercase characters. Camel case is not allowed for this annotation.

The yml
```
info:
  version: v1
  title: My Default API
```

Output is:
```
  @Bean
  @ConditionalOnProperty("myTestApi.security.macma.client-id")
  @ConfigurationProperties("myDefaultApi.security.x")
  public ClientCredentialsResourceDetails macmaResourceDetails() {
    ClientCredentialsResourceDetails details = new ClientCredentialsResourceDetails();
    details.setAccessTokenUri("test");
    return details;
  }
```

`@ConfigurationProperties("myDefaultApi.security.x")` is not allowed

see https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html
```
The prefix value for the annotation must be in kebab case (lowercase and separated by -, such as acme.my-project.person).

```

4.0.3

##### Suggest a fix

Camel case title should be seperated by a dot with just lower case characters or should be completely lower case.
```my.default.api.security.x```

or
```mydefaultapi.security.x```

### PR DESCRIPTION
**Pull request by SirWayne**
fixes #3381

### PR checklist

- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master` . Default: `master`.
- [x] @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04)
--

### Description of the PR

Fix invalid spring properties prefix

### DIFF
--- a/modules/openapi-generator/src/main/resources/JavaSpring/libraries/spring-cloud/clientConfiguration.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/libraries/spring-cloud/clientConfiguration.mustache
@@ -24,10 +24,10 @@ public class ClientConfiguration {
 
 {{#authMethods}}
     {{#isBasic}}
-  {{=<% %>=}}@Value("${{{#lambda.lowercase}}<%title%>{{/lambda.lowercase}}.security.<%name%>.username:}")<%={{ }}=%>
+  {{=<% %>=}}@Value("${{{#lambda.lowercase}}{{{title}}}{{/lambda.lowercase}}.security.{{{name}}}.username:}")<%={{ }}=%>
   private String {{{name}}}Username;
 
-  {{=<% %>=}}@Value("${{{#lambda.lowercase}}<%title%>{{/lambda.lowercase}}.security.<%name%>.password:}")<%={{ }}=%>
+  }@Value("$'{'{{#lambda.lowercase}}{{{title}}}{{/lambda.lowercase}}.security.{{{name}}}.password:'}'")
   private String {{{name}}}Password;
 
   @Bean
--- a/modules/openapi-generator/src/main/resources/JavaSpring/libraries/spring-cloud/clientConfiguration.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/libraries/spring-cloud/clientConfiguration.mustache
@@ -31,7 +31,7 @@ public class ClientConfiguration {
   private String {{{name}}}Password;
 
   @Bean
-  @ConditionalOnProperty(name = "{{#lowercase}}{{{title}}}{{/lowercase}}.security.{{{name}}}.username")
+  @ConditionalOnProperty(name = "{{#lambda.lowercase}}{{{title}}}{{/lambda.lowercase}}.security.{{{name}}}.username")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: modules/openapi-generator/src/main/resources/JavaSpring/libraries/spring-cloud/clientConfiguration.mustache [hunk 1]
```diff
@@ -31,7 +31,7 @@ public class ClientConfiguration {
   private String {{{name}}}Password;
 
   @Bean
-  @ConditionalOnProperty(name = "{{{title}}}.security.{{{name}}}.username")
+  @ConditionalOnProperty(name = "{{#lowercase}}{{{title}}}{{/lowercase}}.security.{{{name}}}.username")
```
macjohnny: shouldn't the lambda be invoked with `{{#lambda.lowercase}}`, as described in https://github.com/OpenAPITools/openapi-generator/blob/master/docs/templating.md#mustache-lambdas ?

SirWayne:
@macjohnny Will this work?

```
 {{=<% %>=}}@Value("${{{#lambda.lowercase}}<%title%>{{/lambda.lowercase}}.security.<%name%>.username:}")<%={{ }}=%>
  private String {{{name}}}Username;
```

macjohnny:
> Will this work?

I guess it wont, because the `{{=<% %>=}}` indicates erb style tags, see http://mustache.github.io/mustache.5.html

maybe this should be converted to normal mustache tags, and escape the `{` and `}`

SirWayne:
Ok got it. Yes it would be better to escape the `{`, but there is no default way to escape braces.

It think a variable is a good solution:
```
 public char openbrace= `{`;
 public char closebrace= `{`;
```

Then i could use:
```
@Value("${{openBrace}}{#lambda.lowercase}}{{{title}}}{{/lambda.lowercase}}.security.{{{name}}}.username:{{closebrace}}")
```

Do you have a proposal,where i should add the `openbrace` variable?

macjohnny:
how about
```mustache
 @Value("${{=<% %>=}}{<%={{ }}=%>{{#lambda.lowercase}}{{{title}}}{{/lambda.lowercase}}.security.<%name%>.key:{{=<% %>=}}}<%={{ }}=%>")
 private String {{{name}}}Key; 
```

SirWayne:
This will work, but is not really readable. 
That's why I like the workaround with the variable. It's readable and easy to understand.

macjohnny:
ok, a variable is a good solution. you can add it e.g. here
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java#L115

wing328:
@SirWayne thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

PR #3502, opened 2019-07-30
[java-client][okhttp-gson] fixes for interceptors in ApiClient.java

### LINKED ISSUE #3497: [BUG] [JAVA] okhttp-gson generates broken method in ApiClient.java
##### Description

The current Okhttp3 - ApiClient [Mustache](https://github.com/OpenAPITools/openapi-generator/blob/f04ef76d0763984c2c37056d94c4d500e627baf0/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache#L195-L209) generates a broken method.
the current `setHttpClient(OkHttpClient newHttpClient)` copies the interceptors from the current field to the new variable, ignoring the fact that the interceptors are already present.
The result is, that each call to setHttpClient duplicates the current interceptors.

Okhttp made the HttpClient immutable and provided a builder-pattern which already captures the current state.

4.0.2

##### Steps to reproduce

- Configure Generator to use okhttp-gson
- Generate some API
- Prepare ApiClient
- Add Interceptor `apiClient.setHttpClient(apiClient.getHttpClient().newBuilder().addInterceptor(interceptor).build())`
- Repeat the previous call and check the Interceptors (the first one is now duplicated)

##### Suggest a fix

```
public ApiClient setHttpClient(OkHttpClient newHttpClient) {
   this.httpClient = newHttpClient;
}
```

###### Workaround with Groovy
We faced this problem in our Spock-Tests (Groovy) and thanks to Groovy we can bypass the setHttpClient method and set the new client directly using the @-Syntax

`apiClient.@httpClient = apiClient.httpClient.newBuilder().addInterceptor(interceptor).build()`

lostiniceland:
ApiClient for Okhttp3 must not copy the interceptors when setting the HttpClient.

fixes #3497

### DIFF
--- a/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
+++ b/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
@@ -335,7 +335,7 @@ public class ApiClientTest {
         OkHttpClient oldClient = apiClient.getHttpClient();
         assertEquals(1, oldClient.networkInterceptors().size());
 
-        OkHttpClient newClient = new OkHttpClient();
+        OkHttpClient newClient = apiClient.getHttpClient().newBuilder().build();
         apiClient.setHttpClient(newClient);
         assertEquals(1, apiClient.getHttpClient().networkInterceptors().size());
         apiClient.setHttpClient(newClient);
--- a/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
+++ b/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
@@ -10,6 +10,7 @@ import java.util.TimeZone;
 
 import org.junit.*;
 import static org.junit.Assert.*;
+import static org.hamcrest.CoreMatchers.*;
 
 
 public class ApiClientTest {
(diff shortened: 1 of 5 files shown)


samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
```diff
@@ -335,7 +335,7 @@ public void testInterceptorCleanupWithNewClient() {
         OkHttpClient oldClient = apiClient.getHttpClient();
         assertEquals(1, oldClient.networkInterceptors().size());
 
-        OkHttpClient newClient = new OkHttpClient();
+        OkHttpClient newClient = apiClient.getHttpClient().newBuilder().build();
```
jmini: CircleCI reports that this change gets overridden when the samples are updated:

```diff
diff --git a/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java b/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
index 3675d8645a..6181e8afa2 100644
--- a/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
+++ b/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/ApiClientTest.java
@@ -335,7 +335,7 @@ public class ApiClientTest {
         OkHttpClient oldClient = apiClient.getHttpClient();
         assertEquals(1, oldClient.networkInterceptors().size());
 
-        OkHttpClient newClient = apiClient.getHttpClient().newBuilder().build();
+        OkHttpClient newClient = new OkHttpClient();
         apiClient.setHttpClient(newClient);
         assertEquals(1, apiClient.getHttpClient().networkInterceptors().size());
         apiClient.setHttpClient(newClient);
```

This is because the source for this file is in the `CI/samples.ci` folder. Can you update this file:
https://github.com/OpenAPITools/openapi-generator/blob/6824bd8b2445e54c25f26589ac51eac3de62f0e9/CI/samples.ci/client/petstore/java/test-manual/okhttp-gson/ApiClientTest.java#L338

jmini:
@lostiniceland: Thank you a lot for this PR, can you also change the mentioned file?

lostiniceland:
Updated as requested by review.

I would argue though that the Tests I've fixed are not very usefull at all. They might have been with the old "logic" but now the tested method is a plain setter, except the check for equality, which could also be removed.
A usefull invariant would be that the passed client must not be null, because the ApiClient cannot function without a httpClient.

lostiniceland:
@jmini I've added another commit which contains the mentioned invariant and removed the unnecessary tests

wing328:

lostiniceland:
@wing328 thanks for pointing that out. I've added my work-email to Github so it should match now.

wing328:
@lostiniceland thanks for the PR, which has been included in the 4.1.0 release: https://twitter.com/oas_generator/status/1160000504455319553

Issue #1475, opened 2018-11-17
[PHP] Invalid enum const names


AlwinGarside:
##### Description

When generating clients for the Bungie OpenAPI v2 spec, certain enum types with unnamed integer values result in invalid constant names, consisting only of integers.

What I mean, concretely, is that this:

```json
    "GroupsV2.GroupType": {
      "format": "int32",
      "enum": [
        "0",
        "1"
      ],
      "type": "integer",
      "x-enum-values": [
        {
          "numericValue": "0",
          "identifier": "General"
        },
        {
          "numericValue": "1",
          "identifier": "Clan"
        }
      ]
    }, 
```

results in this:

```php
class GroupsV2GroupType
{
    /**
     * Possible values of this enum
     */
    const 0 = 0;
    const 1 = 1;
    
    /**
     * Gets allowable values of the enum
     * @return string[]
     */
    public static function getAllowableEnumValues()
    {
        return [
            self::0,
            self::1,
        ];
    }
}
```
(note the `const 0 = 0;`)

For the complete spec file, check:
https://raw.githubusercontent.com/Bungie-net/api/master/openapi-2.json

##### openapi-generator version

3.3.4-SNAPSHOT, commit 32b8d7fee7886c13a8ccd9523dd31791cdf6371b 

##### OpenAPI declaration file content or url

https://raw.githubusercontent.com/Bungie-net/api/master/openapi-2.json

##### Command line used for generation

```bash
java -jar openapi-generator-cli.jar generate \
      --generator-name php \
      --config         swagger-codegen-config.php.json \
      --input-spec     https://raw.githubusercontent.com/Bungie-net/api/master/openapi-2.json \
      --output         bungie-sdk-php \
      -DapiTests=false \
      -DmodelTests=false
```

swagger-codegen-config.php.json:
```json
{
  "composerVendorName": "yogarine",
  "composerProjectName": "bungie-sdk-php",
  "artifactVersion": "2.3.2",
  "invokerPackage" : "Bungie",
  "modelPackage": "Model",
  "apiPackage": "API",
  "variableNamingConvention": "camelCase",
  "packagePath": "bungie-sdk-php",
  "srcBasePath": "src",
  "gitUserId": "Yogarine",
  "gitRepoId": "bungie-sdk-php",
  "validateSpec": false
}
```

##### Steps to reproduce

Generate with the command provided above.

##### Related issues/PRs

None I could find.

##### Suggest a fix/enhancement

[...]

PR #3524, opened 2019-08-01
[php][DefaultCodegen.java] Fix invalid enum const names

githubERIK:
@jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @ybelenko, @renepardon

Fix https://github.com/OpenAPITools/openapi-generator/issues/1475


githubERIK:
The initial failing tests show that some enum number variable names were converted (-1.2 ENUM_NUMBER_MINUS_1_DOT_2) previously as well. The new changes added prefix `N` to the non-converted enum numbers and affected some of those previously converted variables too.

githubERIK:
How is it possible that some enum `1` is transformed to `ENUM_INTEGER_1` while some numeric enums just stay the same?

githubERIK:
Reverted add-N-prefix changes with https://github.com/OpenAPITools/openapi-generator/pull/3524/commits/aa36b80d9dc2d82c81aecddae8985b36a4ca1616 and https://github.com/OpenAPITools/openapi-generator/pull/3524/commits/d07efac2008644697b5edcf64842daa71a8efbb0.

The new approach, however, does not affect the clients which means that the changed method `toEnumVarName` is not called.

githubERIK:
It seems to be a good idea to debug the generated jar and start with a breakpoint in DefaultCodegen and then perform "Step Into" when "importPath→OpenAPI\Client\Model.EnumTest" appears.

Screenshot 2019-08-09 at 10 20 29


and then see what happens with outerEnumInteger

```
    Enum_Test:
      type: object
      required:
        - enum_string_required
      properties:
        enum_string:
          type: string
          enum:
            - UPPER
            - lower
            - ''
        enum_string_required:
          type: string
          enum:
            - UPPER
            - lower
            - ''
        enum_integer:
          type: integer
          format: int32
          enum:
            - 1
            - -1
        enum_number:
          type: number
          format: double
          enum:
            - 1.1
            - -1.2
        outerEnum:
          $ref: '#/components/schemas/OuterEnum'
        outerEnumInteger:
          $ref: '#/components/schemas/OuterEnumInteger'
        outerEnumDefaultValue:
          $ref: '#/components/schemas/OuterEnumDefaultValue'
        outerEnumIntegerDefaultValue:
          $ref: '#/components/schemas/OuterEnumIntegerDefaultValue'
```

githubERIK:
The final approach: change `postProcessModelsEnum` in DefaultCodegen.java to prevent adding enum variable names that are numbers. If number is detected, add a prefix `NUMBER_`.

Generated all samples by executing shell scripts in bin and bin/openapi3 folders.

ybelenko:
Is this PR is still relevant or it should be closed?

asabramo:
Hi guys,
I see this issue was reported 2 years ago.
When I use the latest:
swagger-codegen-cli-3.0.20.jar generate -l php -i api.yaml -o generated_api

I'm still seeing this issue, which means I can't use the files as they are generated - I have to add another step manually, which beats the purpose.

Oddly enough, when I use the version that the Jetbrains plugin retrieves automatically (https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.3/swagger-codegen-cli-3.0.3.jar) , the enum problem doesn't exist (but then there's another bug - BigDecimal instead of float). 

Is there a version where both issues are fixed?
Is there any workaround for the enum issue?
Thanks,
Assi

ybelenko:
@asabramo It seems that you're talking about `Swagger Codegen` which is different project. More information you can read at: https://github.com/OpenAPITools/openapi-generator/blob/master/docs/qna.md

About bug itself, I don't know whether it fixed or not. Maybe we should add part of the spec from related issue to samples and check PHP syntax of generated models with `phplint` package during CI.

Issue #3168, opened 2019-06-18
[BUG] [PYTHON] model attribute names wrong


adrpp:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

There are 3 attributes on model 'Resources', each shows buggy behavior of client / server generated attribute names.

1. 
yaml: `Resources._attributePrefixedWithUnderscore`
generated: `Resources.attribute_prefixed_with_underscore`
- why is leading underscore taken away while generating model ?
- leading underscore might be used for example for some meta information of object and thus might be wanted to be kept

2.
yaml: `Resources.property`
generated: `Resources._property`
- generated naming is clearly wrong, it should be `property_` instead of `_property`
- this violates PEPs as leading underscore is considered as 'private' attr

3.
yaml: `Resources.attributeWithMoreThanOneCapitalLETTERs`
generated: `Resources.attribute_with_more_than_one_capital_lette_rs`
- generator does not care about multiple capital letters which should be kept as they are, instead it's
trying to sanitize to snake_case at a point where it should not

##### openapi-generator version

4.0.1

##### OpenAPI declaration file content or url

https://gist.github.com/adrpp/0df250c32075415c1a73fbba4adb0155

##### Command line used for generation

java -jar openapi-generator-cli.jar generate -g python -o github-test/ -i openapi1.yaml

##### Steps to reproduce

just have a look on generated model `Resources` (github-test/openapi_client/models/resources.py) 

##### Related issues/PRs

no one found describing provided info

##### Suggest a fix

1. `Resources._attributePrefixedWithUnderscore` -> leading underscore should not be removed
2.  `Resources._property` -> leading underscore should not be used, instead append underscore at the end of attribute name
3.  `Resources.attribute_with_more_than_one_capital_lette_rs` -> multiple upper case letters in row should not be snake_cased

PR #3548, opened 2019-08-03
Problem: leading _ in name is removed by PythonClientCodegen

dkliban:
### PR checklist

### Description of the PR
Solution: move leading _ to the end of the name
    
Double leading underscores in Python attribute names cause them to have their names
mangled as described in PEP 8[0]. That's why they were being removed completely before
this patch.
    
Due to the implementation details of the python models that are generated, leaving a
single leading underscore causes a conflict between setters of variables name that differ
only by a leading underscore. e.g.: 'name' and '_name'.
    
This patch changes how leading underscores in variables and parameters are handled. All
leading underscores are moved to the end of the name.
    
[0] https://www.python.org/dev/peps/pep-0008/

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
@@ -451,7 +451,7 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
         // remove dollar sign
         name = name.replaceAll("$", "");
 
-        // if it's all uppper case, convert to lower case
+        // if it's all upper case, convert to lower case
         if (name.matches("^[A-Z_]*$")) {
             name = name.toLowerCase(Locale.ROOT);
         }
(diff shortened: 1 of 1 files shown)


jimschubert:
cc python technical community for review:
@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess

The target spec for this work defined `type` as well as `_type`, and this caused issues. I provided some feedback in Slack, but I've only used Python minimally. The approach in this PR looks good to me, but does anyone else have any concerns with merging it?

dkliban:
This issue resolves some of the concerns raised here https://github.com/OpenAPITools/openapi-generator/issues/3168

wing328:
@dkliban Please add a few tests to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientCodegenTest.java

A good reference is https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptnode/TypeScriptNodeClientCodegenTest.java#L20 (you can simply copy all these tests to start with)

wing328:
Looks like this is a breaking change without fallback?

dkliban:
> Looks like this is a breaking change without fallback?

Yes, this is a breaking change. However, I don't think the existing behavior was correct. 

How can we make this change less painful for existing users?

dkliban:
I decided to change the parameter names in my OpenAPI schema and no longer have this problem. I am closing this PR.

PR #3552, opened 2019-08-04
[Rust Server] Frunk  - LabelledGeneric - support

richardwhiuk:
### PR checklist

  - @frol, @farcaller, @bjgill 

### Description of the PR

Frunk `LabelledGeneric` support

Derive the LabelledGeneric trait on Swagger models.  This allows conversion between structurally similar types by using https://docs.rs/frunk/0.2.2/frunk/#transmogrifying.

This makes it easier to use the models - especially when writing API proxies which convert between two different APIs.

This is an opt in feature on the generated crate. As part of adding the feature, the generated Cargo files have been tidied up.

Authored by @bossmc


bin/rust-server-petstore.sh
```diff
@@ -38,4 +38,8 @@ for spec_path in modules/openapi-generator/src/test/resources/*/rust-server/* ;
 		 $@"
 
   java $JAVA_OPTS -jar $executable $args
+
+  if [ $? -ne 0 ]; then
+    exit $?
+  fi
```
bjgill: Good catch. Any reason to not just `set -e`, though?

modules/openapi-generator/src/main/resources/rust-server/README.mustache
```diff
@@ -73,43 +73,23 @@ The examples can be run in HTTPS mode by passing in the flag `--https`, for exam
 cargo run --example server -- --https
 ```
 
-This will use the keys/certificates from the examples directory. Note that the server chain is signed with
-`CN=localhost`.
-
-
-## Writing a server
```
bjgill: What was wrong with this section?

bin/rust-server-petstore.sh
richardwhiuk: Opinion is divided on whether `set -e` is a good idea - see http://mywiki.wooledge.org/BashFAQ/105 for some opinions.

modules/openapi-generator/src/main/resources/rust-server/README.mustache
richardwhiuk: It's largely wrong and unhelpful.

Just copying and pasting the dependencies from the API crate is unhelpful at best.

Using a workspace isn't really a sensible idea - you should really just treat the generated crate as a dependency in it's own right - you shouldn't need to modify it.

The examples are demonstrations of how the crate can be used - they aren't templates for people to slavishly copy.

PR #3568, opened 2019-08-06
[OCaml] Added optional params support in API operations

cgensoul:
### Added optional params support in API operations

This PR translates optional OpenAPI operation parameters into optional OCaml function parameters. To allow more flexible use of generated API functions, all function parameters are now named (even the required ones).

Default values given for API operation parameters are also turned into default values for corresponding parameters of OCaml generated API function. Default values are also automatically generated for parameters that are of an enum type with a uniq enum case (useless parameter since it's value is constant but encountered...).

A bug has been fixed in path param replacement.
Reformatted api-impl template to make it easier to read.


wing328:
> Reformatted api-impl template to make it easier to read.

Shall we use something like [ocamlformat](https://github.com/ocaml-ppx/ocamlformat) to auto-format the auto-generated code?

We can add post-processing file hook to allow any code formatter to format the code based on user preference.

cgensoul:
Yes we could, I agree that users might want to format it so that it follows
some internal company and/or project guidelines.

It's not in my priorities at the moment though. The goal of the commit you
referenced was simply to make it easier for me to modify the code of the
template (I was unaware of the jMustache rule by which begin/end mustache
tags alone on their line don't generate line breaks in the output, and I
took advantage of that rule as soon as I learned about it).

As far as I am concerned, both the template and the generated code are
quite easy to read now.

cgensoul:
Thanks for the merge ! I think optional parameters was quite an important
feature to have, it makes the generated API functions much nicer to use and
document them at the same time (since function arguments are all named now).

wing328:
> It's not in my priorities at the moment though.

I've filed #3583 to allow file post processing in the OCaml client generator

PR #3586, opened 2019-08-08
typescript-fetch: allow configuration of headers and credentials

### LINKED ISSUE #3402: [BUG] Typescript-fetch : The `options` parameter is gone. No way to pass options.
#### Bug Report Checklist

##### Description

The new version of the typescript-fetch generator seems to have completely dropped the second parameters for each call that provided a way to pass options. For example, specific header or configuration.

As you can see in version 3.3.4 tag, there is always a second parameter to each calls.
https://github.com/OpenAPITools/openapi-generator/blob/v3.3.4/modules/openapi-generator/src/main/resources/typescript-fetch/api.mustache#L93

In the current master, there is no such call and I don't see any way to pass this information.

For example, in our application, we were always passing this:

```
{
    credentials: "include",
    headers: {
        'Cache-Control': 'no-cache, no-store'
    }
}
```

Is there another way to achieve the same thing using the new generator or should I create a new PR to add back this functionality?

4.X.X (master)

Any spec

Any typescript-fetch generation

##### Steps to reproduce

Generate a typescript-fetch client and notice the issue

##### Related issues/PRs

Didn't find anything related to this

##### Suggest a fix

Add the functionality back (have a second parameter everywhere)

JFCote:
### PR checklist

### Description of the PR

I've change my mind on how to fix that "bug".  

In the bug itself, I was suggesting to do like in version 3.X.X and have an "option" parameter in each call but I realized that was no a very good design.

Once I understood how the code was refactored in 4.X.X, the only thing missing was to be able to provide additional `headers` with each Fetch query in the `Configuration` object. Also, the `credentials` option was missing, so I added it. Very simple PR with no breaking change and no risk :)

fixes #3402

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
@@ -52,10 +52,13 @@ export class BaseAPI {
         const body = (context.body instanceof FormData || isBlob(context.body))
 	    ? context.body
 	    : JSON.stringify(context.body);
+
+        const headers = Object.assign(context.headers, this.configuration.headers);
         const init = {
             method: context.method,
-            headers: context.headers,
+            headers: headers,
             body,
+            credentials: this.configuration.credentials
         };
         return { url, init };
     }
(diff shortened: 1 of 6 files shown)


modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -52,10 +52,13 @@ export class BaseAPI {
         const body = (context.body instanceof FormData || isBlob(context.body))
 	    ? context.body
 	    : JSON.stringify(context.body);
+
+        const headers = Object.assign(context.headers, this.configuration.headers);
```
macjohnny: allow overriding configuration headers with custom headers:
```suggestion
        const headers = Object.assign(this.configuration.headers, context.headers);
```

JFCote: My lack of knowledge of typescript! Thanks for the suggestion. Doing the change right now + generating the samples. Should be up in a couple of minutes.

PR #3616, opened 2019-08-12
Fix composed properties missing from allVars

### LINKED ISSUE #3613: [BUG] Composed properties are missing from allVars
##### Description

When a model includes properties by composition (`allOf`, without a discriminator), these properties end up in the `vars` template variable but they are missing from the `allVars` variable.

I would expect `allVars` to include, at minimum, everything in `vars`. Some generators only iterate `allVars` when producing a model class, so they will be missing these composed properties.

4.1.0

https://gist.github.com/glennschmidt/d5e87fe518ac5afc7c39032ae1df479e

`openapi-generator generate -i openapi.json -g swift4 -o apigen`

##### Steps to reproduce

1. Execute the above CLI with the schema from the linked gist
2. Look at the generated output `apigen/OpenAPIClient/Classes/OpenAPIs/Models/EventCollection.swift`

It will contain
```swift
public struct EventCollection: Codable {


    public init() {
    }


}
```

ie. The EventCollection model has no properties, but it should have
* page
* page_count
* page_size
* total_items
* events

##### Suggest a fix

In `DefaultCodegen.fromModel()`, when adding composed properties to `properties`, also add them to `allProperties` (i will submit a PR with this).

glennschmidt:
### Description of the PR

Fixes #3613 

When a model includes properties by composition (`allOf`, without a discriminator), these properties end up in the `vars` template variable but they are missing from the `allVars` variable.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1822,6 +1822,7 @@ public class DefaultCodegen implements CodegenConfig {
                         } else {
                             // composition
                             addProperties(properties, required, refSchema);
+                            addProperties(allProperties, allRequired, refSchema);
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


macjohnny:

wing328:
@glennschmidt thanks for the PR. Can you please update the test case in https://github.com/OpenAPITools/openapi-generator/blob/d7b390f328a597438edd6a72a9f1ff495ca7080a/modules/openapi-generator/src/test/java/org/openapitools/codegen/ruby/RubyClientCodegenTest.java#L526 to test allVars of `superman` as well?

wing328:
@glennschmidt thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

Issue #3485, opened 2019-07-29
[BUG][Golang][Client] reserved word collision for required API parameter


grokify:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When a required parameter with the name `url` is included, `url` is used as the query parameter name in the function parameter list which conflicts with the `net/url` package that is loaded for query and form requests. This results in the following compile time error:

`../../engagedigital/api_webhooks.go:68:28: url.Values undefined (type string has no field or method Values)`

The actual output is:

```go
func (a *WebhooksApiService) CreateWebhook(ctx context.Context, accessToken string, label string, url string, registeredEvents []string, localVarOptionals *CreateWebhookOpts) (Webhook, *http.Response, error) {
	var (
		localVarHttpMethod   = http.MethodPost
		localVarPostBody     interface{}
		localVarFormFileName string
		localVarFileName     string
		localVarFileBytes    []byte
		localVarReturnValue  Webhook
	)

	// create path and map variables
	localVarPath := a.client.cfg.BasePath + "/webhooks"

	localVarHeaderParams := make(map[string]string)
	localVarQueryParams := url.Values{}
	localVarFormParams := url.Values{}
```

The expected output is no collision between parameter name `url` and the `net/url` library when instantiating `url.Values{}`.

##### openapi-generator version

4.1.0-SNAPSHOT

##### OpenAPI declaration file content or url

https://raw.githubusercontent.com/grokify/go-ringcentral-engage/8e520ea8355190939c1cc8990ba562fe487095e7/codegen/partial-specs_v3.0.0/openapi-spec_webhooks.json

```json
          {
            "name":"url",
            "in":"query",
            "required":true,
            "description":"The url of a webhook. This is used to determine the endpoint of your webhook, where",
            "schema":{
              "type":"string"
            }
          }
```

##### Command line used for generation

[...]

PR #3634, opened 2019-08-13
Import Go package with leading underscore

wing328:
### PR checklist

### Description of the PR

To fix https://github.com/OpenAPITools/openapi-generator/issues/3485

Tested locally with `url`, `http`, `ioutil` and `context` as required parameters.

### DIFF
--- a/modules/openapi-generator/src/main/resources/go/api.mustache
+++ b/modules/openapi-generator/src/main/resources/go/api.mustache
@@ -3,17 +3,17 @@ package {{packageName}}
 
 {{#operations}}
 import (
-	"context"
-	"io/ioutil"
-	"net/http"
-	"net/url"
+	_context "context"
+	_ioutil "io/ioutil"
+	_nethttp "net/http"
+	_neturl "net/url"
 {{#imports}}	"{{import}}"
 {{/imports}}
 )
 
 // Linger please
 var (
-	_ context.Context
+	_ _context.Context
 )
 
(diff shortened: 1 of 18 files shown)


bkabrda:
IIUC this tries to address the problem by using more "obscured" version of the words - so technically if someone used a query parameter `_context`, this would still fail. I guess there's a low enough probability of getting in that situation that I'm 👍  (as I don't see any better solution other than obscuring the import names even more, e.g. `_openapiGeneratorImportContext`, but I guess that's too far fetched).

So overall, 👍 from me.

wing328:
> so technically if someone used a query parameter _context, this would still fail.

Very good point. My understanding is that the query (and other forms of) parameter names are sanitized to conform to Go variable naming convention so the leading understood will be removed.

I did a test locally with `_context` and confirmed no variable naming conflict but I'm happy to take another look if you can repeat the issue.

bkabrda:
I can't reproduce such an issue, so you're right, it's ok like this 👍 

Could you please add the same modification to the go-experimental client to keep them in sync as much as possible? Thanks!

wing328:
> Could you please add the same modification to the go-experimental client to keep them in sync as much as possible? Thanks!

Sure I can do it later this week.

wing328:
I will also add more test cases in a separate PR.

wing328:
FYI. Added more test cases via https://github.com/OpenAPITools/openapi-generator/pull/3640

PR #3434, opened 2019-07-24
Java 11 native HTTP client library

bbdouglas:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. If contributing template-only or documentation-only changes which will change sample output, be sure to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) first.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.1.x`, `5.0.x`. Default: `master`.
- [x] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This is a set of mustache templates to create a new Java library called "native" that is backed by the standard Java 11 [HTTPClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html) class.

The majority of files are only lightly edited from the default Java templates. The model classes, for example, are entirely untouched. The main two edits are:

- APIClient.mustache
- api.mustache

It follows the pattern used by Jersey, where the APIClient acts as a mechanism for configuring the clients along with a small repository of utility functions. The individual api classes are created from the settings and helper classes held by the APIClient, and are immutable and thread-safe. Object serialization/deserialization is handled by Jackson, as is common in the other libraries.

We at Yelp currently use these templates to generate clientlibs used in our production microservices architecture. This library lacks some of the features seen in other openapi-generator libraries, but represents a fully functional subset for our use case. Additional features could be added on as needed. Here is a summary of the limitations of the current clientlibs:

- No Validation
- No Auth
- No HTML Forms
- Only native Java dates
- Only application/json mimetype

resolves #3433

PR #3637, opened 2019-08-13
Use javax.annotation.processing.Generated for Java 11 native client

### LINKED ISSUE #3636: [BUG][Java][native] Wrong `Generated` annotation package for Java 9+
#### Bug Report Checklist

- [x] What's the version of OpenAPI Generator used? 4.1.0

##### Description

Most clients use the following `Generated` annotation for the autogenerated code. This is correct for Java 8 and below, see [Java 8 Generated javadocs](https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html).

    @javax.annotation.Generated(value = "org.openapitools...", date = "...")

Starting with Java 9, however, it is now under a new package name (added the "processing" sub-package). See the [Java 9 Generated javadocs](https://docs.oracle.com/javase/9/docs/api/javax/annotation/processing/Generated.html) for more info.

    @javax.annotation.processing.Generated(value = "org.openapitools...", date = "...")

The current native HttpClient Generated annotation template still uses the old package name, and so if you set `hideGenerationTimestamp=false`, the code doesn't compile, since it runs in Java 11.

4.1.0

Any

Fails with all command lines that don't include `hideGenerationTimestamp=true`.

##### Steps to reproduce

See above

##### Related issues/PRs

N/A

##### Suggest a fix

I have a PR coming up to fix this for the native HTTP clientlib. It creates the correct annotation for Java 9+ inside a custom `generatedAnnotation.mustache` template inside the native resources directory.

Once more people start using Java 9+, a better alternative might be to have a `Java 9` mode that would be incorporated into the shared `generatedAnnotation.mustache` file. But for now, I think this quick fix is easier.

The samples don't have to be re-created since they don't have the Generated annotation.

bbdouglas:
Before Java 9, the `Generated` annotation was part of JEE and in the
`javax.annotation` package. Starting with Java 9, it is now part
of the standard SDK, under the new `javax.annotation.processing`
package.

This change creates a custom Generated annotation template for the
native clientlib, since it is guaranteed to be run in Java 11+.

### PR checklist

### Description of the PR

Adds a new custom `generatedAnnotation.mustache` template for the native HttpClient library with the correct `Generated` package name for Java 9+.

Fixes #3636

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/native/generatedAnnotation.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/native/generatedAnnotation.mustache
@@ -1,0 +1,1 @@
+{{^hideGenerationTimestamp}}@javax.annotation.processing.Generated(value = "{{generatorClass}}", date = "{{generatedDate}}"){{/hideGenerationTimestamp}}


bbdouglas:
CC @cbornet @jeff9finger @karismann @Zomzog Java technical committee

 @macjohnny as reviewer of #3434 that introduced this new generator.

This is meant to address bug #3636

macjohnny:
@bbdouglas where/how is this file included/referenced?

jmini:
@macjohnny:

This is with usage of `{{>generatedAnnotation}}` in the templates.

---

But you are right, when this construct is used, inside `Java/pojo.mustache`:
https://github.com/OpenAPITools/openapi-generator/blob/94d4e0cfac90964b82962f74d9f320171a22e278/modules/openapi-generator/src/main/resources/Java/pojo.mustache#L5

Then I think that `Java/generatedAnnotation.mustache` is used and not `Java/libraries/native/generatedAnnotation.mustache` even if you are generating the native client.

So I guess this PR does not really work.

bbdouglas:
@macjohnny @jmini As you rightly pointed out, this is referenced in `pojo.mustache`. I do believe that this PR is working, as I have tested it locally. Unfortunately, the checked-in samples have the timestamp omitted, so it can't be checked there.

I remember looking this up before, and there is a hierarchy of file locations that are searched when looking for a mustache file. First it tries the library directory, and if it doesn't exist there, it backs off to the general Java directory. That's how I can have my own custom `pom.mustache` file, for example. I can't remember where I saw the code that does the mustache file locating, though, so let me see if I can find it.

jmini:
> I can't remember where I saw the code that does the mustache file locating, though, so let me see if I can find it.

The place I know is:
https://github.com/OpenAPITools/openapi-generator/blob/2d7785ac6309cf38dc99894e642e699f6286e45a/modules/openapi-generator/src/main/java/org/openapitools/codegen/AbstractGenerator.java#L153

But I thought this is not used when a template calls an other template.

wing328:
@bbdouglas thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

evanlott:
Does this issue still exist? I have tried plugin versions 4.2.2 and 4.1.3 and keep getting javax.annotations.Generated in my gen'd API. Was the solution not carried over to versions beyond 4.1.1 or something? Perhaps I'm not using the solution correctly, like maybe missing a config option?

dmitry-weirdo:
Still not clear how to use it. 

When I add
                            <configOptions>
                                <hideGenerationTimestamp>true</hideGenerationTimestamp>
                            </configOptions>

, no `@Generated` annotation is generated at all.

Can you please explain how to set up the custom annotation?

Issue #3518, opened 2019-07-31
[BUG][Go] enum name collision


kraney:
##### Description

In the generated Go client, specific values of enums are declared at package scope using only the value name. This causes name collisions between enums with overlapping value names.

##### openapi-generator version

4.1.0-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
components:
  schemas:
    monitor_action_type:
      enum:
      - TRAP
      - SYSLOG
      type: string
    csr_use:
      enum:
      - TLS
      - SYSLOG
      type: string
```
This results in two package-level variables both named "SYSLOG" causing build failure.

##### Command line used for generation

```
docker run --name oas --rm openapitools/openapi-generator-cli:latest generate -g go -i openapi.yaml
```

##### Steps to reproduce

Generate code for a schema with the two models above, and try to build it.

##### Suggest a fix

Possibly prefix the enum value name with the enum type name, e.g. `CsrUse_SYSLOG`.

PR #3644, opened 2019-08-14
Guarantee a unique name for the constants that represent enum values

kraney:
Proposed fix for https://github.com/OpenAPITools/openapi-generator/issues/3518

This change adds a prefix to the consts that are defined to represent enum values. This prevents name collisions between different enums having the same value within.


samples/client/petstore/go/go-petstore-withXml/model_outer_enum.go
```diff
@@ -13,7 +13,7 @@ type OuterEnum string
 
 // List of OuterEnum
 const (
-	PLACED OuterEnum = "placed"
-	APPROVED OuterEnum = "approved"
-	DELIVERED OuterEnum = "delivered"
+	OuterEnum_PLACED OuterEnum = "placed"
```
wing328: Would it be better if it's all upper case, e.g. OUTERENUM_PLACED?

Ref: https://stackoverflow.com/a/22688926/677735

We also need an option so that users can stick with the current behavior instead (e.g. PLACED instead of OUTERENUM_PLACED)

kraney: Well, the Go guides are pretty adamant about using mixed case, but I elected not to change the part that was already there that didn't. But I can uppercase the typename for consistency.

Unfortunately I doubt I'll dig into the code enough to learn how to add an option flag.

wing328: I think I can help on that.

wing328:
I'll add the option in another PR.

wing328:
Filed https://github.com/OpenAPITools/openapi-generator/pull/3675

wing328:
#3675 has been merged into master. To enable it, please do `--additional-properties enumClassPrefix=true`

wing328:
@kraney thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

PR #3654, opened 2019-08-15
Manage enum properly on Dart Jaguar generator

jaumard:
Enum where not generate properly at all, now should be ok


samples/client/petstore/dart-jaguar/openapi/lib/model/order.dart
```diff
@@ -5,22 +5,24 @@ part 'order.jser.dart';
...
+  @Alias('quantity', isNullable: false,  )
   final int quantity;
   
-  @Alias('shipDate', isNullable: false)
+  @Alias('shipDate', isNullable: false,  )
   final DateTime shipDate;
    /* Order Status */
-  @Alias('status', isNullable: false)
+  @Alias('status', isNullable: false,
+             processor:  const StringFieldProcessor(),
```
jaumard: @wing328 is there a way to know if the enum is from and #ref or inline ? Because this will not work :/ 
#ref enum are generating a new file and class, but inline are not for some reason and does this instead

jaumard: I have ignored inline enum and handle them as simple string, I don't think their is another easy way to do that for now, at least the compiled code works like this

Issue #318, opened 2018-06-14
[rust-server] lacks support for `additionalProperties`


bjgill:
##### Description

See https://github.com/swagger-api/swagger-codegen/issues/8089#issuecomment-389117445 - our main bar to using the same swagger spec for sample generation as the other generators is that `rust-server` does not yet support `additionalProperties`.

Admittedly, this seems like a non-trivial problem. `rust-server` is intended to provide a distributed strongly-typed system. As soon as you add `additionalProperties`, however, we lose the ability to be able to verify the contents of said additional properties. 

A naive solution might be:

```
struct Foo {
    bar: Option<i32>,
    additionalProperties: HashMap<String, json::Object>,
};
```

However, this will fail as soon as we introduce non-json media types (e.g. XML, which `rust-server` supports). As such, I'm not sure what a complete solution would look like...

bjgill:
This may be at least partially a regression. The following spec worked in swagger-codegen:

```yaml
swagger: '2.0'
info:
  description: "This spec is mainly for testing Petstore server and contains fake endpoints, models. Please do not use this for any other purpose. Special characters: \" \\"
  version: 1.0.0
  title: OpenAPI Petstore
  license:
    name: Apache-2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
host: petstore.swagger.io:80
basePath: /v2
schemes:
  - http
paths:
  /pet:
    post:
      summary: Add a new pet to the store
      description: ''
      operationId: addPet
      consumes:
        - application/json
      produces:
        - application/json
      parameters:
        - in: body
          name: body
          description: Pet object that needs to be added to the store
          required: true
          schema:
            $ref: '#/definitions/additionalPropertiesObject'
      responses:
        '405':
          description: Invalid input
definitions:
  additionalPropertiesObject:
    description: An additionalPropertiesObject
    type: object
    additionalProperties:
      type: string
    example: "foo"
```

bjgill:
OK - what's happened is that we never supported additional properties. openapi-generator introduced a regression that caused a compile failure. I've put in a workaround in #671 to restore things to how they previously were - i.e. that the `additionalProperties` directive is simply ignored.

richardwhiuk:
We still don't entirely support additional properties - e.g. we don't support objects which mix additional properties and non-additional properties.

PR #3666, opened 2019-08-17
[Rust Server] Support types with additional properties

richardwhiuk:
### PR checklist

   - @frol, @farcaller, @bjgill 

### Description of the PR

This adds support for types which only contain additional properties, which is a common use case. It doesn't handle the thornier case of additional properties within a type with defined properties, which is allowed by the OpenAPI spec, but is more dubious API design in practice (as the API is mixing schema-ed and un-schema-ed data).

Note, this will also only fix the case where the type is defined (e.g. string) - if it's untyped, it remains unsupported.

This thus partially fixes https://github.com/OpenAPITools/openapi-generator/issues/318

An example of something this will support is already defined in the sample API:

```
  additionalPropertiesObject:
    description: An additionalPropertiesObject
    type: object
    additionalProperties:
      type: string
```

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
@@ -909,6 +909,12 @@ public class RustServerCodegen extends DefaultCodegen implements CodegenConfig {
             additionalProperties.put("usesXmlNamespaces", true);
         }
 
+        Schema additionalProperties = ModelUtils.getAdditionalProperties(model);
+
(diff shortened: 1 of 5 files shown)


richardwhiuk:
Oh, this is technically a breaking change, because objects of this form with previously constructible (although with no properties in them) as `AdditionalPropertiesObject {}` and the equivalent creation is `AdditionalPropertiesObject(HashMap::new())` but that's basically unavoidable.

wing328:
@richardwhiuk  thanks for the PR, which has been included in v4.2.0 release: https://twitter.com/oas_generator/status/1189824932345069569

Issue #3191, opened 2019-06-21
[BUG] [Java] Escaped bean property name should be resumed


yue9944882:
##### Description

xref: https://github.com/kubernetes-client/java/pull/595

in the kubernetes openapi spec, there's some escaped property name in the schema:

```json
    "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps": {
      "description": "JSONSchemaProps is a JSON-Schema following Specification Draft 4 (http://json-schema.org/).",
      "properties": {
        "$ref": {
          "type": "string"
        },
        "$schema": {
          "type": "string"
        }
....
```
the generator doesn't seem to properly handle the name and my java output is failing the compiliation:

```java
@ApiModel(description = "JSONSchemaProps is a JSON-Schema following Specification Draft 4 (http://json-schema.org/).")
@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2019-06-20T12:55:05.588Z[Etc/UTC]")
public class V1beta1JSONSchemaProps {
  public static final String SERIALIZED_NAME_$_REF = "$ref";
  @SerializedName(SERIALIZED_NAME_$_REF)
  private String $ref;

  public static final String SERIALIZED_NAME_$_SCHEMA = "$schema";
  @SerializedName(SERIALIZED_NAME_$_SCHEMA)
  private String $schema;
```

##### openapi-generator version

v4.0.0

##### Suggest a fix

it seems that in the swagger-codegen we won't have the issue.

[https://github.com/kubernetes-client/java/blob/dc82549dde1eeea7593203dec788af0fe33685a7/kubernetes/src/main/java/io/kubernetes/client/models/V1beta1JSONSchemaProps.java#L38-L42](https://github.com/kubernetes-client/java/blob/dc82549dde1eeea7593203dec788af0fe33685a7/kubernetes/src/main/java/io/kubernetes/client/models/V1beta1JSONSchemaProps.java#L38-L42)

yue9944882:
anything i can do to move this forward? it's critical for migrating kubernetes client from swagger-gen to openapi-gen..

davidxia:
@yue9944882 do you have repro steps for this?

yue9944882:
yes, try patch https://github.com/kubernetes-client/gen/pull/117 locally and you'll hit this.

wing328:
[...]

PR #3676, opened 2019-08-19
Remove dollar sign from Java const name

wing328:
### PR checklist

### Description of the PR

Remove dollar sign from Java constant name as only letters and underscore are allowed: https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

@bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -17,6 +17,7 @@
 
 package org.openapitools.codegen.languages;
 
+import com.google.common.base.CaseFormat;
 import com.google.common.base.Strings;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
@@ -984,6 +985,9 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
             model.imports.add("ApiModelProperty");
             model.imports.add("ApiModel");
         }
+
+        // add java constancct name
+        property.vendorExtensions.put("x-java-const-name", CaseFormat.UPPER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, property.nameInCamelCase.replaceAll("\\$", "")));
     }
 
     @Override
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -986,7 +986,7 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
             model.imports.add("ApiModel");
         }
 
-        // add java constancct name
+        // store java constant property name in vendor extension, remove dollar sign from the constant name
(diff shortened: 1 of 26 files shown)


jmini:
Does it works if you have 2 properties `$_type` and `type` in the same object?
=> This is a requirement we have.

---

`$` is allowed in constant names in Java.
Maybe not recommended by the linked google guidelines, but this is not the Java specification (that is hosted somewhere on the Oracle website)

wing328:
Fair enough. I'll make an option, e.g. `googleStyleGuide`

yue9944882:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8

> The "Java letters" include uppercase and lowercase ASCII Latin letters A-Z (\u0041-\u005a), and a-z (\u0061-\u007a), and, for historical reasons, the ASCII underscore (_, or \u005f) and dollar sign ($, or \u0024). The $ character should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

@jmini my understanding oracle's offical site, it's not recommended to use $ in the field name intentionally. the supporting for $ is mainly for (a) generated source (b) compatibility for legacy system. considering that this repo is indeed a code-generator so the it's still complying w/ (a).. i will have a sync-up w/ kubernetes maintainers to discuss about the very issue.

yue9944882:
the issue is that it's breaking backward compatibility when we migrate from `swagger-codegen` to `openapi-codegen`, the former has generated non-dollar fields already. let's hold this for more inputs? wdyt @wing328

jmini:
Swagger-Codegen was not generating those constants, but it supports the valid OpenAPI Spec (its v2 counter-part of course) containing both `type` and `$_type` in the same JSON object.

I am not saying am against any change here.

I am just pointing out that this change, the way it is implemented produces a regression.

In some other places, special chars like this one are replaced with its written form.
Code that does that:

https://github.com/OpenAPITools/openapi-generator/blob/4575b3074ab6f2628837ca84ed5c287f2afaceee/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1123-L1125

Can't we use this?

jmini:
> Fair enough. I'll make an option, e.g. `googleStyleGuide`

I do not think that we need an other option here. Can you consider:

1. replacing `$` with `DOLLAR` in the constant name?
2. or find a way to ensure that the generated constants are unique (with a suffix like `_1`, `_2` .. in case of collision)?

Both would be acceptable for me, I prefer the first solution.

wing328:
> Swagger-Codegen was not generating those constants

Very good point on this. We need to better understand the backward compatibility first: https://github.com/OpenAPITools/openapi-generator/issues/3191

I'll continue the discussion in https://github.com/OpenAPITools/openapi-generator/issues/3191 by asking for more information (e.g. error message) to start with

wing328:
> I do not think that we need an other option here. Can you consider:

Personally of course I don't want another option but if someone is referencing these constant in their code, it will break if they upgrade openapi-generator to a newer version with the fix.

Let's put this PR on hold for the time being.

jmini:
I have added the test case discussed in https://github.com/OpenAPITools/openapi-generator/pull/3676#pullrequestreview-276936294 as PR https://github.com/OpenAPITools/openapi-generator/pull/3774. 

A change as proposed in this PR would produce a test failure.

PR #3683, opened 2019-08-19
Two tiny fixes for Java Vertx client

FlorianBruckner:
# PR checklist

CC: @bbdouglas, @sreeshas, @jfiala, @lukoyanov, @cbornet, @jeff9finger, @karismann @Zomzog

### Description of the PR

1.) ApiClient already defines and configures an objectMapper to not fail on unknown properties, but it is not used when parsing the response. The fix uses the pre-configured object mapper instead of the vertx default one

2.) When an operation has no response (or just ones without content), the accept array passed to ApiClient is emtpy. This makes the null check in ApiClient useless, as it still tries to set a null Accept header, which is refused with an NPE. Amend the check with .length > 0 to catch this case.

This fix is for 4.1, which I believe is the version currently living on master, so filing the PR against this branch.

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/vertx/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/vertx/ApiClient.mustache
@@ -441,7 +441,7 @@ public class ApiClient {
 
         updateParamsForAuth(authNames, queryParams, headerParams);
 
-        if (accepts != null) {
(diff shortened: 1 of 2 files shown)


jmini:
In my opinion the AppVeyor build failure is not related to this change

wing328:
@FlorianBruckner thanks for the PR, which has been included in the v4.1.1 release: https://twitter.com/oas_generator/status/1165944867391860737

PR #3688, opened 2019-08-19
Improve cpp qt5 client

MartinDelille:
### PR checklist

### Description of the PR

This PR add two improvements to the qt5 client:

- I removed the host parameter which since it's buggy: The base path add to be set using `setHost` which is not appropriate. Ideally we could compose the request path with the scheme/host/port and base path (without host). Unfortunately the scheme is not accessible from the mustache template (I managed to access it locally by modifying [this function](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCppCodegen.java#L297-L308).
- I simplified timeout management using `QTimer::singleShot()`
- I used simpler `#pragma once` directive


modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
```diff
@@ -313,11 +305,8 @@ void {{prefix}}HttpRequestWorker::execute({{prefix}}HttpRequestInput *input) {
         buffer->setParent(reply);
 #endif
     }
-    if(timeout > 0){
-        timer->setSingleShot(true);
-        timer->setInterval(timeout);
-        connect(timer, &QTimer::timeout, this, [=](){ on_manager_timeout(reply); });
-        timer->start();
+    if(_timeOut > 0){
+        QTimer::singleShot(_timeOut, [&](){ on_manager_timeout(reply); });
```
etherealjoy: What happens if the timer elapsed at the same time response is received? Would there be a crash because the object is still referenced in the Qt event Loop? Won't this happen since the QNetworkReply is async?

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-header.mustache
```diff
@@ -1,6 +1,5 @@
 {{>licenseInfo}}
-#ifndef {{prefix}}_{{classname}}_H
-#define {{prefix}}_{{classname}}_H
+#pragma once
```
etherealjoy: I wonder if the `#pragma once` causes same behaviour in all compilers, when same files with different namespaces in different locations are included. The guard is useful to prevent the compiler from not including it again, because the namespace is different and the C++ linker has to link to the client object from the same namespace.
e.g. when having multiple clients, which is fairly common. One project we are doing has 10+ clients

etherealjoy:
I need to check the host, because the cpp uses it and I wonder how it compiles 
We are using the `host`, e.g. like `http://someurl.com/`

then set the `basePath` at runtime to 
`/api/v1.0`
`/api/v2.0`
`/api/v3.0`

And using a single client to talk to different versions of a service depending on what service is available on the field.

etherealjoy:
> I removed the host parameter which since it's buggy: The base path add to be set using setHost which is not appropriate. Ideally we could compose the request path with the scheme/host/port and base path (without host). Unfortunately the scheme is not accessible from the mustache template (I managed to access it locally by modifying this function.

The issue is in the `ModelUtils` not in the `AbstractCppCodegen`

wing328:
> I removed the host parameter which since it's buggy:

Can you please elaborate on the issue with the host parameter?

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-header.mustache
MartinDelille: Ok

MartinDelille:
Replaced by #4429 and #4430

modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
MartinDelille: In #4430, I replaced [&] by [=]

PR #3695, opened 2019-08-19
 [typescript-fetch] Fix #1998: namespace reqs.

### LINKED ISSUE #1998: [BUG] [typescript-fetch] Request Parameter Interfaces Conflict with remove-operation-id-prefix
#### Bug Report Checklist

##### Description

When an API spec contains two similarly named operations that differ only in prefix:

```
      operationId: tag_new_create
...
      operationId: unit_new_create
```

then generating the `typescript-fetch` client with the `--remove-operation-id-prefix` option will result in an error like:

```
apis/index.ts:2:1 - error TS2308: Module './TagApi' has already exported a member named 'NewCreateRequest'. Consider explicitly re-exporting to resolve the ambiguity.
```

This is in 4.0.0beta. The same error did not occur with the old 3.x typescript-fetch, though this new implementation is generally much better.

https://gist.github.com/bradenmacdonald/061c37b42af22832a14574e4aeaa5c59

```
java -jar openapi-generator-cli-4.0.0-beta.jar generate \
    --input-spec api-spec.yaml \
    --generator-name typescript-fetch \
    --output api_client \
    --config openapi-generator-config.json \
    --remove-operation-id-prefix
```

##### Steps to reproduce

Simply generate the typescript-fetch client using the sample gist and the command above, then run `cd api_client; npm install; npm run build`

##### Related issues/PRs

I couldn't find any existing reports of this issue.

##### Suggest a fix

Ideally just never omit the prefixes when generating the request parameter interfaces, so that the interface is still called e.g. `TagNewCreateRequest` (instead of `interface NewCreateRequest`) regardless of the `--remove-operation-id-prefix` option (the main reason I use that option is for nicer method names; the names of the parameter interfaces aren't often used directly so it's fine if they're more verbose).

Alternately, put each module's interfaces into a namespace for that module, so that one must use e.g. `TagApi.NewCreateRequest` to get the interface after importing the overall API.

jgiles:
### PR checklist

### Description of the PR

Fix #1998 by namespacing the generated request object types used in the typescript-fetch client codegen.


macjohnny:
cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10) @akehir (2019/07)

macjohnny:
cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10) @akehir (2019/07)

macjohnny:
@jgiles thanks for your PR!

samples/client/petstore/typescript-fetch/builds/default/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-4.1.1-SNAPSHOT
\ No newline at end of file
+unset
```
jgiles: I'm not sure why the version is clobbered when I run the sample code re-generation. I can manually undo if necessary.

macjohnny: are you sure you ran `mvn clean package` before generating the samples? are you based on the latest master?

jgiles:
@macjohnny thanks for the speedy review!

Per your suggestion, I have introduced a new CLI option for namespacing the request objects. I patterned the added option after existing options in the codegen class, let me know if adjustments are needed.

Currently, the option defaults to false so that this would be mergeable into master for inclusion in the next patch release (I'm interested in making this available as soon as possible, since the fixed issue blocks my team's migration from the old swagger-codegen). However, it might be worth considering:

- Making the option default to true in the next minor release branch
- Removing the option in the 5.0 major release branch

so that the fix is present by default.

macjohnny:
@jgiles thanks for the changes! I will review them soon

samples/client/petstore/typescript-fetch/builds/default/.openapi-generator/VERSION
jgiles: I'm based on 024f46814cd7071841c08a4363fa73c7d26a2aac (master circa 2019-08-18).

If I run
```
mvn clean package
./bin/typescript-fetch-petstore-all.sh
```

on my current branch (where the `.openapi-generator/VERSION` files already have "unset"), no sample files are changed. If I run it on master, the `.openapi-generator/VERSION` files are overwritten to "unset".

FWIW, I (but no one else on my team) had this problem when running a brew-installed jar of of the old swagger-codegen, which is part of why we switched to running it in Docker... there could just be something weird in my local env.

macjohnny:
@jgiles I think we can merge this and have this released in `4.1.1`
if you want you could copy and adapt the existing builds and/or tests to have your feature compiled/tested by CI and prevent regressions, see e.g.

https://github.com/OpenAPITools/openapi-generator/blob/4575b3074ab6f2628837ca84ed5c287f2afaceee/pom.xml#L834-L845
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-fetch/builds/with-npm-version/pom.xml

macjohnny:
@jgiles by the way, would you like to add your company to https://github.com/OpenAPITools/openapi-generator#4---companiesprojects-using-openapi-generator ?

jgiles:
I will add the Maven builds, and will check about adding company name :-)

(I'm dropping offline for a little bit though, so it will be this evening probably)

jgiles:
@macjohnny added the maven builds, fixed the version files.

samples/client/petstore/typescript-fetch/builds/default/.openapi-generator/VERSION
jgiles: fixed manually

jgiles:
There's a build failure in the Maven stuff. The error looks to be for a different typescript-fetch sample, but I think the new one might also have problems. There's a bit of messiness in the setup of the sample tests, I might need some help getting that bit right.

macjohnny:
@jgiles the travis-error seems unrelated, but you need to register your `pom.xml` here:

https://github.com/OpenAPITools/openapi-generator/blob/4575b3074ab6f2628837ca84ed5c287f2afaceee/pom.xml#L834-L845

moreover, the `package.json` is missing to successfully run `npm install` or `npm run build` in https://github.com/OpenAPITools/openapi-generator/tree/dddc201f05e7167651c012c55742bae0179b5b30/samples/client/petstore/typescript-fetch/builds/namespace-parameter-interfaces, so I would recommend setting the `npmVersion` in https://github.com/OpenAPITools/openapi-generator/blob/dddc201f05e7167651c012c55742bae0179b5b30/bin/typescript-fetch-petstore-multiple-parameters.json similar to https://github.com/OpenAPITools/openapi-generator/blob/dddc201f05e7167651c012c55742bae0179b5b30/bin/typescript-fetch-petstore-with-npm-version.json

If we don't manage to have the CI tests run, I would recommend to merge it without them, so this can be part of the next release, and add the CI tests in a separate PR.

jgiles:
Thanks @macjohnny for the pointers!

I am registering the `pom.xml` here: https://github.com/OpenAPITools/openapi-generator/pull/3695/files#diff-600376dffeb79835ede4a0b285078036R846

I'm now doing the codegen for CI in the same way as the with-npm-version variant, which does produce the needed `package.json`. Hopefully that will resolve the build issues.

jgiles:
I would also be OK merging without the CI. What needs to happen in order to merge?

jgiles:
@macjohnny also added Paxos to the list of companies using openapi-generator!

macjohnny:
> I would also be OK merging without the CI. What needs to happen in order to merge?

I will merge it tomorrow if the ci build succeeds

macjohnny:
Can you please merge the most recent master into your branch? The unrelated typescript-axios build failure should be disabled

jgiles:
Merged in latest master

jgiles:
Using the "Requests" name invariably led to name conflicts between the Requests namespaces when there were multiple services (tags). This issue did reproduce in petstore - I'm not sure why CI didn't catch the compilation error.

I've updated to use the service basename as a prefix (like "PetRequests")

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -264,11 +264,13 @@ private void addOperationObjectResponseInformation(Map<String, Object> operation
...
     private void addOperationNamespaceParameterInterfacesInformation(Map<String, Object> operations) {
-        if (getNamespaceParameterInterfaces()) {
+        Map<String, Object> _operations = (Map<String, Object>) operations.get("operations");
+        List<CodegenOperation> operationList = (List<CodegenOperation>) _operations.get("operation");
+        if (!operationList.isEmpty() && getNamespaceParameterInterfaces()) {
             operations.put("namespaceParameterInterfaces", true);
             operations.put("paramIfaceIndent", "    ");
             operations.put("paramIfaceSuffix", "");
-            operations.put("paramIfaceNsPrefix", "Requests.");
+            operations.put("paramIfaceNsPrefix", operationList.get(0).baseName + "Requests.");
```
macjohnny: maybe this has to be sanitized, but we will see later

PR #3703, opened 2019-08-20
[java][client][resteasy][jersey2][google-api-client][okhttp-json] several fixes to make sent requests more accurate

### LINKED ISSUE #3276: [BUG][JAVA][Client] Regression: Unconditionally sends empty object request body
#### Bug Report Checklist

##### Description

The java `okhttp-gson` client (and likely other Java clients) always send an empty object in the request body for `DELETE` (and likely `POST` and `PUT`) operations which do not define `requestBody`.  Since [`DELETE` request payload has no defined semantics](https://tools.ietf.org/html/rfc7231#section-4.3.5) this is likely to cause compatibility issues (and causes 4XX responses from the API I am working with).

The issue first appeared in 0fb1ffa88b (#98) and still occurs on `master` (0cb921251d).

```yaml
openapi: '3.0.2'
info:
  title: delete example
  version: '1.0.0'
servers:
- url: http://example.com/api
components:
  schemas:
    Error:
      type: object
      properties:
        message:
          type: string
paths:
  /:
    delete:
      responses:
        '204':
          description: Deleted
        default:
          description: Error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
```
(Note: At least one schema must be defined for the generated code to compile, so I added `Error` for this purpose.)

`openapi-generator-cli.jar generate -i openapi.yaml -g java -o openapi-generator-test-client`

##### Steps to reproduce

1.  Generate the code using the YAML spec and command from above, replacing `example.com/api` with a test server.
2.  Create `src/main/java/TestApp.java` with content
```java
import org.openapitools.client.*;
import org.openapitools.client.api.*;

public class TestApp {
    public static void main(String[] args) throws ApiException {
        ApiClient apiClient = new ApiClient();
        DefaultApi defaultApi = new DefaultApi(apiClient);
        defaultApi.rootDelete();
    }
}
```
3.  Run `gradle -DmainClass=TestApp execute` and observe traffic on the test server.

The client generated from code before 0fb1ffa88b will send:

    DELETE /api/ HTTP/1.1
    Accept: application/json
    Content-Type: application/json
    User-Agent: OpenAPI-Generator/1.0.0/java
    Host: example.com
    Connection: Keep-Alive
    Accept-Encoding: gzip

The client generated with 0fb1ffa88b or later will send:
    
    DELETE /api/ HTTP/1.1
    Accept: application/json
    User-Agent: OpenAPI-Generator/1.0.0/java
    Content-Type: application/json; charset=utf-8
    Content-Length: 2
    Host: example.com
    Connection: Keep-Alive
    Accept-Encoding: gzip
    
    {}

##### Suggest a fix

This was fixed for resttemplate by #605 but not for `jersey2`, `okhttp-gson`, or `resteasy`.  Reverting 0fb1ffa88b for the other clients would fix this, but presumably reintroduce #98 (although I'm not sure why Jackson would be called to serialize `null`, presumably we could just stop doing that).

Thoughts from the participants on #98 and #605?  @bmordue, @jmini, @rubms

### LINKED ISSUE #3719: [BUG][Java][client] Not possible to send payloads on DELETE requests
#### Bug Report Checklist

##### Description

I've just hit an issue with Jersey client which doesn't make it possible at all to send payloads with DELETE requests. While the HTTP spec says that semantics of DELETE payload is undefined, there is nothing that strictly prohibits sending payloads with DELETE requests (and I'm actually generating a client for API that needs payloads on DELETE :/ ).

I think this should be allowed in the following way:

* If the body passed in is null, no payload will be sent (not even empty object).
* If the body passed in is not null, payload will be sent in the same way as it is for e.g. POST and other methods.

Does that sound like a good compromise?

4.1.0 (but seems to be problem in all versions AFAICS)

##### Steps to reproduce

Generate a jersey2 client from OpenAPI spec that defines a json payload on DELETE request, try to run invoke the method for deletion with a non-null payload - no payload is sent.

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/issues/3276

##### Suggest a fix

I'm using the following patch to make this work. If it looks ok, I can submit it as a PR (probably after https://github.com/OpenAPITools/openapi-generator/pull/3703 is merged, since it touches same parts of code and my patch might need to change depending on the final changes in #3703).

```diff
diff --git a/Java/libraries/jersey2/ApiClient.mustache b/Java/libraries/jersey2/ApiClient.mustache
index ab6a4a1353..579dc1ae1c 100644
--- a/Java/libraries/jersey2/ApiClient.mustache
+++ b/Java/libraries/jersey2/ApiClient.mustache
@@ -716,7 +716,12 @@ public class ApiClient {
       } else if ("PUT".equals(method)) {
         response = invocationBuilder.put(entity);
       } else if ("DELETE".equals(method)) {
-        response = invocationBuilder.delete();
+        // if body is null or Object, no entity was passed in => call delete without payload
+        if (body == null || body.getClass() == java.lang.Object.class) {
+          response = invocationBuilder.delete();
+        } else {
+          response = invocationBuilder.method("DELETE", entity);
+        }
       } else if ("PATCH".equals(method)) {
         response = invocationBuilder.method("PATCH", entity);
       } else if ("HEAD".equals(method)) {
@@ -776,6 +781,8 @@ public class ApiClient {
     clientConfig.register(json);
     clientConfig.register(JacksonFeature.class);
     clientConfig.property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true);
+    // turn off compliance validation to be able to send payloads with DELETE calls
+    clientConfig.property(ClientProperties.SUPPRESS_HTTP_COMPLIANCE_VALIDATION, true);
     if (debugging) {
 {{^supportJava6}}
       clientConfig.register(new LoggingFeature(java.util.logging.Logger.getLogger(LoggingFeature.DEFAULT_LOGGER_NAME), java.util.logging.Level.INFO, LoggingFeature.Verbosity.PAYLOAD_ANY, 1024*50 /* Log payloads up to 50K */));
@@ -786,6 +793,9 @@ public class ApiClient {
 {{#supportJava6}}
       clientConfig.register(new LoggingFilter(java.util.logging.Logger.getLogger(LoggingFilter.class.getName()), true));
 {{/supportJava6}}
+    } else {
+      // suppress warnings for payloads with DELETE calls
+      java.util.logging.Logger.getLogger("org.glassfish.jersey").setLevel(java.util.logging.Level.SEVERE);
     }
     performAdditionalClientConfiguration(clientConfig);
     return ClientBuilder.newClient(clientConfig);
```

### LINKED ISSUE #3720: [java-client][resteasy] use real PATCH instead of POST with "X-HTTP-Method-Override"
For PATCH call, the `resteasy` client should use a real "PATCH" method instead of using "POST" with "X-HTTP-Method-Override"

Example spec:

```yaml
openapi: 3.0.1
info:
  title: OpenAPI Test
  description: Test Spec for a "Simple Client"
  version: 1.0.0
  license:
    name: Apache-2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
servers:
  - url: 'http://localhost:8090/'
paths:
  /pingPatch:
    patch:
      operationId: pingPatch
      responses:
        '201':
          description: OK
```

Code that must be changed:

https://github.com/OpenAPITools/openapi-generator/blob/10a310ef69d49742405b4b520b1d53b20b16e845/modules/openapi-generator/src/main/resources/Java/libraries/resteasy/ApiClient.mustache#L665-L666

jmini:
### PR checklist

### Description of the PR

Fixes #3276
Fixes #3719
Fixes #3720


jmini:
I found one issue with Jersey2 and sending a `null` entity:

```
java.lang.IllegalStateException: Entity must not be null for http method PUT.
    at org.glassfish.jersey.client.JerseyInvocation.validateHttpMethodAndEntity(JerseyInvocation.java:163)
    at org.glassfish.jersey.client.JerseyInvocation.<init>(JerseyInvocation.java:112)
    at org.glassfish.jersey.client.JerseyInvocation.<init>(JerseyInvocation.java:108)
    at org.glassfish.jersey.client.JerseyInvocation.<init>(JerseyInvocation.java:99)
    at org.glassfish.jersey.client.JerseyInvocation$Builder.method(JerseyInvocation.java:445)
    at org.glassfish.jersey.client.JerseyInvocation$Builder.put(JerseyInvocation.java:334)
```

I think the 2 solutions [mentioned on StackOverflow](https://stackoverflow.com/q/34759569/91497) might be interesting:

* Disable those checks with `SUPPRESS_HTTP_COMPLIANCE_VALIDATION`.
* Send an empty text entity.

Any opinion (can be solved in a follow-up PR)

jmini:
WIP: 

For the jersey2 client, for `null` value, I will stop calling jackson with the `serialize` method.

https://github.com/OpenAPITools/openapi-generator/blob/4575b3074ab6f2628837ca84ed5c287f2afaceee/samples/client/petstore/java/jersey2/src/main/java/org/openapitools/client/ApiClient.java#L691

Instead when body is null `Entity entity` must be set to `Entity.json("")`.

bkabrda:
The jersey2 part seems to be fine, it manages to solve https://github.com/OpenAPITools/openapi-generator/issues/3719

I don't really have too many experience with the rest of the libraries, so it's hard for me to comment them, but 👍 from me on the jersey2 part and the general aim of this PR.

jmini:
Approved by @kevinoid in https://github.com/OpenAPITools/openapi-generator/issues/3276#issuecomment-523449644

Discussed with @wing328 on Slack.

PR #3715, opened 2019-08-21
[java] use [javaparser] library to detect syntax errors

sullis:
### PR checklist

### Description of the PR

### DIFF
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
@@ -1,8 +1,12 @@
 package org.openapitools.codegen;
 
-import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.fail;
 import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.assertFalse;
 
+import com.github.javaparser.ParseProblemException;
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
 import io.swagger.parser.OpenAPIParser;
 import io.swagger.v3.oas.models.Components;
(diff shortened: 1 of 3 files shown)


wing328:
The Travis CI failure is not related to this change.

jmini:
+1

This is an interesting change, allowing checking the validity of the code that would have been generated without having to generate it and to run maven/gradle on it.

It will not catch all cases as the compiler would to it with the compile dependencies on the classpath, but because this is earlier in the process (unit-test).

---

What version of Java syntax is checked?
For example, can we leverage it to check the generated code is Java-6 compliant when the java6 option is used?
Same question with java 11 (for the moment, I do not think that the java-native client uses java-11 construct such as `var` introduced with java 10)

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
```diff
@@ -81,4 +83,19 @@ public static void ensureDoesNotContainsFile(Map<String, String> generatedFiles,
...
+                    assertValidJavaSourceCode(fileContents);
+                }
+            }
+        );
+    }
+
+    public static void assertValidJavaSourceCode(String javaSourceCode) {
+        // note: "StaticJavaParser.parse" throws an exception if the parser
+        // detects a syntax error
+        assertTrue(StaticJavaParser.parse(javaSourceCode).getTypes().size() > 0);
```
jmini: Could the error message be improved when the assertion is failing? 
Maybe if you pass the `fileName` in `validateJavaSourceFiles(..)` you could add this to the error message printed when the condition is `false`. This will help to understand where to look to reproduce the issue.

sullis:
@jmini  ready for review

jmini:
Thank you for the this contribution!

PR #3771, opened 2019-08-27
[core] [regression] set parentName when a single possible parent exists

### LINKED ISSUE #2845: [BUG] Inheritance is broken
#### Bug Report Checklist

##### Description

Basic inheritance, [as described in the OpenAPI spec](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) does not work.

Instead of creating a class / interface hierarchy using inheritance to represent the composition, all properties are flattened into the subclass.

[4.0.0-20190508.072036-627](https://oss.sonatype.org/content/repositories/snapshots/org/openapitools/openapi-generator-cli/4.0.0-SNAPSHOT/openapi-generator-cli-4.0.0-20190508.072036-627.jar) or master

schema.yaml:

```yaml
openapi: "3.0.1"
info:
  version: 1.0.0
  title: Inheritance test
paths:
  /order:
    get:
      summary: Product order details
      operationId: getOrder
      responses:
        '500':
          description: Successful load
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ExtendedErrorModel'
components:
  schemas:
    BasicErrorModel:
      type: object
      required:
        - message
        - code
      properties:
        message:
          type: string
        code:
          type: integer
          minimum: 100
          maximum: 600
    ExtendedErrorModel:
      allOf:
        - $ref: '#/components/schemas/BasicErrorModel'
        - type: object
          required:
            - rootCause
          properties:
            rootCause:
              type: string

```

##### Expected output

The following class/interface hierarchy: to be generated:

- `BasicErrorModel` <- `ExtendedErrorModel`

##### Actual output

Three unrelated classes/interfaces:

 - `BasicErrorModel`
 - `ExtendedErrorModel`
 - `ExtendedErrorModelAllOf`

```
java -jar openapi-generator-cli.jar generate -i ./schema.yaml -g typescript-angular -o src
```

I have also tried with the Java generator, and it creates the same hierarchy, so it is not just a Typescript problem.

##### Steps to reproduce

1. Copy the above yaml into `schema.yaml` locally.
2. Run the command above.
3. Examine the generated code.

##### Suggest a fix

I'm not familiar with the codebase, but it appears that `getParentName` in `ModelUtils.java` only considers a class as a viable parent for inheritance if it has a discriminator, but the above inheritance pattern does not use discriminators.

### LINKED ISSUE #3523: [BUG][JAVA] Generator does not generate inheritance when first element of allOf is not a ref
#### Bug Report Checklist

##### Description

When declaring a schema as a composition of classes using allOf, having the first element of the allOf (which is meant to be the main inherited class extended in my Java subclass) not being a reference to a pre-existing class, it is simply ignored by the generator.

Swagger-editor and openapi-codegen 4.1.0

```yaml
  schemas:
    Example1:
      type: object
      properties:
        Example1Property:
          type: string
    Example2:
      allOf:
        - type: object
          properties:
            Example2Property:
              type: string
        - $ref: '#/components/schemas/Example1'
```

The Swagger-editor generated Example2 class writes:
```java
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen", date = "2019-08-01T11:15:04.683Z[GMT]")
public class Example2 implements  Example1 {
  @SerializedName("Example1Property")
  private String example1Property = null;

  public Example2 example1Property(String example1Property) {
    this.example1Property = example1Property;
    return this;
  }

   /**
   * Get example1Property
   * @return example1Property
  **/
  @Schema(description = "")
  public String getExample1Property() {
    return example1Property;
  }

  public void setExample1Property(String example1Property) {
    this.example1Property = example1Property;
  }

// all the usual methods...
```

_<edit>_
The openapi-codegen generated Example2 class writes:

```java
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen", date = "2019-08-02T09:54:03.694+02:00[Europe/Paris]")public class Example2 {

  @JsonProperty("Example1Property")

  private String example1Property = null;
  public Example2 example1Property(String example1Property) {
    this.example1Property = example1Property;
    return this;
  }

  

  /**
  * Get example1Property
  * @return example1Property
  **/
  @Schema(description = "")
  public String getExample1Property() {
    return example1Property;
  }
  public void setExample1Property(String example1Property) {
    this.example1Property = example1Property;
  }

// all the usual methods...
```
_</edit>_

Expected output would have been (from openapi-codegen):
```java
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen", date = "2019-08-02T09:54:03.694+02:00[Europe/Paris]")public class Example2 {

  @JsonProperty("Example2Property")
  @JsonProperty("Example1Property")

  private String example2Property = null;
  private String example1Property = null;

// ...
```

`
"C:\Program Files\Java\jdk1.8.0_72\bin\java" -jar swagger-codegen-cli.jar generate -i myDefs.yml  -l java  --library resttemplate  -DsupportJava6=true -DdateLibrary=legacy     -o out2/myproject --api-package com.my.project.client.api --model-package com.my.project.client.model -DhideGenerationTimestamp=true -DdisableApiModelAnnotations=true
`

##### Suggest a fix

Having the same behaviour as when the non pre-existing class is set as more than the first occurrence in the allOf and is generated as an interface, that is copying the attributes of the non pre-existing class in the subclass.

mnahkies:
### PR checklist

### Description of the PR
Whilst the spec states that the 'allOf' relationship does not imply a hierarchy:

> While composition offers model extensibility, it does not imply a hierarchy between the models.
> To support polymorphism, the OpenAPI Specification adds the discriminator field.

Unfortunately this does not make sense for many existing use cases, that were supported by older
versions of the generator. Therefore, I've restored the older behavior, specifically
in the case that only a single possible parent schema is present.

I think a more complete solution would generate interfaces for the composed schemas,
and mark the generated class as implementing these.

@wing328 I guess you're probably the best person to review this, given that you overhauled the support for `anyOf`, `allOf`, `oneOf` originally. 

I'd be interested if you had an alternative suggestion for achieving this without requiring the parent schema to know the identity of all the extending schema's - in our real definition collection, we have a set of common models that many different entities extend from in other definition files, and we have also build some common library functions that we use on the generated sub classes that accept objects extending from the parent.

Admittedly, our method of using allOf to express this feels awkward. Ideally I'd like a way to generate a generic `Page<T>` class that could then be used as `Page<Cat>`, `Page<Dog>`, etc but I'm not aware of anyway to do this with the openapi spec / code gen tooling.

Example YAML:
```
openapi: 3.0.0
info:
  title: example
  description: example
  version: 0.0.0
paths:
  /test:
    get:
      tags:
        - Test
      summary: test
      description: test
      operationId: getCats
      responses:
        '200':
          description: successful
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/PagedCats'
components:
  schemas:
    Page:
      properties:
        page:
          type: integer
      required:
        - page
    Cat:
      properties:
        id:
          type: string
          format: uuid
        required:
          - id
    Dog:
      properties:
        id:
          type: string
          format: uuid
        required:
          - id
    PagedCats:
      allOf:
        - $ref: '#/components/schemas/Page'
        - properties:
            items:
              type: array
              items:
                $ref: '#/components/schemas/Cat'
          required:
            - items
    PagedDogs:
      allOf:
        - $ref: '#/components/schemas/Page'
        - properties:
            items:
              type: array
              items:
                $ref: '#/components/schemas/Dog'
          required:
            - items
```

fixes #2845, and fixes #3523

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -896,6 +896,8 @@ public class ModelUtils {
     public static String getParentName(ComposedSchema composedSchema, Map<String, Schema> allSchemas) {
         List<Schema> interfaces = getInterfaces(composedSchema);
 
+        List<String> refedParentNames = new ArrayList<>();
+
         if (interfaces != null && !interfaces.isEmpty()) {
             for (Schema schema : interfaces) {
                 // get the actual schema
@@ -911,11 +913,19 @@ public class ModelUtils {
                     } else {
                         LOGGER.debug("Not a parent since discriminator.propertyName is not set {}", s.get$ref());
                         // not a parent since discriminator.propertyName is not set
+                        refedParentNames.add(parentName);
                     }
                 } else {
                     // not a ref, doing nothing
                 }
             }
+        }
+
+        // parent name only makes sense when there is a single obvious parent
+        if (refedParentNames.size() == 1) {
+            LOGGER.warn("[deprecated] inheritance without use of 'discriminator.propertyName' is deprecated " +
+                    "and will be removed in a future release. Generating model for {}", composedSchema.getName());
+            return refedParentNames.get(0);
         }
 
         return null;
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/java/JavaInheritanceTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/java/JavaInheritanceTest.java
@@ -32,9 +32,10 @@ import org.testng.annotations.Test;
 
 public class JavaInheritanceTest {
 
-    @Test(description = "convert a composed model without discriminator")
+
+    @Test(description = "convert a composed model with parent")
     public void javaInheritanceTest() {
-        final Schema allOfModel = new Schema().name("Base");
+        final Schema parentModel = new Schema().name("Base");
 
         final Schema schema = new ComposedSchema()
                 .addAllOfItem(new Schema().$ref("Base"))
@@ -42,7 +43,7 @@ public class JavaInheritanceTest {
 
         OpenAPI openAPI = TestUtils.createOpenAPI();
         openAPI.setComponents(new Components()
-                .addSchemas(allOfModel.getName(), allOfModel)
+                .addSchemas(parentModel.getName(),parentModel)
                 .addSchemas(schema.getName(), schema)
         );
 
@@ -52,7 +53,7 @@ public class JavaInheritanceTest {
 
         Assert.assertEquals(cm.name, "sample");
         Assert.assertEquals(cm.classname, "Sample");
-        Assert.assertEquals(cm.parent, null);
+        Assert.assertEquals(cm.parent, "Base");
         Assert.assertEquals(cm.imports, Sets.newHashSet("Base"));
     }
 


macjohnny:
@mnahkies thanks a lot for your PR!

macjohnny:

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -909,15 +911,19 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
                         // discriminator.propertyName is used
                         return parentName;
                     } else {
-                        LOGGER.debug("Not a parent since discriminator.propertyName is not set {}", s.get$ref());
-                        // not a parent since discriminator.propertyName is not set
```
wing328: Can we keep these comments so that people know the discriminator.propertyName in the spec is not set and therefore the model is not a "parent"?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -909,15 +911,19 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
...
                     }
                 } else {
                     // not a ref, doing nothing
                 }
             }
         }
 
+        // parent name only makes sense when there is a single obvious parent
+        if (refedParentNames.size() == 1) {
+            return refedParentNames.get(0);
```
wing328: Can we throw a warning about this behavior and advise the user to use "discriminator.propertyName" in OAS v3 instead to set a proper parent model as such behavior will be removed in the future?

Our goal is to move users to adopt OAS v3 instead of staying with OAS v2 (which is not the best specification to work with inheritance/composition)

wing328:
@mnahkies thanks a lot for the PR 👍 

I've left some comments. Please take a look when you've time.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
mnahkies: Sure, I'd still be interested in how you would recommend applying the discriminator approach to a collection of schema's like I describe in the MR description - the parent in this case should not have knowledge of its children, and also I don't particularly want to add an arbitrary field like `type: string` solely for code generation purposes.

I'd be more than happy to refactor our definitions to do things the OAS v3 way over time - in general I've found v3 much better to work with than v2 ever was.

macjohnny: we should raise an issue at the openapi 3.0 specification to allow this type of inheritance without discriminator

wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
wing328: Another way is to use vendor extension to describe this if the OAS 3.0 spec doesn't officially support it at the moment.

(nullable was not supported in OAS v2 and we'd to use x-nullable: true before OAS v3 officially supports it)

macjohnny: this is a great idea, and probably the easiest way to support this behavior in the future.

mnahkies: That sounds like a good plan. @macjohnny would you be comfortable raising such an issue? I don't really have the bandwidth to write a well thought proposal at this time.

Also I've added a warning, and fixed the commit author (thanks for pointing that out). Let me know if you're happy with the wording

macjohnny:
@wing328 can we merge this?

wing328:
Yes, the revised change/message looks good.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
wing328: Btw, if this fallback is breaking some existing use cases (allOf with just one schema), we will need to add an option later to cater to both cases.

smasala:
@wing328 @mnahkies this commit breaks typescript generation for models which only have a single `$ref` item listed

**Example**
```
MyModel:
      description: a model
      allOf:
        - $ref: '#/components/schemas/MyOtherModel'
```

**Generated output**
```
import { SomeImportPropertyType } from './propModel';    // this is correct

/**
 * MyModel
 */
export interface MyModel { 
}
export namespace MyModel {
}

```

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
smasala: @wing328 in regards to your https://github.com/OpenAPITools/openapi-generator/pull/3771#discussion_r318663859 and mine  https://github.com/OpenAPITools/openapi-generator/pull/3771#issuecomment-538341900 I assume that's what is happening in my case.

wing328:
> Unfortunately this does not make sense for many existing use cases, that were supported by older
versions of the generator. Therefore, I've restored the older behavior, specifically
in the case that only a single possible parent schema is present.

Heads-up: we'll remove this "old" behavior in 5.0.x (the upcoming major release)

PR #3780, opened 2019-08-28
[core] do not always cast to ArraySchema

jmini:
### PR checklist

### Description of the PR

Similar to PR #3765 but solved at core Level.

Casting an `ObjectSchema` to `ArraySchema` after a `ModelUtils.isArraySchema(schema)` produces `java.lang.ClassCastException: io.swagger.v3.oas.models.media.ObjectSchema cannot be cast to io.swagger.v3.oas.models.media.ArraySchema` error in some cases.

This is because of the implementation of `ModelUtils.isArraySchema(Schema)` (the second part of the method):

https://github.com/OpenAPITools/openapi-generator/blob/5a54aa5726342e2a469f943883b2ad44c22abbdc/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L361-L370

In the case commented with "// assume it's an array if maxItems, minItems is set" we have an `ObjectSchema`.
When we try to read the `items` value, we need to check if we are allowed to cast or not.


wing328:
@jmini thanks for the PR.

What about removing the following instead?
```java
     // assume it's an array if maxItems, minItems is set 
     if (schema != null && (schema.getMaxItems() != null || schema.getMinItems() != null)) { 
         return true; 
     } 
```

As ModelUtils.isArraySchema(Schema) should be the source of truth to tell if a schema is ArraySchema or not.

jmini:
@wing328: this was also my thought. 
Maybe this was included a long time ago because checking for `instanceOf ArraySchema` was not sufficient. Maybe Swagger-Parser did improve in this area.

---
I will change this PR to remove the second part of the `ModelUtils.isArraySchema(Schema)` method, but I will keep the `getSchemaItems()` in the default codegen (it was a lot of copy-paste).

jmini:
@wing328 `ModelUtils.isArraySchema(Schema)` is modified as discussed.

wing328:
> Maybe this was included a long time ago because checking for instanceOf ArraySchema was not sufficient. 

That's what I recalled

PR #3837, opened 2019-09-05
[C++][Pistache] Add missing setter for arrays

### LINKED ISSUE #3769: [BUG] No setter function generated for arrays
#### Bug Report Checklist

- [ x] Have you provided a full/minimal spec to reproduce the issue?
- [ x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ x] What's the version of OpenAPI Generator used?
- [ x] Have you search for related issues/PRs?
- [ x] What's the actual output vs expected output?

##### Description

No setter function is generated for an array.

4.1.1

```yaml
openapi: 3.0.0
servers:
  - description: SwaggerHub API Auto Mocking
    url: 'https://virtserver.swaggerhub.com/'
info:
  description: API
  version: 1.0.1
  title: API
  termsOfService: 'http://swagger.io/terms/'
tags:
  - name: FuelCosts
paths:
  '/FuelCosts':
    get:
      description: Obtain fuel costs
      tags:
        - FuelCosts
      summary: Get fuel cost
      operationId: GetFuelCostQuery
      responses:
        '200':
          description: Successful operation
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/FuelCost'
      security:
        - ApiKeyAuth: []
components:
  schemas:
    FuelCost:
      type: object
      properties:
        Remarks:
          type: array
          items:
            type: string
```

java -jar openapi-generator-cli-4.1.1.jar generate -i ./Bug.json -g cpp-pistache-server -o generated

##### Steps to reproduce

Generate the C++ code from the spec. In the generated FuelCost.h, there is a getter function for Remarks but no setter:
    /// <summary>
    ///
    /// </summary>
    std::vector<std::string>& getRemarks();
    bool remarksIsSet() const;
    void unsetRemarks();

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/issues/3271

##### Suggest a fix
As a workaround, you can derive a class from FuelCost and add a setter function manually, but this should really be handled by the generator.

muttleyxd:
### PR checklist

### Description of the PR

Fixes #3769 
I would like to refactor this for 5.X releases with #3417


muttleyxd:
Shall we generate setter for required fields?

etherealjoy:
I think we need to do for all.

muttleyxd:
That's how it's done here, so I'm waiting for review :smile:

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-header.mustache
```diff
@@ -35,12 +35,10 @@ public:
     /// <summary>
     /// {{description}}
     /// </summary>
-    {{#isContainer}}{{{dataType}}}& {{getter}}();
-    {{/isContainer}}{{^isContainer}}{{{dataType}}} {{getter}}() const;
-    void {{setter}}({{{dataType}}} const{{^isPrimitiveType}}&{{/isPrimitiveType}} value);
-    {{/isContainer}}{{^required}}bool {{nameInCamelCase}}IsSet() const;
-    void unset{{name}}();
-    {{/required}}
+    {{{dataType}}}{{#isContainer}}&{{/isContainer}} {{getter}}(){{^isContainer}} const{{/isContainer}};
```
etherealjoy: It is still funny if this setter is available only for non required fields.

I think we need to add it.

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
```diff
@@ -29,7 +29,7 @@ void to_json(nlohmann::json& j, const {{classname}}& o)
 {
     j = nlohmann::json();
     {{#vars}}
-    {{#required}}j["{{baseName}}"] = o.m_{{name}};{{/required}}{{^required}}if(o.{{nameInCamelCase}}IsSet())
+    {{#required}}j["{{baseName}}"] = o.m_{{name}};{{/required}}{{^required}}if(o.{{nameInCamelCase}}IsSet(){{#isContainer}} || !o.m_{{name}}.empty(){{/isContainer}})
```
etherealjoy: We can leave here as is.

muttleyxd: > Just a small proposal to your thought process here
> What about set setting the isSet value.
> If the array is empty and the isSet is set then we send anyway. Covers use case of user wanting to send empty array
>
> If the array is not empty we send anyway, covers all exisiting cases. User could have accessed it with the get, or might have explicitly set with the set method.

This covers the second case, ```!array.empty() && !isSet```

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-header.mustache
muttleyxd: With my changes there are setters for both (please check the second commit, where I update the samples, `PhotoUrls` is required and has a setter

etherealjoy: My mistake, only the unset/IsSet is available for non required fields.

etherealjoy: I was thinking you would do same approach for all.
But the setter is there for all for sure.

Issue #3618, opened 2019-08-12
[BUG][typescript-inversify] Type mismatch between service model and HttpClient requirement


siada:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When using typescript in `strict` mode, the code won't compile if your spec has an optional requestBody because the HttpClient defines `body` as `{}` which means 'any except null or undefined'

##### openapi-generator version

4.1.0

##### Steps to reproduce

Generated service method:

```typescript
    public apiAccountLoginPost(loginViewModel?: LoginViewModel, observe?: 'body', headers?: Headers): Promise<LoginViewModelResult>;
    public apiAccountLoginPost(loginViewModel?: LoginViewModel, observe?: 'response', headers?: Headers): Promise<HttpResponse<LoginViewModelResult>>;
    public apiAccountLoginPost(loginViewModel?: LoginViewModel, observe: any = 'body', headers: Headers = {}): Promise<any> {
```

HttpClient:

```typescript
    post(url: string, body: {}|FormData, headers?: Headers): Observable<HttpResponse> {
        return this.performNetworkCall(url, "POST", this.getJsonBody(body), this.addJsonHeaders(headers));
    }
```

Error: 

![image](https://user-images.githubusercontent.com/177663/62855211-810c9c80-bce9-11e9-9b34-5a7c4a477405.png)

##### Related issues/PRs

#3607

##### Suggest a fix

I believe changing the `{}` to `any` in the HttpClient will resolve this, however I don't know enough about typescript to understand the impact this may have

I'm happy to organise the PR for this if you agree with this change

macjohnny:
how about changing it to `body?: Object|FormData` and also adapting 
https://github.com/siada/openapi-generator/blob/b54a5266ac9824306e5904554275d809dcd66057/modules/openapi-generator/src/main/resources/typescript-inversify/HttpClient.mustache#L40
to handle undefined/null values?

macjohnny:
[...]

PR #3849, opened 2019-09-06
typescript-inversify: improve check for required parameters, support multiple media types

### LINKED ISSUE #3838: [REQ] [typescript-inversify] Three small changes and one big?
We have adapted the typescript-inversify template for our own use. Some changes are definitely worth integrating into upstream. As per the contributing guidelines, let us discuss these changes. Pull request should follow.

I think for fixing these small issues, one pull request would be ok. I would also include a fix of #3618 

## Falsy Required Parameters
When a required parameter is falsy, an error is thrown. But `""` and `false` can be valid values. Cf. https://github.com/OpenAPITools/openapi-generator/blob/9cc7bd15f2884b12b373b172fa175063897724ff/modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache#L63-L65

## Multiple Media Types
The Accept header can only contain one media type in the current implementation. Cf. https://github.com/OpenAPITools/openapi-generator/blob/9cc7bd15f2884b12b373b172fa175063897724ff/modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache#L138-L151

## WithInterfaces
The template provides the `withInterfaces` option. When you enable it, `apis.mustache` still generates exports for the implementations, not the interfaces. *Should this be changed?* (Would be a breaking change I guess.)

## Class Based Service Identificators
*This is a major breaking change, so I'm unsure whether it should be integrated.*

Currently the dependency injection provided by the template uses string based service identifiers. This means that type checking for the services has to be setup manually at the point of composition. Especially when interfaces are changing, this can lead to overlooked problems. See [inversify issue 911](https://github.com/inversify/InversifyJS/issues/911) for a discussion.

The solution we came up with, is using abstract classes as service identifiers. This way the composition of the services changes from
```typescript
container.bind<IHttpClient>("IApiHttpClient").to(HttpClient).inSingletonScope();
container.bind<IApiConfiguration>("IApiConfiguration").to(ApiConfiguration).inSingletonScope();
```
to
```typescript
container.bind(IHttpClient).to(HttpClient).inSingletonScope();
container.bind(IApiConfiguration).to(ApiConfiguration).inSingletonScope();
```
and the types will be checked automatically.

After typescript compilation the abstract classes are still there, but because they don’t contain any implementation, they don’t use more resources than the strings.

I could create a separate pull-request for this change, but only if there is a genuine chance that it will be included.

bodograumann:
### PR checklist

- [X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, ~`4.1.x`~, ~`5.0.x`~. Default: `master`.

### Description of the PR

Solves two small issues as described in #3838 (Falsy Required Types and Multiple Media Types) as well as the issue in #3618 (IHttpClient interface).

There is no testing setup for typescript-inversify. These changes are very simple though.

fixes #3838, fixes #3849 
 
### Technical Committee ###


modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
```diff
@@ -60,7 +60,7 @@ export class {{classname}} {
     public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}observe: any = 'body', headers: Headers = {}): {{#usePromise}}Promise{{/usePromise}}{{^usePromise}}Observable{{/usePromise}}<any> {
 {{#allParams}}
 {{#required}}
-        if (!{{paramName}}){
+        if ({{paramName}} == null){
```
macjohnny:
```suggestion
        if ({{paramName}} === null || {{paramName}} === undefined){
```

macjohnny:
@bodograumann thanks for your PR

modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
bodograumann: Is there a specific reason for this change?
The semantics for both versions are identical. Cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#Loose_equality_using

macjohnny: strict equality check is more explicit and easier to read, as it indicates that both cases are considered. I agree this does not change the functionality, but only the readability

bodograumann: Ok, sure, that makes sense.

bodograumann:
You are absolutely right.

PR #3857, opened 2019-09-07
Bugfix/3248

peyerroger:
### PR checklist

### Description of the PR


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -525,31 +525,6 @@ public static boolean isEmailSchema(Schema schema) {
         return false;
     }
 
-    /**
-     * Check to see if the schema is a model with at least one properties
-     *
-     * @param schema potentially containing a '$ref'
-     * @return true if it's a model with at least one properties
-     */
-    public static boolean isModel(Schema schema) {
```
jimschubert: Removing a public helper is a breaking change (these are used by programmatic consumers, and this would therefore need to be deprecated before removal).

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2303,7 +2302,6 @@ public CodegenProperty fromProperty(String name, Schema p) {
             //    property.baseType = getSimpleRef(p.get$ref());
             //}
             // --END of revision
-            property.isModel = ModelUtils.isModel(p);
```
jimschubert: If you're changing the template to not use `isModel`, we should keep this as the change affects all generators. See my resolution in #3855, which sets `isModel` as expected (after the naive attempt done in `setNonArrayMapProperty`.

modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
```diff
@@ -836,21 +836,6 @@ private CodegenProperty codegenPropertyWithXEnumVarName(List<String> values, Lis
         return objs;
     }
 
-    @Test
-    public void objectQueryParamIdentifyAsObject() {
```
jimschubert: NOTE: Removing this test/functionality would be considered a breaking change and would need to be documented as a regression.

PR #3871, opened 2019-09-10
Change Uint8list by List<int> because it cause trouble with last versions of jaguar

jaumard:


wing328:
@jaumard thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #3876, opened 2019-09-11
WIP: Use HttpError class when rejecting promises

### LINKED ISSUE #3872: [REQ] Feature Request Description

In generated typescript code promises are rejected with plain objects instead of error instances.
By not using errors we loose stacktraces and thus easy debugging.

Reject Promises with proper Error instances. This would be done [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-node/api-single.mustache#L225)

I can provide a PR, if you are going to accept it. This would be a breaking change.
I would add a new `HttpError` class with a `statusCode` property. Additionally We can add `body` and `response`. This would be a breaking change, but in my opinion a good one.

I don't think there is any viable alternative
## Additional context
I come from kubernetes-client/javascript#336

aeb-sia:
Fixes #3872

### PR checklist

### Description of the PR


modules/openapi-generator/src/main/resources/typescript-node/api-all.mustache
```diff
@@ -8,5 +8,9 @@ import { {{ classname }} } from './{{ classFilename }}';
 export * from './{{ classFilename }}Interface'
 {{/withInterfaces}}
 {{/apis}}
+import http = require('http');
+export class HttpError {
```
macjohnny: shouldn't this extend the javascript `Error` class to get the stack trace functionality that you are looking for?
```suggestion
export class HttpError extends Error {
```
See https://javascript.info/custom-errors

modules/openapi-generator/src/main/resources/typescript-node/api-all.mustache
```diff
@@ -8,5 +8,9 @@ import { {{ classname }} } from './{{ classFilename }}';
...
+    constructor (public response: http.{{#supportsES6}}IncomingMessage{{/supportsES6}}{{^supportsES6}}ClientResponse{{/supportsES6}}, public body: any, public statusCode?: number) { }
```
macjohnny:
```suggestion
    constructor (public response: http.{{#supportsES6}}IncomingMessage{{/supportsES6}}{{^supportsES6}}ClientResponse{{/supportsES6}}, public body: any, public statusCode?: number) {
      this.name = 'HttpError'
    }
```

macjohnny:
please update the samples with `https://github.com/OpenAPITools/openapi-generator/blob/master/bin/typescript-node-petstore-all.sh`

modules/openapi-generator/src/main/resources/typescript-node/api-all.mustache
aeb-sia: yeah.. 🤦‍♂

modules/openapi-generator/src/main/resources/typescript-node/api-all.mustache
aeb-sia: should I also add a message? Something like `HTTP request failed`?

macjohnny: yes, this would be helpful

samples/client/petstore/typescript-node/default/api/apis.ts
```diff
@@ -4,4 +4,8 @@ export * from './storeApi';
 import { StoreApi } from './storeApi';
 export * from './userApi';
 import { UserApi } from './userApi';
+import http = require('http');
+export class HttpError {
```
macjohnny: please run `mvn clean package` before re-generating the samples

wing328:
@aeb-sia thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #3878, opened 2019-09-12
Fix on PR: #888 RFC: [go] define "any-type" schema as an empty interface in generated go code.

mattiabertorello:
### PR checklist

### Description of the PR
Fix issue point out here: https://github.com/OpenAPITools/openapi-generator/pull/888#discussion_r212615913

@wing328 @ashanbrown @eclipse1985
@antihax (2017/11) @bvwells (2017/12) @grokify (2018/07)


wing328:
@mattiabertorello what about using `--type-mapping object="interface{}"` instead?

if it works, we can add this mapping to the Abstract Go class.

eclipse1985:
> @mattiabertorello what about using `--type-mapping object="interface{}"` instead?
> 
> if it works, we can add this mapping to the Abstract Go class.

@wing328 Actually seems that with this parameter the generations works pretty good even without adding the mapping to the Abstract Go class. Can you confirm?

wing328:
If it works, we can hard code the mapping in the Abstract Go class rather than supplying this everytime via --type-mapping when running the generator.

eclipse1985:
Looks very good to me, but we see that at this [commit](https://github.com/OpenAPITools/openapi-generator/commit/4664c07034fac99cfb6354fae6301d85f4f96810) it was already changed in the past from `interface{}` to `map[string]interface{}`. But if you confirm that is ok on your side we fix it exactly as you said.

wing328:
This is already done: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java#L133

Please give the latest master a try.

Closing this PR for the time being.

PR #3881, opened 2019-09-12
typescript-fetch: fix missing comma in multiple imports

rcarton:
When generating the import list for a super class, the comma separating the objects was missing:
The code generated in `SuperClass.ts` was
```ts
import {
     SubClassAFromJSONTyped <-- missing comma
     SubClassBFromJSONTyped
} from './';
```
It was missing the comma between the imported entries.

I tested by building the latest version off of master and generating the client for my app, then applied the fix and validated that the new version correctly passed the typescript compilation.


wing328:
@rcarton thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #3911, opened 2019-09-18
Added Dart config parameters to fulfill pubspec requirements for publishing to pub

benjaminSchilling33:
### PR checklist

@ircecho @swipesight @jaumard @nickmeinhold
### Description of the PR

Added three configuration values for Dart 2 to be able to publish the generated code to pub without changes:

pubAuthor - contains the name of the author
pubAuthorEmail - contains the email address of the author
pubHomepage - contain the homepage of the project

Changed the name of the output directory from docs to doc as this follows the dart [package layout](https://dart.dev/tools/pub/package-layout)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
@@ -39,16 +39,22 @@ public class DartClientCodegen extends DefaultCodegen implements CodegenConfig {
     public static final String PUB_NAME = "pubName";
     public static final String PUB_VERSION = "pubVersion";
     public static final String PUB_DESCRIPTION = "pubDescription";
+    public static final String PUB_AUTHOR = "pubAuthor";
+    public static final String PUB_AUTHOR_EMAIL = "pubAuthorEmail";
+    public static final String PUB_HOMEPAGE = "pubHomepage";
     public static final String USE_ENUM_EXTENSION = "useEnumExtension";
     public static final String SUPPORT_DART2 = "supportDart2";
     protected boolean browserClient = true;
     protected String pubName = "openapi";
     protected String pubVersion = "1.0.0";
     protected String pubDescription = "OpenAPI API client";
+    protected String pubAuthor = "Author";
+    protected String pubAuthorEmail = "author@homepage";
+    protected String pubHomepage = "homepage";
     protected boolean useEnumExtension = false;
     protected String sourceFolder = "";
-    protected String apiDocPath = "docs" + File.separator;
-    protected String modelDocPath = "docs" + File.separator;
+    protected String apiDocPath = "doc" + File.separator;
+    protected String modelDocPath = "doc" + File.separator;
     protected String apiTestPath = "test" + File.separator;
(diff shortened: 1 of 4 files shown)


wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
```diff
@@ -39,16 +39,22 @@
     public static final String PUB_NAME = "pubName";
     public static final String PUB_VERSION = "pubVersion";
     public static final String PUB_DESCRIPTION = "pubDescription";
+    public static final String PUB_AUTHOR = "pubAuthor";
+    public static final String PUB_AUTHOR_EMAIL = "pubAuthorEmail";
+    public static final String PUB_HOMEPAGE = "pubHomepage";
```
wing328: What about using the following as the default value instead?

```
    public static final String PUB_AUTHOR = "OpenAPI Generator Contributors";
    public static final String PUB_AUTHOR_EMAIL = "team@openapitools.org";
    public static final String PUB_HOMEPAGE = "https://openapi-generator.tech";
```

nickmeinhold:
Added a comment to #3912

nickmeinhold:
The number of CLI options seems to be getting unwieldy to me, the only thing that comes to mind in terms of a solution is having pubspec arguments combined in a json file or similar. That's probably a terrible solution, if it's even possible to mix things together like that, but do any of you know if there is a simple way to combine CLI arguments when they are related?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
benjaminSchilling33: For the author this could be done, because the publisher would have to change this before uploading since the author's email used to authorize further updates of the package

For the homepage I wanted to prevent that the publisher uploads a dart package which links to openapi and thus indicating that the different projects are somehow linked
If that's no problem from your point of view I will change it.

benjaminSchilling33:
@wing328 regarding the commit author it seems like I forgot to change the local git config and then my global one was used, I will fix it.

nickmeinhold:
Thanks @benjaminSchilling33, I was aware of the json file option but was thinking that for people using standard cli arguments, the number was getting unwieldy.  But I think you're right, if the number of cli args you need to use gets large, it becomes time to switch to using a json file.  Thanks

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
nickmeinhold: What do you think about setting all of the PUB_ default values (except PUB_NAME) to null and then not adding anything to the pubspec when the value is null?  I'm wondering if nothing is better than a dummy value that might be confusing or problematic.  What do you both think?

benjaminSchilling33: would be a possibility, everything except PUB_NAME is required only for packages which are published to pub

PR #3918, opened 2019-09-20
Feature/api name suffix

bgong-mdsol:
### PR checklist

### Description of the PR

The problem:
API generation for C#, Java and Ruby creates a class based on the tag and then adds "Api" after it. 
Ideally though we have a class like  "Study" or "Study{{Suffix}}" not "StudyApi".

I introduced a apiNameSuffix parameter in order to add suffixes to the generated api class/file/document names. Added the option "--api-name-suffix" to allow the user specify api name suffix via openapi-generator command line, the value of apiNameSuffix is "Api" if the option "--api-name-suffix" is not provided or empty. 

Sample of command line:
java -jar openapi-generator-cli.jar generate g java --api-name-suffix=client -i petstore.yaml -o out/petstore

This PR works for Java, Ruby, C# and Python 

@jcarres-mdsol
@bkashyapmd
@jfeltesse-mdsol
@prajon84
@ykitamura-mdsol
@cabbott

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -85,6 +85,7 @@ public class DefaultCodegen implements CodegenConfig {
     protected Map<String, String> importMapping = new HashMap<String, String>();
     protected String modelPackage = "", apiPackage = "", fileSuffix;
     protected String modelNamePrefix = "", modelNameSuffix = "";
+    protected String apiNameSuffix = "Api";
     protected String testPackage = "";
     protected Map<String, String> apiTemplateFiles = new HashMap<String, String>();
     protected Map<String, String> modelTemplateFiles = new HashMap<String, String>();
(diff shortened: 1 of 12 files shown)


jfeltesse-mdsol:
At a glance, how about making "Api" as the default suffix for backwards compatibility?
Also, you need to update the README (and other places?) to document that new option.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1693,17 +1706,17 @@ public String toSetter(String name) {
...
      * @param name the name of the Api
-     * @return capitalized Api name ending with "Api"
+     * @return capitalized Api name
      */
     public String toApiName(String name) {
         if (name.length() == 0) {
             return "DefaultApi";
         }
-        return camelize(name) + "Api";
+        return camelize(name + "_" + apiNameSuffix);
```
prajon84: Is not this code is used by Java and C#, then there should be no `_` before `apiNameSuffix` ?

bgong-mdsol: Method will take care all "_" based on languages

bgong-mdsol:
> At a glance, how about making "Api" as the default suffix for backwards compatibility?
> Also, you need to update the README (and other places?) to document that new option.

@jfeltesse-mdsol updated README and making "Api" as the default suffix  https://github.com/OpenAPITools/openapi-generator/pull/3918/commits/eedc03989ac285fc57aa58a94989c5f862f0fd2d

wing328:
Looks good to me. Thanks for the PR. 

The reason why we add "Api" as the default suffix is to avoid naming collisions with model name (e.g. PetApi, Pet (model))

I'll submit another PR to revise the description a little bit so as to make it clearer that the option only applies to Ruby, Python and JaxRS generator.

Also I'll add the new option to maven-plugin as well as gradle-plugin

wing328:
Doc update done via https://github.com/OpenAPITools/openapi-generator/pull/4045

PR #3919, opened 2019-09-20
[typescript][fetch] Fix null typing errors

allejo:
### PR checklist

### Description of the PR

The current `typescript-fetch` template generates some invalid Typescript where the typing for the function is either `undefined` or the model. However, in the other generated code it can also be `null` and the function itself is generating a null check.

Additionally, for each nullable entity in the return statement, checking against `== null` will essentially check for both `undefined` and `null` since in JS, `null == undefined` is true.

**Current**

```typescript
export function ModelToJSON(value?: Model): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return {
        
        'primitive_value': value.primitiveValue,
        'nullable_date': value.nullableDate === undefined ? undefined : value.nullableDate.toISOString(),
        'regular_date': value.regularDate.toISOString(),
    }
}
```

**After PR**

```typescript
export function ModelToJSON(value?: Model | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return {
        
        'primitive_value': value.primitiveValue,
        'nullable_date': value.nullableDate == null ? undefined : value.nullableDate.toISOString(),
        'regular_date': value.regularDate.toISOString(),
    }
}
```

Additionally, the indentation of a generated comment is wrong, so that fix is included in here as well.

~~I don't currently have a Java dev environment setup on my computer but I've tested these template changes through the `-t` flag and would appreciate any early feedback while I get Java set up to check that last box.~~

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -229,14 +229,14 @@ export class {{classname}} extends runtime.BaseAPI {
         {{/returnType}}
     }
 
-   /**
-    {{#notes}}
-    * {{&notes}}
-    {{/notes}}
-    {{#summary}}
-    * {{&summary}}
-    {{/summary}}
-    */
+    /**
+     {{#notes}}
+     * {{&notes}}
+     {{/notes}}
+     {{#summary}}
+     * {{&summary}}
+     {{/summary}}
+     */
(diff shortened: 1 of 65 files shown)


wing328:
@allejo thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #3923, opened 2019-09-20
[java][client] Don't include nullable attributes twice in serialized JSON

bkabrda:
### PR checklist

### Description of the PR

The problem that this PR solves is that for `JsonNullable` fields of POJOs, the field would appear twice, once in snake case and once in camel case (e.g. both `some_field` and `someField` would appear in the JSON). The cause is:

* `JsonNullable` fields have two sets of getters and setters - "simple ones" (e.g. `getSomeField`) and `JsonNullable` ones (e.g. `getSomeField_JsonNullable`).
* For the `JsonNullable` fields, the `getSomeField_JsonNullable` getter is marked as the attribute and explicitly marked to be serialized as `some_field`.
* By default, the Jackson serialization library takes all getters of the object and serializes them. This means that Jackson would also see the non-`JsonNullable` getter and add it to the serialized object.
* This would only demonstrate on fields that have at least two words in their name. For example, `getEnd` would work fine, as there would be `getEnd` and `getEnd_JsonNullable`. The `getEnd_JsonNullable` getter would be marked to be serialized as `end`, which would make Jackson override `getEnd`. But for `getSomeField` and `getSomeField_JsonNullable`, Jackson would serialize `getSomeField` as `someField` and `getSomeField_JsonNullable` as `some_field`.
* Note that this fix also uncovered the need to add a private setter if the nullable field is read-only (Jackson can use that to properly set the value of the readOnly nullable attribute).

This patch fixes the above by explicitly marking the simple getters as ignored during serialization.

There's a slight problem now that this adds an extraneous newline in between the getters' annotations, so I'm trying to get rid of that.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
@@ -676,15 +676,17 @@ public class JavaClientCodegen extends AbstractJavaCodegen
                     var.getVendorExtensions().put("isJacksonOptionalNullable", isOptionalNullable);
                 }
                 if (addImports) {
-                    cm.imports.add("JsonNullable");
-                    Map<String, String> itemJsonNullable = new HashMap<String, String>();
-                    itemJsonNullable.put("import", "org.openapitools.jackson.nullable.JsonNullable");
-                    imports.add(itemJsonNullable);
-
-                    cm.imports.add("NoSuchElementException");
-                    Map<String, String> itemExc = new HashMap<String, String>();
-                    itemExc.put("import", "java.util.NoSuchElementException");
-                    imports.add(itemExc);
+                    Map<String, String> imports2Classnames = new HashMap<String, String>() {{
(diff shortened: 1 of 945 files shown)


wing328:
Looks like the samples (Java client) are not up-to-date: https://circleci.com/gh/OpenAPITools/openapi-generator/8989#tests/containers/2

bkabrda:
@wing328 yeah, there's a problem with an extra added newline that really shouldn't be there - this change shouldn't affect (most of) the generated samples, but it does add an extra newline when [1] does not get rendered (the condition evaluates to false). I couldn't find any way to remove this extra newline. Do you know how I could get rid of it?

[1] https://github.com/OpenAPITools/openapi-generator/pull/3923/files#diff-48702a3c157b1d9b0d527e28c5f99c13R183

bkabrda:
Allright, so I managed to get rid of the extra added whitespace by this change and I even got rid of some whitespace that shouldn't have been there in the first place. Hopefully this is now ready for review.

wing328:
@bkabrda  thanks for the PR, which has been included in v4.2.0 release: https://twitter.com/oas_generator/status/1189824932345069569

Issue #3944, opened 2019-09-24
TypeScript (Node): Better support for binary upload and download


clemens-smartparking:
### Is your feature request related to a problem? Please describe.

When having a form data parameter, e.g.:
```
{
    "name": "file",
    "required": true,
    "in": "formData",
    "description": "",
    "type": "file"
}
```
the generated API simply takes a Buffer object. However, I think it would be nice to support the other request options, especially the one to pass meta data along with the file (also see the example on https://github.com/request/request in the 'multipart/form-data (Multipart Form Uploads)' section).

When downloading a file the generated API seems to ignores the `produces` swagger spec, e.g. when using:
```
"produces": [
    "application/octet-stream"
],
```
the request is done with `application/json` (or am I doing something wrong?)

## Describe the solution you'd like
This is how I currently do the file upload and download:
```
  const part = {
        value: Buffer.from('Test file content'),
        options: {
            filename: 'filename'
        }
  }
  const putResponse = await api.filePut(part as any);

  const getResponse = await api.filesGet(putResponse.body.uri,
    {
      headers: {
        Accept: 'application/octet-stream'
      }
    });
```
There are two hacks here:
1) the `part` object is cast to any because the api expect a Buffer (it works anyway)
2) the `Accept: 'application/octet-stream'` header must be passed in manually

macjohnny:
it would be nice if you could come up with a backwards-compatible implementation for this feature

clemens-smartparking:
I am not aware of backwards-compatibility problems. What do you have in mind?

macjohnny:
Just go ahead, if there is any problem, we will analyze it in the review

macjohnny:
closed with https://github.com/OpenAPITools/openapi-generator/pull/3966 and https://github.com/OpenAPITools/openapi-generator/pull/3967

PR #3967, opened 2019-09-27
typescript-node: form data file

clemens-smartparking:
### PR checklist

@akehir @macjohnny

### Description of the PR

Add options to specify a from data file

Be able to specify file options as described on
https://github.com/request/request in the 'multipart/form-data
(Multipart Form Uploads)' section).

Related to #3944

I am new to the OpenApiGenerator and unsure if I did the java part correctly...

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -17,7 +17,9 @@
 
 package org.openapitools.codegen.languages;
 
+import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.responses.ApiResponse;
 import org.openapitools.codegen.*;
 import org.openapitools.codegen.utils.ModelUtils;
 import org.slf4j.Logger;
@@ -40,8 +42,12 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
     public TypeScriptNodeClientCodegen() {
         super();
 
-        typeMapping.put("file", "Buffer");
+        typeMapping.put("file", "RequestFile");
+        // RequestFile is defined as: `type RequestFile = string | Buffer | ReadStream | RequestDetailedFile;`
         languageSpecificPrimitives.add("Buffer");
+        languageSpecificPrimitives.add("ReadStream");
+        languageSpecificPrimitives.add("RequestDetailedFile");
+        languageSpecificPrimitives.add("RequestFile");
 
         // clear import mapping (from default generator) as TS does not use it
         // at the moment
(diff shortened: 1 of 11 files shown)


samples/client/petstore/typescript-node/default/api/petApi.ts
```diff
@@ -83,17 +83,17 @@ export class PetApi {
     /**
      * 
      * @summary Add a new pet to the store
-     * @param body Pet object that needs to be added to the store
+     * @param pet Pet object that needs to be added to the store
      */
-    public async addPet (body: Pet, options: {headers: {[name: string]: string}} = {headers: {}}) : Promise<{ response: http.ClientResponse; body?: any;  }> {
+    public async addPet (pet: Pet, options: {headers: {[name: string]: string}} = {headers: {}}) : Promise<{ response: http.ClientResponse; body?: any;  }> {
```
macjohnny: please run the [bin/typescript-node-petstore-all.sh](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/typescript-node-petstore-all.sh) script only, not the `openapi3` version

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -70,13 +71,13 @@ public String getHelp() {
 
     @Override
     public boolean isDataTypeFile(final String dataType) {
-        return "Buffer".equals(dataType);
+        return dataType != null && (dataType.equals("Buffer") || dataType.equals("FormDataFile"));
```
macjohnny: I guess `dataType` should be split by `|`, right?

macjohnny:
thanks for the PR

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -70,13 +71,13 @@ public String getHelp() {
...
     }
 
     @Override
     public String getTypeDeclaration(Schema p) {
         if (ModelUtils.isFileSchema(p)) {
-            return "Buffer";
+            return "Buffer|FormDataFile";
```
macjohnny: I think this also affects the api *response* type, so it would be a breaking change, right?

clemens-smartparking:
I moved the type details to the Typescript mustache definitions. I hope I handle the java part correctly now. This made it also possible to add `string` and `ReadStream` types.

What do you mean with `I think this also affects the api response type`? Do you have an example? As far as I see the changes should be transparent to current users (but I also don't know all the swagger details ;) ).

macjohnny:
what is the result if you have something like
```yaml
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Petstore
paths:
  /download:
    get:
      responses:
        '200':
          description: 'An array of pets'
          content:
            application/octet-stream:
              schema:
                type: file
```

clemens-smartparking:
Ok see what you mean. For OpenApi 3 this shouldn't be a problem (https://swagger.io/docs/specification/describing-responses/#response-that-returns-a-file)

However, how to handle it for swagger? Is there a way to find out if a Schema comes from a parameter or a return value? i.e. if it is a parameter then use `RequestFile` otherwise its a return value and `Buffer` should be used. Couldn't find a way to do that. Don't like to return `RequestFile` as a return type as it's always a Buffer. Any suggestions?

macjohnny:
@wing328 any idea?

macjohnny:
@clemens-smartparking maybe you could add the decision logic here:
https://github.com/OpenAPITools/openapi-generator/blob/c136b83bdb5e56ddb297de7f0d5bb9af447a9e54/modules/openapi-generator/src/main/resources/typescript-node/api-single.mustache#L132
and set `Buffer` if the response type is a file.
Or find the method that returns the API Response type and override it there.

clemens-smartparking:
I think I can't do it in the mustache file (no ifs). However, this pointed me in the right direction. I now overwrite the returnType in handleMethodResponse(). This seems to work and I hope it is the right way to do it. However, there is no example in the samples for returning a file. Should, and if yes, how can I add a sample for that?

macjohnny:
@clemens-smartparking can you please merge the latest master into your branch?

clemens-smartparking:
I rebased on the current master

wing328:
@clemens-smartparking thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

Issue #3990, opened 2019-09-30
[REQ] Allow multiple bodyParams for multiple Content Types in RequestBody


jharriman:
### Is your feature request related to a problem? Please describe.

The `requestBody` specification for OpenAPI 3.0 (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#requestBodyObject) allows you to specify multiple MediaType names and schemas for the `Content` section of the `requestBody`. 

Leveraging the ability to use multiple MediaTypes in the `Content` section of the `requestBody` is quite helpful when your MediaTypes are more meaningful than `application/json` e.g.
```
    put:
      operationId: updatePet
      requestBody:
        content:
          application/vnd.petstore.pet.v1+json:
            schema:
              $ref: '#/components/schemas/PetV1'
          application/vnd.petstore.pet.v2+json:
            schema:
              $ref: '#/components/schemas/PetV2'
```

Unfortunately, the current implementation of openapi-generator assumes that you will only ever specify _one_ MediaType in you content definition: 
* In the `fromRequestBody` method of `DefaultCodegen` (https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L4776) only the a single `CodegenParameter` is returned. 
* Looking deeper in the stack, it looks like `ModelUtils.getSchemaFromRequestBody` and `ModelUtils.getSchemaFromContent` (https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L766) assume that there will only ever be one MediaType. At least an warning is printed letting you know that something is wrong, but that's still unhelpful when you actually want to support multiple MediaTypes in the Content section.

## Describe the solution you'd like

It looks like the `DefaultCodegen` class is already setup to support multiple `bodyParams`, but the current implementation only ever adds _one_ `bodyParam` to the list.

Specifically the following changes should be made
* Parse all possible Content `MediaTypes` as `bodyParams` and return a list of `CodegenParameter` for `fromRequestBody`
* Add a new `contentType` property to the `CodegenParameter` object to allow generators to know which `contentType` is associated with each `bodyParam`

## Describe alternatives you've considered

[...]

PR #3991, opened 2019-09-30
Store multiple bodyParams with mediaType appropriately

jharriman:
### PR checklist

New to the project, so while I did my best to answer these questions, I know I did not fully grok the release process and some of the larger concerns that may need to be addressed by my change. 

Not updating any language templates, so I'm guessing I don't need to do this? 

This doesn't seem to be a breaking change (new to this project though, so I could be missing something), so master?

PR targets core, not a specific programming language.

### Description of the PR

Addresses  the issue raised in #3990:
*  Converts `DefaultCodgen.fromRequestBody` to return a `List<CodegenParameter>` instead of a single `CodegenParameter` (since requestBody's often contain _multiple_ Content Types.
* Adds a test for creating multiple `bodyParams` when there are multiple Content Types.
* Updates callers that depend on the singleton functionality to operate as thought there is a singleton (tried to keep the change as simple as possible for now, but these call sites should properly handle receiving multiple `CodegenParameter`s.


spacether:
Thank you for your PR. We do not need it because this information was added in https://github.com/OpenAPITools/openapi-generator/pull/10973

Issue #3992, opened 2019-09-30
[BUG][Kotlin] Kotlin client  variable name


andrewemery:
##### Description

Kotlin client generates invalid variable names for reserved words.

##### openapi-generator version

Working in: 4.1.2
Broken in: 4.1.3-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
responses:
  '200':
    schema:
      properties:
        data: # kotlin reserved word
          type: string
```

Expected:
```kotlin
val `data`: String
```

Actual:
```kotlin
val ``data`` String
```

##### Command line used for generation

```bash
openapi-generator generate -g kotlin
```

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/issues/3899

wing328:
Closed via #3993

PR #3993, opened 2019-09-30
Fixes invalid Kotlin client variable names for reserved words

andrewemery:
The Kotlin Multiplatform client introduced an bug that rendered variable names that were also Kotlin reserved words incorrectly. This change reverts the template to use the previous mechanism for rendering variable names.

https://github.com/OpenAPITools/openapi-generator/issues/3992

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
@@ -19,24 +19,16 @@ package org.openapitools.codegen.languages;
 
 import org.openapitools.codegen.CliOption;
 import org.openapitools.codegen.CodegenConstants;
-import org.openapitools.codegen.CodegenModel;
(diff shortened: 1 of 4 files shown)


andrewemery:
@wing328 Apologies. I introduced a breaking change to Kotlin clients with the addition of Kotlin Multiplatform. 

The original branch (from Swagger Codegen) required a workaround to escape Kotlin identifiers. This workaround is no longer necessary (on OpenAPI) and breaks existing Kotlin clients that have identifers that are Kotlin reserved words.

This change removes the workaround in favour of the existing approach.

wing328:
> Apologies. I introduced a breaking change to Kotlin clients with the addition of Kotlin Multiplatform.

That's ok. Thanks for letting us know and the quite fix.

wing328:
The fix looks good.

wing328:
@andrewemery thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #4023, opened 2019-10-02
[java][client] Fix regression in Java XML serialization. Fixes #4014

### LINKED ISSUE #4014: [BUG][JAVA][XML] Java generator with withXml option produces duplicate XML properties
#### Bug Report Checklist

##### Description

When generating a Java POJO using "withXml" option the `@XmlElement` ends up above `public static final String SERIALIZED_NAME_%` constant, which in turn marks this constant as an XML property. On the other hand, `@XmlAccessorType` is set to `FIELD`, which in turn marks the field, thus producing duplicated definition of the property.

Example:

    /**
     * ValidTakeoverProfilesProfile
     */
    @javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2019-10-01T16:24:48.053+02:00[Europe/Prague]")
    @XmlRootElement(name = "ValidTakeoverProfilesProfile")
    @XmlAccessorType(XmlAccessType.FIELD)
    
    public class ValidTakeoverProfilesProfile {
      @XmlElement(name = "profileName") // this marks the constant below as a property
      public static final String SERIALIZED_NAME_PROFILE_NAME = "profileName";
      @SerializedName(SERIALIZED_NAME_PROFILE_NAME)
      private String profileName; // this becomes a property since XmlAccessorType is FIELD

The outcome is that produced model is not usable for Spring app, as there are multiple properties with a same name inside single class.

4.1.2

After some investigation, version 4.1.0 result produces usable model for my use case.

Complete Maven example is here, with JUnit test: https://github.com/RockyMM/openapi-modelgen-test

    mvn clean test

##### Steps to reproduce

Use the following options to generate the model from YAML

                            <generateApis>false</generateApis>
                            <generateSupportingFiles>true</generateSupportingFiles>
                            <generateModelDocumentation>false</generateModelDocumentation>
                            <skipValidateSpec>true</skipValidateSpec>
                            <supportingFilesToGenerate>pom.xml</supportingFilesToGenerate>
                            <additionalProperties>
                                <additionalProperty>java8=true</additionalProperty>
                                <additionalProperty>dateLibrary=java8</additionalProperty>
                                <additionalProperty>jackson=false</additionalProperty>
                                <additionalProperty>withXml=true</additionalProperty>
                            </additionalProperties>

##### Related issues/PRs

Could not find any.

##### Suggest a fix

In my perspective, in https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/pojo.mustache

both `gson` and `jackson` sections above a property **must** come above `withXml` sections. This could solve the issue.

Either this, or I am doing something very very wrong and missing some obvious thing.

bkabrda:
### PR checklist

### Description of the PR

Fixes regression reported in #4014.

@RockyMM since the current test cases don't cover this, could you try to grab this patch and verify that it fixes the regresssion? Thanks!

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/pojo.mustache
@@ -26,6 +26,9 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
     {{/mostInnerItems}}
     {{/isContainer}}
     {{/isEnum}}
+  {{#gson}}
+  public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
+  {{/gson}}
   {{#withXml}}
   {{#isXmlAttribute}}
   @XmlAttribute(name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
@@ -48,7 +51,6 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   {{/isXmlAttribute}}
   {{/withXml}}
   {{#gson}}
-  public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
   @SerializedName(SERIALIZED_NAME_{{nameInSnakeCase}})
   {{/gson}}
   {{#jackson}}
--- a/modules/openapi-generator/src/main/resources/Java/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/pojo.mustache
@@ -29,6 +29,9 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   {{#gson}}
   public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
   {{/gson}}
+  {{#jackson}}
+  public static final String JSON_PROPERTY_{{nameInSnakeCase}} = "{{baseName}}";
+  {{/jackson}}
   {{#withXml}}
   {{#isXmlAttribute}}
   @XmlAttribute(name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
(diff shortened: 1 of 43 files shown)


RockyMM:
I have two points:

1) this change needs to be propagated to all pojo.mustache in other Java-based generators :( (I was expecting some inheritance, but seems it needs to be copied manually to JAX-RS, Spring and the likes)

2) this segment

    {{#jackson}}
    public static final String JSON_PROPERTY_{{nameInSnakeCase}} = "{{baseName}}";
    {{/jackson}}

also needs to be moved up.

bkabrda:
So I moved the jackson's `JSON_PROPERTY_*` declaration as well, which should solve the second point. I honestly have no idea why this is a problem in all the other pojo templates, as I didn't touch them in my original PR that seemed to have caused this. Perhaps the issues with these were caused by different PRs?

bkabrda:
I think the commit 1ab7b9ceb686d0c600581a08aca6af03bbef81be might have broken the other POJOs. I can try to fix that, but I'd prefer not mixing it with this PR, as the offending commits are different and this PR is for java clients.

RockyMM:
Hm, in commit 1ab7b9ceb686d0c600581a08aca6af03bbef81be I checked the sample of Animal.java in "withXml" resttemplate example and it looked fine. So it might be that this is all the fix that we need.

It's interesting that when you've moved the jacksons `JSON_PROPERTY_*` that the generated samples also changed. I'm just getting a hold in this project so everything is new to me 😄 

This means it is possible to write some automated tests for this so this won't appear again. Like check what is marshalled, and what is unmarsahlled and compare. I could help with this next week, but I think it is not really a show-stopper for the time being.

Update: I just saw #689 - I think it's very good idea. 👍

wing328:
> This means it is possible to write some automated tests for this so this won't appear again. Like check what is marshalled, and what is unmarsahlled and compare. I could help with this next week, but I think it is not really a show-stopper for the time being.

Yes, we welcome additional unit tests to ensure this issue won't happen again.

wing328:
@RockyMM thanks for reviewing the fix.

PR #4024, opened 2019-10-02
[PHP] Correctly format JSON in headers

hinrik:
`ObjectSerializer::toHeaderValue()` in the generated PHP code calls
`toString()` on the values, which formats JSON with the `JSON_PRETTY_PRINT`
option. This will result in a multi-line header which cannot be parsed
since linebreaks aren't allowed by RFC 7230.

In my case I have a header schema called `UpdateUser` which I had hoped
would be serialized as `{"type":"staff","id":123}`.

This PR adds a `toHeaderValue()` method to models which doesn't use `JSON_PRETTY_PRINT`. Works for me.

FYI @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko @renepardon

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -140,7 +140,11 @@ class ObjectSerializer
      */
     public static function toHeaderValue($value)
     {
-       return json_encode(ObjectSerializer::sanitizeForSerialization($value));
+        if (method_exists($value, 'toHeaderValue')) {
+            return $value->toHeaderValue();
+        }
+
+        return self::toString($value);
     }
 
     /**
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -140,7 +140,7 @@ class ObjectSerializer
      */
     public static function toHeaderValue($value)
     {
-        return self::toString($value);
+       return json_encode(ObjectSerializer::sanitizeForSerialization($value));
     }
 
     /**
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -140,7 +140,11 @@ class ObjectSerializer
      */
     public static function toHeaderValue($value)
     {
-        return str_replace(PHP_EOL, '', (string)$value);
+        if (method_exists($value, 'toHeaderValue')) {
+            return $value->toHeaderValue();
+        }
+
+        return self::toString($value);
     }
 
     /**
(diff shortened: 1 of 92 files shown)


hinrik:
It seems there are some cases I didn't cover. Let me fix that up real quick...

modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -140,7 +140,11 @@ class ObjectSerializer
      */
     public static function toHeaderValue($value)
     {
-       return json_encode(ObjectSerializer::sanitizeForSerialization($value));
+        if (method_exists($value, 'toHeaderValue')) {
+            return $value->toHeaderValue();
+        }
+
+        return self::toString($value);
```
ackintosh: How about this instead of adding `toHeaderValue()` to `model_generic.mustache` ? 💡 

```php
// Strip newlines and spaces which comes from JSON_PRETTY_PRINT
return str_replace([PHP_EOL, ' '], '', (string)$value);
```

This will reduce the `if` condition.
>        if (method_exists($value, 'toHeaderValue')) {
>            return $value->toHeaderValue();
>        }

hinrik: There's no need to remove spaces, as they are allowed in a header value (just not at the beginning/end, but JSON wouldn't begin/end with spaces anyway). But yeah, otherwise this approach would work I suppose.

hinrik: I pushed this change. It is simpler, but now there are extra tabs in the JSON since it was pretty-printed, which doesn't really make sense for a single line. We could strip out the tabs as well, I suppose.

ackintosh: Yes, that is why I removed spaces in the sample. 💡

hinrik: Yeah, I see. We should still allow spaces inside strings in the JSON. Would you prefer the previous solution then, with `toHeaderValue()` on the model, or keep it like this?

ackintosh: Oh... sorry for my oversight... I would prefer the previous solution, which makes clean output.

hinrik: :+1:

ackintosh:
The circle CI failure will be fixed via https://github.com/OpenAPITools/openapi-generator/pull/4041 . nothing to do with this PR.

I've checked that unit tests of PHP client has passed locally.

```sh
$ pwd
/Users/akihito1/src/github.com/ackintosh/openapi-generator-1
$ php vendor/bin/phpunit tests
PHPUnit 7.5.16 by Sebastian Bergmann and contributors.

................................................................. 65 / 70 ( 92%)
.....                                                             70 / 70 (100%)

Time: 12.44 seconds, Memory: 10.00 MB

OK (70 tests, 412 assertions)
```

wing328:
@hinrik thanks for the PR, which has been included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #4032, opened 2019-10-02
[Java] Add fluent getters

hinrik:
All Java generators except JavaVertXServer already produce fluent setters,
so this adds fluent getters to those as well.

Motivation: I like having fluent accessors, but the OpenAPI code in my
project is inconsistent with the rest, as I have to resort to `getWhatever()`
accessors instead of `whatever()`.

FYI @bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger
@karismann @Zomzog

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/pojo.mustache
@@ -95,6 +95,10 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   {{/parcelableModel}}
   {{#vars}}
 
+  public {{{datatypeWithEnum}}} {{name}}() {
+    return {{name}};
+  }
+
   {{^isReadOnly}}
   public {{classname}} {{name}}({{{datatypeWithEnum}}} {{name}}) {
     {{#vendorExtensions.isJacksonOptionalNullable}}this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{name}});{{/vendorExtensions.isJacksonOptionalNullable}}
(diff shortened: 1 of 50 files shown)


hinrik:
That single CircleCI test failure (Ruby Petstore test) seems unrelated to my change.

cbornet:
This is redundant with the getXXX getters. I don't understand why you'd want getters without the get prefix. It's a well known convention in Java that comes from the Java Beans contract.

hinrik:
The fluent setters are also redundant with the setXXX setters, but it's still a convention that some use. Example: A project using Lombok might use the [`fluent` option for its accessors](https://projectlombok.org/features/experimental/Accessors).

Would it be better to wrap these in an `additionalProperties` flag that needs to be enabled?

cbornet:
No the regular setters and fluent setters are not the same code.
We keep regular setters just to honor the Java beans contract.

hinrik:
What do you mean they are not the same code? By default two types of setters (`setXXX` and `XXX`) are generated, and only one type of getter (`getXXX`). Why not also a `XXX` getter?

cbornet:
The setXXX is there to follow the Java beans contract and returns `void`. The XXX setter is there to have a fluent way to build the DTO and returns `this`. They are different methods with distinct signature.

hinrik:
Yes, it's true that the fluent setter has a different signature from a non-fluent one, since it returns `this`. Why is that an argument against having a fluent getter as well (which, of course, has the same signature as the normal one, but a more fluent name).

cbornet:
Fluent is not about naming. It's about giving the possibility to chain operations. Which doesn't make sense with getters. Having 2 methods strictly identical except for their names is just duplication and clutter.

wing328:
@hinrik please consider using customized template with the `-t` option if you want to use the suggested change with openapi-generator.

However we won't be able to accept this enhancement into the official repo based on the reasons @cbornet pointed out.

PR #4041, opened 2019-10-03
[Ruby] fix ruby test, update error message

wing328:
### PR checklist

### Description of the PR

- fix ruby test
- update error message

### DIFF
--- a/samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
+++ b/samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
@@ -379,7 +379,7 @@ module Petstore
         # return the array directly as typhoeus will handle it as expected
         param
       else
-        fail "unknown collection format: #{collection_format}"
+        fail "unknown collection format: #{collection_format.inspect}"
       end
     end
   end
--- a/samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
+++ b/samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
@@ -379,7 +379,7 @@ module Petstore
         # return the array directly as typhoeus will handle it as expected
         param
       else
-        fail "unknown collection format: #{collection_format.inspect}"
+        fail "unknown collection format: #{collection_format}"
       end
     end
   end
--- a/samples/openapi3/client/petstore/ruby/spec/api_client_spec.rb
+++ b/samples/openapi3/client/petstore/ruby/spec/api_client_spec.rb
@@ -156,7 +156,7 @@ describe Petstore::ApiClient do
     end
 
     it 'fails for invalid collection format' do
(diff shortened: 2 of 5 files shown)


samples/openapi3/client/petstore/ruby/spec/api_client_spec.rb
```diff
@@ -156,7 +156,7 @@
     end
 
     it 'fails for invalid collection format' do
-      expect(proc { api_client.build_collection_param(param, :INVALID) }).to raise_error(RuntimeError, 'unknown collection format: :INVALID')
+      expect(api_client.build_collection_param(param, :INVALID)).to raise_error(RuntimeError, 'unknown collection format: :INVALID')
```
autopp: It should be a block:
```ruby
expect { api_client.build_collection_param(param, :INVALID) }.to
```

samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
```diff
@@ -379,7 +379,7 @@ def build_collection_param(param, collection_format)
         # return the array directly as typhoeus will handle it as expected
         param
       else
-        fail "unknown collection format: #{collection_format.inspect}"
+        fail "unknown collection format: #{collection_format}"
```
autopp: I think it's better not to fix it here.
For example, nothing is output when `collection_format` is nil.

samples/openapi3/client/petstore/ruby/spec/api_client_spec.rb
wing328: I simply followed lines 155, 151, 147, etc.

Not sure why `proc` was used in the first place but looks like it doesn't work with the newer version of the dependency or ruby version (updated in the CI container)

samples/openapi3/client/petstore/ruby/lib/petstore/api_client.rb
wing328: Sure. Will revert the change as there should be a better fix

samples/openapi3/client/petstore/ruby/spec/api_client_spec.rb
autopp: If wee use `raise_error`, wee need to pass a block to capture the exception.
(And 155, 151, 147, etc don't capture the exception)

> it doesn't work with the newer version of the dependency or ruby version (updated in the CI container)

Hmm... I don't know the details, but the behavior of RSpec may have changed.
In any case, it worked as expected with `expect { api_client.build_collection_param(param, :INVALID) }.to` in my environment. Can you give it a try?

wing328: Yup, you're right that the behavior has changed somehow. Just tried it locally and using block fixes it 👍

josh08h:
Is there a timeframe on this being released and deployed into the `latest` image?

wing328:
@josh08h It's already included in the v4.1.3 release: https://twitter.com/oas_generator/status/1180123829626003456

PR #4055, opened 2019-10-04
[C] Update build-and-test.bash

zhemant:
add check for non autogenerated test file

### PR checklist

### Description of the PR

Due to build of manual test files is commented out in CMakeLists.txt, they are not generated hence build fails.

@Fjolnir-Dvorak

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/README.md.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/README.md.mustache
@@ -64,6 +64,8 @@ once compile, you can run it with ``` ./main ```
 
 Note: You dont need to specify includes for models and include folder seperately as they are path linked. You just have to import the api.h file in your code, the include linking will work.
 
+## Test petstore sample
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/resources/C-libcurl/README.md.mustache
```diff
@@ -64,6 +64,8 @@ once compile, you can run it with ``` ./main ```
 
 Note: You dont need to specify includes for models and include folder seperately as they are path linked. You just have to import the api.h file in your code, the include linking will work.
 
+## Test petstore sample
```
wing328: @zhemant thanks for the RP but I don't think we should add petstore specified instruction in the template as the spec provided by the user may have nothing to do with Petstore.

zhemant: Ok I understand, Then I will modify the README to guide how to compile the user code with the generated code. That was the reason I referenced how it will happen in petstore. I will push another commit

wing328: Closing this as there's no update for a while. Please open a new one when you've time.

PR #4056, opened 2019-10-04
[FEATURE][Haskell] Add Middleware support for the haskell servant generator

Fjolnir-Dvorak:
### PR checklist

- - @jonschoning <-- You are not responsible but the only official haskell maintainer mentioned in the documentation :D

### Description of the PR

I added middleware support to the haskell servant generator and extended the example to handle the additional middleware parameter.


jonschoning:
reviewers/cc: @gibiansky @f-f @mandrean

modules/openapi-generator/src/main/resources/haskell-servant/API.mustache
```diff
@@ -211,13 +212,13 @@ call{{title}} env f = do
 -- | Run the {{title}} server at the provided host and port.
 run{{title}}Server
   :: (MonadIO m, MonadThrow m)
-  => Config -> {{title}}Backend (ExceptT ServerError IO) -> m ()
-run{{title}}Server Config{..} backend = do
+  => Config -> Middleware -> {{title}}Backend (ExceptT ServerError IO) -> m ()
```
f-f: I'd say instead of changing the signature of this function we could add a `run{{title}}ServerWithMiddleware` function, and provide the default implementation of the middleware here

In this way we wouldn't have to document how to get a default middleware, since we'd have that already implemented

Fjolnir-Dvorak: That would be a good Idea to keep backwards-compability, though I would keep the documentation since I think that the call with middleware should be the preferred one since an ID-function does not add any runtime overhead.

Drezil: > In this way we wouldn't have to document how to get a default middleware, since we'd have that already implemented

The 'default'-Middleware is literally just `id` and the example-server would have that as default.
I fear that if one adds wrappers for this it could start to set a precedence for every case where this could be extended.

Another "clean" solution could be to add a datatype:
```haskell
data ServerConfig = ServerConfig { serverMiddleware :: Middleware }
  deriving (....)

defaultServerConfig = ServerConfig id
-- or with data-default
instance Default ServerConfig where
  def = ServerConfig id
```
and people using `defaultServerConfig` would not be affected if we change things in any way in the future. Especially if the use the `Default`-Pattern: http://hackage.haskell.org/package/data-default 
```haskell
let myConfig = defaultServerConfig { serverMiddleware = myMiddleware }
-- or with data-default
let myConfig = def { serverMiddleware = myMiddleware }
```

This way we would never have to update/break the type-signature of the implementation for all future versions.

Issue #4057, opened 2019-10-04
[REQ] [Swift] Swift Package Manager integration


4brunu:
### Is your feature request related to a problem? Please describe.

Xcode 11 was released with Swift Package Manager integration.
At this moment, the Swift 4 generator supports Cocoapods and Carthage integration.
It would be nice to integrate Swift Package Manager also.

## Describe the solution you'd like

Add support for Swift Package Manager

Issue #2721, opened 2019-04-23
New Generator: Swift5 client generator with multiple HTTP library support 


wing328:
##### Description

We would like to start a discussion on a new Swift5 client generator that supports multiple HTTP libraries.

Currently, the Swift (2.x, 3.x, 4x) generators use [Swift Alamofire](https://github.com/Alamofire/Alamofire). For Swift 5, we want to add a new generator that can supports different HTTP libraries via the `library` option similar to what we've done in the Java client generator.

The implementation can be a swappable API client class (the approach we used in Java) or other approach that makes sense in Swift.

This is not to say we'll no longer use Swift Alamofire, which is still a decent choice, so we may want to add a new swift5-generator to support Swift 5.x (only) for time being while we work on the new generator with multiple HTTP library support.

cc @d-date

markfinlabs:
A great idea. There is obviously URLSession in Foundation and now, for server-side Swift, seems a new HTTP client library is in development....
https://forums.swift.org/t/generic-http-client-library-pitch/23341
https://github.com/swift-server/swift-nio-http-client
Our interest would be in this new library for server-side Swift.

mac2000:
While this is in progress is there any known workaround for swift4 generated code to be working in swift5? (e.g. by modifying templates and/or podspec)

martinwiboe:
@mac2000 the swift4 generated code is working fine for me using swift 5. What error are you seeing?

d-date:
Don't worry. Swift4 generator still can be used for existing users (including need objcCompatible option). Swift 5 and Swift 5.1 has so many great features for OAS 3.0. Until Swift 4 generator we have only for OAS 2.0(Swagger).

For various API Clients, this generator only provide interface and protocols with `onlyPalinInterface` option default is `true` . So we can use URLSession in Foundation, Alamofire and any other API Clients.

d-date:
I'd like to ask you how many PromiseKit users exist now? From Swift 5, Result is introduced in Standard Library. If you have no problems, I want to deprecate PromiseKit support since this generator has too many option to keep maintain.

markfinlabs:
We're not using PromiseKit and also are trying to adopt the latest Swift features 😀

lightbow:
AlamoFire was fun while it lasted, but I'd love one fewer dependency. A generator using just URLSession / Foundation would be wonderful.

spprichard:
Is there any progress on this? 

[...]

Issue #230, opened 2018-06-05
[Swift 4] Add ability to pass in a dedicated queue for processing network response


joeboyscout04:
##### Description

The Swift generator uses Alamofire under the hood, which provides this ability to pass in a dedicated background queue for processing responses.  The default is the Main thread. This can cause performance problems when the body of the request is large. 

##### openapi-generator version
3.0.0

##### Suggest a fix/enhancement

It would be simple to alter the Swift generator's AlamofireImplementations `execute(completion: error:)` method to include an optional argument for `queue` (default to `nil`).  The queue could then be passed to `processRequest(completion:)` and then on to the various Alamofire `response()` methods.

Then the user could implement their own `AlamofireRequestBuilder` to override the queue.

PR #4086, opened 2019-10-07
New generator swift5

### LINKED ISSUE #699: [SWIFT4] JSON Decoding fails to parse all valid date-time values
##### Description

The OpenAPI 3.0 specification specifies that string with format: `date-time` should match the definition of `date-time` specified in 
https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14.

The current implementation only handles date-time values that contains micro-seconds, which is per the RFC optional. The Swift 4 version of the library should support all valid date-time values.

3.1.2

https://gist.github.com/grEvenX/089a8c9ecf029560299523ff048cfd5b

docker run --rm -v $(pwd)/generated:/out openapitools/openapi-generator-cli:latest generate -I https://gist.githubusercontent.com/grEvenX/089a8c9ecf029560299523ff048cfd5b/raw/375e5a1759d3a92edab2060d08ecf0650ba8d177/example-openapi-datetime.yaml  -g swift4 -o /out/swift4

##### Steps to reproduce

1. Generate Swift4 client
2. Use the generated client with a GET request containing model with date-time format
3. Have the server/stub return a time in date-time format without microseconds (e.g. `2018-04-30T00:00:00+00:00`)
4. Run the code, the client will crash due to not being able to decode the time properly

Currently there is a `CodableHelper.swift` class generated that is used in decoding JSON.
This class opens up for providing your own DateFormatter used for parsing the dates and also provides a default code if the date formatter is not set. 
The default one provided by openapi-generator should support date-time with and without microseconds, and I think we could by default set the `dataformatter` to an instance of a class that we can provide, e.g. a new class `OpenAPIDateFormatter` which can handle this case.

I've worked around this issue by specifying my own implementation of the DateFormatter with the following code:

```swift

import Foundation

class OpenAPIDateFormatter: DateFormatter {
    override init() {
        super.init()
        
        self.calendar = Calendar(identifier: .iso8601)
        self.locale = Locale(identifier: "en_US_POSIX")
        self.timeZone = TimeZone(secondsFromGMT: 0)
        self.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
    }
    
    required init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }
    
    override func date(from string: String) -> Date? {
        if let validDate = super.date(from: string) {
            return validDate
        }
        
        // first parse-attempt failed, try parsing date without micro-seconds
        let newFormatter = self.copy() as! DateFormatter
        newFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZZ"
        return newFormatter.date(from: string)
    }
}
```

### LINKED ISSUE #4842: [REQ][Swift] make Enum's conform to CaseIterable protocol

It would be nice to make the enums conform to CaseIterable protocol for the Swift compiler to be able to generate a property called `allCases` that provides an array of all cases in the enum.

I will try to create a PR that makes the enum's conform to this protocol.

4brunu:
### PR checklist

@jgavris (2017/07) @ehyche (2017/08) @Edubits (2017/09) @jaz-ah (2017/09) @d-date (2018/03)

### Description of the PR

This PR adds basic support for Swift 5 and it lowers the barrier of any Swift 5 contribution.
It's basically a copy of the Swift 4 generator, but it adds support for Swift Package Manager (fix https://github.com/OpenAPITools/openapi-generator/issues/4057), updates the dependencies version.

***EDIT:***
I have updated this PR to introduce some new features in along side with the new Swift 5 generator.
The new features that the Swift 5 generator has, compared with the Swift 4 generator are:
- Add nullable (or x-nullable in OAS2) support. This will replace the unwrapRequired option.
- Add support for URLSession
- Date formatter add support for ISO8601 with and without milliseconds (fixes #699 and it's a continuation of #1454)
- make enums conform to CaseIterable (fixes #4842)
- Add support for Combine framework 
- cocoapods remove deprecated docset_url and add watchos deployment target
- Add ability to pass in a dedicated queue for processing network response (Fix for https://github.com/OpenAPITools/openapi-generator/issues/230)

This PR fixes https://github.com/OpenAPITools/openapi-generator/issues/2721


d-date:
@4brunu Thanks for contribution but we plan to make another generator which reduce dependencies. #4086

Since Xcode 11 has officially released so I restart to work on. Sorry for late work.

4brunu:
@d-date No problem 🙂
Can you elaborate on which dependencies are you removing? Is is Alamofire or another one?

wing328:
Like other generators such as Java, Ruby that are more mature, I foresee Swift 5 generators will support the `library` option for different HTTP libraries so my take is to accept this one before anyone has time to add the library option for this Swift 5 generator.

d-date:
@wing328 I don't agree with you. IMO, all dependencies should be removed since no more maintain so many templates for response types and APIClients.
We should provide only interfaces and models.

d-date:
@4brunu I'm try to remove all dependencies including Alamofire since this client shouldn't have any dependencies. We want to select own API Clients and response type, including Combine, by myself.

Instead, in other repository, we should provide example API Client for users so we can focus on implementing OAS itself.

We Swift developers hope to make this generator stable for reducing developing time.

wing328:
I see what you mean. I can only share my experience working on this project. You will always find users who want to use a particular HTTP library: Java client now supports 10+ HTTP libraries: feign, okhttp, Dart supports 2 or 3 via different generators. TS supports a lot via different generators.

Besides Alamofire, there's another one by IBM: https://github.com/IBM-Swift/SwiftyRequest

I'm not surprised if we are going to see a request from the user to support SwiftyRequest in the not too distant future.

My take is to support a swappable ApiClient class from day one and leave it up to the user to choose which HTTP library they want to use.

4brunu:
@d-date I also prefer to reduce the number of dependencies, so it would be nice to rely on `URLSession` instead of `Alamofire` and I even considered creating a PR that removes `Alamofire`.

But on the other hand I think it would be nice to provide an optional integration with `RxSwift.

My concern with a big new Swift 5 generator is how long it will take, because we could introduce so many improvements, `Result`, `Combine`, `URLSession`, etc, that can take a long time.

This PR actually doesn't provide any value in itself, because it's the same as in Swift 4, but it will lower the barrier to anyone that wants to contribute with improvements to the Swift 5 generator.

What do you think of a more incremental approach?

By the way, I'm totally ok closing this PR 🙂

Tovkal:
I would merge this as is. My current use case is: I am using Swagger 2 until now, with Swift 5, and want to update to Swagger 3. So I just want to update the Swagger version while keeping the rest as it is.

I understand that there is room for improvement (decoupling from networking dependencies, adding other networking options) but I too believe this should be delivered incrementally.

wing328:
## IMPORTANT

This generator is relatively new and being actively developed. It's considered as beta/alpha and there will be breaking changes (new features, bug fixes, etc) to this generator without further notice.

Issue #4088, opened 2019-10-07
[REQ] Kotlin generator with non public API


4brunu:
### Is your feature request related to a problem? Please describe.

I'm integrating the generated code inside a library I'm creating and I would like for the generated code to no be exposed.

## Describe the solution you'd like

It would be nice to have an option for the generated code to be internal.

PR #4089, opened 2019-10-07
[kotlin] add option for non public api

4brunu:
### PR checklist

@jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04)

### Description of the PR

I'm integrating the generated code inside a library I'm creating and I would like for the generated code to no be exposed.
It would be nice to have an option for the generated code to be internal.
This fix https://github.com/OpenAPITools/openapi-generator/issues/4088
I used the same approach that is used in the [C# generator](https://github.com/OpenAPITools/openapi-generator/blob/d2ce584a3a902e5921a7aa5db3bc8793932a995f/modules/openapi-generator/src/main/resources/csharp-netcore/visibility.mustache)


modules/openapi-generator/src/main/resources/kotlin-client/visibility.mustache
```diff
@@ -0,0 +1 @@
+{{#nonPublicApi}}internal {{/nonPublicApi}}
```
jimschubert: Question: Why put a single value into a new mustache file rather than just set a `visibility` property to be consumed in the templates? This would make sense if there were two condition (nonPublicApi=false => `public` and nonPublicApi=true => internal), but as a single value where Kotlin's default visibility is public probably makes this confusing from a user perspective.

modules/openapi-generator/src/main/resources/kotlin-client/api.mustache
```diff
@@ -17,7 +17,7 @@ import {{packageName}}.infrastructure.Success
 import {{packageName}}.infrastructure.toMultiValue
 
 {{#operations}}
-class {{classname}}(basePath: kotlin.String = "{{{basePath}}}") : ApiClient(basePath) {
+{{>visibility}}class {{classname}}(basePath: kotlin.String = "{{{basePath}}}") : ApiClient(basePath) {
```
jimschubert: All `{{>visibility}}` should be followed by a space. I see that there's a space in `visibility.mustache`, but if a user were to modify `visibility.mustache` to remove the tags, the trailing space may be removed (for example, if the user's git configuration removes trailing spaces). Rather than open to such bugs, we should aim for mustache files to generate valid code regardless of spacing in imported templates.

modules/openapi-generator/src/main/resources/kotlin-client/visibility.mustache
4brunu: I followed the C# example, but I can change that 👍  Should I change it?

modules/openapi-generator/src/main/resources/kotlin-client/api.mustache
4brunu: The problem is that the space is only needed if the the class is internal, if it's `public` it's not needed.
And if I would add an extra space in the `public` case ktlint would fail with and error.

Maybe this it's easier to explain with an example of the generated code:
```
    internal class UserInternal() // note 4 spaces
     class UserPublic() // note 5 spaces ktlint error
```

Do you have any idea on how to fix this and get this output?
```
    internal class UserInternal() // note 4 spaces
    class UserPublic() // note 4 spaces
```

I'm not sure if I explained it right 😟

jimschubert: If that's the case, I think it would be better to not have `visibility.mustache` and instead inline it like this:

```
{{#nonPublicApi}}internal {{/nonPublicApi}}class {{classname}}
```

This way there's not an additional layer of templating for people to change/remove `internal`, and it'll be more clear how the space applies at the usage site.

modules/openapi-generator/src/main/resources/kotlin-client/visibility.mustache
jimschubert: Based on your other response, I think it would be best to remove `visibility.mustache` and inline this one-liner everywhere instead of the import.

modules/openapi-generator/src/main/resources/kotlin-client/api.mustache
4brunu: Thanks for the suggestion, it's done 👍

priettt:
@jimschubert @4brunu Sorry to revive such an old PR, but I have two questions:
- Should `nonPublicApi` appear in the [kotlin generator docs](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/kotlin.md)?
- Should adding this property through `configOptions` also be supported? It's only working if I add it in additionalProperties

PR #4094, opened 2019-10-08
Fix example docker command line

bclermont:
Problem:

```bash
cd ./modules/openapi-generator/src/test/resources/3_0/
docker run --rm \
> -v ${PWD}:/local openapitools/openapi-generator-cli generate \
> -i petstore.yaml \
> -g go \
> -o /local/out/go
[error] The spec file is not found: petstore.yaml
[error] Check the path of the OpenAPI spec and try again.
```

Solution: specify absolute path to input file

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -68,7 +68,7 @@ Here's an example generating a Go client:
 ```bash
 docker run --rm \
   -v ${PWD}:/local openapitools/openapi-generator-cli generate \
-  -i petstore.yaml \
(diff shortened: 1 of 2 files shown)


wing328:
@bclermont thanks for the fix.

I may later switch it to URL instead of local file.

wing328:
@bclermont thanks for the PR, which has been included in v4.2.0 release: https://twitter.com/oas_generator/status/1189824932345069569

PR #4131, opened 2019-10-11
[typescript-inversify] Fix multipart form uploads

### LINKED ISSUE #4129: [BUG][typescript-inversify] hasFormData request requires a 'body' parameter that does not exist
##### Description

`body` parameter is present on a hasFormData request, but `body` is not defined.

Under normal circumstances (a JSON POST) 'body' is defined as a typed view model parameter, but if the request is for a file upload, the post request tries to use a `body` parameter is that is not defined. 

Should this be formData?

image

4.1.2

siada:
### PR checklist

### Description of the PR

This PR fixes an issue with the typescript-inversify generator for multipart file uploads.

When trying to use this generator within my own code I found that for file uploads, it was adding a content-type header of `application/x-www-form-urlencoded;charset=UTF-8` which was incorrect

Based on my own testing I found that when doing a multipart upload using fetch(), the content-type header should be left unset as the browser will itself handle adding the multipart boundary fields

This PR also fixes #4129 as the 'body' param isn't set for www-form or mutlipart requests


modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
```diff
@@ -152,7 +152,12 @@ export class {{classname}} {
 
 {{#hasFormParams}}
         let formData: FormData = new FormData();
+        {{#isMultipart}}
+        headers['Content-Type'] = 'multipart/form-data';
```
macjohnny: It should‘t be set explicitly, as there would be a problem with the separator, or am I missing something?

siada: Yes that’s technically correct, it shouldn’t be explicitly set. But because of the way the HttpClient does an Object.assign to get some default headers, if you leave content-type undefined it becomes application/json which is still wrong

So alongside setting this content-type, I added a check to HttpClient to see if it is a multipart upload and do a `delete headers[“content-type”]`

macjohnny: ok, I see. maybe
https://github.com/OpenAPITools/openapi-generator/blob/07d145828f33e9e2d9e5be6694bfb1d29db4fe07/modules/openapi-generator/src/main/resources/typescript-inversify/HttpClient.mustache#L46-L51

should be fixed instead? @gualtierim what is your opinion?

modules/openapi-generator/src/main/resources/typescript-inversify/HttpClient.mustache
```diff
@@ -51,6 +51,12 @@ class HttpClient implements IHttpClient {
     };
 
     private performNetworkCall(url: string, method: string, body?: any, headers?: Headers): Observable<HttpResponse> {
+
+        // when using fetch & a multipart upload, the requests content-type is handled by the browser, so should be left unset otherwise the multipart boundry is not added
+        if(headers && headers["Content-Type"] == "multipart/form-data") {
```
macjohnny:
```suggestion
        if(headers && headers["Content-Type"] === "multipart/form-data") {
```

macjohnny:
@siada can you please re-generate the samples and merge the latest master?

wing328:
@siada thanks for the PR, which has been included in v4.2.0 release: https://twitter.com/oas_generator/status/1189824932345069569

PR #4137, opened 2019-10-13
[python] Add option to skip client validations

slash-arun:
### PR checklist

### Description of the PR

Adds option to skip client side validations


slash-arun:

cc: @wing328

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -215,10 +229,13 @@ class {{classname}}(object):
         if not isinstance(other, {{classname}}):
             return False
 
-        return self.__dict__ == other.__dict__
+        return self.to_dict() == other.to_dict()
```
wing328: What's the reason to use `to_dict()` instead of `__dict__`?

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -215,10 +229,13 @@ class {{classname}}(object):
...
 
     def __ne__(self, other):
         """Returns true if both objects are not equal"""
-        return not self == other
```
wing328: Performance would `not self == other` be better?

modules/openapi-generator/src/main/resources/python/model.mustache
slash-arun: So far configuration was not part of the model, now configuration is. So if the user does not provide configuration. The configuration is initialized since two objects have different configuration self.__dict__ == other.__dict__ will return false even though its the same object.

Test("abc") will not be equal to Test("abc") since configuration objects inside both are different. This change is required to prevent this behaviour

modules/openapi-generator/src/main/resources/python/model.mustache
slash-arun: This again relates to the configuration object. Since Test("abc") will not be equal to Test("abc") because config is initialized separately for both the objects.

modules/openapi-generator/src/main/resources/python/model.mustache
wing328: @slash-arun thanks for the explanation. I wonder if you can add test cases to https://github.com/OpenAPITools/openapi-generator/tree/master/samples/client/petstore/python/tests to cover the issue moving forward.

modules/openapi-generator/src/main/resources/python/model.mustache
wing328: @slash-arun thanks for the explanation. I wonder if you can add test cases to https://github.com/OpenAPITools/openapi-generator/tree/master/samples/client/petstore/python/tests to cover the issue moving forward.

modules/openapi-generator/src/main/resources/python/model.mustache
slash-arun: Sure, will work on the test cases

modules/openapi-generator/src/main/resources/python/model.mustache
slash-arun: Sure, will work on the test cases

spacether:
@slash-arun can you describe the use case here more in depth?
Why would one want to turn off client side validations?
Is this when receiving data from the server, instantiating an instance of the model on the client?
Why not change the spec to remove the validations?
Will the server accept these values that don't pass validation checks?

slash-arun:
The reason is we would want to run integration tests, similar to https://github.com/swagger-api/swagger-codegen/issues/5530 but for python client

wing328:
Yup, that's a use case we've encounterd previously. Ruby client generator (and another one I forgot) supports similar option to disable client side validation so that one can use it to test the API backend

wing328:
I've created https://github.com/OpenAPITools/openapi-generator/issues/4168 for tracking instead.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -149,6 +149,8 @@ class Configuration(six.with_metaclass(TypeWithDefault, object)):
         self.retries = None
         """Adding retries to override urllib3 default value 3
         """
+        # Disable client side validation
```
jonashaag: @wing328 typo? Shouldn't this be "Enable"

grking8:
Can this option only be set in the code, or via the CLI also?

PR #4138, opened 2019-10-13
[python] Add ability to get allowed values

slash-arun:
### PR checklist

### Description of the PR

Currently, there is no way to get the list of allowed values for an enum. This PR adds a class variable to get all the allowed values

Why this is needed?
fixes https://github.com/OpenAPITools/openapi-generator/issues/4139

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/model.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model.mustache
@@ -77,8 +77,7 @@ class {{classname}}(object):
 {{#allowableValues}}
     @staticmethod
     def get_allowed_values():
-        allowed_values = [{{#enumVars}}{{classname}}.{{name}}{{^-last}}, {{/-last}}{{/enumVars}}]  # noqa: E501
-        return allowed_values
+        return [{{#enumVars}}{{classname}}.{{name}}{{^-last}}, {{/-last}}{{/enumVars}}]  # noqa: E501
 
 {{/allowableValues}}
 {{#vars}}
--- a/modules/openapi-generator/src/main/resources/python/model.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model.mustache
@@ -74,6 +74,13 @@ class {{classname}}(object):
 {{/required}}
 {{/vars}}
 
+{{#allowableValues}}
+    @staticmethod
+    def get_allowed_values():
+        allowed_values = [{{#enumVars}}{{classname}}.{{name}}{{^-last}}, {{/-last}}{{/enumVars}}]  # noqa: E501
+        return allowed_values
+
+{{/allowableValues}}
 {{#vars}}
     @property
     def {{name}}(self):
--- a/modules/openapi-generator/src/main/resources/python/model.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model.mustache
@@ -25,6 +25,10 @@ class {{classname}}(object):
 {{/-last}}
 {{/enumVars}}{{/allowableValues}}
 
+{{#allowableValues}}
+    allowable_values = [{{#enumVars}}{{classname}}.{{name}}{{^-last}}, {{/-last}}{{/enumVars}}]  # noqa: E501
+
+{{/allowableValues}}
     """
     Attributes:
       openapi_types (dict): The key is attribute name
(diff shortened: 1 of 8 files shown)


slash-arun:

cc: @wing328

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -74,6 +74,12 @@ class {{classname}}(object):
 {{/required}}
 {{/vars}}
 
+{{#allowableValues}}
+    @staticmethod
+    def get_allowed_values():
+        return [{{#enumVars}}{{classname}}.{{name}}{{^-last}}, {{/-last}}{{/enumVars}}]  # noqa: E501
+
+{{/allowableValues}}
```
spacether: @slash-arun why not add this data as a dict in the python class, not at the instance level?
The data does not change from instance to instance.
See some parallel work on that here:
- template https://github.com/OpenAPITools/openapi-generator/pull/4154/files#diff-03d0471b3e76c94b0460b466312a2332
- result https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/models/enum_test.py

slash-arun: Agreed, that makes more sense will push the change, Thanks for reviewing @spacether

slash-arun:
Removed the static method and created a variable instead. cc: @spacether @wing328

slash-arun:
Hi, Sorry to ping in your busy schedules. Could you please let me know if there are any other suggestions or alternative ways you would want to implement this, I will work on it. Would really help if this goes out in the next release.
Thanks!

samples/client/petstore/python/petstore_api/models/outer_enum.py
```diff
@@ -30,6 +30,8 @@ class OuterEnum(object):
     APPROVED = "approved"
     DELIVERED = "delivered"
 
+    allowable_values = [PLACED, APPROVED, DELIVERED]  # noqa: E501
```
spacether: @slash-arun is this what you intend?
These allowable_values are the keys or accessors on this enum class, not the accepted values

spacether: Also, do you want this ability on api endpoints too?

If so could the python-experimental generator meet your needs? It has this feature for models:
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/models/enum_class.py#L48
and for api_endpoints:
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/api/fake_api.py#L819

slash-arun: These will evaluate to allowed values only right ['placed', 'approved', 'delivered'] ?

Could you let me know if python-experimental is fairly stable ?

spacether: I'm not sure, you will have to import your python package and see if it works that way. My python enum game is not strong.

So I made python-experimental recently as a fork of the python client to add a new additionalproperties feature and to structure the data better so users can automatically have features like this one that we are discussing. It may change more quickly than the python client, but as long as you version pin to the version that you want it should be stable.
It has not existed for very long, maybe a couple of months. It also is structured in a way where we will more easily be able to implement oneOf, allOf, anyOf inheritance in the generator.

slash-arun: Thanks for the detailed explanation @spacether :) I am just a bit concerned about using it in production at this point, to be honest.

spacether:
No worries. It's your call

samples/client/petstore/python/petstore_api/models/outer_enum.py
spacether: No worries. It's your call

slash-arun:
Can anyone please confirm if this idea is rejected or if it will be included in the next release

Thanks

cc: @wing328

wing328:
@slash-arun we can consider this feature as beta but it may be rolled back without considering as a breaking change if the Python community does not want this feature or prefer something else.

slash-arun:
@wing328, Noted. Thanks.

PR #4182, opened 2019-10-19
java: allow aliased types to have custom mappings

### LINKED ISSUE #2858: [BUG] Generator 3.3.4 does not respect language-specific primitive for string refs
##### Description

I am trying to map a custom data type to a JSON string-valued property (more or less equivalent to a custom format); Jackson will transparently handle the conversion for me in Java. However, the generator does not observe user language-specific primitives when a `$ref` type is marked as a string.

This behavior is observed with 3.3.4

##### OpenAPI content

```yaml
# correctly identifies that EntityIdentifier is a string and produces a Map<String, String>,
# but does not use the LSP to use a custom data type
components:
  schemas:
    EntityIdentifier:
      type: string
      pattern: '^\w+:[\w\d-]+$'
    ResolveExpressionRequest:
      type: object
      properties:
        expression:
          type: string
          minLength: 1
        rootObject:
          type: object
          additionalProperties:
              $ref: '#/components/schemas/EntityIdentifier'
```

```yaml
# Generates a Map<String, EntityIdentifierDto>
components:
  schemas:
    EntityIdentifier:
      type: object
    ResolveExpressionRequest:
      type: object
      properties:
        expression:
          type: string
          minLength: 1
        rootObject:
          type: object
          additionalProperties:
              $ref: '#/components/schemas/EntityIdentifier'
```

Invoked via Maven plugin:
```xml
<plugin>
    <version>3.3.4</version>
    <executions>
        <execution>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>${project.basedir}/src/main/openapi/openapi.yaml</inputSpec>
                <languageSpecificPrimitives>EntityIdentifier</languageSpecificPrimitives>
                <importMappings>EntityIdentifier=foo.EntityIdentifierDto</importMappings>
                <configOptions>
                  <library>resttemplate</library>
                  <java8>true</java8>
                </configOptions>
            </configuration>
        </execution>
    </executions>
  </plugin>
```

##### Steps to reproduce

As above, a `$ref` that is of `type: string` does not get overridden by LSP.

##### Suggest a fix

I'm not familiar enough with the codebase to be certain, but it seems that what may be happening is that the `type: string` is triggering the `isPrimitiveType` logic and short-circuiting the lookup that would happen otherwise.

fabiokung:
~Fixes #2858.~ UPDATE: I was wrong and misundertood the code, this PR does not fix the issue, but https://github.com/OpenAPITools/openapi-generator/pull/4182#issuecomment-546092134 discusses an alternative on how to accomplish the same. I'd still consider the original issue to be a bug, it should be possible to define type mappings for any type definition in the spec.

This will allow `kubernetes-client-java` to switch to openapi-generator (kubernetes-client/java#709).

The kubernetes API swagger spec has types aliased to primitive types, e.g.: 

```
    "intstr.IntOrString": {
      "description": "IntOrString is a type that can hold an int32 or a string.  When used in JSON or YAML marshalling and unmarshalling, it produces or consumes the inner type.  This allows you to have, for example, a JSON field that can accept a name or number.",
      "format": "int-or-string",
      "type": "string"
    },
```

However, these types need to be mapped to custom types in the Java client, so we can have custom serialization for them. The generator is being invoked with the following options:

```
  <type-mappings>intstr.IntOrString=IntOrString,resource.Quantity=Quantity</type-mappings>
  <import-mappings>IntOrString=io.kubernetes.client.custom.IntOrString,Quantity=io.kubernetes.client.custom.Quantity</import-mappings>
```

Without changes in this PR, `type-mappings` gets ignored, since `IntOrString` is first resolved to its alias (`string`), before mappings are evaluated.

cc @wing328 @brendandburns @yue9944882

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1666,7 +1666,11 @@ public class DefaultCodegen implements CodegenConfig {
      * @return The alias of the given type, if it exists. If there is no alias
      * for this type, then returns the input type name.
      */
-    public String getAlias(String name) {
+    private String getAlias(String name) {
+        if (typeMapping != null && typeMapping.containsKey(name)) {
+            // allow aliased types to have custom mappings, the mapping will be resolved later
+            return name;
+        }
         if (typeAliases != null && typeAliases.containsKey(name)) {
             return typeAliases.get(name);
         }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractApexCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractApexCodegen.java
@@ -207,14 +207,6 @@ public abstract class AbstractApexCodegen extends DefaultCodegen implements Code
     }
 
     @Override
-    public String getAlias(String name) {
-        if (typeAliases != null && typeAliases.containsKey(name)) {
-            return typeAliases.get(name);
-        }
-        return name;
-    }
-
-    @Override
     public String toDefaultValue(Schema p) {
         if (ModelUtils.isArraySchema(p)) {
             final ArraySchema ap = (ArraySchema) p;
@@ -399,8 +391,6 @@ public abstract class AbstractApexCodegen extends DefaultCodegen implements Code
     @Override
     public String getSchemaType(Schema p) {
         String schemaType = super.getSchemaType(p);
-
-        schemaType = getAlias(schemaType);
 
         // don't apply renaming on types from the typeMapping
         if (typeMapping.containsKey(schemaType)) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -704,6 +704,10 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
 
     @Override
     public String getAlias(String name) {
+        if (typeMapping != null && typeMapping.containsKey(name)) {
+            // allow aliased types to have custom mappings, the mapping will be resolved later
+            return name;
+        }
         if (typeAliases != null && typeAliases.containsKey(name)) {
             return typeAliases.get(name);
         }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -703,18 +703,6 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
     }
 
     @Override
-    public String getAlias(String name) {
-        if (typeMapping != null && typeMapping.containsKey(name)) {
-            // allow aliased types to have custom mappings, the mapping will be resolved later
-            return name;
-        }
-        if (typeAliases != null && typeAliases.containsKey(name)) {
-            return typeAliases.get(name);
-        }
-        return name;
-    }
-
-    @Override
     public String toDefaultValue(Schema p) {
         p = ModelUtils.getReferencedSchema(this.openAPI, p);
         if (ModelUtils.isArraySchema(p)) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaPlayFrameworkServerCodegen.java
@@ -284,7 +284,6 @@ public class ScalaPlayFrameworkServerCodegen extends AbstractScalaCodegen implem
     @Override
     public String getSchemaType(Schema p) {
         String openAPIType = super.getSchemaType(p);
-        openAPIType = getAlias(openAPIType);
 
         // don't apply renaming on types from the typeMapping
         if (typeMapping.containsKey(openAPIType)) {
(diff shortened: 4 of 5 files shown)


brendandburns:
Thank you for figuring this out!!!

yue9944882:
awesome! politely bump @wing328 for a review!

wing328:
Looks good but I'll need to test it locally tomorrow (as it's 1am here ...) before merging it into master

fabiokung:
Please hold on merging. I suspect there's more involved here. `kubernetes-client-java` seems to be hitting a different code path of the generator.

fabiokung:
I believe this is now good to be merged.

I did some more digging, and found that `typeMappings` on aliased types will only be applied when they are of type `string`, **and have a `format`** field specified:

From `DefaultCodegen#getPrimitiveType()`:

```
        } else if (ModelUtils.isStringSchema(schema)) {
            if (typeMapping.containsKey(schema.getFormat())) {
                // If the format matches a typeMapping (supplied with the --typeMappings flag)
                // then treat the format as a primitive type.
                // This allows the typeMapping flag to add a new custom type which can then
                // be used in the format field.
                return schema.getFormat();
            }
            return "string";
        }
```

With the changes in this PR I was able to tweak code generation for `kubernetes-client-java` and have it apply custom type mappings, by rewriting the spec to make sure types with custom mapping also have a custom `format`.

fabiokung:
The tweaks to the configuration, and rewriting of the spec I had to do are here: yue9944882/gen#1

yue9944882:
great, i will rebase https://github.com/kubernetes-client/gen/pull/117 after this pull is merged&released

fabiokung:
@yue9944882 we will also need OpenAPITools/openapi-generator#4226

wing328:
Thanks for the update. I'll review and run some tests locally.

wing328:
@fabiokung I tested it locally but no luck getting it working locally.  Here is how I tested with [fake petstore spec](https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml):

1) Pull the changes into a branch and build the project locally
2) Run `./bin/java-petstore-okhttp-gson.sh --type-mappings OuterString=CustomString`
3) Some aliases to OuterString ([example](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1725)) should be mapped to `CustomString` instead of `string`
4) Grep `CustomString` in the output but didn't find anything

I also tested with https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json but no luck with `--type-mappings io.k8s.apimachinery.pkg.util.intstr.IntOrString=intstr.IntOrString`

Am I testing the change correctly?

fabiokung:
I take that back, the changes are not going to be trivial. For some reason `swagger-parser-core` does not seem to be filling `Schema#name` properties for each definition, so checking if a type definition has a type mapping installed will be a bit tricky, since once the spec has been parsed, we lose access to the name of each `Schema` node.

I will leave that up to others that are more familiar with the codebase. @wing328 feel free to close and track improvements to type mappings in a different GH issue.

yue9944882:
```
    "io.k8s.apimachinery.pkg.util.intstr.IntOrString": {
      "description": "<trimmed>",
      "type": "string",
      "format": "int-or-string"
    },
```

@wing328 @fabiokung thank you for the deep insight. iiuc, tweaking a custom format is not supported in openapi specification, so the `"format": "int-or-string"` from kuberenetes' spec is applying invalid schema here. otoh currently we have another vendor extension `x-kubernetes-int-or-string` as a replacement of replacing $ref to `intstr.IntOrString` definition, it will be better if there's a way to do type-mappings from extension. anway i think https://github.com/kubernetes-client/java/issues/721 can be considered solved, right?

wing328:
@fabiokung @yue9944882 I think I may have a better solution to come up with something you guys need.

wing328:
The following seems to work for me to meet your requirement mapping IntOrString (schema) to something else and do a proper import accordingly: 
```
$ mvn clean package -DskipTests
$ java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g java -i https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json -o /tmp/k8sc --type-mappings int-or-string=IntOfString --import-mappings IntOfString=io.kubernetes.client.custom.IntOrString
```
Can you plesae give it a try?

fabiokung:
This seems to be exactly what I did already, unless I’m missing something?

fabiokung:
> (snip) anway i think [kubernetes-client/java#721](https://github.com/kubernetes-client/java/issues/721) can be considered solved, right?

Yes, I believe the changes I sent you through yue9944882/gen#1 fixes kubernetes-client/java#721, this PR is not needed because it relies on the custom format.

fabiokung:
Edited the original PR description to make it clear this PR per-se is not fixing anything, and while I think it improves the code a bit, it is not really necessary.

wing328:
> I edited the original PR description to make it clear this PR per-se is not fixing anything, and while I think it improves the code a bit, it is not really necessary.

Thanks for the clarification. I'll let others take a look and see if they've any comments on the change.

yue9944882:
> @yue9944882 yes, I believe the changes I sent you through yue9944882/gen#1 fixes kubernetes-client/java#721, this PR is not needed because it relies on the custom format.

@fabiokung i re-generated w/ https://github.com/kubernetes-client/gen/pull/117, the it doesn't seem to work, all the fields supposed to be `IntOrString` turn out to become `String`. am i missing anything?

fabiokung:
@yue9944882 I got a notification from your comment saying you couldn't make custom type mappings work with kubernetes-client/gen#117, but the comment seems to have gone away. Were you able to make it work?

fabiokung:
Closing as this is not really needed, and ended up just being a red herring and suggestion for improvements to the code.

PR #4209, opened 2019-10-21
typescript-angular: Allow for custom naming of ApiModule

### LINKED ISSUE #4101: [REQ][typescript-angular] Option to rename ApiModule and Configuration

Generated angular client module will always be called `ApiModule`. 
The module also exposes a `Configuration` object which also cannot be renamed.

This isn't great when multiple generated clients are needed in a single project.

I'd like an Config Option for typescript-angular that can configure the `Module` and `Configuration` prefix. 

Option | Description | Values | Default
-- | -- | -- | --
modulePrefix | The prefix of the file and class of the generated module (\<prefix>.module.ts and `class <prefix>Module`). |   | Api
configuraionPrefix | The prefix of the file and class of the generated configuration object (\<prefix>Configuration.ts and `class <prefix>Configuration`). |   | null

**Note** _the names are in the default fileNaming: camelCase format, but kebab-case should also be supported_

mrmeku:
In cases where a user wants to generate multiple module's, it is convenient to have these modules have different names.

This change introduces a new additional property to the typescript-angular generator. The property is named `apiModulePrefix` and will serve as the prefix for the generated ApiModule class. For example, if `apiModulePrefix=MyPrefix` were to be set, the generated module would be named `MyPrefixApiModule`.

The inspiration for this change is that in many projects I contribute to, we generate many ApiModules for different openapi services. Since all of these modules have the same name it forces to developers to use the `import { ApiModule as BlahModule } from ...` syntax which is slightly annoying and is incompatible with some code inspection tools which expect that all classes have different names.

The change is non breaking and tests have been added.

### PR checklist

fixes #4101


bin/typescript-angular-v8-petstore-provided-in-root-with-prefixed-module-name.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+executable="./modules/openapi-generator-cli/target/openapi-generator-cli.jar"
+
+if [ ! -f "$executable" ]
+then
+  mvn -B clean package
+fi
+
+# if you've executed sbt assembly previously it will use that instead.
```
macjohnny: please also add a windows version of this script and add and entry to https://github.com/OpenAPITools/openapi-generator/blob/master/bin/typescript-angular-petstore-all.sh

mrmeku: Done! Thanks for the review @macjohnny , I very much appreciate your responsiveness. You're awesome sauce

wing328:
@mrmeku thanks for the PR, which has been included in v4.1.2 release: https://twitter.com/oas_generator/status/1189824932345069569

PR #4226, opened 2019-10-22
okhttp-gson: fix SSL settings with okhttp3

fabiokung:
The old code boostrapping TLS settings used to work with older okhttp (< 3), but will throw NullPointerExceptions with okhttp3, which I noticed while working on kubernetes-client/java#709.

cc @wing328 @yue9944882 @brendandburns

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
@@ -4,6 +4,7 @@ package {{invokerPackage}};
 
 import okhttp3.*;
 import okhttp3.internal.http.HttpMethod;
+import okhttp3.internal.tls.OkHostnameVerifier;
 import okhttp3.logging.HttpLoggingInterceptor;
 import okhttp3.logging.HttpLoggingInterceptor.Level;
 import okio.BufferedSink;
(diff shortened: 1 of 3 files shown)


fabiokung:
CI failures seem unrelated.

yue9944882:
@fabiokung can you paste a re-pro for the NPE issue? thanks

fabiokung:
@yue9944882 any usage of `ApiClient` that leads to `applySslSettings()` falling into the default case (no custom CA cert, `veryfyingSsl=true`) will throw the NPE.

For example, without changes in here, the snippet below throws an NPE:

```java
        ApiClient api = new ApiClient();
        api.setVerifyingSsl(true);
```

yue9944882:
> ApiClient api = new ApiClient();

the default construction of `ApiClient` is targeting an http endpoint on localhost. so the second line doesn't make sense, isn't it?

wing328:
I'm able to repeat the NPE issue with `api.setVerifyingSsl(true);` and confirm your PR fixes it.

wing328:
> the default construction of ApiClient is targeting an http endpoint on localhost. so the second line doesn't make sense, isn't it?

ApiClient can be treated as a wrapper of the underlying HTTP library for any endpoints/hosts specified in the OpenAPI doc/spec

wing328:
Filed https://github.com/OpenAPITools/openapi-generator/pull/4252 to add a test covering the issue moving forward.

Issue #4159, opened 2019-10-15
[BUG] [typescript-rxjs] Typescript error in generated files for query parameters


balazs-zsoldos:
##### Description

In case Typescript strict mode is on, there is typescript error in the generated code if a function has one or more query parameters.

##### openapi-generator version

4.0.3 - 4.1.3

##### OpenAPI declaration file content or url

The standard petstore yaml file that is used to generate the samples.

##### Command line used for generation

See the samples directory for the generated files.

##### Steps to reproduce

* Download the typescript-rxjs/default (or any other) sample from the github master branch
* Copy it into a project where typescript strict mode is on, or
* Just open it in Visual Code to see the error in UserApi.ts loginUser function.

Generated code:

`   const query: HttpQuery = {
            ...(requestParameters.username && { 'username': requestParameters.username }),
            ...(requestParameters.password && { 'password': requestParameters.password }),
        };`

Error message:

`Type '{} | { 'username': string; } | { 'password': string; } | { 'password': string; 'username': string; }' is not assignable to type 'HttpQuery'.
  Type '{}' is not assignable to type 'HttpQuery'.
    Index signature is missing in type '{}'.`

##### Related issues/PRs

none

##### Suggest a fix

The generated code was ok until 4.0.2. It would be good to check why the mustache files were changed to the new code.

macjohnny:
changing the generated code to
```typescript
const query: HttpQuery = <HttpQuery>{ ...(requestParameters.username && { 'username': requestParameters.username }), ...(requestParameters.password && { 'password': requestParameters.password }), };
```
instead of
```typescript
const query: HttpQuery = { ...(requestParameters.username && { 'username': requestParameters.username }), ...(requestParameters.password && { 'password': requestParameters.password }), };
```
seems to fix the issue.
[...]

PR #4250, opened 2019-10-24
[typescript-rxjs] performance improvements, bugfix for falsy parameters

denyo:
Based on the discussion and feedback in #4159 I spent some time testing different approaches to compose the `HttpQuery` parameters. 
Of course using `if` checks and manipulating the object over spreading an expression is a lot more performant. Further improvements can be achieved by destrucuring the parameters.

Results of the tests can be found here 
* <https://jsperf.com/object-creation-or-spreadoperator-extended> with 6/6 parameters 
* <https://jsperf.com/object-creation-or-spreadoperator-extended-2> with 2/6 parameters

Since the pet store example is a bit limited here are some before and afters from another project:

```diff
- throwIfRequired(requestParameters, 'skillIds', 'deleteMySkillsUsingDELETE');
+ throwIfNullOrUndefined(skillIds, 'deleteMySkillsUsingDELETE');
 
- const query: HttpQuery = {
-     ...(requestParameters.skillIds && { 'skillIds': requestParameters.skillIds }),
+ const query: HttpQuery = { // required parameters are used directly since they are already checked by throwIfNullOrUndefined
+     'skillIds': skillIds,
  };
```

```diff
- const query: HttpQuery = {
-     ...(requestParameters.page && { 'page': requestParameters.page }),
-     ...(requestParameters.size && { 'size': requestParameters.size }),
-     ...(requestParameters.skillIds && { 'skillIds': requestParameters.skillIds }),
- };
+ const query: HttpQuery = {};
+
+ if (page != null) { query['page'] = page; }
+ if (size != null) { query['size'] = size; }
+ if (skillIds != null) { query['skillIds'] = skillIds; }
```

```diff
- const query: HttpQuery = {
-     ...(requestParameters.keyword && { 'keyword': requestParameters.keyword }),
-     ...(requestParameters.lang && { 'lang': requestParameters.lang }),
+ const query: HttpQuery = { // required parameters are used directly since they are already checked by throwIfNullOrUndefined
+     'keyword': keyword,
  };
 
+ if (lang != null) { query['lang'] = lang; }
```

Sizewise after bundeling one of my projects with CRA2 the main chunk went from 344 KB to 340 KB. After gzip it's about 0.1 KB less. When looking at the character count it went from 343498 to 339245. That's not huge but still.

In Summary:
* fixed a bug where parameters like `false` and `0` were lost since they were evaluated with `&&` inside the expression
* improved performance
* less redundancy
* smaller bundle
* `throwIfRequired` is marked as deprecated and unused within the generated code

### PR checklist


modules/openapi-generator/src/main/resources/typescript-rxjs/apis.mustache
```diff
@@ -37,11 +37,11 @@ export class {{classname}} extends BaseAPI {
      * {{&summary}}
      {{/summary}}
      */
-    {{nickname}} = ({{#allParams.0}}requestParameters: {{operationIdCamelCase}}Request{{/allParams.0}}): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}> => {
+    {{nickname}} = ({{#allParams.0}}{ {{#allParams}}{{paramName}}{{^-last}}, {{/-last}}{{/allParams}} }: {{operationIdCamelCase}}Request{{/allParams.0}}): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}> => {
```
macjohnny: i suggest not to unpack the parameters, since it will lead to many naming conflicts

denyo: You're right, they may conflict with `query`, `headers`, `body` and `formData` now. 
Those could either be suffixed with something like `nickname` or while destructuring the params the ones that collide get renamed. 
I am going to try the latter since it seems cleaner.

denyo: I got it working like this where params are checked for the keys `query`, `headers` and `formData` (`body` wasn't an issue).

The result for those looks like 
```diff
-    loginUser = ({ formData, query }: LoginUserRequest): Observable<string> => {
-        throwIfNullOrUndefined(formData, 'loginUser');
-        throwIfNullOrUndefined(query, 'loginUser');
+    loginUser = ({ formData: formDataAlias, query: queryAlias }: LoginUserRequest): Observable<string> => {
+        throwIfNullOrUndefined(formDataAlias, 'loginUser');
+        throwIfNullOrUndefined(queryAlias, 'loginUser');
 
         const query: HttpQuery = { // required parameters are used directly since they are already checked by throwIfNullOrUndefined
-            'formData': formData,
-            'query': query,
+            'formData': formDataAlias,
+            'query': queryAlias,
         };
```

```diff
-    findPetsByTags = ({ headers }: FindPetsByTagsRequest): Observable<Array<Pet>> => {
-        throwIfNullOrUndefined(headers, 'findPetsByTags');
+    findPetsByTags = ({ headers: headersAlias }: FindPetsByTagsRequest): Observable<Array<Pet>> => {
+        throwIfNullOrUndefined(headersAlias, 'findPetsByTags');

        const headers: HttpHeaders = {
            ...
        };

         const query: HttpQuery = { // required parameters are used directly since they are already checked by throwIfNullOrUndefined
-            'headers': headers.join(COLLECTION_FORMATS['csv']),
+            'headers': headersAlias.join(COLLECTION_FORMATS['csv']),
         };
```

To get there I extended `CodegenParameter` by `paramNameAlternative` and `paramNameOrAlternative` in order to keep templates simple.
Inheriting from `CodegenParameter` an creating a new class would be possible but this would probably be a first in the project and I wanted to avoid a copy constructor because of maintainablity.

So now my question are:
1) Is this acceptable or should I extend `CodegenParameter` locally in the generator?
2) Is extending the java code favored over having logic in templates? In this case the only other way would be to have `{{#paramNameAlternative}}{{paramNameAlternative}}{{/paramNameAlternative}}{{^paramNameAlternative}}{{paramName}}{{/paramNameAlternative}}` in like 12 places instead of `{{paramNameOrAlternative}}`.

macjohnny: thanks for your improvements.
I still don't understand why such an effort has to be made instead of just using the current `requestParameters`

denyo: The main effort is already done. 
The result is more performant code at a reduced file size and believe anyone does welcome this.
Perhaps @wing328 can answer the two questions from above?

wing328: Please ping me via Slack to have a discussion on this in Jan 2020 when you've time.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenParameter.java
```diff
@@ -27,7 +27,7 @@
     public boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
             isCookieParam, isBodyParam, hasMore, isContainer,
             secondaryParam, isCollectionFormatMulti, isPrimitiveType, isModel, isExplode;
-    public String baseName, paramName, dataType, datatypeWithEnum, dataFormat,
+    public String baseName, paramName, paramNameAlternative, dataType, datatypeWithEnum, dataFormat,
```
wing328: I missed the reason to introduce `paramNameAlternative`. What about using vendor extensions instead similar to what we've done in other generators?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
```diff
@@ -250,38 +253,65 @@ private void updateOperationParameterEnumInformation(Map<String, Object> operati
         operations.put("hasEnums", hasEnums);
     }
 
+    private void setParamNameAlternative(CodegenParameter param, String paramName, String paramNameAlternative) {
+        if (paramName == param.paramName) {
```
wing328: About `==`, did you mean to use `.equals()` instead to compare the actual string value?

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenParameter.java
denyo: Thanks for the hint. I changed it to `vendorExtensions`.

macjohnny:
@denyo are these changes ready to be part of the next release?

denyo:
> @denyo are these changes ready to be part of the next release?

Yes.

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -121,11 +115,14 @@ export class BaseAPI {
      * Create a shallow clone of `this` by constructing a new instance
      * and then shallow cloning data members.
      */
-    private clone = <T extends BaseAPI>(): T =>
```
ermik, line 124: Why was this change made?

denyo: I think I saw now point in having this generic there. 
Now I can see that people might want to extend `BaseAPI` and it won't work then? Perhaps the generic could be avoided by using `private clone = (): this =>`?

ermik: Yes, it should be `this` in the return type, because without `clone` and inheriting methods are not polymorphic.

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -167,12 +163,21 @@ const queryString = (params: HttpQuery): string => Object.keys(params)
...
 export const throwIfRequired = (params: {[key: string]: any}, key: string, nickname: string) => {
-    if (!params || params[key] === null || params[key] === undefined) {
+    if (!params || params[key] == null) {
         throw new RequiredError(`Required parameter ${key} was null or undefined when calling ${nickname}.`);
     }
 };
 
+export const throwIfNullOrUndefined = (value: any, nickname?: string) => {
+    if (value == null) {
+        throw new Error(`Parameter "${value}" was null or undefined when calling "${nickname}".`);
```
jtvd78, line 177: Won't this here  just print `Parameter <null/undefined> was null or undefined when calling <method>`? Previously, `throwIfRequired` passed in the param name as a string, but here we are only printing the value.

macjohnny: @jtvd78 would you like to file a PR to fix this?

jtvd78: I can do that

jtvd78: See https://github.com/OpenAPITools/openapi-generator/pull/6083

PR #4254, opened 2019-10-24
[Kotlin] Correct isInherited flag for Kotlin generators

mtraynham:
Most of the Kotlin generator templates loop `requiredVars` and `optionalVars` ([1](https://github.com/OpenAPITools/openapi-generator/blob/v4.1.3/modules/openapi-generator/src/main/resources/kotlin-client/data_class.mustache#L32-L36), [2](https://github.com/OpenAPITools/openapi-generator/blob/v4.1.3/modules/openapi-generator/src/main/resources/kotlin-spring/dataClass.mustache#L11-L13)) .  Unfortunately, it looks like during the [`CodegenModel.removeAllDuplicatedProperty`](https://github.com/OpenAPITools/openapi-generator/blob/v4.1.3/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L527), the CodegenProperty's are all [cloned](https://github.com/OpenAPITools/openapi-generator/blob/v4.1.3/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L550-L552) and therefore the current implementation of only updating the `allVars` list isn't sufficient to reflect that in `requiredVars` or `optionalVars`.

All 3 lists (`allVars`, `requiredVars`, `optionalVars`) should be updated with the `isInherited` property.

Also, the `KotlinSpringServerCodegen` was doing the same thing as `AbstractKotlinCodegen` so I removed it.

### PR checklist

- [X] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.1.x`, `5.0.x`. Default: `master`.

@jimschubert @dr4ke616 @karismann @Zomzog

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
@@ -33,7 +33,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.openapitools.codegen.utils.StringUtils.*;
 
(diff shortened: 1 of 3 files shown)


mtraynham:
@jimschubert @dr4ke616 @karismann @Zomzog @wing328  any feedback on this would be appreciated!

jimschubert:
@mtraynham sorry. I've been super busy. I'm trying to get caught up on some issues over the next couple of days.

Good catch on the clone call in `CodegenModel.removeAllDuplicatedProperty`... that seems a little suspect to me and I can't tell if it was an intentional side effect there or not.

Looking over your code, it looks good and the comments are helpful. I was wondering if it would be possible to check the properties on the "final" model in the test, to avoid regressions? For example, if `pm.requiredVars` and `pm.optionalVars` were ever inadvertently truncated or didn't hold exactly the intended properties, it seems this test would still pass.

mtraynham:
Hey @jimschubert, I'll rebase so they aren't conflicts and add a check to verify that all the names of the properties in `requiredVars` and `optionalVars`.   Sound good?

jimschubert:
@mtraynham Thanks! I checked the latest commit and it looks good. We'll merge once CI checks have completed.

mtraynham:
Awesome thanks @jimschubert!

I wanted to point out one thing, my test case change also coincides with a separate bug I found, #4253, where the `optionalVars` would have actually also contained `a`.  This occurs when adding the `requiredItems` to the outer model of an `allOf` composed object.  If I instead move the required items to child of an `allOf` it works as you expect.  Swagger-jaxrs is generating the former, where `requiredItems` is a combination of both the child and parent properties that lives on the outer properties of a composed model.

mtraynham:
@jimschubert @wing328 any chance you guys can take a look at this again?

mtraynham:
Ping @jimschubert @dr4ke616 @karismann @Zomzog @wing328

jimschubert:
@mtraynham Thanks for pinging again. The holidays were super hectic for me, sorry for the long delay.

I appreciate that you merged with latest :)

Issue #4268, opened 2019-10-25
[BUG][SPRING] Spring's validation behavior is different depending on how annotations are used 


lehphyro:
##### Description

Spring's behavior is different depending on whether `@Validated` is present on the class-level:
- If it's present, `ConstraintViolationException` is thrown if there are validation errors.
- If it's not present, `MethodArgumentNotValidException` is thrown if there are validation errors.

This affects the way validation errors are converted to response to clients because the conversion implementation needs to be completely different.

##### openapi-generator version

openapi-generator-maven-plugin:4.1.2

##### Command line used for generation

Plugin configuration used:

```
                <configuration>
                    <inputSpec>${project.basedir}/../../api/service-api/src/main/resources/openapi.yaml</inputSpec>
                    <output>${project.build.directory}/generated-sources/api</output>
                    <generateSupportingFiles>true</generateSupportingFiles>
                    <generatorName>spring</generatorName>
                    <strictSpec>true</strictSpec>
                    <configOptions>
                        <apiPackage>mypackage</apiPackage>
                        <modelPackage>mypackage</modelPackage>                        
                        <library>spring-mvc</library>
                        <dateLibrary>legacy</dateLibrary>
                        <interfaceOnly>true</interfaceOnly>
                        <skipDefaultInterface>true</skipDefaultInterface>
                        <useBeanValidation>true</useBeanValidation>
                        <useTags>true</useTags>
                        <java8>true</java8>
                        <useOptional>false</useOptional>
                    </configOptions>
                </configuration>

```
##### Steps to reproduce

Just enable or disable `useBeanValidation` to see that `@Validated` gets added to the Api interface.

##### Suggest a fix

Add a flag to indicate whether class-level `@Validated` should be added and make it `true` by default to maintain backwards compatibility.

lehphyro:
What needs to happen to have this issue fixed?

lehphyro:
@wing328 Could you please help me get this merged? It's open for 5 months now.

lehphyro:
@wing328 This is still scheduled to go into 5.0.0 right?

singulart:
Is it possible to upvote this PR somehow?

PR #4269, opened 2019-10-25
Use class level validation

lehphyro:
@diyfr 
Suggested fix for #4268, add a new parameter to choose which Spring validation behavior should be applied to the generated code.

### DIFF
--- a/docs/generators/spring.md
+++ b/docs/generators/spring.md
@@ -62,6 +62,7 @@ sidebar_label: spring
 |title|server title name or client service name| |OpenAPI Spring|
 |unhandledException|Declare operation methods to throw a generic exception and allow unhandled exceptions (useful for Spring `@ControllerAdvice` directives).| |false|
 |useBeanValidation|Use BeanValidation API annotations| |true|
+|useClassLevelBeanValidation|Add @Validated to class-level Api interfaces| |true|
 |useOptional|Use Optional container for optional parameters| |false|
 |useTags|use tags for creating interface and controller classnames| |false|
 |virtualService|Generates the virtual service. For more details refer - https://github.com/elan-venture/virtualan/wiki| |false|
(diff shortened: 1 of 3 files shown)


wing328:

wing328:
@lehphyro what about adding class-level validation without an additional option (which means it should be enabled by default)?

wing328:
cc @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04)

lehphyro:
> @lehphyro what about adding class-level validation without an additional option (which means it should be enabled by default)?

That's the case currently, `@Validated` is always added to the controller interface which means Spring always throws `ConstraintViolationException` when there are validation errors, the suggested change is to make that optional.

lehphyro:
@wing328 this PR is open for quite some time even though it's very small, can we please merge it?

lehphyro:
@wing328 Please, could you help me merge this? I'm maintaining a fork of the project just because of it...

lehphyro:
@diyfr @bkabrda Could you please have a look and help me merge this PR?

wing328:
Sorry for late reply on this. Please PM via Slack to discuss on how to proceed with this PR.

wing328:
@lehphyro as discussed in Slack, what about using customized templates with the `-t` option instead as the option only changes one line in the template?

wing328:
To elaborate a bit more, the spring generator has too many options and we prefer to avoid adding another option to change just one line in the template.

lehphyro:
As discussed in Slack, this option will allow users to select which Spring out-of-the-box behavior they'd like to use, so it's really about supporting Spring features.

The suggested change is to make spring throw `MethodArgumentNotValidException` instead of `ConstraintViolationException`, serialization of validation errors into a response body and amount of detail available change depending on which exception is thrown.

lehphyro:
Closed due to inactivity.

Issue #115, opened 2018-05-20
Vert.x Web Client & Server code generator


slinkydeveloper:
##### Description
Code generator both for client and server with Eclipse Vert.x

I'm working on it

PR #4286, opened 2019-10-26
Implement Server Stub Code for Vert.x Web #115

lwlee2608:
### PR checklist

### Description of the PR
Add server generation `java-vertx-web` which use [Vert.x-Web](https://vertx.io/docs/vertx-web/java/)

cc @wing328 @jimschubert @cbornet @ackintosh @jmini @etherealjoy


bin/java-vertx-web-rx-petstore-server.sh
```diff
@@ -0,0 +1,36 @@
...
+  mvn -B clean package
+fi
+
+SPEC="modules/openapi-generator/src/test/resources/3_0/petstore.yaml"
+GENERATOR="java-vertx-web"
+STUB_DIR="samples/server/petstore/java-vertx-web/rx"
+
+# if you've executed sbt assembly previously it will use that instead.
+export JAVA_OPTS="${JAVA_OPTS} -Xmx1024M -DloggerPath=conf/log4j.properties"
+ags="generate -i $SPEC -g $GENERATOR --artifact-id java-vertx-web-rx-server -o $STUB_DIR --additional-properties hideGenerationTimestamp=true $@"
```
wing328: Minor suggestion: adding `-t` with the resource (template) folder so that we don't need to recompile the whole project if we're only changing the templates.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ * Copyright 2018 SmartBear Software
```
wing328: @lwlee2608 I don't think we need line 3 as per https://opensource.stackexchange.com/questions/7300/copyright-notice-in-the-file-header-apache-v2-license/7301#7301

modules/openapi-generator/src/main/resources/JavaVertXWebServer/README.mustache
```diff
@@ -0,0 +1,14 @@
...
+Project generated on : {{generatedDate}}
+{{/hideGenerationTimestamp}}
+
+## Getting Started
+
+This document assumes you have maven available.
+
+To build the project using maven, run:
+```bash
+mvn package && java -jar target/target/java-vertx-web-rx-server-1.0.0-SNAPSHOT-fat.jar
```
wing328: Do we need to use {{artifactVersion}} instead of hard-coded `1.0.0-SNAPSHOT`?

wing328:
Tested locally and the result is good:
```
[INFO] skip non existing resourceDirectory /private/tmp/vertx2/src/test/resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.3:testCompile (default-testCompile) @ openapi-java-vertx-web-server ---
[INFO] No sources to compile
[INFO] 
[INFO] --- maven-surefire-plugin:2.12.4:test (default-test) @ openapi-java-vertx-web-server ---
[INFO] No tests to run.
[INFO] BUILD SUCCESS
[INFO] Total time: 1.732 s
[INFO] Finished at: 2019-10-27T12:29:13+08:00
```

wing328:
Please add a new entry for testing in https://github.com/OpenAPITools/openapi-generator/blob/master/pom.xml#L1166

If the output works with JDK7 as well, please also add an entry to https://github.com/OpenAPITools/openapi-generator/blob/master/pom.xml#L1205

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
```diff
@@ -0,0 +1,162 @@
...
+    public Map<String, Object> postProcessOperationsWithModels(Map<String, Object> objs, List<Object> allModels) {
+        Map<String, Object> newObjs = super.postProcessOperationsWithModels(objs, allModels);
+        Map<String, Object> operations = (Map<String, Object>) newObjs.get("operations");
+        if (operations != null) {
+            List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
+            for (CodegenOperation operation : ops) {
+                operation.httpMethod = operation.httpMethod.toLowerCase(Locale.ROOT);
+
+                if (operation.returnType == null) {
+                    operation.returnType = "Void";
```
wing328: Suggestion: you may want to use {{^returnType}} ... {{/returnType}} to identify void instead.

If you assign "Void" to returnType, there will be no easy way to detect void return in the mustache template.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    public CodegenOperation fromOperation(String path, String httpMethod, Operation operation, List<Server> servers) {
+        CodegenOperation codegenOperation =
+                super.fromOperation(path, httpMethod, operation, servers);
+
+        for (CodegenParameter codegenParameter : codegenOperation.allParams) {
+            if (codegenParameter.dataType.equals("File")) {
+                codegenParameter.dataType = "FileUpload";
+                codegenOperation.imports.remove("File");
+                codegenOperation.imports.add("FileUpload");
```
wing328: What about update the importMapping instead of overriding fromOperation?

lwlee2608: I think an even better solution is to override the typeMapping
 `typeMapping.put("file", "FileUpload");`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
lwlee2608: I was using `{{^returnType}}` in the beginning, but soon realised forcing all null return to Void makes the template file cleaner since all API will be wrapped in RxJava `Single<ApiResponse<{{returnType}}>>`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
lwlee2608: ok, done

bin/java-vertx-web-rx-petstore-server.sh
lwlee2608: i see, good idea.

modules/openapi-generator/src/main/resources/JavaVertXWebServer/README.mustache
lwlee2608: my mistake, done

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaVertXWebServerCodegen.java
wing328: Up to you. (later you may have a use case in which you need to determine whether the operation is returning anything and we can always find a solution/workaround later as this is not a day 1 requirement)

lwlee2608:
@wing328 all comments have been addressed. Please review

The server uses Java8 feature extensively, therefore does not work in JDK7 or lower.

wing328:
@lwlee2608 thanks for the new generator, which has been merged into master.

For your upcoming PRs, I would recommend creating a new branch as per git best practice.

wing328:
@lwlee2608 thanks for the PR, which has been included in v4.1.2 release: https://twitter.com/oas_generator/status/1189824932345069569

Issue #3473, opened 2019-07-26
[C#][csharp-netcore] Error responses do not correctly deserialise


rizwansaeed:
##### Description

If an API returns a response in the 400 range e.g. 404 along with a body, it attempts to deserialise into the target model object rather than `ApiException` resulting in a deserialisation error and loses the body of the response

##### openapi-generator version

4.0.3

##### OpenAPI declaration file content or url

https://fbn-prd.lusid.com/api/swagger/v0/swagger.json

##### Command line used for generation

java -jar openapi-generator-cli.jar generate -g csharp-netcore -c config.json

```config.json
{
  "packageName": "Lusid.Sdk",
  "sourceFolder": ".",
  "targetFramework": "netcoreapp2.0",
  "useDateTimeOffset": true,
  "netCoreProjectFile": true,
  "returnICollection": true,
  "packageVersion": "0.10.482"
}
```
##### Suggest a fix

The issue looks related to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/csharp-netcore/ApiClient.mustache#L126 but an `ApiException` would needed to be propagated back up the cll chain

PR #4373, opened 2019-11-05
[csharp-netcore] Propagate raw content to the ApiException error content.

yzhoholiev:
This PR resolves an issue with the ApiException error content described in #3473 .

The issue is linked to migration to the generic `IRestResponse<T>`. The deserialization happens before the exception is actually checked by `ExceptionFactory`.

Possible hotfix, for now, is to propagate both raw content from `IRestResponse.Content` as well as `IRestResponse<T>.Data` to the `ApiResponse`.

But I think it makes sense to roll back to the older implementation, where deserialization happed if an exception hadn't been constructed.


yzhoholiev:
@mandrean (2017/08), @jimschubert (2017/09) ❤️ @frankyjuang (2019/09)

PR #4381, opened 2019-11-05
[csharp-netcore] Propagate raw content to the ApiException error content.

yzhoholiev:
This PR resolves an issue with the ApiException error content described in #3473 .

The issue is linked to migration to the generic `IRestResponse<T>`. The deserialization happens before the exception is actually checked by `ExceptionFactory`.

Possible fix is to propagate both raw response content from `IRestResponse.Content` as well as `IRestResponse<T>.Data` to the `ApiResponse`.

But I think it makes sense to roll back to the older implementation, where deserialization happed if an exception hadn't been constructed.

Technical committee: @mandrean (2017/08), @jimschubert (2017/09) ❤️ @frankyjuang (2019/09)

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/ApiResponse.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/ApiResponse.mustache
@@ -128,12 +128,29 @@ namespace {{packageName}}.Client
         /// Initializes a new instance of the <see cref="ApiResponse{T}" /> class.
         /// </summary>
         /// <param name="statusCode">HTTP status code.</param>
+        /// <param name="headers">HTTP headers.</param>
         /// <param name="data">Data (parsed HTTP body)</param>
-        public ApiResponse(HttpStatusCode statusCode, T data, string rawContent)
+        public ApiResponse(HttpStatusCode statusCode, Multimap<string, string> headers, T data) : this(statusCode, headers, data, null)
         {
-            StatusCode = statusCode;
-            Data = data;
-            RawContent = rawContent;
+        }
+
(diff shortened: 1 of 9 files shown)


modules/openapi-generator/src/main/resources/csharp-netcore/ApiResponse.mustache
```diff
@@ -105,22 +115,25 @@ namespace {{packageName}}.Client
         /// <param name="statusCode">HTTP status code.</param>
         /// <param name="headers">HTTP headers.</param>
         /// <param name="data">Data (parsed HTTP body)</param>
-        public ApiResponse(HttpStatusCode statusCode, Multimap<string, string> headers, T data)
+        /// <param name="rawContent">Raw content.</param>
+        public ApiResponse(HttpStatusCode statusCode, Multimap<string, string> headers, T data, string rawContent)
```
jimschubert: These would probably be overloaded constructors to avoid breaking existing code (e.g. tests with stubs).

yzhoholiev: Added few overloads for binary compatibility.

yzhoholiev:
@jimschubert it looks like `continuous-integration/drone/pr` was broken because of some access issue. Could you help with a restart of this check?

rizwansaeed:
Hi all, thanks for fixing up this issue, do you know when it might get merged and released?

wing328:
@yzhoholiev  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

Issue #4342, opened 2019-10-31
[BUG][C++][Qt5] Object received via HTTP request callback have false isSet values


javorszkymazsi:
#### Bug Report Checklist

- [ x] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ x] What's the version of OpenAPI Generator used?
- [x ] Have you search for related issues/PRs?
- [ x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I needed a tool for accessing a C# REST API via a C++ application, and I used the openapi-generator to generate Qt code through which I could make HTTP requests to the server and get a response.

If an HTTP Post request is sent to the server and information bundled in a custom object is received in the callback, that object has all its variables filled with valuable information, that's good. But all variables have a corresponding `m_whatever_isSet` bool variable next to them and if that object came from a callback, those bool variables are false. It is no problem if you only need the data but if that object is used again for another HTTP request, all the variables with a false `m_whatever_isSet` value aren't considered. Currently, I'm using a workaround which is painful cpu usage-wise and reset all the variables with the values in them.

##### openapi-generator version

4.1.3

##### OpenAPI declaration file content or url

I did not use such thing. The HTTP request was handled in the C# REST API server.

##### Command line used for generation

`java -jar openapi-generator-cli-4.1.3.jar generate ^
-i http://localhost:5000/swagger/.../swagger.json ^
-g cpp-qt5-client ^
-o ./client ^
-DmodelTests=false ^
-DapiTests=false ^
-DpackageName="ApiClient" ^
-DoptionalProjectFile=true`

##### Steps to reproduce

1. Send HTTP request to the server which gives back a custom object
2. Use that object as it is in an another HTTP request.

##### Related issues/PRs

No.

##### Suggest a fix

Custom objects coming in from an HTTP request's response should have `m_whatever_isSet` as true where the values are filled out.

javorszkymazsi:
Thanks so much for fixing this so quick! :)

PR #4385, opened 2019-11-06
[C++] [Qt5]Update isSet when the object is received from callback.

etherealjoy:
### PR checklist

`Changes:`
- set flag updated when deserializing from response, enabling object reuse
- forward declaration in template functions in helper class due to templates calling one another
- change name of `init` function due to symbol collision with generated code

Fixes: #4342
`cc:`
@muttleyxd @MartinDelille @stkrwork @ravinikam


modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-header.mustache
```diff
@@ -128,7 +149,7 @@ namespace {{this}} {
     bool fromJsonValue(QList<T> &val, const QJsonValue &jval) {
         bool ok = true;
         if(jval.isArray()){
-            for(const auto& jitem : jval.toArray()){
+            for(const auto jitem : jval.toArray()){
```
muttleyxd: Is there a reason for removing reference here?

etherealjoy: The compiler gives a warning that the method `QJsonValue .toArray()` cannot return a reference. So I removed it.

muttleyxd: Looks good to me then!

DevASG:
The same problem is related to the C++ Qt Server.
Is it possible to **generate Objects for Qt Server in the same way**?
Thanks.

etherealjoy:
> The same problem is related to the C++ Qt Server.
> Is it possible to **generate Objects for Qt Server in the same way**?
> Thanks.

@DevASG I created #5634 for this.

PR #4387, opened 2019-11-06
[typescript-fetch] Add (de)serializers for oneOfs

eriktim:
### PR checklist

Using `oneOf`s with `typescript-fetch` results in union types, but lacks decoders and encoders for them. This PR patches that. I've also added support for the discriminator (resulting in tagged unions).

@wing328 Is there already a YAML available for testing oneOfs/allOfs/discriminators? Otherwise I would like to add a generic one.

cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig @topce @akehir

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/modelOneOf.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/modelOneOf.mustache
@@ -3,6 +3,10 @@ import {
     {{#imports}}
     {{{.}}},
     {{/imports}}
+    {{#oneOf}}
+    {{{.}}}FromJSONTyped,
+    {{{.}}}ToJSON,
+    {{/oneOf}}
 } from './';
 
 {{/hasImports}}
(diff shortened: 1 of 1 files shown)


wing328:
@eriktim  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

Bessonov:
@eriktim I'm not sure that I understand how it supposed to work without a discriminator. I've some sort of MongoDb-Api description:
```yaml
    DateSearch:
      oneOf:
      - type: object
        required:
        - $eq
        properties:
          $eq:
            type: string
            format: date
      - type: object
        required:
        - $ne
        properties:
          $ne:
            type: string
            format: date
[...]
```
The generated union type is correct:
```typescript
export type DateSearch = DateSearchOneOf | DateSearchOneOf1 | [...]
```
But `DateSearchToJSON` seems to be wrong (and doesn't compile):
```typescript
export function DateSearchToJSON(value?: DateSearch | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return { ...DateSearchOneOfToJSON(value), ...DateSearchOneOf1ToJSON(value), [...] };
}
```

From logical point of view, the last line try to merge all properies, which isn't what for an union type should happens.

Technically, `DateSearchOneOfToJSON` expects `DateSearchOneOf | null` and therefore doesn't compile.

From my point of view, a check should be generated, which, in sense of `oneOf`, try to determine the type of `DateSearch` and return result of correct `ToJSON` transformer.

Bessonov:
Ok, it's worser, I have no idea how to check in this case :dango:

eriktim:
@Bessonov I currently do not have tsc at hand so I cannot reproduce your error (it would be nice to post that one as well). It looks like TypeScript does not take the spread operator as I had hoped. As stated in my description I would like to have this case tested as well, hence this PR should not have been merged in yet. But even if I would have had, using a `oneOf` without discriminator is not a great choice for TypeScript. Other languages are more capable of dealing with this trial-and-error decoding. Is there any reason you cannot use a discriminator?

SimpleCookie:
Is this really resolved? I am not receiving the problem
```java
@Data
@SuperBuilder(toBuilder = true)
@EqualsAndHashCode(callSuper = false)
@Schema(
    type = "object",
    oneOf = { First.class, Second.class }
)
public abstract class Foo {
    @Schema(required = true)
    private String type;
}

@Data
@EqualsAndHashCode(callSuper = true)
@SuperBuilder(toBuilder = true)
@Schema(type = "object")
@ApiModel(value = "Bar", parent = Foo.class)
public abstract class Bar {
    @Schema(required = true)
    private String id;
}
```
> error TS2305: Module '"."' has no exported member 'FooFromJSON'.
> error TS2305: Module '"."' has no exported member 'FooToJSON'.

_The example above might be slightly off, I cannot copy my code directly due to NDA._
Or am I doing something wrong?

Issue #4363, opened 2019-11-04
[BUG][Ruby] Set NULL is not possible


slemrmartin:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used? 
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

In **Ruby client**, it's not possible to set NULL on any value (I suppose it should be "nullable": true in yaml definition)

##### openapi-generator version

3.3.4, 4.0.0, 4.1.1, 4.2.0

##### OpenAPI declaration file content or url

https://github.com/ManageIQ/topological_inventory-ingress_api/blob/12b92fbdb9485d01b7840318dc42a384f557286a/public/doc/openapi-3-v0.0.2.json#L4243
https://github.com/ManageIQ/topological_inventory-ingress_api/blob/12b92fbdb9485d01b7840318dc42a384f557286a/public/doc/openapi-3-v0.0.2.json#L4252
(2nd case: Is it possible to set nullable to `ref`?)

##### Command line used for generation

```
java -jar openapi-generator-cli-3.4.4.jar generate -i public/doc/opeanpi-3-v0.0.2.json.generator.yaml -c .openapi_generator_config.json -g ruby -o topological_inventory-ingress_api-client-ruby
```
https://github.com/ManageIQ/topological_inventory-ingress_api/blob/master/.openapi_generator_config.json

with version 4.0.0 validator generates 100+ errors, version 4.1+ must be run with `--skip-validate-spec`

##### Steps to reproduce

`TopologicalInventoryIngressApiClient::ServiceOffering.new(:source_ref => '123', :name => 'Sample',...).to_hash`
`TopologicalInventoryIngressApiClient::ServiceOffering.new(:source_ref => '123', :name => nil,...).to_hash`

##### Suggest a fix

Following line causes that keys with `nil` values are deleted, so it's not possible to recognize by server if they're not set or set to nil.

[...]

PR #4391, opened 2019-11-06
Ruby client nullable

slemrmartin:
This PR fixes Ruby clients on columns with "nullable": true. Attribute's key with nil value isn't skipped, but sent to server with nil value.

**Issue**: https://github.com/OpenAPITools/openapi-generator/issues/4363#issuecomment-550199932

### PR checklist


slemrmartin:
@cliffano @zlx @autopp please review

slemrmartin:
cc @wing328

autopp:
@slemrmartin Thanks for PR, I'll review it.

slemrmartin:
@autopp If nullable attribute is ommited, then `!self.class.openapi_nullable.include?(attr)` condition is true, isn't it? Which means that null isn't sent to server. Or which case I'm missing?

autopp:
@slemrmartin Sorry for the incomprehension.
Let me illustrate my concern with the following model example.

```yaml
Foo:
  type: object
  properties:
    x:
      type: integer
      nullable: true
```

On current master:
```ruby
p OpenapiClient::Foo.new({ x: 1 }).to_hash   #=> {:x=>1}
p OpenapiClient::Foo.new({ x: nil }).to_hash #=> {} <- always empty
p OpenapiClient::Foo.new({}).to_hash         #=> {}
```

And on this branch:
```ruby
p OpenapiClient::Foo.new({ x: 1 }).to_hash   #=> {:x=>1}
p OpenapiClient::Foo.new({ x: nil }) #=> {:x=>nil}
p OpenapiClient::Foo.new({})         #=> {:x=>nil} <- always nil
```

For these, I think the following are expected:
```ruby
p OpenapiClient::Foo.new({ x: 1 })   #=> {:x=>1}
p OpenapiClient::Foo.new({ x: nil }) #=> {:x=>nil}
p OpenapiClient::Foo.new({})         #=> {}
```

Please let me know if I misunderstand.

slemrmartin:
@autopp now I understand, thanks for clarification. 
Then it should be combination of `openapi_nullable` and `instance_variable_defined?` I think, because if `nullable` is not defined or false in definition, then variable `x` set to nil shouldn't be allowed
```
    x:
      type: integer
      nullable: true
```
```
p OpenapiClient::Foo.new({ x: nil }) #=> {:x=>nil}
p OpenapiClient::Foo.new({ }) #=> {}
```
, but
```
    x:
      type: integer
      nullable: false
```
```
p OpenapiClient::Foo.new({ x: nil }) #=> {}
p OpenapiClient::Foo.new({ }) #=> {}
```

I don't know openapi specification to the depth so correct me if I'm wrong

modules/openapi-generator/src/main/resources/ruby-client/partial_model_generic.mustache
```diff
@@ -51,6 +51,17 @@
       }
     end
 
+    # List of attributes with nullable: true
+    def self.openapi_nullable
+      [
+        {{#vars}}
+        {{#isNullable}}
+          :'{{{name}}}'{{#hasMore}},{{/hasMore}}
```
autopp: indent seems to be too depth.

autopp:
@slemrmartin your suggestion sounds good 👍 
Can you fix it with a comment about the indentation?

modules/openapi-generator/src/main/resources/ruby-client/partial_model_generic.mustache
wing328: And what about defining it as a hash instead similar to self.openapi_types?

If an object contains a lot of nullable properties, using `include?` may result in performance issues.

slemrmartin: I don't see usage of key-value, but what about using `Set` instead of `Array`, it should have better performance for include?

slemrmartin: I don't like the format 
```
def self.openapi_nullable
      Set.new([
      ])
end
```
is it possible to define it in template in a way that empty set won't generate new line?

autopp: Hmm... it seems to be difficult because there is no way to directly enumerate nullable attributes in template.
@wing328 Do you have any ideas?

By the way, I don't like this format, but it's not acceptable (Rubocop also allows this), so I think this PR can be merged.
Isn't it okay to make another PR for reformatting?

wing328: Set is definitely better than array in this case.

About empty set, I think it's acceptable for day 1 requirement. We can always improve it later with another PR.

wing328:
@slemrmartin  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

slemrmartin:
@wing328 thanks for you help

PR #4407, opened 2019-11-07
[REQ][typescript-angular]: support for object as query parameters

### LINKED ISSUE #4404: [REQ] [typescript-angular]: support for object as query parameters

At present, typescript-angular only supports arrays and simple types (string, 
number, Date) as query parameters. However, when you use openapi-generator in 
conjuction with springdoc and (java) [spring](https://spring.io/), you wish 
to have support for **objects** as query parameters as well.

In spring-data, there is REST support for an idea called 
[`Pageable`](https://docs.spring.io/spring-data/rest/docs/current/reference/html/#paging-and-sorting.sorting). This would 
be the openapi schema for it:

```json
      "Pageable": {
        "nullable": false,
        "type": "object",
        "required": [
          "size",
          "page"
        ],
        "properties": {
          "size": {
            "format": "int32",
            "type": "integer"
          },
          "sort": {"$ref": "#/components/schemas/Sort"},
          "page": {
            "format": "int32",
            "type": "integer"
          }
        }
      },
      "Sort": {
        "nullable": false,
        "type": "array",
        "items": {"type": "string"}
      },
```

The `items` of Sort are the names of the properties to sort the data with. 
Optionally, there could be a sort direction (ASC, DESC) appended like in 
`ansatz,DESC`. The comma (,) should _not be urlencoded_.

What I want is sane support for `Pageable` and other objects.

Lets think about a (GET) REST endpoint that takes an `Pageable` and a map 
with parameters as query parameters. `openapi-generator` will emits something
like:

```typescript
    public getAllAnsatz(pageable: Pageable, map: { [key: string]: Array<string>; }, observe?: 'body', reportProgress?: boolean): Observable<Array<AnsatzDTO>>;
    public getAllAnsatz(pageable: Pageable, map: { [key: string]: Array<string>; }, observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<Array<AnsatzDTO>>>;
    public getAllAnsatz(pageable: Pageable, map: { [key: string]: Array<string>; }, observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<Array<AnsatzDTO>>>;
    public getAllAnsatz(pageable: Pageable, map: { [key: string]: Array<string>; }, observe: any = 'body', reportProgress: boolean = false ): Observable<any> {
        if (pageable === null || pageable === undefined) {
            throw new Error('Required parameter pageable was null or undefined when calling getAllAnsatzfreigabes.');
        }
        if (map === null || map === undefined) {
            throw new Error('Required parameter map was null or undefined when calling getAllAnsatzfreigabes.');
        }

        let queryParameters = new HttpParams({encoder: this.encoder});
        if (pageable !== undefined && pageable !== null) {
            queryParameters = queryParameters.set('pageable', <any>pageable);
        }
        if (map !== undefined && map !== null) {
            queryParameters = queryParameters.set('map', <any>map);
        }
``` 

Certainly, this will not lead to the desired result, as implicit `toString` is called on `pageable`
and `map`.

The idea is to have a method that will act according to the dynamic type of the variable 
that should be added to `queryParameters`:

```typescript
        let queryParameters = new HttpParams({encoder: this.encoder});
        if (pageable !== undefined && pageable !== null) {
            queryParameters = this.addToHttpParams(queryParameters, <any>pageable, 'pageable')
        }
        if (map !== undefined && map !== null) {
            queryParameters = this.addToHttpParams(queryParameters, <any>map, 'map')
        }
```

## Proposed implementation of `addToHttpParams`

```typescript
    private addToHttpParams(httpParams: HttpParams, value: any, key?: string): HttpParams {
        if (typeof value === "object") {
            httpParams = this.addToHttpParams1(httpParams, value);
        } else {
            httpParams = this.addToHttpParams1(httpParams, value, key);
        }
        return httpParams;
    }

    private addToHttpParams1(httpParams: HttpParams, value: any, key?: string): HttpParams {
        if (typeof value === "object") {
            if (this.isArrayLike(value)) {
                (value as []).forEach( elem => httpParams = this.addToHttpParams1(httpParams, elem, key));
            } else if (value instanceof Date) {
                if (key != null) {
                    httpParams = httpParams.append(key,
                        (value as Date).toISOString().substr(0, 10));
                } else {
                   throw Error("key may not be null if value is Date");
                }
            } else {
                Object.keys(value).forEach( k => httpParams = this.addToHttpParams1(
                    httpParams, value[k], key != null ? `${key}.${k}` : k));
            }
        } else if (key != null) {
            httpParams = httpParams.append(key, value);
        } else {
            throw Error("key may not be null if value is not object or array");
        }
        return httpParams;
    }

    private isArrayLike(obj: any): boolean {
        return (Array.isArray(obj) ||
            typeof obj === "object" &&
            typeof (obj.length) === "number" &&
            (obj.length === 0 ||
                (obj.length > 0 && (obj.length - 1) in obj))
        );
  }
```

The implementation will flatten the top level `object` to properties, but nested
are added with `.` Syntax to the key (like in 
`toplevelProperty.propertyOfNestedObject`). Arrays and simple types are treated 
like before (no change). The advantage of the new solution is that REST endpoints
from spring-data are supported out of the box.

aanno2:
fixes #4404

@macjohnny: Thank you for review

I've updated the PR and hope it will be more in-line with what is expected.

## PR checklist

- [x] Read the contribution guidelines.
- [x] If contributing template-only or documentation-only changes which will change sample output, build the project before.
- [x] Run the shell script(s) under ./bin/ (or Windows batch scripts under.\bin\windows) to update Petstore samples related to your fix. This is important, as CI jobs will verify all generator outputs of your HEAD commit, and these must match the expectations made by your contribution. You only need to run ./bin/{LANG}-petstore.sh, ./bin/openapi3/{LANG}-petstore.sh if updating the code or mustache templates for a language ({LANG}) (e.g. php, ruby, python, etc).
- [x] File the PR against the correct branch: master, 4.3.x, 5.0.x. Default: master.
- [x] Copy the technical committee to review the pull request if your PR is targeting a particular programming language.

@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig @topce @akehir 

issue by @aanno2

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -244,7 +244,7 @@ export class {{classname}} {
         {{/isCollectionFormatMulti}}
         {{^isCollectionFormatMulti}}
             {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
-                {{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}'])), '{{baseName}}');
+                {{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}']), '{{baseName}}');
         {{/isCollectionFormatMulti}}
         }
         {{/isListContainer}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -151,6 +151,47 @@ export class {{classname}} {
 {{/operation}}
 {{/useHttpClient}}
 
+    private addToHttpParams(httpParams: HttpParams, value: any, key?: string): HttpParams {
+        if (typeof value === "object") {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParamsRecursive(httpParams, value);
+        } else {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParamsRecursive(httpParams, value, key);
+        }
+        return httpParams;
+    }
+
+    private addToHttpParamsRecursive(httpParams: HttpParams, value: any, key?: string): HttpParams {
+        if (typeof value === "object") {
+            if (this.isArrayLike(value)) {
+                (value as []).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
+            } else if (value instanceof Date) {
+                if (key != null) {
+                    {{#useHttpClient}}httpParams = {{/useHttpClient}}httpParams.append(key,
+                        (value as Date).toISOString(){{^isDateTime}}.substr(0, 10)){{/isDateTime}};
+                } else {
+                   throw Error("key may not be null if value is Date");
+                }
+            } else {
+                Object.keys(value).forEach( k => {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParamsRecursive(
+                    httpParams, value[k], key != null ? `${key}.${k}` : k));
+            }
+        } else if (key != null) {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}httpParams.append(key, value);
+        } else {
+            throw Error("key may not be null if value is not object or array");
+        }
+        return httpParams;
+    }
+
+    private isArrayLike(obj: any): boolean {
+        return (Array.isArray(obj) ||
+            typeof obj === "object" &&
+            typeof (obj.length) === "number" &&
+            (obj.length === 0 ||
+                (obj.length > 0 && (obj.length - 1) in obj))
+        );
+  }
+
 {{#operation}}
     /**
 {{#summary}}
@@ -206,11 +247,13 @@ export class {{classname}} {
         if ({{paramName}}) {
         {{#isCollectionFormatMulti}}
             {{paramName}}.forEach((element) => {
-                {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.append('{{baseName}}', <any>element);
+                {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
+                  <any>element, '{{baseName}}');
             })
         {{/isCollectionFormatMulti}}
         {{^isCollectionFormatMulti}}
-            {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.set('{{baseName}}', {{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}']));
+            {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
+              <any>{{paramName}}, '{{baseName}}');
         {{/isCollectionFormatMulti}}
         }
         {{/isListContainer}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -244,7 +244,7 @@ export class {{classname}} {
         {{/isCollectionFormatMulti}}
         {{^isCollectionFormatMulti}}
             {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
-              <any>{{paramName}}, '{{baseName}}');
+                {{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}'])), '{{baseName}}');
         {{/isCollectionFormatMulti}}
         }
         {{/isListContainer}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -250,12 +250,8 @@ export class {{classname}} {
         {{/isListContainer}}
         {{^isListContainer}}
         if ({{paramName}} !== undefined && {{paramName}} !== null) {
-        {{#isDateTime}}
-            {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.set('{{baseName}}', ({{paramName}} as any instanceof Date) ? ({{paramName}} as any).toISOString(): {{paramName}});
-        {{/isDateTime}}
-        {{^isDateTime}}
-            {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.set('{{baseName}}', <any>{{paramName}});
-        {{/isDateTime}}
+          {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
+            <any>{{paramName}}, '{{baseName}}');
         }
         {{/isListContainer}}
 {{/queryParams}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -163,7 +163,7 @@ export class {{classname}} {
     private addToHttpParamsRecursive(httpParams: HttpParams, value: any, key?: string): HttpParams {
         if (typeof value === "object") {
             if (Array.isArray(value)) {
-                (value as []).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
+                (value as any[]).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
             } else if (value instanceof Date) {
                 if (key != null) {
                     {{#useHttpClient}}httpParams = {{/useHttpClient}}httpParams.append(key,
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -162,7 +162,7 @@ export class {{classname}} {
 
     private addToHttpParamsRecursive(httpParams: HttpParams, value: any, key?: string): HttpParams {
         if (typeof value === "object") {
-            if (this.isArrayLike(value)) {
+            if (Array.isArray(value)) {
                 (value as []).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
             } else if (value instanceof Date) {
                 if (key != null) {
@@ -182,15 +182,6 @@ export class {{classname}} {
         }
         return httpParams;
     }
-
-    private isArrayLike(obj: any): boolean {
-        return (Array.isArray(obj) ||
-            typeof obj === "object" &&
-            typeof (obj.length) === "number" &&
-            (obj.length === 0 ||
-                (obj.length > 0 && (obj.length - 1) in obj))
-        );
-  }
 
 {{#operation}}
     /**
--- a/samples/client/petstore/typescript-angular-v8-provided-in-root/builds/with-npm/package.json
+++ b/samples/client/petstore/typescript-angular-v8-provided-in-root/builds/with-npm/package.json
@@ -1,38 +1,21 @@
 {
-  "name": "@openapitools/typescript-angular-petstore",
+  "name": "default",
   "version": "1.0.0",
-  "description": "OpenAPI client for @openapitools/typescript-angular-petstore",
-  "author": "OpenAPI-Generator Contributors",
-  "keywords": [
-    "openapi-client",
-    "openapi-generator"
-  ],
-  "license": "Unlicense",
+  "description": "### Building",
+  "main": "index.js",
   "scripts": {
-    "build": "ng-packagr -p ng-package.json"
+    "test": "echo \"Error: no test specified\" && exit 1"
   },
-  "peerDependencies": {
-    "@angular/core": "^8.0.0",
-    "@angular/common": "^8.0.0",
-    "@angular/compiler": "^8.0.0",
-    "core-js": "^2.4.0",
-    "reflect-metadata": "^0.1.3",
-    "rxjs": "^6.5.0"
-  },
+  "author": "",
+  "license": "ISC",
   "devDependencies": {
-    "@angular/common": "^8.0.0",
-    "@angular/compiler": "^8.0.0",
-    "@angular/compiler-cli": "^8.0.0",
-    "@angular/core": "^8.0.0",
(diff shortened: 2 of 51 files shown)


macjohnny:
Thanks for your PR
Please follow the steps described in https://github.com/OpenAPITools/openapi-generator/blob/master/.github/PULL_REQUEST_TEMPLATE.md

modules/openapi-generator/src/main/resources/typescript-angular/encoder.mustache
```diff
@@ -15,7 +15,8 @@ export class CustomHttpParameterCodec implements HttpParameterCodec {
         return encodeURIComponent(k);
     }
     encodeValue(v: string): string {
-        return encodeURIComponent(v);
+        return encodeURIComponent(v)
+          .replace(/%2C(ASC|DESC)$/i, ",$1");
```
macjohnny: I dont think this should be here

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -130,6 +130,53 @@ export class {{classname}} {
...
+    {
+        if (typeof value === "object") {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParams1(httpParams, value);
+        } else {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParams1(httpParams, value, key);
+        }
+        return httpParams;
+    }
+
+    private addToHttpParams1(
```
macjohnny: How could we improve the naming of this method?

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -130,6 +130,53 @@ export class {{classname}} {
...
+            }
+        } else if (key != null) {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}httpParams.append(key, value);
+        } else {
+            throw Error("key may not be null if value is not object or array");
+        }
+        return httpParams;
+    }
+
+    private isArrayLike(obj: any): boolean {
```
macjohnny: Type should be object

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -130,6 +130,53 @@ export class {{classname}} {
...
+    private addToHttpParams1(
+        httpParams: {{#useHttpClient}}HttpParams{{/useHttpClient}}{{^useHttpClient}}URLSearchParams{{/useHttpClient}},
+        value: any, key?: string): {{#useHttpClient}}HttpParams{{/useHttpClient}}{{^useHttpClient}}URLSearchParams{{/useHttpClient}}
+    {
+        if (typeof value === "object") {
+            if (this.isArrayLike(value)) {
+                (value as []).forEach( elem => httpParams = this.addToHttpParams1(httpParams, elem, key));
+            } else if (value instanceof Date) {
+                if (key != null) {
+                    {{#useHttpClient}}httpParams = {{/useHttpClient}}httpParams.append(key,
```
macjohnny: Might also be a date time

modules/openapi-generator/src/main/resources/typescript-angular/encoder.mustache
aanno2: Well, see #4404: Optionally, there could be a sort direction (ASC, DESC) appended like in
ansatz,DESC. The comma (,) should not be urlencoded.

It is for this special case (but it's a bit hacky - should I obmit it?)

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
aanno2: No, the method should work for anything given to it (as it does now).

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
aanno2: Sorry, I did not understand. Handling for Date is the same as before. I'm not aware of a JS Datetime standard object...

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
aanno2: Well, this is our (project) standard naming for this kind of recursion. What about `addToHttpParamsRec`?

aanno2:
> Thanks for your PR
> Please follow the steps described in https://github.com/OpenAPITools/openapi-generator/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Well, I hope that I have not done something outright wrong. Not really to sure about the branch to target. And I have no idea what to do for 'Copy the technical committee'.

BTW, travis dies with 'A tuple type element list cannot be empty' but my local build with maven has not reported a problem. I'm not sure how to reproduce the error locally.

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
macjohnny: in the typescript-angular generator the `type: string; format: date-time` maps to the type `Date` in javascript, while `type: string; format: date` maps to `string` by default.
So here something like `{{^isDateTime}}.substr(0, 10){{isDateTime}}` should help

modules/openapi-generator/src/main/resources/typescript-angular/encoder.mustache
macjohnny: I think this should not be done here

aanno2: Hhm, I have no idea where else to put this. Hence I reverted the change...

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
aanno2: I see now and changed it. Thank you!

aanno2:
I have updated and merge-squashed my PR with an appropriate author email. No 'real' changes.

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -130,6 +130,47 @@ export class {{classname}} {
...
+        } else {
+            throw Error("key may not be null if value is not object or array");
+        }
+        return httpParams;
+    }
+
+    private isArrayLike(obj: any): boolean {
+        return (Array.isArray(obj) ||
+            typeof obj === "object" &&
+            typeof (obj.length) === "number" &&
```
macjohnny: I think an actual array should be required, otherwise, the following object is also recognized as array-like, but throws an error due to the `.forEach` missing:
```typescript
{
  length: 5
  4: 'something'
}
```

aanno2: My feeling is that most JS is written with 'duck typing' in mind. Your example object _is looking for trouble_ because (duck typing aware) code could think that the object is an array. But my `isArrayLike` method will _not_ be fooled because it also checks `(obj.length - 1) in obj`. It will be fooled only with `{ length: 1, 0: 'something'}`.

I'm using here code from https://stackoverflow.com/questions/24048547/checking-if-an-object-is-array-like . In principle there are 2 option: only accepting real arrays or accept array-like objects.

If you search for 'array-like', you will find all kind of opinions to the subject. AFAIK, there are some standard objects that should be treated as arrays but aren't 'real' arrays.

macjohnny: i think it should only check for `Array.isArray()`, any other array would need to be converted first
I verified my example above in the browser:
```javascript
isArrayLike({length: 5, 4: 'asdf'})
> true
```
the code should be robust for any kind of input object

aanno2: * Nope. I tried it in Firefox and it works like I said.
* There are even modules (like https://www.npmjs.com/package/@gyumeijie/is-array-like) that works like I described.
* **However, I'm easy with the Array check and will change the code**

aanno2:
@macjohnny: Thank you for accepting my contribution.

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -171,22 +203,20 @@ export class {{classname}} {
...
         {{#isCollectionFormatMulti}}
             {{paramName}}.forEach((element) => {
-                {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.append('{{baseName}}', <any>element);
+                {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
+                  <any>element, '{{baseName}}');
             })
         {{/isCollectionFormatMulti}}
         {{^isCollectionFormatMulti}}
-            {{#useHttpClient}}queryParameters = {{/useHttpClient}}queryParameters.set('{{baseName}}', {{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}']));
+            {{#useHttpClient}}queryParameters = {{/useHttpClient}}this.addToHttpParams(queryParameters,
```
macjohnny: I just noticed that this is a breaking change...
@wing328 @akehir can we still include it?

macjohnny: or would there be a way to avoid a breaking change?

aanno2: Hhm, just to understand this: The generator is supporting OpenApi v2 (Swagger) _and_ v3? (Because according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md collectionFormat is only in v2.)

The meaning of the (old) line is that the given array is joined with some separator and then treated as string? If this is the case, it is possible leaf this as before. The patch is only about allowing (js) objects as query parameters.

macjohnny: I think it should be fine now

macjohnny:
please merge the current master into your branch

samples/client/petstore/typescript-angular-v8-provided-in-root/builds/with-npm/package.json
```diff
@@ -1,38 +1,21 @@
 {
-  "name": "@openapitools/typescript-angular-petstore",
+  "name": "default",
   "version": "1.0.0",
-  "description": "OpenAPI client for @openapitools/typescript-angular-petstore",
```
macjohnny: I think this should be reverted

aanno2: Oops. Not done on purpose. Sorry!

macjohnny:
@aanno2 can you please merge the current master and update the samples?

dougal83:
@aanno2 Thanks for this pull request.  Just run into the issue myself.  All set for next release?

greghopkins:
Would love for this to get merged soon... 🙏 😉

wing328:
@aanno2 please resolve the merge conflicts and merge the latest master into this branch. Let me know if you need help on this.

aanno2:
@wing328: As requested, I have rebased the PR onto the current _master_ branch.

macjohnny:
@aanno2 can you please fix the build error:
```
ERROR in builds/default/api/pet.service.ts(78,27): error TS1122: A tuple type element list cannot be empty.
builds/default/api/store.service.ts(64,27): error TS1122: A tuple type element list cannot be empty.
builds/default/api/user.service.ts(64,27): error TS1122: A tuple type element list cannot be empty.
```

see https://travis-ci.org/OpenAPITools/openapi-generator/builds/640473532?utm_source=github_status&utm_medium=notification

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -151,6 +151,38 @@ export class {{classname}} {
...
+        } else {
+            {{#useHttpClient}}httpParams = {{/useHttpClient}}this.addToHttpParamsRecursive(httpParams, value, key);
+        }
+        return httpParams;
+    }
+
+    private addToHttpParamsRecursive(httpParams: HttpParams, value: any, key?: string): HttpParams {
+        if (typeof value === "object") {
+            if (Array.isArray(value)) {
+                (value as []).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
```
macjohnny:
```suggestion
                (value as any[]).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
```

macjohnny:
please re-generate the samples

kenisteward:
@macjohnny. Has this been released in a version yet?  I was legit looking into fixing this but found this or and am super happy haha

dougal83:
> @macjohnny. Has this been released in a version yet? I was legit looking into fixing this but found this or and am super happy haha

Yes, it is in the latest version.

spencerfontein:
looks like this was merged to 4.3.0, but the latest available on npm is 4.2.3 https://www.npmjs.com/package/@openapitools/openapi-generator-cli. Is there a way I can use the fix ?

dougal83:
> looks like this was merged to 4.3.0, but the latest available on npm is 4.2.3 https://www.npmjs.com/package/@openapitools/openapi-generator-cli. Is there a way I can use the fix ?

Are you sure?  I thought that the milestone on the right was an at-a-glance indicator and it says 4.2.3.  Also, if I recall correctly then I've used this feature from the current npm version.  Where are you seeing 4.3.0?

spencerfontein:
I clicked on this link https://github.com/OpenAPITools/openapi-generator/issues/5174 that daniel-frank posted a few comments back and it mentioned 4.3.

I am currently using 4.2.3

I have the Pagination object in my Java code as a query param and this is the TS I get


```
import { Sort } from'./sort';
exportinterfacePageable {
offset?: number;
sort?: Sort;

pageSize?: number;
pageNumber?: number;
paged?: boolean;
unpaged?: boolean;
}
```

```
export interface Sort {
    sorted?: boolean;
    unsorted?: boolean;
    empty?: boolean;
}
```


Maybe I'm missing something 

This is the command I'm running

`openapi-generator generate -g typescript-angular -i ../../backend/escalation-service/target/openapi.json -o swagger/escalation-api --additional-properties npmName=@mycustom/escalation-api,ngVersion=8.1.2,npmVersion=1.0.0`

Thanks for the help

dougal83:
I think it would be best to open a new issue to query this and include your 'openapi.json'.  Hopefully, someone can point out the issue.

PR #4412, opened 2019-11-07
[Core] Add warning on headers with underscores

### LINKED ISSUE #4217: [REQ] Add warning on headers with underscores

Already discussed a bit in #3658. Quote:

> _Apache and Nginx webservers avoid underscores in HTTP headers, so `api_key` header received as `api-key`. Refs: [Cannot retrieve all request headers using PHP slim framework](https://stackoverflow.com/questions/36316454/cannot-retrieve-all-request-headers-using-php-slim-framework) and [Why underscores are forbidden in HTTP header names](https://stackoverflow.com/questions/22856136/why-underscores-are-forbidden-in-http-header-names/22856867)._

Maybe we should add warnings when headers contains underscores. I would add this to `DefaultCodegen`. I've read somewhere that headers with underscores can cause server security problems.

We can add generator option to disable this warnings. This might be useful when user already configured Nginx/Apache to allow headers with underscores.

## Additional context

ybelenko:
### PR checklist

Closes #4217

I didn't add any tests, because underscore search is pretty basic in current PR. I can add tests if necessary.

Examples of new warnings:
```
[main] WARN  o.o.codegen.DefaultCodegen - Security schema 'api_key' mapped to 
underscored header 'api_key' which is forbidden in Apache and Nginx webservers 
by default. 
https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names/22856867#22856867
```
```
[main] WARN  o.o.codegen.DefaultCodegen - Request header parameter name 
'enum_header_string_array' contains underscore which is forbidden in Apache 
and Nginx webservers by default. 
https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names/22856867#22856867
```
```
[main] WARN  o.o.codegen.DefaultCodegen - Response header name 'X_Rate_Limit' 
contains underscore which is forbidden in Apache and Nginx webservers by 
default. 
https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names/22856867#22856867
```

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -489,6 +489,28 @@ public class DefaultCodegen implements CodegenConfig {
     // override with any special post-processing
     @SuppressWarnings("static-method")
     public Map<String, Object> postProcessOperationsWithModels(Map<String, Object> objs, List<Object> allModels) {
+        Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
+        List<CodegenOperation> operationList = (List<CodegenOperation>) operations.get("operation");
+        for (CodegenOperation op : operationList) {
+            if (op.getHasHeaderParams()) {
+                List<CodegenParameter> headerParams = (List<CodegenParameter>) op.headerParams;
+                for (CodegenParameter param : headerParams) {
+                    String paramBaseName = (String) param.baseName;
+                    if (paramBaseName.contains("_")) {
+                        LOGGER.warn("Request header parameter name '" + paramBaseName + "' contains underscore which is forbidden in Apache and Nginx webservers by default. https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names/22856867#22856867");
+                    }
+                }
+            }
+            if (op.getHasResponseHeaders()) {
+                List<CodegenProperty> responseHeaders = (List<CodegenProperty>) op.responseHeaders;
+                for (CodegenProperty prop : responseHeaders) {
+                    String propBaseName = (String) prop.baseName;
+                    if (propBaseName.contains("_")) {
+                        LOGGER.warn("Response header name '" + propBaseName + "' contains underscore which is forbidden in Apache and Nginx webservers by default. https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names/22856867#22856867");
+                    }
+                }
+            }
+        }
         return objs;
     }
 
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -489,28 +489,6 @@ public class DefaultCodegen implements CodegenConfig {
     // override with any special post-processing
     @SuppressWarnings("static-method")
     public Map<String, Object> postProcessOperationsWithModels(Map<String, Object> objs, List<Object> allModels) {
-        Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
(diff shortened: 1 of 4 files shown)


wing328:
Thanks for the PR. Instead of adding the warning in the default codegen, what about adding it to the PHP Slim 4 generator to start with?

If I'm just using the documentation generator, I don't think the warning is relevant to me.

ybelenko:
@wing328 Hmmm, maybe you right. I've chosen default codegen because it's an api design problem, not framework. Wanted to hear other opinions, but your argument seems reasonable.

I would add it to all server generators, but there is no `AbstractServerCodegen`.

What do you think about log message? Do you like it? Maybe I should remove **stackoverflow** link from it?

ybelenko:
Warning triggers only with server generators from now.

jimschubert:
I think this would be better at spec validation, since users can skip those warnings. This is excellent quality of life improvement for the community.

ybelenko:
@jimschubert Do we have our own spec validator module/package? I used `swagger-cli` npm package for validation before. Btw I've never seen a warning from `swagger-cli` validator before, only errors.

jimschubert:
@ybelenko our validations are done and hoc in the generate and validate tasks. See for example:

https://github.com/OpenAPITools/openapi-generator/blob/de162f7f34e6ac2bf91b289fd7c6ae9f82e03f4b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Validate.java#L54

I created a simple validation framework in #3183 which I plan to consume in work done in the PR linked in that PR. I think I can move your validations when I do the validation extraction.

jimschubert:
Mind if we remove the security label? We should only use this for issues related to security of using the tool (e.g. CVE for Jackson Databind). This isn't related to issues with authorization defined in spec.

ybelenko:
@jimschubert So, what with this PR? Should I close it without merge?

Btw, personally me, I usually skip warnings because they looks like debug logs right now. There is no color difference between `info` and `warn` in terminal output. Can we enhance logging to use colors?

jimschubert:
@ybelenko Just reviewed the PR more in depth…

Would it be possible to add a flag and display this message only once? Consider if someone has 100 endpoints with 3-4 headers per endpoint, all following this format which is valid according to RFC. That would be a ton of noise in the console, and would be frustrating for anyone _not_ deploying to Apache/Nginx (e.g. standalone Spring Boot server). Other than that concern, I think this looks great and I appreciate the detailed message.

> There is no color difference between info and warn in terminal output. Can we enhance logging to use colors?

I think this would be somewhat low priority, but definitely something we could evaluate.

ybelenko:
> That would be a ton of noise in the console, and would be frustrating for anyone not deploying to Apache/Nginx

I would add an option to turn this validation off completely. I filed this issue because I encountered this problem twice(every time like the first time, I didn't remember about Apache specifics and been frustrated). This PR heals my personal pain :smile:

The other thing that drives me crazy is that Apache doesn't allow dashes or underscores in domain names... well I already forget which one, see.

jimschubert:
@ybelenko following up, what about William's suggestion to move this to PHP generators only, or my suggestion to move this out of code which will run for every operation to something like the custom validation?  As the PR stands at the moment I think this will add too much noise and I'd like to avoid doing this since it's done similarly in many different places.  For example, see these from ModelUtils and DefaultCodegen:

image

ybelenko:
@jimschubert I think this is up to you and @wing328.

> my suggestion to move this out of code which will run for every operation to something like the custom validation?

I've never used that validator before, but it sounds like a good idea to me.

jimschubert:
I opened #4979 to move the recommendations from our validate command into new validator types, which will allow us to provide recommendations on OpenAPI Specification docs up front from each of our interfaces rather than just the CLI.

Please check it out and let me know what you think.

jimschubert:
Included in #4979

Issue #4418, opened 2019-11-08
[REQ][Kotlin-Client] Add nullable return type for api classes


Mordag:
### Is your feature request related to a problem? Please describe.

My project is using certain endpoints that is returning nullable objects. Currently the api itself only supports non-null return types for api classes.

First issue is part of each api class which the client (e.g. mobile app) is going to use:
`ResponseType.Success -> (response as Success<*>).data as MyModel`

The other issue is the creation of the Success object in the ApiClient:
```
response.isSuccessful -> return Success(
       responseBody(response.body, accept),
       response.code,
       response.headers.toMultimap()
)
```

responseBody is returning a nullable object but Success expects a non-null body. Here the application will throw a kotlin NPE.

## Describe the solution you'd like

Preferable I would like to have a new option in the config file. When generating all classes, this option should define if all api methods are nullable or not (default could be non-null to avoid breaking code).

The output of this could be the following api method:
```
fun getData() : MyDataModel? {
 ...

     return when (response.responseType) {
            ResponseType.Success -> (response as Success<*>).data as MyDataModel?
            ...
      }
}
```

Addition to this change, the ApiInfrastructureResponse would need to get some adjustments as well to allow the "data" class member to be nullable. Maybe there could be a NullableSuccess object that is used instead of the Success object when the config sets the nullable flag to true.

PR #4422, opened 2019-11-08
[Kotlin][Client] Add option to make all api method return a nullable model

Mordag:
Closes https://github.com/OpenAPITools/openapi-generator/issues/4418

- added new additional-parameter **nullableReturnType**
   - converts the api related methods from non-null to nullable
   - change the class member `data` of `ApiInfrastructureResponse.Success` to a nullable type
- added new kotlin-client sample to show the nullable changes
- added new shell and bash script files

It probably makes sense to merge https://github.com/OpenAPITools/openapi-generator/pull/4419 first. I will take care of any merge conflicts.

### PR checklist


Mordag:
PR is ready for review.

@jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10)

Mordag:
@wing328 Since I dragged in your PR, the circleci step is failing. Do you have an idea for how to fix that? Besides that, the PR is ready for review. I fixed the merge conflicts that were caused by your PR.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
```diff
@@ -90,7 +90,7 @@ import java.io.File
         }
 
     protected inline fun <reified T: Any?> responseBody(body: ResponseBody?, mediaType: String? = JsonMediaType): T? {
-        if(body == null) {
+        if(body == null || mediaType == null) {
```
jimschubert: This is potentially "breaking". Rather than throwing an unsupported error for invalid media type, it'll now result in null without providing further details to the user.

If the intent is to return null for unsupported media types, can this be moved to another if block and add a warning log message?

jimschubert: Same for all ApiClient mustache changes.

samples/client/petstore/kotlin-nullable/src/main/kotlin/org/openapitools/client/apis/PetApi.kt
```diff
@@ -0,0 +1,318 @@
...
+    * Multiple tags can be provided with comma separated strings. Use tag1, tag2, tag3 for testing.
+    * @param tags Tags to filter by 
+    * @return kotlin.Array<Pet> or null
+    * @throws UnsupportedOperationException If the API returns an informational or redirection response
+    * @throws ClientException If the API returns a client error response
+    * @throws ServerException If the API returns a server error response
+    */
+    @Suppress("UNCHECKED_CAST")
+    @Throws(UnsupportedOperationException::class, ClientException::class, ServerException::class)
+    fun findPetsByTags(tags: kotlin.Array<kotlin.String>) : kotlin.Array<Pet>? {
```
jimschubert: This is an odd case, maybe. Why return a nullable array rather than just an empty array?

modules/openapi-generator/src/main/resources/kotlin-client/api.mustache
```diff
@@ -68,7 +68,7 @@ import {{packageName}}.infrastructure.toMultiValue
         )
 
         return when (localVarResponse.responseType) {
-            ResponseType.Success -> {{#returnType}}(localVarResponse as Success<*>).data as {{{returnType}}}{{/returnType}}{{^returnType}}Unit{{/returnType}}
+            ResponseType.Success -> {{#returnType}}(localVarResponse as Success<*>).data as {{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}}
```
jimschubert: Shouldn't this be `as? Type` rather than `as Type?`? I've only used the first form and not the second.

See https://kotlinlang.org/docs/reference/typecasts.html#safe-nullable-cast-operator

Mordag: Both solutions are working. I would say the "cleaner" solution is to use `Type?` instead of `as?`. Have a look at the ""Unsafe" cast operator". It is true that `val x: String = y as String` crashes but `val x: String? = y as String?` would solve that. Does that makes sense to you?

samples/client/petstore/kotlin-nullable/src/main/kotlin/org/openapitools/client/apis/PetApi.kt
Mordag: Generally I would agree but it would make the actual method implementation much more complex. I would also need to know if the Type is a List/Array or just a single object (not sure if that is currently possible?). What do you think?

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
Mordag: Sure I can do that. Would you reckon an error log like "mediaType is null" is sufficient or would you prefer some more details in the log?

Mordag:
Thanks for your feedback. 👍 

> Should we also maybe consider adding a default of `false` to the generator, so it's clear in code that this is an option? I eventually want to make additionalProperties sorted by embedded templates to be discoverable.

Do you mean adding the option to this readme >> https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/kotlin.md

modules/openapi-generator/src/main/resources/kotlin-client/api.mustache
jimschubert: Yeah, I guess that's fine. As I understand it, `as T?` will throw if the cast fails while the other doesn't. If it becomes an issue, we could change it later.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
jimschubert: I don't think "mediaType is null" would work here, because this isn't a value which would be null because of programmatic error.

This argument is pulled from the 'Content-Type' header. If one doesn't exist for whatever reason, this would just be null and that could be confusing.

I think something explaining that media type is null and we expected Json at this point?

Again, I'm not following exactly what the intent is with the early return if media type is null. That's covered by the last block of the method. If we're going to "swallow" a response, we'll want to explain why. For body, it makes sense because a null body means a null response. I hope that helps clarify my concern.

samples/client/petstore/kotlin-nullable/src/main/kotlin/org/openapitools/client/apis/PetApi.kt
jimschubert: There are helper properties like `isArray`. You're right that it may make the templates more complex. Let's roll with this and get feedback from the community for any future change.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
Mordag: I actually had cases where a certain api provided a null media type but the body was not null. But I agree that these cases would rather confuse than help. After thinking about that `mediaType == null`-condition, I think the best would be to leave it out (also to avoid breaking changes). I will do some more research and will provide a new PR (if needed).

jimschubert:
@Mordag regarding the default in the generator, I was referring to something like this: https://github.com/OpenAPITools/openapi-generator/blob/11bcb661a15db57f9dc71995e97437c21b6c9434/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java#L394

However, based on the other discussions, it looks like there won't be other changes so I can just pick up the additionalProperties thing later.

My only remaining question on the PR is related to the mediaType guard.

Let me know if you'd like help looking into the CI error. You mentioned it failing after pulling in changes, but I don't see an additional commit (so I didn't know where to begin). It seems from looking at the logs that it failed on Kotlin generation. It could be as simple as missing the executable flag on `bin/kotlin-client-nullable.sh`

Mordag:
@jimschubert I removed the mediaType for now and updated the samples. If all tests are green, the PR should be ready to 🚢.

Mordag:
@jimschubert The circleci is going to fail with the same error again. Can you help me try to figure out the root problem?

jimschubert:
@Mordag I've opened Mordag/openapi-generator#1 against your branch. This should resolve the CircleCI issue.

I didn't merge and fix up in master because your PR is still marked WIP, and GitHub makes it look like you did a force push after you gave thumbs up.  If you merge the PR and remove WIP, then I think your changes look good and can be merged once CI is green.

Looks like the Shippable failure is just Shippable being flaky on container initialization, it was green on the previous commit, prior to the force push (which removes links to those checks)… see [Shippable runs/12015](https://app.shippable.com/github/OpenAPITools/openapi-generator/runs/12015/1/console)

Mordag:
Thanks for helping me resolving the issue with the circleci. 👍 

> I didn't merge and fix up in master because your PR is still marked WIP, and GitHub makes it look like you did a force push after you gave thumbs up. If you merge the PR and remove WIP, then I think your changes look good and can be merged once CI is green.

How do I remove the WIP tag? It doesn't seem that I have the needed permissions for that. The force push is only related to a rebase to master.

jimschubert:
No problem! I've gone ahead and removed the WIP. I see now that it was added by auto-labeler and you have since changed the title (hadn't noticed that earlier).

wing328:
@Mordag  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

Issue #802, opened 2018-08-13
Rewrite of TypeScript Generators


TiFu:
# Motivation

We currently have 7 (!) different generators for TypeScript, which do not share common code. This makes introducing bug fixes to the TypeScript generator family very hard.

# Goal

Create a general framework which can be used in any OOP language - unifying the generator structure and making the project easier to work with. At the beginning, I would like to look into merging the TypeScript generators into one and improve the architecture during the implementation.  

# Feedback

Tear the architecture apart! Focus on the bigger picture first (e.g. missing some important feature? something which is hard to implement with this architecture?) and then go into details (e.g. missing variables).
The TODOs in the text below mark parts which I did not think (enough) about yet. Feel free to fill in what's missing.

# Architecture

[Architecture (looks weird if I include it as an image. sorry!)](https://gist.github.com/TiFu/44abab1bc3a430f7ecad5fed06416445)

The key idea is to extract everything into its own component and hide it behind an interface. The key advantage of this architecture is 

- extendability e.g. new http libraries or more modifiers for requests e.g. a request is fed through multiple components which modify it according to their rules e.g. server sets the correct request url
- testability - the components can be tested in isolation and can be mocked
- less duplication of code for different frameworks in the same language

I'll go through the different components one by one and try to explain the idea.

## HTTP Library

This part should be clear - hide the details of making a HTTP request behind an interface so that we can use different libraries without changing the API Client. To achieve this we create our own data structures for requests and responses.

TODO: Use promises/futures?

## Authentication

Same idea as HTTP library.
I want to highlight one part: the authentication interface has a method `apply(Request r)`. This method applies the authentication scheme to the given request e.g. in the case of an API Key it would append the api key.

TODO: OAuth work flow with this implementation?

## Response Models & Param Models

TODO: Should this be separated or can response & param models be used interchangeably? i.e. param Pet can also be a response model? If yes these two components should be merged.

This component contains model classes for each parameter & response as well as a (object) serializer and deserializer for these models (e.g. `Array<JsonObject>` in the API response is transformed to `Array<Pet>` by the ObjectDeserializer). 

## Server

This component contains the different servers as defined in OAS. 

TODO: handling of variables. I would propose to offer setters for each variable.
Variables are not yet supported in OpenAPITools. See #793 

## Config

See current implementation.

## API Client

The API client coordinates all other components.

1. Serializes the input parameters
2. creates a new request `r`
3. calls `Server#apply(r)`
[...]

PR #4424, opened 2019-11-08
Typescript refactor: stub rxjs

bodograumann:
As discussed in #802 I have replaced RxJS with a stub when it is not really needed.
There is a new option for the typescript generator `useRxJS` which defaults to `false`.

Additionally I fixed some minor issues.

I noticed that the tests in `samples/client/petstore/typescript/tests` are not automatically generated, but statically committed. So that is another open todo?

CC @TiFu


modules/openapi-generator/src/main/resources/typescript/api/middleware.mustache
```diff
@@ -55,12 +55,12 @@ export interface PromiseMiddleware {
 	 * @returns an observable of the updated request context
 	 *
 	 */
-    pre(context: RequestContext): Observable<RequestContext>;
```
TiFu: Why was this changed to `Promise`?

bodograumann:
### Compilation errors
I compiled locally. The `mvn` and generating the samples worked fine.
The tests did not work. I am not sure why that is. It was indeed confusing because the last time it worked, but when I checked out the typescript-refactor branch and rebuild everything, they still would not work. I guess I need to take a little more time to investigate this issue.

### Readme
Yeah, I did not run ensure-up-to-date.sh again. I was hoping it wasn't needed because last time it changed unrelated things differently than the CI.
Have added the docs line for the useRxJS switch manually now.

### Middleware
I changed it to what I think the original intention was. This is the only way that makes sense to me. Previously both interfaces `Middleware` and `PromiseMiddleware` where identical.
When thinking about how middlewares should work with RxJS stubed, I came to the conclusion, that only `PromiseMiddleware` which is then wrapped, can be used. The user should not be aware of the `Observable` stub, so he must work only with `Promise`.

bodograumann:
Ok, got the tests to run on the original branch and found the problem:
We require the node types, but only implicitly depended on them via `@types/form-data`.
I have now added them as an explicit dependency.
Furthermore I have adjusted the dependencies in the tests according to the changes to the template and samples.

For the jquery tests to run I had to install webpack.

The tests where using `Observable.subscribe`. Afaics this is currently not really necessary, so I replaced it with the stubbed `Observable.toPromise().then`. If in the future it is needed, we can of course add it to the stub.

Not sure whether the jquery tests ran properly on my local machine, because I got 
> \> node ./dist/test-runner.js
>
> [error] Failed to load resource: the server responded with a status of 404 (Not Found)

But let's see what the ci says.

bodograumann:
The failures don’t seem to be related to my changes.
What do you think, @TiFu?

TiFu:
Thanks a lot for investigating this issue and the PR! 

- CircleCI is unrelated
- DroneIO as well
Going to close and reopen this PR to rerun the tests. If it fails again, i'll just merge anyway. 

> Compilation errors
> Readme
Sounds good

> Middleware
Oh the change was in `PromiseMiddleware`. Good catch.

bodograumann:
Ok, forgot to remove `subscribe` from the jquery tests. Now it really is complete :-D
Tests seem good. Failure because of some OCAML stuff. Different from before but still unrelated to these changes.

TiFu:
Great! Again, thanks for the PR which has been merged into ts-refactor.

wing328:
@bodograumann thanks for the PR, which has been included in the v4.2.2 release: https://twitter.com/oas_generator/status/1201432648544972800

Issue #3399, opened 2019-07-19
[BUG] OAS3 server URL/host not parsed


MartinDelille:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

Hi,

After a long time working on other subject, I come back to openapi-generator because I seriously consider using it for my project.

##### Description

I'm encountering a issue when generating a Qt5 client where there is a URL generation error: the protocol is not added when building the request:

```cpp
OAIDefaultApi::OAIDefaultApi() : basePath("/v1"),
    host("api.myapp.com"),
    timeout(0){

}

...

void
OAIDefaultApi::findContactById(const QString& contact_id) {
    QString fullPath;
    fullPath.append(this->host).append(this->basePath).append("/contact/{contactId}");
    ...
}
```

This code generate the following url: `api.myapp.com/v1/contact/{contactId}` instead of `https://api.myapp.com/v1/contact/{contactId}`.

There is a easy way to fix it is to call `defaultApi.setHost("https://api.myapp.com")` before `findContactById` but I guess it might broke in the long term.

##### openapi-generator version

4.0.3

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.1
servers:
  - url: 'https://api.myapp.com/v1'
info:
  description: This is myapp.com API.
  version: 1.0.0
  title: myapp.com
components:
  schemas:
    Contact:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: string
        name:
          type: string
paths:
  /contact/{contactId}:
    get:
      operationId: findContactById
      parameters:
        - name: contactId
          in: path
          schema:
            type: string
          required: true
      responses:
        '200':
          description: An contact object
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Contact'
```

##### Command line used for generation

```
openapi-generator generate -i api.yaml -g cpp-qt5-client
```

##### Steps to reproduce

[...]

PR #4429, opened 2019-11-09
cpp-qt5-client: remove host since it is not well handled

MartinDelille:
This PR replace part of the job done here: #3688 

Now that the base path is well handled, the API class can use the one provided in the API description.

There was previously a mismatch between host and base path as mentionned here: #3399

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCppCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCppCodegen.java
@@ -300,11 +300,16 @@ abstract public class AbstractCppCodegen extends DefaultCodegen implements Codeg
         URL url = URLPathUtils.getServerURL(openAPI, serverVariableOverrides());
         String port = URLPathUtils.getPort(url, "");
         String host = url.getHost();
+        String scheme = url.getProtocol();
+
         if(!port.isEmpty()) {
             this.additionalProperties.put("serverPort", port);
         }
         if(!host.isEmpty()) {
             this.additionalProperties.put("serverHost", host);
+        }
+        if(!scheme.isEmpty()) {
+            this.additionalProperties.put("scheme", scheme);
         }
     }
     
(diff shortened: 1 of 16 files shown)


etherealjoy:
@MartinDelille 
May I propose a different approach to solving this? 

1. Leave the base path in the API as it is.
```
    fullPath.append(this->host).append(this->basePath).append("{{{path}}}");
```
There are many use cases to have base path as is. e.g.
- when same client is using multiple hosts at runtime.
- When using REST over an existing socket. (By subclassing CNetworkManager)

2. Properly parse the host and update/add constructor if necessary?

MartinDelille:
You're right there is indeed a problem in this code I wrote this summer.

The problem I raised [here](https://github.com/OpenAPITools/openapi-generator/issues/3399) and that I'm trying to solve is that `http://api.example.com` is not a host but its not a base path neither.

I need to dig into this

etherealjoy:
I tried something quick on my side to add parsing of scheme

in `AbstractCppCodegen.java` add the parsing of the scheme like below
```
    @Override
    public void preprocessOpenAPI(OpenAPI openAPI) {
        URL url = URLPathUtils.getServerURL(openAPI, serverVariableOverrides());
        String port = URLPathUtils.getPort(url, "");
        String host = url.getHost();
        String scheme = url.getProtocol();
        
        if(!port.isEmpty()) {
            this.additionalProperties.put("serverPort", port);
        }
        if(!host.isEmpty()) {
            this.additionalProperties.put("serverHost", host);
        }
        if(!scheme.isEmpty()) {
            this.additionalProperties.put("scheme", scheme);
        }
    }
```

Then you can remove all these lines from the three header files
```
private:
    const QString PetStoreHost = QStringLiteral("http://petstore.swagger.io");
```
And from all three source files
```
    api.setHost(PetStoreHost);
```

MartinDelille:
The only thing is that as I said above, `http://api.example.com` is more a base url that a host (the host is `api.example.com`), so maybe we should rename `host` to `baseUrl` in the api class?

etherealjoy:
> so maybe we should rename host to baseUrl in the api class?

Ok for me

MartinDelille:
Ok I finally change to the following implementation.

I had to rebase #4435 so maybe we should merge it first if everything is fine with it.

PR #4435, opened 2019-11-10
fixup! Simpler timeout with QTimer::singleShot (#4430)

MartinDelille:
This fix the problem mentionned here: https://github.com/OpenAPITools/openapi-generator/pull/4430#issuecomment-552183699

### DIFF
--- a/samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
+++ b/samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
@@ -100,7 +100,7 @@ QByteArray *PFXHttpRequestWorker::getMultiPartField(const QString &fieldname){
 }
 
 void PFXHttpRequestWorker::setTimeOut(int timeOut){
-    _timeOut = _timeOut;
+    _timeOut = timeOut;
 }
 
 void PFXHttpRequestWorker::setWorkingDirectory(const QString &path){
--- a/samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
+++ b/samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
@@ -374,6 +374,7 @@ void PFXHttpRequestWorker::on_manager_finished(QNetworkReply *reply) {
             headers.insert(item.first, item.second);
         }
     }
+    disconnect(this, nullptr, nullptr, nullptr);
     reply->deleteLater();
     process_form_response();
     emit on_execution_finished(this);
@@ -383,7 +384,6 @@ void PFXHttpRequestWorker::on_manager_timeout(QNetworkReply *reply) {
     error_type = QNetworkReply::TimeoutError;
     response = "";
     error_str = "Timed out waiting for response";
-    disconnect(manager, nullptr, nullptr, nullptr);
     reply->abort();
     reply->deleteLater();
     emit on_execution_finished(this);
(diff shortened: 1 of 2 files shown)


MartinDelille:
This mistake leverage a lack in our testing: the CI should test that client has been updated.

samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
```diff
@@ -374,6 +374,7 @@ void PFXHttpRequestWorker::on_manager_finished(QNetworkReply *reply) {
             headers.insert(item.first, item.second);
         }
     }
+    disconnect(this, nullptr, nullptr, nullptr);
```
etherealjoy: Actually this change broke tests.

etherealjoy: We can wait for the build to finish. CI should fail.

MartinDelille: Yes it fails locally on my side...

etherealjoy:
https://doc.qt.io/qt-5/qobject.html#disconnect
Disconnect everything connected to an object's signals:
```
disconnect(myObject, 0, 0, 0);
```

So indeed original code below is correct.
```C++
void PFXHttpRequestWorker::on_manager_timeout(QNetworkReply *reply) {
    error_type = QNetworkReply::TimeoutError;
    response = "";
    error_str = "Timed out waiting for response";
    disconnect(manager, nullptr, nullptr, nullptr);
    reply->abort();
```

I was trying to suggest this below, but since the code was reworked in your PR I got confused.

```
void PFXHttpRequestWorker::on_manager_finished(QNetworkReply *reply) {
    if(timer->isActive()){
        timer->stop();
    }
    disconnect(timer, nullptr, nullptr, nullptr);
    error_type = reply->error();
    response = reply->readAll();
    error_str = reply->errorString();
```

MartinDelille:
Do we really need to disconnect in the end?

etherealjoy:
Yes we do.
I edited my post above.

etherealjoy:
So we dont get both the callbacks for each API call.

MartinDelille:
Ok so I push the new fix

etherealjoy:
> Ok so I push the new fix

This should solve one topic related to disconnect response callback when timeout.
We need to disconnect the timer callback if we receive the response.

etherealjoy:
There is another bug in this branch.
```
void PFXHttpRequestWorker::setTimeOut(int timeOut){
    _timeOut = _timeOut;
}
```
self assignment.

MartinDelille:
> We need to disconnect the timer callback if we receive the response.

But the worker is deleted later in the callback. Shouldn't it disconnect from the futur timer event?

> self assignment.

My bad! 😅

etherealjoy:
> But the worker is deleted later in the callback. Shouldn't it disconnect from the futur timer event?

I think it is OK.

```
void QObject::deleteLater()

Note: It is safe to call this function more than once; when the first deferred deletion event is delivered, any pending events for the object are removed from the event queue.
```
Lets see if there are issues reported later.

MartinDelille:
Ok I pushed the self assignement fix.

I think this should be the worker responsability to delete itself later (and not the callback).

wing328:
@MartinDelille  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

PR #4443, opened 2019-11-10
[Python] Use Python 2.6 compatible dict definition

richardwhiuk:
This changes the generated client to use a Python 2.6 compatible dictionary definition, which allows the client (on at least some APIs) to work on Python 2.6

This work was contributed by @nick-slack at @Metaswitch.

## Python Technical Committee

@taxpon, @frol, @mbohlool, @cbornet, @kenjones-cisco, @tomplus, @Jyhess, @slash-arun

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api_client.mustache
@@ -231,12 +231,12 @@ class ApiClient(object):
             # and attributes which value is not None.
             # Convert attribute name to json key in
             # model definition for request.
-            obj_dict = {obj.attribute_map[attr]: getattr(obj, attr)
-                        for attr, _ in six.iteritems(obj.openapi_types)
-                        if getattr(obj, attr) is not None}
+            obj_dict = dict([(obj.attribute_map[attr], getattr(obj, attr))
+                             for attr, _ in six.iteritems(obj.openapi_types)
+                             if getattr(obj, attr) is not None])
 
-        return {key: self.sanitize_for_serialization(val)
-                for key, val in six.iteritems(obj_dict)}
+        return dict([(key, self.sanitize_for_serialization(val))
+                     for key, val in six.iteritems(obj_dict)])
 
     def deserialize(self, response, response_type):
         """Deserializes response into an object.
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -231,12 +231,12 @@ class ApiClient(object):
             # and attributes which value is not None.
             # Convert attribute name to json key in
             # model definition for request.
-            obj_dict = {obj.attribute_map[attr]: getattr(obj, attr)
-                        for attr, _ in six.iteritems(obj.openapi_types)
-                        if getattr(obj, attr) is not None}
+            obj_dict = dict([(obj.attribute_map[attr], getattr(obj, attr))
```
tomplus: This version will be slower and the syntax looks an ancient for me. I'm not sure if we should support the Python 2.6, the days before end of life Python 2.x

jimschubert: I agree, or at least make the syntax conditional by some config value.
I'd personally recommend users requiring 2.6 support to customize the template rather than adding that support here. This would mean that we now have to test and support 2.6 and 3.x differently.

spacether: @tomplus @jimschubert testing confirms that you said:
```
import time
import logging

iterable = []
for i in range(30):
    iterable.append((chr(i), i))

logging.basicConfig(level=logging.INFO)

def test_runner(fn, qty):
    start_time = time.time()
    for i in range(qty):
        fn()
    execution_time = time.time() - start_time
    logging.info("fn=%s execution_time=%s" % (fn.__name__, execution_time))

def make_dict_new():
    return {key: value for (key, value) in iterable}

def make_dict_old():
    return dict(tup for tup in iterable)

test_runner(make_dict_old, 100000)
test_runner(make_dict_new, 100000)
```
-->
```
~/programming/scratch$ python3 python_dict_comprehension.py
INFO:root:fn=make_dict_new execution_time=0.2842717170715332
INFO:root:fn=make_dict_old execution_time=0.41632080078125
~/programming/scratch$ python python_dict_comprehension.py
INFO:root:fn=make_dict_new execution_time=0.255958080292
INFO:root:fn=make_dict_old execution_time=0.378924131393
```
@richardwhiuk can you update your code to only use the dict(...) version if the python version <= 2.6?

spacether:
Python 2.7 end of life was Jan 1, 2020 per https://github.com/python/devguide/pull/344
For now we will limit our syntax to 2.7 and higher syntax
If users want to use older versions of python with the generator, then they can customize the templates per @jimschubert's suggestion

modules/openapi-generator/src/main/resources/python/api_client.mustache
richardwhiuk: We aren't going to merge this, so resolving this discussion.

PR #4444, opened 2019-11-10
[C++] [Qt5] Clang format Code

MartinDelille:
This is an attempt to set several [clang-format](https://clang.llvm.org/docs/ClangFormat.html) rules on the generated code.

What do you think about it?

### PR checklist


etherealjoy:
This is a great PR. 
However we need to agree on the style.

https://clang.llvm.org/docs/ClangFormatStyleOptions.html


Possible values:
```
LLVM A style complying with the LLVM coding standards
Google A style complying with Google’s C++ style guide
Chromium A style complying with Chromium’s style guide
Mozilla A style complying with Mozilla’s style guide
WebKit A style complying with WebKit’s style guide
Microsoft A style complying with Microsoft’s style guide
```

jimschubert:
I've had to cancel this PR's CircleCI build because it was blocking our queue for 6 hours. Once the discussion above about style is addressed, please trigger CircleCI again.

MartinDelille:
I open to any suggestion concerning the style. I think I used the `LLVM` style since it seems to be the default. I just needed to apply some tweak to make it possible to use it (see `.clang-format` comments)

samples/client/petstore/cpp-qt5/.clang-format
```diff
@@ -0,0 +1,7 @@
+---
+Language:        Cpp
+AccessModifierOffset: -4
+AllowShortFunctionsOnASingleLine: Empty
```
MartinDelille: default LLVM is All

samples/client/petstore/cpp-qt5/.clang-format
```diff
@@ -0,0 +1,7 @@
...
+BreakBeforeBinaryOperators: NonAssignment
```
MartinDelille: It allowed me to `if` statement with many condition into several lines.

samples/client/petstore/cpp-qt5/.clang-format
```diff
@@ -0,0 +1,7 @@
...
+ColumnLimit: 0
```
MartinDelille: The main issue I was dealing with this setting is that if let to default, the description was splitted into several lines which is not easy to do in the mustache template.

MartinDelille:
This might be some help: https://zed0.co.uk/clang-format-configurator/

wing328:
My suggestion is to include a default .clang-format (auto-generated) and let the user customize it to meet their requirements/preferences.

MartinDelille:
@wing328 This could make the templating easier to write but wouldn't it require to have clang-format as a dependency?

wing328:
> @wing328 This could make the templating easier to write but wouldn't it require to have clang-format as a dependency?

To be clear, we don't need to update the template. We simply leverage code formatter to do the job for us.

Users can use file post-processing (as part of openapi-generator) or use the code formatter on the output after the code generation completes. 

Using Ruby as an example, the generator outputs the `.rubocop.yml` file as a starting point. Users can of course customize it with any rules they want.

MartinDelille:
@wing328  I totally agree with you about not modifying the template: tweaking it to respect the rules I used for the example was not an easy task.

I will look into the ruby generator to mimic the behavior.

MartinDelille:
@etherealjoy Regarding the style, we should maybe have `llvm` as a default style (since it's `clang-format` default style)  and allow the user to specify it's own?

wing328:
> maybe have llvm as a default style (since it's clang-format default style) 

Agreed with using clang-format's default style as a starting point.

MartinDelille:
After investigating I realised that the processing already work and is even more generic than `clang-format` and can work with other tools like `uncrustify` for example.

We just need to set `CPP_POST_PROCESS_FILE` to `/usr/local/bin/clang-format -i` for example or `uncrustify --replace --no-backup` and pass `--enable-post-process-file` to the command line.

The question is: do we want to enable post processing for the sample?

wing328:
> The question is: do we want to enable post processing for the sample?

We'd a discussion on that before (we also tried) but the problem is not all contributors have the code formatter installed.

MartinDelille:
Yes that is what I was thinking.

Do you think improving the style of the generated code is still relevant?

wing328:
@MartinDelille I definitely like your change.

A little background on `--enable-post-process-file`. We found it beneficial to post-process the file (e.g. formatting using different formatter, etc) and it's not easy to get every space, code alignment correct with the mustache templates.

We still try to make the default output looks good in terms of code format so we definitely welcome your PR to update the template with better code format.

Please resolve the merge conflicts and I'll merge this PR.

MartinDelille:
Please consider merging this PR (and #4526) before other cpp-qt5 template modification because rebasing and merging conflict can be quite long! 😅

samples/client/petstore/cpp-qt5/client/PFXHttpRequest.cpp
```diff
@@ -205,13 +189,12 @@ void PFXHttpRequestWorker::execute(PFXHttpRequestInput *input) {
...
-    else {
+    } else {
         // variable layout is MULTIPART
 
-        boundary = "__-----------------------"
-            + QString::number(QDateTime::currentDateTime().toTime_t())
-            + QString::number(qrand());
+        boundary = QString("__-----------------------%0%1")
+                       .arg(QDateTime::currentDateTime().toTime_t())
+                       .arg(qrand());
```
wing328: @etherealjoy switched to `.arg` with `%0%1` instead. Does it look good to you? I"ve not tested it myself.

From the doc (https://doc.qt.io/qt-5/qstring.html), looks like it usually starts with %1 instead of %0?

cc @MartinDelille

etherealjoy: It should be %1 %2, not from %0 indeed.

MartinDelille: Indeed my mistake. The thing is that on my side it works but the documentation say we should start from %1: https://doc.qt.io/qt-5/qstring.html#arg

modules/openapi-generator/src/main/resources/cpp-qt5-client/api-body.mustache
```diff
@@ -9,166 +9,150 @@
...
+void {{classname}}::addHeaders(const QString &key, const QString &value) {
     defaultHeaders.insert(key, value);
 }
 
 {{#operations}}
 {{#operation}}
-void
-{{classname}}::{{nickname}}({{#allParams}}const {{{dataType}}}& {{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) {
+void {{classname}}::{{nickname}}({{#allParams}}const {{{dataType}}} &{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) {
     QString fullPath = QString("%0://%1%2%3%4")
```
etherealjoy: same issue here

MartinDelille: Yes it's fixed here: 2cdf238b74

samples/client/petstore/cpp-qt5/.clang-format
wing328: @MartinDelille Are we going to restore .clang-foramt to the default?

modules/openapi-generator/src/main/resources/cpp-qt5-client/model-body.mustache
```diff
@@ -22,24 +21,19 @@ namespace {{this}} {
...
 
-{{classname}}::~{{classname}}() {
+{{classname}}::~{{classname}}() {}
 
-}
-
-void
-{{classname}}::initializeModel() {
-    {{^isEnum}}{{#vars}}
+void {{classname}}::initializeModel() {{=<% %>=}}{<%={{ }}=%>{{^isEnum}}{{#vars}}
```
etherealjoy: what is this change for?

samples/client/petstore/cpp-qt5/client/PFXStoreApi.cpp
```diff
@@ -18,86 +17,75 @@
...
 
-void PFXStoreApi::addHeaders(const QString& key, const QString& value){
+void PFXStoreApi::addHeaders(const QString &key, const QString &value) {
     defaultHeaders.insert(key, value);
 }
 
-void
-PFXStoreApi::deleteOrder(const QString& order_id) {
+void PFXStoreApi::deleteOrder(const QString &order_id) {
     QString fullPath = QString("%0://%1%2%3%4")
```
etherealjoy: still with %0 starting.

samples/client/petstore/cpp-qt5/.clang-format
MartinDelille: I could remove this file since it is not really used.

I used it to help me to generate a better output but was unable to respect the specific `ColumnLimit` rule, in particular when writing the header documentation.

The current template generates:

```
/**
 * OpenAPI Petstore
 * This is a sample server Petstore server. For this sample, you can use the api key `special-key` to test the authorization filters.
 *
 * The version of the OpenAPI document: 1.0.0
 *
 * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
 * https://openapi-generator.tech
 * Do not edit the class manually.
 */
```

Setting `ColumnLimit` to the default would require the following generation:

```
/**
 * OpenAPI Petstore
 * This is a sample server Petstore server. For this sample, you can use the api
 * key `special-key` to test the authorization filters.
 *
 * The version of the OpenAPI document: 1.0.0
 *
 * NOTE: This class is auto generated by OpenAPI Generator
 * (https://openapi-generator.tech). https://openapi-generator.tech Do not edit
 * the class manually.
 */
```

I have no idea how to split the description from the mustache template.

etherealjoy:
Just one additional comment, rest looks fine.

modules/openapi-generator/src/main/resources/cpp-qt5-client/model-body.mustache
MartinDelille: This (kind of dirty) trick allowed me to replace temporaly the template delimiter from `{{` / `}}` to `<%` / `%>`.

My first iteration was:

```
void {{classname}}::initializeModel() {{{^isEnum}}{{#vars}}
    m_{{name}}_isSet = false;
    ...
```

But if fails since mustache didn't like the `{{{`.

samples/client/petstore/cpp-qt5/client/PFXStoreApi.cpp
MartinDelille: I forgot (again) to update the model. I amended my commit: 05c5f8cb5d

modules/openapi-generator/src/main/resources/cpp-qt5-client/model-body.mustache
etherealjoy: you can give a space.

MartinDelille: I will rather give a carriage return to avoid extra space at the end of the line: a84a0038c6

MartinDelille:
I used `git commit --fixup`. Tell me if you want me to clean the history.

etherealjoy:
It will be squashed on merge so not necessary.

MartinDelille:
Is something blocking this PR to be merged?

If someone else is working on the qt5 client template, there is a risk of ruining the job done here.

samples/client/petstore/cpp-qt5/.clang-format
wing328: > I have no idea how to split the description from the mustache template.

My take is to leverage code formatter to take care of the line split.

MartinDelille: Yes I got it! This `.clang-format` will only be used if we want to check the template generate a clang format compatible code.

PR #4449, opened 2019-11-11
Dart generate keywords

### LINKED ISSUE #4448: [REQ] [Dart] Use Dart SDK to generate Dart keywords for reserved words list

A recent PR improved the Dart generator by updating the reserved words.  When I went through and checked if there were any more Dart keywords missing, there were a few still missing and it was quite time consuming to determine exactly which words were missing as it is a hard-coded list.

Use the Dart SDK to automatically generate the Dart keywords and have the dart generator use the generated set of words. 

We could keep the list of reserved words hard-coded in and just add a comment that the list was generated by the SDK with a date and Dart version so we know when the list needs to be updated.

nickmeinhold:
Potentially fixes #4448 
- a Dart program that generates a text file containing keywords for the current Dart version 
- changed the dart generator to use the dart keywords as the set of reserved words 

Questions 
- project structure: I imagine there is a better place to put these files but nothing jumped out at me when I looked through 
- the list of reserved words is now missing a few of the previous reserved words (just the ones that are a keyword with an asterisk, eg. async*, yield*) 
  - not sure if this is a problem, if it is maybe we should add a separate check for special characters? 

### PR checklist

- [x] Run the shell script(s) 
- [x] Copy the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee)
- @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09)


wing328:
@nickmeinhold  thanks for the PR, which has been included in the v4.2.1 release: https://twitter.com/oas_generator/status/1195339336922759168

PR #4453, opened 2019-11-11
[kotlin][client] Add inheritance to Kotlin Client templates

mtraynham:
Adds inheritance to the Kotlin Client codegen templates using interfaces as the parent class.

### PR checklist

@jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10)


samples/client/petstore/kotlin/src/main/kotlin/org/openapitools/client/models/Order.kt
```diff
@@ -38,8 +38,12 @@ data class Order (
     val status: Order.Status? = null,
     @Json(name = "complete")
     val complete: kotlin.Boolean? = null
-) 
-: Serializable 
+)
+
+
+
+ : Serializable
```
jimschubert: I think we'll need to clean up the newly added empty lines before merge.

modules/openapi-generator/src/main/resources/kotlin-client/data_class.mustache
```diff
@@ -46,6 +52,14 @@ import java.io.Serializable
...
 	}
 {{/serializableModel}}
+{{#discriminator}}
+{{#requiredVars}}
+{{>interface_req_var}}
+{{/requiredVars}}
+{{#optionalVars}}
+{{>interface_opt_var}}
+{{/optionalVars}}
+{{/discriminator}}
```
jimschubert: Not tested, but I think this would need to be the following to avoid empty newlines:

```
{{/serializableModel}}{{#discriminator}}{{#requiredVars}}
{{>interface_req_var}}{{/requiredVars}}{{#optionalVars}}
{{>interface_opt_var}}{{/optionalVars}}{{/discriminator}}
```

jimschubert:
@mtraynham thanks for keeping updated with master. That's much appreciated.  My only concern about the newly added code is the empty newlines it introduces for noop conditions.

I wonder if we can target 4.3.x with this new code? I have concerns that the introduction of the new template files may cause issue, and just to be safe we could release in the minor version which is "breaking change with fallback". The fallback being users could use 4.2.x templates if they have issues.

Any concern with retargeting? As an alternative we could get the vendor extension into master and the template changes against the 4.3.x branch.

mtraynham:
Thanks @jimschubert for taking a look!  I'm not against flattening the new lines, it just makes the template kinda of unreadable.  I'll rebase against the 4.3.x branch as well.

samples/client/petstore/kotlin/src/main/kotlin/org/openapitools/client/models/Order.kt
mtraynham: Corrected by putting all import variants on the same line.

modules/openapi-generator/src/main/resources/kotlin-client/data_class.mustache
mtraynham: Corrected, thanks!

mtraynham:
@jimschubert , I corrected the issues you mentioned and rebased my commits on to OpenAPITools/4.3.x.  

Looks like the checks are failing because it's attempting to merge the feature branch into master and getting some merge failures.  I can rebase again when those are corrected if need be.

jimschubert:
Verified inheritance after testing a merge of master. Looks good!

Issue #4241, opened 2019-10-23
[REQ] [csharp-netcore] Custom JsonSerializer needed


PeterTriebe:
### Is your feature request related to a problem? Please describe.

Lib: csharp-netcore

I want to do a "Patch" request to update some entries.
For this reason its needed that "null" values are not serialized. If a property is not serialized it will be ignored from the service.

## Describe the solution you'd like

To solve this i want to add a custom Json serializer. Inside this I'm able to say that null values will be ignored.

example how it worket after i changed the generated code:

```
        private RestRequest newRequest(
            HttpMethod method,
            String path,
            RequestOptions options,
            IReadableConfiguration configuration)
        {

....
            RestRequest request = new RestRequest(Method(method))
            {
                Resource = path,
                JsonSerializer = new CustomJsonCodec(new JsonSerializerSettings
                {
                  NullValueHandling = NullValueHandling.Ignore
                },configuration)
            };
```
But we want to make it more generic :-)
The creation of this request is wrapped by the generated client. And there is actual no way to pass a custom Json serializer. I think its a good idea to pass a custon serializer together wirh the configuration. With this the client can configure what he want.

## Describe alternatives you've considered

Its also possible to add some configurations like `NullValueHandling = NullValueHandling.Ignore` directly to the config. But i think thats not so suitable.

jimschubert:
Are you specifically looking for something that can be manually set by the user?

I'd like to keep RestSharp and JSON.net types out of the Configuration object, because the design allows users to replace the underlying accessor completely. If we begin adding these types back to the Configuration object, the decision to support these types moves from runtime to codegen time.

For your use case, would it make sense to move the RestSharp client construction to it's own template, which you could override as you see fit? The tooling allows you to override only the templates you care to modify, falling back to built-in templates.

I'm hesitant as well to introduce a new option for this patch use case because it would be "global", meaning the client wouldn't allow you to explicitly null out a property in non-PATCH endpoints.

jimschubert:
On second thought, we could also add an option to use the ignore nulls for patch only. Is this the intended suggestion? I may have originally misunderstood.

PeterTriebe:
Ok I understand.
Default it was also set a JsonSerializer from type CustomJaonCodec :

```
        private readonly JsonSerializerSettings _serializerSettings = new JsonSerializerSettings
        {
            // OpenAPI generated types generally hide default constructors.
[...]

Issue #4236, opened 2019-10-23
[BUG] [csharp-netcore] No way to pass proxy


PeterTriebe:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Sorry I'm not shure if this a bug or feature request.

My client is generated with csharp-netcore.
I want to setup a proxy with the generated client. Here is mandatory not changig the generated code. 

But there is actually no way to pass a proxy to the RestSharp client.
Everything is wrapped.

In Net framwork and Net core 3 its possible to setup a global proxy. But in Net core 2 its not possible.

Before using csharp-netcore i generated the client the client under targetframework net45.
Here i was able to use a proxy.

Example: 

```
     var proxy = new WebProxy("http://localhost:3128")
      {
        Credentials = new NetworkCredential("USER", "PASS")
      };

      var config = new Configuration
      {
        BasePath = "https://my-example-api.com/api/v1",
        AccessToken = token.access_token
      };

      config.ApiClient.RestClient.Proxy = proxy;
```

In the same way its missed for netstandard.

##### openapi-generator version

4.1.3

##### Command line used for generation

java -jar "openapi-generator-cli-4.1.3.jar" generate -i api.yaml -g csharp-netcore -o Example.Client -p packageName=Example.Client

##### Steps to reproduce

Generate a csharp-netcore client and try to setup a proxy

##### Suggest a fix

I think its needed to add api client configuration to the config in the same way like in Net framework

wing328:
> Before using csharp-netcore i generated the client the client under targetframework net45.
Here i was able to use a proxy.

Yes, I agree the csharp-netcore client should provide the same.

A good starting point is to  customize the templates  and use the `-t` option:

Ref: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/csharp-netcore/ApiClient.mustache

PR #4466, opened 2019-11-12
#4241, #4236 Add cutom Json serializer and proxy settings to config

PeterTriebe:
**#4236 Add proxy settings to config**
This was needed because in net-core 2x its not possible to set a global proxy.

#4241 Add json serializer settings to config
This was needed to change the serializer settings. In my case i wanted to ignore null entries.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/Configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/Configuration.mustache
@@ -8,7 +8,9 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Text;
+using Newtonsoft.Json;
 
 namespace {{packageName}}.Client
 {
@@ -223,6 +225,13 @@ namespace {{packageName}}.Client
         /// <value>The access token.</value>
         public virtual string AccessToken { get; set; }
 
+
+        /// <inheritdoc />
+        public JsonSerializerSettings SerializerSettings { get; set; } = new JsonSerializerSettings();
+
+        /// <inheritdoc />
+        public WebProxy Proxy { get; set; }
+
         /// <summary>
         /// Gets or sets the temporary folder path to store the files downloaded from the server.
         /// </summary>
@@ -408,7 +417,9 @@ namespace {{packageName}}.Client
                 Password = second.Password ?? first.Password,
                 AccessToken = second.AccessToken ?? first.AccessToken,
                 TempFolderPath = second.TempFolderPath ?? first.TempFolderPath,
-                DateTimeFormat = second.DateTimeFormat ?? first.DateTimeFormat
+                DateTimeFormat = second.DateTimeFormat ?? first.DateTimeFormat,
+				Proxy = second.Proxy ?? first.Proxy,
+                SerializerSettings = second.SerializerSettings ?? first.SerializerSettings
             };
             return config;
         }
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
@@ -2,6 +2,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Net;
+using Newtonsoft.Json;
 
 namespace {{packageName}}.Client
 {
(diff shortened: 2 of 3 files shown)


PeterTriebe:
@mandrean (2017/08), @jimschubert (2017/09) ❤️ @frankyjuang (2019/09) 
Plsease take a look. Thanks :-)

PeterTriebe:
How I can rebuild  CircleCI? I generated some pedstore things..

jimschubert:
@PeterTriebe Thanks for the contribution! I've triggered the build for you.

modules/openapi-generator/src/main/resources/csharp-netcore/Configuration.mustache
```diff
@@ -223,6 +225,13 @@ namespace {{packageName}}.Client
         /// <value>The access token.</value>
         public virtual string AccessToken { get; set; }
 
+
+        /// <inheritdoc />
+        public JsonSerializerSettings SerializerSettings { get; set; } = new JsonSerializerSettings();
```
jimschubert: I think this will need to be moved elsewhere… maybe as a constructor argument for `ApiClient`.

`Configuration` should be decoupled from framework implementation. For example, `ApiClient` is tightly coupled to RestSharp and JSON.net.  Having `Configuration` rely only on "standard" types means that users may replace `ApiClient` with their desired accessor implementations. Including Json.NET types in Configuration will make this unusable for users who don't pull that type into their client.

modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
```diff
@@ -89,5 +91,15 @@ namespace {{packageName}}.Client
         /// <param name="apiKeyIdentifier">API key identifier (authentication scheme).</param>
         /// <returns>API key with prefix.</returns>
         string GetApiKeyWithPrefix(string apiKeyIdentifier);
+		
+		/// <summary>
+		/// Gets the json serializer settings
+		/// </summary>
+		JsonSerializerSettings SerializerSettings { get; }
```
jimschubert: See previous comment about scoping this to `ApiClient` only.

modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
```diff
@@ -89,5 +91,15 @@ namespace {{packageName}}.Client
...
+		
+		/// <summary>
+		/// Gets the proxy
+		/// </summary>
+		WebProxy Proxy { get; }
```
jimschubert: Note: changing this interface is a breaking change for existing client consumers.
Also, formatting is off by 4 spaces too many here.

jimschubert:
@PeterTriebe I've left a few comments.  CircleCI has failed because you'll need to rebuild the project and run the relevant csharp-netcore scripts under `./bin` to regenerate affected samples.

modules/openapi-generator/src/main/resources/csharp-netcore/Configuration.mustache
PeterTriebe: Its possible to move it to the constructor. But everything is configurated inside the `Configuration `type but just not this? I think every C# developer use `Newtonsoft.Json`. Even if the `ApiClient ` changes. When it moved to the constructor of the `ApiClient ` , the class that wrap this must also use `Newtonsoft.Json`. 
If you want to use this type just inside the `ApiClient `, we have to create a custom type and map everything. But this is not a good way.

modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
PeterTriebe: I thought this interface is just used from the `Configuration` ? And this class is also generated. Using this proxy is optional

PeterTriebe:
@jimschubert I have run the scripts and pushed it after the run fails the first time. Hm. I can try it again

wing328:

hacki11:
is the 4.2.3 milestone correct? does the release 4.2.3 contain this pull request?

modules/openapi-generator/src/main/resources/csharp-netcore/IReadableConfiguration.mustache
jimschubert: Since 5.0 is released soon, I think the breaking change here will be fine for client release. I still think the JsonSerializerSettings should stay in ApiClient, however, because this isn't API configuration but an implementation detail of ApiClient. Since we allow for users to define their own ApiClients (and bring their own serializers), it doesn't make a lot of sense to put ApiClient implementation-specific details into the Configuration object. That is, this configuration object should refer to `System` namespaces only.

I'd be happy to resolve this in a separate commit.

jimschubert:
@wing328 the author of this PR hasn't responded to the request to tie the commits to the GitHub profile, and the implementation still needs a little work to follow the client's patterns. Namely, some formatting needs to be fixed and the NewtonSoft settings should be moved out of the configuration object into the ApiClient implementation.


Is it acceptable to fix this via separate PR and merge with the author details as-is?

PR #4471, opened 2019-11-12
[ci] Removes elm from ensure-up-to-date

jimschubert:
bin/elm-petstore-all.sh invokes elm-petstore.sh and
elm-0.18-petstore.sh. Both of these define `ELM_POST_PROCESS_FILE` for
post-processing the generated files. If a user doesn't have elm-format
installed, they may not realize that ensure-up-to-date has failed which
causes CI to fail due to differences in the ELM generated outputs.

This confusion can lead to a lot of downtime for contributors. For
example, I encountered this while adding feature set information to all
generators. I thought I had introduced the error and spent too long
looking through my changeset and re-running `ensure-up-to-date`
in the background before noticing the failed output. I was able to
generate proper output by installing elm-format. With 80+ languages/frameworks
and a rule for contributors to unblock CI by re-generating any failed
samples, it's not feasible (in some cases, not possible) to ask
contributors to install tooling specific post-processors. We'll have to
rely on elm contributors to run the script manually.

Ideally, elm generator templates should be updated to have properly
formatted outputs as a default.

We may want to consider documenting standards of what we put in the
scripts under bin/*sh and bin/utils/ensure-up-to-date, one of those
standards being that we omit toolchain specific post-processors.

cc @eriktim 

### PR checklist

### DIFF
--- a/bin/utils/ensure-up-to-date
+++ b/bin/utils/ensure-up-to-date
@@ -68,7 +68,7 @@ declare -a scripts=(
 "./bin/dart-petstore.sh"
 "./bin/dart2-petstore.sh"
 "./bin/java-play-framework-petstore-server-all.sh"
-"./bin/elm-petstore-all.sh"
+#"./bin/elm-petstore-all.sh"
 "./bin/meta-codegen.sh"
 # OTHERS
 "./bin/utils/export_docs_generators.sh"


jimschubert:
For reference, here's what the diff without the formatter look like for me:

Screen Shot 2019-11-12 at 4 19 07 PM

wing328:
👌 we'll figure out other ways to ensure the Elm client is up-to-date with better code formatter.

PR #4482, opened 2019-11-13
[Java/Core] Adds properties to ComposedSchema models when they exist

spacether:
This PR fixes this issue:
https://github.com/OpenAPITools/openapi-generator/issues/4515

Currently, when we process composedSchemas, we are not including any properties that are described in the `properties` section of that model's schema. An example is the below fruit model, which leaves out the color property when we ingest it and make a model.
```
openapi: 3.0.1
info:
   title: fruity
   version: 0.0.1
paths:
   /:
      get:
         responses:
            '200':
               description: desc
               content:
                  application/json:
                     schema:
                        $ref: '#/components/schemas/fruit'
components:
   schemas:
      fruit:
         title: fruit
         properties:
            color:
               type: string
         oneOf:
            - $ref: '#/components/schemas/apple'
            - $ref: '#/components/schemas/banana'
resolved
      apple:
         title: apple
         type: object
         properties:
            kind:
               type: string
      banana:
         title: banana
         type: object
         properties:
            count:
               type: number
```
This spec passes validation at: https://apitools.dev/swagger-parser/online/#

This PR fixes this issue, adding properties to the composed model.
We also add a test in DefaultCodegenTest.java where we check that the fruit model contains the color property.

This issue was also recently noticed and fixed in swagger-codegen at: https://github.com/swagger-api/swagger-codegen/pull/9827

Note: additional properties should also be allowed in composed schemas, but that feature is blocked by an upstream bug in swagger-parser. I've included a commented out additionalproperties property in the fruit spec, commented out code to process it in DefaultCodegen.java. Data in both locations links to the open swagger-parser issue which I created at: https://github.com/swagger-api/swagger-parser/issues/1252

Java Technical Committee

Core Team
@wing328 (2015/07) ❤️
@jimschubert (2016/05) ❤️
@cbornet (2016/05)
@ackintosh (2018/02) ❤️
@jmini (2018/04) ❤️
@etherealjoy (2019/06)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1787,6 +1787,17 @@ public class DefaultCodegen implements CodegenConfig {
             Map<String, Schema> allProperties = new LinkedHashMap<String, Schema>();
             List<String> allRequired = new ArrayList<String>();
 
+            // if schema has properties outside of allOf/oneOf/anyOf also add them to m
+            if (schema.getProperties() != null) {
+                addVars(m, unaliasPropertySchema(schema.getProperties()), schema.getRequired(), null, null);
+            }
+
+            // uncomment this when https://github.com/swagger-api/swagger-parser/issues/1252 is resolved
+            // if schema has additionalproperties outside of allOf/oneOf/anyOf also add it to m
+            // if (schema.getAdditionalProperties() != null) {
+            //     addAdditionPropertiesToCodeGenModel(m, schema);
+            // }
+
             // parent model
             final String parentName = ModelUtils.getParentName(composed, allDefinitions);
             final List<String> allParents = ModelUtils.getAllParentsName(composed, allDefinitions);
--- a/modules/openapi-generator/src/test/resources/3_0/oneOf.yaml
+++ b/modules/openapi-generator/src/test/resources/3_0/oneOf.yaml
@@ -22,6 +22,9 @@ components:
          oneOf:
             - $ref: '#/components/schemas/apple'
             - $ref: '#/components/schemas/banana'
+         # additionalProperties:
+         #    type: string
+         # uncomment this when https://github.com/swagger-api/swagger-parser/issues/1252 is resolved
       apple:
          title: apple
          type: object
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
@@ -269,6 +269,41 @@ public class DefaultCodegenTest {
     }
 
     @Test
+    public void testComposedSchemaOneOfWithProperties() {
+        final OpenAPI openAPI = TestUtils.parseSpec("src/test/resources/3_0/oneOf.yaml");
+        final DefaultCodegen codegen = new DefaultCodegen();
+
+        final Schema schema = openAPI.getComponents().getSchemas().get("fruit");
+        codegen.setOpenAPI(openAPI);
+        CodegenModel fruit = codegen.fromModel("Fruit", schema);
+
+        Set<String> oneOf = new TreeSet<String>();
+        oneOf.add("Apple");
+        oneOf.add("Banana");
+        Assert.assertEquals(fruit.oneOf, oneOf);
+        Assert.assertEquals(fruit.optionalVars.size(), 3);
+        Assert.assertEquals(fruit.vars.size(), 3);
+        // make sure that fruit has the property color
+        boolean colorSeen = false;
+        for (CodegenProperty cp : fruit.vars) {
+            if (cp.name.equals("color")) {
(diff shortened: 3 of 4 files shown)


spacether:
> Looks good to me. It is is identical with my PR #4515. I shall cancel my PR since this is done earlier.
> 
> I believe you should cc the core member instead of java technical committee.

Ah, didn't know that. I just cced them, thanks!

spacether:
Core Team: @wing328 @jimschubert @cbornet @ackintosh @jmini @etherealjoy 
Gentle reminder that this still needs a review
- it fixes https://github.com/OpenAPITools/openapi-generator/issues/4515
- the PR includes a test
- this fixes an issue that swagger-codegen already fixed here https://github.com/swagger-api/swagger-codegen/pull/9827

spacether:
@wing328 these changes were approved 8 days ago. Is there a milestone that they can be added to? Is any more work required in the PR?

modules/openapi-generator/src/test/resources/3_0/oneOf.yaml
```diff
@@ -22,6 +22,9 @@ components:
          oneOf:
             - $ref: '#/components/schemas/apple'
             - $ref: '#/components/schemas/banana'
+         # additionalProperties:
+         #    type: string
+         # uncomment this when https://github.com/swagger-api/swagger-parser/issues/1252 is resolved
```
jimschubert: I wouldn't imagine this will get fixed anytime soon in swagger-parser for the 3.0 specification (the revision versions 3.0.1/3.0.2 are supposedly documentation-only changes).

The behavior you're seeing where this returns `MapSchema` is defined behavior: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#model-with-mapdictionary-properties

The requested behavior is undefined, as there's not enough clarity in the specification around constraints of using `oneOf`. It _should_ be that `oneOf` can only coexist with `discriminator`.

`oneOf` is taken from JSON Schema's specification [Section 9.2.1.3](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.2.1.3), which says:

> This keyword's value MUST be a non-empty array. Each item of the array MUST be a valid JSON Schema.
> 
> An instance validates successfully against this keyword if it validates successfully against exactly one schema defined by this keyword's value.

Meaning your example is only a valid object if it is exactly `apple` or exactly `banana`, and no dynamic properties via `additionalProperties` are allowed due to the use of `MUST` rather than `SHOULD` in the JSON Schema specification.

The `discriminator` support for `oneOf` is a JSON Schema extension defined in the OAS [Composition and Inheritance (Polymorphism)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#composition-and-inheritance-polymorphism) section.

Ideally, a spec with `oneOf` any _anything_ besides discriminator should fail validation. We could consider adding this strict validation to OpenAPI Generator, or at least as a recommendation via the `--recommend` flag to the validate command.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1787,6 +1787,17 @@ public CodegenModel fromModel(String name, Schema schema) {
             Map<String, Schema> allProperties = new LinkedHashMap<String, Schema>();
             List<String> allRequired = new ArrayList<String>();
 
+            // if schema has properties outside of allOf/oneOf/anyOf also add them to m
+            if (schema.getProperties() != null) {
+                addVars(m, unaliasPropertySchema(schema.getProperties()), schema.getRequired(), null, null);
+            }
+
+            // uncomment this when https://github.com/swagger-api/swagger-parser/issues/1252 is resolved
```
jimschubert: See my comment on the example spec document in this PR.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1787,6 +1787,17 @@ public CodegenModel fromModel(String name, Schema schema) {
             Map<String, Schema> allProperties = new LinkedHashMap<String, Schema>();
             List<String> allRequired = new ArrayList<String>();
 
+            // if schema has properties outside of allOf/oneOf/anyOf also add them to m
+            if (schema.getProperties() != null) {
+                addVars(m, unaliasPropertySchema(schema.getProperties()), schema.getRequired(), null, null);
```
jimschubert: @wing328 See my [other comment](https://github.com/OpenAPITools/openapi-generator/pull/4482/files#r353522507) in this PR. Constraints around structures containing `oneOf` aren't very clearly defined in OAS 3.x, since the extension to include `discriminator` makes it look like any Schema properties are valid.

Do you think we should log a warning here, or wrap it the a ! strict spec check, or both? I've commented elsewhere as well that we might want to add to the validate and/or recommends about this.

jimschubert:
> This spec passes validation at: https://apitools.dev/swagger-parser/online/#

Just a heads up that our CLI also validates:

```
openapi-generator validate -i openapi.yml
```

Issue #4389, opened 2019-11-06
[BUG][Python] Configuration objects share a common api_key dict


arbelpeled:
##### Description

When creating multiple Configurations using `Configuration()`, the dictionary in the field `api_key` is shared across these multiple (different) configurations.

##### openapi-generator version

3.3.3

##### OpenAPI declaration file content or url

Doesn't matter which declaration

##### Steps to reproduce

In any generated API:

```
c1 = Configuration()
c1.api_key['x-auth-token']="mytoken"
c2 = Configuration()
assert(c2.api_key['x-auth-token']=="mytoken")
```

The assertion should fail but it actually does not.

##### Suggest a fix

It's probably a problem with the default value supplied to the constructor of `Configuration`. If you supply a `{}` default value the same dict will be shared between different invocations of the constructor. The standard way to solve this is by putting a default `None` value and if the value is indeed `None` then create a new `{}` inside the constructor.

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/4485

Issue #3577, opened 2019-08-07
[BUG] [Python] Default configuration gives unexpected behaviour


TimoMorris:
##### Description

After it has been called once, the `Configuration` class constructor effectively ignores the parameters passed to it on subsequent calls, and instead returns a copy of the configuration created the first time it was called.

##### openapi-generator version

v4.0.1 (this is the version I was using, and also where the issue first appears)

##### OpenAPI declaration file content or url

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/configuration.mustache

##### Command line used for generation

```cmd
java -jar openapi-generator-cli-4.0.1.jar generate -i my-spec.json -g python -o .\my_package -DprojectName=my-project -DpackageName=my_package
```

##### Steps to reproduce

Once you've generated the api client package as above, run:

```python
import my_package

config1 = my_package.Configuration(host="https://my-first-site.com")
config2 = my_package.Configuration(host="https://a-different-site.com")

print(config1.host)  # https://my-first-site.com
print(config2.host)  # https://my-first-site.com
```

Or alternatively:

```python
import my_package

config1 = my_package.Configuration()  # this call uses the default host from the spec
config2 = my_package.Configuration(host="https://a-different-site.com")

print(config1.host)  # https://default-host-from-spec.com
print(config2.host)  # https://default-host-from-spec.com
```

The two hosts for the two different configurations should be different.

[**Note:** it may seem strange to want to set different hosts for one API, but the software I'm working with doesn't have a single central API, but rather allows users to set up their own instance(s) of the API. Each user's API has the same spec, except for being hosted wherever the user has set it up. I'm working to develop tools that any user could use, but obviously they have to provide the host to begin with.]

##### Related issues/PRs

* [this commit](https://github.com/OpenAPITools/openapi-generator/commit/d2b91073e1fc499fea67141ff4c17740d25f8e83#diff-24d407589f9c6ea1733ad8ed80a75d66) - this commit introduced the `TypeWithDefault` class which `Configuration` inherits from.
* #3002 (for issue #3003) - this added some parameters to the Configuration constructor, which is a great enhancement, but the way it interacts with the `TypeWithDefault` parent class leads to the above unexpected and undesired behavior.

##### Suggest a fix/enhancement

It depends what the purpose of `TypeWithDefault` is supposed to be. I can see two potential purposes:

[...]

PR #4485, opened 2019-11-14
[Python] each Configuration instance now has its own dicts

spacether:
Per this issue: https://github.com/OpenAPITools/openapi-generator/issues/4389
The dictionaries used in Configuration instances are shared across instances.
This is not the intended function.
Each instance should use its own unique dict to store api_key and api_key_prefix
This PR makes that update and adds a test at samples/client/petstore/python/tests/test_configuration.py for verification

Related issue that may be closed by this PR:
- https://github.com/OpenAPITools/openapi-generator/issues/3577

### PR checklist

Python committee
@taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @slash-arun (2019/11)


samples/client/petstore/python/tests/test_configuration.py
```diff
@@ -0,0 +1,36 @@
...
+$ nosetests -v
+"""
+from __future__ import absolute_import
+
+import unittest
+
+import petstore_api
+
+
+class TestConfiguration(unittest.TestCase):
```
arbelpeled: Nice

TimoMorris:
Thanks for fixing this! 😃🚀

wing328:
@spacether thanks for the PR, which has been included in the v4.2.2 release: https://twitter.com/oas_generator/status/1201432648544972800

tomplus:
This is a breaking change. Some libraries use the feature of creating default configuration and this PR removed it. Method `Configuration.set_default(cls, default)` disappeared. I'll try to restore it in a different way.

spacether:
Sorry that we broke this functionality. We look forward to your PR.
Can you share more about your use case?
have you considered these solutions:
- storing your configuration inputs in environment variables
- making one configuration instance in a module, then importing it into all the modules which need it

TimoMorris:
@tomplus Sorry to hear we've caused you difficulties 😞

FWIW I did quite a detailed write-up of the issues I found with the way the behavior around default configurations when I raised issue #3577 (essentially the same as this issue), which you might find helpful to refer to.

Essentially, it wasn't only caching a copy of the first configuration instantiated to use as a default, but it would then *always* return this default, even if `Configuration` was called with other parameters explicitly passed. So it was over-strongly enforcing the default.

It seemed to me the expected behavior would be to (at most) cache the first configuration as a default and return this if you called `Configuration()` with no parameters. Though even that might be unexpected for some users. But to have the class silently ignore different parameters explicitly passed on subsequent calls led to some very confusing and hard-to-trace behavior.

I imagine it would be relatively straightforward to re-implement a `set_default()` method, or perhaps have a keyword-only argument e.g. `set_as_default` for the `__init__()` method. But however it's implemented, it seems to me that in the Pythonic spirit of "explicit is better than implicit" the user should have to opt-in to it, rather than the default-setting/default-getting behavior being 'on' by default.

And I'd agree with @spacether's suggestions that the user implementing the defaulting behaviour in their own code seems better.

tomplus:
Thanks for the discussion.

I know that it can be implemented in many ways but for me previous implementation was very handy (apart from bugs which were found). I have one server with many API exposed. I was able to create a well-prepared default configuration on startup and then used it without worrying about it.

I think the static method set_default() doesn’t break the principle “explicit is better than implicit” because I have to explicitly set configuration as default configuration. 

I'm sending a PR soon.

tomplus:
Here is my PR https://github.com/OpenAPITools/openapi-generator/pull/5315, Thanks for reviewing.

Issue #4328, opened 2019-10-30
[BUG][PHP] Setters for referenced enums are unusable


hinrik:
I'm using version 4.1.3.

Consider this definition:
```yaml
components:
  schemas:
    Foo:
      type: object
      properties:
        inline_enum:
          type: string
          enum: [one, two, three]
        referenced_enum:
            $ref: '#/components/schemas/ReferencedEnum'
    ReferencedEnum:
      type: string
      enum: [one, two, three]
```

For `inline_enum`, `Foo`'s `setInlineEnum()` method accepts a `string`, the enum values are constants on the `Foo` class, and the value being set is validated to be one of the constants.

However, for `referenced_enum`, `setReferencedEnum()` accepts a `ReferencedEnum` object. But that class only has the constants along with a `getAllowableEnumValues()` function. It carries no state, making it impossible to call `setReferencedEnum()` with a meaningful value.

I've seen that at least the Java generator handles the `referenced_enum` case correctly, as it uses a real `enum` type just like it does for `inline_enum`.

I would very much like to use the `referenced_enum` style in my project because it would allow using the same enum definition in multiple places. This is especially handy when one of the consumers of the API definition is a Java application, where this would mean a single `enum` class instead of multiple duplicate ones where you might have to convert values between each.

Was this just an oversight? If so I can try to whip up a PR to make `setReferencedEnum()` accept a string, and include the same validation that `setInlineEnum()` has (but using the values from the `ReferencedEnum` class).

PR #4488, opened 2019-11-14
[PHP] Support referenced enums

hinrik:
Resolves issue #4328 and allows reusing enums in definitions consumed by
PHP applications.

Without this change, referenced enums are unusable in PHP (and possibly
other languages which don't have native enums). This is because the
setter for such a property would expect an instnace of the external enum
class as the value, and that class has no state (and was therefore a
meaningless value).

On properties which reference enum schemas, we now set `referencedEnumType`
which contains the data type of the enum value. This allows languages
which don't have native enums to handle referenced enums correctly.

I adjusted the PHP model template to make use of this new variable, so
that it now generates very similar code for referenced enums as it does
for inline ones: the setter accepts a primitive type and validates the
value. The only difference is that the allowed values are stored in the
external enum class rather than inline on the model.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenProperty.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenProperty.java
@@ -58,6 +58,7 @@ public class CodegenProperty implements Cloneable {
             isBoolean, isDate, isDateTime, isUuid, isUri, isEmail, isFreeFormObject;
     public boolean isListContainer, isMapContainer;
     public boolean isEnum;
+    public String referencedEnumType;
     public boolean isReadOnly;
     public boolean isWriteOnly;
     public boolean isNullable;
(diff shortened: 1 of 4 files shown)


hinrik:
FYI @jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @ybelenko, @renepardon

ybelenko:
I would take a look at this PR. 

@hinrik can you rebase on latest master and resolve merge conflicts? Or it's not relevant already?

hinrik:
Yeah I can do that.

etremblay:
This feature would be awesome. psalm and phpstorm always complain about badly typed enum arguments generated by the php generator.

Cubid:
I have done the rebase work as of today https://github.com/Cubid/openapi-generator/tree/php_fix_referenced_enums
If any one is interested to go further, she can use it.

Issue #4403, opened 2019-11-07
[BUG][PHP] Model name suffix incorrectly added to inline enum names


hinrik:
Consider this definition:

```yaml
components:
  schemas:
    Foo:
      type: object
      properties:
        inline_enum:
          type: string
          enum: [one, two, three]
        referenced_enum:
            $ref: '#/components/schemas/ReferencedEnum'
    ReferencedEnum:
      type: string
      enum: [one, two, three]
```

If you generate PHP from this defintion using a `modelNameSuffix=DTO` option on the generator, the values of `inline_enum` will now change from being e.g. `INLINE_ENUM_ONE` to `INLINE_ENUM_DTO_ONE`, which doesn't make sense.

Values of `ReferencedEnum` are untouched, still `ReferencedEnum::ONE` (though they are unusable regardless [for another reason](https://github.com/OpenAPITools/openapi-generator/issues/4328)).

ackintosh:
@hinrik Thanks for the PR! It has been merged into master. 🎉

PR #4489, opened 2019-11-14
[PHP] Remove model prefix/suffix from inline enum var names

hinrik:
This resolves issue #4403.

On a model `Foo` with an inline enum `Bar` which has a possible string
value `baz`. we should generate a `FooDTO::BAR_BAZ` constant, not
`FooDTO::BAR_DTO_BAZ`.


hinrik:
FYI @jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @ybelenko, @renepardon

wing328:
@hinrik thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #4500, opened 2019-11-15
Typescript refactor: Platform select for browser and node

bodograumann:
This adds a new option to select the platform as node or browser.
When no platform is selected, a default is chosen by the framework option and likewise the file data type option is implied by the platform.

I did not have time to overhaul the tests, so I currently left the *default* sample build as node version and the *jquery* sample build as browser, as the latter uses proper e2e tests.

This PR assumes that my [previous PR](https://github.com/OpenAPITools/openapi-generator/pull/4424) is merged, so currently those commits are listed as well.

I also included some other minor improvements and fixes.

Cf. #802 
CC @TiFu

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
@@ -707,12 +707,7 @@ public class TypeScriptClientCodegen extends DefaultCodegen implements CodegenCo
             setModelPropertyNaming((String) additionalProperties.get(CodegenConstants.MODEL_PROPERTY_NAMING));
         }
 
-        if (additionalProperties.containsKey(CodegenConstants.SUPPORTS_ES6)) {
-            // convert to boolean
-            additionalProperties.put(CodegenConstants.SUPPORTS_ES6,
-                Boolean.valueOf(additionalProperties.get(CodegenConstants.SUPPORTS_ES6).toString())
-            );
-        }
+        convertPropertyToBooleanAndWriteBack(CodegenConstants.SUPPORTS_ES6);
 
         // change package names
         apiPackage = this.apiPackage + ".apis";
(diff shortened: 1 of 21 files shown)


TiFu:
I just merged #4424 and it seems like this PR also included the changes from that PR. Could you please resolve the conflicts?

bodograumann:
Ok, have cherry picked all the new commits onto the merged typescript-refactor.

bodograumann:
@TiFu Do you think you can merge this? ;-)

TiFu:
Done! Sorry for the delay :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
```diff
@@ -142,6 +145,14 @@ public TypeScriptClientCodegen() {
         }
         frameworkOption.defaultValue(FRAMEWORKS[0]);
 
+        cliOptions.add(new CliOption(TypeScriptClientCodegen.PLATFORM_SWITCH, TypeScriptClientCodegen.PLATFORM_SWITCH_DESC));
```
TiFu, line 148: @bodograumann Is it possible that there's a bug here? First a new `CliOption` object is created and added to `cliOptions`, then another `CliOption` is created, which isn't added to `cliOptions`.

bodograumann: I think it is fine. The framework option is added to cliOptions further down:
```cliOptions.add(frameworkOption);```

TiFu: But `platformOption`/`PLATFORM_SWITCH` (see l.148, l. 149-154) isn't or am i blind?

bodograumann: Ah now I see it. Yes I guess there is some redundant code.
Creating the duplicate CliOption object doesn't seem to cause any problems ^^.
Still, good catch, I should fix that.

Issue #4072, opened 2019-10-05
[BUG][SCALAZ] tests are failing with generator 4.2.0


Fjolnir-Dvorak:
#### Bug Report Checklist

- [x] What's the version of OpenAPI Generator used?
       Latest master (4.2.0-SNAPSHOT)
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
```
[error] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/scalaz/src/main/scala/org/openapitools/client/api/PetApi.scala:68: not found: type ListBuffer
[error]   def findPetsByStatus(host: String, status: List[String] = new ListBuffer[String]() )(implicit statusQuery: QueryParam[List[String]]): Task[List[Pet]] = {
[error]                                                                 ^
[error] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/scalaz/src/main/scala/org/openapitools/client/api/PetApi.scala:89: not found: type ListBuffer
[error]   def findPetsByTags(host: String, tags: List[String] = new ListBuffer[String]() )(implicit tagsQuery: QueryParam[List[String]]): Task[List[Pet]] = {
[error]                                                             ^
[error] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/scalaz/src/main/scala/org/openapitools/client/api/PetApi.scala:235: not found: type ListBuffer
[error]   def findPetsByStatus(status: List[String] = new ListBuffer[String]() )(implicit statusQuery: QueryParam[List[String]]): Task[List[Pet]] = {
[error]                                                   ^
[error] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/scalaz/src/main/scala/org/openapitools/client/api/PetApi.scala:256: not found: type ListBuffer
[error]   def findPetsByTags(tags: List[String] = new ListBuffer[String]() )(implicit tagsQuery: QueryParam[List[String]]): Task[List[Pet]] = {
[error]                                               ^
[warn] two warnings found
[error] four errors found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 16 s, completed Oct 5, 2019 6:14:02 PM
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
	at org.apache.commons.exec.DefaultExecutor.executeInternal(DefaultExecutor.java:404)
[...]

PR #4508, opened 2019-11-17
Fixes issues with Scalaz outputs

jimschubert:
Reported in #4072, discussed in #4454

Scalaz is not re-generated by users or CI (only verified), so some
compilation issues have been introduced into the generator.
Specifically, the generator previously didn't handle defaults well (or
even correctly, maybe?). This broke when array models were added to the
openapi document used to generate samples. The inclusion of Date-related
mappings being mapped to joda time in DefaultCodegen also caused some
issues with new DateTime properties on models.

Over the course of what appears to be Nov 10-17 2019, CircleCI seems to
be having intermittent issues with Scalaz verification. I found that
green builds were picking up SBT 0.13.x and failed builds were SBT
1.1.0. It's not clear where the system level SBT is being defined, but a
simple fix has been to enforce the sbt version in the generator.

For those unfamiliar with SBT; the SBT command acts as a launcher script
which may switch to older/newer versions of SBT. A Scala project invoked
with SBT 1.1.0 will look for an sbt.version override and happily attempt
compilation with the available SBT 1.1.0. The problem is that SBT 1.1.0
uses Scala 2.12 and this is not binary compatible with Scala 2.11. This
can cause issues with builds due to plugins or incompatible Java version.

### PR checklist


jimschubert:
FYI regarding scalaz change @clasnake, @shijinkui, @ramzimaalej
FYI regarding compilation fix @Fjolnir-Dvorak

PR #4527, opened 2019-11-18
[Java][Go] Unalias schema in getTypeDeclaration when not generating model for alias

### LINKED ISSUE #4362: [BUG][Java][Go-experimental] Referenced schema for `array` type badly generated
#### Bug Report Checklist

##### Description

Defining a schema for an array type, and referencing it as the `items` type in another array doesn't generate valid code. It happened for me in Java and go-experimental that I'm using, but might be the case for other languages as well.

4.1.3

The following schemas definition exhibit the behavior 
```yaml
NestedArray:
  type: array
  items:
    type: string
MainSchema:
  type: object
  properties:
    nested_array:
      type: array
      items:
        $ref: '#/components/schemas/NestedArray'
``` 

In go, the generated output looks like this, which doesn't even compile since `Array` is undefined:
```go
type MainSchema struct {
	NestedArray *[]Array `json:"nested_array,omitempty"`
}
```

instead of the expected output of either
```go
type MainSchema struct {
	NestedArray *[]NestedArray `json:"nested_array,omitempty"`
}
```
if we generated a model for the `NestedArray` schema, or simply
```go
type MainSchema struct {
	NestedArray *[][]string `json:"nested_array,omitempty"`
}
```
if we don't generate a model (which it seems is what is supposed to happen when we create a schema for a primitive type).

In java, the output is
```java
public class MainSchema {
  public static final String JSON_PROPERTY_NESTED_ARRAY = "nested_array";
  private List<List> nestedArray = null;
}
```
instead of 
```java
public class MainSchema {
  public static final String JSON_PROPERTY_NESTED_ARRAY = "nested_array";
  private List<List<String>> nestedArray = null;
}
```

Please note that the correct output is generated when inlining the `NestedArray` definition instead of referencing it in the `MainSchema`, like so:
```yaml
MainSchema:
  type: object
  properties:
    nested_array:
      type: array
      items:
        type: array
        items:
          type: string
``` 

openapi-generator generate --http-user-agent DataDog/1.0.0/java -g java -c config/languages/java_v1.json -i spec/v1/full_spec.yaml -o generated/datadog-api-client-java/v1 -t templates/java'

##### Steps to reproduce

Make a spec with the above schema definitions, and inspect the generated model for `MainSchema`

zippolyte:
This is my attempt to fix #4362 for the two languages that I'm having this issue with.
What do you think ?

@bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger @karismann @Zomzog @lwlee2608

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
@@ -278,10 +278,16 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
         if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return "[]" + getTypeDeclaration(inner);
+            if (ModelUtils.isGenerateAliasAsModel())
+                return "[]" + getTypeDeclaration(inner);
+            else
+                return "[]" + getTypeDeclaration(ModelUtils.unaliasSchema(this.openAPI, inner));
         } else if (ModelUtils.isMapSchema(p)) {
             Schema inner = ModelUtils.getAdditionalProperties(p);
-            return getSchemaType(p) + "[string]" + getTypeDeclaration(inner);
+            if (ModelUtils.isGenerateAliasAsModel())
+                return getSchemaType(p) + "[string]" + getTypeDeclaration(inner);
+            else
+                return getSchemaType(p) + "[string]" + getTypeDeclaration(ModelUtils.unaliasSchema(this.openAPI, inner));
         }
         //return super.getTypeDeclaration(p);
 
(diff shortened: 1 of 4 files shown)


wing328:
@zippolyte thanks for the PR. I think it's better to fix it in the DefaultCodegen. I'll try to fix it later this week.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
```diff
@@ -278,10 +278,16 @@ public String getTypeDeclaration(Schema p) {
         if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return "[]" + getTypeDeclaration(inner);
+            if (ModelUtils.isGenerateAliasAsModel())
```
wing328: @zippolyte I don't think you need `ModelUtils.isGenerateAliasAsModel()` as `unaliasSchema` has already taken that into account:  https://github.com/OpenAPITools/openapi-generator/blob/10627ed5cc3f4703b0f02662ea568353e8d26e00/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L812

wing328:
UPDATE: I've revised your fix and I think it's the right approach. I've left a comment suggesting a minor change. Please review when you've time. Thanks again for the PR.

zippolyte:
Indeed it seems I don't need that after all. Thanks for your review, updating.

wing328:
@zippolyte thanks for the PR, which has been merged into master. I've filed https://github.com/OpenAPITools/openapi-generator/issues/4563 to track the fixes for other generators.

wing328:
@zippolyte thanks for the PR, which has been included in the v4.2.2 release: https://twitter.com/oas_generator/status/1201432648544972800

PR #4546, opened 2019-11-19
[Go] Bypass object decode in case of empty body

### LINKED ISSUE #4545: [BUG][Go] Go client decode returns error for empty body response
HTTP 204 does not consider a response body decode method for APIClient object does not check that response is empty so json.Unmarshall returns error ins this case which is not correct behaviour

dukov:
Fixes #4545


dukov:
Could you please take a look?
@antihax 
@bvwells 
@grokify 
@kemokemo 
@bkabrda

wing328:
@dukov thanks for the PR, which has been included in the v4.2.2 release: https://twitter.com/oas_generator/status/1201432648544972800

PR #4547, opened 2019-11-19
Use bin/cpp-qt5-petstore.json

MartinDelille:
This PR use the new ``--config`` option in the `bin/cpp-qt5-petstore.sh` script.

### PR checklist

@ravinikam @stkrwork @etherealjoy @muttleyxd @jimschubert

### DIFF
--- a/bin/cpp-qt5-petstore.json
+++ b/bin/cpp-qt5-petstore.json
@@ -1,0 +1,8 @@
+{
+  "inputSpec": "modules/openapi-generator/src/test/resources/2_0/petstore.yaml",
+  "outputDir": "samples/client/petstore/cpp-qt5",
+  "additionalProperties": {
+    "cppNamespace": "test_namespace",
+    "modelNamePrefix": "PFX"
+  }
+}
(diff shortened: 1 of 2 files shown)


wing328:
No worry. Let me try to fix it.

wing328:
Fixed via 96bbab98a4

We'll find a better way to catch such issue moving forward.

MartinDelille:
I would be cool to have a job that launch the generating scripts, check if they fails and check if there is some diff with `git diff --exit-code`.

Is there an issue about it or shall I create it?

wing328:
We've that in the CI already but the problem with missing config file is that the codegen process still proceeds with default config and completes successfully (exit status code: 0) so it's hard to catch.

We'll instead throw a Runtime exception similar to what we've done for other file input.

jimschubert:
@wing328 we should explicitly reference the root directory. Many of our scripts rely on the original execution path.

wing328:
> Many of our scripts rely on the original execution path.

That's what it's from day 1 I started contributing to this project. Agreed with your suggestion but it will be a lot of work that I would rather suggest spend on other tasks with higher priority instead.

PR #4553, opened 2019-11-20
Add isCircularReference to properties

eriktim:
As a preparation to a major refactoring of the Elm generator I would like to add the ability of detecting circular references. Support for _self_ referring types was already in place, but this PR also detects types that _indirectly_ refer back to themselves. Every property that indirectly refers back to itself gets marked with `isCircularReference`.

### PR checklist

cc @wing328 @jimschubert @cbornet @ackintosh @jmini @etherealjoy

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
@@ -347,6 +347,7 @@ public class ElmClientCodegen extends DefaultCodegen implements CodegenConfig {
                 });
             }
         }
+        setCircularReferences(allModels);
         for (Map.Entry<String, Object> entry : objs.entrySet()) {
             Map<String, Object> inner = (Map<String, Object>) entry.getValue();
             List<Map<String, Object>> models = (List<Map<String, Object>>) inner.get("models");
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ElmClientCodegen.java
```diff
@@ -347,6 +347,7 @@ public int compare(CodegenModel cm1, CodegenModel cm2) {
                 });
             }
         }
+        setCircularReferences(allModels);
```
wing328: Can we add this to the default codegen instead as users would expect `isCircularReference` with a valid value (true, false) for all generators?

eriktim: @wing328 that was my intention, that's why I added it to `updateAllModels` in `DefaultCodegen`. Somehow this did got not triggered in Elm, so I thought I'd probably overridden it. Is `updateAllModels` the correct place to put this?

wing328: Can you please move it to updateAllModels and I'll run some tests later to troubleshoot?

eriktim: @wing328 I already put it in `updateAllModels`. Do I need to remove it from the Elm codegen?

PR #4566, opened 2019-11-21
Add support for dumping request and response in Go generated clients

dnaeon:
The following change adds a new configuration setting, which
controls whether clients want to dump the HTTP request and response.

Useful when debugging API calls and clients.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/go/client.mustache
+++ b/modules/openapi-generator/src/main/resources/go/client.mustache
@@ -11,6 +11,7 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -161,7 +162,25 @@ func parameterToJson(obj interface{}) (string, error) {
 
 // callAPI do the request.
 func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
-	return c.cfg.HTTPClient.Do(request)
+	if c.cfg.Debug {
+	        dump, err := httputil.DumpRequestOut(request, true)
+		if err != nil {
+		        return nil, err
+		}
+		fmt.Println(string(dump))
+	}
+
+	resp, err := c.cfg.HTTPClient.Do(request)
+
+	if c.cfg.Debug {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			return resp, err
+		}
+		fmt.Println(string(dump))
+	}
+
+	return resp, err
 }
 
 // ChangeBasePath changes base path to allow switching to mocks
--- a/modules/openapi-generator/src/main/resources/go/client.mustache
+++ b/modules/openapi-generator/src/main/resources/go/client.mustache
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"mime/multipart"
 	"net/http"
 	"net/http/httputil"
@@ -167,7 +168,7 @@ func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
 		if err != nil {
 		        return nil, err
 		}
-		fmt.Println(string(dump))
+		log.Printf("\n%s\n", string(dump))
 	}
 
 	resp, err := c.cfg.HTTPClient.Do(request)
@@ -177,7 +178,7 @@ func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
 		if err != nil {
 			return resp, err
 		}
-		fmt.Println(string(dump))
+		log.Printf("\n%s\n", string(dump))
 	}
 
 	return resp, err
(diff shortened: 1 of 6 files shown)


dnaeon:
CC @antihax, @bvwells, @grokify, @kemokemo, @bkabrda

wing328:
@dnaeon thanks for the PR. What about using a logger (e.g. https://github.com/google/logger) instead?

dnaeon:
Hey @wing328, 

I've tested it with a logger first, but the results weren't pretty, that's why I've left it using just `fmt.Println`.

antihax:
This should really be done with middleware, not in the client directly. Is the a reason it cant be?

wing328:
> I've tested it with a logger first, but the results weren't pretty, that's why I've left it using just fmt.Println.

Can you please elaborate?

wing328:
FYI. For other clients (e.g. Ruby), we also use a logger for debugging purpose.

dnaeon:
> > I've tested it with a logger first, but the results weren't pretty, that's why I've left it using just fmt.Println.
> 
> Can you please elaborate?

@wing328 Here's one example.

https://play.golang.org/p/o5LE9R5iCpj

Basically the timestamp of the logger is attached to just the first line of the output, which makes it less pretty, IMO.

Also, I found `fmt.Println` to be less distractive for my use cases, but I can switch that over to a logger, if you think that's better.

> This should really be done with middleware, not in the client directly. Is there a reason it cant be?

@antihax That should work I guess. Can you point me to an example middleware I can lookup?

Thanks!

antihax:
Here is an example of two chained middleware for the API client.
Setup of the chain `LimitedTransport` and `APICacheTransport`.
https://github.com/antihax/evedata/blob/master/internal/apicache/apicache.go#L50

Here are the implementations:
https://github.com/antihax/evedata/blob/master/internal/apicache/limitedTransport.go
https://github.com/antihax/evedata/blob/master/internal/apicache/transport.go#L20

You would create another of these that would dump the request, then a second to dump the response.

In the case here you would do something like this:

```
func (t *DumpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	reqb, _ := httputil.DumpRequest(req, true)
	res, err:= t.Transport.RoundTrip(req)
	resb, _ := httputil.DumpRequest(res, true)
        ... log the request and response ...
	return res, err
}```

wing328:
> Basically the timestamp of the logger is attached to just the first line of the output, which makes it less pretty, IMO.

What about putting a "\n" at the beginning of the log?

dnaeon:
> Here is an example of two chained middleware for the API client.
> Setup of the chain `LimitedTransport` and `APICacheTransport`.
> https://github.com/antihax/evedata/blob/master/internal/apicache/apicache.go#L50
> 
> Here are the implementations:
> https://github.com/antihax/evedata/blob/master/internal/apicache/limitedTransport.go
> https://github.com/antihax/evedata/blob/master/internal/apicache/transport.go#L20
> 
> You would create another of these that would dump the request and response.
> 
> In the case here you would do something like this:
> 
> ```
> func (t *DumpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
> 	reqb, _ := httputil.DumpRequest(req, true)
> 	res, err:= t.Transport.RoundTrip(req)
> 	resb, _ := httputil.DumpResponse(res, true)
>         ... log the request and response ...
> 	return res, err
> }
> ```

@antihax Thanks, that would work.


> > Basically the timestamp of the logger is attached to just the first line of the output, which makes it less pretty, IMO.
> 
> What about putting a "\n" at the beginning of the log?

@antihax, @wing328 I'll leave it to you whether it's worth merging this PR considering that we can have the same thing done with either a middleware or custom transport.

wing328:
@dnaeon the debug flag in other clients can be used not only for dumping request/response but also other troubleshooting (e.g. object serialization, etc).

so if you're using a logger with the debug flag, I'll be happy to merge this into master.

dnaeon:
> @dnaeon the debug flag in other clients can be used not only for dumping request/response but also other troubleshooting (e.g. object serialization, etc).
> 
> so if you're using a logger with the debug flag, I'll be happy to merge this into master.

@wing328, I can take care of updating the PR.

Which logger you'd prefer to have as part of the implementation?

https://github.com/google/logger or the builtin `log.Logger`?

wing328:
Please start with the built-in one. Thanks


在 2019年11月23日週六 下午3:49，Marin Atanasov Nikolov <notifications@github.com> 寫道：

dnaeon:
@wing328 just pushed a new commit which uses `log.Logger`.

wing328:
@dnaeon thanks for the PR, which looks good to me.

wing328:
@dnaeon when you've time, I wonder if you can file the same PR for `go-experimental` generator as well. Thank you.

dnaeon:
@wing328 Sure, here it is #4649

PR #4569, opened 2019-11-21
 Use model as body param for generateAliasAsModel

### LINKED ISSUE #3602: [BUG][resttemplate] Object with additionalProperties (HashMap) does not generate a model
##### Description

I am trying to generate a HashMap class, but no class is generated and a plain Java Map is used where the class `ContextHeaders` should be used. 

```
  @JsonProperty("contextHeaders")

  private Map contextHeaders = null;
```

4.0.3

[swagger.json](https://gist.github.com/Na3blis/93dbca819faf2598fdae5f595dbf66e9)
[swagger-client-config.json](https://gist.github.com/Na3blis/b175b9db83fde16f3229708416c102a1)

`curl -s -o ./codegen.jar http://central.maven.org/maven2/org/openapitools/openapi-generator-cli/4.0.3/openapi-generator-cli-4.0.3.jar`
`java -jar ./codegen.jar generate -c ./swagger-client-config.json -i ./swagger.json -g java -o generated-client/`

##### Steps to reproduce

Run the codegen 4.0.3 against the swagger.json file with the config swagger-client-config.json. Notice there is no `com.example.authrules.v1.client.model.ContextHeaders` class, and `com.example.authrules.v1.client.model.RulesUser` references a plain Java Map instead of a class

bodograumann:
Fixes #3602:

When enabling `generateAliasAsModel`, schemes which are aliases for maps or arrays are generated as models by all templates. Those models where previously not imported and used for the relevant body parameters though.

I have not added any tests, but for my own use case the change works and the one sample where the `generateAliasAsModel` option is activated also shows the change in effect.

### Example
Considering the following OpenAPI definition.
```yaml
openapi: 3.0.2
info:
    title: Map Alias Example
    version: 0.1.0
    description: Use a data scheme which is only an alias for a map.
paths:
    '/users/{userId}/roles':
        patch:
            requestBody:
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/UserRoleUpdateRequest'
                required: true
            parameters:
                -
                    name: userId
                    schema:
                        type: string
                    in: path
                    required: true
            responses:
                '200':
                    description: Update successfull.
            operationId: updateUserRoles
components:
    schemas:
        UserRoleUpdateRequest:
            type: object
            additionalProperties:
                type: boolean
            example:
                admin: true
                user: false
```

Usually it would generate typescript-inversify code:
```typescript
  public updateUserRoles(
    userId: string,
    requestBody: { [key: string]: boolean },
    observe?: "body",
    headers?: Headers
  ): Promise<any>;
```

Instead it should generate:
```typescript
  public updateUserRoles(
    userId: string,
    userRoleUpdateRequest: UserRoleUpdateRequest,
    observe?: "body",
    headers?: Headers
  ): Promise<any>;
```


bodograumann:
Could you please have a look at this, @wing328?

spacether:
@bodograumann
This is a very helpful PR. Thank you for filing it.
When working on the Python-experimental generator I also saw this problem.
Does your fix also cover operation parameters not in the body?
For example a string alias model used in a header. 
Can you add tests in java verifying that an alias model in the body and in another location also point to the correct alias model?

bodograumann:
I think this is only concerned with the body parameter.
In principal I could add more tests and also cover other kinds of parameters, but I’m not sure when I’ll find the time for that, @spacether.

spacether:
Limiting it to only body parameters now works. Are you open to help? If you added me as a contributor I could add a java test for this update on this branch.

bodograumann:
Like this? https://github.com/bodograumann/openapi-generator/invitations

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4823,7 +4823,11 @@ public CodegenParameter fromRequestBody(RequestBody body, Set<String> imports, S
         }
         schema = ModelUtils.getReferencedSchema(this.openAPI, schema);
 
-        if (ModelUtils.isMapSchema(schema)) {
+        boolean useAliasReferenceName = ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name) &&
+          (ModelUtils.isMapSchema(schema) || ModelUtils.isArraySchema(schema)) &&
+          (schema.getProperties() == null || schema.getProperties().isEmpty());
+
```
spacether: @bodograumann what if the alias schema is a primitive like string/integer etc?
Why not allow those?
This update currently limits it to mapSchemas or ArraySchemas.

spacether: Why aren't we using model.isAlias?

spacether: Investigation into this led to this alternate PR:
https://github.com/OpenAPITools/openapi-generator/pull/4597
@bodograumann can you fully describe the models that you need adjusted?
It looks like they are not the primitive non-object non-array models that I am writing about.

bodograumann:
Thanks for your feedback @spacether. According to what I read, the option `generateAliasAsModel` only creates aliases for map and array models. E.g. 

https://github.com/OpenAPITools/openapi-generator/blob/d9bde12ebbb9b2c3a695edbd1da8fd3d9082ded9/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java#L313-L314

I guess this also explains why `isAlias` is not used in `DefaultGenerator` and why it cannot be used here.

Personally I encountered a body parameter which is an alias to a map.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4934,7 +4938,7 @@ public CodegenParameter fromRequestBody(RequestBody body, Set<String> imports, S
                 codegenParameter.isModel = true;
             }
 
-            if (codegenModel != null && !codegenModel.emptyVars) {
+            if (codegenModel != null && (codegenModel.hasVars || useAliasReferenceName)) {
```
spacether: @bodograumann why route the aliased models to these two clauses?
* isObjectschema
* Codegenmodel != null
Per our discussion doesn't this only apply to array and map aliases?
Why not handle those cases in those clausee above only?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
spacether: @bodograumann whaf does a non primitive non-object non-array look like? Can you paste in a spec sample of one?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
bodograumann: The clauses for array and map aliases above are special cases where the model is not referenced by name, but included as an array or object literal.
With `generateAliasAsModel` we disable this special handling, so they should return to where they would normally be handled if they weren’t aliases. @spacether

spacether: @bodograumann I am having trouble understanding. Can you paste in a spec example here or in the PR description?

bodograumann: I have added an example to the description @spacether.

spacether: Thank you for adding that. It has helped me understand this use case.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
spacether: @bodograumann this update looks good but it is a bit difficult to understand the way you wrote it because for array and map isAlias use cases we are not going down the isMapSchema and isArraySchema routes.

Can you please update your PR to something like the following:
```
privat void addBodyModelSchema(blah...) {
  // this function adds a model schema to a request body
}

public CodegenParameter fromRequestBody(...) {
  if (ModelUtils.isMapSchema(schema) {
    if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
      addBodyModelSchema(blah..)
    } else {
      // non-alias model code here
    } 
  } else if (ModelUtils.isArraySchema(schema) {
    if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
      addBodyModelSchema(blah..)
    } else {
      // non-alias model code here
    } 
  } else if (ModelUtils.isObjectSchema(schema) {
    addBodyModelSchema(blah..)
  }
}
```
That way, it is very clear to the maintainers that
- we are using an isAlias feature
- we only use it in array schemas + map schemas
- and that we are using it to link up to an existing model

spacether:
@wing328 @jimschubert what do you think about connecting operation.request.body to an isAlias model (type array and type map only) in DefaultCodegen.java. Should we make this change here or only in a specific generator?
This issue where isAlias models are not connected up probably exists at:
- operation.request.body
- operation.response.datatype

bodograumann:
Are you sure about that structure @spacether?
I think it would have to look like this:
```java
private void addBodyModelSchema(
  CodegenParameter codegenParameter,
  Set<String> imports,
  String bodyParameterName,
  boolean forceSimpleRef
) {
    CodegenModel codegenModel = null;
    if (StringUtils.isNotBlank(name)) {
        schema.setName(name);
        codegenModel = fromModel(name, schema);
    }
    if (codegenModel != null) {
        codegenParameter.isModel = true;
    }

    if (codegenModel != null && (codegenModel.hasVars || forceSimpleRef)) {
        ...
    } else {
        ...
    }
}

public CodegenParameter fromRequestBody(...) {
  ...
  if (ModelUtils.isMapSchema(schema) {
    if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
      addBodyModelSchema(codegenParameter, imports, bodyParameterName, true);
    } else {
      // non-alias model code here
    }
  } else if (ModelUtils.isArraySchema(schema) {
    if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
      addBodyModelSchema(codegenParameter, imports, bodyParameterName, true);
    } else {
      // non-alias model code here
    }
  } else if (ModelUtils.isObjectSchema(schema) || ModelUtils.isComposedSchema(schema)) {
    addBodyModelSchema(codegenParameter, imports, bodyParameterName, false);
  } else ...
}
```

That doesn’t seem much more readable to me.

I agree that refactoring, especially of the whole `isObjectSchema || isComposedSchema` case is in order, but that is beyond my understanding of the code.

spacether:
> Are you sure about that structure @spacether?
> I think it would have to look like this:
> 
> ```java
> private void addBodyModelSchema(
>   CodegenParameter codegenParameter,
>   Set<String> imports,
>   String bodyParameterName,
>   boolean forceSimpleRef
> ) {
>     CodegenModel codegenModel = null;
>     if (StringUtils.isNotBlank(name)) {
>         schema.setName(name);
>         codegenModel = fromModel(name, schema);
>     }
>     if (codegenModel != null) {
>         codegenParameter.isModel = true;
>     }
> 
>     if (codegenModel != null && (codegenModel.hasVars || forceSimpleRef)) {
>         ...
>     } else {
>         ...
>     }
> }
> 
> public CodegenParameter fromRequestBody(...) {
>   ...
>   if (ModelUtils.isMapSchema(schema) {
>     if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
>       addBodyModelSchema(codegenParameter, imports, bodyParameterName, true);
>     } else {
>       // non-alias model code here
>     }
>   } else if (ModelUtils.isArraySchema(schema) {
>     if (ModelUtils.isGenerateAliasAsModel() && StringUtils.isNotBlank(name)) {
>       addBodyModelSchema(codegenParameter, imports, bodyParameterName, true);
>     } else {
>       // non-alias model code here
>     }
>   } else if (ModelUtils.isObjectSchema(schema) || ModelUtils.isComposedSchema(schema)) {
>     addBodyModelSchema(codegenParameter, imports, bodyParameterName, false);
>   } else ...
> }
> ```
> 
> That doesn’t seem much more readable to me.
> 
> I agree that refactoring, especially of the whole `isObjectSchema || isComposedSchema` case is in order, but that is beyond my understanding of the code.

@bodograumann nope I am not sure about the structure. What a wrote was just a guideline. Yup, I agree that it would look something like what you wrote.
I feel that what you wrote is more readable for the following reasons:
- it would remove ~130 lines from the fromRequestBody method, moving that code into the addBodyModelSchema method, a shorter fromRequestBody is easier to understand. This method is currently 241 lines long.
- using addBodyModelSchema in three locations allows maintainers to more easily see the 3 call sites and that that method is used for both alias models and objectschema models

When I read the code as-is, it some ways that I can be confused are:
- the useAliasReferenceName boolean is defined with 6 conditions and applies to two use cases
- many if statements omit being used by using !useAliasReferenceName so we have to keep it in mind
- when we finally do use useAliasReferenceName in an if case, that if case now covers 3 use cases (object models, array alias models, map alias models) which is again a lot to keep in mind and understand

What are the reasons that you feel your current code is more readable?

spacether:
Core Team, what do you think of this PR? 
This is a core update to DefaultCodegen.java
where Array and Map alias models would be directly hooked up to their generated classes.

@wing328 (2015/07) 
@jimschubert (2016/05) 
@cbornet (2016/05)
@ackintosh (2018/02) 
@jmini (2018/04) 
@etherealjoy (2019/06)

samples/client/petstore/dart2/petstore_client_lib/lib/api_client.dart
```diff
@@ -137,6 +137,8 @@ class ApiClient {
           return client.delete(url, headers: headerParams);
         case "PATCH":
           return client.patch(url, headers: headerParams, body: msgBody);
+        case "HEAD":
+          return client.head(url, headers: headerParams);
```
spacether: Why is this PR updating dart2 and go-experimental files?
Does it need a rebase on master?

bodograumann:
You make good points @spacether and I agree that the current code is not that easy to understand.
The reasons I have reservations are:
- The new method `addBodyModelSchema` is tightly coupled to the context established in `fromRequestBody` which is passed through the parameters. That means it is hard to understand on its own.
- It still contains two mostly independent cases; cf. `codegenModel.hasVars || forceSimpleRef` above. I would rather keep the else-case in the original position, because it is not used in the other two call sites. At the third call site it is only apparent which of the two cases has to be used after creating `codegenModel` though.

I’m not certain which variant is easier to read, but if you think creating the new method is an improvement, I can do the changes.

samples/client/petstore/dart2/petstore_client_lib/lib/api_client.dart
bodograumann: Yeah at the time some samples where outdated. I have have rebased now.
I could not run `ensure_up_to_date.sh` successfully though, so not sure what’s happening there on master...

spacether:
> You make good points @spacether and I agree that the current code is not that easy to understand.
> The reasons I have reservations are:
> 
> * The new method `addBodyModelSchema` is tightly coupled to the context established in `fromRequestBody` which is passed through the parameters. That means it is hard to understand on its own.
> * It still contains two mostly independent cases; cf. `codegenModel.hasVars || forceSimpleRef` above. I would rather keep the else-case in the original position, because it is not used in the other two call sites. At the third call site it is only apparent which of the two cases has to be used after creating `codegenModel` though.
> 
> I’m not certain which variant is easier to read, but if you think creating the new method is an improvement, I can do the changes.

Thanks for letting me know your concerns. The propose newer code is more tightly coupled, yes.
That's something that we can fix. What is we made the following changes?
- add the import in fromRequestBody, and remove it from the inputs for addBodyModelSchema
- have addBodyModelSchema return type codegenParameter and use it like `codegenParameter = addBodyModelSchema(bodyParameterName);

Also, are you sure that we need the forceSimpleRef boolean?
Why don't we just check one of these?
- `codegenModel.hasVars || codegenModel.isAlias`
- `codegenModel.hasVars || codegenModel. isArrayModel  || codegenModel. isMapModel`

With these changes our code in addBodyModelSchema will be more decoupled, yes?

What do you think?
I do think that creating the new method will help us here.

bodograumann:
Sorry @spacether, but I don’t see how that could work. Adding `codegenParameter` as return value does not make sense to me. It is created at the beginning of `fromRequestBody`, so must be passed into `addBodyModelSchema`. Then it is modified inside `addBodyModelSchema` and `fromRequestBody` can still use the same object.
I guess I could return an array of imports from `addBodyModelSchema` and add them to `imports` at the three call sites inside `fromRequestBody`, but that makes the code more complicated while only adding a bit of decoupling.

In principle `forceSimpleRef` could be replaced by
```java
ModelUtils.isMapSchema(schema) || ModelUtils.isArraySchema(schema)
```
I cannot say if this is equivalent to the conditions you listed, without further research in the code.
Removing the parameter could make sense I guess. Keeping the parameter makes the behaviour more explicit, but may lead to an unnecessary general scope of the `addBodyModelSchema` method, making it harder to refactor further.

bodograumann:
While working on a fix for #4968 I stumbled upon the `ModelUtils.unaliasScheme` which seems to work recursively and respect `isGenerateAliasAsModels`. Maybe `addBodyModelSchema` can be rewritten to use `unaliasScheme` once and do the rest independent from whether `isGenerateAliasAsModels` is set.
What do you think @spacether?

spacether:
So in Java it is valid to do.
String a = "hi";
a = "there";
If we wanted to we could do the same thing with codegenParameter. TO your point though we already need to pass it in, so refactoring `codegenParameter = addBodyModelSchema` doesn't really save us much.

If you want to decouple addBodyModelSchema more, I would suggest replacing the forceSimpleRef with `ModelUtils.isMapSchema(schema) || ModelUtils.isArraySchema(schema)`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -4977,87 +5054,94 @@ public CodegenParameter fromRequestBody(RequestBody body, Set<String> imports, S
...
-                    codegenParameter.baseName = "request_body";
+                if (StringUtils.isEmpty(bodyParameterName)) {
+                    if (StringUtils.isEmpty(mostInnerItem.complexType)) {
+                        codegenParameter.baseName = "request_body";
+                    } else {
+                        codegenParameter.baseName = mostInnerItem.complexType;
+                    }
                 } else {
-                    codegenParameter.baseName = mostInnerItem.complexType;
+                    codegenParameter.baseName = bodyParameterName;
```
spacether: Why did this line change?

spacether:
These updates looks great and are much easier to understand now. Thanks for making those tweak @bodograumann 

Ready for a review, Core Team:
@wing328 (2015/07) 
@jimschubert (2016/05) 
@cbornet (2016/05)
@ackintosh (2018/02) 
@jmini (2018/04) 
@etherealjoy (2019/06)

wing328:
@bodograumann sorry for taking so long to review. I wonder if you can resolve the merge conflicts when you've time and ping me via Slack or my email address in my Github profile when it's ready for another review. Thanks.

wing328:
LGTM. Thanks again for the PR and sorry for taking a bit long to review.

PR #4570, opened 2019-11-21
[BUG][Python-Flask] controller parameter may not be sanitized

Fjolnir-Dvorak:
Parameters in python are written snake_case. If there is a get-parameter for a request which is in random-case it will be snake-cased. That does not work with connexion since it searches for a parameter which is not sanitized.

This bug exists since  2019-01-11 when the python generators got reworked and cleaned up.

### PR checklist

@Jyhess

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonFlaskConnexionServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonFlaskConnexionServerCodegen.java
@@ -59,6 +59,6 @@ public class PythonFlaskConnexionServerCodegen extends PythonAbstractConnexionSe
     public String toParamName(String name) {
         // Connexion is broken and does not sanitize the parameter names.
         // Since Connexion does no sanitization we cannot do it here sadly.
-        return toVarName(name);
+        return name;
     }
 }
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonFlaskConnexionServerCodegen.java
```diff
@@ -54,4 +54,11 @@ protected void addSupportingFiles() {
         supportingFiles.add(new SupportingFile("__init__.mustache", packagePath(), "__init__.py"));
         testPackage = packageName + "." + testPackage;
     }
+
+    @Override
+    public String toParamName(String name) {
+        // Connexion is broken and does not sanitize the parameter names.
+        // Since Connexion does no sanitization we cannot do it here sadly.
+        return name;
```
wing328: What if the parameter name contains something like "-" (dash)? Would it still work without sanitization?

Fjolnir-Dvorak:
Well, I have to admit: it got fixed in __main__.py via the parameter "pythonic_params=True" in app.add_api
Due to abstraction to make the generated code modular and thus usable I never saw that fix. Originally it was a bug in connexion since that didn't allow pythonic parameters.

Issue #4586, opened 2019-11-22
[BUG][python-flask] converting to and from dict with camelCase properties 


Shogoki:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used? --> **4.2.1**
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?

##### Description

When i have a model defined in OpenAPI, which uses camelCase Properties the generator automatically converts the python props to pythonic snake_case Names. So far, so good.

When using the from_dict(dict) method of the Model class it is taking a dict with the snakeCase properties as Keys and converts them at instantiation. This is super cool and useful, as we can directly pass a dict, created from JSON payload, as is. 

But, with an existing model, when I call the to_dict() method on it, i get a dict, which has the snake_case prop names as Key!

This means, that have inconsistency in converting a model to dict, and creating one from the dict again. as this will fail for all props having a camelCase name (see example). I believe this behaviour is not intended. I´d expect the to_dict() method to produce the camelCase prop names (which are available in the attribute_map anyway.

**Example:**

```python
 # create a new instance of MyModel
    model = MyModel(demo_name="name", id=1)
    # Property has snake_case Name: demoName --> demo_name
    log.warn("Model is: {}".format(model))
    # OUTPUT: WARNING in demo_controller: Model is: {'demo_name': 'name', 'id': 1}

    # using to_dict method to get a dict from Model
    model_dict = model.to_dict()
    log.warn("Model.to_dict() is: {}".format(model_dict))
    # OUTPUT: WARNING in demo_controller: Model.to_dict() is: {'demo_name': 'name', 'id': 1}

    # now we want to convert that dict (probably fetched from DB) back to our Model
    new_model = MyModel.from_dict(model_dict)
    log.warn("New Created Model is: {}".format(new_model))
    # OUTPUT: WARNING in demo_controller: New Created Model is: {'demo_name': None, 'id': 1}
    # --> demo_name = None !!!!
```

##### openapi-generator version

4.2.1

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
  title: DEMO for issue
  version: "1.0"
servers:
- url: http://localhost:8080
tags:
- name: demo
paths:
  /demo:
    get:
      operationId: get_demo
      responses:
[...]

PR #4587, opened 2019-11-22
patched to_dict method to use prop_name from attribute_map

Shogoki:
Fix for #4586 as asuggested.

### PR checklist

- [X ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [X ] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.3.x`, `5.0.x`. Default: `master`.
- [X ] Copy the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

Technical Comitee.


spacether:
How about having a default argument in the to_dict method:
'to_dict(json_keys=true)'
that way we allow both types of keys but default to
what you prefer?@wing328 should we default it to false to not be a breaking change?

Shogoki:
Hey, thanks for your quick reply. 
I think this would make sense, to still allow the old way.
If doing this, wouldnt it be a good idea to add the same argument, with same default, to the from_dict method to? Cause at the and, the thing that i would expect them to be behave the similiar.
What do you think?

spacether:
@shogoki adding the same input argument to from_dict also sounds good to me for the reasons that you mentioned.

Shogoki:
just updated the PR with an optional param json_keys which defaults to `True` for both `to_dict` as well as `from_dict`

modules/openapi-generator/src/main/resources/python-flask/base_model_.mustache
```diff
@@ -22,34 +22,37 @@ class Model(object):
...
         """Returns the model properties as a dict
-
+        
+        :param json_keys: Defines if json_keys are used in dict
+        :type: bool
         :rtype: dict
         """
         result = {}
 
         for attr, _ in six.iteritems(self.openapi_types):
```
spacether: This could be changed to:
`for attr in self.openapi_types:`
because we are only using the key, not the value

spacether:
@Shogoki thanks for making those update and adding the json_keys=True parameter and default.
Just a couple more points and then we can move forward
- the circleci tests is failing, looks like the kotlin sample needs an update, can you run `./bin/utils/ensure-up-to-date` then commit the updates to your branch? That should fix it
- can you add a test verifying the behavior or to_dict and from_dict? with False and True values for json_keys?
   - it looks like the tests for python-flask live here: https://github.com/OpenAPITools/openapi-generator/tree/master/samples/server/petstore/python-flask/openapi_server/test

Shogoki:
i did not have time to work on this in the last months.
Now i ran the command you suggested. For the tests I am not 100% sure. 
I believe the path you suggested contains generated files!?

However, i took a look into the corresponding mustache template file, but I am not quite sure, if this is the right place to add some tests for this kind of functionality

spacether:
Why are so many files in this PR? Does it need a rebase on master?

guillaumedsde:
hello :smiley:  kindly bumping this PR, would be interested in a merge

spacether:
@shogoki can you address the points that we raised above?
- The current pr is 1,600+ files and is very difficult to review. Can you rebase it on master?
- Can you address the open unresolved conversations?

Shogoki:
Sorry for the long time, not working on this. 
But, i finally made it to fix this one.
I also implemented the change requested by @spacether 

The PR is 19 Files now ;-)

modules/openapi-generator/src/main/resources/python-flask/base_model_.mustache
```diff
@@ -22,34 +22,37 @@ class Model(object):
...
+        :param json_keys: Defines if json_keys are used in dict
+        :type: bool
         :rtype: dict
         """
         result = {}
 
-        for attr, _ in six.iteritems(self.openapi_types):
+        for attr in self.openapi_types:
             value = getattr(self, attr)
+            dict_attr = self.attribute_map[attr] if json_keys else attr
```
spacether: Why not reassign the value of attr here?
```
if if json_keys:
    attr = self.attribute_map[attr]
```
We don't use it anywhere else below.

modules/openapi-generator/src/main/resources/python-flask/base_model_.mustache
```diff
@@ -22,34 +22,37 @@ class Model(object):
     attribute_map = {}
 
     @classmethod
-    def from_dict(cls{{^supportPython2}}: typing.Type[T]{{/supportPython2}}, dikt){{^supportPython2}} -> T{{/supportPython2}}:
```
spacether: Why did you remove this code which works for python 2.7?
We still need this because we support python 2.7 and higher.
Can you add it back in?

modules/openapi-generator/src/main/resources/python-flask/model.mustache
```diff
@@ -62,15 +62,17 @@ class {{classname}}(Model):
 {{/vars}}
 
     @classmethod
-    def from_dict(cls, dikt){{^supportPython2}} -> '{{classname}}'{{/supportPython2}}:
```
spacether: Why did you remove this code which works for python 2.7?
We still need this because we support python 2.7 and higher.
Can you add it back in?

bukowa:
Hello what is missing?

spacether:
Please rebase this on master branch and address the open comments

bukowa:
@spacether that looks simple enough but it is not clear for me what files should be changed - from what you wrote above - folders with "samples" are automatically generated (exluding tests) - so the only files that needs changing are in `modules/openapi-generator/src/main/resources` ?

also `from_dict` is pointing directly at `util.deserialize_model` - can this be done with `to_dict` aswell ?

spacether:
A couple o places will need to change
- you will need to update the mustache template files in modules/openapi-generator/src/main/resources
- then you need to regenerate the samples per the `Run the shell script(s` PR template
- if you are adding a test it should go in https://github.com/OpenAPITools/openapi-generator/tree/master/samples/server/petstore/python-flask/openapi_server/test where it will not me overwritten when samples are generated

MartijnGevaert:
Kindly bumping this PR.
Not having this functionality means not being able to use camelCase in json bodies in openApi spec.
for example, `userId` in openApi becomes `user_id` in Python.
If I then return from a REST endpoint:
```
response = new UserIdResponseObject()
response.user_id = "some user id"
return response.to_dict()
```
It returns `{user_id: "some user id"}` with `user_id` instead of `userId`

PR #4588, opened 2019-11-22
Add nonDiscriminatorVars to CodegenModel and use it instead of vars…

### LINKED ISSUE #3796: [BUG][JAVA] Discriminator property gets serialized twice

3.3.4

I'm working with a JHipster API-First project, which uses openapi-generator-maven-plugin (v3.3.4).

##### Description

By following the OpenAPI 3.x documentation/examples, it seems that the property used as discriminator must also be specified in the properties list of the schema.

Example:
```yaml
TicketEvent:
  type: object
  description: A generic event
  discriminator:
    propertyName: type
  required:
    - id
    - sequenceNumber
    - timestamp
    - type
  properties:
    id:
      type: integer
      format: int64
      readOnly: true
    sequenceNumber:
      type: integer
      readOnly: true
    timestamp:
      type: string
      format: date-time
      readOnly: true
    type:
      type: string
      readOnly: true
TicketMovedEvent:
  description: A ticket move event
  allOf:
    - $ref: '#/components/schemas/Event'
    - type: object
      required:
        - source
        - target
      properties:
        source:
          $ref: '#/components/schemas/Queue'
        target:
          $ref: '#/components/schemas/Queue'
```

The generated Java class for the code above seems to differ from the [Jackson guidelines](https://www.baeldung.com/jackson-annotations) for polymorphic type handling with annotations, where the property used as discriminator must not be present in the class. 
The generated code also includes this property as a class attribute with getter/setter. 

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type", visible = true)
@JsonSubTypes({
  @JsonSubTypes.Type(value = TicketMovedEvent.class, name = "TicketMovedEvent")
})

public class TicketEvent   {
   ...

   @JsonProperty("type")
   private String type;
```

This causes the serialization output to include the property twice, as shown below.

```
{
    ...
    "type": "TicketMovedEvent",
    "type": null,
    ...
}
```

I've also tried to remove the property from the OpenAPI properties list, leaving intact the discriminator part; this way the generated code corresponds to the Jackson's guidelines and the serialization works just fine. On the other hand, I get the following error during the deserialization process.

```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "type" (class it.blutec.bludesk.web.api.model.TicketMovedEvent), not marked as ignorable ...
```

To fix this, i've configured the Jackson `ObjectMapper` object for ignoring this kind of situations.

```java
ObjectMapper om = new ObjectMapper()
				.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```

By the way, this looks like a workaround to me beacuse my OpenAPI file doesn't (strictly) follows the spec.

##### Suggest a fix

The generator should not include an additional attribute (with getter/setter) for a discriminator property.

keenanpepper:
…for JavaSpring. Fixes #3796

This may also be appropriate for other generators but I am only changing it for the JavaSpring template now. Ran bin/spring-all-petstore.sh and all the changes are expected.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
@@ -54,6 +54,7 @@ public class CodegenModel {
     public List<CodegenProperty> readOnlyVars = new ArrayList<CodegenProperty>(); // a list of read-only properties
     public List<CodegenProperty> readWriteVars = new ArrayList<CodegenProperty>(); // a list of properties for read, write
     public List<CodegenProperty> parentVars = new ArrayList<CodegenProperty>();
+    public List<CodegenProperty> nonDiscriminatorVars = new ArrayList<CodegenProperty>(); // properties other than the discriminator (without parent's properties)
     public Map<String, Object> allowableValues;
 
     // Sorted sets of required parameters.
@@ -61,7 +62,8 @@ public class CodegenModel {
     public Set<String> allMandatory = new TreeSet<String>(); // with parent's required properties
 
     public Set<String> imports = new TreeSet<String>();
-    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, isNullable, hasRequired, hasOptional, isArrayModel, hasChildren, isMapModel;
+    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, isNullable, hasRequired, hasOptional,
+            isArrayModel, hasChildren, isMapModel, hasNonDiscriminatorVars;
     public boolean hasOnlyReadOnly = true; // true if all properties are read-only
     public ExternalDocumentation externalDocumentation;
 
@@ -582,6 +584,18 @@ public class CodegenModel {
                 }
             }
         }
+    }
+
+    /**
+     * Copy vars to nonDiscriminatorVars, except the discriminator
+     */
+    public void populateNonDiscriminatorVars() {
+        for (CodegenProperty var : vars) {
+            if (!var.name.equals(getDiscriminatorName())) {
+                nonDiscriminatorVars.add(var);
+            }
+        }
+        hasNonDiscriminatorVars = !nonDiscriminatorVars.isEmpty();
     }
 
     /**
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1943,6 +1943,8 @@ public class DefaultCodegen implements CodegenConfig {
         // remove duplicated properties
         m.removeAllDuplicatedProperty();
 
+        m.populateNonDiscriminatorVars();
+
         // post process model properties
         if (m.vars != null) {
             for (CodegenProperty prop : m.vars) {
--- a/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache
@@ -8,7 +8,7 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
   private static final long serialVersionUID = 1L;
 
 {{/serializableModel}}
-  {{#vars}}
+  {{#nonDiscriminatorVars}}
     {{#isEnum}}
     {{^isContainer}}
 {{>enumClass}}
@@ -34,8 +34,8 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
   private {{>nullableDataType}} {{name}}{{#isNullable}} = JsonNullable.undefined(){{/isNullable}}{{^isNullable}}{{#defaultValue}} = {{{.}}}{{/defaultValue}}{{/isNullable}};
   {{/isContainer}}
 
-  {{/vars}}
-  {{#vars}}
+  {{/nonDiscriminatorVars}}
+  {{#nonDiscriminatorVars}}
   public {{classname}} {{name}}({{{datatypeWithEnum}}} {{name}}) {
     this.{{name}} = {{#isNullable}}JsonNullable.of({{name}}){{/isNullable}}{{^isNullable}}{{name}}{{/isNullable}};
     return this;
@@ -92,7 +92,7 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
     this.{{name}} = {{name}};
   }
 
-  {{/vars}}
+  {{/nonDiscriminatorVars}}
 
   @Override
   public boolean equals(java.lang.Object o) {
@@ -101,17 +101,17 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
     }
     if (o == null || getClass() != o.getClass()) {
       return false;
-    }{{#hasVars}}
+    }{{#hasNonDiscriminatorVars}}
     {{classname}} {{classVarName}} = ({{classname}}) o;
-    return {{#vars}}Objects.equals(this.{{name}}, {{classVarName}}.{{name}}){{#hasMore}} &&
-        {{/hasMore}}{{/vars}}{{#parent}} &&
-        super.equals(o){{/parent}};{{/hasVars}}{{^hasVars}}
-    return true;{{/hasVars}}
+    return {{#nonDiscriminatorVars}}Objects.equals(this.{{name}}, {{classVarName}}.{{name}}){{#hasMore}} &&
+        {{/hasMore}}{{/nonDiscriminatorVars}}{{#parent}} &&
+        super.equals(o){{/parent}};{{/hasNonDiscriminatorVars}}{{^hasNonDiscriminatorVars}}
+    return true;{{/hasNonDiscriminatorVars}}
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash({{#vars}}{{name}}{{#hasMore}}, {{/hasMore}}{{/vars}}{{#parent}}{{#hasVars}}, {{/hasVars}}super.hashCode(){{/parent}});
+    return Objects.hash({{#nonDiscriminatorVars}}{{name}}{{#hasMore}}, {{/hasMore}}{{/nonDiscriminatorVars}}{{#parent}}{{#hasNonDiscriminatorVars}}, {{/hasNonDiscriminatorVars}}super.hashCode(){{/parent}});
   }
 
   @Override
@@ -119,8 +119,8 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
     StringBuilder sb = new StringBuilder();
     sb.append("class {{classname}} {\n");
     {{#parent}}sb.append("    ").append(toIndentedString(super.toString())).append("\n");{{/parent}}
-    {{#vars}}sb.append("    {{name}}: ").append(toIndentedString({{name}})).append("\n");
-    {{/vars}}sb.append("}");
+    {{#nonDiscriminatorVars}}sb.append("    {{name}}: ").append(toIndentedString({{name}})).append("\n");
+    {{/nonDiscriminatorVars}}sb.append("}");
     return sb.toString();
   }
 
--- a/samples/server/petstore/spring-mvc-j8-async/src/main/java/org/openapitools/model/Animal.java
+++ b/samples/server/petstore/spring-mvc-j8-async/src/main/java/org/openapitools/model/Animal.java
@@ -22,32 +22,8 @@ import javax.validation.constraints.*;
 })
 
 public class Animal   {
-  @JsonProperty("className")
-  private String className;
-
   @JsonProperty("color")
   private String color = "red";
-
-  public Animal className(String className) {
-    this.className = className;
-    return this;
-  }
-
-  /**
-   * Get className
-   * @return className
-  */
-  @ApiModelProperty(required = true, value = "")
-  @NotNull
-
-
-  public String getClassName() {
-    return className;
-  }
-
-  public void setClassName(String className) {
-    this.className = className;
-  }
 
   public Animal color(String color) {
     this.color = color;
@@ -79,13 +55,12 @@ public class Animal   {
       return false;
     }
     Animal animal = (Animal) o;
-    return Objects.equals(this.className, animal.className) &&
-        Objects.equals(this.color, animal.color);
+    return Objects.equals(this.color, animal.color);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(className, color);
+    return Objects.hash(color);
   }
 
   @Override
@@ -93,7 +68,6 @@ public class Animal   {
     StringBuilder sb = new StringBuilder();
     sb.append("class Animal {\n");
     
-    sb.append("    className: ").append(toIndentedString(className)).append("\n");
     sb.append("    color: ").append(toIndentedString(color)).append("\n");
     sb.append("}");
     return sb.toString();
--- a/samples/server/petstore/spring-mvc-j8-localdatetime/src/main/java/org/openapitools/model/Animal.java
+++ b/samples/server/petstore/spring-mvc-j8-localdatetime/src/main/java/org/openapitools/model/Animal.java
@@ -22,32 +22,8 @@ import javax.validation.constraints.*;
 })
 
 public class Animal   {
-  @JsonProperty("className")
-  private String className;
-
   @JsonProperty("color")
   private String color = "red";
-
-  public Animal className(String className) {
-    this.className = className;
-    return this;
-  }
-
-  /**
-   * Get className
-   * @return className
-  */
-  @ApiModelProperty(required = true, value = "")
-  @NotNull
-
-
-  public String getClassName() {
-    return className;
-  }
-
-  public void setClassName(String className) {
-    this.className = className;
-  }
 
   public Animal color(String color) {
     this.color = color;
@@ -79,13 +55,12 @@ public class Animal   {
       return false;
     }
     Animal animal = (Animal) o;
-    return Objects.equals(this.className, animal.className) &&
-        Objects.equals(this.color, animal.color);
+    return Objects.equals(this.color, animal.color);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(className, color);
+    return Objects.hash(color);
   }
 
   @Override
@@ -93,7 +68,6 @@ public class Animal   {
     StringBuilder sb = new StringBuilder();
     sb.append("class Animal {\n");
     
-    sb.append("    className: ").append(toIndentedString(className)).append("\n");
     sb.append("    color: ").append(toIndentedString(color)).append("\n");
     sb.append("}");
     return sb.toString();
--- a/samples/server/petstore/spring-mvc/src/main/java/org/openapitools/model/Animal.java
+++ b/samples/server/petstore/spring-mvc/src/main/java/org/openapitools/model/Animal.java
@@ -22,32 +22,8 @@ import javax.validation.constraints.*;
 })
 
 public class Animal   {
-  @JsonProperty("className")
-  private String className;
-
   @JsonProperty("color")
   private String color = "red";
-
-  public Animal className(String className) {
-    this.className = className;
-    return this;
-  }
-
-  /**
-   * Get className
-   * @return className
-  */
-  @ApiModelProperty(required = true, value = "")
-  @NotNull
-
-
-  public String getClassName() {
-    return className;
-  }
-
-  public void setClassName(String className) {
-    this.className = className;
-  }
 
   public Animal color(String color) {
     this.color = color;
@@ -79,13 +55,12 @@ public class Animal   {
       return false;
     }
     Animal animal = (Animal) o;
-    return Objects.equals(this.className, animal.className) &&
-        Objects.equals(this.color, animal.color);
+    return Objects.equals(this.color, animal.color);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(className, color);
+    return Objects.hash(color);
   }
 
   @Override
@@ -93,7 +68,6 @@ public class Animal   {
     StringBuilder sb = new StringBuilder();
     sb.append("class Animal {\n");
     
-    sb.append("    className: ").append(toIndentedString(className)).append("\n");
     sb.append("    color: ").append(toIndentedString(color)).append("\n");
     sb.append("}");
     return sb.toString();
--- a/samples/server/petstore/springboot-beanvalidation/src/main/java/org/openapitools/model/Animal.java
+++ b/samples/server/petstore/springboot-beanvalidation/src/main/java/org/openapitools/model/Animal.java
@@ -22,32 +22,8 @@ import javax.validation.constraints.*;
 })
 
 public class Animal   {
-  @JsonProperty("className")
-  private String className;
-
   @JsonProperty("color")
   private String color = "red";
-
-  public Animal className(String className) {
-    this.className = className;
-    return this;
-  }
-
-  /**
-   * Get className
-   * @return className
-  */
-  @ApiModelProperty(required = true, value = "")
-  @NotNull
-
-
-  public String getClassName() {
-    return className;
-  }
-
-  public void setClassName(String className) {
-    this.className = className;
-  }
 
   public Animal color(String color) {
     this.color = color;
@@ -79,13 +55,12 @@ public class Animal   {
       return false;
     }
     Animal animal = (Animal) o;
-    return Objects.equals(this.className, animal.className) &&
-        Objects.equals(this.color, animal.color);
+    return Objects.equals(this.color, animal.color);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(className, color);
+    return Objects.hash(color);
   }
 
   @Override
@@ -93,7 +68,6 @@ public class Animal   {
     StringBuilder sb = new StringBuilder();
     sb.append("class Animal {\n");
     
-    sb.append("    className: ").append(toIndentedString(className)).append("\n");
     sb.append("    color: ").append(toIndentedString(color)).append("\n");
     sb.append("}");
     return sb.toString();
--- a/samples/server/petstore/springboot-delegate-j8/src/main/java/org/openapitools/model/Animal.java
+++ b/samples/server/petstore/springboot-delegate-j8/src/main/java/org/openapitools/model/Animal.java
@@ -22,32 +22,8 @@ import javax.validation.constraints.*;
 })
 
 public class Animal   {
-  @JsonProperty("className")
-  private String className;
-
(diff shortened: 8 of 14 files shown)


keenanpepper:
I was going along with this comment from the issue #3796: "If removing discriminator property addresses the issue, we can certainly do that."

davinchia:
Anyone else want to weigh in on this? The proposed fix seems sensible as the discriminator is mainly an artifact of strongly-typed/object-oriented languages (we are using Java). Since information is within a type, persisting the discriminator field seems extraneous.

davinchia:
@keenanpepper i took a second look  at your changes since my comment yesterday and i believe we also need to remove the `visible` jackson annotation here https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaSpring/typeInfoAnnotation.mustache#L3. Otherwise the generated code attempts to pass the discriminator to the deserialiser and fails as it is no longer a field on the model.

keenanpepper:
@davinchia done

davinchia:
Much much appreciated @keenanpepper, I definitely didn't mean to be 'assigning' you work :) 

Two questions from me:
1) Any reason why we would not also include the changes in the Java client? Seems strange misleading to fix this on the server-side and not on the client side. e.g. as a dev I am now able to implement this correctly, but my clients still send out misleading serialised json and can lead to confusion when debugging. If there aren't good reasons, I'd be happy to do so either in this same PR, or in a separate PR.
2) What are conventions around merging/releasing this? If a branch passes all tests and has an approval, is it on maintainers or patchers to make sure the PR gets merged in? Furthermore, when can we expect the patch to be released? @jeff9finger can you shed some light?

keenanpepper:
1. Yes, that's the main thing I'm uncertain of, since there are so many different templates for all the different generators, I don't know which should have this change and which shouldn't. It's possible this bug even affects languages other than Java. I figured it was better to fix it for just one generator, and it can always be fixed in other generators later.

keenanpepper:
I just found and fixed an issue with this PR regarding hasMore for the new nonDiscriminatorProperties. We need hasMore to be computed independently for them, which means we also need to clone the properties themselves.

I could add a test for this case (discriminator property happens to be the last, so hasMore is incorrectly true on the non-discrimaintor property that becomes the last after the discriminator is removed), but I thought it better to get the code fixed first.

keenanpepper:
Okay I'm having a different thought about this now after we're trying to use my fix internally. We still want to have a getter method for the discriminator property. So my new idea is to scratch this approach (nonDiscriminatorVars), and instead use the @JsonIgnore annotation to mark the discriminator as ignored for serialization.

Any comments on that?

keenanpepper:
Okay I updated this code review with the new approach (to preserve the context & discussion) and edited the subject since nonDiscriminatorVars is no longer a part of it. Let me know what you think.

keenanpepper:
@davinchia @jeff9finger what do you think of this (somewhat less invasive) approach?

jeff9finger:
The Java templates use the `x-discriminator-value` vendor extension to allow definition of discriminator values. This is a useful feature for me at least.

Marking a discriminator property as ignored in all cases seems to invasive.  There are times when the discriminator is needed and other times when it isn't. One example I can think of is when the discriminator has meaning other than just a discriminator.  Would it make sense to only ignore the property when the `x-discriminator-value` vendor extension is null? (If the property is ignored, would the getter for the discriminator still provide the discriminator value?)

keenanpepper:
> (If the property is ignored, would the getter for the discriminator still provide the discriminator value?)

Yes, the @JsonIgnore annotation only affects serialization, not deserialization. So if the descriminator is PetType, you can still call instance.getPetType() and get the value of the pet type that was passed in.

That was actually the main reason for me updating this PR to use @JsonIgnore instead of just removing the discriminator from the generated model.

jeff9finger:
> Yes, the @JsonIgnore annotation only affects serialization, not deserialization.

OK. That is good. 

IIRC, the `@JsonTypeInfo` annotation instructs jersey to add the discriminator to the payload on serialization.  Is that correct?

keenanpepper:
Yes, the @JsonTypeInfo stuff causes the correct subtype to be instantiated during deserialization, and also causes the discriminator property to be added based on the concrete type during serialization. The discriminator field of the object will be ignored in favor of the concrete type during serialization.

jeff9finger:
Great.
LGTM 👍

nickshoe:
Another possibility is to remove the discriminator property from the generated class altogether.

This would initially lead to the following exception during deserialization:
`com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: ...`

But this can be easily fixed by configuring the Jackson ObjectMapper to ignore non-existing class properties when it encounters them in JSON during deserialization.
```java
    ObjectMapper om = new ObjectMapper();
    om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
    ...
```

Please see the following code example.

Event.java
```java
import com.fasterxml.jackson.annotation.JsonIgnore;
import com.fasterxml.jackson.annotation.JsonSubTypes;
import com.fasterxml.jackson.annotation.JsonTypeInfo;

@JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        include = JsonTypeInfo.As.PROPERTY,
        property = "type",
        visible = true
)
@JsonSubTypes({
        @JsonSubTypes.Type(value = CreationEvent.class, name = "CreationEvent"),
        @JsonSubTypes.Type(value = DeletionEvent.class, name = "DeletionEvent")
})
public class Event {

    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

}
```

CreationEvent.java
```java
public class CreationEvent extends Event {
}
```

DeletionEvent.java
```java
public class DeletionEvent extends Event {
}
```

JacksonExperiment.java
```java
    public static void main(String[] args) {
        ObjectMapper om = new ObjectMapper();
        om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);

        try {
            Event deserializedCreationEvent = om.readValue("{ \"type\": \"CreationEvent\", \"name\": \"Foo\" }", Event.class);
            System.out.println(deserializedCreationEvent.getClass().getName());

            Event deserializedDeletionEvent = om.readValue("{ \"type\": \"DeletionEvent\", \"name\": \"Bar\" }", Event.class);
            System.out.println(deserializedDeletionEvent.getClass().getName());

            CreationEvent creationEvent = new CreationEvent();
            creationEvent.setName("Foo");
            String serializedCreationEvent = om.writeValueAsString(creationEvent);
            System.out.println(serializedCreationEvent);

            DeletionEvent deletionEvent = new DeletionEvent();
            deletionEvent.setName("Bar");
            String serializedDeletionEvent = om.writeValueAsString(deletionEvent);
            System.out.println(serializedDeletionEvent);
        } catch (JsonProcessingException e) {
            e.printStackTrace();
        }
    }
```

This code will produce the following output:
```
CreationEvent
DeletionEvent
{"type":"CreationEvent","name":"Foo"}
{"type":"DeletionEvent","name":"Bar"}
```

nickshoe:
If adding the `om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);` config might be an hassle, we could open an issue on Jackson lib and ask them not to raise an error when a JSON key is used as the discriminator property (i.e. it is specified in the `@JsonTypeInfo( ..., property = "type", ... )`) _and_ the class is missing the respective property.

holomekc:
This issue is not only related to JavaSpring, because it is an overall Jackson topic. As mentioned by @nickshoe the it must not be JsonTypeInfo.As.EXISTING_PROPERTY because due to Jackson documentation would result in client being responsible for setting appropriate type values and this would make JsonSubTypes annotation useless. So the changes I can see in the commit mentioned in this issue are not enough.

We are using jersey2 and what works for is is to overwrite typeInfoAnnotation.mustache and pojo.mustache:

typeInfoAnnotation.mustache:
```mustache
{{#jackson}}

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "{{{discriminator.propertyBaseName}}}", visible = true)
{{#discriminator.mappedModels}}
{{#-first}}
@JsonSubTypes({
{{/-first}}
  @JsonSubTypes.Type(value = {{modelName}}.class, name = "{{^vendorExtensions.x-discriminator-value}}{{mappingName}}{{/vendorExtensions.x-discriminator-value}}{{#vendorExtensions.x-discriminator-value}}{{{vendorExtensions.x-discriminator-value}}}{{/vendorExtensions.x-discriminator-value}}"),
{{#-last}}
})
{{/-last}}
{{/discriminator.mappedModels}}
{{#isClassnameSanitized}}
@JsonTypeName("{{name}}")
{{/isClassnameSanitized}}
{{/jackson}}
```

pojo.mustache:
```mustache
{{#jackson}}
import com.fasterxml.jackson.annotation.JsonIgnore;
{{/jackson}}
/**
 * {{#description}}{{.}}{{/description}}{{^description}}{{classname}}{{/description}}
 */{{#description}}
@ApiModel(description = "{{{description}}}"){{/description}}
{{#jackson}}
@JsonPropertyOrder({
{{#vars}}
  {{classname}}.JSON_PROPERTY_{{nameInSnakeCase}}{{^-last}},{{/-last}}
{{/vars}}
})
{{/jackson}}
{{>additionalModelTypeAnnotations}}{{>generatedAnnotation}}{{#discriminator}}{{>typeInfoAnnotation}}{{/discriminator}}{{>xmlAnnotation}}
public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#vendorExtensions.x-implements}}{{#-first}}implements {{{.}}}{{/-first}}{{^-first}}, {{{.}}}{{/-first}}{{#-last}} {{/-last}}{{/vendorExtensions.x-implements}}{
{{#serializableModel}}
  private static final long serialVersionUID = 1L;

{{/serializableModel}}
  {{#vars}}
    {{#isEnum}}
    {{^isContainer}}
    {{^vendorExtensions.x-enum-as-string}}
{{>modelInnerEnum}}
    {{/vendorExtensions.x-enum-as-string}}
    {{/isContainer}}
    {{#isContainer}}
    {{#mostInnerItems}}
{{>modelInnerEnum}}
    {{/mostInnerItems}}
    {{/isContainer}}
    {{/isEnum}}
  {{#gson}}
  public static final String SERIALIZED_NAME_{{nameInSnakeCase}} = "{{baseName}}";
  {{/gson}}
  {{#jackson}}
  public static final String JSON_PROPERTY_{{nameInSnakeCase}} = "{{baseName}}";
  {{/jackson}}
  {{#withXml}}
  {{#isXmlAttribute}}
  @XmlAttribute(name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
  {{/isXmlAttribute}}
  {{^isXmlAttribute}}
    {{^isContainer}}
  @XmlElement({{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
    {{/isContainer}}
    {{#isContainer}}
  // Is a container wrapped={{isXmlWrapped}}
      {{#items}}
  // items.name={{name}} items.baseName={{baseName}} items.xmlName={{xmlName}} items.xmlNamespace={{xmlNamespace}}
  // items.example={{example}} items.type={{dataType}}
  @XmlElement({{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
      {{/items}}
      {{#isXmlWrapped}}
  @XmlElementWrapper({{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}name = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}")
      {{/isXmlWrapped}}
    {{/isContainer}}
  {{/isXmlAttribute}}
  {{/withXml}}
  {{#gson}}
  @SerializedName(SERIALIZED_NAME_{{nameInSnakeCase}})
  {{/gson}}
  {{#vendorExtensions.x-is-jackson-optional-nullable}}
  {{#isContainer}}
  private JsonNullable<{{{datatypeWithEnum}}}> {{name}} = JsonNullable.<{{{datatypeWithEnum}}}>undefined();
  {{/isContainer}}
  {{^isContainer}}
  private JsonNullable<{{{datatypeWithEnum}}}> {{name}} = JsonNullable.<{{{datatypeWithEnum}}}>{{#defaultValue}}of({{{.}}}){{/defaultValue}}{{^defaultValue}}undefined(){{/defaultValue}};
  {{/isContainer}}
  {{/vendorExtensions.x-is-jackson-optional-nullable}}
  {{^vendorExtensions.x-is-jackson-optional-nullable}}
  {{#isContainer}}
  private {{{datatypeWithEnum}}} {{name}}{{#required}}{{#defaultValue}} = {{{.}}}{{/defaultValue}}{{/required}}{{^required}} = null{{/required}};
  {{/isContainer}}
  {{^isContainer}}
  private {{{datatypeWithEnum}}} {{name}}{{#defaultValue}} = {{{.}}}{{/defaultValue}};
  {{/isContainer}}
  {{/vendorExtensions.x-is-jackson-optional-nullable}}

  {{/vars}}
  {{#parcelableModel}}
  public {{classname}}() {
  {{#parent}}
    super();
  {{/parent}}
  {{#gson}}
  {{#discriminator}}
    this.{{{discriminatorName}}} = this.getClass().getSimpleName();
  {{/discriminator}}
  {{/gson}}
  }
  {{/parcelableModel}}
  {{^parcelableModel}}
  {{#gson}}
  {{#discriminator}}
  public {{classname}}() {
    this.{{{discriminatorName}}} = this.getClass().getSimpleName();
  }
  {{/discriminator}}
  {{/gson}}
  {{/parcelableModel}}
  {{#vars}}

  {{^isReadOnly}}
  {{#vendorExtensions.x-enum-as-string}}
  public static final Set<String> {{{nameInSnakeCase}}}_VALUES = new HashSet<>(Arrays.asList(
    {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}}, {{/-last}}{{/enumVars}}{{/allowableValues}}
  ));

  {{/vendorExtensions.x-enum-as-string}}
  public {{classname}} {{name}}({{{datatypeWithEnum}}} {{name}}) {
    {{#vendorExtensions.x-enum-as-string}}
    if (!{{{nameInSnakeCase}}}_VALUES.contains({{name}})) {
      throw new IllegalArgumentException({{name}} + " is invalid. Possible values for {{name}}: " + String.join(", ", {{{nameInSnakeCase}}}_VALUES));
    }

    {{/vendorExtensions.x-enum-as-string}}
    {{#vendorExtensions.x-is-jackson-optional-nullable}}
    this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{name}});
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    {{^vendorExtensions.x-is-jackson-optional-nullable}}
    this.{{name}} = {{name}};
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    return this;
  }
  {{#isArray}}

  public {{classname}} add{{nameInCamelCase}}Item({{{items.datatypeWithEnum}}} {{name}}Item) {
    {{#vendorExtensions.x-is-jackson-optional-nullable}}
    if (this.{{name}} == null || !this.{{name}}.isPresent()) {
      this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{{defaultValue}}});
    }
    try {
      this.{{name}}.get().add({{name}}Item);
    } catch (java.util.NoSuchElementException e) {
      // this can never happen, as we make sure above that the value is present
    }
    return this;
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    {{^vendorExtensions.x-is-jackson-optional-nullable}}
    {{^required}}
    if (this.{{name}} == null) {
      this.{{name}} = {{{defaultValue}}};
    }
    {{/required}}
    this.{{name}}.add({{name}}Item);
    return this;
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
  }
  {{/isArray}}
  {{#isMap}}

  public {{classname}} put{{nameInCamelCase}}Item(String key, {{{items.datatypeWithEnum}}} {{name}}Item) {
    {{#vendorExtensions.x-is-jackson-optional-nullable}}
    if (this.{{name}} == null || !this.{{name}}.isPresent()) {
      this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{{defaultValue}}});
    }
    try {
      this.{{name}}.get().put(key, {{name}}Item);
    } catch (java.util.NoSuchElementException e) {
      // this can never happen, as we make sure above that the value is present
    }
    return this;
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    {{^vendorExtensions.x-is-jackson-optional-nullable}}
    {{^required}}
    if (this.{{name}} == null) {
      this.{{name}} = {{{defaultValue}}};
    }
    {{/required}}
    this.{{name}}.put(key, {{name}}Item);
    return this;
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
  }
  {{/isMap}}

  {{/isReadOnly}}
   /**
  {{#description}}
   * {{description}}
  {{/description}}
  {{^description}}
   * Get {{name}}
  {{/description}}
  {{#minimum}}
   * minimum: {{minimum}}
  {{/minimum}}
  {{#maximum}}
   * maximum: {{maximum}}
  {{/maximum}}
   * @return {{name}}
  **/
{{#required}}
{{#isNullable}}
  @javax.annotation.Nullable
{{/isNullable}}
{{/required}}
{{^required}}
  @javax.annotation.Nullable
{{/required}}
{{#useBeanValidation}}{{>beanValidation}}{{/useBeanValidation}}  @ApiModelProperty({{#example}}example = "{{{example}}}", {{/example}}{{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
{{#vendorExtensions.x-extra-annotation}}
  {{{vendorExtensions.x-extra-annotation}}}
{{/vendorExtensions.x-extra-annotation}}
{{#vendorExtensions.x-is-jackson-optional-nullable}}
  {{!unannotated, Jackson would pick this up automatically and add it *in addition* to the _JsonNullable getter field}}
  @JsonIgnore
{{/vendorExtensions.x-is-jackson-optional-nullable}}
{{#jackson}}{{#isDiscriminator}}
  @JsonIgnore
{{/isDiscriminator}}{{/jackson}}
{{^vendorExtensions.x-is-jackson-optional-nullable}}{{#jackson}}{{> jackson_annotations}}{{/jackson}}{{/vendorExtensions.x-is-jackson-optional-nullable}}
  public {{{datatypeWithEnum}}} {{getter}}() {
    {{#vendorExtensions.x-is-jackson-optional-nullable}}
    {{#isReadOnly}}{{! A readonly attribute doesn't have setter => jackson will set null directly if explicitly returned by API, so make sure we have an empty JsonNullable}}
    if ({{name}} == null) {
      {{name}} = JsonNullable.<{{{datatypeWithEnum}}}>{{#defaultValue}}of({{{.}}}){{/defaultValue}}{{^defaultValue}}undefined(){{/defaultValue}};
    }
    {{/isReadOnly}}
    return {{name}}.orElse(null);
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    {{^vendorExtensions.x-is-jackson-optional-nullable}}
    return {{name}};
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
  }
  {{#vendorExtensions.x-is-jackson-optional-nullable}}
{{> jackson_annotations}}
  public JsonNullable<{{{datatypeWithEnum}}}> {{getter}}_JsonNullable() {
    return {{name}};
  }
  {{/vendorExtensions.x-is-jackson-optional-nullable}}{{#vendorExtensions.x-is-jackson-optional-nullable}}
  @JsonProperty(JSON_PROPERTY_{{nameInSnakeCase}})
  {{#isReadOnly}}private{{/isReadOnly}}{{^isReadOnly}}public{{/isReadOnly}} void {{setter}}_JsonNullable(JsonNullable<{{{datatypeWithEnum}}}> {{name}}) {
    {{! For getters/setters that have name differing from attribute name, we must include setter (albeit private) for jackson to be able to set the attribute}}
    this.{{name}} = {{name}};
  }
  {{/vendorExtensions.x-is-jackson-optional-nullable}}
  {{^isReadOnly}}
  public void {{setter}}({{{datatypeWithEnum}}} {{name}}) {
    {{#vendorExtensions.x-enum-as-string}}
    if (!{{{nameInSnakeCase}}}_VALUES.contains({{name}})) {
      throw new IllegalArgumentException({{name}} + " is invalid. Possible values for {{name}}: " + String.join(", ", {{{nameInSnakeCase}}}_VALUES));
    }
    {{/vendorExtensions.x-enum-as-string}}
    {{#vendorExtensions.x-is-jackson-optional-nullable}}
    this.{{name}} = JsonNullable.<{{{datatypeWithEnum}}}>of({{name}});
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
    {{^vendorExtensions.x-is-jackson-optional-nullable}}
    this.{{name}} = {{name}};
    {{/vendorExtensions.x-is-jackson-optional-nullable}}
  }
  {{/isReadOnly}}
  {{/vars}}
{{>libraries/jersey2/additional_properties}}
  /**
   * Return true if this {{name}} object is equal to o.
   */
  @Override
  public boolean equals(Object o) {
  {{#useReflectionEqualsHashCode}}
    return EqualsBuilder.reflectionEquals(this, o, false, null, true);
  {{/useReflectionEqualsHashCode}}
  {{^useReflectionEqualsHashCode}}
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }{{#hasVars}}
    {{classname}} {{classVarName}} = ({{classname}}) o;
    return {{#vars}}{{#isByteArray}}Arrays{{/isByteArray}}{{^isByteArray}}Objects{{/isByteArray}}.equals(this.{{name}}, {{classVarName}}.{{name}}){{^-last}} &&
        {{/-last}}{{/vars}}{{#additionalPropertiesType}}&&
        Objects.equals(this.additionalProperties, {{classVarName}}.additionalProperties){{/additionalPropertiesType}}{{#parent}} &&
        super.equals(o){{/parent}};{{/hasVars}}{{^hasVars}}
    return {{#parent}}super.equals(o){{/parent}}{{^parent}}true{{/parent}};{{/hasVars}}
  {{/useReflectionEqualsHashCode}}
  }
  @Override
  public int hashCode() {
  {{#useReflectionEqualsHashCode}}
    return HashCodeBuilder.reflectionHashCode(this);
  {{/useReflectionEqualsHashCode}}
  {{^useReflectionEqualsHashCode}}
    return Objects.hash({{#vars}}{{^isByteArray}}{{name}}{{/isByteArray}}{{#isByteArray}}Arrays.hashCode({{name}}){{/isByteArray}}{{^-last}}, {{/-last}}{{/vars}}{{#parent}}{{#hasVars}}, {{/hasVars}}super.hashCode(){{/parent}}{{#additionalPropertiesType}}{{#hasVars}}, {{/hasVars}}{{^hasVars}}{{#parent}}, {{/parent}}{{/hasVars}}additionalProperties{{/additionalPropertiesType}});
  {{/useReflectionEqualsHashCode}}
  }
  @Override
  public String toString() {
    StringBuilder sb = new StringBuilder();
    sb.append("class {{classname}} {\n");
    {{#parent}}
    sb.append("    ").append(toIndentedString(super.toString())).append("\n");
    {{/parent}}
    {{#vars}}
    sb.append("    {{name}}: ").append(toIndentedString({{name}})).append("\n");
    {{/vars}}
    {{#additionalPropertiesType}}
    sb.append("    additionalProperties: ").append(toIndentedString(additionalProperties)).append("\n");
    {{/additionalPropertiesType}}
    sb.append("}");
    return sb.toString();
  }
  /**
   * Convert the given object to string with each line indented by 4 spaces
   * (except the first line).
   */
  private String toIndentedString(Object o) {
    if (o == null) {
      return "null";
    }
    return o.toString().replace("\n", "\n    ");
  }
{{#parcelableModel}}
  public void writeToParcel(Parcel out, int flags) {
{{#model}}
{{#isArray}}
    out.writeList(this);
{{/isArray}}
{{^isArray}}
{{#parent}}
    super.writeToParcel(out, flags);
{{/parent}}
{{#vars}}
    out.writeValue({{name}});
{{/vars}}
{{/isArray}}
{{/model}}
  }
  {{classname}}(Parcel in) {
{{#isArray}}
    in.readTypedList(this, {{arrayModelType}}.CREATOR);
{{/isArray}}
{{^isArray}}
{{#parent}}
    super(in);
{{/parent}}
{{#vars}}
{{#isPrimitiveType}}
    {{name}} = ({{{datatypeWithEnum}}})in.readValue(null);
{{/isPrimitiveType}}
{{^isPrimitiveType}}
    {{name}} = ({{{datatypeWithEnum}}})in.readValue({{complexType}}.class.getClassLoader());
{{/isPrimitiveType}}
{{/vars}}
{{/isArray}}
  }
  public int describeContents() {
    return 0;
  }
  public static final Parcelable.Creator<{{classname}}> CREATOR = new Parcelable.Creator<{{classname}}>() {
    public {{classname}} createFromParcel(Parcel in) {
{{#model}}
{{#isArray}}
      {{classname}} result = new {{classname}}();
      result.addAll(in.readArrayList({{arrayModelType}}.class.getClassLoader()));
      return result;
{{/isArray}}
{{^isArray}}
      return new {{classname}}(in);
{{/isArray}}
{{/model}}
    }
    public {{classname}}[] newArray(int size) {
      return new {{classname}}[size];
    }
  };
{{/parcelableModel}}
{{#discriminator}}
static {
  // Initialize and register the discriminator mappings.
  Map<String, Class<?>> mappings = new HashMap<String, Class<?>>();
  {{#mappedModels}}
  mappings.put("{{mappingName}}", {{modelName}}.class);
  {{/mappedModels}}
  mappings.put("{{name}}", {{classname}}.class);
  JSON.registerDiscriminator({{classname}}.class, "{{propertyBaseName}}", mappings);
}
{{/discriminator}}
}
```

I mean for Jackson it seems quite pointless to integrate the discriminator but I guess this is the easiest fix. So basically we tell Jackson: Set type due to annotations (which is what we want) and then we tell Jackson ignore one property (on getter) which is the discriminator and we do not care about.

nickshoe:
> Okay I'm having a different thought about this now after we're trying to use my fix internally. We still want to have a getter method for the discriminator property. So my new idea is to scratch this approach (nonDiscriminatorVars), and instead use the @JsonIgnore annotation to mark the discriminator as ignored for serialization.
> 
> Any comments on that?

Hey @keenanpepper, could you give an example where you need a getter for the discriminator property?
You could devise the type of an object by its class.

nickshoe:
@keenanpepper please, see also this comment https://github.com/OpenAPITools/openapi-generator/issues/3796#issuecomment-1199994908.

It should be the approach proposed by @davinchia at the beginning of the thread.

We just have to figure out why we would need a discriminator property in the POJO. I cant' imagine a real need for that, or I can think of writing code that uses the object class to infer the discriminator property value.

Issue #362, opened 2018-06-20
Better handling of alias


wing328:
##### Description

We want to provide better handling of alias:

- [x] Alias to primtive type
- [ ] Alias to map
- [ ] Alias to Array
- [ ] Alias of Alias

for the following components/scopes:
- [ ] requestBody
- [ ] parameters in the operation
- [ ] model properites
- [ ] response

##### openapi-generator version

Latest master 

##### Suggest a fix/enhancement

As a starting point, we've filed https://github.com/OpenAPITools/openapi-generator/pull/360. Please expect more PRs to come.

PR #4597, opened 2019-11-25
[Core] use alias models in request.body when GenerateAliasAsModel=true

spacether:
When we generate alias models using GenerateAliasAsModel
the operation.body incorrectly describes an alias model using the primitive datatype, not the alias model's class name.

For this spec:
```
paths:
  /fake/outer/number:
    post:
      tags:
        - fake
      description: Test serialization of outer number types
      operationId: fakeOuterNumberSerialize
      parameters:
        - name: body
          in: body
          description: Input number as post body
          schema:
            $ref: '#/definitions/OuterNumber'
      responses:
        '200':
          description: Output number
          schema:
            $ref: '#/definitions/OuterNumber'
definitions:
  OuterNumber:
    type: number
```
We produce a client with GenerateAliasAsModel=true
and our master branch would describe the operation.body as type number
When it should be described as type OuterNumber because we are producing a model
OuterNumber.

### Test Verififcation
This PR fixes that issue, describing operation.body as type OuterNumber.
A test showing that an alias model's baseType, dataType, and isModel properties are set correctly for an alias model:
- https://github.com/bodograumann/openapi-generator/blob/body-parameter-as-alias-spacether/modules/openapi-generator/src/test/java/org/openapitools/codegen/java/JavaClientCodegenTest.java#L87

### Open Issues:
- https://github.com/OpenAPITools/openapi-generator/issues/362
   - This PR fixes request body parameters which point to models

Core Team:
@wing328 (2015/07)
@jimschubert (2016/05)
@cbornet (2016/05)
@ackintosh (2018/02)
@jmini (2018/04)
@etherealjoy (2019/06)


samples/client/petstore/dart2/petstore_client_lib/lib/api_client.dart
```diff
@@ -137,6 +137,8 @@ class ApiClient {
           return client.delete(url, headers: headerParams);
         case "PATCH":
           return client.patch(url, headers: headerParams, body: msgBody);
+        case "HEAD":
+          return client.head(url, headers: headerParams);
```
spacether: this was generated by ensure-up-to-date

samples/client/petstore/go-experimental/go-petstore/client.go
```diff
@@ -350,6 +350,9 @@ func (c *APIClient) prepareRequest(
 }
 
 func (c *APIClient) decode(v interface{}, b []byte, contentType string) (err error) {
+	if len(b) == 0 {
+		return nil
+	}
```
spacether: this was generated by ensure-up-to-date

samples/client/petstore/python-experimental/docs/AnimalFarm.md
```diff
@@ -0,0 +1,9 @@
+# AnimalFarm
+
+## Properties
+Name | Type | Description | Notes
+------------ | ------------- | ------------- | -------------
+
+[[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
+
+
```
spacether: This is an array model that is not used in any endpoints

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -375,41 +357,6 @@ public void updateCodegenPropertyEnum(CodegenProperty var) {
...
-            // set the example value
-            if (modelProp.isEnum) {
-                String value = modelProp._enum.get(0).toString();
-                result.example = modelName + "(" + toEnumValue(value, simpleDataType) + ")";
-            } else {
-                result.example = modelName + "(" + result.example + ")";
-            }
-        }
-        return result;
-    }
```
spacether: We can delete this because it was only responsible for setting the result.example
That is now handled by setParameterExampleValue

samples/openapi3/client/petstore/go/go-petstore/client.go
```diff
@@ -353,6 +353,9 @@ func (c *APIClient) prepareRequest(
 }
 
 func (c *APIClient) decode(v interface{}, b []byte, contentType string) (err error) {
+	if len(b) == 0 {
+		return nil
+	}
```
spacether: this is from ensure-up-to-date

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -95,7 +95,7 @@ public void processOpts() {
 
         // default this to true so the python ModelSimple models will be generated
         ModelUtils.setGenerateAliasAsModel(true);
-        LOGGER.info(CodegenConstants.GENERATE_ALIAS_AS_MODEL + " is hard coded to true in this generator. Alias models will only be generated if they contain validations or enums");
+        LOGGER.info(CodegenConstants.GENERATE_ALIAS_AS_MODEL + " is hard coded to true in this generator.");
```
spacether: Now that the Java generator connects all request.body parameters to the correct aliasModels, we need to connect up all of those alias models in:
- request.body
- response.body
Earlier, python-experimental was deleting alias models that did't have validations or enums.
Now that Java is using all alias models, even those without validations and enums, this generator should use all of those alias models.

wing328:
GenerateAliasAsModel is only for array, map at the moment.

With this PR, you want to extend this option to cover all primitive types (e.g. integer, string, etc) as well?

spacether:
Correct. The Python generator treats primitives (string/int etc) as alias models too. If this is not a general case then I can limit it to Python or Python-experimental only.

wing328:
Please do it for `python-experimental` only.

If I remember correctly, the reason you mentioned to me before is that the API server is not returning the correct data type specified in the OpenAPI spec/doc and that's why you wrap everything (request, response, etc) as an object in Python, right?

spacether:
Thanks, I will do it for Python-experimental only.
I do it there for consistencywwith the spec and because a primitive model could contain validations an enums and our models are unable to store validations so I make an object model with a property and store the validations/enums there.

wing328:
>  I make an object model with a property and store the validations/enums there.

Instead of defining the data with primitive type, what about updating the spec to define it as a model (with a property) instead?

spacether:
That is another possible solution.
What if we added the ability to store validations in a model? That would be a simpler solution.

wing328:
> What if we added the ability to store a validation in a model? That would be a simpler solution.

Very interesting idea but according to the spec, only the properties should have validation (pattern, max, min, etc).

A workaround can be using vendor extensions on the model to document those validations.

spacether:
But we can define a string model with a validation. Is that model not conforming to the spec?

wing328:
> But we can define a string model with a validation

That's debatable. It really comes down to how we define "model" (and alias) and that's why we have got the option `generateAliasAsModel` as some languages allow defining array, map as model/object.

```
definitions:
  OuterNumber:
    type: number
```

The above doesn't have "properties" so we've NOT been treating it as a model in pretty much all generators. It's treated as an alias (that can be reused) instead.

spacether:
Ah, I didn't realize that a primitive model with validations was a fuzzy area. Reading the spec I also see that enums are not defined at the model level but we do it in the java generator. Why not follow that same pattern for a validation on a primitive?

spacether:
Closing this diff because it should be limited to the python-experimental generator only.
If I want to make these updated, I will reopen them in a different PR.

Issue #3844, opened 2019-09-05
[BUG][PYTHON] Multiple authentication schemes are used, only want to use one


phillip-elliott:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I have a spec with the following securitySchemes:
```
"securitySchemes": {
      "APIKeyHeader": {
        "type": "apiKey",
        "name": "API-Key",
        "in": "header"
      },
      "basicAuth": {
        "type": "http",
        "scheme": "basic"
      }
    }
```

I am applying these schemes globally, indicating that either APIKeyHeader OR basicAuth can be used:
```
security": [
    {
      "APIKeyHeader": []
    },
    {
      "basicAuth": []
    }
  ],
```

When using a python client, I do the following:
```
config = Configuration()
config.host = base_url
config.api_key['API-Key'] = api_key

api_client = ApiClient(configuration=config)
project_api = ProjectsApi(api_client)
```

I am not providing a username or password. I would expect the python client to ignore the basic auth scheme and just use the apiKey scheme instead. Instead, both are used and all requests are denied (403).

##### openapi-generator version

4.1.1

##### OpenAPI declaration file content or url

https://gist.github.com/phillip-elliott/0438a8159f138e1f270a7814b81e65d9

##### Command line used for generation

openapi-generator generate -i openapispec.json -g python 

##### Suggest a fix

The python client's Configuration class has a method named get_basic_auth_token. This method could return None if username and password are both set to None. If it returns None, then I believe the basic auth header will not be added to the requests (see ApiClient.update_params_for_auth).

wing328:
> I would expect the python client to ignore the basic auth scheme and just use the apiKey scheme instead. Instead, both are used and all requests are denied (403).

In other clients, if username and password are not set, then the HTTP basic header won't be sent out.

I wonder if you've time to make a similar enhancement to the Python client.

martyzz1:
[...]

PR #4605, opened 2019-11-25
[Python] conditionally set auth attributes

sebastien-rosset:
Fix for #3844:
- Do not add auth parameter in HTTP request if auth parameter is set None
- Add attribute to control which auth schemes can be used.

Currently, the python client obtains the list of security schemes that have been specified in the OAS spec, and automatically adds the authentication attributes in the header, cookie or query. This happens even if the parameters are unset. For example, the "Authorization" header may be set with an empty Bearer token:
```
'Authorization': 'Bearer '
```
With this PR, the bearer access token is no longer set if its value is None. 

Tests performed:
1) Use OAS spec that specifies more than one security scheme for a given operation, e.g. OAuth2 and cookie
2) Generate the Python SDK using the master branch
3) Write a Python client that sets the cookie but not the OAuth2 access token
4) Invoke the client and verifies both auth attributes have been set (causing auth failure)
5) Generate the Python SDK using code from this PR
6) Invoke the client again, verify the auth parameters are set conditionally.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -23,10 +23,12 @@ class Configuration(object):
 
     :param host: Base url
     :param api_key: Dict to store API key(s).
-      The dict key should be the name of the api key, as specified in the
-      OAS specification.
-      The dict value should be the value of the api key (i.e. the secret).
-    :param api_key_prefix: Dict to store API prefix (e.g. Bearer)
+      Each entry in the dict specifies an API key. The dict key is the name
+      of the api key, as specified in the OAS specification.
+      The dict value is itself a dict that must have a 'value' entry.
+      The 'value' entry is required and specifies the api key secret.
+      The 'prefix' entry is optional and specifies an api key prefix when
+      generating the auth data.
     :param username: Username for HTTP basic authentication
     :param password: Password for HTTP basic authentication
     :param access_token: The OAuth2 access token
@@ -43,15 +45,14 @@ class Configuration(object):
 
     You can programmatically set the cookie:
       conf = {{{packageName}}}.Configuration(
-        api_key={'JSESSIONID': 'abc123'}
-        api_key_prefix={'JSESSIONID': 'JSESSIONID='}
+        api_key={'JSESSIONID': {prefix='JSESSIONID=', value='abc123'}}
       )
     The following cookie will be added to the HTTP request:
        Cookie: JSESSIONID=abc123
     """
 
     def __init__(self, host="{{{basePath}}}",
-                 api_key=None, api_key_prefix=None,
+                 api_key=None,
                  username=None, password=None,
                  access_token=None):
         """Constructor
@@ -67,11 +68,6 @@ class Configuration(object):
         if api_key:
             self.api_key = api_key
         """dict to store API key(s)
-        """
-        self.api_key_prefix = {}
-        if api_key_prefix:
-            self.api_key_prefix = api_key_prefix
-        """dict to store API prefix (e.g. Bearer)
         """
         self.refresh_api_key_hook = None
         """function hook to refresh API key if expired
@@ -248,11 +244,13 @@ class Configuration(object):
         """
         if self.refresh_api_key_hook is not None:
             self.refresh_api_key_hook(self)
-        key = self.api_key.get(identifier)
-        if key:
-            prefix = self.api_key_prefix.get(identifier)
-            if prefix:
-                return "%s%s" % (prefix, key)
+        key_entry = self.api_key.get(identifier)
+        if not isinstance(key_entry, dict):
+            raise Exception("The api_key attribute must be a dictionary")
+        if key_entry:
+            key = key_entry["value"]
+            if "prefix" in key_entry:
+                return "%s%s" % (key_entry["prefix"], key)
             else:
                 return key
 
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -22,11 +22,31 @@ class Configuration(object):
     Do not edit the class manually.
 
     :param host: Base url
-    :param api_key: Dict to store API key(s)
+    :param api_key: Dict to store API key(s).
+      The dict key should be the name of the api key, as specified in the OAS specification.
+      The dict value should be the value of the api key (i.e. the secret).
     :param api_key_prefix: Dict to store API prefix (e.g. Bearer)
     :param username: Username for HTTP basic authentication
     :param password: Password for HTTP basic authentication
     :param access_token: The OAuth2 access token
+
+    :Example:
+
+    Given the following security scheme in the OpenAPI specification:
+      components:
+        securitySchemes:
+          cookieAuth:         # name for the security scheme
+            type: apiKey
+            in: cookie
+            name: JSESSIONID  # cookie name
+
+    You can programmatically set the cookie:
+      conf = {{{packageName}}}.Configuration(
+        api_key={'JSESSIONID': 'abc123'}
+        api_key_prefix={'JSESSIONID': 'JSESSIONID='}
+      )
+    The following cookie will be added to the HTTP request:
+       Cookie: JSESSIONID=abc123
     """
 
     def __init__(self, host="{{{basePath}}}",
@@ -231,7 +251,7 @@ class Configuration(object):
         if key:
             prefix = self.api_key_prefix.get(identifier)
             if prefix:
-                return "%s %s" % (prefix, key)
+                return "%s%s" % (prefix, key)
             else:
                 return key
 
@@ -258,7 +278,7 @@ class Configuration(object):
         auth = {}
 {{#authMethods}}
 {{#isApiKey}}
-        if self.api_key and '{{keyParamName}}' in self.api_key:
+        if '{{keyParamName}}' in self.api_key:
             auth['{{name}}'] = {
                 'type': 'api_key',
                 'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -71,11 +71,6 @@ class Configuration(object):
         """
 {{/hasBearerMethods}}
 {{/hasOAuthMethods}}
-        self.security_schemes = None
-        """The security schemes to use when authenticating and authorizing the client.
-        While the OpenAPI specification may support multiple security schemes (e.g. OAuth2, basic),
-        the client may want to explicitly use a specific security scheme.
-        """
         self.logger = {}
         """Logging Settings
         """
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -30,7 +30,7 @@ class Configuration(object):
 
     def __init__(self, host="{{{basePath}}}",
                  api_key=None, api_key_prefix=None,
-                 username="", password=""):
+                 username=None, password=None):
         """Constructor
         """
         self.host = host
@@ -60,17 +60,22 @@ class Configuration(object):
         """Password for HTTP basic authentication
         """
 {{#hasOAuthMethods}}
-        self.access_token = ""
+        self.access_token = None
         """access token for OAuth/Bearer
         """
 {{/hasOAuthMethods}}
 {{^hasOAuthMethods}}
 {{#hasBearerMethods}}
-        self.access_token = ""
+        self.access_token = None
         """access token for OAuth/Bearer
         """
 {{/hasBearerMethods}}
 {{/hasOAuthMethods}}
+        self.security_schemes = None
+        """The security schemes to use when authenticating and authorizing the client.
+        While the OpenAPI specification may support multiple security schemes (e.g. OAuth2, basic),
+        the client may want to explicitly use a specific security scheme.
+        """
         self.logger = {}
         """Logging Settings
         """
@@ -256,51 +261,51 @@ class Configuration(object):
 
         :return: The Auth Settings information dict.
         """
-        return {
+        auth = {}
 {{#authMethods}}
 {{#isApiKey}}
-            '{{name}}':
-                {
-                    'type': 'api_key',
-                    'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
-                    'key': '{{keyParamName}}',
-                    'value': self.get_api_key_with_prefix('{{keyParamName}}')
-                },
+        if self.api_key:
+            auth['{{name}}'] = {
+                'type': 'api_key',
+                'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
+                'key': '{{keyParamName}}',
+                'value': self.get_api_key_with_prefix('{{keyParamName}}')
+            }
 {{/isApiKey}}
 {{#isBasic}}
   {{^isBasicBearer}}
-            '{{name}}':
-                {
-                    'type': 'basic',
-                    'in': 'header',
-                    'key': 'Authorization',
-                    'value': self.get_basic_auth_token()
-                },
+        if self.username and self.password:
+            auth['{{name}}'] = {
+                'type': 'basic',
+                'in': 'header',
+                'key': 'Authorization',
+                'value': self.get_basic_auth_token()
+            }
   {{/isBasicBearer}}
   {{#isBasicBearer}}
-            '{{name}}':
-                {
-                    'type': 'bearer',
-                    'in': 'header',
-                    {{#bearerFormat}}
-                    'format': '{{{.}}}',
-                    {{/bearerFormat}}
-                    'key': 'Authorization',
-                    'value': 'Bearer ' + self.access_token
-                },
+        if self.access_token:
+            auth['{{name}}'] = {
+                'type': 'bearer',
+                'in': 'header',
+                {{#bearerFormat}}
+                'format': '{{{.}}}',
+                {{/bearerFormat}}
+                'key': 'Authorization',
+                'value': 'Bearer ' + self.access_token
+            }
   {{/isBasicBearer}}
 {{/isBasic}}
 {{#isOAuth}}
-            '{{name}}':
-                {
-                    'type': 'oauth2',
-                    'in': 'header',
-                    'key': 'Authorization',
-                    'value': 'Bearer ' + self.access_token
-                },
+        if self.access_token:
+            auth['{{name}}'] = {
+                'type': 'oauth2',
+                'in': 'header',
+                'key': 'Authorization',
+                'value': 'Bearer ' + self.access_token
+            }
 {{/isOAuth}}
 {{/authMethods}}
-        }
+        return auth
 
     def to_debug_report(self):
         """Gets the essential information for debugging.
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -259,7 +259,7 @@ class Configuration(object):
         auth = {}
 {{#authMethods}}
 {{#isApiKey}}
-        if self.api_key:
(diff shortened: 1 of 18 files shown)


sebastien-rosset:
@taxpon @frol @mbohlool  @cbornet @kenjones-cisco @tomplus @Jyhess @slash-arun @spacether

samples/client/petstore/python-asyncio/petstore_api/configuration.py
```diff
@@ -35,7 +35,7 @@ class Configuration(object):
 
     def __init__(self, host="http://petstore.swagger.io:80/v2",
                  api_key=None, api_key_prefix=None,
-                 username="", password=""):
+                 username=None, password=None):
```
spacether: If you make these default to None, then invoking 
config. get_basic_auth_token() will raise
`TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'`

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -30,7 +30,7 @@ class Configuration(object):
 
     def __init__(self, host="{{{basePath}}}",
                  api_key=None, api_key_prefix=None,
-                 username="", password=""):
+                 username=None, password=None):
```
spacether: If you make these default to None, then invoking
config. get_basic_auth_token() will raise
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
Truth testing for empty string returns false, so it makes sense to leave these as empty strings, that way your `if self.username and self.password` will work, and we will not raise an exception when we run self.get_basic_auth_token

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -60,17 +60,22 @@ class Configuration(object):
         """Password for HTTP basic authentication
         """
 {{#hasOAuthMethods}}
-        self.access_token = ""
+        self.access_token = None
```
spacether: Down below this is added to a string when we return r=from self.auth_settings()
can you keep this as a string default so it is impossible to throw an exception?
How about passing `access_token = ''` as an init arg, and setting
`self.access_token = access_token`?
That way we have that line only once in the {{#hasBearerMethods}} tag.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -60,17 +60,22 @@ class Configuration(object):
...
 {{/hasOAuthMethods}}
 {{^hasOAuthMethods}}
 {{#hasBearerMethods}}
-        self.access_token = ""
+        self.access_token = None
         """access token for OAuth/Bearer
         """
 {{/hasBearerMethods}}
 {{/hasOAuthMethods}}
+        self.security_schemes = None
```
spacether: Why do we need to add this parameter?
Why not just iterating over the keys of self.auth_settings()?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -256,51 +261,51 @@ class Configuration(object):
...
 {{#authMethods}}
 {{#isApiKey}}
-            '{{name}}':
-                {
-                    'type': 'api_key',
-                    'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
-                    'key': '{{keyParamName}}',
-                    'value': self.get_api_key_with_prefix('{{keyParamName}}')
-                },
+        if self.api_key:
```
spacether: Why not add a condition to only set this if the key is in api_key?
Something like if `self.api_key and '{{keyParamName}}' in self.api_key`?

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
```diff
@@ -526,7 +526,11 @@ class ApiClient(object):
...
             return
-
+        if self.configuration.security_schemes:
+            auth_settings = [
+                value for value in auth_settings
+                if value in self.configuration.security_schemes
+            ]
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
```
spacether: Now that we will be returning only auth_settings that exist, we can remove the:
`if not auth_settings['value']` code here.
Per its history: https://github.com/OpenAPITools/openapi-generator/commit/271661aec79771e15ce65943a1ed1e8af1ca4d34
It was only added to skip over handling an auth_setting where values were empty.
The proposed addition will no longer return an empty auth_setting, so we can remove it.

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
```diff
@@ -526,7 +526,11 @@ class ApiClient(object):
         """
         if not auth_settings:
             return
-
+        if self.configuration.security_schemes:
+            auth_settings = [
+                value for value in auth_settings
+                if value in self.configuration.security_schemes
+            ]
```
spacether: Is this still needed if we only return valid fully filled out auth_setting items in from 
self.configuration.auth_settings()?

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: You are right, initially I thought this would be needed, I will remove it.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: My reasoning for having both None and empty string was a concern some existing SDK users may complain this would break compatibility. For example, if somebody is using the SDK to test a system, they would probably want to test with empty user, empty password, empty access token, so I thought it wouldn't be a good idea to omit the authorization attribute (e.g. query, cookie, header) just because a string is empty.

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
sebastien-rosset: I am removing.

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
sebastien-rosset: I have removed the test.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: Sure. BTW, is there a good explanation of the api key and api key prefix? It's a bit confusing and I can add documentation. For example, somebody who reads the OAS cookie example at https://swagger.io/docs/specification/authentication/cookie-authentication/ may not be clear how api_key and api_key_prefix are supposed to be set. Initially I thought it had to be
self.api_key["JSESSIONID"]="abcde12345", but that's not how it works.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: My reasoning for having both None and empty string was a concern about backward compatibility. For example, if somebody is using the generated SDK to test a system, they would probably want to be able to test their API service with an empty user string, empty password, or empty access token. Also, potentially some SDK users are using the SDK in a unit test scenario where the access token is an empty string and the UT is accepting empty string as a valid token. I don't know enough about the usage to determine if it's really important to minimize any potential backward compatibility breakage. Because of these use cases, I thought it wouldn't be a good idea to omit the authorization attribute (e.g. query, cookie, header) just because a string is empty.

sebastien-rosset: True. What do you think about my other comment where I elaborate my reasoning about the "None" value? I just realized I would need to change the test, and I could also change config. get_basic_auth_token()

sebastien-rosset: I have also added some unit tests.

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: I don't know exactly how the api_key vs api_key_prefix work.
The best way to learn about it would be to read the current code and the PRs the created/edited those features.
Feel free to add comments describing those variables in the code

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -247,60 +240,66 @@ class Configuration(object):
...
 {{#authMethods}}
 {{#isApiKey}}
-            '{{name}}':
-                {
-                    'type': 'api_key',
-                    'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
-                    'key': '{{keyParamName}}',
-                    'value': self.get_api_key_with_prefix('{{keyParamName}}')
-                },
+        if self.api_key and '{{keyParamName}}' in self.api_key:
```
spacether: api_key will be a dict per our __init__ code, this could be changed to:
`if '{{keyParamName}}' in self.api_key`

sebastien-rosset:
Not directly related to this PR, is there any reason the generated Python classes do not leverage class inheritance?

samples/client/petstore/python-experimental/tests/test_pet_api.py
```diff
@@ -71,9 +71,10 @@ def request(self, *args, **kwargs):
 class PetApiTests(unittest.TestCase):
 
     def setUp(self):
-        config = Configuration()
-        config.host = HOST
-        self.api_client = petstore_api.ApiClient(config)
+        self.config = Configuration()
```
spacether: Optionally pass in the host an access_token inputs into the Configuration
instance rather than setting them with .host .access_token

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: sure, will do.

spacether:
> Not directly related to this PR, is there any reason the generated Python classes do not leverage class inheritance?

Which classes, the Configuration class, the model class?
What is the context around your question?

sebastien-rosset:
> > Not directly related to this PR, is there any reason the generated Python classes do not leverage class inheritance?
> 
> Which classes, the Configuration class, the model class?
> What is the context around your question?

Sorry, in the model class. For example, in the generated Java, "Cat extends Animal" and "Dog extends Animal". This is done when the OAS spec has the "allOf" keyword and discriminants, and the mustache template uses the "parent" tag. I know doing class inheritance is not required and classes can be flattened, but still it's nice to have class inheritance because the generated classes are less verbose and easier to read.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: I found out how api keys work. For example, given the following security scheme named 'cookieAuth' in the OpenAPI specification:

```
      components:
        securitySchemes:
          cookieAuth:         # name for the security scheme
            type: apiKey
            in: cookie
            name: JSESSIONID  # cookie name
```

The client should configure the cookie like this:
```
       configuration = openapi.Configuration(
          api_key={'JSESSIONID': the_cookie_value},
          api_key_prefix={'JSESSIONID': 'JSESSIONID='})
```

But I think there is a bug at line https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/configuration.mustache#L241

There should be no space: 
```
return "%s %s" % (prefix, key)
```

Replace with:
```
return "%s%s" % (prefix, key)
```

Otherwise there will be an extra space in the cookie in the HTTP request.
Cookie: JSESSIONID=abcde\r\n

sebastien-rosset: I have added a code sample.

sebastien-rosset:
Never mind, I see there is already a PR: #4446

> > > Not directly related to this PR, is there any reason the generated Python classes do not leverage class inheritance?

spacether:
> > > Not directly related to this PR, is there any reason the generated Python classes do not leverage class inheritance?
> > 
> > 
> > Which classes, the Configuration class, the model class?
> > What is the context around your question?
> 
> Sorry, in the model class. For example, in the generated Java, "Cat extends Animal" and "Dog extends Animal". This is done when the OAS spec has the "allOf" keyword and discriminants, and the mustache template uses the "parent" tag. I know doing class inheritance is not required and classes can be flattened, but still it's nice to have class inheritance because the generated classes are less verbose and easier to read.
> 
> I'm asking because I have submitted a PR for golang to support embedded structs (which is one way in golang to model type inheritance), and I wonder if I should submit a PR for Python too.

composed inheritance models can have the same properties in multiple child schemas.
That's why we don't do it with inheritance. I have an open PR adding composed schemas to the python-experimental generator here: https://github.com/OpenAPITools/openapi-generator/pull/4446

sebastien-rosset:
@spacether , I think I have addressed all your comments.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -22,15 +22,38 @@ class Configuration(object):
...
+      components:
+        securitySchemes:
+          cookieAuth:         # name for the security scheme
+            type: apiKey
+            in: cookie
+            name: JSESSIONID  # cookie name
+
+    You can programmatically set the cookie:
+      conf = {{{packageName}}}.Configuration(
+        api_key={'JSESSIONID': 'abc123'}
```
spacether: Thanks for adding this to the docs
A nice refactor here in the future (not now) would be
api_key['JSESSIONID'] = {
   'value': 'abc123',
   'prefix': 'JSESSIONID='
}
And we could get the optional prefix with
api_key_prefix = api_key['JSESSIONID'].get('prefix')
Then we could delete the api_key_prefix dict

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -238,7 +252,7 @@ class Configuration(object):
         if key:
             prefix = self.api_key_prefix.get(identifier)
             if prefix:
-                return "%s %s" % (prefix, key)
+                return "%s%s" % (prefix, key)
```
wing328: What about keeping the space in between?

Users upgrading to a new version with this change will encounter issues as the space is now gone.

sebastien-rosset: True. I think to address this I need to introduce one more configuration parameter that specifies whether there should be a space or not. This is more complicated to configure for the end user but at least it retains the backward compatibility.

spacether: Removing the space and storing the space in the prefix would be a general solution. But adding a new spacer parameter lets the user control it and ia a backwards compatible fix. Then you could use
%s%s%s (prefix, spacer, key)
where we default spacer to " " if it was not set in api_key or api_key_prefix

wing328: What about targetting this fix for 4.3.0 (upcoming minor release - breaking changes with fallback)?

I personally prefer not to introduce another parameter as too many options in the configuration is not a good thing.

Is it correct to say that the space is causing an issue in your use case?

spacether: If we do that could we also refactor to remove the api_key_prefix dict mentioned above? All this info is about the api key value pair, it is unnecessarily complex to use two dicts to store it.

sebastien-rosset: > What about targetting this fix for 4.3.0 (upcoming minor release - breaking changes with fallback)?
> 
> I personally prefer not to introduce another parameter as too many options in the configuration is not a good thing.
> 
> Is it correct to say that the space is causing an issue in your use case?
> 
> If it's not causing the issue, then I suggest we leave it as it's to make it consistent with other clients (e.g. ruby, php, etc) and revisit when someone has such a use case.

Yes, the space is causing a problem in my case. Naively it makes sense to automatically add a space when the generated auth is "PREFIX KEY". But what about the case when the generated auth must be "FOO=BAR" (which is my case)? In that case you need to set the prefix to "FOO=", and there is no reason why the generated auth would become "FOO= BAR" with a space in between.

sebastien-rosset: > If we do that could we also refactor to remove the api_key_prefix dict mentioned above? All this info is about the api key value pair, it is unnecessarily complex to use two dicts to store it.

Sure, I can do that.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: Should I rename "api_key" to "api_keys"? Since this will break backward compatibility, it will be a hint to the user.

spacether: In my opinion I like that. That way they hit client side exceptions rather than server side auth errors

modules/openapi-generator/src/main/resources/python/configuration.mustache
wing328: > But what about the case when the generated auth must be "FOO=BAR"

A workaround is to set the API key directly as "FOO=BAR" without using the prefix.

sebastien-rosset: I have committed a fix that should address the above comments. Let me know if this works for you.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: I noticed some of the arguments are not pluralized. So I kept the argument name as "api_key", but I've added a type check on the client side. If the argument is not a dict, then raise an Exception on the client side before anything is sent to the server.

samples/client/petstore/python-experimental/tests/test_api_client.py
```diff
@@ -41,20 +43,32 @@ def test_configuration(self):
         client = petstore_api.ApiClient(config)
 
         # test prefix
-        self.assertEqual('PREFIX', client.configuration.api_key_prefix['api_key'])
+        self.assertEqual('PREFIX', client.configuration.api_key['api_key']['prefix'])
```
spacether: @sebastien-rosset optionally include "=" or " " at the end of 
config.api_key['api_key']["prefix"]

samples/client/petstore/python-experimental/tests/test_pet_api.py
```diff
@@ -115,20 +116,40 @@ def test_preload_content_flag(self):
         resp.close()
         resp.release_conn()
 
+    def test_config(self):
+        config = Configuration(host=HOST, access_token='ACCESS_TOKEN')
+        self.assertIsNotNone(config.get_host_settings())
+        self.assertEquals(config.get_basic_auth_token(),
+                          urllib3.util.make_headers(basic_auth=":").get('authorization'))
+        self.assertEquals(len(config.auth_settings()), 1)
```
spacether: This is great!

samples/client/petstore/python/tests/test_pet_api.py
```diff
@@ -222,7 +224,10 @@ def test_find_pets_by_status(self):
 
     def test_find_pets_by_tags(self):
         self.pet_api.add_pet(self.pet)
-
+        with open("file_out.txt", "w") as fout:
```
spacether: @sebastien-rosset why are we writing to disk here?
It looks like this may be for debugging on your machine only. If it is, can you delete it?

samples/client/petstore/python/tests/test_pet_api.py
```diff
@@ -13,6 +13,7 @@
 
 import os
 import unittest
+import pprint
```
spacether: If we only need this for writing to disk and we will remove the disk writing below, then please delete this too

sebastien-rosset: I am removing, this was only for testing purpose.

samples/client/petstore/python/tests/test_pet_api.py
sebastien-rosset: yes this was for debugging, I am removing.

samples/client/petstore/python-experimental/tests/test_api_client.py
sebastien-rosset: Do you mean something like this?
```
config.api_key['api_key'] = {'value': '123456', 'prefix': 'PREFIX '}
```
instead of:
```
config.api_key['api_key'] = {'value': '123456', 'prefix': 'PREFIX'}
```

spacether: Yes, something like that. Again this is just a nice to have not required. Your code properly exercises the feature and shows that it is working.

sebastien-rosset: I made the change.

wing328:
When you have time, please PM me via Slack to have a discussion on this in Jan 2020.

sebastien-rosset:
Based on feedback, I have opened a new PR to replace this one: #4988

Issue #4606, opened 2019-11-25
[BUG][Go] Support allOf inheritance with embedded go structs


sebastien-rosset:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Currently, golang structs are generated as a flat structure, even when the OAS spec uses "allOf" and discriminators. I understand generating flat structures is valid, but IMO it would be better to leverage the "supportsInheritance" attribute and generate embedded go structs, or at least have an option to generate embedded go structs. This would make the generated code less verbose, easier to read and use, reduce the size of the generated code.

##### openapi-generator version

4.2.1

##### OpenAPI declaration file content or url

https://gist.github.com/sebastien-rosset/f32f0abea11416bf82175577b48c251e

##### Command line used for generation

```
JAR_FILE=openapi-generator-cli.jar
config=config.yaml
spec=vehicles.yaml
java -jar ${JAR_FILE} generate \
   -c $config \
   -i ${spec} \
   -g go-experimental \
   -o out \
   --enable-post-process-file
```

##### Steps to reproduce

1. Checkout OpenAPITool from master and build it
2. Generate go client code from the OAS spec: https://gist.github.com/sebastien-rosset/f32f0abea11416bf82175577b48c251e
3. Notice the model files are flattened and do not use supportsInheritance

Also, the AbstractGoCodegen.java file does not set the "supportsInheritance" field to true, unlike the AbstractJavaCodegen:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java#L97

##### Related issues/PRs

PR to inherit the discriminators from all ancestors:
https://github.com/OpenAPITools/openapi-generator/pull/4454

Issue with proposal to use interfaces:
https://github.com/OpenAPITools/openapi-generator/issues/4559

##### Suggest a fix

Strawman proposal:
[...]

Issue #4559, opened 2019-11-20
[Go] Generate go interfaces to handle polymorphic substitution


sebastien-rosset:
##### Description

This is a suggestion to generate go interfaces in the go code.

The OAS spec supports polymorphism and inheritance. One area that seems to be problematic in the generated go code is when a schema refers to a type that has multiple sub-types, and the actual sub-type is determined at run-time by the api service. The go code is generated with a pointer to a struct (not an interface), which means there is no possibility of dynamically determining the proper go struct based on the discriminator value.

For example, suppose an OAS API specifies a path that returns a heterogenous list of animals. The returned list may include dogs, cats, and any "subclass" of animal. The api service may return a list of animals, and each object in the array may have a different discriminator value (e.g. Dog, Cat). This scenario can also apply to a component: for example, the Animal type may have a property which specifies a list of visited places, and each "Place" may have sub-types. A place could be a city, country or other types of places.

With the OpenAPITools generated code, the go client will unmarshal the list of Animals into the "Animal" go struct, and it will lose all the properties that were specified in the sub-types (such as Dog and Cat).

To address this problem, we have created a custom go implementation, and we generate a golang interface for each specified type in a OAS schema. The go type of the generated property is the (generated) go interface, not a pointer to a go struct. This makes it a bit harder to unmarshal because the default go unmarshaler does not know which type to instantiate. But this can be handled with a custom, generated JSON unmarhshaler; the unmarshaler reads the value of the discriminator property to determine the actual type, then it instantiates the proper go struct (e.g. Dog or Cat).

When there is a small and fixed number of sub-types, one could argue "oneOf" should be used to provide an explicit list of all sub-types. Besides the fact that currently the generators do not handle oneOf, it would ugly when the number of sub-types grows to a large value.

Is there a strong reason to generate pointers to go structs as opposed to go interfaces? Are there other approaches that work for polymorphic substitution? It seems the OAS spec is not very clear on that topic.

##### openapi-generator version

4.2.1

##### OpenAPI declaration file content or url

```yaml
paths:
 /Animals:
    get:
      responses:
        '200':
          description: a heterogeneous (dogs, cats...) list of animals.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Animals'

components:
  schemas:
    Resource:
      discriminator:
        propertyName: objectType
      properties:
        objectType:
          type: string
    Place:
      allOf:
        - $ref: '#/components/schemas/Resource'
        - type: object
[...]

PR #4607, opened 2019-11-25
[Golang] generate embedded structs and set supportsInheritance to true

sebastien-rosset:
This PR addresses #4606: generate embedded Go structs to model the class inheritance hierarchy.

I am a new contributor to OpenAPITools, so before I continue with a full PR or abandon the work, can I get feedback to determine if leveraging the "supportsInheritance" attribute is the right thing to do for golang? If yes, should there be a flag to control whether or not the generated structs should be flat or use embedded structs? I know the PR is not complete, I just want to know if this is headed in the right direction.

### PR checklist


sebastien-rosset:
@antihax  @bvwells  @grokify  @kemokemo  @bkabrda

sebastien-rosset:
If someone forgets to run the shell scripts under ./bin, are the CI jobs supposed to catch the problem and fail the tests? I am asking because this PR passed the tests, but subsequently I found out there are additional scripts under ./bin/openapi3, which presumably I have to run.
```
./bin/go-*
./bin/openapi3/go-*
```

sebastien-rosset:
The generated go code does not compile for a reason that seems unrelated to my PR.
```
$ cd samples/openapi3/client/petstore/go-experimental/go-petstore
$ go build
./model_nullable_class.go:27:29: syntax error: unexpected [, expecting semicolon or newline or }
./model_nullable_class.go:234:56: syntax error: unexpected [ after top level declaration
./model_nullable_class.go:236:19: syntax error: unexpected [ after top level declaration
./model_nullable_class.go:244:50: syntax error: mixed named and unnamed function parameters
./model_nullable_class.go:246:19: syntax error: unexpected [ at end of statement
./model_nullable_class.go:249:2: syntax error: non-declaration statement outside function body
./model_nullable_class.go:262:56: syntax error: unexpected [, expecting comma or )
./model_nullable_class.go:267:64: syntax error: unexpected [ after top level declaration
./model_nullable_class.go:269:19: syntax error: unexpected [ after top level declaration
./model_nullable_class.go:277:58: syntax error: mixed named and unnamed function parameters

antihax:
Catching up on this one. Go does not support inheritance so I am trying to understand what this is going to end up as?

sebastien-rosset:
> Catching up on this one. Go does not support inheritance so I am trying to understand what this is going to end up as?

I'm going to run the scripts in the bin directory and commit to show the proposed diff. The properties of the "parent" are defined in a separate go struct, and the go structs are embedded.

OpenAPI allOf/anyOf/oneOf is not necessarily implemented with the traditional "inheritance" construct of a programming language. In the case of golang, I am suggesting to use embedded structs. The benefits are:
1) Avoid having large, flattened structs
2) Avoid repeating the same set of properties across many structs, especially when there is a complex OpenAPI model with both breadth and depth of the "allOf" hierarchy.

A related problem is that currently, it's not possible to have collections of heterogeneous classes (#4559). This particular PR is not address the heterogeneous collection problem.

bkabrda:
@sebastien-rosset so what is the problem with having large flattened structures and repeating the same set of properties across many structs? To me, when the code is generated it doesn't matter too much when it's slightly larger.
I'm not against doing this change assuming you help me better understand the motivation behind it.

sebastien-rosset:
> @sebastien-rosset so what is the problem with having large flattened structures and repeating the same set of properties across many structs? To me, when the code is generated it doesn't matter too much when it's slightly larger.
> I'm not against doing this change assuming you help me better understand the motivation behind it.

Sure, here are some considerations:

* Ability to use polymorphic type substitution, e.g. https://play.golang.org/p/hlFnRJAomJ0
For example, this makes it possible to iterate over a collection of heterogeneous elements, and more generally to execute algorithms that can leverage type substitution. 
* SDK Code is easier to read and understand. Properties are not repeated, sometimes hundreds of times with large OpenAPI specs.
* Smaller footprint in generated code. In our case, we have small embedded devices, where 10KB can make a difference. I've measured a 10% to 20% decrease in code size depending on the input OpenAPI spec.
* It's easier to track SDK changes when the model changes. If the generated SDK is committed to git and the model changes in the parent type, the number of files that change is smaller. For example, if one property is added to a parent type and there are 100 sub-types, that would cause one change versus 101 changes.

Potentially there could be a golang configuration parameter to specify whether or not to flatten. In general, inheritance and composition are useful constructs, but if you want to support both flat and inheritance, it shouldn't be too hard to make this configurable (although too much configurability is sometimes not so good).

sebastien-rosset:
It looks like toInstantiationType(Schema property) must be overridden in src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java. The default implementation from the superclass does not work in the case that "additionalProperties" is set.

sebastien-rosset:
Mistakenly deleted branch, re-opening as https://github.com/OpenAPITools/openapi-generator/pull/4934

mariuspavelnistor:
Hi Sebastien. So let me understand , I run go-experimental-pestore.sh, but it seems this will not generate go-server models and it uses 'go-experimental' target. So it seems there is no way to generate go-server models now with embedded structures using this feature ?

sebastien-rosset:
> Hi Sebastien,
> 
> I couldn't make it work when having multiple allOf . It seems it embeds only the first structure.
> 
> schemas:
> RangeBase:
> allOf:
> - $ref: 'common.yaml#/components/schemas/EnableBase'
> - $ref: 'common.yaml#/components/schemas/CommonRangeBase'
> - $ref: 'common.yaml#/components/schemas/DutRangeBase'

Yes, that is true, currently the go template only works with the first `allOf` reference. The code generator has support for `parent` and allParents. The template would have to be extended to support allParents.

The relevant codegen code is:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L36,L40

And the relevant go template is:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/go/model_simple.mustache#L3-L12

sebastien-rosset:
Also, can you clarify if you are asking about the client side or server side? The PR I submitted was for the client side, not for the server side.

mariuspavelnistor:
I use it for server side.

sebastien-rosset:
> I use it for server side but the generated code is not different.

Then my recommendation is to take #4934 as a starting point and apply it to the go server generator. Is this something you could do?

mariuspavelnistor:
I will give a try.  One more thing I don't understand is why did you use discriminator to decide  embedding the parent into child ?  The thing is we have external yamls and it's difficult to change them. Can I change the code to avoid using this field ? Is it difficult to do it ?

I did a trick like this to avoid that field but I still need to change the yaml 

    CommonRangeBase:
      type: object
      discriminator:
        propertyName: CommonRangeBase      
      properties:
        id:
          type: string
          x-id: true
          x-auto: "increment"
        count:
          type: integer
          x-auto: default(1)
      required: [id,count]

sebastien-rosset:
> I will give a try. One more thing I don't understand is why did you use discriminator to decide embedding the parent into child ? The thing is we have external yamls and it's difficult to change them. Can I change the code to avoid using this field ? Is it difficult to do it ?
> 
> I did a trick like this to avoid that field but I still need to change the yaml
> 
> ```
> CommonRangeBase:
>   type: object
>   discriminator:
>     propertyName: CommonRangeBase      
>   properties:
>     id:
>       type: string
>       x-id: true
>       x-auto: "increment"
>     count:
>       type: integer
>       x-auto: default(1)
>   required: [id,count]
> ```

Support for embedded structs in go was one of my first PRs in openapi-generator, I wanted to minimize the changes. The OpenAPI specification has a section about inheritance, unfortunately it's not very precise and therefore it causes some controversy, which I didn't want to go into.
My recommendation is to add a go-specific CLI option that controls whether `allOf` $ref children are generated as a flat data structure or go embedded structs. This will mean people who want to keep the flat structure have control. The default would be "flat".

PR #4613, opened 2019-11-26
[Python] Do not use mutable default argument

jirikuncar:
Avoid using mutable default arguments in Python.

### PR checklist

--
cc @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @slash-arun (2019/11) @spacether (2019/11)

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -143,14 +143,7 @@ class Configuration(object):
 
     @property
     def logger_file(self):
-        """The logger file.
-
-        If the logger_file is None, then add stream handler and remove file
-        handler. Otherwise, add file handler and remove stream handler.
-
-        :param value: The logger_file path.
-        :type: str
-        """
+        """The logger file."""
         return self.__logger_file
 
     @logger_file.setter
@@ -174,16 +167,12 @@ class Configuration(object):
 
     @property
     def debug(self):
-        """Debug status
-
-        :param value: The debug status, True or False.
-        :type: bool
-        """
+        """Return debug status."""
         return self.__debug
 
     @debug.setter
     def debug(self, value):
-        """Debug status
+        """Set debug status.
 
         :param value: The debug status, True or False.
         :type: bool
@@ -205,18 +194,12 @@ class Configuration(object):
 
     @property
     def logger_format(self):
-        """The logger format.
-
-        The logger_formatter will be updated when sets logger_format.
-
-        :param value: The format string.
-        :type: str
-        """
+        """Return the logger format."""
         return self.__logger_format
 
     @logger_format.setter
     def logger_format(self, value):
-        """The logger format.
+        """Set the logger format.
 
         The logger_formatter will be updated when sets logger_format.
 
@@ -227,7 +210,7 @@ class Configuration(object):
         self.logger_formatter = logging.Formatter(self.__logger_format)
 
     def get_api_key_with_prefix(self, identifier):
-        """Gets API key (with prefix if set).
+        """Get API key (with prefix if set).
 
         :param identifier: The identifier of apiKey.
         :return: The token for api key authentication.
@@ -243,7 +226,7 @@ class Configuration(object):
                 return key
 
     def get_basic_auth_token(self):
-        """Gets HTTP basic authentication header (string).
+        """Get HTTP basic authentication header (string).
 
         :return: The token for basic HTTP authentication.
         """
@@ -252,7 +235,7 @@ class Configuration(object):
         ).get('authorization')
 
     def auth_settings(self):
-        """Gets Auth Settings dict for api client.
+        """Get Auth Settings dict for api client.
 
         :return: The Auth Settings information dict.
         """
@@ -303,7 +286,7 @@ class Configuration(object):
         }
 
     def to_debug_report(self):
-        """Gets the essential information for debugging.
+        """Get the essential information for debugging.
 
         :return: The report for debugging.
         """
@@ -315,7 +298,7 @@ class Configuration(object):
                format(env=sys.platform, pyversion=sys.version)
 
     def get_host_settings(self):
-        """Gets an array of host settings
+        """Get an array of host settings
 
         :return: An array of host settings
         """
@@ -349,10 +332,10 @@ class Configuration(object):
             {{/servers}}
         ]
 
-    def get_host_from_settings(self, index, variables=None):
+    def get_host_from_settings(self, index=0, variables=None):
         """Get host URL based on the index and variables.
 
-        :param index: array index of the host settings
+        :param index: array index of the host settings (default: 0)
         :param variables: mapping of variable and the corresponding value
         :return: URL based on host settings
         """
@@ -363,23 +346,23 @@ class Configuration(object):
             server = servers[index]
         except IndexError:
             raise ValueError(
-                "Invalid index {0} when selecting the host settings. Must be less than {1}"  # noqa: E501
-                .format(index, len(servers)))
+                "Invalid index {0} when selecting the host settings. "
+                "Must be less than {1}".format(index, len(servers)))
 
         url = server['url']
 
-        # go through variable and assign a value
-        for variable_name in server['variables']:
-            server_variable = server['variables'][variable_name]
-            value = variables.get(variable_name, server_variable['default_value'])
+        # go through variables and replace placeholders
+        for variable_name, variable in server['variables'].items():
+            used_value = variables.get(variable_name, variable['default_value'])
 
-            if 'enum_values' in server_variable and value not in server_variable['enum_values']:
+            if 'enum_values' in variable \
+                    and used_value not in variable['enum_values']:
                 raise ValueError(
-                    "The variable `{0}` in the host URL has invalid value {1}. Must be {2}."  # noqa: E501
-                    .format(
+                    "The variable `{0}` in the host URL has invalid value "
+                    "{1}. Must be {2}.".format(
                         variable_name, variables[variable_name],
-                        server['variables'][variable_name]['enum_values']))
+                        variable['enum_values']))
 
-            url = url.replace("{" + variable_name + "}", value)
+            url = url.replace("{" + variable_name + "}", used_value)
 
         return url
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -349,41 +349,37 @@ class Configuration(object):
             {{/servers}}
         ]
 
-    def get_host_from_settings(self, index, variables={}):
-        """Gets host URL based on the index and variables
+    def get_host_from_settings(self, index, variables=None):
+        """Get host URL based on the index and variables.
+
         :param index: array index of the host settings
-        :param variables: hash of variable and the corresponding value
+        :param variables: mapping of variable and the corresponding value
         :return: URL based on host settings
         """
-
+        variables = {} if variables is None else variables
         servers = self.get_host_settings()
 
-        # check array index out of bound
-        if index < 0 or index >= len(servers):
+        try:
+            server = servers[index]
+        except IndexError:
             raise ValueError(
-                "Invalid index {} when selecting the host settings. Must be less than {}"  # noqa: E501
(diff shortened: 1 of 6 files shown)


modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
         ]
 
-    def get_host_from_settings(self, index, variables={}):
-        """Gets host URL based on the index and variables
+    def get_host_from_settings(self, index, variables=None):
+        """Get host URL based on the index and variables.
+
         :param index: array index of the host settings
-        :param variables: hash of variable and the corresponding value
+        :param variables: mapping of variable and the corresponding value
```
spacether: Technically a dict in python but mapping is fine too. They are the same.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
         :return: URL based on host settings
         """
-
+        variables = {} if variables is None else variables
```
spacether: Much better

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
             raise ValueError(
-                "Invalid index {} when selecting the host settings. Must be less than {}"  # noqa: E501
+                "Invalid index {0} when selecting the host settings. Must be less than {1}"  # noqa: E501
                 .format(index, len(servers)))
 
-        server = servers[index]
         url = server['url']
 
         # go through variable and assign a value
         for variable_name in server['variables']:
```
spacether: Hey, you are iterating through a dict and using its key and value here.
Why not use `for var_name, var_value_dict in six.iteritems(server['variables']):`?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
-                        .format(
-                            variable_name, variables[variable_name],
-                            server['variables'][variable_name]['enum_values']))
-            else:
-                # use default value
-                url = url.replace(
-                    "{" + variable_name + "}",
-                    server['variables'][variable_name]['default_value'])
+            server_variable = server['variables'][variable_name]
+            value = variables.get(variable_name, server_variable['default_value'])
```
spacether: What if there is no default value? We would get a KeyError. Should we handle the key error?
The names used here are confusing, could you use something more clearly named like:
```
used_value = variables.get(server_var_name, server_var_dict['default_value'])
if `enum_values` in server_var_dict and used_value not in server_var_dict['enum_values']:
    raise ValueError(blah...)
url = url.replace("{" + server_var_name + "}", used_value)
```
The consolidation that you have done here is a lot better and simpler, thank you.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
         :return: URL based on host settings
         """
-
+        variables = {} if variables is None else variables
         servers = self.get_host_settings()
 
-        # check array index out of bound
-        if index < 0 or index >= len(servers):
+        try:
+            server = servers[index]
```
spacether: Are defaulting this to 0 elsewhere?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +349,37 @@ class Configuration(object):
...
-                            server['variables'][variable_name]['enum_values']))
-            else:
-                # use default value
-                url = url.replace(
-                    "{" + variable_name + "}",
-                    server['variables'][variable_name]['default_value'])
+            server_variable = server['variables'][variable_name]
+            value = variables.get(variable_name, server_variable['default_value'])
+
+            if 'enum_values' in server_variable and value not in server_variable['enum_values']:
```
spacether: The line lengths are throwing linting errors our continuous integration tests per:
```
petstore_api/configuration.py:321:80: E501 line too long (82 > 79 characters)
            value = variables.get(variable_name, server_variable['default_value'])
                                                                               ^
petstore_api/configuration.py:323:80: E501 line too long (96 > 79 characters)
            if 'enum_values' in server_variable and value not in server_variable['enum_values']:
```
Can you make sure your long lines get shortened to 79 characters or less?
Remember that if we are doing assignment we can use parens like:
```
blah = (
   "some very long value here.........."
)
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: I think so

jirikuncar:
@spacether thank you for the review. I have cleaned up the configuration template even more based on your suggestions.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +332,37 @@ class Configuration(object):
             {{/servers}}
         ]
 
-    def get_host_from_settings(self, index, variables={}):
-        """Gets host URL based on the index and variables
-        :param index: array index of the host settings
-        :param variables: hash of variable and the corresponding value
+    def get_host_from_settings(self, index=0, variables=None):
```
jirikuncar: Or we can keep it without a default value.
```suggestion
    def get_host_from_settings(self, index, variables=None):
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -349,41 +332,37 @@ class Configuration(object):
...
+        """Get host URL based on the index and variables.
+
+        :param index: array index of the host settings (default: 0)
```
jirikuncar:
```suggestion
        :param index: array index of the host settings
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -143,14 +143,7 @@ class Configuration(object):
...
     @property
     def logger_file(self):
-        """The logger file.
-
-        If the logger_file is None, then add stream handler and remove file
-        handler. Otherwise, add file handler and remove stream handler.
-
-        :param value: The logger_file path.
-        :type: str
-        """
```
spacether: Why are we removing this comment?
It provides useful info about the type of this property and how to use it

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -174,16 +167,12 @@ class Configuration(object):
 
     @property
     def debug(self):
-        """Debug status
-
-        :param value: The debug status, True or False.
-        :type: bool
-        """
```
spacether: Why are we removing this comment?
It provides useful info about the type of this property

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -205,18 +194,12 @@ class Configuration(object):
 
     @property
     def logger_format(self):
-        """The logger format.
-
-        The logger_formatter will be updated when sets logger_format.
-
-        :param value: The format string.
-        :type: str
-        """
```
spacether: Why are we removing this comment?
It provides useful info about the type of this property and how to use it

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: While I like this default, adding it will break legacy code because it was a positional argument before and we are changing it to a named argument.
Can we remove the default and keep it as positional?
Sorry for the new different guidance here. More thought on it led me to the above conclusion.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: But this is a getter not setter. The signature is wrong.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: This is a getter not setter so the signature is wrong.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: This is a getter not setter so the signature is wrong.

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Please add these docstrings back in.
When one uses help to find out how to use the methods of a class, it only gives back the docstring in the getter per:
https://stackoverflow.com/questions/16025462/what-is-the-right-way-to-put-a-docstring-on-python-property
Also having the getter describe the type returned by the getter his helpful to users

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Please add these docstrings back in.
When one uses help to find out how to use the methods of a class, it only gives back the docstring in the getter per:
https://stackoverflow.com/questions/16025462/what-is-the-right-way-to-put-a-docstring-on-python-property
Also having the getter describe the type returned by the getter his helpful to users

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Please add these docstrings back in.
When one uses help to find out how to use the methods of a class, it only gives back the docstring in the getter per:
https://stackoverflow.com/questions/16025462/what-is-the-right-way-to-put-a-docstring-on-python-property
Also having the getter describe the type returned by the getter his helpful to users

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: If the reason is that the setter docstrings are ignored then I've removed them and adjusted the getter ones.

spacether:
@jirikuncar it sounds like there may have been a misunderstanding here. My comments were asking that you preserve the docstrings in the setters as you were deleting the types shown there. My comments also give a reason why the docstring for both the getter and the setter exists there. In my opinion having the docstring in the getters is also helpful. As a dev I want to be able to:
- read a getter docstring to understand what I get
- read a setter docstring to understand what I need to set
- use the help(ClassName) in the python terminal to read getter docstring and read what we need to get/set
In my comments I didn't ask that you delete the getter docstrings.
If you want to delete the getter docstrings I am okay with that as the information still exists in the setter docstrings.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -174,7 +166,7 @@ class Configuration(object):
 
     @property
     def debug(self):
-        """Debug status
```
spacether: Please add a type description to this docstring
We are losing the information when we delete it from the setter

spacether:
@jirikuncar can you check out the Travis CI error here?
https://travis-ci.org/OpenAPITools/openapi-generator/builds/618563611?utm_source=github_status&utm_medium=notification
It looks like there are python errors in CI testing there.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: @spacether I don't get your comment.

```python
    @property
    def debug(self):
        """The debug status.

        :param value: The debug status, True or False.
        :type: bool
        """
```

jirikuncar:
I might revisit this one in the future by updating docstrings to follow PEP 257 (`pydocstyle`).

A minimal changed for mutable defaults has been moved to #4665.

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Below you are deleting the type description from @debug.setter
So because the getter does not describe the type, the dev user no longer has any documentation in the getter or setter describing what type this parameter should be.

spacether: Sorry, my mistake, I missed the remaining bool type here.

PR #4617, opened 2019-11-27
[typescript-angular] Support text response types

### LINKED ISSUE #2035: [BUG] [typescript-angular] Missing handling of text responses
##### Description

When the returned content type of an API is a text type (e.g. `text/plain`), the Angular generated service doesn't set the `responseType: text` option (as shown in Angular documentation: https://angular.io/guide/http#requesting-non-json-data)

3.3.4

```yaml
openapi: 3.0.0
info:
  title: MyAPI
  description: MY API
  version: 0.0.0
servers:
  - url: http://localhost:8080/api/
paths:
  /example:
    get:
      operationId: getExample
      responses:
        "200":
          description: "Success"
          content:
            text/plain:
              schema:
                type: string
```

`docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli:v3.3.4 generate -i /local/swagger.yaml  -g typescript-angular\ -o /local/src/app/api\ -DngVersion=6.0.0

##### Suggest a fix
This is the generated service:
```typescript
    public getExample(observe?: 'body', reportProgress?: boolean): Observable<string>;
    public getExample(observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<string>>;
    public getExample(observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<string>>;
    public getExample(observe: any = 'body', reportProgress: boolean = false ): Observable<any> {

        let headers = this.defaultHeaders;

        // to determine the Accept header
        let httpHeaderAccepts: string[] = [
            'text/plain'
        ];
        const httpHeaderAcceptSelected: string | undefined = this.configuration.selectHeaderAccept(httpHeaderAccepts);
        if (httpHeaderAcceptSelected !== undefined) {
            headers = headers.set('Accept', httpHeaderAcceptSelected);
        }

        // to determine the Content-Type header
        const consumes: string[] = [
        ];

        return this.httpClient.get<string>(`${this.configuration.basePath}/example`,
            {
                withCredentials: this.configuration.withCredentials,
                headers: headers,
                observe: observe,
                reportProgress: reportProgress
            }
        );
    }
```

Instead the generator should call the get method with this format:
```typescript
        return this.httpClient.get(`${this.configuration.basePath}/example`,
            {
                withCredentials: this.configuration.withCredentials,
                headers: headers,
                observe: observe,
                reportProgress: reportProgress,
                responseType: 'text'
            }
        );
``` 

Without the setting of this option. the http client will return an error instead of the text content of the called API

djnalluri:
Fixes #2035 
If a mime type is selected that starts with 'text' ('text/plain', 'text/csv', etc.), the appropriate response type of string is returned.

### PR checklist

Technical Committee: @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig @topce @akehir


modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -385,20 +385,32 @@ export class {{classname}} {
...
+            withCredentials: this.configuration.withCredentials,
+            headers: headers,
+            observe: observe,
+            reportProgress: reportProgress
+        }
+
         return this.httpClient.{{httpMethod}}{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}(`${this.configuration.basePath}{{{path}}}`,{{#isBodyAllowed}}
             {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},{{/isBodyAllowed}}
-            {
-{{#hasQueryParams}}
```
macjohnny: I suggest to leave the inline params here instead of using an object with type any

macjohnny:
Thanks for your PR

macjohnny:
please dont run the bin/openapi3 scripts and revert the samples

djnalluri:
Should I revert the samples to the state prior to my first commit or should the samples be in a state where the bin/openapi3 scripts have not been run but the bin/*-petstore.sh have been run?

macjohnny:
simply revert the samples folder to your previous commit and re-run the bin/typescript-angular-petstore-all.sh script

djnalluri:
9d87689 are the changes after running just the bin/typescript-angular-petstore-all.sh on ef7c790.

djnalluri:
73f1087 was my first run of the /bin/openapi3 scripts. I reverted the commit and reran /bin/typescript-angular-petstore-all.sh. Should be fixed now.

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -385,15 +385,25 @@ export class {{classname}} {
 
 {{/hasFormParams}}
 {{#useHttpClient}}
+    {{^isResponseFile}}
+        let responseType: any = undefined;
```
macjohnny: I am concerned that we lose type-checking for the httpClient method invocation, since it uses the exact value of `responseType` (`json`, `text`, `arraybuffer`, `blob`) to determine the correct method signature, see e.g.
https://github.com/angular/angular/blob/168abc6d6f52713383411b14980e104c99bfeef5/packages/common/http/src/client.ts#L963-L970

a better way to achieve this would be to implement this check in https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java and export a variable (similar to `isResponseFile`) that can be checked in the template to generate either
```
responseType: 'text'
```
or
```
responseType: 'blob'
```
or nothing (for the json case)

@wing328 maybe there is already a variable that tells whether we expect a text response?

djnalluri: The problem with this is the 'Accept' header is determined at run time. The configuration object supports a user-defined function for choosing an 'Accept' header making the response type unpredictable (at least for operations supporting multiple content types). This could be mitigated by explicitly providing the desired content type in the function parameters and generate typescript definitions for each content type.

macjohnny:
```suggestion
        let responseType: 'text' | 'json' = undefined;
```

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -385,15 +385,25 @@ export class {{classname}} {
...
+        if(httpHeaderAcceptSelected !== undefined) {
```
macjohnny:
```suggestion
        if(httpHeaderAcceptSelected && httpHeaderAcceptSelected.startsWith('text')) {
```

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -385,15 +385,25 @@ export class {{classname}} {
...
+            responseType = httpHeaderAcceptSelected.startsWith('text') ? 'text' : 'json';
```
macjohnny:
```suggestion
            responseType = 'text';
```

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -386,9 +386,9 @@ export class {{classname}} {
 {{/hasFormParams}}
 {{#useHttpClient}}
     {{^isResponseFile}}
-        let responseType: any = undefined;
-        if(httpHeaderAcceptSelected !== undefined) {
-            responseType = httpHeaderAcceptSelected.startsWith('text') ? 'text' : 'json';
+        let responseType: 'text' | 'json' = undefined;
```
macjohnny: = json

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -385,15 +385,25 @@ export class {{classname}} {
...
 {{/hasFormParams}}
 {{#useHttpClient}}
+    {{^isResponseFile}}
+        let responseType: 'text' | 'json' = 'json';
+        if(httpHeaderAcceptSelected && httpHeaderAcceptSelected.startsWith('text')) {
+            responseType = 'text';
+        }
+
+    {{/isResponseFile}}
         return this.httpClient.{{httpMethod}}{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}(`${this.configuration.basePath}{{{path}}}`,{{#isBodyAllowed}}
```
djnalluri: We'll need to strip the generics on 388/395 if we want to keep the type checking on responseType.

djnalluri: Actually, there is no overload without the generics that accepts 'json' response type and including generics forces 'json'. So I'm not really seeing a way for this to work out without discarding type checks somewhere.

macjohnny: Yeah, I guess there is no way to solve this, but your current code is fine, it enables using text responses. The returned observable is typed correctly anyway

macjohnny: Yeah, I guess there is no way to solve this, but your current code is fine, it enables using text responses. The returned observable is typed correctly anyway

Xambey:
@macjohnny , @jerome-jutteau 
djnalluri* 04.12.19 12:25 [typescript-angular] Support text response types (#4617) 
   {{^isResponseFile}}
                responseType: <any>responseType,
    {{/isResponseFile}} 

breaks all http requests, because of this problem: https://github.com/angular/angular/issues/18586
must be responseType: <'text' | 'json'> responseType

djnalluri:
Could you clarify as to what errors you are receiving and with which Angular version?

Xambey:
> Could you clarify as to what errors you are receiving and with which Angular version?

http requests are not sent, and the interceptor receives an exception object, instead of an error response. Latest version of v8 angular

djnalluri:
I have tested this change on v8 with text responses and seems to work fine. One quirk is that you must cast the return type from the strongly typed return object to a string instead. These changes should only affect endpoints that provide non-JSON MIME responses that begin with 'text'. The default selection function chooses a JSON response type when available so this change shouldn't apply under those conditions. I don't see how this code could break any previous functionality unless a MIME type starting with 'text' was returning JSON and existing code relied on the returned data to be parsed.

Xambey:
@djnalluri I was mistaken, but there is still a small problem. options is checked only on undefined, which is not very good (I inserted null, because of this there was a problem) 

```
let httpHeaderAcceptSelected: string | undefined = options && options.httpHeaderAccept;
        if (httpHeaderAcceptSelected === undefined) {
            // to determine the Accept header
            const httpHeaderAccepts: string[] = [
                'application/json'
            ];
            httpHeaderAcceptSelected = this.configuration.selectHeaderAccept(httpHeaderAccepts);
        }
```

djnalluri:
In most code I've seen, usage of null is largely avoided. If there is a use case for using null instead of undefined or if it should indeed be supported by convention, I can submit a PR to fix this. You can open an issue and tag me in it.

PR #4628, opened 2019-11-27
[Python] Add 'supportLargeModel' tag to handle unusually large OAS data type

### LINKED ISSUE #4600: [BUG][Python] "SyntaxError: more than 255 arguments" when a data type has 255 or more properties
#### Bug Report Checklist

##### Description

The Python model.mustache template generates a \_\_init\_\_ function. The function arguments are the properties defined in the data type. However, because the OpenAPI spec does not set a max limit for the number of properties in a data type, this means a data type could have a list of 300 properties. But until version 3.7, python functions cannot have more than 255 arguments. 
The generated python code will fail with a syntax error:
```
SyntaxError: more than 255 arguments
```
Note that as a comparison point, Java also has a limit of 255 function arguments, but the generated Java class does not exhibit a similar problem. The Java constructor is not generated with a long list of arguments.

4.2.1

I don't think a OAS content sample is needed. It's obvious from the mustache template that the generated __init__ function could have more than 255 arguments:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/model.mustache#L60

##### Steps to reproduce

1) Create a OpenAPI spec with a data type that has 255 or more properties (possibly from "inherited" classes using allOf
2) Run the Python code generator. In the generated model file, the \_\_init\_\_ method will have more than 255 arguments.
3) Try to execute the SDK using Python 3.6 or earlier. The python interpreter will report a syntax err. With python 3.7 and above, the python interpreter will execute without error.

##### Suggest a fix

One workaround is to use Python 3.7 or above. However, several Linux distributions do not support 3.7 yet, so it may be operationally complex to require SDK users to install Python 3.7.
Another workaround is to ensure that OAS data types should not have more than 254 arguments, which is a good idea anyway. The first argument of \_\_init\_\_ is "self", hence 254. This could be documented in the wiki. Having caveats in the wiki would be helpful.
A fix would be to remove the arguments from the \_\_init\_\_ function, but this will break backward compatibility.

sebastien-rosset:
This PR fixes #4600 by introducing a new **supportLargeModel** tag. Some languages have limitations such as max 255 arguments per function (Java, Python < 3.7) or 256 arguments (Eiffel).

This can be a problem for some of the generated code when the OAS data types are unusually large. For example, the Python model.mustache template generates a __init__ function. The function arguments are the properties defined in the data type. However, because the OpenAPI spec does not set a max limit for the number of properties in a data type, this means a data type could have a list of 300 properties. But until version 3.7, python functions cannot have more than 255 arguments.

This PR includes modifications for Python, and potentially in the future it could be done for other languages. This is because every language has its own set of limits.
If the code generator is invoked without the supportLargeModel tag, the generator is exactly the same as before. This ensures this does not break any backward compatibility.

### PR checklist

### DIFF
--- a/docs/generators/python-experimental.md
+++ b/docs/generators/python-experimental.md
@@ -12,4 +12,5 @@ sidebar_label: python-experimental
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |hideGenerationTimestamp|Hides the generation timestamp when files are generated.| |true|
 |generateSourceCodeOnly|Specifies that only a library source code is to be generated.| |false|
+|supportLargeModel|Generate files that can support unusually large data types, such as data types with more than 255 properties.| |false|
 |library|library template (sub-template) to use: asyncio, tornado, urllib3| |urllib3|
--- a/docs/generators/python.md
+++ b/docs/generators/python.md
@@ -12,4 +12,5 @@ sidebar_label: python
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
 |hideGenerationTimestamp|Hides the generation timestamp when files are generated.| |true|
 |generateSourceCodeOnly|Specifies that only a library source code is to be generated.| |false|
+|supportLargeModel|Generate files that can support unusually large data types, such as data types with more than 255 properties.| |false|
 |library|library template (sub-template) to use: asyncio, tornado, urllib3| |urllib3|
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java
@@ -245,6 +245,12 @@ public class CodegenConstants {
     public static final String SOURCECODEONLY_GENERATION = "generateSourceCodeOnly";
     public static final String SOURCECODEONLY_GENERATION_DESC = "Specifies that only a library source code is to be generated.";
 
+    // Some languages may have limitations such as max 255 arguments per function (e.g. Java, Python < 3.7).
+    // This may cause problems during compilation if the generated code exceeds language-specific limits, which could
+    // happen when at least one OpenAPI data type is unusually large.
+    public static final String SUPPORT_LARGE_MODEL = "supportLargeModel";
+    public static final String SUPPORT_LARGE_MODEL_DESC = "Generate files that can support unusually large data types, such as data types with more than 255 properties.";
+
     public static final String PARCELIZE_MODELS = "parcelizeModels";
     public static final String PARCELIZE_MODELS_DESC = "toggle \"@Parcelize\" for generated models";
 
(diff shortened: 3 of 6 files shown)


spacether:
@sebastianhaas what quantity of arguments are you talking about?
How many of them are required vs optional?
The python-experimental generator passes optional arguments using the kwargs input here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_init_properties.mustache#L1
So if using it is a possibility, it could reduce the likelihood that you would hit that argument count max
If they were all required, then the python-experimental generator would not solve your problem

sebastien-rosset:
> @sebastien-rosset what quantity of arguments are you talking about?
In one specific OAS spec, I have seen 266 properties. This is a long list of BIOS tokens, where each BIOS token is modeled as a property.

> How many of them are required vs optional?
Most of the arguments are optional.

> The python-experimental generator passes optional arguments using the kwargs input here:
> https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_init_properties.mustache#L1
> So if using it is a possibility, it could reduce the likelihood that you would hit that argument count max
> If all of your arguments are required, then the python-experimental generator will not solve your problem.

This will work for me.

> 
> Also, can you `Copy the Python technical committee to review the pull request` per the PR checklist?

Sure, will do.

sebastien-rosset:

spacether:
> > @sebastien-rosset what quantity of arguments are you talking about?
> 
> In one specific OAS spec, I have seen 266 properties. This is a long list of BIOS tokens, where each BIOS token is modeled as a property.
> 
> > How many of them are required vs optional?
> 
> Most of the arguments are optional.
> 
> > The python-experimental generator passes optional arguments using the kwargs input here:
> > https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_init_properties.mustache#L1
> > So if using it is a possibility, it could reduce the likelihood that you would hit that argument count max
> 
> Yes, that will work. In fact, for other reasons, I am finding out the "python" code generator won't work for me anyway, in particular because it does not handle allOf and discriminators properly. I am assuming most of these enhancements would go in the "python-experimental" area.
> 
> > If all of your arguments are required, then the python-experimental generator will not solve your problem.
> 
> This will work for me, most of the arguments are optional.
> 
> > Also, can you `Copy the Python technical committee to review the pull request` per the PR checklist?
> 
> Sure, will do.

@sebastien-rosset if you need composed schema support for oneOf/anyOf, python-experimental will support it soon per https://github.com/OpenAPITools/openapi-generator/pull/4446

sebastien-rosset:
> @sebastien-rosset if you need composed schema support for oneOf/anyOf, python-experimental will support it soon per #4446

Yes, I am watching this PR, this is great. There is also the related #4454.

Issue #4636, opened 2019-11-28
[BUG][PHP] Unable to generate working client based on API-Platform OAS3 definition


pascal-splotches:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

With an API generated with [API-Platform](https://api-platform.com/) I am unable to generate a working API client. Based on the API you build within API platform it automatically generates a JSON-LD API with an OpenAPI definition file.

The API endpoints do not seem to be able to parse, load and handle the JSON-LD fields in the response, as fields like `@id`, `@context` and `@type` remain empty on individual items. On collections fields like `hydra:member`, `hydra:totalItems` etc. also do not seem to be parsed and I end up with an empty collection.

I don't expect my use-case to be a rare one. Perhaps I'm doing something wrong, but so far have been unable to determine what? Any advice would be welcomed.

##### openapi-generator version

`4.2.1` - _not a regression_

##### OpenAPI declaration file content or url

https://gist.github.com/pascal-splotches/60efb27ff3946dfffd0ac745d4203a27

##### Command line used for generation

```bash
export PHP_POST_PROCESS_FILE="prettier --parser php --write"
openapi-generator generate \
    --input-spec "https://gist.githubusercontent.com/pascal-splotches/60efb27ff3946dfffd0ac745d4203a27/raw/7e0a85028d49a1076548f8b6dc5ee3dce2801493/openapi.definition.json" \
    --generator-name php \
    --api-package Api \
    --model-package Model \
    --invoker-package "Example\Api"
    --additional-properties "sortParamsByRequiredFlag=true,ensureUniqueParams=true,allowUnicodeIdentifiers=true,variableNamingConvention=camelCase" \
    --enable-post-process-file
```
##### Steps to reproduce

When generating the API, the following error gets thrown when not using the `--skip-validate-spec` flag:
```bash
[...]

PR #4684, opened 2019-12-03
[PHP] Preserve JSON accept header

ro0NL:
fixes https://github.com/OpenAPITools/openapi-generator/issues/4636#issuecomment-560574254

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
+++ b/modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
@@ -72,7 +72,7 @@ class HeaderSelector
     {
         if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
             return null;
-        } elseif ($jsonAccept = preg_grep("/application\/(?:\w++\+)?json/i", $accept)) {
+        } elseif ($jsonAccept = preg_grep('~(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$~', $accept)) {
             return implode(',', $jsonAccept);
         } else {
             return implode(',', $accept);
--- a/modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
+++ b/modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
@@ -72,8 +72,8 @@ class HeaderSelector
     {
         if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
             return null;
-        } elseif (preg_grep("/application\/json/i", $accept)) {
-            return 'application/json';
+        } elseif ($jsonAccept = preg_grep("/application\/(?:\w++\+)?json/i", $accept)) {
+            return implode(',', $jsonAccept);
         } else {
             return implode(',', $accept);
         }
--- a/samples/client/petstore/php/OpenAPIClient-php/lib/HeaderSelector.php
+++ b/samples/client/petstore/php/OpenAPIClient-php/lib/HeaderSelector.php
@@ -82,7 +82,7 @@ class HeaderSelector
     {
         if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
             return null;
-        } elseif ($jsonAccept = preg_grep("/application\/(?:\w++\+)?json/i", $accept)) {
+        } elseif ($jsonAccept = preg_grep('~(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$~', $accept)) {
             return implode(',', $jsonAccept);
         } else {
             return implode(',', $accept);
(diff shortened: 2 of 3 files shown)


modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
```diff
@@ -73,7 +73,7 @@ class HeaderSelector
         if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
             return null;
         } elseif (preg_grep("/application\/json/i", $accept)) {
-            return 'application/json';
+            return $accept;
```
jorissteyn: This is incorrect: the method must return a string and you're returning a list of content types here.

Your intent seems to be returning all content types, which is what the fallback else-clause does. The question is: why look for application/json in the first place?

There is a relevant unit test for this specific case in `HeaderSelectorTest` and this is what we're breaking when we remove the `preg_grep` clause:

```
        $headers = $selector->selectHeaders([
            'application/xml',
            'application/json'
        ], []);
        $this->assertSame('application/json', $headers['Accept']);
```

The test case seems to make sure json content types are always used in favour of non-json types. If we want to keep that behaviour I would suggest something like:

```
        if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
            return null;
        } 

        $jsonAcceptTypes = preg_grep('#application/(ld\+)json#i');

        if (!empty($jsonAcceptTypes)) {
            $accept = $jsonAcceptTypes;
        }
        
        return implode(',', $accept);
```

This way the behaviour is unchanged if no json content type is passed to the method.

ro0NL: got ya. I missed the single string return type. Funny this patch still works :) i guess in our case it's something like `[app/json+ld, app/json]` so all match JSON.

Maybe im missing the purpose of this method, as now looking at it.. it looks like it could be simply `implode(', ', $accept)` in all cases... but i'll do a more in-depth analysis tomorrow of what's exactly happening :+1: 

from your patch i figure we're filtering JSON-only accept types, while preserving the original input.

ro0NL:
Hi @jorissteyn , i've updated the PR.

In our case `$accept` is

```
array:3 [▼
  0 => "application/ld+json"
  1 => "application/json"
  2 => "text/html"
]
```

I had to modify the regex to match `ld+json` as well. It now returns `application/ld+json,application/json`, and all works well :+1:

ro0NL:
@wing328 same here :) switched to master.

wing328:
@ro0NL thanks for the PR. Can you reuse the following regular expression to detect JSON MIME instead?

```java
public boolean isJsonMime(String mime) {
    String jsonMime = "(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$";
    return mime != null && (mime.matches(jsonMime) || mime.equals("*/*"));
  }
```
These are Java code and we need to plot the regular expression to PHP.

Ref: https://github.com/OpenAPITools/openapi-generator/blob/d75d089bc96ec0ef8bf90344f952c6590c9596fb/samples/client/petstore/java/jersey1/src/main/java/org/openapitools/client/ApiClient.java#L514

ro0NL:
@wing328 done :) https://3v4l.org/TkPqE

modules/openapi-generator/src/main/resources/php/HeaderSelector.mustache
```diff
@@ -72,8 +72,8 @@ class HeaderSelector
     {
         if (count($accept) === 0 || (count($accept) === 1 && $accept[0] === '')) {
             return null;
-        } elseif (preg_grep("/application\/json/i", $accept)) {
-            return 'application/json';
+        } elseif ($jsonAccept = preg_grep('~(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$~', $accept)) {
```
wing328:
```
(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$
```

That's the one I provided earlier. In the code above, it's prefix and suffix with `~`.

Does it work for you? Have it tested it locally?

ro0NL: yes, see https://3v4l.org/TkPqE. In PHP that's the de delimiter (`~`). I used a diffferent delimiter then `/`, to avoid extra escaping.

wing328: Thanks for the explanation.

wing328:
cc @jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) ❤️, @ybelenko (2018/07), @renepardon (2018/12)

dkarlovi:
Welcome @ro0NL, nice to have you here. :100:

PR #4695, opened 2019-12-04
[core] Provide User feedback on confusing use of oneOf

jimschubert:
OAS 3.x specification isn't entirely clear on oneOf support. JSON Schema
defines oneOf in such a way that the Schema is only valid if it
validates against exactly one of the referenced Schemas. This suggests
that a Schema defined with oneOf can't include additional "dynamic"
properties. OpenAPI extends on this by adding the necessary
discriminator object, which allows tooling to decide the intended
Schema.

As tooling, openapi-generator may support loose or confusing definitions
in the Specification to better support our user's use cases. In this
case, we may warn that while this usage is technically valid the two
target specifications are unclear about the actual constraints regarding
oneOf.

This adds the warning and recommendation I proposed in review of #4482

cc @spacether @OpenAPITools/generator-core-team 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1789,21 +1789,18 @@ public class DefaultCodegen implements CodegenConfig {
             List<String> allRequired = new ArrayList<String>();
 
             // if schema has properties outside of allOf/oneOf/anyOf also add them to m
-            if (schema.getProperties() != null) {
-                addVars(m, unaliasPropertySchema(schema.getProperties()), schema.getRequired(), null, null);
+            if (composed.getProperties() != null && !composed.getProperties().isEmpty()) {
+                if (composed.getOneOf() != null && !composed.getOneOf().isEmpty()) {
+                    LOGGER.warn("'oneOf' is intended to include only the additional optional OAS extension discriminator object. " +
+                            "For more details, see https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.2.1.3 and the OAS section on 'Composition and Inheritance'.");
+                }
+                addVars(m, unaliasPropertySchema(composed.getProperties()), composed.getRequired(), null, null);
             }
-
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1789,21 +1789,18 @@ public CodegenModel fromModel(String name, Schema schema) {
...
-            // if schema has additionalproperties outside of allOf/oneOf/anyOf also add it to m
-            // if (schema.getAdditionalProperties() != null) {
-            //     addAdditionPropertiesToCodeGenModel(m, schema);
-            // }
-
             // parent model
             final String parentName = ModelUtils.getParentName(composed, allDefinitions);
             final List<String> allParents = ModelUtils.getAllParentsName(composed, allDefinitions);
             final Schema parent = StringUtils.isBlank(parentName) || allDefinitions == null ? null : allDefinitions.get(parentName);
-            final boolean hasParent = StringUtils.isNotBlank(parentName);
```
spacether: Are we deleting this because it is unused?

spacether:
@jimschubert this looks good. Should we add tests?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
jimschubert: Yeah. This would fail compilation of it was still expected.

jimschubert:
@spacether I'm not sure how we'd add tests. The changes in non-cli code is logging and a special case to test a minimal scenario which resulted in an NPE. That NPE will need to be investigated later.

For the CLI, we don't have tests that verify generated files. The change in this PR only results in stdout text, for which we don't have tests on other commands.

spacether:
> @spacether I'm not sure how we'd add tests. The changes in non-cli code is logging and a special case to test a minimal scenario which resulted in an NPE. That NPE will need to be investigated later.
> 
> For the CLI, we don't have tests that verify generated files. The change in this PR only results in stdout text, for which we don't have tests on other commands.

God to know, sounds like adding them would be more hassle than the benefit that we would get.

Issue #4705, opened 2019-12-04
[BUG][Kotlin] models properties order change when they become optional's


4brunu:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The order of the parameters in a model in kotlin changes if a property changes from required to optional.
This caused issues in my project.

Imagine that you have the class Person and every parameter is required.
```kotlin
// This is the generated class
data class Person (
    val firstName: kotlin.String,
    val lastName: kotlin.String
)

// And this is how you instantiate the class Person
val person = Person("John", "Smith")
```

This works great, but know the open api spec created by the backend person changes and `firstName` is now optional.
Now the parameter order changed, and the values are assigned to the wrong property, unless you change all the places in your project where you instantiate this class.

```kotlin
// This is the new generated class
data class Person (
    val lastName: kotlin.String,
    val firstName: kotlin.String? = null
)

// But here you are passing "John" as lastName and "Smith" as firstName
val person = Person("John", "Smith")
```

##### openapi-generator version

4.2.2

##### OpenAPI declaration file content or url

// First example, all properties required
https://gist.github.com/4brunu/e8c82fdcd79dbd3546632bae743138cf

// Second example, one property required
https://gist.github.com/4brunu/85487ad068597658f3b831c5f588faeb

##### Command line used for generation

`openapi-generator generate --input-spec openapi/openapi.json --generator-name kotlin --output openapi/`

##### Steps to reproduce

Run the generated with the two spec files and check the diferences.

##### Related issues/PRs

None.

##### Suggest a fix

Keep the order of the parameters despite them being required or not.
Or if it's better for the current behaviour to be keept for backwards compatibility, add an option to keep the order of the model properties.
Something like this 
https://github.com/OpenAPITools/openapi-generator/blob/d0dfe3e4a0f614377e486f170571bc269d9117c1/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1114-L1115

[...]

PR #4707, opened 2019-12-04
[kotlin] add options sortParamsByRequiredFlag and sortModelPropertiesByRequiredFlag

4brunu:
This PR exposes the `sortParamsByRequiredFlag` already implemented in `DefaultCodegen.java` to allow some more customisation.
https://github.com/OpenAPITools/openapi-generator/blob/d0dfe3e4a0f614377e486f170571bc269d9117c1/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1114-L1115

Also also adds a new option `sortModelPropertiesByRequiredFlag` to allow some customisation on the orders of the models.

This also fixes https://github.com/OpenAPITools/openapi-generator/issues/4705 and keeps backward compatibility.

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -71,6 +71,7 @@
     public AbstractKotlinCodegen() {
         super();
         supportsInheritance = true;
+        setSortModelPropertiesByRequiredFlag(true);
```
wing328: I suggest not to set it as the default:

1) it will be a breaking change and can't target master.
2) not everyone wants this to be the default behavior

wing328: As discussed, it's not a breaking change as you've updated the template as well

PR #4708, opened 2019-12-05
[dart-dio] Fixes enum by setting the wireName to the correct value

athornz:
### PR checklist

CC @ircecho @swipesight @jaumard @nickmeinhold

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart-dio/enum.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/enum.mustache
@@ -11,11 +11,10 @@ class {{classname}} extends EnumClass {
       {{#description}}
   /// {{description}}
       {{/description}}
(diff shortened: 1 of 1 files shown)


nickmeinhold:
I haven't worked on this stuff for a while, when I made a generator that used built_value I vaguely remember having to mess around with using value vs name for the wirename and digging into the java to understand the difference but I've forgotten it all now so I'm not much use here sorry.  I'm sure you know what you're doing.  I can't finish without saying something about tests.  There, said it.

athornz:
I'm pretty sure `value` is the right one to use for the `wireName` because that's the value present in json. `name` is a sanitised version of `value`

In my case the values are returned from the api like `Enum_Value` and the `name` ends up being `enumValue`.

nickmeinhold:
Ah right, that’s sounding familiar now, thanks for the explanation.

Issue #4700, opened 2019-12-04
[BUG] [C-libcurl] The name of field in a struct cannot get escaped when it is the keywords of C such "enum, continue, etc" or includes "-"


ityuhui:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When using C-libcurl generator to generate kubernetes API, some field names are keywords of C programming language, e.g.
```c
object_ *default;
list_t *enum;
```
And some field names include "-", e.g.
```c
int x-kubernetes-int-or-string;
```
The source code fails to compile due to these field names.

##### openapi-generator version
master

##### OpenAPI declaration file content or url
https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

##### Command line used for generation
mvn clean package
java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json -g c -o ../c_k8s_api_client

##### Steps to reproduce
1. generate the c api client by the above command
2. check the file "c_k8s_api_client/model/v1_json_schema_props.h"

##### Related issues/PRs
https://github.com/OpenAPITools/openapi-generator/issues/3970

##### Suggest a fix
does openapi-generator provide a method to escape the name of field in a struct ?

wing328:
@ityuhui thanks for reporting the issue. We need to override the toParamName method to handle these cases. 

Please refer to other generators (e.g. ruby) on how toParamName is implemented.

ityuhui:
Thank you @wing328 
I will take a look at the toParamName

ityuhui:
Hi @wing328 

I found "baseName" cannot be escaped for the keyword, only "name" can.
[...]

PR #4735, opened 2019-12-09
[C-libcurl] Change baseName to name to escape the field name of struct when it in…

ityuhui:
Change "baseName" to "name" to escape the field name of struct when it includes C programming language keyword

As described in [Issue 4700](https://github.com/OpenAPITools/openapi-generator/issues/4700)
I found "baseName" cannot be escaped for the keyword, only "name" can.
I checked this part in [cpp-rest-sdk-client](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/cpp-rest-sdk-client/model-header.mustache) and [cpp-qt5-client](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/cpp-qt5-client/model-header.mustache), they do not use "baseName", but use "name". So I have made these code change.

### PR checklist

@wing328 @PowerOfCreation @zhemant

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
@@ -74,17 +74,17 @@ return 0;
     {{^isContainer}}
     {{^isModel}}
     {{#isEnum}}
-    char* {{baseName}}{{classname}}_ToString({{baseName}}_e {{baseName}}){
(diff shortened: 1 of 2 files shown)


wing328:
Right, let's go with `name` instead of `baseName` but we may need to provide an option to let people choose or we need to implement better serialization between property name in the C client (e.g. created_by) and the JSON key name (e.g. created-by).

Issue #4739, opened 2019-12-09
[REQ][kotlin][client] make serializer configurable


4brunu:
### Is your feature request related to a problem? Please describe.

The serializers provided by the kotlin client generator are not configurable.
It would be nice to have a way to specify different type adapters for gson and Moshi.

## Describe the solution you'd like

A solution to this would be exposing the builder and making the instance of gson and Moshi lazy to provide an opportunity to hook different type adapters.
A similar approach is used with the OkHttp library in the kotlin client generator.

## Additional context

I need to use different date formatters but this is not currently possible, this is why I'm opening this issue.

PR #4740, opened 2019-12-09
[kotlin][client] make serializer configurable

4brunu:
The serializers provided by the kotlin client generator are not configurable.
It would be nice to have a way to specify different type adapters for gson and Moshi.
A solution to this would be exposing the builder and making the instance of gson and Moshi lazy to provide an opportunity to hook different type adapters.
A similar approach is used with the OkHttp library in the kotlin client generator.
I need to use different date formatters but this is not currently possible, this is why I'm opening this PR.
This fixes https://github.com/OpenAPITools/openapi-generator/issues/4739

### PR checklist


jimschubert:
@4brunu this change seems reasonable and valuable. Do you happen to know how the Java static method for a lazy val will work? I ask because I imagine there could be threading issues with configuration and first access, and I wonder if these would need synchronization.

4brunu:
@jimschubert I think it will work the same way that works in kotlin, because @JvmStatic it's just an annotation to better intro with java, to inform java that this is a static method or property.
Here is the difference when calling this from java with and without the annotation.
With the @JvmStatic annotation:
```
Serializer.getGson();
```
Without the @JvmStatic annotation:
```
Serializer.INSTANCE.getGson();
```
When working with lazy, the default thread-safety mode is LazyThreadSafetyMode.SYNCHRONIZED.
I'm not sure if this answers your question, but basically this approach was inspired here
https://github.com/OpenAPITools/openapi-generator/blob/00abb4780c4ac8cfde9301071b653f062555cc79/samples/client/petstore/kotlin-string/src/main/kotlin/org/openapitools/client/infrastructure/ApiClient.kt#L30-L36

PR #4762, opened 2019-12-10
[core] Validation properties standardization

jimschubert:
This should address #4653

The request was to add validation properties to CodegenModel, where CodegenModel may be used to represent reusable schemas which don't necessary result in a generated model (see `isString`, `isInteger`, etc).  Those use cases are legacy functionality from before there was a standardized structure in swagger-parser (now, `Schema`). I couldn't find any usages in template code, but that's not to say that users with customized templates aren't relying heavily on CodegenModel and legacy functionality.  I've added the set of validation properties to `CodegenModel` to address this issue, but I think we should work out a deprecation strategy if CodegenModel is no longer meant to be used for primitive schemas. That is, an OpenAPI Schema representing a primitive type should _never_ be a model in generated code.

The heart of the discussion in the linked issue is this:
* CodegenModel appears to be considered the "most proper" transformation from Schema to some type within OpenAPI Generator
* The ask is to not drop JSON Schema Validation Properties whenever we convert from a `Schema` instance to one of the `Codegen*` types in OpenAPI Generator
* The use case this would allow is model reuse in generated code which requires pre-condition and post-condition checks.

This PR will address issues in our other `Codegen*` types, such as the copy method on `CodegenParameter` would previously not copy pattern, minimum or maximum. CodegenResponse previously had not validation properties. The previous assumption seems to be that CodegenResponse, for instance, didn't require validation; that's not really the case in microservices, and especially not in OLTP where data consistency is key and something like `pattern` or `length` checks could prevent security breaches.

I'm opening this for feedback. If reviewers feel that CodegenModel does not require these validation properties, I'd prefer to remove them from that class.

cc @spacether 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
@@ -19,15 +19,31 @@ package org.openapitools.codegen;
 
 import java.util.*;
 
-public class CodegenResponse {
+public class CodegenResponse implements IJsonSchemaValidationProperties {
     public final List<CodegenProperty> headers = new ArrayList<CodegenProperty>();
-    public String code, message;
+    public String code;
+    public String message;
     public boolean hasMore;
     public List<Map<String, Object>> examples;
-    public String dataType, baseType, containerType, pattern;
+    public String dataType;
+    public String baseType;
+    public String containerType;
     public boolean hasHeaders;
-    public boolean isString, isNumeric, isInteger, isLong, isNumber, isFloat, isDouble, isByteArray, isBoolean, isDate,
-            isDateTime, isUuid, isEmail, isModel, isFreeFormObject;
+    public boolean isString;
+    public boolean isNumeric;
+    public boolean isInteger;
+    public boolean isLong;
+    public boolean isNumber;
+    public boolean isFloat;
+    public boolean isDouble;
+    public boolean isByteArray;
+    public boolean isBoolean;
+    public boolean isDate;
+    public boolean isDateTime;
+    public boolean isUuid;
+    public boolean isEmail;
+    public boolean isModel;
+    public boolean isFreeFormObject;
     public boolean isDefault;
     public boolean simpleType;
     public boolean primitiveType;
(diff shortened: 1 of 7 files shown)


wing328:
> but I think we should work out a deprecation strategy if CodegenModel is no longer meant to be used for primitive schemas. That is, an OpenAPI Schema representing a primitive type should never be a model in generated code

Our replies in the discussion are pretty clear on this. I'll add an FAQ later to make sure users are aware of this.

jimschubert:
@wing328 our replies in the other thread are clear that we don't intend to use CodegenModel to represent primitive schemas, and we've all agreed on this.

What wasn't clear to me, and most likely to the community as well, is that these Boolean fields on CodegenModel which appear to support primitives are really meant to only refer to enum types. You clarified this with me privately, but I don't think it was clear in the linked thread and it wasn't clear to me when I opened this issue. I plan to add comments to each of these fields to clarify in code.

Rather than my proposed deprecation, we might want to consider clarification on the names themselves for 5.0. Example: change `isNumber` to `isEnumNumber`.

wing328:
Added to the FAQ: https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#hows-the-primitive-type-handled-in-the-generator

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
```diff
@@ -79,30 +113,152 @@ public boolean equals(Object o) {
...
+    }
+
+    @Override
+    public void setMaxProperties(Integer maxProperties) {
+        this.maxProperties = maxProperties;
     }
 
     @Override
     public String toString() {
-        final StringBuffer sb = new StringBuffer("CodegenResponse{");
```
wing328: FYI. This is auto-generated by Intelllji IDEA

jimschubert: Are you commenting about the removed line or my modified line? My line was my IntelliJ default auto-generated output.

spacether:
For context here adding IJsonSchemaValidationProperties to CodegenParameter and CodegenResponse is problematic in 3.X.X specs because:
[Parameter](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#parameterObject)
- A parameter holds a schema but it is not a schema
- Both Paramerer.schema (CodegenProperty) and Parameter (CodegenParameter) contain required as a property and the values can disagree.
- Our java tooling conflates Parameter.schema with Parameter so values for ref, and required have unclear data sources at the CodegenParameter level if Parameter.schema data is loaded into CodegenParameter
For this reason, I think that we should
- Remove IJsonSchemaValidationProperties from CodegenParameter
- Switch all users of those properties in CodegenParameter to use CodegenParameter.schema instead which is an instance of CodegenProperty and has the same schema properties

The same reasoning applies to CodegenResponse
[Response](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#responseObject) stores its schemas in: content.contentType.schema
If a response has multiple content types, then CodegenResponse's type properties cannot store them all, one should look in 
CodegenResponse.content and iterate over contentType schemas to use all of the response types. Those are instances of CodegenProperty.
Our tooling conflates the first response's schema with CodegenResponse when it looks like CodegenResponse is meant to be a representation of the higher Response object.

If this is too big of a change, then I am also open to some generators using newer classes that more closely represent
Parameter and Response in our project which are not CodegenParameter and CodegenResponse. Perhaps they could be
CodegenParameterV3Spec and CodegenResponseV3Spec.
What do you think @wing328 ?

PR #4769, opened 2019-12-11
routesのレスポンスバリデーション

takoring:


wing328:

wing328:
Looks like it's not a breaking change so it should target master instead.

modules/openapi-generator/src/main/resources/Javascript/partial_model_generic.mustache
```diff
@@ -77,6 +77,13 @@
       {{/parentModel}}
       {{/useInheritance}}
       {{#vars}}
+      {{#requiredValidate}}
```
wing328: I don't see any reference to `requiredValidate` in the code base. Where does this tag come from?

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
```diff
@@ -21,6 +21,7 @@ import okhttp3.ResponseBody
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 {{/jvm-okhttp4}}
 import okhttp3.Request
+import okhttp3.Credentials
```
4brunu: Why is it adding this import?

modules/openapi-generator/src/main/resources/Javascript/partial_model_generic.mustache
yoshimotocamobile: additional-properties requiredValidate implementation addition

yoshimotocamobile:
> Looks like it's not a breaking change so it should target master instead.

I use model for a response of backend.
As it becomes the analysis error when required does not enter in sdk of kotlin-client or swift-client, I do validation at the time of a response in a server side

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm/infrastructure/ApiClient.kt.mustache
4brunu: @yoshimotocamobile Thanks for your contribution 🙂 
Since https://github.com/OpenAPITools/openapi-generator/pull/4772 was merged and already fixes this, could you please revert this change? Thanks

PR #4781, opened 2019-12-13
[Kotlin][Client] Added new option to enable moshi code generator

Mordag:
Moshi introduced a new code generator that provides a better performance than the current used KotlinJsonAdapterFactory solution. To enable the generator, every data class needs to use the new annotation `@JsonClass(generateAdapter = true)`.

Changes made with this PR:
- Added new option to enable moshi codegen. Option available via `moshiCodeGen`.
- Updated kotlin to 1.3.61.
- Updated moshi related dependencies to 1.9.2.
- Updated samples

### PR checklist


4brunu:
Is there any disadvantage in this approach versus the old one? If not, I wouldn't mind that this would be the default behaviour without adding extras flags.

Mordag:
@4brunu yep there is. The new one improves the performance around converting json to objects + it does not use kotlin-reflects which reduces the project size by around 2-3mb. It should also simplify the usage of R8 (proguard). It would say that is significant.

4brunu:
@Mordag thanks for the explanation, In this case, I think it's better to have a flag and let people choose which approach they prefer.

wing328:

PR #4787, opened 2019-12-13
[go-experimental] Use builder pattern for requests

zippolyte:
Implement builder pattern to construct requests and pass all parameters. Similar construct as #4666 

The motivation for this is to keep a client that will stay backward compatible in case some required parameters become optional which is not a breaking change in itself, but with the current implementation, it would change the method signature, thus introducing a breaking change in the client upon new generation.

Only path parameters are always required and any change would break the API itself, which is why I left them in the function signature. 
cf https://swagger.io/specification/#parameterObject:
> Determines whether this parameter is mandatory. If the parameter location is "path", this property is REQUIRED and its value MUST be true. 

Additionally, this removes the use of the `antihax/optional` module for query parameters, as we don't need the explicit null functionality here. 
cf https://github.com/OpenAPITools/openapi-generator/issues/522#issuecomment-523896869

Any feedback/suggestion would be much appreciated.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
@@ -46,6 +46,8 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
     protected String packageName = "openapi";
     protected Set<String> numberTypes;
 
+    protected boolean usesOptionals = true;
+
     public AbstractGoCodegen() {
         super();
 
@@ -400,7 +402,7 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
                 }
 
                 // import "time" if the operation has a required time parameter.
-                if (param.required) {
+                if (param.required || !usesOptionals) {
                     if (!addedTimeImport && "time.Time".equals(param.dataType)) {
                         imports.add(createMapping("import", "time"));
                         addedTimeImport = true;
@@ -414,7 +416,7 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
                 }
 
                 // import "optionals" package if the parameter is optional
-                if (!param.required) {
+                if (!param.required && usesOptionals) {
                     if (!addedOptionalImport) {
                         imports.add(createMapping("import", "github.com/antihax/optional"));
                         addedOptionalImport = true;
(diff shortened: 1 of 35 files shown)


antihax:
I recall there was a situation where we may need to specify path, qs, or body parameters as you could have a query or path parameter with the same name as a body parameter, etc.

The other projects using a builder pattern (Peal notably iirc) would specify this. @wing328 is this something to consider in this new design? I will see if i can find the example.

zippolyte:
Hmm that seems fair. 
I would argue though that the current approach would have that name clash problem too, since the different kinds of parameters in either the method signature, or the Opts object.

I actually remember having encountered this in our own specs, we had a path parameter that was name `foo_bar` and the endpoint accepted a body parameter of type `FooBar`. It resulted in a method like `test(fooBar string, fooBar FooBar)` so the code did not compile, and we just decided to change the name of the path parameter

antihax:
It does, but so do many of the other clients in other languages. The solution i had seen (i thought was the perl version, but not anymore) was to use something like:
```
client.Function
   .BodyParameters(name, id)
   .FormParameters(name, pet);
```

zippolyte:
The problem with this one is that it still has the problem of generating client that are cannot be backward compatible when for example a body parameter is added in the API.

This is the main problem I am trying to solve with the approach in that PR.

zippolyte:
After giving more thought to this, I think the name clashing issue is not too big a problem anyway. It's easily fixable in the specification itself by changing the names of your path parameters for example, and there even is a `x-codegen-request-body-name` extension for the specific case of the body param name.
As you said also, it looks like things have been refactored away from the separation of parameter types.
With that said, what's your opinion on the approach I'm trying to take to solve the backward compatibility issue, while also retaining a decent UX ?

Also @wing328 or @jimschubert, I'm not too sure what's going on with the circleCI tests, can you advise here ?

Thanks all !

wing328:
I've pushed a fix to the master. Please merge the latest master into this branch to resolve the build error. Sorry for the inconvenience caused.

jimschubert:
AppVeyor looks like an unrelated issue with sockets closing in the C# client test.

zippolyte:
Bumping, @antihax do you have any other feedback regarding this ?

jimschubert:
I've skimmed this quickly (I'm by no means a Go expert), and I think it looks good.

I wonder if we should target the 4.3.x branch because it'll break existing client code (see diff on tests).

zippolyte:
Yeah it will definitely break existing code. I felt like I could given this is the experimental template, but happy to target 4.3.x then.

wing328:
@zippolyte for go-experimental (but not other generators), it's ok to have breaking changes in the master.

jimschubert:
Awesome. Then yeah we can merge if there are no other concerns.

zippolyte:
@jimschubert OK to merge then ?

jimschubert:
Yeah, since it's been almost a week with no other concerns from contributors I think it's good to go.

zippolyte:
Thanks for merging ! 🙇

PR #4793, opened 2019-12-15
[Kotlin][Client] Update Reserved Keywords

javils:
### PR checklist

### PR Description
Update reserved keywords following the official documentation: https://kotlinlang.org/docs/reference/keyword-reference.html and https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/renderer/KeywordStringsGenerated.java


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -91,70 +91,35 @@ public AbstractKotlinCodegen() {
 
         // this includes hard reserved words defined by https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/renderer/KeywordStringsGenerated.java
         // as well as keywords from https://kotlinlang.org/docs/reference/keyword-reference.html
-        reservedWords = new HashSet<String>(Arrays.asList(
-                "abstract",
-                "annotation",
+        reservedWords =  new HashSet<String>(Arrays.asList(
+                "package",
```
4brunu: @javils The list seems a bit incomplete now...
I think you need to add also the keywords from https://kotlinlang.org/docs/reference/keyword-reference.html.
Here are some examples of missing keywords, but there are many more.
```
abstract
lateinit
override
```

javils: @4brunu It's not necessary because abstract, lateinit or override are soft keywords, so you can name your model or your params with these names, you can write something like:
`class abstract { 
val lateinit: Srtring = ""
fun override() = ""
}`

It's totally correct and the compiler work properly, but you can't use the hard keywords in the last example without a compile error.

4brunu: Hum, although that works, I'm not sure if that's a good practice. 
Let's wait for the other reviewers to see what they think.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -91,70 +91,35 @@ public AbstractKotlinCodegen() {
...
-                "val",
-                "var",
-                "vararg",
+                "else",
+                "while",
+                "do",
                 "when",
-                "while"
+                "interface",
+                "typeof"
```
wing328: Can you keep the list sorted with alphabetical order?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractKotlinCodegen.java
wing328: I tested with using `abstract` in model property name, operation ID and the code still compiles. I'm ok these 3 keywords get removed for the time being until someone reports an issue.

wing328:
Did some tests with final, public, enum and the result is good.

PR #4794, opened 2019-12-15
[Kotlin]Remove unnecessary kotlin prefix in .kt files

javils:
### PR checklist

### PR Description
Removed unnecessary "kotlin." prefix in all .kt files.

This PR it's so big because all the examples have had to be regenerated. So please, If someone can take a look at the examples I would be very grateful.


4brunu:
Hum, I'm not sure if they are there for a reason.
@jimschubert can you help here? 🙂

jimschubert:
The `kotlin.` prefix exists because this tool allows users to remap types and instantiation types. If you move away from fully qualified types (remove `kotlin.`), it will break this functionality.

jimschubert:
Forgot to mention, another reason we fully qualified types is to avoid conflict with user types. As an example, try your branch and master with a spec that has a schema called something like `Int`, `List` or `Set`.

javils:
But, this could happen with any language if I understand well. In Java for example you can have the same problem an we don't put any prefix like "java.lang." Instead, this could be broke easily in a .mustache file, it's easy to introduce this kind of bug easily.

Anyway, how can I test this remap? There are any example in the project to try to reproduce and find a solution for this?

Thanks!

wing328:
> In Java for example you can have the same problem an we don't put any prefix like "java.lang." Instead,

In java client generator, there's an option:

```
	fullJavaUtil
	    whether to use fully qualified name for classes under java.util. This option only works for Java API client (Default: false)
```

Not exactly the same thing as discussed in this PR but worth mentioning.

Is it correct to say that the current output (kotlin code) works in your use case but just that you don't like fully-qualified types in the output?

javils:
Yes, the output it's ok, but it's redundant have "kotlin." prefix in all the property models. It's only an aesthetic change.

jimschubert:
> But, this could happen with any language if I understand well.

This is something we'll need to test for. In some generators (like C#), we either consider common data structure names to be "keywords" which triggers a rename, or we fully qualify types where possible.

I wonder if we added support for a linter/formatter to Kotlin, if that would resolve the concerns? This way it offloads the concern of language syntax to tooling where it belongs (not us) and you get the added benefit of cleaning up and sorting unused imports as well.

wing328:
@javils please PM me via Slack to further discussion this when you've time.

javils:
> @javils please PM me via Slack to further discuss this when you've time.

Hi, I'm out of my PC for at least 2 weeks, but when I'm coming back I'll send you a PM. Thank you!

kuFEAR:
@javils Hi, where did you stop in the PR? can you update your the PR with master?

PR #4795, opened 2019-12-15
[Kotlin][Client] Add support for JodaTime

javils:
### PR checklist

### PR Description
Add support for joda time library. I know that it's a not supported library as you can see [here](https://github.com/JodaOrg/joda-time) but there are a lot of java project using this library and they are very coupled to this library. 

So I think that we can offer a easy transition for these project supporting JodaTime and they could migrate to kotlin and after that, remove these dependencies if they want.


samples/client/petstore/kotlin-jodatime/src/main/kotlin/org/openapitools/client/infrastructure/LocalDateAdapter.kt
```diff
@@ -0,0 +1,19 @@
+package org.openapitools.client.infrastructure
+
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import java.time.LocalDate
```
4brunu: @javils Those imports are wrong, they should be from joda time

samples/client/petstore/kotlin-jodatime/src/main/kotlin/org/openapitools/client/infrastructure/LocalDateTimeAdapter.kt
```diff
@@ -0,0 +1,19 @@
+package org.openapitools.client.infrastructure
+
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import java.time.LocalDateTime
```
4brunu: @javils Those imports are wrong, they should be from joda time

jimschubert:
Has this PR been abandoned?

@4brunu should we add support for Joda via separate PR? I've never found a need for Joda, so I don't know how popular it is or why one would prefer this over Java time. We also support template customization with minimal effort, so I'd personally prefer having users make that simple customization rather than implement joda, threetenp, etc across all JVM generators.

4brunu:
@jimschubert I don't need the support for JodaTime. I was just trying to help with the code review 🙂 
I think the only reason to add support for JodaTime is in legacy projects.

wing328:
@javils did you have time to review the feedback? Is this PR still needed?

jimschubert:
@wing328 I think the PR can't be merged until the imports are fixed anyway. Adding JodaTime support as another lib to maintain when Java Time API is now preferred seems to add more maintenance overhead which hasn't had a lot of community request and which is easily addressed using custom templates.

jimschubert:
Haven't heard back in regards to the discussion here.
I think rather than incorporating this into Kotlin, users who need it can use existing functionality (import/type mappings, or template customization) to achieve this.

PR #4812, opened 2019-12-17
[bug][dart] Allow derived generators to call DefaultCodegen#processOpts

jimschubert:
DefaultCodegen#procesOpts sets up quite a bit which may be expected by
users (template directory overrides, git user/repo, etc). The way the
dart-dio and dart-jaguar processOpts methods were written, these things
did not work. This makes it work.

cc all dart contributors for awareness:
   * Dart: @yissachar
   * Dart (refactor): @joernahrens
   * Dart 2: @swipesight
   * Dart (Jaguar): @jaumard
   * Dart (Dio): @athornz

To repro the bug, copy the dart `README.mustache` (for dart-dio or dart-jaguar) to some directory called `template`, make some changes to the README, for example add to the top:

```
# Testing https://{{gitHost}}/{{gitUserId}}/{{gitRepoId}}.git
```

Generate with:

```
openapi-generator generate --generator-name dart-dio -t $(pwd)/template --input-spec openapi.yaml --output client/
```

We'd expect this README to be generated, but as of current master it is not. This was reported as a question by a member of the community in Slack. I thought it was going to be a quick assist, but turned out to be an issue. There's no associated GitHub issue, although I'd image this might resolve other issues with dart-dio and dart-jaguar because there's an attempted workaround in the extended `DartClientCodegen` which leads me to believe there was a known confusion with `templateDir`:

```
        final Object isSupportDart2 = additionalProperties.get(SUPPORT_DART2);
        if (Boolean.FALSE.equals(isSupportDart2) || (isSupportDart2 instanceof String && !Boolean.parseBoolean((String) isSupportDart2))) {
            // dart 1.x
            LOGGER.info("Dart version: 1.x");
            supportingFiles.add(new SupportingFile("analysis_options.mustache", "", ".analysis_options"));
        } else {
            // dart 2.x
            LOGGER.info("Dart version: 2.x");
            // check to not overwrite a custom templateDir
            if (templateDir == null) {
                embeddedTemplateDir = templateDir = "dart2";
            }
        }
```

I don't follow why the if final if condition in the above snippet exists. Maybe someone familiar with these generators could add more of a comment (why are we setting dirs here when they should be handled in constructor and super.processOpts?)

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
@@ -173,9 +173,13 @@ public class DartClientCodegen extends DefaultCodegen implements CodegenConfig {
         return "Generates a Dart (1.x (deprecated) or 2.x) client library.";
     }
 
+    protected void defaultProcessOpts() {
+        super.processOpts();
+    }
+
     @Override
     public void processOpts() {
-        super.processOpts();
+        defaultProcessOpts();
 
         if (StringUtils.isEmpty(System.getenv("DART_POST_PROCESS_FILE"))) {
             LOGGER.info("Environment variable DART_POST_PROCESS_FILE not defined so the Dart code may not be properly formatted. To define it, try `export DART_POST_PROCESS_FILE=\"/usr/local/bin/dartfmt -w\"` (Linux/Mac)");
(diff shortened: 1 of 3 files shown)


wing328:
> I don't follow why the if final if condition in the above snippet exists. 

If I remember correctly, we won't be able to tell if the dart 2 option is enabled in the constructor and that's why I did it in processOpts. That could be a better way to handle this but I think overall using an option to determine the correct template folder (e.g. dart1 vs dart2) is not a good idea.

PR #4872, opened 2019-12-27
scala set and default values for set and array. change in array type

fokot:
Changes:
- if `uniqueItems: true` then generate `Set` otherwise `Seq`
- changed default implementation of `Seq` from `ListBuffer` to `List`
- default value for set is `Set.empy` and for array `List.empy`

### PR checklist


fokot:
I touched only `Scala` and there are failing tests in different languages which were failing even before. Is it ok?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
```diff
@@ -113,7 +113,7 @@ public ScalaAkkaClientCodegen() {
         typeMapping.put("binary", "File");
         typeMapping.put("number", "Double");
 
-        instantiationTypes.put("array", "ListBuffer");
+        instantiationTypes.put("array", "List");
```
jimschubert: This is a breaking change, and won't work because templates expect mutation.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractScalaCodegen.java
```diff
@@ -187,7 +187,9 @@ public String getTypeDeclaration(Schema p) {
         if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return getSchemaType(p) + "[" + getTypeDeclaration(inner) + "]";
+            boolean isSet = ModelUtils.isSet(ap);
```
jimschubert: I feel like this should be done for List as well.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -386,6 +386,11 @@ public static boolean isArraySchema(Schema schema) {
         return (schema instanceof ArraySchema);
     }
 
+    public static boolean isSet(ArraySchema schema) {
```
jimschubert: I think this needs to consider List as well.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalazClientCodegen.java
```diff
@@ -76,8 +76,6 @@ public ScalazClientCodegen() {
         importMapping.remove("List");
         importMapping.remove("Set");
         importMapping.remove("Map");
-        
-        importMapping.put("ListBuffer", "scala.collection.mutable.ListBuffer");
```
jimschubert: This is a ~badass~ breaking change and shouldn't be removed.

Edit: all your changes are badass, but I meant "breaking".

jimschubert:
Thanks for the contribution. Don't worry about the failures for now, we can restart the jobs once master is green. Also... sorry for the individual comments on the review, but GitHub kept crashing Chrome and Firefox submitted then all as regular comments.

My main concern is the change from ListBuffer to List. The templates are written in a way that these almost always expect mutation, and List doesn't have an `.add` function. Once master is stable and we restart the jobs in this PR, we should see compilation errors in Scala samples because of this. I had a couple other questions, like whether we should be considering array and list for conversion to set (in one place, you use List.empy for array instantiation which isn't the same)?

wing328:
I'm looking into the CI failure in the master branch...

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
fokot: When do we access generated array in templates? Do you mean tests? Or am I missing something?

fokot:
@jimschubert I changed the default values to `List.empty` as we see them as `Seq` (`scala.collection.Seq`) fields which are immutable. If we want mutation, templates should import `scala.collection.mutable.Seq`. Or is it meant, that people should cast in Scala?

jimschubert:
> I changed the default values to List.empty as we see them as Seq (scala.collection.Seq) fields which are immutable.

@fokot Seq and List are completely different, though (think List vs LinkedList in Java). Seq is also a trait which can represent non-List collections, while List is very limiting on data type to only types which implement Seq. Why not use `Seq.empty`?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
jimschubert: @fokot "templates" here doesn't just refer to the built-in templates under resources. openapi-generator supports user customized templates, so changing the above will break expectations of anyone with customized templates who expect mutability.  Sorry to not be more clear about that.

Also, `instantiationTypes` really only makes sense as something which can instantiate an object. You can see the difference in your test modifications, for example:

```diff
-        Assert.assertEquals(property1.defaultValue, "new ListBuffer[String]() ");
+        Assert.assertEquals(property1.defaultValue, "List.empty ");
```

This might be used in a class. Now, rather than an immutable property reference with a mutable data structure:

```
  prop: ListBuffer[String] = new ListBuffer[String]()
```

We have an immutable type:

```
   prop: List[String] = List.empty
```

while I agree that the latter is more idiomatic, and you can just use the copy method as case classes intend, anyone who has previously generated against openapi-generator will have to change their client call sites or communicate the change to client SDK users.

If possible, could we instead rollback this change in this PR and open another PR against the 5.0 branch for this change? There is a workaround in 4.x to specify the instantiation types and type mappings manually for anyone who wants the functionality you've implemented here. Unfortunately, doing the right think now means introducing a breaking change for those using the template.

fokot:
@jimschubert now I know the misunderstanding. You must be using pre 2.13 Scala. They changed Seq to immutable.Seq in 2.13 https://github.com/scala/scala/releases/tag/v2.13.0

jimschubert:
@fokot I'm confused by your comment. The things I've mentioned are:

* We can't change from ListBuffer to List without breaking clients generated with openapi-generator, and should probably hold off until 5.0 for that change
* Array generates as Seq, I'm not following why we use Seq/Set in some places, but the default instantiation is `List.empty` rather than `Seq.empty` (are you simply banking on List implements Seq here?)
* Why do we convert only array with uniqueItems=true to a Set and not list? I think it's correct to apply this logic using `ArraySchema`, but what happens if someone sets a type mapping from `array` to `list`? I think ModelUtils.isSet would still return as expected but that some of the ternary conditions wouldn't return the target list type.

I'll try to take some time tomorrow and provide more details if needed. However, I don't think we can merge with the current ListBuffer change.

jimschubert:
@fokot I included your contribution and made adjustments based on my comments above. I opened #4926, which has been merged into master. Please take a look and let me know if you any additional concerns that I might have missed in the above discussion.

Issue #4885, opened 2019-12-29
[Python] Recursive/Cyclic definitions cause import error in Python-experimental


ethan92429:
The problem is that with the changes to have the class property types used directly, circular type referencing or recursive definitions doesn't seem to work anymore:

##### OpenAPI declaration file content or url
```YAML
components:
  schemas:
    Dog:
      properties:
        enemy:
          type: object
          $ref: '#/components/schemas/Dog'
```

This should be a valid spec, but the generated python attempts to reference the Dog class in the definition of the Dog class - which clearly doesn't work. This causes a circular import. This looks to be a regression from just the plain python generator - which uses strings to define openApi types. What are your thoughts on this?

##### Command line used for generation

Normal generation for python-experimental using any spec which has a circular reference.

##### Steps to reproduce

Make a spec with a circular reference, run the python-experimental generator, attempt to use the client, and you'll get an import error.

##### Related issues/PRs

This started happening with #4446 . 

##### Suggest a fix

@spacether suggested simply storing a model's parameter types as a class method rather than a dict. Then they will not be loaded at module load time.

ethan92429:
@spacether I attempted a fix. See here: #4886

spacether:
This issue has been resolved by https://github.com/OpenAPITools/openapi-generator/pull/4888

PR #4888, opened 2019-12-29
[Python] Allow models to have properties of type self

spacether:
This updates the python-experimental generator to allow a model to have properties of type self.
An example model is a Player which stores an enemyPlayer also of type player.

See this use case example:
```
  Player:
    type: object
    required:
      - name
    properties:
      name:
        type: string
      enemyPlayer:
        type: object
        $ref: '#/components/schemas/Player'
```
And our python code demonstrating that it works:
```
        jim = petstore_api.Player(
            name="Jim",
            enemy_player=petstore_api.Player(name="Sam")
        )
```
This was fixed by converting class.openapi_types into a class method, which prevents the openapi_types code from being run at import time.
We also removed the module name from the dataType of the property if the property in self's model.

### Test Verification
A test has been added demonstrating that this works at https://github.com/spacether/openapi-generator/blob/issue_4885_allow_self_properties/samples/client/petstore/python-experimental/test/test_player.py#L32

If merged, this will resolve:
https://github.com/OpenAPITools/openapi-generator/issues/4885

Python technical-committee:

@ethan92429


ethan92429:
@spacether I just ran into another issue using this PR. It is in the resolution of the discriminator. If a model specifies that it is a grandchild of the base model, the discriminator for that grandkid won't be found because the client only looks in the discriminators of the base. The way around this would be to specify _every_ subtype possible, which would be extraordinarily ugly for a large tree that inherits all from one main type (a common pattern for my company). Instead, I think the client should recursively search the discriminator tree to find the correct type. Specifically, [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L594) we should look for a match recursively down the discriminator tree. Would that still honor the OAS?

spacether:
@ethan92429 I won't know until I see an example of your use case. Can you create and share a sample spec?

ethan92429:
Example spec: 
```YAML
components:
  schemas:
    Animal:
      type: object
      required:
      - pet_type
      properties:
        type_key:
          type: string
      mapping:
        pet: '#/components/schemas/Pet'
    Pet:
      type: object
      properties:
        type_key:
          type: string
      discriminator:
        propertyName: type_key
        mapping:
    Cat:
      type: object
      properties:
        type_key:
          age: number
```
Now if the client tries to deserialize something that looks like {'type_key' : 'cat'} and it is known to be of type 'Animal',  it will just not find the correct discriminated class (the Cat class). To work correctly, it would need to recursively search through all the discriminators of Pet as well. However, I'm not sure this is supported in OAS... Is the right way to solve this by making all my discriminators class-specific? For instance, changing 'type_key' to 'animal_type_key' and 'pet_type_key', respectively? That way an instnance of a Cat would specify both  'animal_type_key' and 'pet_type_key'... Thoughts?

ethan92429:
@spacether an example of code that would support this behavior: https://github.com/ethan92429/openapi-generator/commit/ffff978a691d6db10b4f9d7bdb483f78e745fa69

spacether:
Thank you for making that sample. It looks like this issue is unrelated to properties of type self. To help us fix it can you please post it as a new issue?
Should Cat allOf inherit Pet? Shoiuld Pet allOf inherent Animal? Why not combine the Pet and Animal classes?

ethan92429:
Yes - all inherit Animal and Cat inherits Pet. This is just the simplest example I could come up with to illustrate the issue - the spec that I'm using is 77,000 lines of JSON, so posting that wouldn't have been helpful :-) . There are many, many examples of this type of inheritance in our codebase. In our homegrown clients, we have a single 'discriminator mapping' dictionary that can relate _any_ of our type mapping key values to their respective type. This way we flatten the type graph and achieve good performance. It would be really neat if there was some way to support this use case. I can write up an issue, but TBH, I'm not sure what the right solution is...

spacether:
Your posted diff is the right solution. I understand your use case better now. Let's move the discussion there or to an issue if you make it. Either is good.

ethan92429:
#4912 is the issue I posted. Moving discussion there.

PR #4893, opened 2019-12-31
[C-libcurl] The JSON key name in request/response body should not be escaped even though it is a C key word.

ityuhui:
When a struct field in the generated c client is a C key word (e.g. continue), it should be escaped.  This is implemented by the [PR #4735](https://github.com/OpenAPITools/openapi-generator/pull/4735).
But #4735 introduces an issue, that is the JSON key name in request/response body is also escaped.
e.g.
```
./model/v1_list_meta.c:    if(cJSON_AddStringToObject(item, "_continue", v1_list_meta->_continue) == NULL) {
./model/v1_list_meta.c:    cJSON *_continue = cJSON_GetObjectItemCaseSensitive(v1_list_metaJSON, "_continue");
```
This will cause logic error when API server processes the request/response body.
Actually the JOSN key name should not be escaped, e.g.

```
./model/v1_list_meta.c:    if(cJSON_AddStringToObject(item, "continue", v1_list_meta->_continue) == NULL) {
./model/v1_list_meta.c:    cJSON *_continue = cJSON_GetObjectItemCaseSensitive(v1_list_metaJSON, "continue");
```
 

### PR checklist

@wing328 @zhemant

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
@@ -282,18 +282,18 @@ cJSON *{{classname}}_convertToJSON({{classname}}_t *{{classname}}) {
     {{^isContainer}}
     {{#isPrimitiveType}}
     {{#isNumeric}}
-    if(cJSON_AddNumberToObject(item, "{{{name}}}", {{{classname}}}->{{{name}}}) == NULL) {
+    if(cJSON_AddNumberToObject(item, "{{{baseName}}}", {{{classname}}}->{{{name}}}) == NULL) {
     goto fail; //Numeric
     }
     {{/isNumeric}}
     {{#isBoolean}}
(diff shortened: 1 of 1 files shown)


zhemant:
I would like to suggest something, I had faced issue with the struct names, enum values being repeateded or the api contains same struct name but different internal parameters. 

To overcome this add the complete name i.e. project name to make each variable, function unique. This has helped me to overcome a lot of issue. only problem is the length of variables increases. But it helps in enums(when two enums have same value, we get compile error in c if in same project) also when there are two structs with same name.

if (!("object".equals(name))) {
            name = this.projectName + "_" + name;
        }

This is the logic i have added in toModelName function in .java file. If you need to talk more, you can ping me on gitter.

ityuhui:
Thank you @zhemant 
Your suggestion is very helpful! I will try it when facing the name issue next time. 

For this PR, it is used to fix the name that should not be escaped in REST request/response based on the #4735.

wing328:
@zhemant thanks for the review and the suggestion.

Agreed with @ityuhui that this PR is safe to be merged.

PR #4897, opened 2019-12-31
[Rust Server] fix handling of special characters in path param names

aandyl:
The rust server generator used the names of path parameters directly as the names of regex capture groups and format string substitutions. This is invalid when the path parameter is not a legal rust identifier. I couldn't find a specification of what characters OpenAPI permits in these names, and at least one other generator (python-flask, I think) was sanitizing path param names, so that's what I've done here.

Since I've changed the templates to use `{{paramName}}` instead of `{{baseName}}` this has the side effect of changing the previously working camel-case `petId` to `pet_id`. This seems desirable from a rust style perspective, but I'm not sure if it could be a backwards compatibility concern.

### PR checklist

- [x] Copy the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language. @frol @farcaller @bjgill @richardwhiuk

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
@@ -523,6 +523,16 @@ public class RustServerCodegen extends DefaultCodegen implements CodegenConfig {
         Map<String, Schema> definitions = ModelUtils.getSchemas(this.openAPI);
         CodegenOperation op = super.fromOperation(path, httpMethod, operation, servers);
 
+        String pathFormatString = op.path;
+        for (CodegenParameter param : op.pathParams) {
+            // Replace {baseName} with {paramName} for format string
+            String paramSearch = "{" + param.baseName + "}";
+            String paramReplace = "{" + param.paramName + "}";
+
+            pathFormatString = pathFormatString.replace(paramSearch, paramReplace);
+        }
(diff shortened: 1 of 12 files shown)


wing328:
> This seems desirable from a rust style perspective, but I'm not sure if it could be a backwards compatibility concern.

Let's consider this as a bug fix and see if anyone has issues with backward-compatibility.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -523,6 +523,16 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
...
 
+        String pathFormatString = op.path;
+        for (CodegenParameter param : op.pathParams) {
+            // Replace {baseName} with {paramName} for format string
+            String paramSearch = "{" + param.baseName + "}";
+            String paramReplace = "{" + param.paramName + "}";
+
+            pathFormatString = pathFormatString.replace(paramSearch, paramReplace);
+        }
+        op.vendorExtensions.put("pathFormatString", pathFormatString);
```
wing328: I'll rename this with x-path-format-string to conform the vendor extension naming convention (mentioned in the contributing guidelines)

wing328: Filed https://github.com/OpenAPITools/openapi-generator/pull/5020

Issue #4912, opened 2020-01-02
[Python] Multilevel inheritance discriminator support 


ethan92429:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used? 4.2.3
- [x] Have you search for related issues/PRs? Yes - they are #4446
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

@spacether 

When using OAS to describe a complex polymorphic graph, the client is currently unable to serialize grandchildren of any given type (subclasses of subclasses.) I have made a recursive model lookup function, which recursively looks at the discriminators of the subclasses (discriminated classes) until the leaf nodes, but that has very bad performance, reducing performance to 1/100 of normal deserialization. There is a lot of repeated work done for the same model over and over, so I added caching, and the performance is now roughly 1/10 of what it was. Ideally, there would be some way to specify the whole type mapping dictionary.

##### openapi-generator version

4.2.3

##### OpenAPI declaration file content or url

```YAML
components:
  schemas:
    Animal:
      type: object
      required:
      - type_key
      properties:
        type_key:
          type: string
      mapping:
        pet: '#/components/schemas/Pet'
    Pet:
      type: object
      properties:
        type_key:
          type: string
      discriminator:
        propertyName: type_key
        mapping:
          cat: '#/components/schemas/Cat'
    Cat:
      type: object
      properties:
        type_key:
          age: number
```
##### Suggest a fix

See Description

spacether:
Can you please submit your proped PR fromhttps://github.com/ethan92429/openapi-generator/commit/f606cfb1f1d6dbeecd96536f7ef2bcf51a743591
As a fix for this issue?

ethan92429:
@spacether, #4913 is submitted as a PR - still nowhere close to final, it isn't performant for my use case, so I'd like some advice on how to improve that and whether or not this is inline with the OAS.

PR #4913, opened 2020-01-02
adds Recursive Discriminator support

ethan92429:
For now this is just a draft to start discussion with @spacether 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
@@ -592,8 +592,11 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
 
     used_model_class = model_class
     if model_class.discriminator() is not None:
-        used_model_class = model_class.get_discriminator_class(
-            from_server, model_data)
+        used_model_class = recursive_discriminator_lookup(model_class, from_server, model_data)
+
+    # This is the case if the used_model_class isn't a subclass but rather the parent class itself
+    if not used_model_class:
+        used_model_class = model_class
 
     if issubclass(used_model_class, ModelSimple):
         instance = used_model_class(value=model_data, **kw_args)
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         instance = used_model_class(**kw_args)
     return instance
 
+recursive_discriminators_cache = {}
+
+def recursive_discriminators(model_class):
+    """
+    Return the discriminator mapping for all possible subtypes (recursively found down to the leaves)
+    :param model_class:
+    :return:
+    """
+    if model_class in recursive_discriminators_cache:
+        return recursive_discriminators_cache[model_class]
+    recursive_discriminated_types = model_class.discriminator()
+    if not recursive_discriminated_types:
+        return {}
+    for discriminator_key, discriminator_dict in model_class.discriminator().items():
+        for child_class in discriminator_dict.values():
+            for child_discriminator_key, child_discriminator_dict in recursive_discriminators(child_class).items():
+                if child_discriminator_key in recursive_discriminated_types:
+                    recursive_discriminated_types[child_discriminator_key].update(child_discriminator_dict)
+                else:
+                    recursive_discriminated_types[child_discriminator_key] = child_discriminator_dict
+    recursive_discriminators_cache[model_class] = recursive_discriminated_types
+    return recursive_discriminated_types
+
+def recursive_discriminator_lookup(model_class, from_server, model_data):
+    discr_propertyname_py = list(model_class.discriminator().keys())[0]
+    try:
+        return recursive_discriminators(model_class)[discr_propertyname_py][model_data[model_class.attribute_map[discr_propertyname_py]]]
+    except KeyError as e:
+        return
 
 def deserialize_file(response_data, configuration, content_disposition=None):
     """Deserializes body to file
--- a/samples/client/petstore/python-experimental/petstore_api/model_utils.py
+++ b/samples/client/petstore/python-experimental/petstore_api/model_utils.py
@@ -848,6 +848,10 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         used_model_class = model_class.get_discriminator_class(
             from_server, model_data)
 
+    # This is the case if the used_model_class isn't a subclass but rather the parent class itself
+    if not used_model_class:
+        used_model_class = model_class
+
     if issubclass(used_model_class, ModelSimple):
         instance = used_model_class(value=model_data, **kw_args)
         return instance


spacether:
@ethan92429 can you add a link in the description to https://github.com/OpenAPITools/openapi-generator/issues/4912
If this is merged, that issue should be closed

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+recursive_discriminators_cache = {}
+
+def recursive_discriminators(model_class):
+    """
+    Return the discriminator mapping for all possible subtypes (recursively found down to the leaves)
+    :param model_class:
+    :return:
+    """
+    if model_class in recursive_discriminators_cache:
+        return recursive_discriminators_cache[model_class]
```
spacether: How are we sure that this model class is valid?
We are not checking it with from_server and model_data.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         instance = used_model_class(**kw_args)
     return instance
 
+recursive_discriminators_cache = {}
+
+def recursive_discriminators(model_class):
+    """
```
spacether: Can you add a clearer description of our use case here?
Something like
```
A model can return a child model using it's discriminator. But if that new model also has a discriminator we have not yet reached our final model. Keep recursively calling until we get there.
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+    Return the discriminator mapping for all possible subtypes (recursively found down to the leaves)
+    :param model_class:
+    :return:
+    """
+    if model_class in recursive_discriminators_cache:
+        return recursive_discriminators_cache[model_class]
+    recursive_discriminated_types = model_class.discriminator()
+    if not recursive_discriminated_types:
+        return {}
```
spacether: Our old invocation of get_discriminator_class returns a class.
Shouldn't we return model_class in this case?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+        for child_class in discriminator_dict.values():
+            for child_discriminator_key, child_discriminator_dict in recursive_discriminators(child_class).items():
+                if child_discriminator_key in recursive_discriminated_types:
+                    recursive_discriminated_types[child_discriminator_key].update(child_discriminator_dict)
+                else:
+                    recursive_discriminated_types[child_discriminator_key] = child_discriminator_dict
+    recursive_discriminators_cache[model_class] = recursive_discriminated_types
+    return recursive_discriminated_types
+
+def recursive_discriminator_lookup(model_class, from_server, model_data):
```
spacether: Why aren't we recursively calling this?
```
new_model_class = model_class.get_discriminator_class(from_server, model_data)
if new_model_class.discriminator() is None:
   return new_model_class
return recursive_discriminator_lookup(model_class, from_server, model_data)
```

samples/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -848,6 +848,10 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         used_model_class = model_class.get_discriminator_class(
             from_server, model_data)
 
+    # This is the case if the used_model_class isn't a subclass but rather the parent class itself
+    if not used_model_class:
```
spacether: Our old code always returned a model class.
Why do we need to return None now?
Can you remove the case where this is returning None and just returns the correct class?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+    discr_propertyname_py = list(model_class.discriminator().keys())[0]
+    try:
+        return recursive_discriminators(model_class)[discr_propertyname_py][model_data[model_class.attribute_map[discr_propertyname_py]]]
+    except KeyError as e:
```
spacether: Why not raise an exception?
What does our get_discriminator_class class method do in this case?

spacether:
@ethan92429 can you please add
- sample models in the python-experimental spec here https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/python-client-experimental/petstore-with-fake-endpoints-models-for-testing.yaml
  - For clarity could you create new models that are not used by any other tests?
- a test deserializing on of the base models into a grandchild model in https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/tests/test_deserialization.py

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: Can you add a clearer description of our use case here?
Something like
```
A model can return a child model using it's discriminator. But if that new model also has a discriminator we have not yet reached our final model. Keep recursively calling until we get there.
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
ethan92429: hmm, not sure I understand. The only purpose of this method is to search through the discriminators and add them to the cache - not to return the model class itself.  That is done in `recursive_discriminator_lookup`. Does that make more sense? This is essentially a recursive replacement for `model_class.discriminator`... I could put it into the model_class definition instead... maybe that would be cleaner? unclear.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
ethan92429: This isn't equivalent to get_discriminator_class for performance reasons. This instead is equivalent to model_class.discriminator.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
ethan92429: I had done this before, but the issue was that we have a recursive lookup for a single type potentially many, many times. So instead, I put in a simple cache and changed the logic to 'lookup the recursive discriminator in the cache for the given model_class'. If it isn't there, I calculate the whole recursive discriminator for that type and I store it in the cache. Then any subsequent lookup is just a single dictionary lookup. I had the exact logic that you had just laid out and the deserialization process took 30 seconds+ for a complex model. The cache brought that down to ~3

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+                    recursive_discriminated_types[child_discriminator_key].update(child_discriminator_dict)
+                else:
+                    recursive_discriminated_types[child_discriminator_key] = child_discriminator_dict
+    recursive_discriminators_cache[model_class] = recursive_discriminated_types
+    return recursive_discriminated_types
+
+def recursive_discriminator_lookup(model_class, from_server, model_data):
+    discr_propertyname_py = list(model_class.discriminator().keys())[0]
+    try:
+        return recursive_discriminators(model_class)[discr_propertyname_py][model_data[model_class.attribute_map[discr_propertyname_py]]]
```
spacether: This one line is doing the following:
- retrieving a value from and possible storing data in recursive_discriminators_cache
- accessing the discriminator map
- with a value from model data
- using the js key
- using the python key
Please separate this into multiple lines for readability and understanding

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
ethan92429: I'm refactoring this to be identical to get_discriminator_class. That should resolve a couple of these comments - I agree it is fast and dirty ATM

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+    recursive_discriminated_types = model_class.discriminator()
+    if not recursive_discriminated_types:
+        return {}
+    for discriminator_key, discriminator_dict in model_class.discriminator().items():
+        for child_class in discriminator_dict.values():
+            for child_discriminator_key, child_discriminator_dict in recursive_discriminators(child_class).items():
+                if child_discriminator_key in recursive_discriminated_types:
+                    recursive_discriminated_types[child_discriminator_key].update(child_discriminator_dict)
+                else:
+                    recursive_discriminated_types[child_discriminator_key] = child_discriminator_dict
```
spacether: Can you add a test of this function that shows what we store going down both forks of this if/else?

samples/client/petstore/python-experimental/petstore_api/model_utils.py
ethan92429: This was actually the original error I was looking at before I started looking at the recursive discriminator issue. The method returns nothing when the discriminator specified is not in the discriminator mapping. I believe in that case, a reasonable solution is to return the base class itself, but perhaps we should instead return an error here?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+    """
+    if model_class in recursive_discriminators_cache:
+        return recursive_discriminators_cache[model_class]
+    recursive_discriminated_types = model_class.discriminator()
+    if not recursive_discriminated_types:
+        return {}
+    for discriminator_key, discriminator_dict in model_class.discriminator().items():
+        for child_class in discriminator_dict.values():
+            for child_discriminator_key, child_discriminator_dict in recursive_discriminators(child_class).items():
+                if child_discriminator_key in recursive_discriminated_types:
```
spacether: Do we need to do a check to make sure that child_discriminator_key == discriminator_key?
If not then that implies that we have multiple discriminator keys.
How are we making sure that we are using the correct python or javascript variable name to model_data values?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: That makes sense. When we get this working, please remove the get_discriminator_class if we are no longer using it.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
...
+    :return:
+    """
+    if model_class in recursive_discriminators_cache:
+        return recursive_discriminators_cache[model_class]
+    recursive_discriminated_types = model_class.discriminator()
+    if not recursive_discriminated_types:
+        return {}
+    for discriminator_key, discriminator_dict in model_class.discriminator().items():
+        for child_class in discriminator_dict.values():
+            for child_discriminator_key, child_discriminator_dict in recursive_discriminators(child_class).items():
```
spacether: Can you use clearer names here?
Is child_discriminator_key the discriminator propertyName or a model_class_name?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         instance = used_model_class(**kw_args)
     return instance
 
+recursive_discriminators_cache = {}
+
+def recursive_discriminators(model_class):
+    """
+    Return the discriminator mapping for all possible subtypes (recursively found down to the leaves)
+    :param model_class:
+    :return:
```
spacether: Can you describe the structure of the dict that we are returning here?
What are the keys what are the values?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -609,6 +612,35 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
         instance = used_model_class(**kw_args)
     return instance
 
+recursive_discriminators_cache = {}
```
spacether: Can you describe the structure of the dict that we are storing here?
What are the keys what are the values?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: Ah, I understand

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: Why do we need to store an empty dict if we will never key into it?
Shouldn't we avoid making that last recursive call?

samples/client/petstore/python-experimental/petstore_api/model_utils.py
spacether: Per the spec here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism
If a discriminator is used, then the child class must contain the discriminator propertyName, and by inference the child class must be in the discriminator map.
Please raise an exception that lets the user know what went wrong.

spacether:
@ethan92429 are you okay?
Haven't seen anything here in the last week.

ethan92429:
Just busy with a couple other things and on vacation until Wednesday next
week. Feel free to make the updates if you desire, and thanks for the
review! I'll look more on Wednesday 😊

spacether:
Glad to hear it. There's no rush on this. Also you know the use case much better than I do. Hope that you have a fun and relaxing vacation!

spacether:
@ethan92429 any update on your PR?

PR #4915, opened 2020-01-02
swagger-core 2.1.1 && swagger-parser 2.0.17

sullis:
notes: https://github.com/swagger-api/swagger-parser/releases

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/serializer/SerializerUtilsTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/serializer/SerializerUtilsTest.java
@@ -52,7 +52,7 @@ public class SerializerUtilsTest {
                "      description: Some description\n" +
                "      operationId: pingOp\n" +
                "      responses:\n" +
-               "        200:\n" +
+               "        \"200\":\n" +
                "          description: Ok\n" +
                "components:\n" +
                "  schemas:\n" +
@@ -178,7 +178,7 @@ public class SerializerUtilsTest {
                 "      description: Some description\n" +
                 "      operationId: pingOp\n" +
                 "      responses:\n" +
-                "        200:\n" +
+                "        \"200\":\n" +
                 "          description: Ok\n"; 
         assertEquals(content, expected);
     }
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/test/java/org/openapitools/codegen/serializer/SerializerUtilsTest.java
```diff
@@ -178,7 +178,7 @@ public void testToYamlStringMinimalExample() throws Exception {
                 "      description: Some description\n" +
                 "      operationId: pingOp\n" +
                 "      responses:\n" +
-                "        200:\n" +
+                "        \"200\":\n" +
```
wing328: @OpenAPITools/generator-core-team HTTP response code is now in the form of string instead of integer.

I don't see any mention of this change in swagger core/parser change log or release note.

I think we should highlight this change in our release note so users won't be surprised if the HTTP status code is now a string instead of a integer.

etherealjoy: Can't we convert it before consuming?

wing328: We can but someone may have already processed these numeric status codes in their app and this change may break their app. To fix it, the users need to do the conversion.

jimschubert: @wing328 this is documented on the Responses object:

> This field MUST be enclosed in quotation marks (for example, "200") for compatibility between JSON and YAML.

jimschubert: I believe YAML keys are always considered strings, so someone parsing this manually through YAML should be unaffected. We could verify in YAML spec, but probably lower priority than fixing a bug that makes the doc non-compliant with the spec.

wing328: OK. Let's go ahead with this change.

PR #4922, opened 2020-01-03
[C-libcurl] Support setting basePath and apiKeys when creating an apiClient

ityuhui:
After generating a C client,  user will use apiClient_create to create an apiClient to execute REST request operation.

But for now,  apiClient_create does not accept any parameter so apiClient cannot get any user setting. 
e.g.
```c
apiClient_t *apiClient_create() {
    curl_global_init(CURL_GLOBAL_ALL);
    apiClient_t *apiClient = malloc(sizeof(apiClient_t));
    apiClient->basePath = "http://localhost";
    apiClient->dataReceived = NULL;
    apiClient->response_code = 0;
    apiClient->apiKeys = NULL;

    return apiClient;
}
```

I updated the template code, let the apiClient_create can get user settings:
```c
apiClient_t *apiClient_create(const char *basePath
, list_t *apiKeys
) {
    curl_global_init(CURL_GLOBAL_ALL);
    apiClient_t *apiClient = malloc(sizeof(apiClient_t));
    if(basePath){
        apiClient->basePath = strdup(basePath);
    }else{
        apiClient->basePath = strdup("http://localhost");
    }
    apiClient->dataReceived = NULL;
    apiClient->response_code = 0;
    apiClient->apiKeys = apiKeys;

    return apiClient;
}
```

I also updated the header file and apiClient_free.

### PR checklist

@wing328 @zhemant

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
@@ -9,10 +9,22 @@
 
 size_t writeDataCallback(void *buffer, size_t size, size_t nmemb, void *userp);
 
-apiClient_t *apiClient_create() {
+apiClient_t *apiClient_create(const char *basePath
+{{#hasAuthMethods}}
+{{#authMethods}}
+{{#isApiKey}}
+, list_t *apiKeys
+{{/isApiKey}}
+{{/authMethods}}
+{{/hasAuthMethods}}
+) {
     curl_global_init(CURL_GLOBAL_ALL);
     apiClient_t *apiClient = malloc(sizeof(apiClient_t));
-    apiClient->basePath = "{{{basePath}}}";
+    if(basePath){
+        apiClient->basePath = strdup(basePath);
+    }else{
+        apiClient->basePath = strdup("{{{basePath}}}");
+    }
     apiClient->dataReceived = NULL;
     apiClient->response_code = 0;
     {{#hasAuthMethods}}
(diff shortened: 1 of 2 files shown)


ityuhui:
I wonder whether I need remove 
```
    {{#isApiKey}}
    if(apiClient->apiKeys) {
        list_free(apiClient->apiKeys);
    }
    {{/isApiKey}}
```
in the function 
```
apiClient_free(apiClient_t *apiClient)
```
because apiClient->apiKeys is passed by the parameter apiKeys of apiClient_create.
Users should manage(new and delete) the memory of apiKeys themselves,  apiClient_free() should not free apiClient->apiKeys because apiClient_create does not allocate memory for apiClient->apiKeys.

This may cause double-free coredump if users free apiKeys after apiClient_free is called.

zhemant:
Yes, a deep copy is okay, as the generated code will manage the memory. But then we should mention this that we are doing a deep copy and the user has to free the memory after the function call anytime. To avoid memory leaks if happens.

same goes for basepath.

ityuhui:
Thanks @zhemant 

I will close this PR and create a new one to deliver "deep-copy"

wing328:
Also what about creating a new method (e.g. *apiClient_create_with_base_path_and_apikeys()) to make it backward compatible?

ityuhui:
@wing328 

It's a good idea to create a new function to keep compibitily.

I suggest the new funcation's name as "apiClient_create_with_base_path" because "list_t *apiKeys" is generated by the template code automatically. If there is no apiKeys in OpenAPI specification,  "list_t *apiKeys" does not exist in parameter list.

```c
apiClient_t *apiClient_create_with_base_path(const char *basePath
{{#hasAuthMethods}}
{{#authMethods}}
{{#isApiKey}}
, list_t *apiKeys
{{/isApiKey}}
{{/authMethods}}
{{/hasAuthMethods}}
) {
```

Actually I will have another PR that adds a new parameter "caPath" to apiClient_create to support  SSL/HTTPS.  I think the new function's name "apiClient_create_with_base_path_and_capath" is also OK.

wing328:
> I suggest the new funcation's name as "apiClient_create_with_base_path" because "list_t *apiKeys" is generated by the template code automatically

sounds good to me 👍

PR #4926, opened 2020-01-04
[scala] Support for Set when array has uniqueItems=true

jimschubert:
This addresses my comments discussed in #4872 

Specifically, I override type/property evaluation and default values so that Set is used consistently across models and properties. I've also added tests.

If CI passes here, I think we can close #4872 and merge into master. I would consider Arrays with unique items generating as List rather than Set to be a bug rather than a breaking change. @wing328 would you agree?

### PR checklist

cc @clasnake @shijinkui @ramzimaalej


jimschubert:
@ramzimaalej  thanks for the review. I added another commit to use default values for monadic `.empty` in the abstract base (for types that might make sense for users to specify as overrides). The scalatra sample hasn't been generated since 3.0.0 and didn't compile for me before or after my change, so I've just reverted the sample.

jimschubert:
I'll go ahead and merge. If there are concerns about it breaking functionality, we can revert and retarget for 4.3.0. But I think the impact is minimal and considered a bug by those who expected Set anyway.

wing328:
@jimschubert agreed with your change 👍

PR #4929, opened 2020-01-05
[dart-dio] Fixes issues with DateTime import and form date paramaterToString function

athornz:
Fixes issues with DateTime import and form date paramaterToString function

- due to the addition of the timemachine library, models were importing ‘DateTime’ when using the core library, which is not a valid import.

- the parameterToString function was copied from the dart2 generator and had some errors when some enums were classes.

CC: @ircecho @swipesight @jaumard @nickmeinhold @amondnet


amondnet:
LTGM

Issue #2730, opened 2019-04-24
[REQ] Support for custom types based on format field


steco:
**Description**

OpenApi-Gen currently supports the following hard-coded "format"s for a string parameter
 - uuid
 - binary
 - email
 - date
 - date-time
 - byte-array
 - binary

I'd like to make this more generic, ie support additional values for the "format" field and use a type-mapping parameter to map them to a specific type in the generated code.  As per https://swagger.io/specification/#dataTypes:

_the format property is an open string-valued property, and can have any value. Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification_

This [commit](https://github.com/finbourne/openapi-generator/commit/ab99bed7b38112de7306cf862240f8b67b09d841) is the change to support this.  Is this an appropriate change to make ?

wing328:
@steco thanks for the suggestion. Yes, we definitely want to support more `format`.

I would suggest opening a PR to start with and see if the CIs catch any issue as a result of the suggested change.

steco:
Thanks William. I've done it here - https://github.com/OpenAPITools/openapi-generator/pull/2799, and it looks like all the CI tests pass.  What's the next step ?

Thanks

chrylis:
Would this change make it possible for me to use a format `myformat` and provide a custom mapping to `com.company.project.MyFormat`?

steco:
Yes.  I've only used this for a C# client but I believe that the change should work for all languages.  You would just need to pass --type-mappings myformat=com.company.project.MyFormat to swagger-codegen-cli.jar

sigpwned:
I have a similar requirement and independently generated what is (essentially) the same PR as linked above. It worked for my Java application as well. I would love to see this get merged and released soon! My team will likely have to maintain a patched version until then. Thanks to @steco for the issue and PR, and to the broader team for a hopefully quick review and merge!

ShoeBoom:
Is it possible to extend this feature to other types. For example my spec uses something similar to

```
"foo": {
    "enum": [
        "0",
        "1",
        "2"
    ],
    "type": "integer",
    "format": "byte"
}
```

the codgen maps this to Int

crtomirmajer:
Any progress on this?

ShoeBoom:
> Any progress on this?

I made a related change in #4931. But it hasn't been active for a while so idk

marcelstoer:
Since both #2799 and #9285 were merged I suggest you close this.

[...]

PR #4931, opened 2020-01-06
[Enhancement] added support for custom type & format mapping

ShoeBoom:
### Description of the PR
Makes it possible to map format.

For example these args can be used to properly map the following:
```
"foo": {
    "enum": [
        "0",
        "1",
        "2"
    ],
    "type": "integer",
    "format": "byte"
}
```

`generate -i openapi.json -o out -g kotlin --type-mappings integer_byte=kotlin.Byte`

By default it would map `foo` to `Int`

this is similar to #2730

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1631,6 +1631,10 @@ public class DefaultCodegen implements CodegenConfig {
     private String getPrimitiveType(Schema schema) {
         if (schema == null) {
             throw new RuntimeException("schema cannot be null in getPrimitiveType");
+        } else if (typeMapping.containsKey(schema.getType() + "_" + schema.getFormat())) {
+            // allows custom type_format mapping.
+            // use {type}_{format}
+            return typeMapping.get(schema.getType() + "_" + schema.getFormat());
         } else if (ModelUtils.isStringSchema(schema) && "number".equals(schema.getFormat())) {
             // special handle of type: string, format: number
             return "BigDecimal";


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1631,6 +1631,10 @@ private String getSingleSchemaType(Schema schema) {
     private String getPrimitiveType(Schema schema) {
         if (schema == null) {
             throw new RuntimeException("schema cannot be null in getPrimitiveType");
+        } else if (typeMapping.containsKey(schema.getType() + "_" + schema.getFormat())) {
+            // allows custom type_format mapping.
+            // use {type}_{format}
+            return typeMapping.get(schema.getType() + "_" + schema.getFormat());
```
jimschubert: @wing328 I like the idea proposed in this PR of supporting type definitions for type+format combinations. I wonder if instead of an underscore, we may want to use something which is unlikely to exist in the wild (like `+`).

Compare:

```
generate -i openapi.json -o out -g kotlin --type-mappings integer_byte=kotlin.Byte
generate -i openapi.json -o out -g kotlin --type-mappings integer+byte=kotlin.Byte
generate -i openapi.json -o out -g kotlin --type-mappings string_secure_string=org.owasp.passfault.impl.SecureString
generate -i openapi.json -o out -g kotlin --type-mappings string+secure_string=org.owasp.passfault.impl.SecureString
```

ShoeBoom: I think `@format:` could be another option. 

something like

`generate -i openapi.json -o out -g kotlin --type-mappings integer@format:byte=kotlin.Byte`

wing328: I've created https://github.com/OpenAPITools/openapi-generator/pull/9285 based on this PR and use `+` instead. (sorry for delay in reviewing this excellent enhancement)

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/9285

PR #4934, opened 2020-01-07
[Golang] Add support allOf inheritance with embedded go structs

sebastien-rosset:
This PR addresses the following:
* The generated go-experimental code is not compiling in some cases.
* Use an embedded struct to model the codegen parent
* The instantiation type of a parent of "map" type is incorrect

In particular, the generated samples/openapi3/client/petstore/go-experimental/go-petstore/api_fake.go file has the following code:
```
 t := *r.pipe
  if reflect.TypeOf(t).Kind() == reflect.Slice {
     ....
  }
  ....
  t := *r.context
```

As you can see, the same variable **"t"** is defined twice, which causes a compilation error:
```
go-petstore/api_fake.go:1782:4: no new variables on left side of :=
```

This PR depends on #5075.

### PR checklist


sebastien-rosset:

sebastien-rosset:
It looks like "continuous-integration/travis-ci/pr " is failing for a reason other than this PR.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
```diff
@@ -328,6 +329,16 @@ public String getSchemaType(Schema p) {
             type = openAPIType;
         return type;
     }
+    
+    @Override
+    public String toInstantiationType(Schema property) {
```
bkabrda: Could you please provide a comment here on what this method does and why it does that? I *think* I understand it in the context of your PR, but if I was just looking at it in the codebase, I'd probably have no idea what's happening here and why.

modules/openapi-generator/src/main/resources/go-experimental/api.mustache
```diff
@@ -134,14 +134,16 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx _context.Context{{#hasParams
 	{{#queryParams}}
 	{{#required}}
 	{{#isCollectionFormatMulti}}
-	t:={{paramName}}
-	if reflect.TypeOf(t).Kind() == reflect.Slice {
-		s := reflect.ValueOf(t)
-		for i := 0; i < s.Len(); i++ {
-			localVarQueryParams.Add("{{baseName}}", parameterToString(s.Index(i), "{{#collectionFormat}}{{collectionFormat}}{{/collectionFormat}}"))
+	{
```
bkabrda: Why is this change necessary? Is it to make sure that the scope of variables defined in this block is limited?

sebastien-rosset:
> So after the explanation provided in [#4607 (comment)](https://github.com/OpenAPITools/openapi-generator/pull/4607#issuecomment-566662587) I think I'm ok with accepting this change. It seems to work ok and makes sense to me. I left one suggestion to improve and one question inline to get some improvement/explanation.
> 
> What I'm not sure about is whether this should go inside the non-experimental go client, as I think it might result in unexpected changes in JSON payloads.

Using Golang embedded structs have zero impact on the JSON payload (vs a flattened structure). The payload remains exactly the same.

> OTOH this feature never worked before, so I guess adding it can also be considered just an improvement/fix and the different JSON payloads will now actually be correct.
> 
> All in all, I'm ok with this being merged (after the suggested change/reply), assuming it gets a 👍 from someone else as well (@antihax WDYT?)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
sebastien-rosset: sure, will do.

bkabrda:
> Using Golang embedded structs have zero impact on the JSON payload (vs a flattened structure). The payload remains exactly the same.

Yeah, but the serialized children will now contain parent attributes, which they didn't before. So the payload is going to be different, is it not?

modules/openapi-generator/src/main/resources/go-experimental/api.mustache
sebastien-rosset: This is totally independent of the embedded go struct enhancement, maybe I should have raised a separate PR, but it takes time to manage multiple PRs.
Without this change, there can be compilation errors and static analysis issues. This is because the block (which is now in curly braces) is repeated multiple times depending on how many query parameters there are in the OAS. 

The issue is that this line will be generated multiple times:
```
t:={{paramName}}
```

sebastien-rosset:
> > Using Golang embedded structs have zero impact on the JSON payload (vs a flattened structure). The payload remains exactly the same.
> 
> Yeah, but the serialized children will now contain parent attributes, which they didn't before. So the payload is going to be different, is it not?

Hmm, do you have an example, maybe I am missing something? Previously, with the following OAS schema:
Parent:
  property p1
  property p2

Child:
  property p3

The generated code for the child was a flattened struct that had p1, p2, p3. Hence in the JSON document these three properties were present.
Now "Parent" is embedded in the child; when the struct is serialized, it will still have the same p1, p2, p3 properties.

sebastien-rosset:
Darn, the rebase from upstream didn't work as I expected. Let me fix this.

bkabrda:
@sebastien-rosset I was looking wrong in terms of my above comments about different JSON, please disregard that.

In terms of changing the non-experimental go client, I think I was looking at an older version of your PR that has contained the same changes for non-experimental and experimental. With the current version of PR I'm +1 (with my Go language committee member hat on, can't speak for the changes to the core files).

sebastien-rosset:
Sorry, I am still trying to figure out why changes are showing up for the following files, they shouldn't.
```
modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/ConfigHelp.java
modules/openapi-generator/src/main/resources/JavaJaxRS/spec/libraries/quarkus/Dockerfile.jvm.mustache
modules/openapi-generator/src/main/resources/JavaJaxRS/spec/libraries/quarkus/application.properties.mustache
modules/openapi-generator/src/main/resources/JavaJaxRS/spec/libraries/quarkus/pom.mustache
pom.xml
```

sebastien-rosset:
> @sebastien-rosset I was looking wrong in terms of my above comments about different JSON, please disregard that.
> 
> In terms of changing the non-experimental go client, I think I was looking at an older version of your PR that has contained the same changes for non-experimental and experimental. With the current version of PR I'm +1 (with my Go language committee member hat on, can't speak for the changes to the core files).

I have fixed the problem with rebase, the unrelated core changes are no longer in the PR.

sebastien-rosset:
It appears there is an unrelated problem with Swift4 tests in master.

sebastien-rosset:
@antihax, can you review this PR? thank you.

sebastien-rosset:
> @sebastien-rosset I was looking wrong in terms of my above comments about different JSON, please disregard that.
> 
> In terms of changing the non-experimental go client, I think I was looking at an older version of your PR that has contained the same changes for non-experimental and experimental. With the current version of PR I'm +1 (with my Go language committee member hat on, can't speak for the changes to the core files).

@bkabrda , are you still ok with this fix? One big problem IMO is that the golang code does not compile. I tried to make it more clear in the PR summary. It would be great to fix this issue. thank you.

samples/client/petstore/go-experimental/go-petstore/model_dog.go
```diff
@@ -16,59 +16,10 @@ import (
 
 // Dog struct for Dog
 type Dog struct {
-	ClassName string `json:"className"`
-	Color *string `json:"color,omitempty"`
+	Animal
```
jimschubert: This is very cool. I'm not a pro in Go by any means, but I've just learned about field promotion in anonymous nested structs. ;)

PR #4937, opened 2020-01-07
[suggestion] Get apis property from either SystemProperties or additionalProperties

shybovycha:
### Introduction

### This change

I could not really find a good way to pass values through `SystemProperties` when using the maven plugin. Hence it is pretty much impossible from my point of view to set the `apis` property, I think that it could also be passed through the `additionalProperties` option.

Also, see the other PR ( #4938 ) which fixes a bug in parsing the list options in `additionalProperties`.

Does not seem like a breaking change, hence targeting `master`.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -530,7 +530,12 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
         }
         Map<String, List<CodegenOperation>> paths = processPaths(this.openAPI.getPaths());
         Set<String> apisToGenerate = null;
-        String apiNames = GlobalSettings.getProperty("apis");
+        String apiNames = GlobalSettings.getProperty(CodegenConstants.APIS);
+
+        if (apiNames == null || StringUtils.isBlank(apiNames)) {
+            apiNames = (String) opts.getConfig().additionalProperties().get(CodegenConstants.APIS);
+        }
+
         if (apiNames != null && !apiNames.isEmpty()) {
             apisToGenerate = new HashSet<String>(Arrays.asList(apiNames.split(",")));
         }


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -530,7 +530,12 @@ private void generateApis(List<File> files, List<Object> allOperations, List<Obj
         }
         Map<String, List<CodegenOperation>> paths = processPaths(this.openAPI.getPaths());
         Set<String> apisToGenerate = null;
-        String apiNames = GlobalSettings.getProperty("apis");
+        String apiNames = GlobalSettings.getProperty(CodegenConstants.APIS);
+
+        if (apiNames == null || StringUtils.isBlank(apiNames)) {
+            apiNames = (String) opts.getConfig().additionalProperties().get(CodegenConstants.APIS);
```
jimschubert: I don't think this will work… it would mean that generators can no longer support an `apis` additional property for other reasons as any setting here would cause no API classes to get generated.

Can you provide an example of how you've configured the Maven plugin? I think it would be best to fix system property support there.

jimschubert:
@shybovycha I've tested the following, and it works as expected:

```
diff --git a/modules/openapi-generator-maven-plugin/examples/kotlin.xml b/modules/openapi-generator-maven-plugin/examples/kotlin.xml
index 3a2979232c..a28a2b6095 100644
--- a/modules/openapi-generator-maven-plugin/examples/kotlin.xml
+++ b/modules/openapi-generator-maven-plugin/examples/kotlin.xml
@@ -66,6 +66,10 @@
 							<apiPackage>kotlintest.org.openapitools.client.api</apiPackage>
 							<modelPackage>kotlintest.org.openapitools.client.model</modelPackage>
 							<invokerPackage>kotlintest.org.openapitools.client</invokerPackage>
+
+							<environmentVariables>
+								<apis>Pet</apis>
+							</environmentVariables>
 						</configuration>
 					</execution>
 				</executions>

```

This generates only `PetApi` and all associated models. I've looked at the readme for the plugin, and the usage is not clear so I'll get that updated.

jimschubert:
Thanks for submitting this PR and for proposing workarounds here and in #4938. I'm going to close this in favor of #5251, which adds an `apisToGenerate` property to match the existing `modelsToGenerate` property. It looks like the `environmentVariables` documentation was missing and wouldn't have been clear that this option was added for defining `systemProperties` (or, `GlobalSettings`). The changes in #5251 should clarify how to do this without workaround and fallbacks.

PR #4939, opened 2020-01-07
[feature] [jaxrs-spec] Use tag by default to group paths

shybovycha:
### Introduction

### This change

Using tag as the primary key to group paths and operations.

This comes extremely handy when used together with `apis` system property (or `additionalProperties` one, as in #4937 ) so the operations could be grouped into resources.

I tested it on our existing project together with `swagger-maven-plugin` (specifically [this hack](https://github.com/kongchen/swagger-maven-plugin/pull/805)) to minimize the change when migrating the whole project to Open-API (mostly automatically).

This definitely is a breaking change (as the resources generated will have different names), so targeting `5.0.x`.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
@@ -224,21 +224,26 @@ public class JavaJAXRSSpecServerCodegen extends AbstractJavaJAXRSServerCodegen {
             basePath = basePath.substring(0, pos);
         }
 
-        String operationKey = basePath;
-        if (StringUtils.isEmpty(basePath)) {
-            basePath = tag;
-            operationKey = "";
-            primaryResourceName = tag;
-        } else if (basePath.matches("\\{.*\\}")) {
-            basePath = tag;
-            operationKey = "";
-            co.subresourceOperation = true;
-        } else {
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaJAXRSSpecServerCodegen.java
```diff
@@ -224,21 +224,26 @@ public void addOperationToGroup(String tag, String resourcePath, Operation opera
...
-            operationKey = "";
-            primaryResourceName = tag;
-        } else if (basePath.matches("\\{.*\\}")) {
-            basePath = tag;
-            operationKey = "";
-            co.subresourceOperation = true;
-        } else {
-            if (co.path.startsWith("/" + basePath)) {
-                co.path = co.path.substring(("/" + basePath).length());
+        String operationKey = tag;
```
jimschubert: I like this idea, however I think it should be guarded by an option the way other generators are doing this (see `useTags` in `SpringCodegen`. It looks like defaulting to tags will also have a negative impact on user upgrades for those relying on the subresource behaviors (or groupings in general). I think it would be best to have the opt-in option for 5.0 and make it default at some point later, maybe 5.1?

shybovycha: like this idea. this should make this more a workaround rather than a breaking change. will change my approach

PR #4940, opened 2020-01-07
[go-experimental] Support aliasing of API keys

jirikuncar:
Support aliasing of API keys so users of generated clients don't have to specify same value under different names.

Example:
```yaml
components:
  securitySchemes:
    apiKeyAuth:
      type: apiKey
      name: api_key
      in: query
      description: API Key
    apiKeyAuthHeader:
      type: apiKey
      name: SECURITY-API-KEY
      x-auth-id-alias: apiKeyAuth
      in: header
      description: API Key
```

One can also alias both names for compatibility reasons or simply nicer usage:
```yaml
components:
  securitySchemes:
    apiKeyAuth:
      type: apiKey
      name: api_key
      x-auth-id-alias: secret
      in: query
      description: API Key
    apiKeyAuthHeader:
      type: apiKey
      name: SECURITY-API-KEY
      x-auth-id-alias: secret
      in: header
      description: API Key
```

### PR checklist


jirikuncar:
Travis CI failure seems unrelated. See https://travis-ci.org/OpenAPITools/openapi-generator/builds/633835021#L20010

jirikuncar:
@wing328 all check have passed 🎉

modules/openapi-generator/src/main/resources/go-experimental/api.mustache
```diff
@@ -242,7 +242,7 @@ func (r api{{operationId}}Request) Execute() ({{#returnType}}{{{.}}}, {{/returnT
 	if r.ctx != nil {
 		// API Key Authentication
 		if auth, ok := r.ctx.Value(ContextAPIKeys).(map[string]APIKey); ok {
-			if auth, ok := auth["{{keyParamName}}"]; ok {
+			if auth, ok := auth["{{#vendorExtensions.x-lookup}}{{.}}{{/vendorExtensions.x-lookup}}{{^vendorExtensions.x-lookup}}{{name}}{{/vendorExtensions.x-lookup}}"]; ok {
```
sebastien-rosset: It would be good to document this vendor extension somewhere. Currently the vendor extensions are listed at https://github.com/OpenAPITools/openapi-generator/wiki/Vendor-Extensions, but it does not look like only owners can make a change (?). Here the vendor extension is buried in the template, so I think very few people would know this aliasing capability exists.
Relevant issue:
#4976

jirikuncar: @sebastien-rosset I agree we should document it. Personally I would move the wiki to `/docs` so it's editable by contributors.

wing328: I saw the switch from `{{keyParamName}}` to `{{name}}`. Is `keyParamName` casuing issue?

wing328: Another minor feedback is that when I saw `{{#vendorExtensions.x-lookup}}`, I really have no idea what `lookup` means in this case.

What about `{{#vendorExtensions.x-auth-id-alias}}` instead?

jirikuncar: @wing328 other languages are using `{{keyParamName}}` and - **yes** - `{{name}}` can cause problem if query param and header have the same name.

jirikuncar: I don't have any strong opinion about the name. At first I thought about it as key aliasing, but during the implementation I have changed my mind can call it `x-lookup` since it's more about from where the value is looked up.

jimschubert: I like William's suggestion of `x-auth-id-alias`.

Also, it sounds like `{{name}}` should be changed back to `{{keyParamName}}`? Maybe it's usage in `{{^vendorExtensions.x-lookup}}{{name}}{{/vendorExtensions.x-lookup}}` was unintentional?

jirikuncar: - I have changed the `x-lookup` to `x-auth-id-alias`.
- using `keyParamName` is wrong since there can be a collision when using API key in header and query argument under same name: `curl -H 'my-key: header-key' https://example.com/endpoint` vs. `curl https://example.com/endpoint?my-key=query-key`

jirikuncar:
@jimschubert @wing328 I have renamed the `x-lookup` extension also in #4966. Is there a place where we can document it?

jimschubert:
@wing328 did you have any further comments about this PR or #4966?

wing328:
Nope. The renamed extension looks good to me.

Issue #1811, opened 2019-01-04
[REQ] Feature Request Description


jimschubert:
### Is your feature request related to a problem? Please describe.

Yes. The barrier to entry for authoring templates or custom generators should be minimized.

## Describe the solution you'd like

config-help should provide some options for displaying more than CLI-only generation flags.

Examples:

* `--authoring` => could provide the embeddedTemplateDir as this isn't always the same as the generator name
* `--extensions` => could provide a list of vendor extensions for the generator

## Describe alternatives you've considered

We could create a new CLI command, but that seems to add little value over one or more additional switches.

## Additional context

I think any additional information should be hidden by default. That is, authoring and extension information shouldn't display when `config-help` is called for a generator without the additional switches. This is because some of our generators have a ton of options, and there's no need to crowd the terminal output.

The generated output could eventually be used to generate parts of our documentation (see #1770).

jimschubert:
Vendor Extensions, Supporting Files may make more sense as a `--dry-run` option for the `generate` command or something similar.  Features should be doable once #3614 is merged, and I envision either outputting a table to the generator doc page or creating an overall matrix like kangax's ES6 compat page (https://kangax.github.io/compat-table/es6/) or jwt.io' library option support display (https://jwt.io/).

I've output _most_ of the details for modifying a generator via config help changes in #4941. These options (primitives, import/instantiation mapping) make sense as part of config-help because they are CLI options which may need to be modified to generate valid code, especially for specs users don't control. 

Authoring details may make more sense as its own command. Authoring is a more advanced option which might require viewing or dumping template files for a given generator+library combination, with instructions for modifying the template. This could get hairy because some generators switch on options provided to the `generate` command to determine which files are added to the supporting files array.

I wonder if it may make more sense to create "subcommand" for the `generate` command.  In airline, there aren't really subcommands, so it would require that we change `generate` to a group, with a default command of `run`. This would allow us to put the additional functionality here under commands like:

* `simulate` or `dry-run`: acts as a dry-run, and displays a list of files which would have been written, files which would have been excluded (and why), or other details
* `info`: can display a list of info on supporting files, vendor extensions, additional properties, or anything else which is heavily reliant on `processOpts`. I don't know how airline shares args between commands within a group, but I think an info subcommand would have the same requirements as we have on `generate` now.
[...]

PR #4941, opened 2020-01-07
[doc] full generator details

jimschubert:
Looking for feedback on adding this to docs. It's a lot of additional text.

`ul` elements will be single-column in github, but displayed as expected in our doc site.

image

### PR checklist


jimschubert:

spacether:
@jimschubert could you also add the additional-properties options that are available to each generator and include an sample that shows how to use to, or a link to that?

jimschubert:
@spacether additional properties aren't evaluated until processOpts (the `generate` command). This would need to be done in some `--dry-run` option in the generate command (along with supporting files).

jimschubert:
I added more details to #1811 about the options we can only get from the `generate` command.

spacether:
@jimschubert are you sure?
When I look here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java#L145
I see many cliOptions that are available in the PythonClientCodegen method.

jimschubert:
@spacether what you've linked to is CLI options which were already documented before this PR (the `CONFIG OPTIONS` header). See: https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/python.md

Additional Properties are different. It's a map of options which holds values _usually_ including CLI options, plus others that can only be determined after the generate command and subsequently `processOpts`.
Compare the package name CLI option just above the line you've linked and this one in `processOpts`:  https://github.com/OpenAPITools/openapi-generator/blob/a2532cc3c5f665124c9ad36feca962b55c5d3d64/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java#L184

For better examples of why it would only be helpful to dump these from generate, see the following generators which set additionalProperties and/or supportingFiles based on user inputs:

* [Java client](https://github.com/OpenAPITools/openapi-generator/blob/a2532cc3c5f665124c9ad36feca962b55c5d3d64/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java#L227)
* [Aspnetcore](https://github.com/OpenAPITools/openapi-generator/blob/a2532cc3c5f665124c9ad36feca962b55c5d3d64/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java#L284)
* [Csharp](https://github.com/OpenAPITools/openapi-generator/blob/a2532cc3c5f665124c9ad36feca962b55c5d3d64/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CSharpNetCoreClientCodegen.java#L467)

Also, another example of how vendor extensions (those we apply for our tooling) will only be known after the generate command:

* [Finch Server] (https://github.com/OpenAPITools/openapi-generator/blob/a2532cc3c5f665124c9ad36feca962b55c5d3d64/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaFinchServerCodegen.java#L397)

There may be opportunities for CLI options, although we do already output defaults and the CLI option type itself outputs available enum information. I would think any confusion here might be generator specific, like if a generator has a set of values but doesn't define them as an enum but rather a freeform string.

I apologise for any confusion, but I didn't realize folks considered CLI options to also be additional properties. I think it's only coincidental that they usually end up there.

spacether:
Thank you for explaining that. I didn't understand how all of those were related.

spacether:
Also people don't necessarily know about this link that you pointed us to.
https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/python.md
The new doc page that you are generating also looks like it would benefit from including this same info ir linking to the correct md file

jimschubert:
Forgot to mention, we do already document how to pass multiple additional properties.. see https://openapi-generator.tech/docs/usage#additional-properties

I don't know if it makes sense to double-document this, or to add a link to this on the config page. These are generator specific details, while how to define additional properties is tool specific. That is, it's different for CLI, Maven plugin, Gradle Plugin, Online, and the Bazel Plugin. I think we can safely assume that people can find those docs easily via the doc site. If people are navigating docs in GitHub, `usage.md` is in the parent directory from the auto-generated docs.

If we cross-link, I'm pretty sure it won't work in docusaurus unless it's a full http link, and that will introduce weird navigation for those using GitHub.

jimschubert:
@spacether

> Also people don't necessarily know about this link that you pointed us to.

I'm not sure how to solve for that. The project follows a pretty standard pattern of putting documentation under the `docs` folder, and a single `generators` folder there (plus other docs at the root). This generators folder has a README which lists the generators for GitHub navigation.

Do you mean people using CLI don't know to look here? This is why I've added the same info as above to the `config-help` command earlier this week. It's behind a `--full-details` flag for now, so it won't break anyone who might be parsing the output (for whatever reason).

> The new doc page that you are generating also looks like it would benefit from including this same info ir linking to the correct md file

I don't understand the comment. I'll assume you mean the doc page in the screenshot should link to this markdown page? Our doc site, which is linked in the README and GitHub repo description, uses docusaurus to build the site from the markdown under `docs`. It rebuilds the site on every push to master, and what you see in the `python.md` file I've linked gets much prettier after build: https://openapi-generator.tech/docs/generators/python

It's this site generation which is the ultimate reason for those markdown files to exist. However, for docusaurus to build it's graph of cross-links, the markdown files have to be registered in a specific way. These generated markdown files can't be registered in the expected way because docusaurus isn't great at files in nested directories. So, to link from these back to other doc pages, it would have to be the full doc site URL rather than a relative link (which works in all other markdown pages since they're not nested in directories). That's the background for my above comment about making for weird navigation in GitHub.

I hope that clarifies (and was on point for your comment). As always, I'm open for suggestions to improve things.

spacether:
Thank you for clarifying and explaining this.
I didn't know what our md files are being used to autogenerate files at https://openapi-generator.tech/docs/
Understanding this now, I see that Python.md includes config options, and that we are updating that same generator markdown file.
Mistakenly I thought that we we editing other files because the screenshot looked so different from what I knew and the description didn't say we are updating each generator's markdown file, which in turn autogenerates files at https://openapi-generator.tech/docs/generators/python
This includes the data that I want our users to have. On a generator page I want them to be able to answer questions like:
- What reserved words are in this generator
- What CLI options can I use with this generator
- [FUTURE] what Swagger/Openapi features are supported in this generator

Thank you for your patience and thorough explanation of the changes that you made and how our repo and the docs site works

PR #4946, opened 2020-01-08
[fix] [python-flask-python2] [python-flask] Datetime is not handled well by Python

shybovycha:
Python can't really deepcopy objects well https://stackoverflow.com/a/32593735/330471
The API definition contained the datetime and it was parsed by Python' YAML lib as an object,
which caused the deepcopy to go nuts.

Making that timestamp a string, however, solves the issue.

### PR checklist


shybovycha:
cc @slash-arun and @spacether to review

samples/server/petstore/python-flask-python2/openapi_server/openapi/openapi.yaml
```diff
@@ -602,7 +602,7 @@ components:
         petId: 6
         quantity: 1
         id: 0
-        shipDate: 2000-01-23T04:56:07.000+00:00
```
spacether: Should this be loaded as a string -> datetime whether or not the yaml has quotes?
Can you add a test showing how this behavior has changed with your update in python code?

shybovycha: I might. I just don't think this test will have much value - it would simply fail with an exception at setup.

Since the way the API definitions are loaded is described in _flask_, this would be essentially testing somewhere in between flask's test utils and python's `copy.deepcopy` (which doesn't handle objects well). So in my opinion, this has little to do with the **openapi-generator**.

shybovycha: Most importantly, currently `master` is failing because of this ;)

shybovycha: It turned to be a good idea, after all - python2 was affected, whilst python3 was not. Added tests to highlight that.

samples/server/petstore/python-flask-python2/openapi_server/test/test_spec_parsing.py
```diff
@@ -0,0 +1,15 @@
...
+from openapi_server.encoder import JSONEncoder
+from unittest import TestCase
+
+class TestSpecParsing(TestCase):
+    def test(self):
+        with self.assertRaises(TypeError):
+            logging.getLogger('connexion.operation').setLevel('ERROR')
+            app = connexion.App(__name__, specification_dir='../openapi/')
+            app.app.json_encoder = JSONEncoder
+            app.add_api('broken_openapi.yaml', pythonic_params=True)
```
spacether: Can you trim that spec down to just the part that breaks functionality?
Can you add a comment in the yaml at that location?
Which code is failing to handle the spec or the datetime correctly?
Is it us or connexion?
- If it's connection it may make more sense to open an issue with that repo

samples/server/petstore/python-flask/openapi_server/test/test_spec_parsing.py
```diff
@@ -0,0 +1,14 @@
...
+from openapi_server.encoder import JSONEncoder
+from unittest import TestCase
+
+class TestSpecParsing(TestCase):
+    def test(self):
+        logging.getLogger('connexion.operation').setLevel('ERROR')
+        app = connexion.App(__name__, specification_dir='../openapi/')
+        app.app.json_encoder = JSONEncoder
+        app.add_api('broken_openapi.yaml', pythonic_params=True)
+        self.assertTrue(app.app is not None)
```
spacether: Can you also assert that the datatype of the parameter that you tweaked is correct?

spacether:
Closing due to inactivity

PR #4959, opened 2020-01-10
[Go] Support HTTP signature

sebastien-rosset:
This is adding a new "HTTP signature" security schemes (https://datatracker.ietf.org/doc/draft-cavage-http-signatures/). HTTP signatures is still a IETF draft, and hopefully it will become an official RFC this year. On one hand one may argue it shouldn’t be added because it’s still a draft, but on the other hand it is already being used by multiple products, so there may be benefits to support it. It is possible multiple organizations are independently adding the same security scheme.
 
Ideally one way to address the problem would be to make it possible to add new security schemes (such as HTTP signature) without being required to fork the OpenAPITools repo. But given the current code structure, it’s not clear to me how this could be achieved. Adding new security schemes involves surgery in multiple locations (Java codegen, templates, mustache tags).

I raised this point in Slack before opening the PR.

This PR depends on #4993 and #4934

### PR checklist


sebastien-rosset:
There is a dependency on golang 1.10.
What is the version of golang being used to build the "go" and "go-experimental" artifacts? It looks like it is a version < 1.10. Where is this specified? What is the rule for upgrading the go version? For example, how do you decide whether it's ok to upgrade to 1.13.6? I submitted a PR that uses strings.Builder, but that is defined in 1.10 and above. Should I write code to do the equivalent of strings.Builder? Or submit another PR to upgrade the version of golang?

sebastien-rosset:
@spacether, do you know what is the expected type of "header_params" in call_api?

```
        :param header_params: Header parameters to be
            placed in the request header.
```

It looks like it is a dict of string to string, but I'm not 100% certain. The reason I ask is because technically, an HTTP request may contain the same header multiple times. This is something I need to take into account when signing HTTP requests.

spacether:
Dunno. I would check out our tests to see.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3638,6 +3639,13 @@ public boolean isDataTypeFile(String dataType) {
                 } else if ("bearer".equals(securityScheme.getScheme())) {
                     cs.isBasicBearer = true;
                     cs.bearerFormat = securityScheme.getBearerFormat();
+                } else if ("signature".equals(securityScheme.getScheme())) {
+                    // HTTP signature as defined in https://datatracker.ietf.org/doc/draft-cavage-http-signatures/
+                    // The registry of security schemes is maintained by IANA.
+                    // https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+                    // As of January 2020, the "signature" scheme has not been registered with IANA yet.
+                    // This scheme may have to be changed when it is officially registered with IANA.
```
wing328: What about showing the comment as warning using LOGGER.warn?

I think we need to make it clear to users this is not official and may be subject to change in the future.

sebastien-rosset: ok, sure, doing.

sebastien-rosset: I am adding the following warning:
```
LOGGER.warn("Security scheme 'HTTP signature' is a draft IETF RFC and subject to change.");
```

jimschubert:
I kicked off CircleCI again, since it failed but at quick glance it seemed unrelated to work in this PR.

sebastien-rosset:
I just learned some people save their private key with file extensions other than ".pem", so I am going to relax the validation of the private key suffix.

sebastien-rosset:

A similar implementation has been done for Python and was merged to master.

jimschubert:
I've merged the linked pre-req PRs. This should be good to merge after master builds pass.

PR #4988, opened 2020-01-13
[Python] Conditionally set auth attributes

sebastien-rosset:
Fix for #3844:

Do not add auth parameter in HTTP request if auth parameter is set None
Add attribute to control which auth schemes can be used.
Currently, the python client obtains the list of security schemes that have been specified in the OAS spec, and automatically adds the authentication attributes in the header, cookie or query. This happens even if the parameters are unset. For example, the "Authorization" header may be set with an empty Bearer token:

'Authorization': 'Bearer '
With this PR, the bearer access token is no longer set if its value is None.

Tests performed:

Use OAS spec that specifies more than one security scheme for a given operation, e.g. OAuth2 and cookie
Generate the Python SDK using the master branch
Write a Python client that sets the cookie but not the OAuth2 access token
Invoke the client and verifies both auth attributes have been set (causing auth failure)
Generate the Python SDK using code from this PR
Invoke the client again, verify the auth parameters are set conditionally.

### PR checklist


sebastien-rosset:

spacether:
@sebastien-rosset how is this pull request different than your old one?

sebastien-rosset:
> @sebastien-rosset how is this pull request different than your old one?

Based on feedback from @wing328, I have made the following changes to avoid any breaking change for existing users:
* Retain the space after the api key prefix.
* Retain the "api_key" and "api_key_prefix" variables
* Don't add the OAuth2 access token as a input argument to the Configuration constructor, because it would be better to have a TokenSource. This could be done in a separate PR.

sebastien-rosset:
There seems to be an unrelated build issue:

```
Loading composer repositories with package information
Failed to clone the git@github.com:ybelenko/slim-token-authentication.git repository, try running in interactive mode so that you can enter your GitHub credentials

**Comment by spacether** [comment]
@wing328 why should we keep the api_key_prefix data in a separate dict when that data is storing info about an api_key? If we store it there it would be a breaking change with fallback and it reduces the number or dicts that we use from two to one. That way questions like:
- what value should I use for an api_key
- what prefix is used when sending this key
Can all be answered in the one api_key dict

**Comment by sebastien-rosset** [comment]
> @wing328 why should we keep the api_key_prefix data in a separate dict when that data is storing info about an api_key? If we store it there it would be a breaking change with fallback and it reduces the number or dicts that we use from two to one. That way questions like:
> 
> * what value should I use for an api_key
> * what prefix is used when sending this key
>   Can all be answered in the one api_key dict

My understanding is that we should split in two separate PRs: one that can go in master without breaking the backward compatibility (because master will be used as the basis for the next minor revision), and another PR that should go into the next release that can potentially break backward compatibility.

**Comment by spacether** [comment]
> > @wing328 why should we keep the api_key_prefix data in a separate dict when that data is storing info about an api_key? If we store it there it would be a breaking change with fallback and it reduces the number or dicts that we use from two to one. That way questions like:
> > 
> > * what value should I use for an api_key
> > * what prefix is used when sending this key
> >   Can all be answered in the one api_key dict
> 
> My understanding is that we should split in two separate PRs: one that can go in master without breaking the backward compatibility (because master will be used as the basis for the next minor revision), and another PR that should go into the next release that can potentially break backward compatibility.

Ah understood. Thanks for your explanation.

**Comment by sebastien-rosset** [comment]
Unrelated network error during build

**Comment by sebastien-rosset** [comment]
Build failure due to unrelated transient network issue.

**Comment by spacether** [comment]
This looks good. Thanks for this PR @sebastien-rosset !

File: modules/openapi-generator/src/main/resources/python/api_client.mustache [hunk 1]
```diff
@@ -520,9 +520,7 @@ class ApiClient(object):
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
-                if not auth_setting['value']:
```
martyzz1: This change breaks my client using openapi-generator-cli-5.0.0-20200127.021040-34.jar

We have api endpoints which allow Any Auth method. This function needs an overhaul, because it simply picks the last configured setting in the array, instead of the one which has actually been specified in the Configuration instance...
Preserving this line makes the code do the right thing...

    if not auth_setting['value']:
                    continue

martyzz1: Traceback (most recent call last):
  File "/usr/local/lib/python3.6/unittest/mock.py", line 1183, in patched
    return func(*args, **keywargs)
  File "/home/circleci/app/snap_usertracking/tests/test_client_api.py", line 665, in test_post_apikey
    response = api_instance.user_event(correct_post_data)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/api/tracking_api.py", line 287, in user_event
    **kwargs)  # noqa: E501
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/api/tracking_api.py", line 402, in user_event_with_http_info
    collection_formats=collection_formats)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/api_client.py", line 378, in call_api
    _request_timeout, _host)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/api_client.py", line 190, in __call_api
    _request_timeout=_request_timeout)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/api_client.py", line 422, in request
    body=body)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/rest.py", line 313, in POST
    body=body)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/snap_usertracking_api/rest.py", line 170, in request
    headers=headers)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/urllib3/request.py", line 80, in request
    method, url, fields=fields, headers=headers, **urlopen_kw
  File "/home/circleci/app/venv/lib/python3.6/site-packages/urllib3/request.py", line 171, in request_encode_body
    return self.urlopen(method, url, **extra_kw)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/urllib3/poolmanager.py", line 330, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/home/circleci/app/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/home/circleci/app/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 387, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/local/lib/python3.6/http/client.py", line 1262, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/local/lib/python3.6/http/client.py", line 1303, in _send_request
    self.putheader(hdr, value)
  File "/usr/local/lib/python3.6/http/client.py", line 1239, in putheader
    if _is_illegal_header_value(values[i]):
TypeError: expected string or bytes-like object

sebastien-rosset: > We have api endpoints which allow Any Auth method. 

Can you clarify a bit what you mean by "Any" auth?

martyzz1: ahh, I have a custom configuration.mustache but based on an older version.... I'll try updating it and see if that works...

sebastien-rosset: > ahh, I have a custom configuration.mustache but based on an older version.... I'll try updating it and see if that works...

I see. Let me know how it goes.

spacether:
Did we removed a feature where
auth_settings['value'] stored a specific auth data store?

sebastien-rosset:
> Did we removed a feature where
> auth_settings['value'] stored a specific auth data store?

No, I don't think so. I think the problem is happening because @martyzz1  has a configuration.mustache template based on the old version. In the new configuration.mustache template, the code makes sure to skip auth_settings items if the corresponding auth parameters are not set in the Configuration class.

sebastien-rosset:
> Did we removed a feature where
> auth_settings['value'] stored a specific auth data store?

Just to make sure, are you referring to @martyzz1's issue, or something else?

spacether:
> > Did we removed a feature where
> > auth_settings['value'] stored a specific auth data store?
> 
> Just to make sure, are you referring to @martyzz1's issue, or something else?

I was referring to this PR but upon closer reading of it, I don't think that we removed a feature here. We're okay.

martyzz1:
From @sebastien-rosset 's detailed response - and knowing the changes I made to my configuration.mustache - I'd agree that no features are likely to have been removed...

sebastien-rosset:
> From @sebastien-rosset 's detailed response - and knowing the changes I made to my configuration.mustache - I'd agree that no features are likely to have been removed...

Thank you for your response. I will mark the issue resolved for now, feel free to reopen if needed.

Issue #590, opened 2018-07-18
[OAS 3.0.0] yaml with multiple server urls generates code using first url only  


hippoKnight:
##### Description
I wrote a yaml following OAS 3.0, its servers part looked like this:
```yaml
...
#--- server ---#
servers:
  # server object
  - url: http://1.1.1.1:9200/
  - url: http://1.1.1.1:9201/v2
...
```
I used this yaml generating python code, and didn' t see the second url('http://1.1.1.1:9201/v2') 

So, my question is: What is the desired result if the yaml contains multiple urls?

##### openapi-generator version

I installed openapi-generator using Homebrew yesterday.

##### OpenAPI declaration file content or url

```yaml

#--- OpenAPI version ---#
openapi: '3.0.0'

#--- metadata about the api ---#
info:
  # title of the application
  title: '9'

  # description of the application
  description: '9'

  # terms of service for the API, must be a URL
  termsOfService: ''

  # contact information for the exposed API
  contact:
    name: 'myname'
    url: 'myurl'
    email: 'my@mail.com'

  # license information for the exposed API
  license:
    name: 'myname'
    url: 'myurl'

  # version of the OpenAPI document
  version: '1.0.1'

#--- additional external documentation ---#
externalDocs:
  description: '9'
  url: ''

#--- server ---#
servers:
  # server object
  - url: http://1.1.1.1:9200/
  - url: http://1.1.1.1:9201/v2
    # URL to the target host

    # map between a variable name and its value(used for substitution in the server's URL template)
    variables:
      basePath:
        enum:
          - ''
          - 'website/face/v2/'
          - 'business/api/'
        default: ''
        description: 'basePath'

      username:
        default: 'admin'
        description: 'username'

    description: 's'

#--- tags for specification with additional metadata ---#
tags:
  - name: 'website'
    description: 'API of website operations'

security:
  # only one of the security requirement objects need to be satisfied to authorize a request
  # mentioned names must in components-seuritySchemes
  -
    cookieAuth_sessionId: []
    cookieAuth_ytClusterId: []
    # lists the required security schemes to execute this operation
    # objects that contain multiple schemes require all satisfication

#--- reusable objects for the specification ---#
components:


  # Security scheme definitions (see Authentication)
  securitySchemes:
    # https://swagger.io/specification/#securitySchemeObject
    cookieAuth_sessionId:
      type: 'apiKey'
      in: 'cookie'
      name: 'session_id'

[...]

PR #4998, opened 2020-01-14
[java] Support templated servers

jirikuncar:
Adds support for multiple templated servers from OpenAPI specification.

See related PR for Go: https://github.com/OpenAPITools/openapi-generator/pull/4635
Addresses #590

### PR checklist

---
cc @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04) @lwlee2608 (2019/10


modules/openapi-generator/src/main/resources/Java/ApiClient.mustache
```diff
@@ -626,8 +684,15 @@ public class ApiClient {
    * @return The full URL
    */
   private String buildUrl(String path, List<Pair> queryParams, List<Pair> collectionQueryParams) {
+    String baseURL;
+    if (serverIndex != null) {
+      baseURL = servers[serverIndex].URL(serverVariables);
```
bkabrda: We should probably check that `serverIndex` is within the length of `servers`, as it's a user-set value and can be wrong.

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
```diff
@@ -684,12 +785,24 @@ public class ApiClient {
    * @return The response body in type of string
    * @throws ApiException API exception
    */
-  public <T> ApiResponse<T> invokeAPI(String path, String method, List<Pair> queryParams, Object body, Map<String, String> headerParams, Map<String, String> cookieParams, Map<String, Object> formParams, String accept, String contentType, String[] authNames, GenericType<T> returnType) throws ApiException {
+  public <T> ApiResponse<T> invokeAPI(String operation, String path, String method, List<Pair> queryParams, Object body, Map<String, String> headerParams, Map<String, String> cookieParams, Map<String, Object> formParams, String accept, String contentType, String[] authNames, GenericType<T> returnType) throws ApiException {
```
bkabrda: Maybe we could preserve backwards compatibility here by also keeping the method with its original signature, which would call this new method while passing `operation=null`?

modules/openapi-generator/src/main/resources/Java/ApiClient.mustache
jirikuncar: It will raise anyway.

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
jirikuncar: I am adding:
```diff
diff --git a/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache b/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
index ee4d7a6d10..bef8929e66 100644
--- a/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
@@ -906,6 +906,14 @@ public class ApiClient {
     }
   }
 
+  /**
+   * @deprecated Add qualified name of the operation as a first parameter.
+   */
+  @Deprecated
+  public <T> ApiResponse<T> invokeAPI(String path, String method, List<Pair> queryParams, Object body, Map<String, String> headerParams, Map<String, String> cookieParams, Map<String, Object> formParams, String accept, String contentType, String[] authNames, GenericType<T> returnType) throws ApiException {
+    return invokeAPI(null, path, method, queryParams, body, headerParams, cookieParams, formParams, accept, contentType, authNames, returnType);
+  }
+
   /**
    * Build the Client used to make HTTP requests.
    * @param debugging Debug setting
```

modules/openapi-generator/src/main/resources/Java/ApiClient.mustache
wing328: I agreed with @bkabrda to have a better error/exception message similar to what we've done in other generators, e.g. Python
```
        except IndexError:
            raise ValueError(
                "Invalid index {0} when selecting the host settings. "
                "Must be less than {1}".format(index, len(servers)))
```


https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/petstore_api/configuration.py#L316

jirikuncar: I have added the range check.

modules/openapi-generator/src/main/resources/Java/ServerConfiguration.mustache
```diff
@@ -0,0 +1,60 @@
+package {{invokerPackage}};
+
+import java.util.Map;
+
+import com.datadog.api.v1.client.ServerVariable;
```
jirikuncar:
```suggestion
import {{invokerPackage}}.ServerVariable;
```

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
```diff
@@ -684,12 +783,36 @@ public class ApiClient {
...
     updateParamsForAuth(authNames, queryParams, headerParams, cookieParams);
 
-    // Not using `.target(this.basePath).path(path)` below,
+    // Not using `.target(targetURL).path(path)` below,
     // to support (constant) query string in `path`, e.g. "/posts?draft=1"
-    WebTarget target = httpClient.target(this.basePath + path);
+    String targetURL;
+    if (serverIndex != null) {
```
jimschubert: Notice here to consumers that if new server variable functionality causes issue, one may call `setServerIndex(null)` to revert to previous behavior. Current default behavior with the server index 0 behaves in the same way as before, but moves the URL configuration to the servers array.

jimschubert:
This is great and adds functionality that most clients don't support. The samples are out of date which caused the build failures. I'll fix that and let the builds run, but then I'd like to merge this so people have some time to evaluate on SNAPSHOT before the 4.3.0 release in a couple weeks.

ChristianCiach:
Please see https://github.com/OpenAPITools/openapi-generator/commit/a09271e76d088ff2f32c8d5a89e881314d743147#r40779493

PR #5006, opened 2020-01-15
[go-experimental] Preserve capitalized acronyms in structure names

bkabrda:
Currently an openapi model called e.g. `EntityURL` would have a structure generated with name `EntityUrl`. I think this is both against the intention of the spec author and it is also against [Go naming recommendations](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). This PR fixes that to preserve the capitalized acronyms.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java
@@ -235,6 +235,10 @@ public abstract class AbstractGoCodegen extends DefaultCodegen implements Codege
     }
 
     public String toModel(String name) {
+        return toModel(name, true);
+    }
+
+    public String toModel(String name, boolean doUnderscore) {
         if (!StringUtils.isEmpty(modelNamePrefix)) {
             name = modelNamePrefix + "_" + name;
         }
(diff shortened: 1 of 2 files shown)


wing328:
I think https://github.com/OpenAPITools/openapi-generator/pull/4900 aims to fix similar issue.

bkabrda:
@wing328 yeah, it's similar. What's different is that my PR doesn't touch the existing stable Go client (because this change would break backwards compatibility) and I think my PR is much simpler and it addresses the root cause of this issue (the solution in #4900 feels more like a workaround).

Do note that this issue isn't present in e.g. Java client generator (or other generators that I've seen), so I think it doesn't need a solution on the level of `DefaultCodegen`. I'm not saying that the solution outlined in #4900 is wrong, I just think it might have a slightly different usecase.

jimschubert:
I agree that #4900 seems more like a workaround. It also seems like it will break literally every client with something that _looks like_ an initialism or acronym when the user may not have intended this. For projects with static analysis rules, it may trigger warnings about identifiers not matching casing rules. For example, if you take any of the Java/jvm clients and push them to sonarcloud, you'll get a ton of warnings.

I much prefer the approach in this PR, since rules around language supported capitalization should be left to the language/framework and not forced globally.

PR #5015, opened 2020-01-16
[Core, Rust Server] Support multipart/related requests

richardwhiuk:
This adds support for multipart/related requests.

This includes changes to the core of OpenAPI Generator to add a new parameter type, `relatedParams`.

I'm putting this against 5.0.x, because the Rust code has been developed against the Hyper 0.12 version of the Rust Server generator.

Contribution by @JamesTurner, myself and others from @Metaswitch.

### Rust Server Technical Committee

@frol @farcaller @bjgill @richardwhiuk

### PR checklist


richardwhiuk:
I don't think either of the failures in CI are related to this code:
- https://travis-ci.org/OpenAPITools/openapi-generator/builds/637920927?utm_source=github_status&utm_medium=notification looks like a failure to pull a PHP library from GitHub 
- https://app.bitrise.io/build/a79ca15bce20e884#?tab=log is a failure in the Swift tests

richardwhiuk:
@OpenAPITools/generator-core-team anyone fancy reviewing the core changes here?

jimschubert:
I've reviewed the code in DefaultCodegen and it looks solid. I have one comment and one request.

I'm a little concerned about adding support for mime types not defined in the specification, and I wonder if we might want to come up with another way to support custom content types in the 5.0 release. This isn't a blocker for this PR, just a comment. I think it's great that this exposes our need to construct a template object which can allow different behaviors for different mime types rather than just exposing the name.

For my request, could you add a `Custom` enum to https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-core/src/main/java/org/openapitools/codegen/meta/features/ParameterFeature.java#L76 and declare support for this in the constructor of the generator (`featureSet`)? Ideally, the enum should be annotated with OAS2 or OAS3 if you've pulled from one of those specs (which I don't see). You could also annotate with `ToolingExtension`. The `featureSet` items will be used to generate a feature matrix in our docs shortly.

jimschubert:
Another comment: CircleCI failed due to output changes in rust server after your branch merge. Can you double check that?

richardwhiuk:
@jimschubert 

`multipart/related` is intended to be supported by the OpenAPI specification - see https://github.com/OAI/OpenAPI-Specification/issues/520 , https://github.com/OAI/OpenAPI-Specification/issues/303, resulting in https://github.com/OAI/OpenAPI-Specification/pull/878/files which have added support for roughly arbitrary multipart types.

As such, I think a generator could/should support `multipart/form-data`, `multipart/related`, `multipart/mixed`, and `multipart/alternative` to be complete spec compliant. I can add enum options for all of those, mark Rust Server as supporting `multipart/related` and tag them as OAS3?

I'll fix up CI.

richardwhiuk:
Hmm, I wonder whether looking at these as multipart related parameters is actually the right fix.

Really, form parameters and related parameters are the same - it's just the operation takes a different content type. multipart/form parameters can have content types.

jimschubert:
@richardwhiuk So rather than `FormMultipart` documented as just OAS2 support, it should just be `Multipart` targeting both spec versions? Or are you considering that there are some missing points in DefaultCodegen which should account for these separately?

richardwhiuk:
Yes, I think so. I've pushed a change which is much less invasive in `DefaultCodegen.java` - see https://github.com/OpenAPITools/openapi-generator/pull/5015/commits/9c97fef7f96127d9da7cb82b7befde2c9b58b5f4

I'm also suspicious that we need to track which generators support what content types. It's not clear to me that `ParameterFeature` is the right mechanism to do that.

With this change, the Rust Server generator should support `multipart/form-data`, `multipart/related`, `application/json`, `application/octet-stream`, `text/plain` and `application/xml`.

In future, we might support `application/yaml`, `multipart/alternative`, `multipart/mixed`, etc.

jimschubert:
@richardwhiuk `ParameterFeature` is intended to track the data types defined explicitly in OAS2/3, as we do have some generators which currently don't support these fully (example: go-server doesn't handle int32 as reported by someone in Slack this week). Allowing generators to explicitly include things they don't support would make these types of misses clear to users.

jimschubert:
Looks good. I think it makes a lot of sense to just have contentType as you've added.

PR #5019, opened 2020-01-17
[Python] fix numeric enum in python flask, aiohttp

wing328:
- Plotted the fix from Python client generator
- Tested locally with numeric enum

cc @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @slash-arun (2019/11) @spacether (2019/11)

### PR checklist


modules/openapi-generator/src/main/resources/python-flask/model.mustache
```diff
@@ -116,6 +118,7 @@ class {{classname}}(Model):
 {{/isMapContainer}}
 {{/isContainer}}
 {{^isContainer}}
+        allowed_values = [{{#isNullable}}None,{{/isNullable}}{{#allowableValues}}{{#values}}{{#isString}}"{{/isString}}{{{this}}}{{#isString}}"{{/isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
```
spacether: We are tying to minimize the usage of flake8 E501 and other flake8 opt out lines.
Can you spread this to a multiline statement so we won't need  `# noqa: E501`?

modules/openapi-generator/src/main/resources/python-aiohttp/model.mustache
```diff
@@ -89,8 +89,8 @@ class {{classname}}(Model):
         :type {{name}}: {{dataType}}
         """
 {{#isEnum}}
-        allowed_values = [{{#allowableValues}}{{#values}}"{{{this}}}"{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]
 {{#isContainer}}
+        allowed_values = [{{#isNullable}}None,{{/isNullable}}{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
```
spacether: We are tying to minimize the usage of flake8 E501 and other flake8 opt out lines.
Can you spread this to a multiline statement so we won't need # noqa: E501?

modules/openapi-generator/src/main/resources/python-aiohttp/model.mustache
```diff
@@ -109,6 +109,7 @@ class {{classname}}(Model):
 {{/isMapContainer}}
 {{/isContainer}}
 {{^isContainer}}
+        allowed_values = [{{#isNullable}}None,{{/isNullable}}{{#allowableValues}}{{#values}}{{#isString}}"{{/isString}}{{{this}}}{{#isString}}"{{/isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
```
spacether: We are tying to minimize the usage of flake8 E501 and other flake8 opt out lines.
Can you spread this to a multiline statement so we won't need # noqa: E501?

modules/openapi-generator/src/main/resources/python-flask/model.mustache
```diff
@@ -96,8 +98,8 @@ class {{classname}}(Model):
         :type {{name}}: {{dataType}}
         """
 {{#isEnum}}
-        allowed_values = [{{#allowableValues}}{{#values}}"{{{this}}}"{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
 {{#isContainer}}
+        allowed_values = [{{#isNullable}}None,{{/isNullable}}{{#allowableValues}}{{#values}}{{#items.isString}}"{{/items.isString}}{{{this}}}{{#items.isString}}"{{/items.isString}}{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}]  # noqa: E501
```
spacether: We are tying to minimize the usage of flake8 E501 and other flake8 opt out lines.
Can you spread this to a multiline statement so we won't need # noqa: E501?

samples/server/petstore/python-aiohttp/openapi_server/openapi/openapi.yaml
```diff
@@ -32,7 +32,7 @@ paths:
         required: true
         x-body-name: body
       responses:
-        405:
+        "405":
```
spacether: These spec updates are unexpected in this pull review because we did not edit the spec file.
Can you please add the below sample generation files to https://github.com/OpenAPITools/openapi-generator/blob/master/bin/utils/ensure-up-to-date?
- python-aiohttp
- python-flask-python2
- python-flask

spacether:
> * Tested locally with numeric enum

Looking at our repo I don't see a single example of a spec with a numeric enum.
Could you add a minimal spec with a numeric enum and add JAVA test [here](modules/openapi-generator/src/test/java/org/openapitools/codegen/python) here of one of fromModel/fromOperation/fromProperty etc verifying that the generated enum is correct?
Adding this spec will also allow other generators to test for this condition.

modules/openapi-generator/src/main/resources/python-flask/model.mustache
wing328: Let's do this in a separate PR instead. I'll open an issue to track it to see if anyone has the time to improve it.

(this PR simply plots the fix from the python client template)

samples/server/petstore/python-aiohttp/openapi_server/openapi/openapi.yaml
wing328: Added to ensure-up-to-date.

These scripts are using petstore.yaml (original) but not the fake petstore spec (which has numeric enums)

modules/openapi-generator/src/main/resources/python-flask/model.mustache
spacether: Are you suggesting that we remove the E501s in a separate PR? For this one I prefer to omit the E501 for these lines. This update is adding them when they are not needed. Why add them when it throws a linting error which we can avoid?

samples/server/petstore/python-aiohttp/openapi_server/openapi/openapi.yaml
spacether: Can we add a model with numeric enums in petstore.yaml?
That lets us verify this fix in the repo and this PR.

wing328: We can update the script to use the fake petstore spec instead (petstore.yaml shouldn't be updated)

When a generator gets mature, we will use fake petstore spec instead of petstore.yaml/json as the generator should be able to handle those edge cases in fake petstore spec.

I've not sure if these generators can handle the edge cases in fake petstore spec.

modules/openapi-generator/src/main/resources/python-flask/model.mustache
wing328: Created https://github.com/OpenAPITools/openapi-generator/issues/5030 for tracking.

wing328:
> Can you add JAVA test here with that spec using one of fromModel/fromOperation/fromProperty etc verifying that the generated enum is correct?

If you look at the fix, it cannot be tested in the Java class because it uses `isString` in the mustache template to determine whether a double quote is required.

Later after switching these scripts from petstore.yaml to fake petstore petstore, we can add test(s) in the samples to ensure the output is correct.

PR #5033, opened 2020-01-18
[CORE] replace http with https. Replace http://central.maven.org with https:repo1.maven.org

sebastien-rosset:
Trying to eliminate the use of http as much as possible.
Some of these changes are documentation only and do not affect the build but we should use https nonetheless.
There are multiple use of http://central.maven.org, but that site returns error 403 with a message stating to use http://repo1.maven.org. I've replaced these.

There are some http URL that don't exist anymore. I'm not sure what to do about it, and it's not clear they are used anyway.

I did a cursory search to find use of http://:  
``` 
grep -r 'http://'  . | grep -v xmlns |grep -v schemaLocation   | wc
```
Right now it yields over 6000 results. Unfortunately we can't do a simple search and replace.

Slightly better way to find references to http://
```
grep -r 'http://'  . | grep -v xmlns |grep -v schemaLocation  | sed -rn 's/.*(http:\/\/[^ )"]+).*/\1/p' | sort | uniq
```
It yields 779 results.

### PR checklist


jimschubert:
I'm going to close this in favor of #5034. I copied the wring PR number in my LGTM comment. ❌ 

Additional issue is `mavenCentral()` for gradle prior to 4.8.1 doesn't default to https, so we have to force the HTPS URL by explicitly defining the repository. It passed on my above PR, but I missed regenerating some of the samples.

PR #5036, opened 2020-01-18
[core] Move overwrite ownership to SupportingFile

### LINKED ISSUE #4582: [BUG] DefaultCodegen#writeOptional should honor options such as skipOverwrite/codegen ignore
##### Description

Rules for skipping 50-100 files across a handful of generators is not intuitive.

all

##### Steps to reproduce

* Add any text to `samples/openapi3/client/petstore/ruby/spec/spec_helper.rb`
* Commit a WIP commit
* Run `openapi-generator batch --verbose bin/ci/openapi3-ruby-client-petstore.json`
* Notice no diff because the file is silently skipped due to `writeOptional`
* Run `./bin/openapi3/ruby-client-petstore.sh`
* Notice diff because file is deleted in the bash script via:

```bash
find samples/openapi3/client/petstore/ruby/spec -type d -not -name spec -not -name custom | xargs rm -Rf
find samples/openapi3/client/petstore/ruby/spec -type f -not -name petstore_helper.rb -not -iwholename '*/spec/custom/*' | xargs rm -Rf
```

##### Suggest a fix

Remove support for `writeOptional` in favor of newer methods for controlling writes. If outputs under `samples` require these to be unmodified, they should be added to respective `.openapi-generator-ignore` files.

jimschubert:
Previously, there was a writeOptional method in DefaultCodegen which
allowed *Codegen instances to immediately write out a supporting file if
it did not exist. This would allow a codegen implementation to skip
user-facing options such as definitions in .openapi-codegen-ignore, the
"supportingFiles" system property, and support for the experimental
handlebars templating engine. While our implementation only modified the
supportingFiles list conditionally, it added confusion as it seemed to
imply that file writes were somewhat the responsibility of
DefaultCodgen (it's DefaultGenerator which handles file manipulation).

This commit moves the definition of whether a file supports overwriting
existing files into the SupportingFile type itself, allowing that
functionality to be determined at time-of-write rather than
time-of-definition. This would allow us, for example, to dump the list
of files which would be generated using a --dry-run option or similar.

This will be a breaking change for anyone who has extended
DefaultCodegen and called "writeOptional". The path to migrate is to add
the SupportingFile to the supportingFiles list and chain the method call
`.doNotOverwrite()` on the instance.

This has the added benefit of clarifying this behavior, considering the
write behavior wasn't previously "optional" writes but optionally
defining the list of supportingFiles based on the state of the file
system.

Fixes #4582 

This is both a bug fix (due to previous behavior of potentially circumventing user options) and a breaking change (due to workflow options exposed to users via DefaultCodegen rather than DefaultGenerator).

### PR checklist


jimschubert:
I should clarify:

Our `writeOptional` implementation would previously only add the file to the `supportingFiles` list if the file didn't exist. Being a method in `DefaultCodegen`, there's nothing that would have prevented someone from writing immediately (like if someone said "Why isn't writeOptional not writing?").

There's technically nothing stopping users from circumventing DefaultGenerator's workflow and writing files whenever they want from a Codegen implementation, but this aims to clarify ownership of that behavior as before the `SupportingFile` would sometimes be in the list of supporting flies and sometimes would not be in the list.

jimschubert:
I've squashed and force-pushed, trying to add clarification to the commit comment and PR description about my concerns with the previous behavior.

Issue #33, opened 2018-05-13
Discuss static analysis options


jimschubert:
##### Description

I raised the question in #26 about how we could verify standard usage of slf4j loggers at compile time, and whether there is a static analysis tool we could use.  We could look into SonarQube, but I've also found a Findbugs plugin at https://github.com/KengoTODA/findbugs-slf4j which would do exactly this.

Should we consider incorporating this type of static analysis now? With so many open source contributors, it would be helpful to standardize on some coding patterns to reduce maintenance overhead.

##### openapi-generator version

all

##### OpenAPI declaration file content or url

n/a

##### Command line used for generation

n/a

##### Steps to reproduce

n/a

##### Related issues/PRs

* #26 

##### Suggest a fix/enhancement

Incorporate FindBugs or similar into the build.

jmini:
This is a great idea. Thank you a lot for opening this issue.

---

I thought Findbugs is no longer developed and was forked to Spotbugs.
https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2016-November/004321.html

---

One thing we need to consider is the integration of this with the build and/or with the PR workflow...
Do we want the build to fail if one of the rule is broken? This is hard, but it will bring us quality.

---

SonarQube: they have an online version and it is free for OpenSource project:
https://about.sonarcloud.io/sq/

I have heard they are working on an integration with the PR workflow, I am not sure if the feature was released yet or not.

wing328:
UPDATE: I've used sonarcloud.io to generate the [report](https://sonarcloud.io/project/issues?id=org.openapitools%3Aopenapi-generator-project&resolved=false&types=BUG)

Ref: https://github.com/OpenAPITools/openapi-generator/issues/788

tomasbjerre:
I'm missing this as well. I would propose doing this with PMD and Spotbugs, using maven-plugins.

 * Add maven-plugins to parent pom
   * https://spotbugs.readthedocs.io/en/latest/maven.html (and in the future with Gradle https://spotbugs.readthedocs.io/en/latest/gradle.html)
   * https://maven.apache.org/plugins/maven-pmd-plugin/ (and in the future with Gradle https://docs.gradle.org/current/userguide/pmd_plugin.html)
 * Configure plugins to not fail the build on errors.
[...]

PR #5040, opened 2020-01-19
Spotbugs, PMD and Checkstyle #33

tomasbjerre:
This is a proposed solution to #33

It introduces static code analysis with:

 * Checkstyle
 * PMD
 * Spotbugs

It relies on Maven plugins, that all have equivalent Gradle plugins.

It is automatically used in any CI tool as well as when working locally, as it is executed with `mvn verify`.

### PR checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ packages/
 .vagrant/
 .vscode/
 **/.vs
+bin
+.factorypath
 
 .settings
 
--- a/modules/openapi-generator-cli/pom.xml
+++ b/modules/openapi-generator-cli/pom.xml
@@ -28,6 +28,33 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <configLocation>${project.basedir}/../../google_checkstyle.xml</configLocation>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
+                <configuration>
+                    <excludeFilterFile>${project.basedir}/../../spotbugs-exclude.xml</excludeFilterFile>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-pmd-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>se.bjurr.violations</groupId>
+                <artifactId>violations-maven-plugin</artifactId>
+                <configuration>
+                    <!-- Should be decreased continuousle as issues are fixed, and will eventually 
+                        reach 0. -->
+                    <maxViolations>274</maxViolations>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <configuration>
                     <archive>
--- a/modules/openapi-generator-cli/pom.xml
+++ b/modules/openapi-generator-cli/pom.xml
@@ -30,14 +30,14 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <configuration>
-                    <configLocation>${project.basedir}/../../google_checkstyle.xml</configLocation>
+                    <configLocation>${project.parent.basedir}${file.separator}google_checkstyle.xml</configLocation>
                 </configuration>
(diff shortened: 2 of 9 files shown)


jimschubert:
This is awesome. One thing I'd comment is that we'll need to double check and make sure PMD and spotbugs artifacts don't get included into the CLI fat jar.

I see you're still committing and the test suite is running as I type. Are you ready for reviews?

tomasbjerre:
Glad you like it =)

I'm ready for review. I'm just trying to figure out why the build is failing, it works locally. But I'll give that a rest and await review.

We can probably find a bunch of rules that can be ignored. There are quite a few issues reported now.

jimschubert:
Bitrise failed for an unrelated reason. I have a fix in #5043 which will be merged once those tests succeed.

jimschubert:
I've done an initial pass-through, and have some questions/comments.

* Is there a benefit to using the relative path from project.basedir rather than the parent.basedir (configured at the top of each pom)? I made [these local changes](https://gist.github.com/jimschubert/488c6cca32b0c6f7b54df7db64ba9007) and it works as I'd expect, but I've never seen the relative usage for config and was wondering if I was missing something.
* I have my terminal buffer set to 30,000 lines. The tabular report I saw blew this away. I'm wondering how we might integrate such reports into our process. I've setup sonarcloud to evaluate, because I'm familiar with using Sonar… but it doesn't support Java for automatic analysis and would require that I install and configure sonar-scanner. Not a big deal, but will probably be a week or so before I have time to look into it further.
* I have a pretty beefy development machine (3.5 GHz dual core i7, 16GB 2133 MHz LPDDR3 RAM, Intel Iris Plus Graphics 650 1536 MB). These additions _double_ the project build time from 01:34 min to 03:17 min for full build/analysis/install. Builds on travis on this branch are taking 5-8 minutes longer than builds without this change (See screenshot, which doesn't include the lower end 25m 11s). I wonder if it might be good to start small, like with checkstyle or spotbugs only? * Could we run the "biggest bang" on PRs and run the full quality scan off pushes to master? Rather than hook this up into travis/circleci --both of which have had issues with pushing to external systems-- maybe we could explore GitHub Actions for pushes to master to run PMD+Checkstyle+SpotBugs and push that to Sonar?

image

I'd also be interested in what others from @OpenAPITools/generator-core-team and @OpenAPITools/openapi-generator-collaborators suggest.

tomasbjerre:
Made a new commit:

 * Is now using `parent.basedir`.
 * Reduced Spotbugs effort from `max` to `min`.

The large amount of issues can be taken care of by:

 * Setting `<detailLevel>VERBOSE</detailLevel>` to WARN or ERROR.
 * Adding more exclusions in the PMD/Spotbugs/Checkstyle configurations.
 * Actually working on correcting the issues =)

Hopefully lowering effort of `Spotbugs` reduces time to an acceptable level. If not, I would suggest adding a maven-profile that enables static code analysis. But if that can be avoided, if it can be in the default profile, I think the static code analysis will be of better use. My opinion is that issues from static code analysis should really be as severe as unit tests failing.

The current proposed solution here requires zero integration with any external service. I think you should think about what a big advantage that is and how much more complex things become when you add Sonar. The goal should really be to get these numbers down to 0 and that just shows there is not need for all the fancy graphs and statistics that Sonar produces.

jimschubert:
I agree the goal is to get those numbers to zero. But, I'm also looking at how maintainers may more easily inspect the results. Right now, it's a pain to view logs across numerous CI providers.

What would more concern me is build time. On my older MacBook Pro, the full project took 10 minutes where it only took 1.5 min after my upgrade. I'm thinking about how to get the quality without negative impact to community contributions.

I'll check the changes you've made later. I think 2 minutes added to builds would be our max increase in PRs.

tomasbjerre:
I found an issue in the Violations Magen Plugin, released that, and now Travis is failing because that version is not in Central yet... should be green any minute now.

My opinion is that the easiest way of working with static code analysis is to print it in the build log when doing `mvn verify`. Then I can easily use the analysis without having to wait for CI builds and I also don't need to setup any special tool, just execute Maven. But I definitely see why you are worried, this is just a proposed solution, feel free to close the PR if you don't want it =)

jimschubert:
To clarify, I'm not against the approach, just looking at how to avoid causing something like Circle CI from choking on display (the 30k lines comment), where it already struggles.

I agree the full quality profile on master is the way to go, because it's fine if master takes a few more minutes then each PR.

jimschubert:
For the core team, I've run a Gradle Enterprise scan on this branch and here's the hotspots

## mvn install

image

## mvn -T 4 install

image

## mvn -T 24 install

image

tomasbjerre:
Filtering out the `samples` would also reduce the time spend on analysis. Perhaps something to consider.

PR #5045, opened 2020-01-19
[gradle] consistent use of maven url in gradle files

sebastien-rosset:
According to https://docs.gradle.org/current/userguide/declaring_repositories.html, the syntax for maven repositories is:
```
repositories {
    maven {
        url "https://maven.springframework.org/release"
    }
}
```

There is no equal sign between the 'url' keyword and the value.

### PR checklist

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -51,7 +51,7 @@ To include in your project, add the following to `build.gradle`:
 buildscript {
   repositories {
     mavenLocal()
-    maven { url = "https://repo1.maven.org/maven2" }
+    maven { url "https://repo1.maven.org/maven2" }
   }
   dependencies {
     classpath "org.openapitools:openapi-generator-gradle-plugin:3.3.4"
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -51,7 +51,7 @@ To include in your project, add the following to `build.gradle`:
 buildscript {
   repositories {
     mavenLocal()
-    maven { url = uri("https://repo1.maven.org/maven2") }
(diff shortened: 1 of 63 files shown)


jimschubert:
I did some investigation into why both of these work (with and without the equals).

All of these appear to be the same in `build.gradle`:

```
maven { url "https://repo1.maven.org/maven2" }
maven { url("https://repo1.maven.org/maven2") }
maven { url = "https://repo1.maven.org/maven2" }
```

But, I'm not sure of what the underlying Groovy support is that enables this. See the below example where I have a property and a setter.

```
groovy:000> class Example {
groovy:001>     def url
groovy:002>     void setUrl(Object url) {
groovy:003>         this.url = url
groovy:004>     }
groovy:005>
groovy:005>     String toString() {
groovy:006>         return "Example: $url"
groovy:007>     }
groovy:008> }
===> true
groovy:000> example = new Example()
===> Example: null
groovy:000> example.url = "http://direct.example.com"
===> http://direct.example.com
groovy:000> example.setUrl("http://setter.example.com")
===> null
groovy:000> println example.toString()
Example: http://setter.example.com
===> null
groovy:000> example.url "http://sugared.example.com"
ERROR groovy.lang.MissingMethodException:
No signature of method: Example.url() is applicable for argument types: (String) values: [http://sugared.example.com]
Possible solutions: use([Ljava.lang.Object;), any(), dump(), grep(), getUrl(), use(java.lang.Class, groovy.lang.Closure)
```

It probably has something to do with the Closure. Anyway, best to stick with idiomatic Gradle and not use something which could potentially break in the future. Thanks for fixing it up!

Issue #1085, opened 2018-09-22
JAVA-SPRING model generation: disable swagger annotations import


dario-frongillo:
##### Description

Hi i m using java spring generator and i did not find an option to disable the import of 
import io.swagger.annotations.ApiModel;
import io.swagger.annotations.ApiModelProperty;
inside my model code generated.
Checking  the source code openapi-generator-3.2.3\modules\openapi-generator\src\main\resources\JavaSpring\model.mustache
i found that the ApiModel and ApiModelProperty are imported with mustache instructions

{{#imports}}import {{import}};
{{/imports}}

So i cant customize the mustache to disable it because {{import}} contains other useful import.

##### openapi-generator version

3.2.3

dario-frongillo:
Suggest a fix/enhancement : add a proprs to control swagger annotations

wwang2016:
Hi,

I had the same issue with generating model with @ApiModel and @ApiModelProperty

It is not needed in a model. In addition, it is generating the model with the open API v3 spec. That makes it worse.

I am trying to customize the template and have no luck as for now.

Wayne

mbayou:
Hi,

Do you have any news about this fix/enhancement?

Mathieu.

AndreasSi:
Hi all,

i have the same requirements. i will only generate interface an model classes with no dependencies pointing to io.swagger.

I try to fix this issue and will provide a fix/enhancement.

Kind regards,
Andreas

cen1:
This is not only a spring problem but all java in general. Each template wants to add it's own dependencies, you can't generate a clean native java model with any templates right now, at least not to my knowledge. It should be possible to disable any external dependencies with a config option.

ilmirons:
Any news on this? I'd need this feature too.

crtab:
Ran into the same issue - as I'm using openapi-generator to create a client to consume an existing external API, I don't need (or want) the swagger annotations as I won't be reproducing documentation already available from an external party.

hotzen:
we are generating from a proper OASv3, we dont need or want any swagger-v2 dependencies in our generated model - please make this configurable!

melissapalmer:
Having the same issue as this and would be great if this could be a configurable flag!

hotzen:
hey, is there any hope to configure this behaviour?

hotzen:
anything yet?

hotzen:
any news?

PR #5058, opened 2020-01-20
Adding useDocumentationAnnotations to JavaSpring

tomasbjerre:
Issue: #1085

Adding `useDocumentationAnnotations` to JavaSpring.

If the annotation is always added, users are forced to have Swagger on classpath during compilation. Something that we want to avoid.

Pretty much same code as in https://github.com/OpenAPITools/openapi-generator/pull/4270 but not mych activity there and we really need this. So this is a basically a rebase of AndreasSi to solve the merge conflict.

### PR checklist


tomasbjerre:
I made 2 commits, one with the feature and one with petstore update. I'm pretty sure I don't understand the petstore scripts yet, a lot of files were deleted when I run `./bin/spring-all-petstore.sh`.

modules/openapi-generator/src/main/resources/JavaSpring/api.mustache
```diff
@@ -7,7 +7,9 @@ package {{package}};
 
 {{#imports}}import {{import}};
 {{/imports}}
+{{#useDocumentationAnnotations}}
```
pixelshaded: so what is removing the io.swagger.annotations from the imports array? Otherwise it still gets added even if useDocumentationAnnotations is false. AND wouldn't it be better to handle this higher up, aka the JavaAbstractGenerator class so that anything using Java will remove the annotations from imports?

samples/server/petstore/spring-mvc-j8-async/src/main/java/org/openapitools/model/TypeHolderExample.java
```diff
@@ -109,7 +109,7 @@ public TypeHolderExample integerItem(Integer integerItem) {
    * Get integerItem
    * @return integerItem
   */
-  @ApiModelProperty(example = "-2", required = true, value = "")
+  @ApiModelProperty(required = true, value = "")
```
pixelshaded: these feel like regressions. Aka if it has an example, it should still have one after your change.

dragonjuggler:
I've been looking for how to do this via the maven or gradle plugin for the generator (since OpenAPI uses the YAML/JSON for documentation, the annotations are just clutter in the generated code).  Running with <configHelp>true</configHelp> does not show a 'useDocumentationAnnotations' option.  Is this implemented/available in the plugins?  I'm using version 5.0.0-beta2 of the openapi-generator-maven-plugin.

Thanks.

-Mark

PR #5061, opened 2020-01-21
[go] fix compilation error in generated go code

sebastien-rosset:
Currently, the generated go-experimental code has compilation errors and static analysis issues. This is because the block is repeated multiple times, depending on how many query parameters there are in the OAS.

It boils down to the fact that this line will be generated multiple times:

t:={{paramName}}
Hence the compiler will fail with the same variable being redefined multiple times.

I've opened a small, specific PR so hopefully it can be merged soon (compared to PR #4934 that has these changes plus other changes as well).

### PR checklist


wing328:
> Hence the compiler will fail with the same variable being redefined multiple times.

I think the best way is to define the variable once only if there is at least 1 query parameter defined.

Your fix is acceptable.

PR #5063, opened 2020-01-21
[test] Removes jmockit in favor of mockito

jimschubert:
We use mockito in many tests. This removes jmockit which is run as a javaagent in favor of Mockito which is not.

This work is in preparation for applying some static analysis tools, while evaluating others such as Jacoco. I'm also look at ways to improve build times while also decreasing "ramp up time" for contributions from the community. Reducing the number of mock frameworks and dependencies is a step toward that goal.

I'm opening this for discussion. It seems like jmockit was used heavily in `GenerateTest` in the CLI to test the internals of the generate task. I think this is the wrong approach; each component of that task should be tested in isolation.

I think this approach cleans up the options tests, as it was never really clear to me and most likely to others how to use the Expectations API in jmockit. I also found while testing that the Expectations of jmockit would silently swallow exceptions. For example, scala-http-client defines this method:

```
public void setModelPropertyNaming(String naming) {
    if ("original".equals(naming) || "camelCase".equals(naming) ||
            "PascalCase".equals(naming) || "snake_case".equals(naming)) {
        this.modelPropertyNaming = naming;
    } else {
        throw new IllegalArgumentException("Invalid model property naming '
                naming + "'. Must be 'original', 'camelCase', " +
                "'PascalCase' or 'snake_case'");
    }
}
```

The expectation in the test passed `modelPropertyNaming` as the option and verified that it was executed once. That execution should have resulted in an IllegalArgumentException on invocation, but jmockit's Expectation only cared that the method was called and hid that it was called with invalid data.

~I plan to look at replacing the `GenerateTest` tests with unit tests, and whether it would make sense to construct the Generate class to evaluate argument parsing.~ The generate task now evaluates that CodgenConfigurator is called as expected. The `DefaultGenerator` implementation should be tested separately, considering previous tests weren't evaluating anything that was generated.

### PR checklist

### DIFF
--- a/modules/openapi-generator-cli/src/test/java/org/openapitools/codegen/cmd/GenerateTest.java
+++ b/modules/openapi-generator-cli/src/test/java/org/openapitools/codegen/cmd/GenerateTest.java
@@ -17,446 +17,413 @@
 
 package org.openapitools.codegen.cmd;
 
-import org.mockito.MockSettings;
-import org.openapitools.codegen.ClientOptInput;
-import org.openapitools.codegen.DefaultGenerator;
-import org.openapitools.codegen.OpenAPIGenerator;
-import org.openapitools.codegen.config.CodegenConfigurator;
+import io.airlift.airline.Cli;
 import org.apache.commons.lang3.ArrayUtils;
+import org.mockito.MockSettings;
+import org.openapitools.codegen.DefaultGenerator;
+import org.openapitools.codegen.Generator;
+import org.openapitools.codegen.config.CodegenConfigurator;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
 
 import static org.mockito.Answers.CALLS_REAL_METHODS;
 import static org.mockito.Mockito.*;
+import static org.testng.Assert.fail;
 
 @SuppressWarnings("unused")
 public class GenerateTest {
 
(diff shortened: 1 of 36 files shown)


modules/openapi-generator-cli/src/test/java/org/openapitools/codegen/cmd/GenerateTest.java
```diff
@@ -17,536 +17,413 @@
...
-                configurator.setArtifactVersion(value);
-                times = 1;
-            }
-        };
+        verifyNoMoreInteractions(configurator);
     }
 
     @Test
-    public void testLibrary() throws Exception {
-        final String value = "library1";
```
jimschubert: `library1` here should have resulted in an exception, but jmockit swallowed it.

modules/openapi-generator/src/test/java/org/openapitools/codegen/options/ScalaHttpClientOptionsProvider.java
```diff
@@ -29,7 +29,7 @@
     public static final String SORT_PARAMS_VALUE = "false";
     public static final String SORT_MODEL_PROPERTIES_VALUE = "false";
     public static final String ENSURE_UNIQUE_PARAMS_VALUE = "true";
-    public static final String MODEL_PROPERTY_NAMING = "modelPropertyNaming";
```
jimschubert: This old value should have resulted in an exception, but jmockit swallowed it.

jimschubert:
fyi for @OpenAPITools/openapi-generator-collaborators 

I'm going to go ahead and merge this as there are usually very few new/updated "Options" tests. Everything else seems to already use mockito.

wing328:
@jimschubert fine with mockito. Thanks for the effort 👍

PR #5085, opened 2020-01-23
[typescript-angular] Fix typescript model kebab-cased filenames

### LINKED ISSUE #5073: [BUG][typescript-angular] Model properties with underscore prefix generates invalid file names/imports using kebab-case file naming
#### Bug Report Checklist

##### Description

I have an API that follows HAL standards. It includes a `_links` property defined as an inline object schema. When the typescript-angular generator is run, it creates a typescript file named `foo-response--links.ts`, but references it as `foo-response-links` in imports from other files.

I am generating this using the NPM client @openapitools/openapi-generator-cli@1.0.8-4.2.2 which bundles the 4.2.2 generator (I think).

Invoking with: `generate --generator-name typescript-angular --input-spec foo.yml --config config.json --skip-validate-spec`

Config:
```json
{
  "ngVersion": "7.2.0",
  "fileNaming": "kebab-case",
  "modelPropertyNaming": "original",
  "serviceSuffix": "Service",
  "sortParamsByRequiredFlag": true,
  "prependFormOrBodyParameters": true
}
```

API example (foo.yml):

```yaml
openapi: 3.0.0
info:
  title: Example API
  description: An Example API
  version: "1.0"
paths:
  /hello:
    get:
      summary: Example endpoint
      operationId: get_hello
      responses:
        200:
          description: Example response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/FooResponse'
components:
  schemas:
    FooResponse:
      type: object
      properties:
        _links:
          required:
            - collection
            - self
          type: object
          properties:
            collection:
              type: string
              format: uri
            self:
              type: string
              format: uri
```

Note that the `FooResponse` schema has a property `_links` defined as an object.

This generates a `foo-response.ts`  and a `foo-response--links.ts` (note the double hyphen.) 

In `foo-response.ts`, it tries to import the model from `foo-response-links.ts` (note the single hyphen) like so:

```typescript
import { FooResponseLinks } from './foo-response-links';

export interface FooResponse { 
  _links?: FooResponseLinks;
}
```

Note that when using 'camelCase' file naming, this issue is not present and the model files are generated as `fooResponse.ts` and `fooResponseLinks.ts` and linked appropriately.

Confirmed that this is an issue with @openapitools/openapi-generator-cli versions:

* 1.0.8-4.2.2 (current latest)
* 1.0.1-4.1.3

`generate --generator-name typescript-angular --input-spec foo.yml --config config.json --skip-validate-spec`

##### Suggest a fix

The double hyphen in the file name is very unusual but technically valid. The issue is that the logic that determines the file name to use, and the logic that determines the module name to import from is out of sync. 

So the fix is either:
* A. Generate a file name `foo-response-links.ts` without the double hyphen, to match the import. or:
* B. Update the import to be `from './foo-response--links'` to match the file name. 

(I personally prefer solution A because it's more natural.)

roddy:
When using "kebab-case" file names, the typescript-angular client generator would create model files with repeating dashes in the filenames, but the import statements would have single dashes. This was caused by the Inline Model Resolver naming the resolved models by concatenating the parent name to the child model name with an underscore separator; if the child model name had a leading underscore, this would cause a double underscore in the new model name. When this model name was converted to kebab case for the file name, each individual underscore was converted to a dash, resulting in a filename with repeated dashes.

To address, updated the `dashize` method in StringUtils to collapse multiple consecutive whitespace and/or underscores to a single dash. This _shouldn't_ be a breaking change, but if need be the fix could easily be amended to make the collapsing part of the operation optional using a boolean parameter and an overloaded method call.

Added test cases to the StringUtils unit tests, Angular Typescript client code gen tests, and Angular Typescript model tests. Ran all of the tests in the project using `mvn test`; no failures. There were no changes to the petstore APIs or templates.  Built the project and verified using the YML and commandline arguments from #5073 (yml included at bottom of PR).

Fixes #5073 

### PR checklist

@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11)

### Verification

Sample YML for verification (make sure to set fileNaming to "kebab-case"):

```yaml
openapi: 3.0.0
info:
  title: Example API
  description: An Example API
  version: "1.0"
paths:
  /hello:
    get:
      summary: Example endpoint
      operationId: get_hello
      responses:
        200:
          description: Example response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/FooResponse'
components:
  schemas:
    FooResponse:
      type: object
      properties:
        _links:
          required:
            - collection
            - self
          type: object
          properties:
            collection:
              type: string
              format: uri
            self:
              type: string
              format: uri
```

Previous behavior would generate files `./models/foo-response.ts` and `./models/foo-response--links.ts`. New behavior is to generate `./models/foo-response-links.ts` (other file unchanged).

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/StringUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/StringUtils.java
@@ -40,7 +40,7 @@ public class StringUtils {
      * @return The dashized version of the word, e.g. "my-name"
      */
     public static String dashize(String word) {
-        return underscore(word).replaceAll("[_ ]", "-");
+        return underscore(word).replaceAll("[_ ]+", "-");
     }
 
     /**
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularClientCodegenTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularClientCodegenTest.java
@@ -5,6 +5,7 @@ import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.PathItem;
 import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.media.StringSchema;
 import io.swagger.v3.oas.models.responses.ApiResponse;
 import io.swagger.v3.oas.models.responses.ApiResponses;
 import org.openapitools.codegen.CodegenOperation;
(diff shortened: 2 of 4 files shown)


macjohnny:
thanks for the PR and the tests!

PR #5094, opened 2020-01-23
[python] Cleanup ThreadPool with atexit rather than __del__

### LINKED ISSUE #5093: [Python] Generated code is out of contract - application hangs indefinitely
Bug originally reported by @gaborbernat in the swagger-codegen project, but the same issue applies here: https://github.com/swagger-api/swagger-codegen/issues/9991

The generated code snippet from - https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api_client.mustache#L78 is not allowed:

    def __del__(self):
        if self._pool is not None:
            self._pool.close()
            self._pool.join()

The very simple program of creating a global client, and then just exiting will hang indefinitely on Python3.8. Initially raised as CPython bug under https://bugs.python.org/issue39360, but @pablogsal (core CPython developer) investigated and concluded that this way of cleaning up pools is not supported. Instead one should either allow the user to close or if the user did not do the generated code should do by registering https://docs.python.org/3.8/library/atexit.html callbacks.

Fix incoming

fabianvf:
This removes the `__del__` function from the generated Python client,
and replaces it with a `cleanup` function. When a ThreadPool is created,
the cleanup function is registered with the `atexit` module.

This fixes #5093, where the API client could hang indefinitely at
garbage collection.

Related to https://github.com/swagger-api/swagger-codegen/pull/10002

### PR checklist

cc @taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess @slash-arun @spacether

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api_client.mustache
@@ -2,6 +2,7 @@
 {{>partial_header}}
 from __future__ import absolute_import
 
+import atexit
 import datetime
 from dateutil.parser import parse
 import json
@@ -75,11 +76,19 @@ class ApiClient(object):
         self.user_agent = '{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}OpenAPI-Generator/{{{packageVersion}}}/python{{/httpUserAgent}}'
         self.client_side_validation = configuration.client_side_validation
 
-    def __del__(self):
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def close(self):
         if self._pool:
             self._pool.close()
             self._pool.join()
             self._pool = None
+            if hasattr(atexit, 'unregister'):
+                atexit.unregister(self.close)
 
     @property
     def pool(self):
@@ -87,6 +96,7 @@ class ApiClient(object):
          avoids instantiating unused threadpool for blocking clients.
         """
         if self._pool is None:
+            atexit.register(self.close)
             self._pool = ThreadPool(self.pool_threads)
         return self._pool
 
--- a/modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
@@ -8,22 +8,26 @@ from pprint import pprint
 {{#hasAuthMethods}}
 # Defining host is optional and default to {{{basePath}}}
 configuration.host = "{{{basePath}}}"
-# Create an instance of the API class
-api_instance = {{{packageName}}}.{{{classname}}}({{{packageName}}}.ApiClient(configuration))
-{{#allParams}}{{paramName}} = {{{example}}} # {{{dataType}}} | {{{description}}}{{^required}} (optional){{/required}}{{#defaultValue}} (default to {{{.}}}){{/defaultValue}}
-{{/allParams}}
+# Enter a context with an instance of the API client
+with {{{packageName}}}.ApiClient(configuration) as api_client:
+    # Create an instance of the API class
+    api_instance = {{{packageName}}}.{{{classname}}}(api_client)
+    {{#allParams}}{{paramName}} = {{{example}}} # {{{dataType}}} | {{{description}}}{{^required}} (optional){{/required}}{{#defaultValue}} (default to {{{.}}}){{/defaultValue}}
+    {{/allParams}}
 {{/hasAuthMethods}}
 {{^hasAuthMethods}}
-# Create an instance of the API class
-api_instance = {{{packageName}}}.{{{classname}}}()
-{{#allParams}}{{paramName}} = {{{example}}} # {{{dataType}}} | {{{description}}}{{^required}} (optional){{/required}}{{#defaultValue}} (default to {{{.}}}){{/defaultValue}}
-{{/allParams}}
+# Enter a context with an instance of the API client
+with {{{packageName}}}.ApiClient() as api_client:
+    # Create an instance of the API class
+    api_instance = {{{packageName}}}.{{{classname}}}(api_client)
+    {{#allParams}}{{paramName}} = {{{example}}} # {{{dataType}}} | {{{description}}}{{^required}} (optional){{/required}}{{#defaultValue}} (default to {{{.}}}){{/defaultValue}}
+    {{/allParams}}
 {{/hasAuthMethods}}
 
-try:
-{{#summary}}    # {{{.}}}
-{{/summary}}    {{#returnType}}api_response = {{/returnType}}api_instance.{{{operationId}}}({{#allParams}}{{#required}}{{paramName}}{{/required}}{{^required}}{{paramName}}={{paramName}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}
-    pprint(api_response){{/returnType}}
-except ApiException as e:
-    print("Exception when calling {{classname}}->{{operationId}}: %s\n" % e)
+    try:
+    {{#summary}}    # {{{.}}}
+    {{/summary}}    {{#returnType}}api_response = {{/returnType}}api_instance.{{{operationId}}}({{#allParams}}{{#required}}{{paramName}}{{/required}}{{^required}}{{paramName}}={{paramName}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}
+        pprint(api_response){{/returnType}}
+    except ApiException as e:
+        print("Exception when calling {{classname}}->{{operationId}}: %s\n" % e)
 ```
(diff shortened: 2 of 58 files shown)


spacether:
Thank you for this PR. Can you add a test verifying that the garbage collection is cleaning up the threadpool?

fabianvf:
@spacether hmm, probably, though I'm not totally sure how off the top of my head. I'll dig into it a bit, did you have an approach in mind?

spacether:
Perhaps use gc.collect per here https://stackoverflow.com/questions/1641717/manual-garbage-collection-in-python

gaborbernat:
> Thank you for this PR. Can you add a test verifying that the garbage collection is cleaning up the threadpool?

No, please don't. This is not supported in CPython. It's a programming paradigm that might work in other languages, but not CPython. The whole point of this PR is that thread pools should not be handled within a garbage collector as that puts you in undefined space. Remember in Python it's not guaranteed the garbage collector will ever be called. If the user decided that they don't manually close down the threadpool the only fallback is to do it via atexit as the final line of defence. In case you have resources though it's recommended to encourage usage of the objects via the ``with`` that guarantees a cleanup operation to be called.

spacether:
> > Thank you for this PR. Can you add a test verifying that the garbage collection is cleaning up the threadpool?

Thank you for letting us know that using the gc library is the wrong way to test this.
My intention is for us to add a test verifying that this is working.
@gaborbernat/@fabianvf what do you think of us also adding an __enter__ and __exit__ methods?
That is suggested here: https://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object
And would allow our users to also use with context like
```
try:
    with petstore_api.PetApi(petstore_api.ApiClient(configuration)) as api_instance:
        body = petstore_api.Pet() # Pet | Pet object that needs to be added to the store
        # Add a new pet to the store
        api_instance.add_pet(body)
except ApiException as e:
    print("Exception when calling PetApi->add_pet: %s\n" % e)
```
@fabianvf 
Can you add a test verifying that the atexit method is working?
- It looks like you can call atexit._run_exitfuncs() to simulate the program exit per https://github.com/python/cpython/blob/master/Lib/test/test_atexit.py#L51
Can you also add this feature to the python-experimental generator?
- In the OpenApi Slack general channel we talked about eventually making python-experimental the primary python generator. Adding helpful fixes like this in both generators would may that future switch easier.

fabianvf:
@spacether well I know @gaborbernat is in favor of it, I was worried it would change the PR from a bugfix to a feature and slow it down (we're hitting issues with hanging in the kubernetes python client). I added it in, working on figuring out the test now.

fabianvf:
Tests added, wasn't totally sure if those are generated tests or not so let me know if I need to put them somewhere else

spacether:
@fabianvf you put them in the right place, thanks!
Do you think that we should update the docs to use a with context manager for the ApiClient?
Keeping it as is or adding the with both make sense to me.
- Adding the with encourages the good practice of using the context for ApiClient
- Keeping it as is: technically we don't need to add the with in the samples because atexit will invoke cleanup when the python program stops running
What do you think?

fabianvf:
@spacether probably makes sense to update the docs to encourage the use of the context manager, since there could be longer running programs that create/destroy multiple clients that could still be impacted

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -75,7 +76,13 @@ class ApiClient(object):
...
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def close(self):
         if self._pool:
             self._pool.close()
             self._pool.join()
```
gaborbernat: after the pool is closed and joined we should unregister the atexit -  https://docs.python.org/3/library/atexit.html#atexit.unregister; this is so that we don't keep the object in memory until the application dies

fabianvf: :+1: Makes sense

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -75,18 +76,27 @@ class ApiClient(object):
...
             self._pool = None
+            if hasattr(atexit, 'unregister'):
```
fabianvf: it looks like `atexit.unregister` doesn't exist in Python 2, so I added this guard

spacether:
> @spacether probably makes sense to update the docs to encourage the use of the context manager, since there could be longer running programs that create/destroy multiple clients that could still be impacted

Can you update these doc templates to use with context managers for api_client?
- https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
- https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/api_doc_example.mustache
- https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/common_README.mustache
- https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/README_common.mustache

spacether:
@fabianvf can you run `bin/utils/ensure-up-to-date` and update you PR?
- python-asyncio
- python-tornado
Need to be updated

fabianvf:
@spacether thank you, fixed!

modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
```diff
@@ -8,22 +8,26 @@ from pprint import pprint
 {{#hasAuthMethods}}
```
spacether: Can you add a line here instantiating configuration?

modules/openapi-generator/src/main/resources/python/python-experimental/api_doc_example.mustache
```diff
@@ -7,51 +7,55 @@ from pprint import pprint
 {{#hasAuthMethods}}
```
spacether: Can you add a line above {{#hasAuthMethods}} instantiating configuration?

modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
```diff
@@ -8,22 +8,26 @@ from pprint import pprint
...
 # Defining host is optional and default to {{{basePath}}}
 configuration.host = "{{{basePath}}}"
-# Create an instance of the API class
-api_instance = {{{packageName}}}.{{{classname}}}({{{packageName}}}.ApiClient(configuration))
-{{#allParams}}{{paramName}} = {{{example}}} # {{{dataType}}} | {{{description}}}{{^required}} (optional){{/required}}{{#defaultValue}} (default to {{{.}}}){{/defaultValue}}
-{{/allParams}}
+# Enter a context with an instance of the API client
```
spacether: Aren't these blocks `{{#hasAuthMethods}}`/`{{^hasAuthMethods}}` identical except for setting 
 configuration.host?
Please move the identical block contents  underneath the `{{/hasAuthMethods}}` on line 25

modules/openapi-generator/src/main/resources/python/python-experimental/README_common.mustache
```diff
@@ -7,18 +7,20 @@ from pprint import pprint
 {{> python_doc_auth_partial}}
 # Defining host is optional and default to {{{basePath}}}
```
spacether: Do we need to instantiate configuration here?

modules/openapi-generator/src/main/resources/python/python-experimental/api_doc_example.mustache
```diff
@@ -7,51 +7,55 @@ from pprint import pprint
...
 # Defining host is optional and default to {{{basePath}}}
 configuration.host = "{{{basePath}}}"
-# Create an instance of the API class
-api_instance = {{{packageName}}}.{{{classname}}}({{{packageName}}}.ApiClient(configuration))
+# Enter a context with an instance of the API client
+with {{{packageName}}}.ApiClient(configuration) as api_client:
+    # Create an instance of the API class
+    api_instance = {{{packageName}}}.{{{classname}}}(api_client)
```
spacether: Aren't these blocks `{{#hasAuthMethods}}`/`{{^hasAuthMethods}}` identical except for setting 
 configuration.host?
Please move the identical block contents  underneath the `{{/hasAuthMethods}}` on line 20

modules/openapi-generator/src/main/resources/python/python-experimental/README_common.mustache
fabianvf: I believe configuration is instantiated in the ` {{> python_doc_auth_partial}}` snippet

modules/openapi-generator/src/main/resources/python/api_doc_example.mustache
fabianvf: I believe configuration is instantiated in the `{{> python_doc_auth_partial}}` snippet

modules/openapi-generator/src/main/resources/python/python-experimental/README_common.mustache
fabianvf: actually it looks like that snippet doesn't even exit in the experimental directory...

Issue #5025, opened 2020-01-17
[PHP] Code generator does not use/validate "scheme" security attribute


sebastien-rosset:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The PHP code generator does not validate the value of the "scheme" attribute in the security scheme object. It assumes that when type is set to **"http"**, the auth scheme is "basic". But that's not always the case. Other HTTP schemes that are defined in the IANA registry.

The "scheme" attribute in the security scheme object is required as specified in [OAS 3.0.2](https://github.com/OAI/OpenAPI-Specification/blob/v3.1.0-dev/versions/3.0.2.md#security-scheme-object)
The value of the "scheme" attribute must be the auth scheme from the IANA registry available at
https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml (as per RFC 7235).

##### openapi-generator version

master branch, January 18 2020

##### OpenAPI declaration file content or url

Consider the following OAS security schemes:

```yaml
securitySchemes:
   http_basic_test:
      type: http
      scheme: basic
   http_scram:
      type: http
      scheme: SCRAM-SHA-256
   http_other_scheme:
      type: http
      scheme: my-security-scheme-value
```
The PHP generator properly generates the code for "http_basic_test". For the other two scheme values (http_scram and http_other_scheme), the generator still produces code assuming the scheme value is "basic". 

I found this issue while working on PR #4993 and #4958

##### Command line used for generation

./bin/openapi3/php-petstore.sh

##### Steps to reproduce

1. Edit modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml and add the following scheme:

```yaml
   http_signature_test:
      type: http
      scheme: signature
```
The "signature" value is NOT HTTP basic. This is just an example, I think any value would produce the same results.

2. Execute ./bin/openapi3/php-petstore.sh

[...]

PR #5095, opened 2020-01-23
[HttpBasicAuth] Use positive isBasicBasic tag instead of negative isBasicBearer

sebastien-rosset:
For HTTP basic auth, many mustache templates use a convention to negate the isBasicBearer tag, i.e. the case for HTTP basic auth is the negation of isBasicBearer. The problem is this only works when there are two possible HTTP auth, 1) basic auth and 2) bearer token. To support more http auth methods, we should stop using the negation of "isBasicBearer".

```
...
{{^isBasicBearer}}
   // do stuff for HTTP basic authentication
{{/isBasicBearer}}
{{#isBasicBearer}}
   // do stuff for bearer authentication
{{/isBasicBearer}}
....
```

I am proposing that instead, we change the templates as follows:
```
{{#isBasicBasic}}
   // do stuff for HTTP basic authentication
{{/isBasicBasic}}
{{#isBasicBearer}}
   // do stuff for bearer authentication
{{/isBasicBearer}}
....
```

I started making the change for Javascript. If that seems a reasonable change to reviewers, I can make the change everywhere.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache
@@ -43,9 +43,9 @@
 {{/emitJSDoc}}{{=< >=}}    this.authentications = {
 <#authMethods>
 <#isBasic>
-<^isBasicBearer>
+<#isBasicBasic>
       '<name>': {type: 'basic'}<^-last>,</-last>
-</isBasicBearer>
+</isBasicBasic>
 <#isBasicBearer>
       '<name>': {type: 'bearer'}<^-last>,</-last><#bearerFormat> // <&.></bearerFormat>
 </isBasicBearer>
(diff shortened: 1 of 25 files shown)


sebastien-rosset:
This should fix #5025

sebastien-rosset:
See also #4993, #4959 and #4958, support for HTTP signature in core codegen, Python and golang.

sebastien-rosset:
@richardwhiuk , can you comment on this line: https://github.com/OpenAPITools/openapi-generator/blame/master/modules/openapi-generator/src/main/resources/rust-server/client-mod.mustache#L453

Logically, the test is "if OAuth2 and NOT BasicBearer". Can this be replaced with a positive test? The problem with not "BasicBearer" is that when new HTTP auth methods are added, this will still match the (NOT BasicBearer) clause, and that's probably not what you want.

richardwhiuk:
@sebastien-rosset I think that code is fine - it's just trying to eliminate a double match on Bearer being generated.

PR #5106, opened 2020-01-25
[scala] fix akka-scala-client template compilation warnings

### LINKED ISSUE #4004: [BUG][Scala-AKKA] The custom package name is being overwritten in reference.conf file
Hi,

I been settings custom `mainPackage` name in the `pom` file.
```
      <plugin>
         ...
            <configuration>
              <inputSpec>${project.basedir}/target/licenseplate-api-contract.json</inputSpec>
              <generatorName>scala-akka</generatorName>
              <configOptions>
                <mainPackage>com.some.package</mainPackage>
                <configKeyPath>com.some.package</configKeyPath>
              </configOptions>
              <output>${project.basedir}</output>
            </configuration>
      ...
      </plugin>
``` 
When code generation complete the `reference.conf` file contains hardcoded package name `org.openapitools.client` which is causing runtime error making HTTP request:
```
.../reference.conf: 1: No configuration setting found for key 'com'
```
I need the same namespace as `mainPackage`.   

At the first glance looking at`ScalaAkkaClientCodegen` it look like the `configKeyPath` is not being set from configuration.
Please can you fix this issue?
Can we have custom api, model & invoker package - it been ignored now?

Regards,
Sergey

chameleon82:
- fix akka-scala-client template compilation warnings
- COOKIE location authorization key added
- fix #4004 custom package name in reference.conf

### PR checklist


wing328:

wing328:
Please update the test case (https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/scalaakka/ScalaAkkaClientCodegenTest.java#L362) to resolve the build failure.

modules/openapi-generator/src/test/resources/3_0/petstore.yaml
```diff
@@ -41,6 +50,15 @@ paths:
...
       responses:
+        '200':
+          description: successful operation
+          content:
+            application/xml:
+              schema:
+                $ref: '#/components/schemas/Pet'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Pet'
```
wing328: @chameleon82 are these a must? In other words, would the change still work without documenting the HTTP 200 response in the spec?

We prefer not to update petstore.yaml to keep it as original as possible.

modules/openapi-generator/src/main/resources/scala-akka-client/model.mustache
```diff
@@ -1,9 +1,10 @@
 {{>licenseInfo}}
 package {{package}}
 
+{{#imports}}
+import {{import}}
+{{/imports}}
 import {{mainPackage}}.core.ApiModel
-import org.joda.time.DateTime
-import java.util.UUID
```
jimschubert: This will be a potential breaking change unless this is added to the constructor:

```
importMapping.put("UUID", "java.util.UUID");
```

modules/openapi-generator/src/test/resources/3_0/petstore.yaml
jimschubert: @wing328 Thanks for calling this out. When you say original, what are we keeping it original compared to? This spec isn't the "official" petstore.yaml spec from OpenAPI Specification 3.0 examples.

I'd have some concerns about _not_ including these changes. It would mean that the PUT and POST methods are only documented to ever return a client error. From https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#responsesObject:

> A container for the expected responses of an operation. The container maps a HTTP response code to the expected response.
> 
> The documentation is not necessarily expected to cover all possible HTTP response codes because they may not be known in advance. However, documentation is expected to cover a successful operation response and any known errors.
> 
> The default MAY be used as a default response object for all HTTP codes that are not covered individually by the specification.
> 
> The Responses Object MUST contain at least one response code, and it SHOULD be the response for a successful operation call.

Although the line "documentation is expected to cover a successful operation response" doesn't use the specification terminology `MUST`, `MAY` or `SHOULD`, I think our test yaml breaking 'expectations' could be confusing especially since there is no 'default' response documented.

chameleon82: @wing328 @jimschubert I've checked than one https://petstore3.swagger.io/api/v3/openapi.yaml which currently has those specified, openApi specs also have specified successful response here https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore.yaml and here https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore-expanded.yaml
However https://petstore.swagger.io/v2/swagger.yaml has not specified even default one. 
Seems that case must be clear in `OpenApi Specification`. 
I suppose `SHOULD` for successful operation means that client has another way to check result. For example, by calling `GET /pet/123` after calling `POST /pet {id = 123, ... }`
For the current case i have added this to validate successful scenarios in the unit tests

modules/openapi-generator/src/main/resources/scala-akka-client/model.mustache
chameleon82: it is already there in `ScalaAkkaClientCodegen extends AbstractScalaCodegen extends` [DefaultCodegen](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L1229) base class.
I will add unit test to check there is no breaking changes for the further versions

modules/openapi-generator/src/test/resources/3_0/petstore.yaml
jimschubert: You're right that's probably why they've used `SHOULD`. However, the common way to do that without a returned body is to use links in the header of a _successful_ response. I think your changes to include 2xx type responses makes sense. If we don't include positive responses for these endpoints, I think we should remove the endpoints entirely, because no realistic API returns only client errors for a path.

@wing328 are you cool with keeping these changes? We should probably also review all our go-to specs and make sure they represent real-world designs.

wing328: Totally agreed with reviewing the spec to ensure it's correct. Let's go with this PR for the time being.

wing328:
For some reasons, I can't comment/reply to https://github.com/OpenAPITools/openapi-generator/pull/5106#discussion_r375040629. Agreed with revising the "original" as it may not be perfectly accurate. I'll create a separate issue to track it.

PR #5116, opened 2020-01-26
[ci][test] Modifications to static analysis contribution via Tomasbjerre

### LINKED ISSUE #33: Discuss static analysis options

jimschubert:
This extends on submission by @tomasbjerre in #5040 by moving the PMD and Spotbugs  plugin execution to a separate profile `static-analysis`. This keeps CheckStyle in the default build because it added a very minimal amount of time to execution.

Once merged, I'd like to extend the Sonar CI GitHub Action to be a static analysis action which caches the xml reports generated by these plugins as build artifacts.  This will give us SonarCloud as a means for evaluating the static analysis, and allow for contributors to download the `*.xml` reports from the action. Once the GitHub action is updated, we can also close #33 as well.

Core contributors may run static analysis with:

```
mvn -Pstatic-analysis install
```

The analysis is separated from default functionality to reduce impact to community contributions. SpotBugs/PMD may add a non-trivial amount of time to builds on some machines.

Closes #5040

### PR checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>3.1.0</version>
+                <version>${checkstyle.plugin.version}</version>
                 <configuration>
                     <consoleOutput>false</consoleOutput>
                     <failsOnError>false</failsOnError>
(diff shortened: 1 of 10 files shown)


spotbugs-exclude.xml
```diff
@@ -0,0 +1,34 @@
...
+			<And>
+				<Class name="io.undertow.server.protocol.ajp.AjpRequestParseState" />
+				<Field name="dataSize" />
+			</And>
+		</Or>
+	</Match>
+
+	<!-- For source/class naming, see https://spotbugs.readthedocs.io/en/stable/filter.html#java-element-name-matching -->
+	<Match>
+		<!-- Ignore all source files with samples in path name -->
```
tomasbjerre: One could argue that the generated code should also be free from issues found by static code analysis. I'm not sure about excluding samples here.

I can see how users have different code standards and running Checkstyle and PMD is not worth the effort, but I'm not sure it is good to exclude Spotbugs here.

pom.xml
```diff
@@ -470,6 +542,73 @@
...
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>com.github.spotbugs</groupId>
+                        <artifactId>spotbugs-maven-plugin</artifactId>
+                        <version>${spotbugs.plugin.version}</version>
+                        <configuration>
+                            <failOnError>false</failOnError>
+                            <!-- https://spotbugs.readthedocs.io/en/stable/effort.html -->
+                            <effort>min</effort>
```
tomasbjerre: Now that it does not run by default, perhaps it does not need to be `min`?

tomasbjerre:
The `travis.yml` can be extended to include this tool:
https://www.npmjs.com/package/violation-comments-to-github-command-line

But perhaps you would rather have Sonar do the commenting.

spotbugs-exclude.xml
jimschubert: Another reason to exclude these here is that I assume most users use custom templates. Analyzing our "samples" is therefore very much like analyzing our tests.

jimschubert: Another reason to exclude these here is that I assume most users use custom templates. Analyzing our "samples" is therefore very much like analyzing our tests.

pom.xml
jimschubert: I can evaluate this once it's hooked into the target pipeline.

spotbugs-exclude.xml
jimschubert: Another concern is that only a very small subset of our samples are Java :\
Maybe we could later work out the pattern to exclude everything but java client and server outputs in the samples directory.

jimschubert:
> The `travis.yml` can be extended to include this tool:
> [npmjs.com/package/violation-comments-to-github-command-line](https://www.npmjs.com/package/violation-comments-to-github-command-line)
> 
> But perhaps you would rather have Sonar do the commenting.

We have some cleanup to do with respect to how much is being logged in Travis. As it is, both Travis and CirciCI UIs lock up my browser because of the amount of text output. I'm looking at modifying our logger to reduce duplicated log lines, and once that's done I will look out the outputs in other CI.

I'm thinking the `static-analysis` profile may also be run in CircleCI for contributor outputs.

PR #5124, opened 2020-01-27
[C++][Pistache] Fix operator== for empty model structs

kuzkry:
When `useStructModel=true` is set and a model specifies no properties, then a generated struct will not compile.
```
(...) error: return-statement with no value, in function returning 'bool' [-fpermissive]
 bool NoProperties::operator==(const NoProperties& other) const { return; }
```

### PR checklist

P.S. I refuse to run `bin/openapi3/cpp-pistache-server-petstore.sh` since the changes I get seem to be unrelated.

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-pistache-server/model-struct-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-pistache-server/model-struct-source.mustache
@@ -23,7 +23,7 @@ nlohmann::json {{classname}}::to_json() const
 
 bool {{classname}}::operator==(const {{classname}}& other) const
 {
-    return {{#vars}}{{baseName}} == other.{{baseName}}{{#hasMore}} && {{/hasMore}}{{/vars}};
+    return true{{#vars}} && {{baseName}} == other.{{baseName}}{{/vars}};
 }
 
 bool {{classname}}::operator!=(const {{classname}}& other) const


modules/openapi-generator/src/main/resources/cpp-pistache-server/model-struct-source.mustache
```diff
@@ -23,7 +23,7 @@ nlohmann::json {{classname}}::to_json() const
 
 bool {{classname}}::operator==(const {{classname}}& other) const
 {
-    return {{#vars}}{{baseName}} == other.{{baseName}}{{#hasMore}} && {{/hasMore}}{{/vars}};
+    return true{{#vars}} && {{baseName}} == other.{{baseName}}{{/vars}};
```
etherealjoy: I think this is incorrect. May I suggest something like this?

```
    return true{{#vars}} && {{baseName}} == other.{{baseName}}{{#hasMore}} && {{/hasMore}}{{/vars}};

**Comment by kuzkry** [comment]
Actually hold on, I also get warnings about unused parameters. I'll try to work this out too.

**Comment by kuzkry** [comment]
> Actually hold on, I also get warnings about unused parameters. I'll try to sort this out too.

Ok, I believe we can proceed with this PR and leave warnings for the other pull request.

**Comment by kuzkry** [comment]
Actually what I think now is that we don't need that. Here is why:

I wanted to have something like that:
```yaml
components:
  schemas:
    Point2D:
      type: object
      properties:
        x:
          type: integer
          format: int32
        y:
          type: integer
          format: int32
    Position:
      $ref: '#/components/schemas/Point2D'
```
where:
- `Point2D` is my common type
- `Position` is my wanna-be alias for `Point2D` which suits better for my particular endpoint specification

This is why I got empty struct of `Position` which caused compilation error.
But now I think this is a wrong use of [the OpenAPI specification](https://swagger.io/docs/specification/data-models/data-types/#object) and I should have only been using `Point2D`.
If I'm not mistaken, then I guess this overuse of specification could be patched so the generator would recognize similar cases and tell users to fix their specification.

So assuming that aliases don't work like that, I think this PR is N/A because such an object:
```yaml
components:
  schemas:
    Empty:
      type: object
```
will anyway result in 
`[main] INFO  o.o.codegen.DefaultGenerator - Model LOL not generated since it's a free-form object`
and not generate an empty struct.

etherealjoy:
Makes sense :)

PR #5125, opened 2020-01-27
[C++][Pistache] Add neutral discards to supress a potential -Wunused-parameter warning

kuzkry:
**Note: this is dependent on #5124, so there is only one commit in this PR.**

When `useStructModel=true` is set and a model specifies no properties and assuming no compilation errors, then a generated struct will not produce -Wunused-parameter warnings.

Unfortunately, I didn't find a nice and clean solution. My ideas include:
- `(void) argument` or `static_cast<void>(argument)`
- conditionally specify only type like `const NoProperties&` instead of `const NoProperties& o`, however that means a messier template
- `__attribute__((unused))` but this isn't standard
- `[[maybe_unused]]` but it's C++17 so I guess it's a no-go
- `template <typename T> void ignore(T) {}`

So I decided to use the first option as it's a noop for compilers but I dunno. If you have any better idea, let me know.

### PR checklist

P.S. I still :wink: refuse to run `bin/openapi3/cpp-pistache-server-petstore.sh` since the changes I get seem to be unrelated.

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-pistache-server/model-struct-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-pistache-server/model-struct-source.mustache
@@ -23,7 +23,7 @@ nlohmann::json {{classname}}::to_json() const
 
 bool {{classname}}::operator==(const {{classname}}& other) const
 {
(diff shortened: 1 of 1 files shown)


kuzkry:
As its parent PR (#5124) is closed, it makes no sense to merge it.

Issue #5127, opened 2020-01-28
[BUG][typescript-angular] Incorrect import path for inline models 


amakhrov:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

Inline models (if there are more than one) together with fileNaming="kebab-case" result in incorrect import paths.
E.g. `InlineObject1` is created in `models/inline-object-1.ts`. But generated api file tries to import it as
```
import {InlineObject1} from '../models/inline-object1'
```

##### openapi-generator version

Master and stable

##### OpenAPI declaration file content or url

[inline-models.yml (gist)](https://gist.github.com/amakhrov/7967679d2945ba1daf30a2313afd69df)

Important part is that there are 2 POST endpoints with inline models in body.

##### Command line used for generation

```
docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate \
    -i /local/inline-models.yml \
    -g typescript-angular \
    --config /local/config.json \
    -o /local/generated
```
and `config.json`:
```
{
  "fileNaming": "kebab-case",
}
```

##### Steps to reproduce

- Generate the files using the spec.
- Compare the name of the file containing InlineObject1 model with the import path used in `api/default.service.ts`
- **Expectation**: they should match
- **Actual**: file name is `inline-object-1`; import path is `inline-object1`

##### Suggest a fix

PR: https://github.com/OpenAPITools/openapi-generator/pull/5128

macjohnny:
@amakhrov

> E.g. `InlineObject1` is created in `models/inline-object-1.ts`. But generated api file tries to import it as
> 
> ```
> import {InlineObject1} from '../models/inline-object-1'
> ```

this seems correct, or am I missing something?

amakhrov:
My bad. The actual import looks like
```
import {InlineObject1} from '../models/inline-object1'
```
see the missed dashed before `1`

PR #5128, opened 2020-01-28
[typescript-angular] Import models from a single model/models.ts

amakhrov:
Fixes https://github.com/OpenAPITools/openapi-generator/issues/5127

The underlying problem causing https://github.com/OpenAPITools/openapi-generator/issues/5127 is that import paths used in api files are produced based on model class names (e.g `InlineObject1`), while the actual model files written by the generator take filenames based on raw model name (`inline_object_1`).

Perhaps a proper way would be to eliminate this inconsistency by using the same logic in both places.
However, import paths for api are assigned in `DefaultGenerator`, so the change can potentially have large impact on all generators.

So instead of fixing this inconsistency, this PR offers an alternative solution: ignore individual import paths prepared for api, and import all models from `../model/models.ts`.
This approach is actually in line with a couple of other typescript generators: typescript-axios, typescript-fetch, typescript-jquery.

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -448,13 +448,6 @@ public void postProcessParameter(CodegenParameter parameter) {
 
         operations.put("hasSomeFormParams", hasSomeFormParams);
 
-        // Add additional filename information for model imports in the services
-        List<Map<String, Object>> imports = (List<Map<String, Object>>) operations.get("imports");
-        for (Map<String, Object> im : imports) {
-            im.put("filename", im.get("import"));
-            im.put("classname", im.get("classname"));
-        }
-
```
amakhrov: NOTE: This is not actually needed for the fix.
However, after the the changes in this PR this `filename` property is no longer used in the templates. And this is not reliable anyway in the case of inline models. So I just did this cleanup rather than accumulating it as tech debt.

Apparently, this might break some setups that use externally-provided templates (via `--template-dir`).
I'm not sure what the policy is for such changes - is it safe to push to master? Should there be any changelog covering that? Or is it ok because codegen does not consider such custom properties a part of a public contract?

Open to suggestions here!

macjohnny: I suggest to leave it there, and remove it later for a major release

amakhrov: I'll restore these shortly and add a comment saying it's not used any more

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -448,13 +448,6 @@ public void postProcessParameter(CodegenParameter parameter) {
 
         operations.put("hasSomeFormParams", hasSomeFormParams);
```
macjohnny:
```suggestion
        operations.put("hasSomeFormParams", hasSomeFormParams);
        
        // Add additional filename information for model imports in the services	
        List<Map<String, Object>> imports = (List<Map<String, Object>>) operations.get("imports");	
        for (Map<String, Object> im : imports) {	
            im.put("filename", im.get("import"));	
            im.put("classname", im.get("classname"));	
        }
        
```

PR #5137, opened 2020-01-29
fix #5136 - adds templateResourcePath option to load custom templates from a configurable resource path instead from a directory.

### LINKED ISSUE #5136: [REQ] Load templates from classpath

Currently the only way to overwrite embedded templates is to use the templateDir. If openapi generator is used in multiple projects that should share all the same customized templates you have to copy the templates to each project or to fork and adjust he generator project itself and adjust the embedded templates directly.

Add a templateResourcePath option that if set defines a resource path in the classpath to look for templates files. If both templateDir and templateResourcePath is set, template files from templateDir should take precedence. By doing so you can define a company wide set of templates that overwrite the embedded templates but still allowing to overwrite templates on a per api basis.  

Extending templateDir to allow a comma-separated list of directories and allowing Spring like "classpath:" prefix denoting paths that should be resolved via classloader. But I did not opt for this option because templateDir is already a File and not a String, so you would need to alter this and in the end it sounded a little bit too much. Thus I decided for a new option.

## Additional context

Pull Request will follow.

eicki:
Added a templateResourcePath option, see RFC 5136 for details.

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -16,9 +16,9 @@ The transform logic needs to implement [CodegenConfig.java](https://github.com/o
 ## Modifying Templates
 
 > OpenAPI Generator applies user-defined templates via options:  
-> * CLI: `-t/--template` CLI options
-> * Maven Plugin: `templateDirectory`
-> * Gradle Plugin: `templateDir`
+> * CLI: `-t/--template` (or `-trp/--template-resource-path` for a classpath resource) CLI options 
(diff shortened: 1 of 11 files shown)


eicki:
#5164 fixed the issue in an appropriate way

jimschubert:
@eicki hey, sorry. I didn't see that you'd had a PR open for that work. Looks like referencing the issue you opened in the PR title doesn't link back to the issue.

Issue #5135, opened 2020-01-28
[BUG][typescript][typescript-angular] enumSuffix appends to modelNameSuffix


amakhrov:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

1. [all typescript generators] When `modelNameSuffix` option is provided, generated enum names concatenate this with enumSuffix, resulting in smth like `BaseNameModelSuffixEnum`.
2. [`typescript-angular`] A custom option `modelSuffix` is added. When specified, it leads to the same issue with enums (when `stringEnums=false`)
3. [`typescript-angular`] The `stringEnums=true` option removes the default `"Enum"` suffix, but still keeps the one coming from `modelSuffix`.

Let's say we have schema with enum:
```
Response:
  type: object
  properties:
    prop_name:
      type: string
      enum: [ Capital, UPPER, lower, snake_case, 'illegal+id' ]
```

Here is what is produced with various generator options:

| generator | options | enum name |
| - | - | - |
| typescript-angular | stringEnums=true | ResponsePropName |
| typescript-angular | stringEnums=true, modelSuffix="Model" | ResponseModelPropNameModel |
| typescript-angular | stringEnums=false | Response.PropNameEnum |
| typescript-angular | stringEnums=false, modelSuffix="Model" | ResponseModel.PropNameModelEnum |
| typescript-axios | - | ResponsePropNameEnum |
| typescript-axios | modelNameSuffix="Model" | ResponseModelPropNameModelEnum |

Other TS generators (typescript-fetch, typescript-rxjs) behave similar to typescript-axios.

##### openapi-generator version

master, stable

##### OpenAPI declaration file content or url

[gist: spec-enum.yml](https://gist.github.com/amakhrov/c2956d36e599d9833a40309e86490484)

##### Suggest a fix

I suggest the following behavior for all TS generators:
- Without modelNameSuffix, keep the existing behavior demonstrated by typescript-axios.
- With modelNameSuffix, only append "Enum" suffix but not model suffix. So instead of `ResponseModelPropNameModelEnum` we should get `ResponseModelPropNameEnum`
[...]

PR #5138, opened 2020-01-29
[typescript] Append enum suffix without model suffix

amakhrov:
Fixes https://github.com/OpenAPITools/openapi-generator/issues/5135

~I was not able to approach this problem without introducing backward-incompatible changes.
A new `enumSuffix` option is introduced to minimize migration efforts, though.~
**UPDATE**: actually, backward compatibility is now ensured by having `v4-compat` value by default for the newly introduced option

- (**breaking change**) With modelNameSuffix, only append "Enum" suffix but not model suffix. So instead of `ResponseModelPropNameModelEnum` we now get `ResponseModelPropNameEnum`.
- Without modelNameSuffix, all typescript generators except typescript-angular keep the existing behavior (emit smth like `ResponsePropNameEnum`)
- (**breaking change**) typescript-angular now follows the same enum naming rules as other TS generators. It still uses its custom `modelSuffix`, though - so it's still possible to configure it via `modelNameSuffix` or `modelSuffix`. Both options should work the same with the updated enum naming schema.
- `enumSuffix` is now configurable via options. This would provide smoother migration for existing users (e.g. one can set it to "ModelEnum" to mimic the old behavior).
- (_bonus_) More robust model names generation. Previously it would be possible to have a spec with a model named `error` (note lowcase!), which would then emit a Typescript interface `Error`. This should not have been possible, because `Error` is a builtin class. Now all the sanity checks are done to a fully-transformed model (after being camelized, and adding prefix and suffix). This addresses a part of the problem described in https://github.com/OpenAPITools/openapi-generator/issues/5139)

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -49,6 +49,11 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
     public static final String NPM_VERSION = "npmVersion";
     public static final String SNAPSHOT = "snapshot";
 
+    public static final String ENUM_NAME_SUFFIX_V4_COMPAT = "v4-compat";
+    public static final String ENUM_NAME_SUFFIX_DESC_CUSTOMIZED = CodegenConstants.ENUM_NAME_SUFFIX_DESC
+            + " A special '" + ENUM_NAME_SUFFIX_V4_COMPAT + "' value enables the backward-compatible behavior (as pre v4.2.3)";
+
+
     // NOTE: SimpleDateFormat is not thread-safe and may not be static unless it is thread-local
     @SuppressWarnings("squid:S5164")
     protected static final ThreadLocal<SimpleDateFormat> SNAPSHOT_SUFFIX_FORMAT = ThreadLocal.withInitial(() -> new SimpleDateFormat("yyyyMMddHHmm", Locale.ROOT));
@@ -59,7 +64,9 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
     protected String npmName = null;
     protected String npmVersion = "1.0.0";
 
-    protected String enumSuffix = "Enum";
+    protected String enumSuffix = ENUM_NAME_SUFFIX_V4_COMPAT;
+    protected Boolean isEnumSuffixV4Compat = false;
+
     protected String classEnumSeparator = ".";
 
     public AbstractTypeScriptClientCodegen() {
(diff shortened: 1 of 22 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -149,11 +149,13 @@ public AbstractTypeScriptClientCodegen() {
         typeMapping.put("DateTime", "Date");
         typeMapping.put("binary", "any");
         typeMapping.put("File", "any");
+        typeMapping.put("file", "any");
```
amakhrov: This is a mapping for the OAS2 built-in `file` type. It's similar to the existing mapping defined in `TypeScriptAngularClientCodegen.java`:
```
typeMapping.put("file", "Blob");
```
Without that, the updated `toModelName()` logic would add a bogus import for api endpoints that deal with type:file (`imageUpload` endpoint in PetStore)

Frankly, I'm not sure if the old mapping in line 151 here is needed at all.

macjohnny:
@amakhrov Thanks a lot for your PR. 
If I understand it correctly, the problem is not a bug itself, but rather a naming inconvenience?

amakhrov:
@macjohnny i guess it depends on a point of view.
I consider it a bug, because it makes modelSuffix rather unusable in typescript-angular, resulting in enum names that are called Models (assuming Model is the suffix used).

However, one can argue that since generator still emits some valid typescript, it's not a bug but a feature.

If you think it's rather a feature than a bugfix - do you think it's still valuable to be merged into the next major version?

amakhrov:
Thinking further about introducing backward incompatible changes.... Perhaps there is a way to keep compatibility by introducing a special `v4-compat` value of  this new enumNameSuffix option, and making it a default. This value will enforce the existing behavior for all generators. This value can then be removed in the major release if needed.
Any thoughts?

macjohnny:
@amakhrov I think it is a good idea to have a backwards-compatible behavior by default, so we can merge it and include it in the next patch release

macjohnny:
concerning bug or feature, since the code compiles as it is generated now, i would consider it a feature/improvement of the naming, which is, of course, favorable, too.

amakhrov:
All right, so I'll make the change described (introduce a special `v4-compat` value for this option), and update the PR - all tomorrow (it's past midnight over here already)

macjohnny:
@wing328 can we include this in the 4.2.3 release? it will be ready in a few minutes, just waiting for the last ci build to complete

florianstuerzlinger:
Is it possible to not add a suffix to enums?
i tried calling the generator with an empty parameter - <enumNameSuffix></enumNameSuffix>
This leads to the suffix "False". Any solution for this?

PR #5152, opened 2020-01-29
[codegen] Performance optimizations: use of cache for camel case and sanitized strings

sebastien-rosset:
While running codegen under a Java profiler, I've noticed more than 80% of the time is spent in a very small number of functions that can be optimized with a cache:
* sanitize words
* underscore words
* camelize words

With a medium to large OpenAPI spec, I see the same function is called with the same arguments tens of millions of time, so there is a great opportunity to optimize performance.

This PR is purely a performance optimization, it does not change any output, as can be verified with no changes in the samples.

### PR checklist


sebastien-rosset:
There seems to be an unrelated build issue.

wing328:
Tested locally with a large spec and did see some improvements. Tested with logger to confirm hits to the cache.

jimschubert:
This should be moved from a Map to caffeine cache to automatically expire records, otherwise OpenAPI Generator will be unusable in hosted environments (such as our online generator).

PR #5154, opened 2020-01-29
[Python] Python HTTP signature update

sebastien-rosset:
This PR is a follow up to the initial commit for HTTP signature support, based on additional testing and review.

* The OAuth2 "access_token" value should be null by default, not empty string. Otherwise, the client will attempt to set the "Authorization" header with an empty bearer token.
* Added more documentation.
* Conditionally generate the examples in the Configuration() object, based on what security schemes are defined in the OAS spec.
* In the Configuration examples, add snippet of OAS spec with a security scheme to better explain the use cases.
* Add missing import 
* Fix typo

### PR checklist


sebastien-rosset:
Build issue:
```
Execution failed for task ':compileKotlin'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not download jackson-coreutils.jar (com.github.fge:jackson-coreutils:1.8)
      > Could not get resource 'https://jcenter.bintray.com/com/github/fge/jackson-coreutils/1.8/jackson-coreutils-1.8.jar'.
         > Could not GET 'https://repo.jfrog.org/artifactory/libs-release-bintray/com/github/fge/jackson-coreutils/1.8/jackson-coreutils-1.8.jar?referrer'.
            > sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

File: modules/openapi-generator/src/main/resources/python/configuration.mustache [hunk 1]
```diff
@@ -123,19 +152,21 @@ class Configuration(object):
...
+{{#hasHttpSignatureMethods}}
         if signing_info is not None:
             signing_info.host = host
         self.signing_info = signing_info
         """The HTTP signing configuration
         """
+{{/hasHttpSignatureMethods}}
 {{#hasOAuthMethods}}
-        self.access_token = ""
+        self.access_token = None
```
spacether: This changes the default behavior of auth_settings.
Before the default of empty string would have the auth_settings() return a dict containing:
```
{
  ...
  'value': 'Bearer ',
  ...
}
```
Whereas this change would omit that entry.
Because this is a breaking change with fallback (users tests could break)
We can either:
- target the 4.3.0 (breaking change with fallback)
- or revert the access_token back to empty string in this PR and release it on master
What do you want to do here?
Did you have any discussions with @wing328 about this?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -123,19 +152,21 @@ class Configuration(object):
...
         """access token for OAuth/Bearer
         """
 {{/hasOAuthMethods}}
 {{^hasOAuthMethods}}
 {{#hasBearerMethods}}
-        self.access_token = ""
+        self.access_token = None
```
spacether: The above comment about targeting a different release or changing this back to empty string applies here.

modules/openapi-generator/src/main/resources/python/configuration.mustache
sebastien-rosset: But I argue this is a bug, just like password should not be the empty string by default. If a client actually wants to use a bearer token or OAuth2 access token, it will have to set the "access_token" attribute to a non-empty value.

wing328: > But I argue this is a bug, just like password should not be the empty string by default. If a client actually wants to use a bearer token or OAuth2 access token, it will have to set the "access_token" attribute to a non-empty value.

I think that's what we do in other languages (e.g. PHP, Ruby) as empty string could be a possible value (e.g. password in HTTP basic auth but of course it's against security best practice) so it should not be set as the default.

wing328: > target the 4.3.0 (breaking change with fallback)

Current master is (4.3.0), which is scheduled by the end of this month (Feb)

sebastien-rosset: So can I mark this conversation resolved?

spacether: Yes, feel free to mark this as resolved

Issue #5117, opened 2020-01-26
[BUG] Need more robust handling of URL operation 'path' segment character encoding


sebastien-rosset:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The generated Python code has problems when the operation "path" segment includes every possible allowed character per RFC 3986, but I suspect this might be a more general problem. I also noticed the OAS spec 3.0.2 is ambiguous about the "path" encoding, so I opened https://github.com/OAI/OpenAPI-Specification/issues/2119.

I think the OpenAPITool makes the assumption that URL paths in the OAS spec are encoded according to RFC 3986 section 3.3, though this may be by accident, not intentional. When I look at the Python implementation, I see the "query" part is URL encoded, but not the "path" part of the URL.  

A path segment (the parts in a path separated by /) in an absolute URI path can contain zero or more of pchar that is defined as follows:

```
  pchar       = unreserved / pct-encoded / sub-delims / ":" / "@"
  pct-encoded = "%" HEXDIG HEXDIG
  unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
  sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
              / "*" / "+" / "," / ";" / "="
```

This means this is a valid "path" segment:

```
/store/orderAZa–z09-._~!$&'(abc)*+ , ;=:@%C5%92%C3%AB%F0%9F%8D%87%E2%84%AC:
```

##### openapi-generator version

master, as of January 26th, 2020

sebastien-rosset:
I see AbstractJavaCodeGen modifies the path with a  call to sanitizePath(), but it's done in a very limited way (only the double quotes are changed). IMO, this should be done in a canonicalized way across all languages.

sebastien-rosset:
[...]

PR #5158, opened 2020-01-29
[c-libcurl]: add support for url encoding

zhemant:
Added support for url encoding (present only in the query params)

### PR checklist

@wing328  @ityuhui

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/build.gradle.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/build.gradle.mustache
@@ -35,9 +35,9 @@ test {
 
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
-    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
     {{#moshi}}
     {{^moshiCodeGen}}
+    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
     compile "com.squareup.moshi:moshi-kotlin:1.9.2"
     {{/moshiCodeGen}}
     compile "com.squareup.moshi:moshi-adapters:1.9.2"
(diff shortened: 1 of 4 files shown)


sebastien-rosset:
Related to this, I have opened https://github.com/OAI/OpenAPI-Specification/issues/2119 and #5117. I think it would be great to address URL path and query encoding in a consistent way across languages.

zhemant:
For URL path encoding, I believe it should have support right from the library instead of reinventing the whole encoding wheel again. I agree with the example you have given. But with libcurl I tried and it encodes "/" also this creates the problem in the server where the handlers do not recognize the URLs as they are encoded. Let me know if you get any update on encodings.

sebastien-rosset:
> For URL path encoding, I believe it should have support right from the library instead of reinventing the whole encoding wheel again.

Yes, I completely agree we should not reinvent the wheel. By library, do you mean codegen? Per-language encoding under modules/openapi-generator/src/main/java/org/openapitools/codegen/languages? A mustache template invoking a language specific library? I think we may need a combination, but ideally as much as possible would be pushed to codegen.

I think codegen should be able to do some sanitization even before the 

> I agree with the example you have given. But with libcurl I tried and it encodes "/" also this creates the problem in the server where the handlers do not recognize the URLs as they are encoded. Let me know if you get any update on encodings.

Yes, I had a similar problem with golang. The standard golang library performs incorrect path encoding, and I had to use a different library that carefully encodes the URL path.

zhemant:
By library I mean for C client, we use the curl library to send the request. Curl library provides URL encoding functionality. I tried using it for the whole path but it creates more problem.  and CURL is a standard library so my guessing is it is updated as per the required standards(what to encode and what not to encode).

modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
```diff
@@ -181,14 +181,20 @@ char *assembleTargetUrl(char    *basePath,
...
         strcat(targetUrl, "?");
         list_ForEach(listEntry, queryParameters) {
             keyValuePair_t *pair = listEntry->data;
             replaceSpaceWithPlus(pair->key);
             strcat(targetUrl, pair->key);
             strcat(targetUrl, "=");
-            replaceSpaceWithPlus(pair->value);
-            strcat(targetUrl, pair->value);
+            if(curl) {
+              char *output = curl_easy_escape(curl, pair->value, 0);
```
sebastien-rosset: PATH   : Http://SomeUrl.com:8080//store/curlybraces-%7B%7D/AZaz-09._~!$&'(abc)*+, ;=:@Œë🍇ℬ/order/template-{orderId}/.././c///g?c=3&a=1&b=9&c=0#target

This is how the above URL is supposed to be converted. Note characters like :, /, $ ! are not percent encoded.

ESCAPED: http://someurl.com:8080//store/curlybraces-%7B%7D/AZaz-09._~!$&'(abc)*+,%20;=:@%C5%92%C3%AB%F0%9F%8D%87%E2%84%AC/order/template-{orderId}/.././c///g?c=3&a=1&b=9&c=0#target

Using curl_easy_escape, I get a lot more character escapes:
Http%3A%2F%2FSomeUrl.com%3A8080%2F%2Fstore%2Fcurlybraces-%257B%257D%2FAZaz-09._~%21%24%26%27%28abc%29%2A%2B%2C%20%3B%3D%3A%40%C5%92%C3%AB%F0%9F%8D%87%E2%84%AC%2Forder%2Ftemplate-%7BorderId%7D%2F..%2F.%2Fc%2F%2F%2Fg%3Fc%3D3%26a%3D1%26b%3D9%26c%3D0%23target

PR #5162, opened 2020-01-30
[typescript] Introduce "enumPropertyNaming" option

amakhrov:
This works similar to the existing "modelPropertyNaming", but applies to enum var names instead of model properties.
Default value is "PascalCase", which reflects the way it works currently.

Personally, I only need this configurable in the typescript-angular generator now. However, since the enum var name transformation is implemented in the abstract typescript class, I thought it would be beneficial to keep consistency across multiple typescript generators. After all, this is quite a generic setting, not specific to a particular framework.

### PR checklist

### DIFF
--- a/docs/generators/javascript-flowtyped.md
+++ b/docs/generators/javascript-flowtyped.md
@@ -7,6 +7,7 @@ sidebar_label: javascript-flowtyped
 | ------ | ----------- | ------ | ------- |
 |allowUnicodeIdentifiers|boolean, toggles whether unicode identifiers are allowed in names or not, default is false| |false|
 |ensureUniqueParams|Whether to ensure parameter names are unique in an operation (rename parameters that are not).| |true|
+|enumPropertyNaming|Naming convention for enum properties: 'camelCase', 'PascalCase', 'snake_case', 'UPPERCASE', and 'original'| |PascalCase|
 |modelPropertyNaming|Naming convention for the property: 'camelCase', 'PascalCase', 'snake_case' and 'original', which keeps the original name| |camelCase|
 |npmName|The name under which you want to publish generated npm package. Required to generate a full package| |null|
 |npmRepository|Use this property to set an url your private npmRepo in the package.json| |null|
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -37,6 +37,7 @@ import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.openapitools.codegen.utils.StringUtils.camelize;
 import static org.openapitools.codegen.utils.StringUtils.underscore;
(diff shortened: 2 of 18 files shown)


docs/generators/javascript-flowtyped.md
```diff
@@ -7,6 +7,7 @@ sidebar_label: javascript-flowtyped
 | ------ | ----------- | ------ | ------- |
 |allowUnicodeIdentifiers|boolean, toggles whether unicode identifiers are allowed in names or not, default is false| |false|
 |ensureUniqueParams|Whether to ensure parameter names are unique in an operation (rename parameters that are not).| |true|
+|enumPropertyNaming|Naming convention for enum properties: 'camelCase', 'PascalCase', 'snake_case', 'UPPERCASE', and 'original'| |PascalCase|
```
macjohnny: why is the javascript-flowtyped affected?

amakhrov: I was surprised also.
But apparently,
```
class JavascriptFlowtypedClientCodegen extends AbstractTypeScriptClientCodegen
```

macjohnny: ah, ok alright

macjohnny:
note: instead of fake commits, you can also close and re-open the PR to trigger CI

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -588,6 +592,37 @@ public String toEnumName(CodegenProperty property) {
         }
     }
 
+    protected void setEnumPropertyNaming(String naming) {
+        try {
+            enumPropertyNaming = ENUM_PROPERTY_NAMING_TYPE.valueOf(naming);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Invalid enum property naming '" +
+                    naming + "'. Must be 'original', 'camelCase', " +
+                    "'PascalCase', 'snake_case', or 'UPPERCASE'");
```
macjohnny: maybe the available values in `ENUM_PROPERTY_NAMING_TYPE` could be listed automatically, so the comment will not be outdated

modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularClientCodegenTest.java
```diff
@@ -16,6 +17,22 @@
...
+    @Test
+    public void toEnumVarName() {
+        TypeScriptAngularClientCodegen codegen = new TypeScriptAngularClientCodegen();
+        codegen.processOpts();
+
+        Assert.assertEquals(codegen.toEnumVarName("valid_id", "string"), "ValidId");
+        Assert.assertEquals(codegen.toEnumVarName("illegal-id+", "string"), "IllegalId");
+
+        codegen = new TypeScriptAngularClientCodegen();
+        codegen.additionalProperties().put(CodegenConstants.ENUM_PROPERTY_NAMING, CodegenConstants.ENUM_PROPERTY_NAMING_TYPE.original.name());
```
macjohnny: maybe you would want to add a test for the non-original case, too?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
amakhrov: Good idea, will do!

macjohnny:
there seems to be some issue with the option parsing, see https://app.shippable.com/github/OpenAPITools/openapi-generator/runs/14327/1/console

amakhrov:
> there seems to be some issue with the option parsing

Looks like Kotlin build is broken - no idea why. Rebasing my branch from the latest master to see if it's gonna self-resolve

modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularClientCodegenTest.java
amakhrov: Done

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
amakhrov: and done

amakhrov:
The last CI step still fails with a kotlin-related error.
I can see that some other open PRs also fail with the same error. Is it a global issue with configuration / 3rd party service?

PR #5175, opened 2020-01-31
[go-experimental] Add model constructors to initialize non-container vars with defaults

bkabrda:
This PR adds a constructor to all models that initializes non-container vars with proper defaults. It might also be possible to do container vars in the future, but that is pretty tricky (and also not as widely used to my knowledge).

# PR checklist


modules/openapi-generator/src/main/resources/go-experimental/model.mustache
```diff
@@ -38,6 +38,19 @@ type {{classname}} struct {
 	{{name}} {{^required}}*{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
 {{/vars}}
 }
+
```
jirikuncar: can you add a docstring?

platinummonkey: Any reason not to require the required Params in the input?

bkabrda: @jirikuncar thanks for pointing out the docs I'll fix that.

@platinummonkey I think the problem here is that it's not straightforward what to do with required params that have defaults set. Should these be required arguments for the constructor or not? I think that this uncertainty is why generated constructors in other similar languages don't have required params either. I'd like to stay consistent here, so that's why I chose this approach.

bkabrda: I pushed a new commit that adds both the docstring and also documentation in `model_doc`.

platinummonkey: `New(args....)` vs `NewDefault()` maybe is more clear?

samples/client/petstore/go-experimental/go-petstore/model_animal.go
```diff
@@ -20,6 +20,13 @@ type Animal struct {
 	Color *string `json:"color,omitempty"`
 }
 
+func NewAnimal() *Animal {
+    this := Animal{}
+    var Color string = "red"
```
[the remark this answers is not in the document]
bkabrda: Yeah, I just used the name of the var. I think there might be a camelCase version of this somewhere in the rendering context, let me try to find and use it.

modules/openapi-generator/src/main/resources/go-experimental/model.mustache
bkabrda: @platinummonkey I just realized one more reason that is problematic with the required properties:

Imagine that you have required properties A and B and code generated for that: `NewFoo(A, B)`. Now a change happens that makes `B` optional. You regenerate the code and suddenly you get `NewFoo(A)` - so that's backwards incompatible. But the change you did in API *is* backwards compatible, so this breaks the promise of not breaking API of the generated client when a non-breaking API change happens.

samples/client/petstore/go-experimental/go-petstore/model_animal.go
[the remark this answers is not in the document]
bkabrda: Fixed.

modules/openapi-generator/src/main/resources/go-experimental/model.mustache
platinummonkey: I'm not actually sure that's a problem, if you need to introduce a var, this this method signature is also changed. The alternative is to list all variables, and if one becomes non-optional though or a new one is added then still the signature is changed. Seem moot given go's syntax.

bkabrda: The signature of this method will never change, as it never accepts any arguments. But I think I'm starting to like the solution proposed in your previous comment with `New(required, args, ...)` and `NewWithDefaults()`. Assuming we clearly document that the contract of the former will always change when the set of required arguments changes, it should be ok. I'll work on this.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientExperimentalCodegen.java
```diff
@@ -122,7 +130,11 @@ public String toDefaultValue(Schema p) {
     @Override
     public CodegenProperty fromProperty(String name, Schema p) {
         CodegenProperty prop = super.fromProperty(name, p);
-        prop.nameInCamelCase = camelize(prop.name, true);
+        String cc = camelize(prop.name, true);
+        if (isReservedWord(cc)) {
+            cc = escapeReservedWord(cc);
```
jirikuncar: 👍

bkabrda:
For anyone reviewing, please hold merging this until https://github.com/OpenAPITools/openapi-generator/pull/5150 is merged. Thanks!

bkabrda:
Ok, rebased and ready to merge now.

bkabrda:
FWIW, bitrise failed because of an unrelated error (at least that's how I understand it):

```
==> Installing maven
Error: An exception occurred within a child process:
  DownloadError: Failed to download resource "maven"
Download failed: Couldn't determine mirror, try again later.
Can't install formulas:  exit status 1
|                                                                              |
+---+---------------------------------------------------------------+----------+
| x | brew-install@0.10.2 (exit code: 1)                            | 2.1 min  |
+---+---------------------------------------------------------------+----------+
| Issue tracker: https://github.com/bitrise-steplib/steps-brew-install/issues  |
| Source: https://github.com/bitrise-steplib/steps-brew-install                |
+---+---------------------------------------------------------------+----------+
```

PR #5181, opened 2020-01-31
[gradle] Add recommendations to validate task

### LINKED ISSUE #335: Add validate --recommend option to maven and gradle plugin options.
##### Description

In #292, I implemented a `--recommend` option for the CLI's validate command. We'll want to add a similar option to the Maven and Gradle plugins.

I'm thinking that we could create a validator result object (e.g. `ValidationSummary`, with properties exposing a set of `errors` and a set of `warnings`) and a validator utility (e.g. `SpecificationValidator`), in which we encapsulate the logic for returning errors/warnings. The CLI command (and probably Maven and Gradle plugins) output a stringified result of errors and warnings, which would be a static utility method on the validator utility type. This would remove any inconsistencies between the tools and would allow other tools (like my IntelliJ plugin) consider "errors" the same as we consider them in openapi-generator; right now it's the same as swagger-parser, but this may not always be the case.

3.0.2+

##### Related issues/PRs

* #229

jimschubert:
Adds recommendation/validation to the validate task in the gradle plugin.
This fixes #335 although there's no addition to the maven plugin, because none of CLI/Gradle/Maven do custom recommendations or validation in the generate task.

### PR checklist

### DIFF
--- a/modules/openapi-generator-gradle-plugin/README.adoc
+++ b/modules/openapi-generator-gradle-plugin/README.adoc
@@ -377,6 +377,11 @@ openApiGenerate {
 |None
 |The input specification to validate. Supports all formats supported by the Parser.
 
+|recommend
+|Boolean
+|true
+|Whether or not to offer recommendations related to the validated specification document.
+
 |===
 
 === openApiMeta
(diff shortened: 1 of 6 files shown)


jimschubert:

jimschubert:
Caches on CI appear to have gotten messed up with this PR. I've opened another to retry.

Issue #4770, opened 2019-12-11
[BUG][typescript-angular] allOf not working without additional properties


joke:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

The typescript generator generators does not include attributes without specifying additional properties:
- The interface `PetListEntry` is empty and does not extend `Pet`

```typescript
export interface PetNew { 
    name: string;
}

export interface Pet { 
    id: string;
    name: string;
}

export interface PetListEntry { 
}
```

##### openapi-generator version
4.2.2

##### OpenAPI declaration file content

```yaml
openapi: 3.0.2
info:
  title: api
  version: 1.0.0
paths:
  /pets:
    get:
      summary: list all pets
      responses:
        200:
          description: An array of pets
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/PetListEntry'
    post:
      summary: create new pet
      requestBody:
        description: Pet to be created
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/PetNew'
      responses:
        201:
          description: Newly created pet
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'
components:
  schemas:
    Pet:
      allOf:
        - type: object
          required:
            - id
          properties:
            id:
              type: string
        - $ref: '#/components/schemas/PetNew'
    PetListEntry:
      allOf:
        - $ref: '#/components/schemas/Pet'
    PetNew:
      type: object
      required:
        - name
      properties:
        name:
          type: string
```

##### Command line used for generation

```bash
$ openapi-generator-cli-4.2.2.jar generate -g typescript-angular -i api.yaml
```

##### Steps to reproduce

- Generate the model files
- Check model files

digaus:
Can confirm this... Fix would be nice

PR #5182, opened 2020-01-31
[core][typescript][perl] Ensure model.parent is also added to model.allParents with multiple inheritance

amakhrov:
Fixes https://github.com/OpenAPITools/openapi-generator/issues/4770

Turned out that generator still supports a deprecated inheritance case with a single oneOf ref without a discriminator. The behavior in this case is to treat it as inheritance rather than composition.

Now, the problem is that even though `model.parent` is properly populated with the referenced model in this situation, `model.allParents` (**used with multiple inheritance**) stays empty. It looks like a bug in the implementation rather than intended behavior - but of course correct me if I'm wrong.

This PR makes sure the same logic supporting the deprecated inheritance case applies both to `parent` and `allParents`.

I also cleaned up a few tests for DefaultCodegen that are related to `allOf` composition - just to give me a bit more safety assurance of the change (better than nothing)

### PR checklist

Typescript generators family use multiple inheritance, hence copying them here:

And also Perl (another representative of multiple inheritance supporting generator):
@wing328  @yue9944882

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -972,7 +972,8 @@ public class ModelUtils {
     public static String getParentName(ComposedSchema composedSchema, Map<String, Schema> allSchemas) {
         List<Schema> interfaces = getInterfaces(composedSchema);
 
-        List<String> refedParentNames = new ArrayList<>();
+        List<String> refedWithoutDiscriminator = new ArrayList<>();
+
         if (interfaces != null && !interfaces.isEmpty()) {
             for (Schema schema : interfaces) {
                 // get the actual schema
@@ -986,9 +987,8 @@ public class ModelUtils {
                         // discriminator.propertyName is used
                         return parentName;
                     } else {
-                        LOGGER.debug("Not a parent since discriminator.propertyName is not set {}", s.get$ref());
                         // not a parent since discriminator.propertyName is not set
-                        refedParentNames.add(parentName);
+                        refedWithoutDiscriminator.add(parentName);
                     }
                 } else {
                     // not a ref, doing nothing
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
```diff
@@ -528,9 +528,31 @@ public void testAllOfRequired() {
         Schema child = openAPI.getComponents().getSchemas().get("clubForCreation");
         codegen.setOpenAPI(openAPI);
         CodegenModel childModel = codegen.fromModel("clubForCreation", child);
-        showVars(childModel);
+        Assert.assertEquals(getRequiredVars(childModel), Collections.singletonList("name"));
```
amakhrov: now sure how we ended up here having a test without any assertions :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -996,13 +997,20 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
                     } else {
                         LOGGER.debug("Not a parent since discriminator.propertyName is not set {}", s.get$ref());
                         // not a parent since discriminator.propertyName is not set
+                        refedParentNames.add(parentName);
```
macjohnny: is the comment above still valid?

amakhrov: Yes it is. Also note the same existing comment in line 961 above.

For some reason, which is not clear to me, openapi generator treats this scenario (oneOf with a single $ref, where the referenced schema has no discriminator) as inheritance, despite the lack of discriminator (which would normally mean composition and not inheritance).

My fix doesn't change this behavior. It just makes it (i believe) more correct

amakhrov: actually, no. you're right, I need to clean this up.
first off, the comment duplicates the log message - obvious redundancy.
second, I think it should be exactly same message as in line 961 (currently it's kind of similar, but with diffirent flavors)

amakhrov: @macjohnny sorry, I'm obviously not thinking clear tonight. Should just stop answering till tomorrow.
Generally, it looks like the same comments and messages are used in both places (here and inside `getParentName()` above). Which kind of is the intention. I'll double check tomorrow to make sure it still makes sense

amakhrov: I believe the comment here is still very precise.
However, I removed the LOGGER.debug above that, which just repeated that comment.
Also renamed the variable to better reflect the intention. Hope it makes more sense now

wing328:
@amakhrov thanks for the PR, which looks good to me. Can you please resolve the merge conflicts when you've time?

amakhrov:
@wing328 rebased from the latest `master`

wing328:

PR #5184, opened 2020-02-01
[codegen] Add code comments and improve "first content" warning

sebastien-rosset:
When the OAS spec has paths with multiple content types, the following obscure warning is logged:
o.o.codegen.utils.ModelUtils - Multiple schemas found in content, returning only the first one

I have added code comments and better log message.
o.o.codegen.utils.ModelUtils - Multiple schemas found in the OAS content section, returning only the first one (‘application/json’)

As a possible enhancement, I suggest supporting a configuration property to specify the desired content type. Sometimes it’s not possible to reorder the OAS spec because you are the consumer of the spec, not the author.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -814,12 +814,11 @@ public class ModelUtils {
             return null;
         }
         Map.Entry<String, MediaType> entry = content.entrySet().iterator().next();
-        MediaType mediaType = content.values().iterator().next();
         if (content.size() > 1) {
             // Other content types are currently ignored by codegen. If you see this warning,
             // reorder the OAS spec to put the desired content type first.
-            LOGGER.warn("Multiple schemas found in the OAS 'content' section, returning only the first one ('" +
-                entry.getKey() + "')");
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -791,15 +791,37 @@ public static Schema getSchemaFromResponse(ApiResponse response) {
...
         if (content == null || content.isEmpty()) {
             return null;
         }
+        Map.Entry<String, MediaType> entry = content.entrySet().iterator().next();
+        MediaType mediaType = content.values().iterator().next();
         if (content.size() > 1) {
-            LOGGER.warn("Multiple schemas found in content, returning only the first one");
+            // Other content types are currently ignored by codegen. If you see this warning,
+            // reorder the OAS spec to put the desired content type first.
+            LOGGER.warn("Multiple schemas found in the OAS 'content' section, returning only the first one ('" +
```
sebastien-rosset: This type of warning tends to be buried among other informational log messages, but it is crucial to how codegen processes the OAS.

As a suggested future enhancement, the tool should generate a list of warnings that can be output to a structured format, such as JSON. The very last log message should be a summary of the warnings, with a way to get the list of warnings.

PR #5191, opened 2020-02-02
[bug] fix null pointer exception while evaluating recommend…

sebastien-rosset:
A null pointer exception occurs when invoking --validate --recommend if at least one of the parameters in the OAS spec is a $ref.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/validations/oas/OpenApiEvaluator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/validations/oas/OpenApiEvaluator.java
@@ -87,7 +87,10 @@ public class OpenApiEvaluator implements Validator<OpenAPI> {
             }
         }
 
-        parameters.forEach(parameter -> validationResult.consume(parameterValidations.validate(parameter)));
+        parameters.forEach(parameter -> {
+            parameter = ModelUtils.getReferencedParameter(specification, parameter);
+            validationResult.consume(parameterValidations.validate(parameter));
+        });
 
         return validationResult;
     }


sebastien-rosset:
@jimschubert , I would like to be able to do a validation based on the OpenAPI version (3.0.x versus 3.1). This is bleeding edge, but there are some nice capabilities coming in 3.1, including 'null' type and type arrays. So the recommendation needs to depend on whether the version is 3.0 or 3.1. Should we make the "OpenAPI" object available to the validators? If so, how?

jimschubert:
@sebastien-rosset that would require updating our parser to one that supports 3.1. Currently, the object model that we pull in is only the 3.0.2 compatible object model. I'd eventually like to have our own parser so we can do the bleeding edge stuff, but that requires getting caught up on a better and maintainable design first.

sebastien-rosset:
> @sebastien-rosset that would require updating our parser to one that supports 3.1. Currently, the object model that we pull in is only the 3.0.2 compatible object model. I'd eventually like to have our own parser so we can do the bleeding edge stuff, but that requires getting caught up on a better and maintainable design first.

Actually, I've been experimenting a bit with the "null" type, which is written as:
```yaml
type: 'null'
```
I don't think this requires a new parser because the value of the "type" attribute is still a string. At least I am able to do the things I would like to do. This may be a different story for type arrays, but I'm not there yet.
Now that "nullable" is becoming deprecated on OAS (see my long text on nullable properties), I think I'd rather work on using "type: null" in OneOf that going with something that will be deprecated soon. I also realized it's ok to use OneOf and avoid ugly code because it's possible to control the names of the generated code.

PR #5192, opened 2020-02-02
[core] Normalizing vendor extension naming

jimschubert:
According to [OAS 2.0][1] and [OAS 3.0][2] Specifications:

> Allows extensions to the OpenAPI Schema. The field name MUST begin with x-,
>  for example, x-internal-id. The value can be null, a primitive, an array or an object.
>  Can have any valid JSON format value.

This commit attempts to define a [clear identifier design format][3] of maintaining lower-kebab casing and following the x- prefix defined by OAI Specification.

Following a convention that matches that used by others (see [autorest][4]), we will remove any confusion about naming strategies for template authors and customizers. Following the lower-kebab convention will allow us to convert from camelCase and missing prefixes to the desired format. For example, these conversions are simple to make for template consistency:

* customValue => x-custom-value
* x-customValue => x-custom-value
* x-custom-value => x-custom-value

This convention also allows us to define a single standard for use across all generators. This means no occurrence of `x-operationId` in one generator and `x-operation-id` in another.

[1]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-objects
[2]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#specificationExtensions
[3]: https://tools.ietf.org/html/draft-wilde-registries-01#section-3.4
[4]: https://github.com/Azure/autorest/tree/master/docs/extensions

Included in this PR is a logger which will log a warning message only once. This could be used for ModelUtils and UrlUtils to reduce stdout/stderr noise. The `LoggerWrapper` implementation is supported by an in-memory (caffeine) cache. I've added support for the following system properties so consumers with more advanced use cases can tweak/disable the caching OnceLogger:

* `org.openapitools.codegen.utils.oncelogger.enabled`: set to `false` to disable the caching OnceLogger. Set to `true` to enable (the default).
* `org.openapitools.codegen.utils.oncelogger.cachesize`: modify the cache size from the default of 200 messages (this is log message, not message count)
* `org.openapitools.codegen.utils.oncelogger.expiry`: set the expiration time in millis to control how long messages remain in-memory. The default is 2000 (2 seconds), but could be made lower for high traffic consumers. For one-off use cases such as CLI, Maven Plugin, and Gradle Plugin, changing this property may not be useful.

**NOTE** This will require additional template work and removal of older extensions in the 5.0 branch. We'll need document the change(s) and upgrade path for the 5.0 release as well.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java
@@ -2,6 +2,7 @@ package org.openapitools.codegen.utils;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Ticker;
 import org.openapitools.codegen.config.GlobalSettings;
 import org.slf4j.Logger;
 import org.slf4j.Marker;
@@ -77,11 +78,16 @@ public class OnceLogger extends LoggerWrapper {
     }
 
     static {
+        caffeineCache(Ticker.systemTicker(), expireMillis);
+    }
+
+    static void caffeineCache(Ticker ticker, int expireMillis) {
         // Initializes a cache which holds an atomic counter of log message instances.
         // The intent is to debounce log messages such that they occur at most [maxRepetitions] per [expireMillis].
         messageCountCache = Caffeine.newBuilder()
                 .maximumSize(maxCacheSize)
                 .expireAfterWrite(expireMillis, TimeUnit.MILLISECONDS)
+                .ticker(ticker)
                 .build();
     }
 
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java
@@ -9,6 +9,7 @@ import org.slf4j.MarkerFactory;
 import org.slf4j.ext.LoggerWrapper;
 
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Provides calling code a way to log important messages only once, regardless of how many times the invocation has occurred.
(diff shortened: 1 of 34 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java
```diff
@@ -0,0 +1,162 @@
...
+     * @param msg
+     */
+    @Override
+    public void trace(String msg) {
+        if (!isTraceEnabled() || !isTraceEnabled(MARKER)) return;
+
+        if (shouldLog(msg)) super.trace(MARKER, msg);
+    }
+
+    private boolean shouldLog(final String msg) {
```
ben-manes: maybe a single operation if you want to avoid races?
`messageCountCache.asMap().merge(msg, 0, count -> count + 1)`

or an atomic counter if you want to avoid locking except on the initial mapping?

jimschubert: Thanks for the quick and helpful review, Ben! I decided to go the `AtomicInteger` route because it's not critical functionality for this method and doesn't need repeated locking, I'm just trying to remove noise for users.

ben-manes: I think that's best choice, too! 

However, it no longer writes into the map on every `shouldLog` call and the `expireAfterWrite` evicts based on the entry's creation time. The old behavior was to reset the expiration each time the method was called, so it requires no occurrences for the expiration duration. That could be done now using `expireAfterAccess` to reset on every read or write, which is fine as this is the sole usage.

I suspect you don't want to suppress an ongoing error for good once the threshold is reached, as that makes it appear to be over, and simply want to debounce over a time interval. This new expiration behavior might be more correct, but I'm unsure. You probably want to think a bit about exactly what expiration behavior you want and maybe write a quick test to verify that.

jimschubert: You're correct that my intent was to debounce over a time internal. I think I mistakenly took `expireAfterWrite` to mean initial writes only, but looking at it after your comment I see that understanding wouldn't have made sense if I'd looked at the method a little more closely.

I'll definitely need to add a comment about intention on the cache member, and write a handful of tests. I'll have to think about usage for errors and other levels a little more. The way I wrote this, it'll require that someone explicitly decorates a Logger instance and I may just throw an error if it's used to log error messages, or just skip the cache evaluation for that method.

PR #5207, opened 2020-02-04
[typescript] Make additional properties access safer

asmundg:
Instead of asserting that any key access returns a valid property, force
the consumer to check that the value is defined.

When declaring an object with additional properties

```yaml
  /store/inventory:
    get:
      tags:
        - store
      summary: Returns pet inventories by status
      description: Returns a map of status codes to quantities
      operationId: getInventory
      produces:
        - application/json
      parameters: []
      responses:
        '200':
          schema:
            type: object
            additionalProperties:
              type: integer
              format: int32
```

The generated interface is unsafe, because it asserts that _any_ property access on the response returns a value. In reality, a property access can return undefined, which will cause a runtime failure.

```ts
async getInventory(): Promise<{ [key: string]: number; }
```

If the compiler is running with strict null checking, we can avoid the problem by declaring that property access _can_ return undefined, forcing the consumer to verity that they got an actual value:

```ts
async getInventory(): Promise<{ [key: string]: number | undefined; }>
```

### PR checklist

cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig @topce @akehir @petejohansonxo


amakhrov:
@asmundg thanks for your PR!

Unfortunately this is a core issue in Typescript, which is currently not resolvable. E.g. see https://github.com/microsoft/TypeScript/issues/23817 (which has a number of links to many related issues of the same kind).

A problem is that defining the keys this way (`| undefined`) would negatively impact using this object in the contexts that only deal with existing object keys (`for .. of`, `Object.values()`, etc).

I don't believe it should be addressed in codegen at all. I would rather wait for the sufficient resolution in the Typescript compiler, or some community-wide workaround for that.

Any thoughts?

asmundg:
Thanks @amakhrov,

You're right that this is a fundamental typescript problem. Changing this behaviour in codegen would be a breaking change.

While the current behaviour is fine if you always access the additional properties through Object.keys(), it's very bug prone if you're trying to access specific properties directly.

As we do have a mechanism for expressing that an object can have undefined properties, I'm not convinced that we unconditionally need to accept reduced safety of typescript's default behaviour.

Would it make sense to make the modified behaviour available behind a flag, so consumers can choose whether to enable stricter types (at the cost of superfluous checking when iterating over object properties)?

amakhrov:
@asmundg introducing a flag for that sounds like a good idea. I believe many users can find the new behavior helpful - but probably not all of them.

Another consideration: this change doesn't have anything specific to the `-fetch` generator alone, right? Perhaps this logic (together with the flag driving it) should be defined in AbstractTypeScript codegen instead?

asmundg:
@amakhrov good point about AbstractTypeScript. I'll take a look at making this shared and behind a flag.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java
```diff
@@ -251,6 +251,9 @@
     public static final String SUPPORTS_ES6 = "supportsES6";
     public static final String SUPPORTS_ES6_DESC = "Generate code that conforms to ES6.";
 
+    public static final String NULL_SAFE_ADDITIONAL_PROPS = "nullSafeAdditionalProps";
+    public static final String NULL_SAFE_ADDITIONAL_PROPS_DESC = "Set to make additional properties types declare that their indexer may return undefined";
```
amakhrov: Suggest moving it to AbstractTypescript codegen. I believe the purpose of CodegenConstants is to store stuff that is applicable to many if not all generators. While the change introduced in this PR is very typescript-specific

asmundg: Gotcha. I was following the pattern of SUPPORTS_ES6 above, but it does seem a bit too broad.

amakhrov: I see! Yeah it looks inconsistent across the app. SUPPORTS_ES6 was introduced years ago, but some later addition (NPM_VERSION, etc) are made directly in AbstractTypescript, which makes more sense to me.

macjohnny:
@amakhrov I agree a unit test for this "small" variation should be enough

asmundg:
Thanks a lot for the feedback! This has been a bit of a monkey see, monkey do process on my part. :)

wing328:
@asmundg thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5209, opened 2020-02-05
[python] use datetime aware object

sebastien-rosset:
Add timezone information to datetime, otherwise the calculation of UNIX time may be incorrect, depending on the platform and local time configuration.

datetime.now() returns the current local date and time. If optional argument tz is None or not specified, this is like today(), but, if possible, supplies more precision than can be gotten from going through a time.time() timestamp (for example, this may be possible on platforms supplying the C gettimeofday() function).
This means the value of now() is not deterministic across platforms unless the timezone information is provided. As a result, on some platforms the "created" and "expires" fields are set incorrectly and cause the server side to conclude the signature is incorrect.

Also, need to handle python 2.7 and python 3.x.

### PR checklist


samples/openapi3/client/petstore/python-experimental/petstore_api/signing.py
```diff
@@ -270,15 +263,16 @@ def _get_signed_header_info(self, resource_path, method, headers, body, query_pa
...
 
-        # Get current time and generate RFC 1123 (HTTP/1.1) date/time string.
-        now = datetime.now()
-        stamp = mktime(now.timetuple())
-        cdate = formatdate(timeval=stamp, localtime=False, usegmt=True)
+        # Get UNIX time, e.g. seconds since epoch, not including leap seconds.
+        now = time()
+        # Format date per RFC 7231 section-7.1.1.2. An example is:
+        # Date: Wed, 21 Oct 2015 07:28:00 GMT
+        cdate = formatdate(timeval=now, localtime=False, usegmt=True)
```
spacether: Should we worry about a server using non-gmt time?

sebastien-rosset: Timezone would be for human visibility only, because what really matters is that we have an unambiguous time independent of the timezone, that both the client and server can agree on. For example, if the client is in PST time 9:00AM, it may be received at 9PM in Europe, but if the client always sends in UTC time and clearly specifies so, there is no ambiguity.

On the other hand, using timezones is not a good idea, because the client and servers may not have the exact set of timezones (since timezone information can change over time). There is less risk of mismatch if UTC is used.

That's what the RFC recommends. The RFC states the Date header can have one of three formats, but there is a preferred format, which must be in UTC. And unless I am mistaken, in RFC 7231 they use GMT for historic reasons, but it means UTC.

```
     Sun, 06 Nov 1994 08:49:37 GMT
```
Examples of the two obsolete formats are
```
     Sunday, 06-Nov-94 08:49:37 GMT   ; obsolete RFC 850 format
     Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
```

sebastien-rosset: > Should we worry about a server using non-gmt time/a specific time zone?

That would be ok. If the server is in PST time zone, and it receives the following date: Date: Wed, 21 Oct 2015 07:28:00 GMT

It will know unambiguously when the client sent the request (of course assuming the client is reasonably in sync with Internet time). It does not matter if the client is in Europe or Asia.

PR #5217, opened 2020-02-05
[csharp] - fix json encoding override on form params

milesressler:
See issue 5215 relating to encoding not supported on form parameters

@mandrean @jimschubert @frankyjuang @shibayan

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenParameter.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenParameter.java
@@ -22,7 +22,7 @@ import java.util.*;
 public class CodegenParameter implements IJsonSchemaValidationProperties {
     public boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
             isCookieParam, isBodyParam, hasMore, isContainer,
-            secondaryParam, isCollectionFormatMulti, isPrimitiveType, isModel, isExplode;
+            secondaryParam, isCollectionFormatMulti, isPrimitiveType, isModel, isExplode, jsonEncoding;
     public String baseName, paramName, dataType, datatypeWithEnum, dataFormat,
             collectionFormat, description, unescapedDescription, baseType, defaultValue, enumName, style;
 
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenParameter.java
```diff
@@ -22,7 +22,7 @@
 public class CodegenParameter implements IJsonSchemaValidationProperties {
     public boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
             isCookieParam, isBodyParam, hasMore, isContainer,
-            secondaryParam, isCollectionFormatMulti, isPrimitiveType, isModel, isExplode;
+            secondaryParam, isCollectionFormatMulti, isPrimitiveType, isModel, isExplode, jsonEncoding;
```
shibayan: In OAI 3.0, encoding can be specified besides `application/json`, so this method seems to have low extensibility.
This change will affect the entire openapi-generator and will require a maintainers review.

cc. @wing328 @jimschubert

jimschubert: I agree. If you need custom logic, I think a vendor extension would be more appropriate.

milesressler: Not quite sure I follow - from what I read about vendor extensions on [the docs] (https://swagger.io/docs/specification/openapi-extensions/), they are meant for _functionality that is not covered by the standard OpenAPI Specification_ .  This isn't custom logic, this is functionality that is part of the [OAI specification](https://swagger.io/docs/specification/describing-request-body/), and is currently ignored by the OAI generator.

I do agree that other encoding types can be specified and should be supported as well, but since the objects already include a ToJson(), it seemed like an easy win for those that need it.

wing328: @milesressler please follow `isJson` approach in Dart client generator: https://github.com/OpenAPITools/openapi-generator/blob/8d6286dfae0c37d53ce9d27cd00b8e8ef87bcfe1/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartJaguarClientCodegen.java#L272

Let me know if you need help. You can PM me via Slack (mentioned in the readme)

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/7293

Issue #1904, opened 2019-01-14
[REQ][Java] Add covariant overrides for inherited allOf setters


doctau:
When generating a model class which is an allOf, the Java generators make it a subclass (of the first allOf entry anyway) with all the getter/setter methods inherited from the parent. This means that if you are using the fluent setters, the result of calling any inherited setter is the parent not the subclass. It would be useful if it returned the subclass.

For example, consider this fragment
`
    Foo:
      type: object
      properties:
        a:
          type: string

    Bar:
      allOf:
      - $ref: '#/components/schemas/Foo'
      - type: object
        properties:
          b:
            type: string
`

The following code will fail, because Bar.a() returns a Foo instead of a Bar. It would be safe to override the method to return Bar, since covariant return types are allowed in Java.
`Bar bar = new Bar().a("a").b("b")`

This would make using the fluent setters easier when allOf is used.

wing328:
Which version are you using? Please try the latest master if you've not yet done so.

doctau:
I was using 3.3.4, but I'll try master shortly. Thanks

sluglit:
The problem still occurs in 4.0.0-SNAPSHOT

I wanted to create a custom pojo.mustache but model.parentVars is not set either

pixelshaded:
I also attempted a custom template, but parentVars is empty along with vars in the subclass. This might be ok if there was a way to determine inside allVars if a var is from a parent. There is no such boolean.

pixelshaded:
isInherited seems like the proper boolean for this but it is false in cases when the var comes from a parent.

Mattias-Sehlstedt:
This should have been fixed with https://github.com/OpenAPITools/openapi-generator/pull/15125

PR #5232, opened 2020-02-07
Fix ParentVars and IsInherited for Java

pixelshaded:
The first step in being able to add overrides to parent fluent setters is to get the isInherited and parentVars props working so the templates can know that a particular var is inherited.

https://github.com/OpenAPITools/openapi-generator/issues/1904

Looking at the other languages, inheritance seems to be handled above the DefaultCodegen class. I take this to mean that inheritance rules likely vary between languages so it needs to go in to a language specific generator class. I figured all Java inheritance is the same, so put these changes in the AbstractJavaCodegen that everything should extend from.

I also copied a test from the JavaCodegenClientTest and made it more robust by actually checking parentVars and isInherited props. These tests validated that the parent child relationship exists between CodegenModels, but not on the child properties etc.

The current implementation is very naive. It assumes that all variables shared between the parent and child are inherited. I chose this route because there may be multiple layers of inheritance, so any inherited props from the grandparent should be considered inherited on the child as well.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
@@ -38,6 +38,7 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.util.*;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


daberni:
anything one can help with? would be very glad if this fix could get merged (@nmuesch)

PR #5242, opened 2020-02-07
[Python-experimental] fix name of generated python class for composed schema

sebastien-rosset:
In some cases there is a mismatch in the name of the generated Python class.
The python-experimental/model_templates/model_composed.mustache template is used to generate a Python class when the corresponding OAS model is a composed schema, i.e. oneOf, anyOf, allAll. The name of the class uses the {{unescapedDescription}} tag, whereas the composed elements used {{.}}, which may not yield the same results.

The name of the class may not match the name of the referenced class in the generated _composed_schemas() function.

Note: The use of "unescapedDescription" as a holder for the name of the Python class is very confusing. I saw the comments indicating there are no other fields that can be used, but still imo it's not a pretty workaround.
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L753

The problem occurs in the following scenario:
```yaml
components:
  schemas:
   os.Install:
      allOf:
        - '#/components/schemas/os.AbstractXYZ'
        - type: object
           properties:
               ...
```

The name of the generated class is "OsIntall", but the code generator also creates a reference to the "OSInstall" type.

Update: the PR as is does not work. @spacether , do you have a suggestion on how to change the template such that the compose schema references will match the unescapedDescription tag?

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
@@ -38,17 +38,17 @@ class {{unescapedDescription}}(ModelComposed):
         return {
           'anyOf': [
 {{#anyOf}}
-              {{{.}}},
+              {{unescapedDescription}},
 {{/anyOf}}
           ],
           'allOf': [
 {{#allOf}}
-              {{{.}}},
+              {{unescapedDescription}},
 {{/allOf}}
           ],
           'oneOf': [
 {{#oneOf}}
-              {{{.}}},
+              {{unescapedDescription}},
 {{/oneOf}}
           ],
         }{{#discriminator}}


spacether:
> Update: the PR as is does not work. @spacether , do you have a suggestion on how to change the template such that the compose schema references will match the unescapedDescription tag?

We do very specific processing of the model names in different locations to make this work.
Can you list the strings that we have to work with, what we currently produce and what we should be using here?

sebastien-rosset:
> > Update: the PR as is does not work. @spacether , do you have a suggestion on how to change the template such that the compose schema references will match the unescapedDescription tag?
> 
> We do very specific processing of the model names in different locations to make this work.
> Can you list the strings that we have to work with, what we currently produce and what we should be using here? These model names should have the form module.ClassName here, unless they are the same type as the class that we are in, then they should be ClassName.

Sure. Here is a specific example:
```yaml
    os.Install:
        title: OS Install
        additionalProperties: true
        allOf:
            # This 'os.Install' class inherits all properties from its parent class os.BaseInstallConfig.
            - $ref: '#/components/schemas/os.BaseInstallConfig'
            # The inline definition of all the properties specified in this 'os.Install' class. 
            - type: object
              required:
                  - ObjectType
              properties:
                  Name:
                          type: string
```

sebastien-rosset:
Root cause is #5248

PR #5258, opened 2020-02-10
[kotlin]Fix ktor doesn't generate nullable types

### LINKED ISSUE #5198: kotlin-server/ktor doesn't generate nullable types
##### Description

The kotlin-server/ Ktor generator generates a `Path.kt` file that doesn't mark optional parameters as nullable.

openapi-generator-cli/4.2.3

```yaml
openapi: 3.0.0
info:
  version: 0.0.1
  title: Optional Parameter Demo
paths:
  /foo:
    get:
      parameters:
        - in: query
          name: optional
          schema:
            type: integer
        - in: query
          name: mandatory
          required: true
          schema:
            type: integer
      responses:
        200:
          description: ''
          content:
            text/plain:
              schema:
                type: integer
```

`java -jar openapi-generator-cli.jar generate -i demo.yaml -g kotlin-server -o test/`

##### Steps to reproduce

 1. Run the above command on the declaration file provided

This will generate the following template:

```kotlin
/**
* Optional Parameter Demo
* No description provided (generated by Openapi Generator https://github.com/openapitools/openapi-generator)
*
* The version of the OpenAPI document: 0.0.1
*
*
* NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
* https://openapi-generator.tech
* Do not edit the class manually.
*/
package org.openapitools.server

import io.ktor.locations.KtorExperimentalLocationsAPI
import io.ktor.locations.Location

object Paths {
    /**
     *
     *
     * @param mandatory
     * @param optional  (optional)
     */
    @KtorExperimentalLocationsAPI
    @Location("/foo") class fooGet(val mandatory: kotlin.Int, val optional: kotlin.Int)

}
```

Optional parameters are not annotated as nullable (`?`).

[Paths.kt.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache#L18) should be ammended to print `?` for optional parameters.

yutaka0m:
fixes #5198

Optional parameters have to annotated as nullable

### PR checklist

I can't find `./bin/openapi3/kotlin-server-petstore.sh`:cry:

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache
@@ -18,7 +18,7 @@ object Paths {
      {{#allParams}}* @param {{paramName}} {{description}} {{^required}}(optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}
      {{/allParams}}*/
     @KtorExperimentalLocationsAPI
-    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
+    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#required}}{{#isNullable}}?{{/isNullable}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
 
     {{/bodyAllowed}}
     {{/operation}}
--- a/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache
@@ -18,7 +18,7 @@ object Paths {
      {{#allParams}}* @param {{paramName}} {{description}} {{^required}}(optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}
      {{/allParams}}*/
     @KtorExperimentalLocationsAPI
-    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
+    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
 
     {{/bodyAllowed}}
     {{/operation}}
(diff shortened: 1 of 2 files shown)


yutaka0m:

ackintosh:
Restarted the TravisCI job.

ackintosh:
If no one has further feedback, I'll merge this tomorrow.

modules/openapi-generator/src/main/resources/kotlin-server/libraries/ktor/Paths.kt.mustache
```diff
@@ -18,7 +18,7 @@ object Paths {
      {{#allParams}}* @param {{paramName}} {{description}} {{^required}}(optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}
      {{/allParams}}*/
     @KtorExperimentalLocationsAPI
-    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
+    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
```
4brunu: Maybe we should also check for {{#isNullable}}?{{/isNullable}}

yutaka0m: Thank you for reviewing!!

How is this? -> 

```suggestion
    @Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{#isNullable}}?{{/isNullable}}{{^isNullable}}{{^required}}? = null{{/required}}{{/isNullable}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
```


I couldn't test because I don't know how to send `null` in query parameters.

4brunu: What about something like this?

`@Location("{{path}}") class {{operationId}}({{#allParams}}val {{paramName}}: {{{dataType}}}{{#isNullable}}?{{^required}} = null{{/required}}{{/isNullable}}{{^isNullable}}{{^required}}? = null{{/required}}{{/isNullable}}{{#hasMore}}, {{/hasMore}}{{/allParams}})`

jimschubert: I think required would have to append `?` if it's nullable, but this would change assignment in the method body not just in the parameter list. Also, rather than assigning `null` I think we'd want to honor defaults. For example, `i: Int? = 1` would be a valid argument.

4brunu: I think this is a bit dangerous, because is a valid value that can be used without the user noticing.

yutaka0m:
@jimschubert @4brunu 
Thank you for reviewing!!!

I added `?` when parameter is `required` and `isNullable`. (@4brunu reccomentded it)

4brunu:
Looks good to me 👍

wing328:
@yutaka0m thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

Issue #5262, opened 2020-02-10
[BUG][typescript-axios] missing discriminator


cervotoc:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description
While using multiple inheritance the generated classes do not include the discriminator field. This mean the server is unable to select correct implementation for the abstract class. 

##### openapi-generator version

4.2.1-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.2
info:
  title: TEST
  version: "1.0"
paths:
  /test:
    post:
      operationId: test
      responses:
        200:
          description: result
          content:
            'application/json':
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/AeItemVO'
components:
  schemas:
    AeItemVO:
      required:
        - itemTypeId
        - textValue
      type: object
      properties:
        id:
          type: integer
          format: int32
        uuid:
          type: string
        itemTypeId:
          type: integer
          format: int32
        itemSpecId:
          type: integer
          format: int32
        textValue:
          type: string
      discriminator:
        propertyName: '@class'
        mapping:
          AeItemInteger: '#/components/schemas/AeItemIntegerVO'
          AeItemString: '#/components/schemas/AeItemStringVO'
    AeItemIntegerVO:
      allOf:
        - $ref: '#/components/schemas/AeItemVO'
        - type: object
          properties:
            value:
              type: integer
              format: int32
          required:
            - value
    AeItemStringVO:
      allOf:
        - $ref: '#/components/schemas/AeItemVO'
        - type: object
          properties:
            value:
              type: string
          required:
            - value
```

##### Command line used for generation

`openapi-generator generate -i openapi.yml -g typescript-axios -o src/api/generated/ -p withSeparateModelsAndApi=true,apiPackage=api,modelPackage=model`

##### Steps to reproduce

run the above command.
The generated `src/api/generated/model/ae-item-vo.ts` contains

```
export interface AeItemVO {
[...]

PR #5263, opened 2020-02-10
Fix for #5262 [BUG][typescript-axios] missing discriminator

cervotoc:
Extended musatche template to include generating of discriminator.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/modelGeneric.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/modelGeneric.mustache
@@ -8,6 +8,14 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
     [key: string]: {{{additionalPropertiesType}}}{{#hasVars}} | any{{/hasVars}};
 
 {{/additionalPropertiesType}}
+{{#discriminator}}
+    /**
+     * Discriminator
+     * @type {string}
+     * @memberof {{classname}}
+     */
+    '{{propertyBaseName}}': string;
+{{/discriminator}}
 {{#vars}}
     /**
      * {{{description}}}


amakhrov:
Looks like this change would put the discriminator property twice in the generated interface, if the property is explicitly listed in api spec. This is the case in the examples provided in OAS3, e.g.
```
      "Pet": {
        "type": "object",
        "discriminator": {
          "propertyName": "petType"
        },
        "properties": {
          "name": {
            "type": "string"
          },
          "petType": {
            "type": "string"
          }
        },
        "required": [
          "name",
          "petType"
        ]
      }
```

amakhrov:
Shouldn't a fix involve adding a missing implicit discriminator into allProps in DefaultCodegen instead? And only in a case when it is missing indeed.
How do other generators handle this scenario?

macjohnny:
@wing328 can you help here?

cervotoc:
Ok, I agree. But when I add an property, the server side (openapi generator: spring) adds that property to class and I don't want that. I just want on java files annotation for discriminator `@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "@class", visible = true)`

amakhrov:
@cervotoc I see.
Apparently, in Java `discriminator` is emitted separately from model properties, hence no collision.

However, in TS you're emitting it as a regular model property. So there must be some flag in the model determining whether we need to add the discriminator as a new property into the generated model or not. Or just add this unconditionally in the Typescript Client (not in the default codegen, as it would break the existing logic for java and probably something else).

The problem is that having the discriminator as a model property in api-spec is a totally valid use case, and is demonstrated in many examples in OAS2 and OAS3. Hence we must make sure the codegen still generates valid typescript code for this case - as well as for your case with no explicit property for discriminator.

wing328:
> @wing328 can you help here?

I'll take a look later this week.

wing328:

cervotoc:
Can some one  help with  #5694 / #5695 ? Nobody response :\

macjohnny:
@amakhrov what do you suggest to do here?

amakhrov:
@macjohnny I could imagine having additional logic in `AbstractTypeScriptClient#postProcessModels`:
- for each model
  - if it has discriminator, check if the property with the same name already exists in properties (allProperties?)
  - if exists, do nothing
  - if doesn't exist - add it as a property of type `string`

PR #5266, opened 2020-02-10
[typescript] Don't hardcode the date type if user has mapped it

asmundg:
If the user has set --type-mappings Date=foo, respect this setting
instead of hardcoding the date type.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -86,6 +86,8 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         apiPackage = "api";
         modelPackage = "model";
 
+        typeMapping.put("DateTime", "Date");
+
         this.cliOptions.add(new CliOption(NPM_REPOSITORY,
                 "Use this property to set an url your private npmRepo in the package.json"));
         this.cliOptions.add(CliOption.newBoolean(WITH_INTERFACES,
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
@@ -134,15 +134,10 @@ public class TypeScriptFetchClientCodegen extends AbstractTypeScriptClientCodege
 
     @Override
     public String getTypeDeclaration(Schema p) {
-        String oasType = getSchemaType(p);
         if (ModelUtils.isFileSchema(p)) {
             return "Blob";
         } else if (ModelUtils.isBinarySchema(p)) {
             return "Blob";
-        } else if (ModelUtils.isDateSchema(p)) {
-            return typeMapping.containsKey("Date") ? typeMapping.get("Date") : "Date";
-        } else if (ModelUtils.isDateTimeSchema(p)) {
-            return typeMapping.containsKey("Date") ? typeMapping.get("Date") : "Date";
         }
         return super.getTypeDeclaration(p);
     }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
@@ -63,6 +63,9 @@ public class TypeScriptFetchClientCodegen extends AbstractTypeScriptClientCodege
         this.modelTemplateFiles.put("models.mustache", ".ts");
         this.addExtraReservedWords();
 
+        typeMapping.put("date", "Date");
+        typeMapping.put("DateTime", "Date");
+
         this.cliOptions.add(new CliOption(NPM_REPOSITORY, "Use this property to set an url your private npmRepo in the package.json"));
         this.cliOptions.add(new CliOption(WITH_INTERFACES, "Setting this property to true will generate interfaces next to the default class implementations.", SchemaTypeUtil.BOOLEAN_TYPE).defaultValue(Boolean.FALSE.toString()));
         this.cliOptions.add(new CliOption(USE_SINGLE_REQUEST_PARAMETER, "Setting this property to true will generate functions with a single argument containing all API endpoint parameters instead of one argument per parameter.", SchemaTypeUtil.BOOLEAN_TYPE).defaultValue(Boolean.TRUE.toString()));
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
@@ -134,14 +134,15 @@ public class TypeScriptFetchClientCodegen extends AbstractTypeScriptClientCodege
 
     @Override
     public String getTypeDeclaration(Schema p) {
+        String oasType = getSchemaType(p);
         if (ModelUtils.isFileSchema(p)) {
             return "Blob";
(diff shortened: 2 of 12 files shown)


asmundg:
cc @amakhrov @macjohnny

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -134,14 +134,15 @@ public void processOpts() {
 
     @Override
     public String getTypeDeclaration(Schema p) {
+        String oasType = getSchemaType(p);
```
amakhrov: is this variable used?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -134,14 +134,15 @@ public void processOpts() {
...
             return "Blob";
         } else if (ModelUtils.isBinarySchema(p)) {
             return "Blob";
         } else if (ModelUtils.isDateSchema(p)) {
-            return "Date";
+            return typeMapping.containsKey("Date") ? typeMapping.get("Date") : "Date";
         } else if (ModelUtils.isDateTimeSchema(p)) {
-            return "Date";
+            return typeMapping.containsKey("Date") ? typeMapping.get("Date") : "Date";
         }
```
amakhrov: I don't think there is a need to have custom type mapping for datetime here at all - because AbstractTsClient already includes it into its `typeMappings` map. That's why it's not a problem for the TsAngularClient, which doesn't define any specific mapping for `DateTime` at all, and it still works.
So we could just remove it from the child TS generators and rely on the abstract one

Any thoughts?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -134,14 +134,15 @@ public void processOpts() {
...
     @Override
     public String getTypeDeclaration(Schema p) {
+        String oasType = getSchemaType(p);
         if (ModelUtils.isFileSchema(p)) {
             return "Blob";
         } else if (ModelUtils.isBinarySchema(p)) {
             return "Blob";
         } else if (ModelUtils.isDateSchema(p)) {
-            return "Date";
+            return typeMapping.containsKey("Date") ? typeMapping.get("Date") : "Date";
```
amakhrov: not sure this is correct.
According to `DefaultCodegen#getPrimitiveType()` implementation, `"Date"` is not a valid primitive type. Instead, `"date"` is (more precisely, `SchemaTypeUtil.DATE_FORMAT`).

Again, the AbstractTsClient already includes a reasonable mapping for it (maps to "string": `typeMapping.put("date", "string");`). So having it in the child TS codegens seems redundant

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
asmundg: The custom mapping was apparently introduced in 1e023f3b6e4d655d6e553bb359ffb128f5a20382 to fix a bug where someone expected the default type for date to be `Date`.

I agree that it seems a bit weird and that changing the defaults is exactly what type-mappings is for. Is it reasonable to change the (now) default behavior for date back to `string`?

amakhrov: There are a few things going on here:
- Abstract TS client has default mapping for date and DateTime (latter maps to typescript Date object - by mistake)
- TsFetch client doesn't define its own mappings, but redefines getTypeDeclaration in a way that it's always Date.
- TsFetch client also defines some runtime logic in the generated TS code to actually instantiate Date objects from the strings. So does TsReduxQuery.
- TsRxjs client doesn't generate any mapping code, so `Date` type in the interfaces is simply incorrect there again.

Based on the above, TsFetch's usage of Date object is legit - because the generated code indeed converts raw http data to and from Date. (the same also true for TsReduxQuery).
The only reason for a user to change this default behavior is if they have custom templates (proivided via `--template-dir`), which do not do any Date serialization, in contrast to built-in templates.

I can imagine the following solution:
- All child TS clients rely on AbstractTs's protected typeMappings. Furthermore, we want to make sure it maps `DateTime` to string, as it already does for `date`.
- **Some** child TS clients (fetch, redux-query) should override those typeMappings - not via custom Java code, but via the same protected `typeMappings`.
- If a user has special needs (because they use custom mustache templates), they provide custom mappings typeMappings via `--type-mappings` option, which takes precedence over values populated in the codegen constructor.

asmundg: Right, that sounds sensible.

modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/fetch/TypeScriptFetchModelTest.java
```diff
@@ -95,7 +95,7 @@ public void simpleModelTest() {
         final CodegenProperty property4 = cm.vars.get(3);
         Assert.assertEquals(property4.baseName, "birthDate");
         Assert.assertEquals(property4.complexType, null);
-        Assert.assertEquals(property4.dataType, "Date");
+        Assert.assertEquals(property4.dataType, "string");
```
macjohnny: this should be `Date`, since typescript-fetch implements date deserialization:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache#L66

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -86,6 +86,8 @@ public TypeScriptAngularClientCodegen() {
         apiPackage = "api";
         modelPackage = "model";
 
+        typeMapping.put("DateTime", "Date");
```
macjohnny: I think it would be good to remove this, i.e. to change the mapping from `Date` to `string` for `format: date-time`, since the typescript-angular generator does not support serialization for the models.
Note: this would pose a problem, since serialization for query params is implemented, see e.g. https://github.com/OpenAPITools/openapi-generator/pull/5225

asmundg: Right, it was unclear to me whether de-serialization was supported for angular. And it would require changing the unit tests.

I'm tempted to just keep the current behavior and punt on "fixing" it for angular.

amakhrov: @macjohnny we can remove it separately (for angular and other generators except fetch/reduxquery), since it **might** break existing behavior for current users

macjohnny:
@asmundg can you please re-generate all typescript samples?
see https://circleci.com/gh/OpenAPITools/openapi-generator/12620#tests/containers/2

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
asmundg: Ok, I updated the rest of the generators to keep the output consistent.

wing328:
@asmundg thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

batousik:
@asmundg Can you explain why the generator is no longer using Date objects?

macjohnny:
@batousik see https://github.com/OpenAPITools/openapi-generator/pull/5314

PR #5269, opened 2020-02-10
[BUG][C++] Avoid using plain underscore when escaping reserved words

### LINKED ISSUE #5268: [BUG][C++] Reserved keywords are escaped via underscore prefix, resulting in compile failure on MSVC 14
#### Bug Report Checklist

##### Description

In code generated by the cpprest client generator, we use the keyword `inline` is a query parameter, which is reserved in C++. By default, reserved keywords are escaped by prefixing them with an underscore.

This is documented as bad practice in C++, as there may still be [conflicts with the global namespace:](https://en.cppreference.com/w/cpp/keyword)

>Also, all identifiers that contain a double underscore __ in any position and each identifier that begins with an underscore followed by an uppercase letter is always reserved and **all identifiers that begin with an underscore are reserved for use as names in the global namespace.**

See the sample OpenAPI file below. When generating the cpprest client for this file, the signature for the generated GET /things call in DefaultApi.h looks like

```
    pplx::task<std::vector<utility::string_t>> thingsGet(
        boost::optional<bool> _inline
    );
```

On MSVC 14, this actually causes a compile failure:

>DefaultApi.h(60): error C2735: 'inline' keyword is not permitted in formal parameter type specifier

4.2.2

```yaml
    openapi: 3.0.0
    info:
      title: Inline API
      version: 0.0.1
    paths:
      /things:
        get:
          summary: Returns a list of things.
          parameters:
            - in: query
              name: inline
              schema:
                type: boolean
              description: Matches a keyword in C++.
          responses:
            '200':    # status code
              description: A JSON array of things.
              content:
                application/json:
                  schema: 
                    type: array
                    items: 
                      type: string

```

```
path/to/java.exe -cp openapi-generator-cli.jar org.openapitools.codegen.OpenAPIGenerator generate -g cpp-restsdk --skip-validate-spec -i test.yml -o output
```

##### Steps to reproduce

1. Generate the cpp-restsdk client for the file given above.
2. Notice the underscores in the generated source.
3. Compile using msvc-14. The argument `_inline` is rejected.

##### Related issues/PRs

I found various language-specific escaping issues, but none which were directly relevant to this.

##### Suggest a fix

We were able to work around this using templates and the preprocessor. We added:
```
#define _inline r_inline
```
to the api-header.mustache template file, effectively replacing the keyword via the preprocessor before it gets compiled.

However, I believe the generator itself should be modified so that it escapes reserved words in a manner which does not conflict with C++ conventions. I opted for r_<keyword>, but anything similar can be used.

I will post a PR immediately following this issue with this fix, although if you think there are any better alternatives, please let me know. I would be happy to adjust as needed. Thanks!

alexweav:
Fixes #5268  as described in the issue. Are there any better preferred ways of escaping this?

Ran the generator against the file given in the issue, and verified that the output now compiled using MSVC-14. Repeated below, for ease of use:

```yaml
    openapi: 3.0.0
    info:
      title: Inline API
      version: 0.0.1
    paths:
      /things:
        get:
          summary: Returns a list of things.
          parameters:
            - in: query
              name: inline
              schema:
                type: boolean
              description: Matches a keyword in C++.
          responses:
            '200':    # status code
              description: A JSON array of things.
              content:
                application/json:
                  schema: 
                    type: array
                    items: 
                      type: string
```

### PR checklist


alexweav:
cc @ravinikam @stkrwork @MartinDelille @muttleyxd

wing328:
@alexweav thanks for the PR.

What about adding an option `reservedWordPrefix` so that people can customize it in whatever way they want (and the change will be backward compatible)? 

We can default it to `r_` like what you've suggested.

alexweav:
@wing328 @etherealjoy Apologies for the slow response on this. I added a CLI option, `reservedWordPrefix`, which allows users to configure the prefix, although it will still default to `r_`.

I made this change in the base abstract generator for C++, but I noticed that the CppRestSdkClientCodegen actually clears all CLI options inherited from its base class. Removing this clear-all brought in more unwanted options, so I instead opted for adding the option directly to the CppRest class too. Is there a better way of working around this?

Since this option was added to the base abstract generator, it will be inherited by all other C++ generators which do not clear their inherited options. I noticed that the CppPistacheServer generator and the CppRestbedServer generator also clear their inherited options, but the other C++ generators don't. If you would like, I can explicitly add this option to these two as well, following the same approach I took for the CppRest generator. However, if there is a different/more generic approach that you would prefer, please let me know.

etherealjoy:
LGTM.
However there is a CI issue because some of the changed files were not archived.
Could you please check in those files as well, they got changed due to this newly added flag
```shell
	modified:   docs/generators/cpp-qt5-client.md
	modified:   docs/generators/cpp-qt5-qhttpengine-server.md
	modified:   docs/generators/cpp-restsdk.md
	modified:   docs/generators/cpp-tizen.md
```

I was going to suggest the reserved keyword prefix be left as **_** but i realized it will not compile on Windows MSVC.

wing328:
>  If you would like, I can explicitly add this option to these two as well, following the same approach I took for the CppRest generator. However, if there is a different/more generic approach that you would prefer, please let me know.

Please file another PR with the same approach to do so. Thank you.

alexweav:
>Please file another PR with the same approach to do so. Thank you.

Posted: #5418

wing328:
@alexweav thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5312, opened 2020-02-13
[RESOLVE] Added Spring JPA annotations support to SpringCodegen

MarkusBansky:
Added Spring JPA annotations support to SpringCodegen.

Added `additionalAnnotations` to **CodegenModel**, **CodegenProperty**.
Added `isJpaEntity` to **CodegenModel**.
Added `useSpringJpa` config option to **SpringCodegen**.

Example: 

```yaml
User:
  type: object
  x-java-jpa-entity: true
  x-java-jpa-table: users
  properties:
    id:
      type: integer
      format: int64
      x-java-jpa-annotations:
        - "@Id"
        - "@GeneratedValue(strategy = GenerationType.IDENTITY)"
    firstName:
      type: string
    lastName:
      type: string
    username:
      type: string
    passwordHash:
      type: string
```

Would generate this class for you:

```java
/**
 * User
 */
@javax.annotation.Generated(value = "org.openapitools.codegen.languages.SpringCodegen", date = "2020-02-13T21:44:31.102132Z[Europe/London]")

@Entity // <---- makes this an Entity
@Table(name = "users") // <---- applies table name
public class User   {

  @Id // <---- adds annotations
  @GeneratedValue(strategy = GenerationType.IDENTITY)
  @JsonProperty("id")
  private Long id;


  @JsonProperty("firstName")
  private String firstName;


  @JsonProperty("lastName")
  private String lastName;


  @JsonProperty("username")
  private String username;


  @JsonProperty("passwordHash")
  private String passwordHash;

  public User id(Long id) {
    this.id = id;
    return this;
  }

  /**
   * Get id
   * @return id
  */
  @ApiModelProperty(value = "")


  public Long getId() {
    return id;
  }

  public void setId(Long id) {
    this.id = id;
  }

/// the rest of the file there
```

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
@@ -74,6 +74,9 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
 
     public Map<String, Object> vendorExtensions = new HashMap<String, Object>();
 
+    public boolean isJpaEntity = false;
+    public Set<String> additionalAnnotations = new TreeSet<String>();
+
     //The type of the value from additional properties. Used in map like objects.
     public String additionalPropertiesType;
 
@@ -97,6 +100,14 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
 
     public void setAdditionalPropertiesType(String additionalPropertiesType) {
         this.additionalPropertiesType = additionalPropertiesType;
+    }
+
+    public Set<String> getAdditionalAnnotations() {
+        return additionalAnnotations;
+    }
+
+    public void setAdditionalAnnotations(Set<String> additionalAnnotations) {
+        this.additionalAnnotations = additionalAnnotations;
     }
 
     public Set<String> getAllMandatory() {
@@ -538,6 +549,7 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
                 hasChildren == that.hasChildren &&
                 isMapModel == that.isMapModel &&
                 hasOnlyReadOnly == that.hasOnlyReadOnly &&
+                isJpaEntity == that.isJpaEntity &&
                 getUniqueItems() == that.getUniqueItems() &&
                 getExclusiveMinimum() == that.getExclusiveMinimum() &&
                 getExclusiveMaximum() == that.getExclusiveMaximum() &&
@@ -580,6 +592,7 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
                 Objects.equals(externalDocumentation, that.externalDocumentation) &&
                 Objects.equals(vendorExtensions, that.vendorExtensions) &&
                 Objects.equals(additionalPropertiesType, that.additionalPropertiesType) &&
+                Objects.equals(getAdditionalAnnotations(), that.getAdditionalAnnotations()) &&
                 Objects.equals(getMaxProperties(), that.getMaxProperties()) &&
                 Objects.equals(getMinProperties(), that.getMinProperties()) &&
                 Objects.equals(getMaxItems(), that.getMaxItems()) &&
@@ -599,14 +612,14 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
                 getInterfaceModels(), getChildren(), anyOf, oneOf, allOf, getName(), getClassname(), getTitle(),
                 getDescription(), getClassVarName(), getModelJson(), getDataType(), getXmlPrefix(), getXmlNamespace(),
                 getXmlName(), getClassFilename(), getUnescapedDescription(), getDiscriminator(), getDefaultValue(),
-                getArrayModelType(), isAlias, isString, isInteger, isLong, isNumber, isNumeric, isFloat, isDouble,
+                getArrayModelType(), isAlias, isString, isInteger, isLong, isNumber, isNumeric, isFloat, isDouble, isJpaEntity,
                 getVars(), getAllVars(), getRequiredVars(), getOptionalVars(), getReadOnlyVars(), getReadWriteVars(),
                 getParentVars(), getAllowableValues(), getMandatory(), getAllMandatory(), getImports(), hasVars,
                 isEmptyVars(), hasMoreModels, hasEnums, isEnum, isNullable, hasRequired, hasOptional, isArrayModel,
                 hasChildren, isMapModel, hasOnlyReadOnly, getExternalDocumentation(), getVendorExtensions(),
                 getAdditionalPropertiesType(), getMaxProperties(), getMinProperties(), getUniqueItems(), getMaxItems(),
                 getMinItems(), getMaxLength(), getMinLength(), getExclusiveMinimum(), getExclusiveMaximum(), getMinimum(),
-                getMaximum(), getPattern(), getMultipleOf());
+                getMaximum(), getPattern(), getMultipleOf(), getAdditionalAnnotations());
     }
 
     @Override
@@ -684,6 +697,8 @@ public class CodegenModel implements IJsonSchemaValidationProperties {
         sb.append(", maximum='").append(maximum).append('\'');
         sb.append(", pattern='").append(pattern).append('\'');
         sb.append(", multipleOf='").append(multipleOf).append('\'');
+        sb.append(", isJpaEntity=").append(isJpaEntity);
+        sb.append(", additionalAnnotations=").append(additionalAnnotations);
         sb.append('}');
         return sb.toString();
     }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
@@ -52,6 +52,7 @@ public class SpringCodegen extends AbstractJavaCodegen
     public static final String CONFIG_PACKAGE = "configPackage";
     public static final String BASE_PACKAGE = "basePackage";
     public static final String INTERFACE_ONLY = "interfaceOnly";
+    public static final String USE_SPRING_JPA = "useSpringJpa";
     public static final String DELEGATE_PATTERN = "delegatePattern";
     public static final String SINGLE_CONTENT_TYPES = "singleContentTypes";
     public static final String VIRTUAL_SERVICE = "virtualService";
@@ -98,6 +99,7 @@ public class SpringCodegen extends AbstractJavaCodegen
     protected boolean hateoas = false;
     protected boolean returnSuccessCode = false;
     protected boolean unhandledException = false;
+    protected boolean useSpringJpa = false;
 
     public SpringCodegen() {
         super();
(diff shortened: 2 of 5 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
```diff
@@ -74,6 +74,9 @@
 
     public Map<String, Object> vendorExtensions = new HashMap<String, Object>();
 
+    public boolean isJpaEntity = false;
+    public Set<String> additionalAnnotations = new TreeSet<String>();
```
MarkusBansky: `additionalAnnotations` can also be used in future to add customised annotations to denigrated model class.

InfoSec812:
+1 This is fantastic! I'll have to see about porting this to work in other Java generators for things like JAX-RS and others. Thanks!

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
InfoSec812: To support additional annotation, we would also probably need to have something like `x-additional-imports` to add those annotations to the imports on the class.

InfoSec812: And that in turn would require some way to add additional dependencies to the POM perhaps?

MarkusBansky: I've added a `spring-data-jpa` dependency to `pom.mustache` when user specifies the `useSpringJpa` in `configOptions`, and it adds the import to model class when that model contains a flag `isJpaEntity`

MarkusBansky:
Short docs:

1. openApiGenerate **configOptions** must contain `useSpringJpa: 'true'` to use JPA annotations.
2. Each model that is supposed to use JPA annotations must have `x-java-jpa-entity` set to `true`.
3. `x-java-jpa-table` takes a string and is optional, it can set the custom name of the table, if not present, then default `@Table` tag is added.
4. Each property in such model can have `x-java-jpa-annotations` - a list of annotations as string, each one would be added to the property, example:
```yaml
x-java-jpa-annotations:
  - "@Id"
  - "@GeneratedValue(strategy = GenerationType.IDENTITY)"
```

Dependency for `spring-data-jpa` is added to `pom` when user specifies to use spring jpa.
Import for `persistence.*` is added to a model that is specified as `x-java-jpa-entity: true`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -52,6 +52,7 @@
     public static final String CONFIG_PACKAGE = "configPackage";
     public static final String BASE_PACKAGE = "basePackage";
     public static final String INTERFACE_ONLY = "interfaceOnly";
+    public static final String USE_SPRING_JPA = "useSpringJpa";
```
InfoSec812: @MarkusBansky I wonder if we could move this from the `SpringCodegen` up to the `AbstractJavaCodegen` so that it can be of benefit to ALL of the Java generators?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -98,6 +99,7 @@
     protected boolean hateoas = false;
     protected boolean returnSuccessCode = false;
     protected boolean unhandledException = false;
+    protected boolean useSpringJpa = false;
```
InfoSec812: @MarkusBansky And here, we could rename this property from `useSpringJpa` to `useJpa`?

samples/client/petstore/spring-cloud-async/src/main/java/org/openapitools/model/Category.java
```diff
@@ -15,9 +15,11 @@
 @ApiModel(description = "A category for a pet")
 
 public class Category   {
+
```
InfoSec812: @MarkusBansky Looks like your code formatter made a LOT of whitespace changes... Perhaps it would be easier to review this PR if there weren't 500+ files with simple whitespace changes?

MarkusBansky: @InfoSec812 these files where generated from a terminal, with a command from the pull request description, why is it inserting a whitespace then?

InfoSec812: It could be that your changes to the template is causing the whitespace to be added?

jimbydamonk: I played around with this. It is because of the mustache template. I tried this:
```diff --git a/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache b/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache
index e26e1d8d9b..5ccb28faf8 100644
--- a/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/pojo.mustache
@@ -19,8 +19,10 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}}{{^parent}}
 {{>enumClass}}
     {{/mostInnerItems}}
     {{/isContainer}}
-    {{/isEnum}}{{#additionalAnnotations}}
-  {{{.}}}{{/additionalAnnotations}}
+    {{/isEnum}}
+{{#additionalAnnotations}}
+  {{{.}}}
+{{/additionalAnnotations}}
   {{#jackson}}
   @JsonProperty("{{baseName}}"){{#withXml}}
   @JacksonXmlProperty({{#isXmlAttribute}}isAttribute = true, {{/isXmlAttribute}}{{#xmlNamespace}}namespace="{{xmlNamespace}}", {{/xmlNamespace}}localName = "{{#xmlName}}{{xmlName}}{{/xmlName}}{{^xmlName}}{{baseName}}{{/xmlName}}"){{/withXml}}

```
And that clean it up.

v0l0dia: Is there any progress on reverting newline/whitespaces from this PR? As this feature is somewhat I've been looking for, I'm very much intrested in it. I was looking into customizing spring generator by myself and it seems almost everything I need is ready. 
Maybe one could help with sanitizing those 500+ newline changes or reimplementing the PR?

evgeniyrymko:
Hello. I am a relatively new person in Java development. I have a problem with applying the described annotations. I use Open API Generator Gradle plugin, version 4.3.1. I added the line "useSpringJpa": "true" in openapi-generator-config.json file, then I modified my OAS file and added properties to schema definition:
 ProjectResponse:
      type: object
      x-java-jpa-entity: true
      x-java-jpa-table: users
      properties:
        id:
          type: integer
          format: int64
          x-java-jpa-annotations:
            - "@Id"
            - "@GeneratedValue(strategy = GenerationType.IDENTITY)"
          description: unique id of the project for the identifier
          example: 1
        name:
          type: string
          description: the name of the project
          example: To-do Application Development
After generating classes, I didn't get annotations that are described in this topic.
Can you help me with that? Thanks)

lgadawski:
When it could potentially be merged?

PedroGuerraPT:
Hello @MarkusBansky @InfoSec812, is this still being pursued?

jasondamour:
Hello @MarkusBansky @InfoSec812, is this still being pursued?

MarkusBansky:
Should look into this. Maybe there was already something similar added

jasondamour:
I poked around, I don't see anything similar yet :(

jasondamour:
@InfoSec812 Can you clarify how to specify additional imports for those annotations? I can add the annotations just fine, but any annotation from a non-standard package fails to resolve. I tried all sorts of variations of import and type mappings, but no beans:

For example: this openapi.yaml:
```
components:
  schemas:
    User:
      type: object
      x-class-extra-annotation: |-
        @Data
        @Document
      required:
        - id
        - name
      properties:
        id:
          type: string
          x-field-extra-annotation: |-
            @Id
        name:
          type: string
```

And this pom.xml config:
```
<plugin>
    <version>6.4.0</version>
    <executions>
        <execution>
            <id>generate-spring</id>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>${project.parent.basedir}/api-schemas/platform.yaml</inputSpec>
                <generatorName>spring</generatorName>
                <typeMappings>
                    <typeMapping>Data=lombok.Data</typeMapping>
                    <typeMapping>Document=org.springframework.data.mongodb.core.mapping.Document</typeMapping>
                    <typeMapping>Id=org.springframework.data.annotation.Id</typeMapping>
                </typeMappings>
                <importMappings>
                    <importMapping>Data=lombok.Data</importMapping>
                    <importMapping>Document=org.springframework.data.mongodb.core.mapping.Document</importMapping>
                    <importMapping>Id=org.springframework.data.annotation.Id</importMapping>
                </importMappings>
                <configOptions>
                    <interfaceOnly>true</interfaceOnly>
                    <openApiNullable>false</openApiNullable>
                    <useTags>true</useTags>
                    <useBeanValidation>true</useBeanValidation>
                    <performBeanValidation>true</performBeanValidation>
                </configOptions>
            </configuration>
        </execution>
    </executions>
</plugin>
```

This fails because the generated models doesn't import lombok or org.springframework.data packages, and theres no way I can find to add imports, hence adding my vote to this MR

InfoSec812:
@jasondamour You cannot add imports. That is why my examples show fully-qualified package and class names for the annotations.

jasondamour:
Ah I see. I made a silly mistake when trying out the FQDN class approach, my openapi maven module didn't have the dependencies declared so the package wasn't found. Thank @InfoSec812 !!

benjaminpochat:
I mention [the problem described here](https://stackoverflow.com/questions/75911300/is-there-a-way-to-specify-jpa-annotations-for-fields-referencing-entities-with-o). It deals with JPA annotation for entity fields (such as @OneToOne, @OneToMany, etc...). It would be great if such annotation could be generated. I suggest the following pattern for that case :

`
components:
  schemas:
    User:
        address:
          allOf:
            - $ref: '#/components/schemas/Address'
          x-field-extra-annotation: |-
            @OneToOne
`

InfoSec812:
> I mention [the problem described here](https://stackoverflow.com/questions/75911300/is-there-a-way-to-specify-jpa-annotations-for-fields-referencing-entities-with-o). It deals with JPA annotation for entity fields (such as @OnetoOne, @OneToMany, etc...). It would be great if such annotation could be generated. I suggest the following pattern for that case :
> 
> ```
> components:
>   schemas:
>     User:
>         address:
>           allOf:
>             - $ref: '#/components/schemas/Address'
>           x-java-jpa-annotation: |-
>             @OneToOne
> ```

Out of curiousity, why would you not define it like so:

```
components:
  schemas:
    User:
        address:
          $ref: '#/components/schemas/Address'
          x-java-field-annotation: |-
            @OneToOne
```

benjaminpochat:
> Out of curiousity, why would you not define it like so:
> 
> components:
>   schemas:
>     User:
>         address:
>           $ref: '#/components/schemas/Address'
>           x-java-field-annotation: |-
>             @OneToOne
> 

Because [OpenApi specs](https://swagger.io/docs/specification/using-ref/#sibling) says explicitly : "Any sibling elements of a $ref are ignored.".

The syntax suggested based on "allOf" property is not very elegant, but it seems to be the only workaround to add additional field properties to referenced type. See [this ticket at openapi-spec.](https://github.com/swagger-api/swagger-ui/issues/4732)

InfoSec812:
@benjaminpochat Since OpenAPI 3.1 $ref siblings should be supported and work: https://github.com/OAI/OpenAPI-Specification/issues/2744#issuecomment-939796784

benjaminpochat:
> @benjaminpochat Since OpenAPI 3.1 $ref siblings should be supported and work: https://github.com/OAI/OpenAPI-Specification/issues/2744#issuecomment-939796784

Great ! Thanks for this information.
Then, if it could be used to add some field annotations in SpringCodeGen, it would be event perfect.
I tried to use it with maven plugin (openapi-generator-maven-plugin) in versions 6.2.1 and 6.5.0. But with both versions, the field annotations were not generated.

alexander793:
Why is this PR not merged yet? I need this feature in my current project :grimacing:

PR #5314, opened 2020-02-13
[typescript] Remove incomplete support of Date type

amakhrov:
**Intro**
Historically, TypeScript generators had typeMapping "DateTime=Date". However only select generators generate code that really does serialization/deserialization of dates at runtime.

Some other generators support serialization of Date query parameters, but not deserialize models returned from api. Without proper model deserialization, the whole DateTime=Date mapping is simply incorrect (declared data type is `Date`, but real value at runtime coming from api is still `string`).

- **Generators that fully implement serialization**: `typescript-fetch`, `typescript-reduxquery`, `typescript-node`.
- **Generators with partial support** (query params only, but not models): `typescript-angular`, `typescript-axios`.

Lately this partial Date support seems to get quite a bit of traction:
https://github.com/OpenAPITools/openapi-generator/pull/4869, https://github.com/OpenAPITools/openapi-generator/pull/5057, https://github.com/OpenAPITools/openapi-generator/pull/5225, https://github.com/OpenAPITools/openapi-generator/pull/5266, https://github.com/OpenAPITools/openapi-generator/pull/5287

**The approach implemented in this PR is**
- Keep Date support as is for generators that fully support serialization/deserialization.
- Set typeMapping `DateTime=string` for all other typescript generators (actually, in `AbstractTypescriptClient`).
- Remove partial support for Dates from mustache templates (for generators with partial support, as listed above).

### PR checklist


modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -168,13 +168,6 @@ export class {{classname}} {
         if (typeof value === "object") {
             if (Array.isArray(value)) {
                 (value as any[]).forEach( elem => httpParams = this.addToHttpParamsRecursive(httpParams, elem, key));
-            } else if (value instanceof Date) {
```
macjohnny: @amakhrov how about we leave this code here for users that manually provide the current type mapping to Date as a fallback?

amakhrov: I still fail to understand how using DateTime=Date mapping can possibly work for anyone.
It emits Models with properties of type Date, while these are in fact strings - enforcing developers to hack it around (like `myModel.dateField as any as string`).

We can probably keep this till the major release (5.0) and clean it up then - what do you think?

macjohnny: currently `DateTime=Date` can only work if no dates are used in any body param model, but only in query params

macjohnny: I think 
https://github.com/OpenAPITools/openapi-generator/pull/5314/files#diff-5a152f3888a8d2712c9824bc1eff5db6L89
should be released with `4.3.0`
while
https://github.com/OpenAPITools/openapi-generator/pull/5314/files#diff-f61af8a4c4eb2720bfb15259e0dca419L155
should be reverted to allow a fallback for backwards compatibility. Moreover, this will be useful when a full date conversion is implemented later.
Leaving it here does not have an impact for the new default `DateTime=string` mapping.

macjohnny: and the https://github.com/OpenAPITools/openapi-generator/pull/5314/files#diff-d270de95051055e6c877f904c360ecadL117-L128 part should also be kept, for backwards compatibility. it is checked for `instanceof Date`

amakhrov: I'm ok with keeping the mapping in templates as is - although I believe the only reason it was ever needed is because the generated code (with DateTime=Date) forced users to pass Dates as params.
Now, if someone still wants to have this mapping for whatever reason (provided via --type-mapping argument), they can support that with a customized template (via --template-dir).

amakhrov: A discarded the commits changing the templates - only keep the ones that change teh default typemapping

macjohnny:
Note: the old type mapping can be achieved by providing a custom type mapping of `--type-mappings=DateTimeDateTime=Date`

SetupCoding:
> Note: the old type mapping can be achieved by providing a custom type mapping of `--type-mappings=DateTimeDateTime=Date`
Just to clarify, it should be `--type-mappings=DateTime=Date`.

wing328:

PR #5315, opened 2020-02-14
[python] add method to set/get default configuration

tomplus:
As a follow up the discussion https://github.com/OpenAPITools/openapi-generator/pull/4485#issuecomment-584915749 I'd like to recovery an option to set default configuration. I prepared very simple solution which provides the same API as we had before.

Thanks for reviewing it.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -4,6 +4,7 @@
 
 from __future__ import absolute_import
 
+import copy
 import logging
 {{^asyncio}}
 import multiprocessing
@@ -116,6 +117,8 @@ class Configuration(object):
 {{/hasHttpSignatureMethods}}
 {{/hasAuthMethods}}
     """
+
+    _default = None
 
     def __init__(self, host="{{{basePath}}}",
                  api_key=None, api_key_prefix=None,
@@ -240,6 +243,31 @@ class Configuration(object):
         """
         # Disable client side validation
         self.client_side_validation = True
+
+    @classmethod
+    def set_default(cls, default):
+        """Set default instance of configuration.
+
+        It stores default configuration, which can be
+        returned by get_instance method.
+
+        :param default: object of Configuration
+        """
+        cls._default = copy.copy(default)
+
+    @classmethod
+    def get_instance(cls):
+        """Return new instance of configuration.
+
+        This method returns newly created, based on default constructor,
+        object of Configuration class or returns a copy of default
+        configuration passed by the set_default method.
+
+        :return: The configuration object.
+        """
+        if cls._default is not None:
+            return copy.copy(cls._default)
+        return Configuration()
 
     @property
     def logger_file(self):
--- a/modules/openapi-generator/src/main/resources/python/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python/configuration.mustache
@@ -245,18 +245,27 @@ class Configuration(object):
         self.client_side_validation = True
 
     @classmethod
+    def copy(cls, source):
+        if source is None:
+            return None
+        ret = copy.copy(source)
+        ret.api_key = copy.copy(source.api_key)
+        ret.api_key_prefix = copy.copy(source.api_key_prefix)
+        return ret
+
+    @classmethod
     def set_default(cls, default):
         """Set default instance of configuration.
 
         It stores default configuration, which can be
-        returned by get_instance method.
+        returned by new_instance method.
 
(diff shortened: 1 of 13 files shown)


samples/openapi3/client/petstore/python/petstore_api/configuration.py
```diff
@@ -169,6 +172,31 @@ def __init__(self, host="http://petstore.swagger.io:80/v2",
...
+        """Return new instance of configuration.
+
+        This method returns newly created, based on default constructor,
+        object of Configuration class or returns a copy of default
+        configuration passed by the set_default method.
+
+        :return: The configuration object.
+        """
+        if cls._default is not None:
+            return copy.copy(cls._default)
```
spacether: Why copy it vs just return the instance?
The CI tests are failing when trying to deepcopy the logger.

spacether: How about returning a new Configuration instance which uses the same instantiation inputs as the default instance?

tomplus: Yes, deepcopy doesn't work. Here as in the previous implementation a copy of default configuration is returned.

The CI tests failed because a remote endpoint returned 404 error on add_pet method. I don't know why, I'll try to run it again.

tomplus: Tests passed. I guess the original intention was to return a copy which looks like a new object but with different default values. Maybe I should name the method `new_instance` instead of `get_instance` to make it more understandable... ?

samples/client/petstore/python/tests/test_configuration.py
```diff
@@ -22,14 +22,25 @@ def setUp(self):
...
+
+    def testDefaultConfiguration(self):
+
+        # prepare default configuration
+        c1 = petstore_api.Configuration(host="example.com")
+        petstore_api.Configuration.set_default(c1)
+
+        # get default configuration
+        c2 = petstore_api.Configuration.get_instance()
+        self.assertEqual(c2.host, "example.com")
```
spacether: Can you also add a test that verifies that the
contents of the api_key and api_key_prefix dicts are the same but that their ids are different?
Or will they be the same because of the shallow copy?

tomplus: I wasn't aware that api_key are dict, thanks.

samples/openapi3/client/petstore/python/petstore_api/configuration.py
tomplus: I changed method name to `new_instance()`.

spacether: new_instance is good.
How do you feel about get_default_copy as a name?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -241,6 +244,40 @@ class Configuration(object):
         # Disable client side validation
         self.client_side_validation = True
 
+    @classmethod
+    def copy(cls, source):
```
spacether: How about changing this to:
`__deepcopy__(self, memodict={})`
That way you control how everything can be copied. For the logger only you can then make a new logger with the same properties as the old one. All other properties can be deep copied. You can use dir(self) or `self.__dict__` to iterate over all of self's properties. 

If you do it this way, then in new_instance you can return copy deepcopy(cls._default)

samples/openapi3/client/petstore/python/petstore_api/configuration.py
tomplus: Updated, both versions look good to me.

modules/openapi-generator/src/main/resources/python/configuration.mustache
tomplus: Yeah, I'd totally forgotten about this possibility. Thanks.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -241,6 +244,45 @@ class Configuration(object):
...
+
+        This method returns newly created, based on default constructor,
+        object of Configuration class or returns a copy of default
+        configuration passed by the set_default method.
+
+        :return: The configuration object.
+        """
+        if cls._default is not None:
+            return copy.deepcopy(cls._default)
+        return Configuration()
```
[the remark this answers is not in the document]
tomplus: I think `get_default_copy` is good enough for this usage.

tomplus:
Thank you.

Issue #5248, opened 2020-02-08
[BUG][codegen] Clarify behavior of inlined schema with "title" attribute


sebastien-rosset:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

It would be useful to clarify the behavior of the inlined "title" attribute. When a OAS model is defined inline, the generator flattens the model and tries to use the inlined "title" attribute in the OAS document to create a model name. If "title" is not present in the spec, it uses the hard-coded 'InlineObject'' model name. This is a surprising feature, as the title is for documentation purpose, as described in the JSON schema spec. It would be good to document that in the wiki and when invoking the validator; maybe also add a warning.

wing328:
A bit more background on why we use `title` to name the inline model. Some users prefer to define the model using the inline schema instead of defining the model separately, which gives users total control in naming the model. We do advise them to define the model separately if they want to have complete control on the model naming but they still prefer to use the inline schema. 

`title` is probably the closest or most reasonable pick to name the inline schema (and we agree it's not the most ideal way to name the model defined in the inline schema) so that's what we've chosen at the moment to name models defined by the inline schemas.

> The generated name ends up being "OSInstall", unlike the outer class name which is "OsInstall".

For the issue you mentioned, it's a bug as the naming should be done using `toModelName` override in the generator (regardless of what attribute, title or something else, to use when naming the inline schema)

```
    os.Install:
        allOf:
            - $ref: '#/components/schemas/os.Abstract'
            - type: object
              title: OS Install
              properties:
                  Name:
                          type: string
```

There was a PR to generate the last inline schema as a separate model while the old behavior was that the last inline schema's properties are added to the object/model/class (e.g. os.Install in this case). There were a couple of issues opened that users do not like the new behavior so we may consider switching back or add an option to make the behavior configurable.

sebastien-rosset:
[...]

PR #5316, opened 2020-02-14
[codegen] Improve java code comments and argument documentation; Fix issue with ComposedSchema that has undeclared properties

sebastien-rosset:
Add code comments in ModelUtils and documentation for the --generate-alias-as-model argument.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -328,6 +328,33 @@ public class InlineModelResolver {
         }
     }
 
+    /**
+     * Flattens properties of inline object schemas that belong to a composed schema into a
+     * single flat list of properties. This is useful to generate a single or multiple
+     * inheritance model.
+     * 
+     * In the example below, codegen may generate a 'Dog' class that extends from the
+     * generated 'Animal' class. 'Dog' has additional properties 'name', 'age' and 'breed' that
+     * are flattened as a single list of properties.
+     * 
+     * Dog:
+     *   allOf:
+     *     - $ref: '#/components/schemas/Animal'
+     *     - type: object
+     *       properties:
+     *         name:
+     *           type: string
+     *         age:
+     *           type: string
+     *     - type: object
+     *       properties:
+     *         breed:
+     *           type: string
+     * 
+     * @param openAPI the OpenAPI document
+     * @param key a unique name ofr the composed schema.
+     * @param children the list of nested schemas within a composed schema (allOf, anyOf, oneOf).
+     */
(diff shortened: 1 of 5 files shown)


sebastien-rosset:
This PR is a fix for #5317

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -338,6 +365,13 @@ private void flattenComposedChildren(OpenAPI openAPI, String key, List<Schema> c
...
                 ObjectSchema op = (ObjectSchema) component;
                 if (op.get$ref() == null && op.getProperties() != null && op.getProperties().size() > 0) {
+                    // Note: the call to op.getTitle() can lead to totally unexpected results and should not be
+                    // used to generate the innerModelName.
+                    // If the value of the 'title' attribute happens to match a schema defined elsewhere in
+                    // the specification, 'innerModelName' will be the same as that other schema.
+                    // The 'title' attribute is supposed to be for human consumption, not for code generation.
+                    // OAS authors should not be expected to set a 'title' value that will control the
+                    // code generation logic.
                     String innerModelName = resolveModelName(op.getTitle(), key);
```
sebastien-rosset: IMO, the "title" attribute in a OAS document should not have any impact whatsover on the code generation (codegen currently uses title to flatten inline schemas). Using 'title' can lead to completely unexpected outcomes for the generated code.

The OAS spec clearly states the 'title' attribute is for human consumption, so codegen should not use it.

I was working with a draft OAS document that has a composed allOf schema. The "title" of the inner schema happens to be the name of the outer (composed) object, but it could also have been any value, including names of totally unrelated schema.

That had the weird side effect to nullify the inheritance hierarchy. Then you have to debug the code to finally understand why the "parent" attribute ends up being nullified as a side effect of the "title" attribute

jimschubert: I agree with this, and always thought it was odd.

Maybe we should consider it a bug and replace it's support with a vendor extension in 4.3.0. We don't have any metrics about who is actually using the feature, but I doubt anyone is because:

1. it's not really documented
2. it's confusing
3. it may block users with well-defined specs from using the tooling

sebastien-rosset: I have opened #5248 for this issue.

sebastien-rosset:
Thank you @jimschubert !

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
wing328: Based on the discussion in https://github.com/OpenAPITools/openapi-generator/issues/5248 I'll file a PR to revise the wording here.

PR #5327, opened 2020-02-14
[Kotlin][Client] Enum toString handling

Mordag:
I have noticed that in certain cases, the enum usage for path parameter and url query parameter is incorrectly implemented. 

In the case I have an enum with MY_ENUM("MyEnum") and I am using the enum as a path parameter. The path would look like that "myPath/.../MY_ENUM/...". This request would always fail. That is due to the different enum value than the server would expect.

To fix this issue, I suggest we add a custom toString to every enum class so that we will not use the enum var name but the value. That way the path would look like this "myPath/.../MyEnum/...".

Changes made with this PR:
- Added custom toString to the enum class script.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/enum_class.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/enum_class.mustache
@@ -41,6 +41,14 @@ import kotlinx.serialization.internal.CommonEnumSerializer
 
 {{/enumVars}}{{/allowableValues}}
 
+	/**
+	This override toString avoids using the enum var name and uses the actual api value instead.
+	In cases the var name and value are different, the client would send incorrect enums to the server.
+	**/
+	override fun toString(): String {
+        return value
+    }
+
 {{#multiplatform}}
     {{#nonPublicApi}}internal {{/nonPublicApi}}object Serializer : CommonEnumSerializer<{{classname}}>("{{classname}}", values(), values().map { it.value.toString() }.toTypedArray())
 {{/multiplatform}}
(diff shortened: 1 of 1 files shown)


4brunu:
Thanks for your contribution 👍 
Can you please run `./mvnw package && sh bin/kotlin-client-all.sh` and commit the changes in the pet projects?

modules/openapi-generator/src/main/resources/kotlin-client/enum_class.mustache
```diff
@@ -41,6 +41,14 @@ import kotlinx.serialization.internal.CommonEnumSerializer
 
 {{/enumVars}}{{/allowableValues}}
 
+	/**
+	This override toString avoids using the enum var name and uses the actual api value instead.
+	In cases the var name and value are different, the client would send incorrect enums to the server.
+	**/
+	override fun toString(): String {
+        return value
```
jimschubert:
```suggestion
        return value{{^isString}}.toString(){{/isString}}
```

Mordag:
@4brunu I already run this but it caused no changes to the sample project. I guess the sample project are not using any enums?

jimschubert:
> I guess the sample project are not using any enums?

Looks like there are only two enums in the test yaml for Kotlin: a query string and a model property. We might want to regenerate at some point later with a yaml that tests more use cases.

The Shippable error was a failure to build Elixir client due to Shippable being flaky. I've kicked off a rebuild. If it fails again, we shouldn't let this block merging.

wing328:
> We might want to regenerate at some point later with a yaml that tests more use cases.

We should switch to the fake petstore spec which contains a lot more tests for enum. Last time I tried to switch but got compilation errors. I'll test again later to see if these PRs for the Kotlin generators fix the issue.

wing328:
@Mordag thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

Issue #5324, opened 2020-02-14
[BUG][Java] Generated API class has wrong dataType and does not compile


sebastien-rosset:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Using OpenAPI v3 document to generate a client-side Java SDK with the default library (okhttp-gson). The generated classes under the "api" directory cause a compilation failure. 

For example, a generated API Java class api/SyslogApi.java has the following createSyslogPolicyCall() method. The requestBody argument is a templatized PolicyAbstractPolicy<String, Object>. 

```
  public okhttp3.Call
  createSyslogPolicyCall(PolicyAbstractPolicy<String, Object> requestBody,
                         final ApiCallback _callback) throws ApiException {
    Object localVarPostBody = requestBody;
```

However, the generated model/PolicyAbstractPolicy.java class is not a generic, hence the compilation fails. There are two problems:
1. The argument uses a generic
1. The argument type is the parent class

I was expecting the generated method signature to be something like:

```
public okhttp3.Call createSyslogPolicyCall(SyslogPolicy syslogPolicy, final ApiCallback _callback)
```

The mustache templates that are used to generate these classes are:
1. API: [Java/libraries/native/api.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/api.mustache)
1. Model: modules/openapi-generator/src/main/resources/Java/pojo.mustache

It looks like the problem is the value of **dataType** in the API template does not match the class name of the POJO
```
{{{dataType}}} {{paramName}}
```

I am using the default Java library, which is okhttp-gson.

##### openapi-generator version

master February 14th 2020

##### OpenAPI declaration file content or url

Operations:
```yaml
  /syslog/Policies:
    post:
      tags:
      - syslog
      summary: Create a 'syslog.Policy' resource.
      operationId: CreateSyslogPolicy
      requestBody:
[...]

PR #5331, opened 2020-02-15
[Java] Generated API class has wrong dataType and does not compile Issue

sebastien-rosset:
Details of the issue are described in #5324.

Without this PR, the generated code is as below. There are two problems: 1) the type of the requestBody argument is wrong (it should be SyslogPolicy, not PolicyAbstractPolicy); 2) The type of the requestBody argument uses generics.
 
```
createSyslogPolicyCall(PolicyAbstractPolicy<String, Object> requestBody ....
```

With this PR, the generated code is as follows, it fixes both problems:
```
createSyslogPolicyCall(SyslogPolicy requestBody ....
```

This is not ready to be merged, but I need a bit of guidance to make progress.

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1752,6 +1752,10 @@ public String getSchemaType(Schema schema) {
      */
     @SuppressWarnings("static-method")
     public String toAllOfName(List<String> names, ComposedSchema composedSchema) {
+        Map<String, Object> exts = composedSchema.getExtensions();
+        if (exts != null && exts.containsKey("x-allOf-name")) {
+            return (String) exts.get("x-allOf-name");
```
wing328: We want to standardize vendor extension naming with lower cases only separated by hyphens so  what about using`x-all-of-name` instead?

wing328:
cc @OpenAPITools/generator-core-team  as the change covers default codegen.

modules/openapi-generator/src/test/java/org/openapitools/codegen/java/JavaModelEnumTest.java
```diff
@@ -172,6 +172,7 @@ public void testEnumTestSchema() {
         codegen.setOpenAPI(openAPI);
 
         Schema enumTest = openAPI.getComponents().getSchemas().get("Enum_Test");
+        Assert.assertNotNull(enumTest);
```
spacether: Do our tests cover this new code?
This model: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1422
does not contain `x-all-of-name`
Also doesn't this test require that the schema not be null?
Down below we test that the CodegenModel generated form the schema contains vars, which means that the schema had to not be null to contain those vars.

sebastien-rosset:
> Do our tests cover this new code?

The reason I added this line is only to make the test more robust, it's not really related to this PR. At some point during my testing, I mistakenly changed the content of the OAS document which is used for this enum test. It ended up returning a null value for "enumTest", and it caused a null pointer exception at the next line. So even if enumTest is null, the next line will fail, which is what is expected, but it's better to assert that enumTest is not null rather than relying on a NPE to be raised.

> Also doesn't this test already require that the schema not be null?
Yes, but this above, this makes the UT code more robust.

sebastien-rosset:
The "avro-schema" fake-petstore generator fails in the CI process. I found out the .drone file invokes bin/utils/test-fake-petstore-for-all.sh, which in turns executes a java command that redirects all output to /dev/null, so it's not possible to troubleshoot by directly looking at the CI logs. I then executed the script locally, but the execution completes successfully in my local workspace. There must be something different in the CI execution environment compared to my local workspace.
To get around the problem, I have changed the bin/utils/test-fake-petstore-for-all.sh script to redirect the output to a file, and if the execution fails, cat the log file.

```
ERROR: Failed to run avro-schema
--
140 | java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -g avro-schema -o /tmp/openapi-generator-test-fake-petstore/2.0/avro-schema
```

wing328:
Looks good but I've not tested it locally

PR #5339, opened 2020-02-16
[Rust Server] Add support for untyped properties and models

richardwhiuk:
If a property has no specified type in the schema then it is allowed to take any type. This is problematic in Rust, where all code is statically typed. As such, this PR makes the assumption that JSON is in use, and thus we can use `serde_json::Value` as an escape hatch. This likely doesn't work for XML based APIs. To support both, theoretically, we could parameterise all models which included untyped properties (e.g. as `ModelName<T>`) and then have `T` be either `serde_json::Value` or `xml::Node` or similar. This isn't implemented here, but is a possible expansion.

This PR also clarifies support regarding unknown mime types. Unknown mime types are now treated as unstructured data - and presented either as a byte array or a String - e.g. `application/yaml` is presented as a string

## Rust Technical Committee

- @frol
- @farcaller
- @bjgill

### PR checklist


richardwhiuk:
https://github.com/OpenAPITools/openapi-generator/issues/5387 raises a question here, which we should probably clarify before merging this.

sebastien-rosset:
How will this PR handle the following schema? Will you only accept maps of strings to values? Or any valid JSON/XML document?

```yaml
components:
  schemas:
    Blob:
      type: object
```

richardwhiuk:
Any valid JSON/XML document due to the limitations of `ModelUtils.isFreeformObject` to distinguish:

```yaml
Foo:
  type: object
```

and

```yaml
Foo: {}
```

as discussed in #5387.

I don't think that's a blocker on merging this - once that's fixed, we can support for the former in the Rust Server Generator.

I think there's a secondary problem that we default to treating additional properties as not present.

sebastien-rosset:
> Any valid JSON/XML document due to the limitations of `ModelUtils.isFreeformObject` to distinguish:
> 
> ```yaml
> Foo:
>   type: object
> ```
> 
> and
> 
> ```yaml
> Foo: {}
> ```
> 
> as discussed in #5387.
> 
> I don't think that's a blocker on merging this - once that's fixed, we can support for the former in the Rust Server Generator.
> 
> I think there's a secondary problem that we default to treating additional properties as not present.

I've added a helper function in this PR to distinguish the two cases: https://github.com/OpenAPITools/openapi-generator/pull/5391/files#diff-57d7532cf464a8d7c24aab4b22ceb993R680

PR #5350, opened 2020-02-17
Use yarn for website builds

jimschubert:
### PR checklist

cc @wing328 fyi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ before_cache:
   - rm -fr $HOME/.m2/repository/org/openapitools/
 
 cache:
+  yarn: true
   directories:
   - $HOME/.m2
   - $HOME/.ivy2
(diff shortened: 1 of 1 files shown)


xvilo:
As this has no description whatsoever, what's the desired reason to use yarn over npm? You're now creating an extra dependency that needs to be installed too. (See the before_install).

jimschubert:
@xvilo this has no effect on users or their processes whatsoever.

Npm doesn't support pinning dependencies in a good way, even with shrinkwrap. Our new site uses an alpha version of Docusaurus, so we don't want Travis to be randomly broken. Using Yarn is the only good way to pin versions with loose semver options.

xvilo:
Instead you could do npm ci instead of npm install. That would probably fix your issues...

jimschubert:
`npm ci` skips user interactivity. It doesn't change npm's poor resolution algorithm. 

Can you explain exactly what you're concerned about, and I'll try to address your concern. If you're concerned about speed, we can evaluate the time differences. However, the solution would probably be to move to yarn for everything as we've had pretty regular issues with npm on Travis.

Issue #5351, opened 2020-02-17
[Python][Validation] Knob to control whether to ignore additional attribute in the HTTP response


sebastien-rosset:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I've noticed SDKs are inconsistent in how they handle additional, undeclared attributes in HTTP response. What do you think about having a language-independent parameter that controls the leniency?

Some clients (such as python-experimental) do strict validation of the response. If the response includes undeclared properties that the client does not know about, the client raises an exception. Other clients (such as go and Java) are lenient and ignore additional attributes from the server JSON response.

Use case: The client has been generated and keeps running for weeks without problem. Then the server is upgraded to a newer version, adding one more declared property in the schema, in a way which is backward compatible. I.e. the OAS spec has a new property to an existing API. The new server sends the new property when the existing API is requested.
The problem is as soon as the server is upgraded, the clients will start failing. If there are thousands of clients, this is bad because as soon as the server is upgraded, all clients fail and must be upgraded.

A user of the client SDK may have the intent to run the same version of the SDK for a while, even if the server is upgraded. They may be perfectly fine doing this because the server provides backward compatibility guarantees and new attributes can be safely ignored. Other users may want to make sure the client fails when a new attribute in sent by the server, because they really want the client and server to be always at the same version.

As a workaround, it's possible to set "additionalproperties" in the spec everywhere, to prepare for additions to the spec. I don't find this solution very appealing because it can't address the above use cases.

##### openapi-generator version

master February 17th 2020

sebastien-rosset:
@spacether

spacether:
Also, I am not sure if additionalProperties is widely supported across our generators yet.
This CLI option sounds useful when people want to allow new properties without raising validation errors.

sebastien-rosset:
Yes, I can confirm support for additional properties is lacking in Java and go

sebastien-rosset:
[...]

PR #5362, opened 2020-02-18
[Python] add discard_unknown_keys parameter

sebastien-rosset:
Fix for #5351

### PR checklist


spacether:
@sebastien-rosset this looks good.
Setting the default to False like you did is my preference. It preserves our current master branch behavior. 
Can you also 
- edit the model classes to not raise exceptions with extra key input?
You could do this by only keeping the known args in kwargs in the init method.
- add a couple of tests 
Thank you!

sebastien-rosset:
> @sebastien-rosset this looks good.
> Setting the default to False like you did is my preference. It preserves our current master branch behavior.
> Can you also
> 
> * edit the model classes to not raise exceptions with extra key input?
>   You could do this by only keeping the known args in kwargs in the init method.
> * add a couple of tests
>   Thank you!

Which init method are you talking about?

sebastien-rosset:
> @sebastien-rosset this looks good.
> Setting the default to False like you did is my preference. It preserves our current master branch behavior.
> Can you also
> 
> * edit the model classes to not raise exceptions with extra key input?
>   You could do this by only keeping the known args in kwargs in the init method.
> * add a couple of tests
>   Thank you!

I modified this code:
```
    if len(unused_args) > 0:
        if len(additional_properties_model_instances) == 0 and \
                    not self._configuration.discard_unknown_keys:
            raise ApiValueError(
                "Invalid input arguments input when making an instance of "
                "class %s. Not all inputs were used. The unused input data "
                "is %s" % (self.__class__.__name__, unused_args)
            )
        if len(additional_properties_model_instances) > 0:
            for var_name, var_value in six.iteritems(unused_args):
                for instance in additional_properties_model_instances:
                    setattr(instance, var_name, var_value)
```

spacether:
This change impacts composed models and non-composed models.
What about the below update here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_shared.mustache#L23

Something like:
```
        self._data_store = {}
        self._check_type = _check_type
        self._from_server = _from_server
        self._path_to_item = _path_to_item
        self._configuration = _configuration

        if self. _configuration. discard_unknown_keys:
            property_names = list(kwargs.keys())
            for property_name in property_names:
                if property_name not in self. attribute_map:
                    del kwargs[property_name]
````

File: modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache [hunk 1]
```diff
@@ -17,6 +17,11 @@
         self._path_to_item = _path_to_item
         self._configuration = _configuration
 
+        if self._configuration.discard_unknown_keys:
```
spacether: Can you handle the case when _configuration is none?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
```diff
@@ -17,6 +17,11 @@
...
+            for property_name in list(kwargs.keys()):
+                if property_name not in self.attribute_map:
+                    del kwargs[property_name]
```
spacether: Can you move this new code and the 
```
        self._data_store = {}
        ...
        self._configuration = _configuration
```
into method_init_shared?
It is always used together and defined only once if we put it there.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1124,9 +1124,10 @@ def validate_get_composed_info(constant_args, model_args, self):
                 "class %s. Not all inputs were used. The unused input data "
                 "is %s" % (self.__class__.__name__, unused_args)
             )
-        for var_name, var_value in six.iteritems(unused_args):
-            for instance in additional_properties_model_instances:
-                setattr(instance, var_name, var_value)
+        else:
+            for var_name, var_value in six.iteritems(unused_args):
+                for instance in additional_properties_model_instances:
+                    setattr(instance, var_name, var_value)
```
spacether: Can you explain why we need this?
I think that unused_args will be length 0 if we are dropping the extra kwargs before we invoke this method.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
sebastien-rosset: done.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
sebastien-rosset: Done.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
sebastien-rosset: No it won't be. This is the case when the unused_args are injected into the additionalproperties map.

sebastien-rosset:
Does this need to be handled too?
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_normal.mustache#L20

spacether:
> Does this need to be handled too? I.e. can there be unused arguments that would cause setattr to raise an exception?

I don't think so, at that point we have already deleted out the unexpected args.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: Right, but didn't we delete those args with:
self._configuration.discard_unknown_keys = True?
So they no longer exist here.

What should we do if:
- What if one of our composed models has additional properties in it + discard_unknown_keys=True?
- What if self model has additional properties defined in self and discard_unknown_keys=True?

sebastien-rosset: No, I removed that code. If additional properties exists, it will populate the undeclared properties into the additionalproperties Map. Then at that point there will be nothing to discard.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
```diff
@@ -34,9 +28,23 @@
...
+                    # discard variable.
+                    pass
+                else:
+                    # This should not happen, any validation problem should have been raised in
+                    # validate_get_composed_info 
+                    raise ApiValueError(
+                        "Invalid input arguments input when making an instance of "
+                        "class %s. Not all inputs were used. The unused input data "
+                        "is %s" % (self.__class__.__name__, unused_args)
+                    )
```
spacether: Can you keep the exception raising in validate_get_composed_info?
That makes it so we only need to look there to understand what is going wrong vs 2 or more locations.
Also, the model class currently automatically throws an exception when attempting to assign to an unknown variable so we should be okay here.

sebastien-rosset: > Can you keep the exception raising in validate_get_composed_info?
I didn't change that. It's still raising the exception. One thing I could do here is do nothing, because validate_get_composed_info has already raised the exception.

> That makes it so we only need to look there to understand what is going wrong vs 2 or more locations.
> Also, the model class setattr currently automatically throws an exception when attempting to assign to an unknown variable so we should be okay here.

spacether: Can you also add code to cover the non-composed schemas use case?

sebastien-rosset: I just refactored a bit.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: So when discard_unknown_keys=True it still allows additionalproperties to process unknown keys?

sebastien-rosset: Yes, I think that should be the behavior.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
```diff
@@ -34,9 +28,14 @@
...
         self._var_name_to_model_instances = composed_info[1]
         self._additional_properties_model_instances = composed_info[2]
+        unused_args = composed_info[3]
 
 {{#requiredVars}}
         self.{{name}} = {{name}}
 {{/requiredVars}}
         for var_name, var_value in six.iteritems(kwargs):
+            if var_name in unused_args and self._configuration is not None and \
+                            self._configuration.discard_unknown_keys:
```
spacether: How about adding `and not self._additional_properties_model_instances`?
That way when we have _additional_properties_model_instances we will assign those additional props.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: In line41 of modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache we are already assigning additional properties.
How about deleting this else clause because we already handle assign these properties there?

sebastien-rosset: I've added documentation in the configuration file to clarify. There is a difference between "undeclared" versus unknown.

sebastien-rosset:
> Can you also add code to cover the non-composed schemas use case?

Can you help me here? I don't know which case you are referring to.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
sebastien-rosset: done

spacether:
> > Can you also add code to cover the non-composed schemas use case?
> 
> Can you help me here? I don't know which case you are referring to.

We also need toy add your fix skipping unused args here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_normal.mustache#L20

spacether:
> > I don't think so, at that point we have already deleted out the unexpected args.

My earlier answer was assuming we deleted the unused args first thing in our `__init__` method, which we are no longer doing.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
spacether: Right but we already assign these properties in line41.
So right now this code is assigning properties in two places.
Can we only assign these properties in one place?
Can that be on line41 in `__init__`?

sebastien-rosset: > In line41 of modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache we are already assigning additional properties.
> How about deleting this else clause because we already handle assign these properties there?

Sorry I am confused, at line 41 in method_init_composed.mustache there is no else clause

sebastien-rosset: I think I understand now. I've submitted another change.

spacether: At that line we are looping over all kwargs.
Some of those are the unused args.
For those unused args we assign the key value properties to self.
Self puts those in its own data store, and the data stores of all composed models that contain that property.

sebastien-rosset: There is some strange delay in github message delivery. I think I have addressed your point.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_normal.mustache
```diff
@@ -8,12 +8,6 @@
...
 
-        self._data_store = {}
-        self._check_type = _check_type
-        self._from_server = _from_server
-        self._path_to_item = _path_to_item
-        self._configuration = _configuration
-
 {{#requiredVars}}
         self.{{name}} = {{name}}
 {{/requiredVars}}
```
spacether: How about adding this if statement in the below for loop?
```
            if var_name not in self.attribute_map and self._configuration is not None and \
                            self._configuration.discard_unknown_keys and \
                            self.additional_properties_type is None:
                # discard variable.
                pass
```
That will let us skip unknown keys for these non composed schema models.

sebastien-rosset: done

> How about adding this if statement in the below for loop?
> 
> ```
>             if var_name not in self.attribute_map and self._configuration is not None and \
>                             self._configuration.discard_unknown_keys and \
>                             self.additional_properties_type is None:
>                 # discard variable.
>                 pass
> ```
> 
> That will let us skip unknown keys for these non composed schema models.

done

sebastien-rosset:
> How about combining this if statement with its outer one?

done

modules/openapi-generator/src/test/resources/3_0/all_of_with_additional_properties.yaml
```diff
@@ -0,0 +1,43 @@
...
+          type: string
+    Adult:
+      description: A representation of an adult
+      additionalProperties: {}
+      allOf:
+      - $ref: '#/components/schemas/Person'
+      - type: object
+        properties:
+          address:
+            type: string
```
spacether: Could you move these into the python-experimental samples spec?
That way we can test instantiating these classes in python.

sebastien-rosset: I remove this code.

sebastien-rosset:
> This update looks good, we just need some test verifying that it works correctly.
> Can you add verification tests of:
> 
> * a composed schema model without additonalproperties + discard_unknown_keys=True
> * a composed schema model WITH additonalproperties + discard_unknown_keys=True
> * a non-composed schema model without additonalproperties + discard_unknown_keys=True
> * a non-composed schema model WITH additonalproperties + discard_unknown_keys=True
> 
> Some things to check could be:
> 
> * that we instantiate the model without an error
> * that accessing the properties that should have been discarded raises exceptions on those model instances
> * when additonalproperties exist, those extra properties are included in the model even when discard_unknown_keys=True

I did that, but I uncovered a separate bug that would have to be addressed in another PR. I don't
 think the current template supports the scenario when additionalProperties is set as a sibling of allOf.

```yaml
    Cat:
      additionalProperties: true
      allOf:	      allOf:
        - type: object	        - type: object
          properties:	          properties:
            declawed:	            declawed:
```

spacether:
> I did that, but I uncovered a separate bug that would have to be addressed in another PR. I don't
> think the current template supports the scenario when additionalProperties is set as a sibling of allOf.

Have you seen this issue?
https://github.com/swagger-api/swagger-parser/issues/1252

sebastien-rosset:
> > I did that, but I uncovered a separate bug that would have to be addressed in another PR. I don't
> > think the current template supports the scenario when additionalProperties is set as a sibling of allOf.
> 
> Have you seen this issue?
> [swagger-api/swagger-parser#1252](https://github.com/swagger-api/swagger-parser/issues/1252)'

I just saw it. I have recently added documentation to isMapSchema: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java#L409

According to @jimschubert , this is by design, and it makes sense. A composed schema can be both a map and composed. So I have added documentation to help clarify.

> 
> What if your composed schema model has a oneOf or allOf mode which contains additonalProperties? That should demonstrate the composedSchema + addtionalProperties use case.

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+                ]
+            }
+        }
+        # The 'unknown_property' is undeclared, which would normally raise an exception, but
+        # when discard_unknown_keys is set to True, the unknown properties are discarded.
+        response = MockResponse(data=json.dumps(data))
+        deserialized = api_client.deserialize(response,
+            ({str: (petstore_api.Dog,)},), True)
+        self.assertTrue(isinstance(deserialized, dict))
+        self.assertTrue(isinstance(deserialized['dog'], petstore_api.Dog))
```
spacether: Can you add a test to check the value of deserialized['dog'].to_dict?
That would prove that the unknown properties have been discarded.

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+                "color": "black",
+                "breed": "husky",
+                "unknown_property": "a-value"
+            }
+        }
+        response = MockResponse(data=json.dumps(data))
+
+        # Deserializing with strict validation raises an exception because the 'unknown_property'
+        # is undeclared.
+        with self.assertRaises(Exception) as cm:
```
spacether: Can you use our specific exception here rather than the general Exception? Is it petstore_api.ApiValueError?

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+MockResponse = namedtuple('MockResponse', 'data')
+
+class DiscardUnknownPropertiesTests(unittest.TestCase):
+
+    def test_deserialize_dog_do_not_discard_unknown_properties(self):
+        """ deserialize str, Dog) with unknown properties, strict validation is enabled """
+        config = Configuration(discard_unknown_keys=False)
+        api_client = petstore_api.ApiClient(config)
+        data = {
+            'dog': {
```
spacether: If you want, you can change this data to just the dog dict.
That would look like:
```
            deserialized = api_client.deserialize(response,
                (petstore_api.Dog,), True)
```

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+                ({str: (petstore_api.Dog,)},), True)
+        self.assertTrue(re.match('.*Not all inputs were used.*unknown_property.*', str(cm.exception)),
+            'Exception message: {0}'.format(str(cm.exception)))
+
+    def test_deserialize_dog_discard_unknown_properties(self):
+        """ deserialize str, Dog) with unknown properties, discard unknown properties """
+        config = Configuration(discard_unknown_keys=True)
+        api_client = petstore_api.ApiClient(config)
+        data = {
+            'dog': {
```
spacether: If you want, you can change this data to just the dog dict.
That would look like:
```
            deserialized = api_client.deserialize(response,
                (petstore_api.Dog,), True)
```

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+            ({str: (petstore_api.Dog,)},), True)
+        self.assertTrue(isinstance(deserialized, dict))
+        self.assertTrue(isinstance(deserialized['dog'], petstore_api.Dog))
+
+    def test_deserialize_cat_do_not_discard_unknown_properties(self):
+        """ deserialize str, Cat) with unknown properties, strict validation is enabled """
+        config = Configuration(discard_unknown_keys=False)
+        api_client = petstore_api.ApiClient(config)
+        data = {
+            'cat': {
```
spacether: Same comment as above, you can strip this down to just the cat dict

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+        self.assertTrue(isinstance(deserialized, dict))
+        self.assertTrue(isinstance(deserialized['cat'], petstore_api.Cat))
+        self.assertEqual(deserialized['cat']['color'], 'black')
+
+    def test_deserialize_cat_discard_unknown_properties(self):
+        """ deserialize str, Cat) with unknown properties, discard unknown properties """
+        config = Configuration(discard_unknown_keys=True)
+        api_client = petstore_api.ApiClient(config)
+        data = {
+            'cat': {
```
spacether: Same comment as above, you can strip this down to just the cat dict

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+                ]
+            }
+        }
+        # The 'unknown_property' is undeclared, which would normally raise an exception, but
+        # when discard_unknown_keys is set to True, the unknown properties are discarded.
+        response = MockResponse(data=json.dumps(data))
+        deserialized = api_client.deserialize(response,
+            ({str: (petstore_api.Cat,)},), True)
+        self.assertTrue(isinstance(deserialized, dict))
+        self.assertTrue(isinstance(deserialized['cat'], petstore_api.Cat))
```
spacether: Can you add a test to check the value of deserialized['cat'].to_dict?
That would prove that the unknown properties have been discarded.

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        response = MockResponse(data=json.dumps(data))
+
+        # Deserializing with strict validation does not raise an exception because the even though
+        # the 'dynamic-property' is undeclared, the 'Cat' schema defines the additionalProperties
+        # attribute.
+        deserialized = api_client.deserialize(response,
+            ({str: (petstore_api.Cat,)},), True)
+        self.assertTrue(isinstance(deserialized, dict))
```
spacether: Can you an an assertion check to confirm that cat has additionalproperties instances
and add a comment saying that this test case tests composed schema with additionalproperties

sebastien-rosset:
I have to work on another fix for golang. So for now I will keep this in our fork, it works for us and I have so much time in my day to work on adding unit tests that did not exist before.

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
sebastien-rosset: done

sebastien-rosset:
> Thanks for adding those additional test cases!
> This looks great, just waiting for CI tests to pass then we can merge it.

Build passed.

Issue #5246, opened 2020-02-08
[REQ] Please Support Angular 9


eiswind:
### Is your feature request related to a problem? Please describe.

Angular 9 has new requirements on typescript and rxjs that require updates.

lewcianci:
Is this just as simple as updating the generator to reference Typescript >= 3.6.4 and rxjs?  Or is more required? If it's that simple I can submit a PR for this issue as I'm keen to get it resolved also.

macjohnny:
anyone up to adapting the code?

eiswind:
I also updated
"ng-packagr": "^9.0.0", still in need a proof that it works because of other issues.

ltomes:
@eiswind so you updated your `typescript`, `rxjs`, and `ng-packagr` versions? Could you list the versions you are using?

I updated all angular versions as well, and although I can build the module, when I include it in my application, and use it I get cyclic dependency errors
```
Error: Cannot instantiate cyclic dependency! InjectionToken HTTP_INTERCEPTORS
    at throwCyclicDependencyError (core.js:7986)
    at R3Injector.hydrate (core.js:16761)
    at R3Injector.get (core.js:16526)
    at HttpInterceptingHandler.handle (http.js:2865)
    at MergeMapSubscriber.project (http.js:1665)
    at MergeMapSubscriber._tryNext (mergeMap.js:46)
    at MergeMapSubscriber._next (mergeMap.js:36)
    at MergeMapSubscriber.next (Subscriber.js:49)
    at Observable._subscribe (subscribeToArray.js:3)
    at Observable._trySubscribe (Observable.js:42)
```

There is a chance I am still just on the wrong rxjs version (`6.5.4`) or some other dependency.

eiswind:
This worked at my side:

```
"peerDependencies": {
    "@angular/core": "^9.0.0",
    "@angular/common": "^9.0.0",
    "@angular/compiler": "^9.0.0",
    "core-js": "^2.4.0",
    "reflect-metadata": "^0.1.3",
    "rxjs": "^6.5.0"
  },
  "devDependencies": {
    "@angular/common": "^9.0.0",
    "@angular/compiler": "^9.0.0",
    "@angular/compiler-cli": "^9.0.0",
    "@angular/core": "^9.0.0",
    "@angular/platform-browser": "^9.0.0",
    "ng-packagr": "^9.0.0",
    "reflect-metadata": "^0.1.3",
    "rxjs": "^6.5.0",
[...]

PR #5370, opened 2020-02-19
typescript-angular: support angular 9

topce:
Add support for angular 9
fix api.module.ts to generate 
```typescript
ModuleWithProviders<ApiModule>
```
instead of 
```typescript
ModuleWithProviders
```
for more info check 
[migration-module-with-providers](https://angular.io/guide/migration-module-with-providers)

fixes https://github.com/OpenAPITools/openapi-generator/issues/5246

### PR checklist

- [x ] If contributing template-only or documentation-only changes which will change sample output, [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) before.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -192,6 +193,12 @@ public void processOpts() {
             additionalProperties.put(PROVIDED_IN_ROOT, false);
         }
 
+        if (ngVersion.atLeast("9.0.0")) {                
+            additionalProperties.put(MODULE_WITH_PROVIDERS, true);
```
amakhrov: Looks like the purpose of this flag is to drive adding a generic param to the ModuleWithProviders (as in angular 9 the default value for param (any) has been removed). It doesn't affect whether the module with providers is generated or not. Is that correct?
If so, perhaps a better name would be "ENFORCE_GENERIC_MODULE_WITH_PROVIDERS" or smth like that?
Or maybe even remove it from the java class, and just do conditionally in the template (based on ngVersion)?

macjohnny: rename is fine. concerning the template condition, I think this is a bit difficult with the mustache template, so I think the current solution is fine.

topce:
Wow one day after PR and already reviewed
And merged into master 
Simply the best!!! @macjohnny @amakhrov

veluria:
There's a typo in TypeScriptAngularClientCodegen.java#L335 which says `8.0.0` instead of `9.0.0`.

wing328:
@veluria do you filing a PR with the fix when you've time?

veluria:
I can send a PR later.

macjohnny:
@veluria thanks for pointing this out, I filed a PR to fix this: https://github.com/OpenAPITools/openapi-generator/pull/5585

wing328:
@topce thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

Issue #4873, opened 2019-12-27
[BUG] [python-flask] invalid `.from_dict` generated when using object as query parameter


Tohnmeister:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
4.2.2
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
I would've expected this to generate code that compiles/runs. But the `.from_dict` seems invalid Python syntax.
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When using this spec as input:

```yaml
openapi: "3.0.0"
info:
  title: "My API"
  version: "0.0.1"
paths:
  /invoices:
    get:
      parameters:
        - name: filter
          in: query
          description: Filter parameters
          required: false
          schema:
            type: object
            properties:
              field:
                type: string
      responses:
        '200':
          description: An array of invoices
          content:
            application/json:
              schema:
                type: array
                items:
                  type: object
                  properties:
                    number:
                      type: integer
                      format: int64
```

a default_controller.py is generated containing the following code:

```python
import connexion
import six

from openapi_server.models.filter import Filter  # noqa: E501
from openapi_server.models.inline_response200 import InlineResponse200  # noqa: E501
from openapi_server import util


def invoices_get(filter=None):  # noqa: E501
    """invoices_get

     # noqa: E501

    :param filter: Filter parameters
    :type filter: dict | bytes

    :rtype: List[InlineResponse200]
    """
    if connexion.request.is_json:
        filter = .from_dict(connexion.request.get_json())  # noqa: E501
    return 'do some magic!'
```
The `.from_dict` is invalid Python syntax.

Trying to run the generated application, results in the following error:

```bash
jsonschema.exceptions.RefResolutionError: Unresolvable JSON pointer: 'components/parameters/filter'
```

##### openapi-generator version

4.2.2

##### OpenAPI declaration file content or url

See above.

##### Command line used for generation

openapi-generator generate -i ./openapi.yaml -g python-flask -o ./generated

##### Steps to reproduce

- Use above command and openapi.yaml to generate python-flask code.

##### Suggest a fix

[...]

Issue #5235, opened 2020-02-07
[CI Issue] Travis CI failed itself due to "ImportError: cannot import name 'FileStorage'"


wing328:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Travis CI failed itself with the following errors:

```
Installing collected packages: filelock, toml, tox, pyflakes, entrypoints, pycodestyle, mccabe, flake8
Successfully installed entrypoints-0.3 filelock-3.0.12 flake8-3.7.9 mccabe-0.6.1 pycodestyle-2.5.0 pyflakes-2.1.1 toml-0.10.0 tox-3.14.3
WARNING: You are using pip version 19.3.1; however, version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
py3 create: /home/travis/build/OpenAPITools/openapi-generator/samples/server/petstore/python-aiohttp/.tox/py3
py3 installdeps: -r/home/travis/build/OpenAPITools/openapi-generator/samples/server/petstore/python-aiohttp/requirements.txt, -r/home/travis/build/OpenAPITools/openapi-generator/samples/server/petstore/python-aiohttp/test-requirements.txt
[...]

PR #5372, opened 2020-02-19
Use the dataType if the baseType is not set

JustinBusschau:
A fix for #4873 

### PR checklist


JustinBusschau:
I have left a single space before the `dataType`

```
{{paramName}} = {{#baseType}}{{baseType}}{{/baseType}}{{^baseType}}{{#dataType}} {{dataType}}{{/dataType}}{{/baseType}}.from_dict(connexion.request.get_json())  # noqa: E501
```

If I remove that space, the `dataType` is not rendered and we end up with the same result as raised in #4873 (the invalid `.from_dict).

Any suggestions on how to correct this are welcome.

spacether:
Thank you for this PR! Can you please add a sample model/endpoint whixh shows this working in the flask samples folder?
- the flask sample is here https://github.com/OpenAPITools/openapi-generator/tree/master/samples/server/petstore/python-flask 
- a test should be added here

JustinBusschau:
> * a test should be added here https://github.com/OpenAPITools/openapi-generator/tree/master/samples/server/petstore/python-flask/openapi_server/test

@spacether 
The specific case that this fix addresses is not covered by the current definition in `petstore.yaml`. This only occurs when you send an object or an enum as a query parameter. Neither of these cases are defined in the `petstore.yaml` - how should I proceed?

spacether:
@JustinBusschau can you make a new version of petstore.yaml which includes your use case?
Then you can regenerate samples and add python tests of it.

spacether:
@JustinBusschau it would be easier if you used the existing sh sample generation scripts. That way,
all you have to do is 
- save off a new version of the spec
- edit the sh file to point to the new spec
- add your new sample models
- run the sh file
- add the tests to the new sample files

Can you make those changes?
Doing it this way will minimize the number of files that need to be reviewed

JustinBusschau:
The difficulty I had doing it that way is that the changes I introduced to the spec are only supported by `openapi: 3.0.0` and not by `swagger: 2.0`

Editing the existing `sh` file to use my altered spec looks like moving away from supporting `swagger` specs with this generator - which was not the intention of this change. I was hoping to have this be a change that supports a feature of `openapi` without affecting those who still need to use `swagger` specs with this generator - which I need to do with some of my older-but-still-used projects. I want to leave the older `swagger`-style test (without the objects in request parameters) and add a new test that checks specifically for the new `openapi`-only feature.

spacether:
> The difficulty I had doing it that way is that the changes I introduced to the spec are only supported by `openapi: 3.0.0` and not by `swagger: 2.0`
> 
> Editing the existing `sh` file to use my altered spec looks like moving away from supporting `swagger` specs with this generator - which was not the intention of this change. I was hoping to have this be a change that supports a feature of `openapi` without affecting those who still need to use `swagger` specs with this generator - which I need to do with some of my older-but-still-used projects. I want to leave the older `swagger`-style test (without the objects in request parameters) and add a new test that checks specifically for the new `openapi`-only feature.

So this generator already has an openapi (v3) spec sample per:
- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/openapi3/python-flask-petstore-python2.sh
- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/openapi3/python-flask-petstore.sh

Both of those use petstore.yaml at `modules/openapi-generator/src/test/resources/3_0/petstore.yaml`
Editing those existing two sh files is not moving away from supporting swagger specs because that definition is v3 definition.
For reference swagger(v2) generation is done in:
- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/python-server-flask-petstore-python2.sh
- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/python-server-flask-petstore.sh
Which uses a different yaml spec to generate samples and should address your concern about supporting swagger users.

My concerns with doing it the way that you have done are:
- it is confusing to users which sample folder to look at for a v3 spec generated sample
  - samples/server/petstore/ (historically we have stored v2 samples here not v3)
  - samples/openapi3/server/petstore/python-flask
  - samples/openapi3/server/petstore/python-flask-python2
- If all other generators follow this precendent, then we could have 3+ samples folders per generator per spec which will lead to:
  - many more samples
  - increased CI runtime
  - user confusion about where to go to look at a sample of the generator output

Why not just save off your new spec at:
`https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/test/resources/3_0/flask/petsote.yaml`?

JustinBusschau:
Of course - I was missing the `openapi3` folder in `/bin` and consequently using the wrong sample generation scripts. Thanks for pointing this out @spacether :)

modules/openapi-generator/src/main/resources/python-flask/requirements.mustache
```diff
@@ -1,4 +1,4 @@
-connexion >= 2.5.0; python_version>="3.6"
+connexion >= 2.6.0; python_version>="3.6"
```
spacether: Why did you change this version?

JustinBusschau: When using 2.5.0 I was getting an error about not being able to import `FileStorage` from `werkzeug `. The only way I could find to reliably fix this was to upgrade connexion as per [this answer](https://github.com/zalando/connexion/issues/1149#issuecomment-583848140) to [this issue](https://github.com/zalando/connexion/issues/1149)

JustinBusschau:
Any advances on this @spacether ?

If there's a strong argument to be made for keeping the older version of connexion for python versions >= 3.6 I'm happy to roll back that change - addressing the `FileStorage` issue for python versions < 3.6 as you suggest in https://github.com/OpenAPITools/openapi-generator/issues/5235 is obviously not within the remit of this change

wing328:
@JustinBusschau thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

Issue #5294, opened 2020-02-11
[BUG][CSharp] Bug generating code with Enum having maxLength property


shwetashukla:
Hi 

We have an open api spec version 2.0 and we are trying to generate code in csharp . 
We have an issue in generated SDKif the property whose type is string is an enum and has maxLength specified . Please refer line 121 in attached contract .
[petstore.txt](https://github.com/OpenAPITools/openapi-generator/files/4189137/petstore.txt)

Below is the code in csharp and the one in bold throws error.

// Status (string) maxLength
            if(this.Status  != null && **this.Status.Length > 20**)
            {
                yield return new System.ComponentModel.DataAnnotations.ValidationResult("Invalid value for Status , length must be less than 20.", new [] { "Status " });
            }

The code should be like this.Status.ToString().Length.
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

PR #5378, opened 2020-02-20
Update modelGeneric.mustache

shwetashukla:
If maxlength is specified for a property type enum it there should be .ToString() appended before length check

#5294 

### PR checklist

@mandrean @jimschubert @frankyjuang @shibayan

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp/modelGeneric.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp/modelGeneric.mustache
@@ -259,22 +259,44 @@ this.{{name}} = {{#lambda.camelcase_param}}{{name}}{{/lambda.camelcase_param}};
             {{/parent}}
             {{#vars}}
             {{#hasValidation}}
+            {{#isEnum}}
+            {{#maxLength}}
+            // {{{name}}} ({{{dataType}}}) maxLength
+            if(this.{{{name}}} != null && this.{{{name}}}.ToString().Length > {{maxLength}})
+            {
+                yield return new System.ComponentModel.DataAnnotations.ValidationResult("Invalid value for {{{name}}}, length must be less than {{maxLength}}.", new [] { "{{{name}}}" });
+            }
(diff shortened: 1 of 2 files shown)


shwetashukla:
> Validating the enum value length looks very strange to me.
> In general, you should validate that the value is defined as an enum values.

> I had looked at the code generated in other languages and in those enums were checked for values. But in C# if enum is assigned an incorrect value it would either throw an exception or if marked nullable enum it would contain null. Also maxLength as per open api specification is valid tag with enums .

shibayan:
The value is always checked when converting to an enum, which means that this validation method will already contain the correct enum value.
It seems strange to treat a value successfully converted as an enum as an error with maxlength. Is it a bug of OpenAPI definition?

shwetashukla:
> The value is always checked when converting to an enum, which means that this validation method will already contain the correct enum value.
> It seems strange to treat a value successfully converted as an enum as an error with maxlength. Is it a bug of OpenAPI definition?

As per OpenAPI specification enum with maxlength is a valid combination which even i also find a bit strange , but even in that case the code generated should not have compilation errors . 
So do you suggest that we skip maxLength and minLength check for enums?

shwetashukla:
> > The value is always checked when converting to an enum, which means that this validation method will already contain the correct enum value.
> > It seems strange to treat a value successfully converted as an enum as an error with maxlength. Is it a bug of OpenAPI definition?
> 
> As per OpenAPI specification enum with maxlength is a valid combination which even i also find a bit strange , but even in that case the code generated should not have compilation errors .
> So do you suggest that we skip maxLength and minLength check for enums?

wing328:
Based on the discussion in https://github.com/OAI/OpenAPI-Specification/issues/2153, there's no need to validate maxLength, minLength for enum.

wing328:
@shwetashukla thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

PR #5394, opened 2020-02-21
[erlang] Do not export function if not defined.

cagdass:
Logic added to logic_handler.mustache so that it does not export authorize_api_key/3 when the function is not defined, leading to compilation error of *_logic_handler.erl, where * is openapi by default. 

```rebar3 shell``` under the root directory of the code stub can be run to compile.

### DIFF
--- a/modules/openapi-generator/src/main/resources/erlang-server/logic_handler.mustache
+++ b/modules/openapi-generator/src/main/resources/erlang-server/logic_handler.mustache
@@ -1,7 +1,14 @@
 -module({{packageName}}_logic_handler).
 
 -export([handle_request/4]).
+{{#authMethods}}
+	{{#isApiKey}}
 -export([authorize_api_key/3]).
+	{{/isApiKey}}
+{{/authMethods}}
+{{^authMethods}}
+-export([authorize_api_key/3]).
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wing328** [comment]
cc @tsloughter (2017/11) @jfacorro (2018/10) @robertoaloi (2018/10)

File: modules/openapi-generator/src/main/resources/erlang-server/logic_handler.mustache [hunk 1]
```diff
@@ -1,7 +1,14 @@
 -module({{packageName}}_logic_handler).
 
 -export([handle_request/4]).
+{{#authMethods}}
+	{{#isApiKey}}
 -export([authorize_api_key/3]).
+	{{/isApiKey}}
+{{/authMethods}}
```
wing328: This works if the spec contains 1 API key definitions, If it has more than 1, then it won't work. I'll switch it to `hasApiKeyAuth` in another PR. Thanks for your contribution.

wing328:
@cagdass thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5398, opened 2020-02-21
[Java][Spring][1975] Add OAuth2 Preauthorize annotations based on scope

### LINKED ISSUE #1975: [REQ] Spring Server generators should allow for adding OAuth2 Preauthorize annotations based on scope
Using OAuth JWT scopes and Spring Security's PreAuthorize annotation requires the PreAuthorize annotation to be on the same method as the Mapping related annotation.

With the current generators (Java/Kotlin), the Mapping annotation is in generated code, so manually adding the PreAuthorize annotation can't be done without modifying generated code.

It would be good if the generator had a flag for Spring Security, and would then leverage the OAuth2 security settings to add @PreAuthorize() annotations to the Controller class methods with the Mapping related annotations.

This is dependent on #392 being corrected first.

ersinciftci:
Fixes #1975 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/JavaSpring/api.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/api.mustache
@@ -17,6 +17,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 {{/jdk8-no-delegate}}
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
 {{#useBeanValidation}}
 import org.springframework.validation.annotation.Validated;
 {{/useBeanValidation}}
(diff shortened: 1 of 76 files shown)


ersinciftci:
@wing328 @cbornet I tried to make the PreAuthorize import conditional but failed, any help would be appreciated.

cbornet:
I think this doesn't work if there's multiple securities defined for an operation (eg. Oauth + basic).

ersinciftci:
@cbornet I just tested with an OpenAPI spec that has both OAuth2 and Basic, and seems to be working, because it has `{{#isOAuth}}` so I think the extra Basic security item doesn't affect it. But I can't figure out for what use case the tests are failing.

cbornet:
What I mean is that an endpoint that has both basic and oauth should be accessible by basic OR oauth. Here I think if you try to access by basic auth, the pre-authorization will fail and you will be rejected.

wing328:
> I think this doesn't work if there's multiple securities defined for an operation (eg. Oauth + basic).

@cbornet good catch. I think such a case rarely occurs (based on my experience dealing with issues reported to this repo). We can document it as a known issue for the time being and work on it when there's a demand to fix it.

Walliee:
I just stumbled here trying to find `@PreAuthorize` support in OpenAPI Generator. 
I am looking to avoid custom templates in our project and I think this feature would get us there. 
 
Anything I can do to help?

ersinciftci:
Hi @Walliee , sorry I didn't have much time recently, your help would be greatly appreciated. 

Firstly, not sure if the tests are failing because of the use case @cbornet mentioned, so that's pending some investigation (though @wing328 mentioned that we can ignore that use case for now, we still need to fix the tests). 

One other thing is this change would fail for people that don't use Spring Security because it imports a class from there, so maybe this should be a configuration that people can turn on (like addPreAuthorize = true).

Thanks

Walliee:
Yes; of course. The pointers you gave make sense. Thanks!

@nhomble and I looked at this briefly today and looks like we can handle bearer auth too. We’ll hopefully have a PR out tomorrow.

nhomble:
Awesome stuff @ersinciftci, we are still polishing a little bit (and I need to regen the samples), but here is a [sneak peak](https://github.com/nhomble/openapi-generator/tree/gh-1975)

ersinciftci:
@nhomble looks great! `useSpringSecurity` sounds much better than `addPreAuthorize` :)

ersinciftci:
@nhomble looks much better than this one, thanks! (I'm closing this one)

Issue #3650, opened 2019-08-15
[REQ] sbt plugin for OpenAPI Generator


wing328:
We already have maven and gradle plugins. As discussed in in the slack channels, there's also demand for an sbt plugin.

If anyone would like to contribute an sbt plugin for openapi generator, please reply to let us know.

YanDoroshenko:
I already have a [barely] working prototype, working on improving it.

jimschubert:
@YanDoroshenko let me know if you need any assistance and I'd be happy to check it out. I have experience writing SBT plugins.

YanDoroshenko:
@jimschubert Is there a way to define an optional settingKey in SBT?

Example:
Api package setting is not mandatory (generator has some default value). I can declare a `settingKey[Option[String]]` and initialize it to `None` as a default. Then if it's present, I update the generator setting. But that means that plugin user has to specify it with Option like `openapiApiPackage := Some("org.something.api")`. Is there a way around it?

Thanks.

jimschubert:
@YanDoroshenko another way to do this would be to have a string and initialize it to a null. Then when you use the string, 

```
Option(yourKeySetting) match {
  case Some(value) => // set it
  case None => // ignore it
}
```

YanDoroshenko:
@jimschubert Nope:
```
[info] [error] Setting value cannot be null: ProjectRef(uri("file:/tmp/sbt_cf846a14/simple/"), "simple") / modelNameSuffix
```

jimschubert:
@YanDoroshenko I could've sworn I had done this with null, but maybe I used an empty string and filtered on nonEmpty values in the past. I don't think the usage of assigning `Some(value)` is bad, as it clearly indicates that you're setting an optional value. This is done using `Option[String]` in some plugins (see https://github.com/sbt/sbt-autoversion/blob/2f7798d8de68cf2b5425423b2c9b702df2cdb5f4/src/main/scala/autoversion/Keys.scala#L10).

If it's possible to get the current setting for a property, I guess you could also create a `settingKey[String => String]` and bind this to the identity function by default.

YanDoroshenko:
https://github.com/upstart-commerce/sbt-openapi-generator

wing328:
@YanDoroshenko is it correct to say that https://github.com/upstart-commerce/sbt-openapi-generator is ready for OpenAPI Generator to try it out and provide feedback?

YanDoroshenko:
[...]

PR #5406, opened 2020-02-23
sbt plugin for OpenAPI Generator

chameleon82:
Regarding to #3650 sbt plugin for  OpenAPI Generator added

Few concerns about this PR:
- [ ] need to define which repositories this artefact will be published. At least, for sbt community it should be published to `https://bintray.com/sbt/sbt-plugin-releases`, I believe @eed3si9n can help us on it. (continued here https://github.com/OpenAPITools/sbt-openapi-generator/issues/5)
- [x] plugin name convention. Following OpenApi artefact should be named as `openapi-generator-sbt-plugin`, but [sbt plugin best practices](https://www.scala-sbt.org/1.x/docs/Plugins-Best-Practices.html) recommends to name it as `sbt-openapi-generator`. 
Resolution: `sbt-openapi-generator` as [separate project](https://github.com/OpenAPITools/sbt-openapi-generator)
- [x] settings name convention. Right now all settings haven't prefix, but sbt has only global domain for all settings which can be conflicted with other sbt plugins. Should we change it following [sbt best practices](https://www.scala-sbt.org/1.x/docs/Plugins-Best-Practices.html)? Sbt recommends to reuse default settings if possible, but not much settings we can. Alternate way to keep original names for settings is to put all openapi settings in separate `openapiSettings` case class, but in this case `sbt` will loose description of each setting (which probably ok)
Resolution: `openApi` prefix
- [x] maven integration. As i understand it is possible to build plugin with maven (`sbt` integration tests still require sbt anyway) and integrate this module as part of project, but have no experience on it. 
Resolution: don't needed as moved to [separate project](https://github.com/OpenAPITools/sbt-openapi-generator)
- [x] should plugin support `src_managed` directory for generated files? 
Resolution: up to user. will be better described in README

### PR checklist

- [x] Copy the (technical committee)[https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee] to review the pull request if your PR is targeting a particular programming language.

@jimschubert @wing328 @YanDoroshenko


jimschubert:
@chameleon82 I think this looks great. I wonder if it would be better to include this as a separate repository under OpenAPI Tools? I think that would address some concerns that might arise from adding another plugin here:

* We can follow SBT Plugin best practices, such as not conflicting with the Maven/Gradle naming conventions
* We can configure deployments in a way that don't interfere with our Maven/Sonatype deploys (to avoid confusion that we've had with like the Gradle POM wrapper)
* We don't have to integrate with Maven. I'd much rather stick with SBT-only in this case.
* We won't introduce additional build time to the core project

I think the separate repo would be good as well because it would allow us to setup GitHub Workflows for build/release. We could also create a cron workflow which checks OpenAPITools/openapi-generator daily for new releases and automatically prepare a new release Pull Request of the SBT Plugin. I'm planning to do something similar for my IntelliJ Plugin as well.

As far as your other questions, I think an `openApi` prefix on settings makes sense. Also, I'm not sure about the `src_managed` question… are you asking if we should default the output directory to `(sourceManaged in Compile).value / "out"`? I think it would be a good default but best to leave this up to the user, maybe just documenting it in the README.

I have a few questions/comments about the implementation. Please forgive me, because it's been about 2.5 years since I last write an SBT plugin.

* In trait OpenApiCodegen, you create a new configuration called `OpenApiCodegen`. I thought it was considered bad practice to create a new config in most cases, is there a specific reason for this or could we extend another config or even simplify as `val OpenApiCodegenConfig = config("openapi")`?
* For defaults, I think we'll also need to separate some like Compile and Test, for example:
  ```
    keyForOutputDir in Compile := (sourceManaged in Compile).value / "out",
    keyForOutputDir in Test := (sourceManaged in Test).value / "out",
  ```
* Usually we'd see an exported sequence of settings which define the keys and their defaults for the plugin, and these are applied to the configuration specifically (see [scrooge](https://github.com/twitter/scrooge/blob/80496d0a14050646038e3890ce09a534df88d38d/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala#L301) for example). Is that done differently here because of the Config file support?

chameleon82:
@jimschubert i've implemented new config as sbt doesn't have default `Generate` or something like this.
I considered next scenario:
- able to build sources as separate project(module) with `Codegen` config
- compile project/module with generated files
- keep generated files in git to able track changes
Make generation under `Compile` config also make sense in most cases and i think it possible to achieve above goals with `Compile` config. Will try to rethink about usage scenarios. 
* yes, settings applied differently because of config file support. It make difficult to override value when default value presents in config as need to apply config file settings first to configurator and then check if setting value not equal to default and not equal to configurator value and have to repeat it for every supported configuration. It will be easier to manage if `codegen` default settings will be public available. For now most values are private.

chameleon82:
@jimschubert revised the code, seems `OpenApiCodegen` config is redundant. I deleted it and renamed settings with prefix.
`sourceManaged` seems a bit complicated, i tried to use options 
```
Compile / sourceGenerators += openApiGenerateTask.taskValue.map{ 
  _.map{ file => file.filter(...).getAbsoluteFile }
}
```
and 
```
val generatedSources = mappings in (Compile, packageSrc) ++= { ... }
```
But as main purpose of plugin is to generate whole project it unclear for me how to `reload` dependencies from generated `build.sbt` file at compilation time.
It could be simple as `sbt "openApiGenerate; reload; compile"` in build flow.

`sourceManaged` makes sense when user don't want to generate `build.sbt` and use predefined instead. In that case `configFile.outputDir` setting must be ignored or sbt `openApiOutputDir` setting changed to a task. But i got confused to separate it from normal project compilation and need help to implement this scenario.

jimschubert:
@chameleon82 I've setup https://github.com/OpenAPITools/sbt-openapi-generator

Should we move your work there?

jimschubert:
Closing this as we created https://github.com/openapitools/sbt-openapi-generator

I'll leave #3650 to address documenting the new plugin

Issue #5278, opened 2020-02-11
[BUG][go-experimental] Explicit nulls support is broken


bkabrda:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The current support for explicit nulls, as modified in PR https://github.com/OpenAPITools/openapi-generator/pull/3989 by @shaxbee (and approved by myself) is broken. I'm opening this issue to discuss the best approach to fix it - I'm ok with implementing the solution for it.

For any given nullable and not required field, we currently generate code like:

```go
type V struct {
    A *NullableInt32 `json:"a,omitempty"`
}
```

The problem right now is that when an API sends back an explicit-null field, it's not deserialized properly as an explicit null. This is because for a pointer field, Go unmarshalling machinery *doesn't even invoke the custom unmarshall method*, it basically considers explicit null to be the same as the field not being present at all.

So there are 2 options:
* Create custom unmarshaller for `V`
* Don't make `A` a pointer

Following the reasoning in #3989 about not overloading marshal/unmarshal for these potentially large structures, let's try going with not making `A` a pointer, hence getting:

```go
type V struct {
    A NullableInt32 `json:"a,omitempty"`
}
```

This would make deserialization work, as the custom unmarshaller for `A` would always run, but there's another problem - the custom marshaller for `A` would always run when serializing. IOW, now there's no way to *not* serialize this field. But the field is supposed to be optional, so we'll have to create a custom marshaller for `V`.

So based on the above (if my thinking is correct), we have two options:
* Leave the field to be a pointer and provide custom unmarshaller
* Make the field non-pointer and provide custom marshaller

I'm leaning towards the latter, as I think it's easier. I've been playing with how this could work and came up with [1] - this solution changes how the `Nullable*` structures work a bit, but it also brings it closer to how this is implemented for Java/jackson clients, using [2].

If everyone agrees, I'll send a PR.

[...]

PR #5414, opened 2020-02-24
[go-experimental] Fix nullable support

bkabrda:
This fixes [1] while also making the support for explicit nulls much more similar to what we have for Java clients with jackson-databind-nullable.

Notes:
* Support for container fields is not implemented due to lack of generics in Go. This might become possible with more work, but I didn't want to further complicate this PR.
* The performance of generated clients might get hit a bit because of the newly added custom marshalling functions. This is a tradeoff taken to achieve correctness of the implementation (see the description of [1] for reasoning of why this is needed).

[1] https://github.com/OpenAPITools/openapi-generator/issues/5278

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
+++ b/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
@@ -105,6 +105,9 @@ func New{{classname}}WithDefaults() *{{classname}} {
 {{#vars}}
 {{#required}}
 // Get{{name}} returns the {{name}} field value
+{{#isNullable}}
+// If the value is explicit nil, the zero value for {{vendorExtensions.x-basetype}} will be returned
+{{/isNullable}}
 func (o *{{classname}}) Get{{name}}() {{vendorExtensions.x-basetype}} {
 	if o == nil {{#isNullable}}{{^isContainer}}|| o.{{name}}.Get() == nil{{/isContainer}}{{/isNullable}} {
 		var ret {{vendorExtensions.x-basetype}}
--- a/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
+++ b/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
@@ -262,7 +262,7 @@ func (v Nullable{{classname}}) Get() *{{classname}} {
 	return v.value
 }
 
-func (v Nullable{{classname}}) Set(val *{{classname}}) {
+func (v *Nullable{{classname}}) Set(val *{{classname}}) {
 	v.value = val
 	v.isSet = true
 }
@@ -271,7 +271,7 @@ func (v Nullable{{classname}}) IsSet() bool {
 	return v.isSet
 }
 
-func (v Nullable{{classname}}) Unset() {
+func (v *Nullable{{classname}}) Unset() {
 	v.value = nil
 	v.isSet = false
 }
--- a/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
+++ b/modules/openapi-generator/src/main/resources/go-experimental/model.mustache
@@ -138,7 +138,7 @@ func (o *{{classname}}) Get{{name}}Ok() ({{dataType}}, bool) {
 		return ret, false
 	}
 {{#isNullable}}
-    return o.{{name}}, o.{{name}}.IsSet()
+    return o.{{name}}, {{#isContainer}}o.{{name}} == nil{{/isContainer}}{{^isContainer}}o.{{name}}.IsSet(){{/isContainer}}
 {{/isNullable}}
 {{^isNullable}}
     return *o.{{name}}, true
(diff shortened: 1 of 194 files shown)


modules/openapi-generator/src/main/resources/go-experimental/model.mustache
```diff
@@ -139,11 +156,50 @@ func (o *{{classname}}) Has{{name}}() bool {
...
+    {{^isContainer}}
+    if {{#required}}true{{/required}}{{^required}}o.{{name}}.IsSet(){{/required}} {
+        toSerialize["{{baseName}}"] = o.{{name}}.Get()
+    }
+    {{/isContainer}}
+    {{/isNullable}}
+    {{! if argument is not nullable, don't set it if it is nil}}
+    {{^isNullable}}
+    if {{#required}}true{{/required}}{{^required}}o.{{name}} != nil{{/required}} {
+        toSerialize["{{baseName}}"] = o.{{name}}
```
jirikuncar:
```suggestion
        toSerialize["{{baseName}}"] = &o.{{name}}
```

wing328:
@bkabrda I notice that some line uses 4-space instead of tabs for indentation. What about using tabs instead?

e.g. https://github.com/OpenAPITools/openapi-generator/pull/5414/files#diff-e1c002d29e771fb5b795a039f401a360R150

wing328:
@bkabrda I may have missed it. What about adding some tests under `samples/client/petstore/go-experimental/go-petstore/`?

bkabrda:
@wing328 yeah, I'll fix the indentation, good point.

As for the tests, I don't see any nullables in the openapi 2.0 spec sample, but there are some in the openapi 3.0 spec sample, so I'll try to add some tests to `samples/openapi3/client/petstore/go-experimental/` if that sounds ok.

bkabrda:
@wing328 so I fixed the tabs/spaces issue. For nullables, it seems that there is no endpoint at the swagger petstore that would be using this feature. Would it be sufficient to just add a unit test that would deserialize/serialize an entity and make sure everything works fine (without hitting an API endpoint)? Thanks!

modules/openapi-generator/src/main/resources/go-experimental/model.mustache
```diff
@@ -111,26 +123,31 @@ func (o *{{classname}}) Set{{name}}(v {{dataType}}) {
...
 // Get{{name}}Ok returns a tuple with the {{name}} field value if set, zero value otherwise
 // and a boolean to check if the value has been set.
 func (o *{{classname}}) Get{{name}}Ok() ({{dataType}}, bool) {
-	if o == nil || o.{{name}} == nil {
+	if o == nil {{^isNullable}}|| o.{{name}} == nil{{/isNullable}} {
 		var ret {{dataType}}
 		return ret, false
 	}
+{{#isNullable}}
+	return o.{{name}}, {{#isContainer}}o.{{name}} == nil{{/isContainer}}{{^isContainer}}o.{{name}}.IsSet(){{/isContainer}}
```
zippolyte: Since we return if the value inside the nullable is set, why don't we return the value inside the nullable too directly ? Might simplify usage of nullables ?
Haven't thought about it deeply, just throwing that out there. Let me know what you think

bkabrda: I think it would then be problematic what `Get{{name}}` should return - because it would return null both if there's explicit null and if there's not, so it wouldn't make much sense to keep that method for nullables, which would lead to inconsistency in the API of the class. Also, getting the whole nullable structure directly lets you call e.g. `Unset()` on it... I do see your point, but I think this is a matter of preference, YMMV :)

bkabrda: Hmm, but it's true that doing this would make it somewhat easier to add implementation of nullables for container types in the future. Let me try to experiment with the idea a bit.

modules/openapi-generator/src/main/resources/go-experimental/model.mustache
```diff
@@ -57,167 +56,319 @@ type {{classname}} struct {
...
 {{/isContainer}}
 {{/defaultValue}}
 {{/vars}}
-    return &this
+	return &this
 }
 
 {{#vars}}
 {{#required}}
 // Get{{name}} returns the {{name}} field value
```
zippolyte: Let's document the caveat in case of explicit null

modules/openapi-generator/src/main/resources/go-experimental/model_doc.mustache
```diff
@@ -36,37 +36,40 @@ but it doesn't guarantee that properties required by API are set
...
 and a boolean to check if the value has been set.
 
+### Set{{name}}
+
+`func (o *{{classname}}) Set{{name}}(v {{vendorExtensions.x-basetype}})`
+
+Set{{name}} sets {{name}} field to given value.
+
+{{#isNullable}}
 ### Has{{name}}
```
zippolyte: `Has...` is defined for every non required types, not just for nullables I believe.

wing328:
@bkabrda thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5422, opened 2020-02-24
Fix #5420 add headers from configuration object

### LINKED ISSUE #5420: [BUG] Can't set headers from configuration object Typescript Axios
Typescript Axios.

Can't set custom headers from configuration object.
`localVarRequestOptions.headers = {...localVarHeaderParameter, ...options.headers};` in `ApiAxiosParamCreator`
In this row you must get headers from `...baseOptions.headers,` for fix this. If I can find this, I will make pull request.

nibytes:
Small, but so important fix. Configuration class has baseOptions key, this is object for set any axios options, but headers from this object was ignored because they shadowing, I fixed this.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -189,7 +189,8 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             localVarUrlObj.query = {...localVarUrlObj.query, ...localVarQueryParameter, ...options.query};
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
-            localVarRequestOptions.headers = {...localVarHeaderParameter, ...baseOptions.headers, ...options.headers};
+            let headersFromBaseOptions = baseOptions && baseOptions.headers ? baseOptions.headers : [];
+            localVarRequestOptions.headers = {...localVarHeaderParameter, ...headersFromBaseOptions, ...options.headers};
     {{#hasFormParams}}
             localVarRequestOptions.data = localVarFormParams{{#vendorExtensions}}{{^multipartFormData}}.toString(){{/multipartFormData}}{{/vendorExtensions}};
     {{/hasFormParams}}
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -189,7 +189,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             localVarUrlObj.query = {...localVarUrlObj.query, ...localVarQueryParameter, ...options.query};
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
-            let headersFromBaseOptions = baseOptions && baseOptions.headers ? baseOptions.headers : [];
+            let headersFromBaseOptions = baseOptions && baseOptions.headers ? baseOptions.headers : {};
             localVarRequestOptions.headers = {...localVarHeaderParameter, ...headersFromBaseOptions, ...options.headers};
     {{#hasFormParams}}
             localVarRequestOptions.data = localVarFormParams{{#vendorExtensions}}{{^multipartFormData}}.toString(){{/multipartFormData}}{{/vendorExtensions}};
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -189,7 +189,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             localVarUrlObj.query = {...localVarUrlObj.query, ...localVarQueryParameter, ...options.query};
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
-            localVarRequestOptions.headers = {...localVarHeaderParameter, ...options.headers};
+            localVarRequestOptions.headers = {...localVarHeaderParameter, ...baseOptions.headers, ...options.headers};
     {{#hasFormParams}}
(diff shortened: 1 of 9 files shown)


ivannnnnnnnnn:
I have the same issue. Approve this request

wing328:
cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)

wing328:
@nibytes thanks for the PR. Please run `./bin/typescript-axios-petstore-all.sh` to update the samples.

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -189,7 +189,7 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             localVarUrlObj.query = {...localVarUrlObj.query, ...localVarQueryParameter, ...options.query};
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
-            localVarRequestOptions.headers = {...localVarHeaderParameter, ...options.headers};
+            localVarRequestOptions.headers = {...localVarHeaderParameter, ...baseOptions.headers, ...options.headers};
```
amakhrov: - Based on the code above, it seems that `baseOptions` can be undefined - probably need null check here before accessing `baseOptions.headers`.
- Shouldn't `localVarHeaderParameter` have priority over `baseOptions.headers`? Right now it looks like base options can override customizations specific to a particular api method

nibytes: 1) I will fix this
2) I dont know, but I think localVarHeaderParameter will override baseOptions, and you can't change variables in localVarHeaderParameter in no way. So my option is allow override localVarHeaderParameter by baseOptions setted by user for all, and they overriding by options from request.

nibytes:
Guys, how I can add new changes for this PR?

akehir:
> In my case, I need set X-CSRF-Token header for all requests, but I can't do it in one place. Setting this token in options of every request, very repeatable.

Unfortunately I'm not deeply familiar with axios, but  couldn't you use an axios request [interceptor](https://github.com/axios/axios#interceptors)  to set the xsrf token for all request in one central location?

nibytes:
> Unfortunately I'm not deeply familiar with axios, but couldn't you use an axios request [interceptor](https://github.com/axios/axios#interceptors) to set the xsrf token for all request in one central location?

Headers will not override by ApiParamsCreator? Anyway any headers from Configuration.baseOptions not are include in request headers. So we need fix this.

nibytes:
> thanks for the PR. Please run `./bin/typescript-axios-petstore-all.sh` to update the samples.

Sorry I have not linux system for do it now. Anyone can help?

macjohnny:
@nibytes the windows-version of the sample generation script is located in `bin/windows`

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -189,7 +189,8 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
             localVarUrlObj.query = {...localVarUrlObj.query, ...localVarQueryParameter, ...options.query};
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
-            localVarRequestOptions.headers = {...localVarHeaderParameter, ...options.headers};
+            let headersFromBaseOptions = baseOptions && baseOptions.headers ? baseOptions.headers : [];
```
macjohnny: I think this should initialize to an empty object, right?

nibytes: Yes

nibytes:
@wing328 Please, can you tell me when this will released?

macjohnny:
it seems that the next release will be 4.3.0 on 2020-03-20, see https://github.com/OpenAPITools/openapi-generator/milestones

amakhrov:
@nibytes also you might want to use [openapi-generator-cli:latest](https://hub.docker.com/r/openapitools/openapi-generator-cli/tags?page=1&name=latest) - which is frequently built off `master` - so you don't have to wait till the next version release

wing328:
@nibytes thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PieterBracke:
@wing328 I am having a similar issue concerning X-XSRF-TOKEN, currently I am using an interceptor to read the token from my cookie and append it to the headers on outgoing requests? I assume that if we wish to do it like this PR suggested we'd need to update the configuration object when the X-XSRF-TOKEN got updated 2? Seems to me using an interceptor in that case and extracting the token the cookie is much easier? I did notice that using the regular httpClient and using the .Post method (in Angular) does append the X-XSRF-TOKEN by default witouth any problems, it is only when using a service generated by openapi generator that this default behavior is overriden by your setting of the headers. Is there a way to not override some tokens (preferrably customizable in settings?) for instance 

`  importProvidersFrom(
      HttpClientModule,
      BrowserAnimationsModule,
      ApiModule.forRoot(
        () =>
          new Configuration({
            basePath: `${environment.BFF_URL}`,
            **maintainHeaders: ['X-XSRF-TOKEN']**
          })
      ),`

Or is that not possible?

XhstormR:
> @wing328 I am having a similar issue concerning X-XSRF-TOKEN, currently I am using an interceptor to read the token from my cookie and append it to the headers on outgoing requests? I assume that if we wish to do it like this PR suggested we'd need to update the configuration object when the X-XSRF-TOKEN got updated 2? Seems to me using an interceptor in that case and extracting the token the cookie is much easier? I did notice that using the regular httpClient and using the .Post method (in Angular) does append the X-XSRF-TOKEN by default witouth any problems, it is only when using a service generated by openapi generator that this default behavior is overriden by your setting of the headers. Is there a way to not override some tokens (preferrably customizable in settings?) for instance
> 
> ` importProvidersFrom( HttpClientModule, BrowserAnimationsModule, ApiModule.forRoot( () => new Configuration({ basePath:`${environment.BFF_URL}`, maintainHeaders: ['X-XSRF-TOKEN'] }) ),`
> 
> Or is that not possible?

+1, need config for `X-XSRF-TOKEN` header

PR #5423, opened 2020-02-25
Add pythonSrcRoot option to python servers (aiohttp/flask/blueplanet) to support src/ layout projects [and reenable/fix all python server tests]

cognifloyd:
This will allow the python project to be in a subdirectory (as specified
in pythonSrcRoot). That could mean following the src layout with sources
under src/ or lib/. Multi-language projects might use a sub directory
like python/, or whatever makes sense for the project.

By default, the pythonSrcRoot is "", meaning the existing behavior is
the default.

### PR checklist


cognifloyd:
Copying python TC:
> @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @slash-arun (2019/11) @spacether (2019/11)

Do I need a sample python server for every variant? I've only include the aiohttp sample, but I can add samples for flask and bluplanet if desired.

spacether:
Thank you for the PR! One sample is fine, no need to do a bunch.

cognifloyd:
I don't see how the dart and swift test failures have anything to do with this PR.

wing328:
Those errors should have been fixed in the master. Please merge the latest master into your branch to give it another try.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
```diff
@@ -165,6 +167,8 @@ public PythonAbstractConnexionServerCodegen(String templateDirectory, boolean fi
                 defaultValue("8080"));
         cliOptions.add(CliOption.newBoolean(USE_NOSE, "use the nose test framework").
                 defaultValue(Boolean.FALSE.toString()));
+        cliOptions.add(CliOption.newBoolean(SRC_LAYOUT, "put python module under src/").
```
spacether: @cognifloyd would there ever be a use case where we would want to name the folder something other than src?
If so why don't we make this parameter more general like sourceFolderName type string.
What do you think?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
```diff
@@ -304,7 +315,20 @@ public String escapeReservedWord(String name) {
      */
     @Override
     public String apiFileFolder() {
-        return outputFolder + File.separator + apiPackage().replace('.', File.separatorChar);
+        String pkgPath = apiPackage().replace('.', File.separatorChar);
+        if (this.srcLayout) {
+            pkgPath = "src" + File.separator + pkgPath;
+        }
```
spacether: Why not edit outputFolder directly when we assign it first?
Then we don't need to put these fi blocks in twice.
Up in processOpts:
```
        if (this.srcLayout) {
            outputFolder = "src" + File.separator + "src";
        }
```

modules/openapi-generator/src/main/resources/python-flask/tox.mustache
```diff
@@ -6,4 +6,4 @@ deps=-r{toxinidir}/requirements.txt
      -r{toxinidir}/test-requirements.txt
 
 commands=
-   {{^useNose}}pytest --cov={{{packageName}}}{{/useNose}}{{#useNose}}nosetests{{/useNose}}
\ No newline at end of file
+   {{^useNose}}pytest --cov={{#srcLayout}}src/{{/srcLayout}}{{{packageName}}}{{/useNose}}{{#useNose}}nosetests{{/useNose}}
```
spacether: Did you test this with useNose=True also?
You might need to hard code in a path for nosetests now that the code is deeper.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
cognifloyd: `src/` is pretty standard, but I suppose I have seen projects that do `lib/` as well.
So, an empty string (or maybe a `.`) means no subdir and a string == a subdir? hmm.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
cognifloyd: Because there are still supporting files (like Readme, .gitignore, etc) that still need to be in the root dir. So we need to selectively move only the source files under the src/ dir.

modules/openapi-generator/src/main/resources/python-flask/tox.mustache
cognifloyd: hmm. I don't know nose.
But from a quick google, it looks like it will use lib and src directories by default.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
cognifloyd: hmm. Here's [someone](https://stackoverflow.com/questions/30461071/configuring-the-directories-that-nosetests-looks-into) that puts his source in a `python` dir. So, there may be use cases where that would be helpful. Like in multi-language projects or where you include client and server libs in the same repo.

So, the name `srcLayout` only makes sense as a boolean. What would an option be for selecting an alternate directory for the python source files? `pythonSrcRoot`? `srcLayoutDir`? or would you do two vars, one for the dir to use and a boolean to enable it?

spacether: How about the string pythonSrcRoot? And make it an empty value by default.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
spacether: Ah, in that case why not use your new variable
pythonSrcRoot? Then it is only defined once and it will include the file separator when you pass in the value "src/"

cognifloyd:
I just finished reworking this to use `pythonSrcRoot` option (a path) instead of `srcLayout` (a boolean).
I squashed into meaningful commits and rebased this on master.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
cognifloyd: done

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
cognifloyd: done

modules/openapi-generator/src/main/resources/python-flask/tox.mustache
cognifloyd: If you can point me to an option for nosetests to set the python sources root, I'm happy to add that.
From what I can tell, you have to add it inside the test code: https://stackoverflow.com/a/30501263
But there isn't any test code to work with. It looks like all of the tests are written with pytest in mind (from my limited understanding of nose).
The things that take `useNose` into account are `tox.mustache` and `test-requirements.mustache`. It looks like the flask template for travis expects to run tests with nosetests without regards to `useNose`.

De-tangling the nose stuff would probably be good for a separate PR (by someone who actually uses / understands nose).

cognifloyd:
@spacether how does this look now?

spacether:
@cognifloyd can you add the following files to your new folder `samples/server/petstore/python-aiohttp-srclayout`
- dev-requirements.txt
- Makefile
- test_python3.sh
- pom.xml
- in the pom.xml at the root of the repo please add `<module>samples/server/petstore/python-aiohttp-srclayout</module>`

Why do I get this error when I try to run `make test` in your sample folder?
Can you fix it?
```
ImportError while loading conftest '/Users/justin/programming/openapi-generator/samples/server/petstore/python-aiohttp-srclayout/tests/conftest.py'.
tests/conftest.py:5: in <module>
    import connexion
.tox/py3/lib/python3.7/site-packages/connexion/__init__.py:3: in <module>
    from .apis import AbstractAPI  # NOQA
.tox/py3/lib/python3.7/site-packages/connexion/apis/__init__.py:1: in <module>
    from .abstract import AbstractAPI  # NOQA
.tox/py3/lib/python3.7/site-packages/connexion/apis/abstract.py:9: in <module>
    from ..operations import make_operation
.tox/py3/lib/python3.7/site-packages/connexion/operations/__init__.py:1: in <module>
    from .abstract import AbstractOperation  # noqa
.tox/py3/lib/python3.7/site-packages/connexion/operations/abstract.py:11: in <module>
    from ..decorators.response import ResponseValidator
.tox/py3/lib/python3.7/site-packages/connexion/decorators/response.py:12: in <module>
    from .validation import ResponseBodyValidator
.tox/py3/lib/python3.7/site-packages/connexion/decorators/validation.py:9: in <module>
    from werkzeug import FileStorage
E   ImportError: cannot import name 'FileStorage' from 'werkzeug' (/Users/justin/programming/openapi-generator/samples/server/petstore/python-aiohttp-srclayout/.tox/py3/lib/python3.7/site-packages/werkzeug/__init__.py)
ERROR: InvocationError for command /Users/justin/programming/openapi-generator/samples/server/petstore/python-aiohttp-srclayout/.tox/py3/bin/pytest --cov=src/openapi_server (exited with code 4)
```

cognifloyd:
OK I am adding those files now. I was confused about why they were in the other python server samples when the generator clearly didn't create them.

cognifloyd:
https://github.com/OpenAPITools/openapi-generator/blob/9adfedbfbb45c7059e64b3d9a285710bded4fb62/pom.xml#L1228-L1229

As far as that error goes, apparently its not just this test experiencing that problem:
https://github.com/OpenAPITools/openapi-generator/issues/5235

That has the smell of a transient dep that changed somewhere. I'll take a look.

modules/openapi-generator/src/main/resources/python-flask/requirements.mustache
```diff
@@ -1,7 +1,15 @@
 connexion >= 2.6.0; python_version>="3.6"
+# connexion 2.6 dropped support for python < 3.6
 connexion >= 2.3.0; python_version=="3.5"
 connexion >= 2.3.0; python_version=="3.4"
+{{#supportPython2}}
 connexion == 2.4.0; python_version<="2.7"
+{{/supportPython2}}
+# werkzeug is a dep of connexion. They need to record their dep.
```
spacether, line 8: How about the simpler
```
# connexion requires werkzeug but connexion < 2.4.0 does not install werkzeug
# we must peg werkzeug versions below to fix connexion
# https://github.com/zalando/connexion/pull/1044
werkzeug == 0.16.1; python_version=="3.4"
werkzeug == 0.16.1; python_version=="3.5"
```
Note: for python versions ==2.7 and >= 3.6 we should not need to install werkzeug
because per https://github.com/zalando/connexion/commit/7111bc91ee9ffcdce1899c06205e61f634972228
this is fixed in connexion 2.4.0, and 2.6.0 which we are installing for those versions of python

cognifloyd: I applied your suggestion, plus I simplifeid it further using the or syntax from PEP 496: https://www.python.org/dev/peps/pep-0496/#examples

cognifloyd:
Well that was a lot more than I planned on changing. So much for a non-invasive feature addition. I have now reenabled and fixed the tests for all of the flask and aiohttp servers.

cognifloyd:
Travis missed posting results (passed). 
:heavy_check_mark: https://travis-ci.org/github/OpenAPITools/openapi-generator/builds/670859311

cognifloyd:
:heavy_check_mark: Travis neglected to report that the build passed.

modules/openapi-generator/src/main/resources/python-aiohttp/requirements.mustache
```diff
@@ -1,3 +1,3 @@
-connexion[aiohttp,swagger-ui] == 2.0.2
-swagger-ui-bundle == 0.0.2
-aiohttp_jinja2 == 1.1.0
+connexion[aiohttp,swagger-ui] == 2.6.0
```
spacether, line 1: This generator states that is supports  python 3.5 but this version of connexion does not work in python 3.5
Can you add the connexion support for python >= 3.5 here?

cognifloyd:
ok. I will adjust the deps. I think I'll go for a >= dep like the python-flask generator instead of pinning.

cognifloyd:
Travis passed. Shippable error is unrelated (a failure to install Kotlin).

modules/openapi-generator/src/main/resources/python-aiohttp/requirements.mustache
```diff
@@ -1,3 +1,4 @@
-connexion[aiohttp,swagger-ui] == 2.6.0
+# 2.3 is the last version that supports python 3.5
+connexion[aiohttp,swagger-ui] >= 2.3.0
```
spacether, line 2: Hey so connexion only limits installation versions by supported versions of python in 3.6.0
per https://github.com/zalando/connexion/commit/f31ac615bfc8a3c696c5ebe53f3c57e5234ff6f8
Please:
- pin the connexion version for python 3.5
- pin the werkzeug version for python 3.5 (per your other work)
- pin `connexion >= 2.6.0; python_version>="3.6"` to cover higher versions of python

modules/openapi-generator/src/main/resources/python-flask/requirements.mustache
```diff
@@ -1,10 +1,20 @@
 connexion >= 2.6.0; python_version>="3.6"
-connexion >= 2.3.0; python_version=="3.5"
-connexion >= 2.3.0; python_version=="3.4"
+# connexion 2.6 dropped support for python < 3.6
+connexion >= 2.3.0; python_version=="3.5" or python_version=="3.4"
```
spacether, line 3: Please change this to `<= 2.3.0` I realize that I pinned this incorrectly when I fixed it.
The releases here: https://github.com/zalando/connexion/releases
Show that 2.3.0 is the last version that supports python 3.4 and 3.5

spacether: Thank you for fixing this

modules/openapi-generator/src/main/resources/python-aiohttp/requirements.mustache
spacether: Thank you for fixing this

spacether:
Just waiting on CI tests before merging, unrelated errors won't hold this up

cognifloyd:
:heavy_check_mark: Travis
:large_orange_diamond: Bitrise timed out

PR #5427, opened 2020-02-25
[typescript] Clean up modelPropertyNaming across generators 

### LINKED ISSUE #2976: [BUG][Typescript] Incorrect types generated due to case/naming differences
I have a schema, which types such as:
image

This generates a typescript interface like:
```typescript
/**
 * 
 * @export
 * @interface VDN
 */
export interface VDN {
    /**
     * 
     * @type {string}
     * @memberof VDN
     */
    vdnExtension?: string;
    /**
     * 
     * @type {string}
     * @memberof VDN
     */
    vdnName?: string;
    /**
     * 
     * @type {Array<string>}
     * @memberof VDN
     */
    skills?: Array<string>;
    /**
     * 
     * @type {number}
     * @memberof VDN
     */
    acdEngineId?: number;
    /**
     * 
     * @type {number}
     * @memberof VDN
     */
    lobId?: number;
    /**
     * 
     * @type {string}
     * @memberof VDN
     */
    lobName?: string;
    /**
     * 
     * @type {string}
     * @memberof VDN
     */
    programId?: string;
}
```

The structure is technically correct, even the property names have been converted from snake-case to camelCase. However this is just for the type, when I actually use the generated client, the property names are same as the original, so really the converted property names in the response are all undefined. Am I doing something wrong?

Generator: `typescript-axios`
OpenAPI-Generator version: `0.0.11-4.0.0-beta3`

amakhrov:
Fixes #2976

**Summary**
- Generators without runtime models conversion use "original" property naming by default. It's still possible to change it via cli options - might be helpful when used together with customized templates. The cli option description has been modified to provide some more context.
- Generators with runtime conversion (typescript-fetch, typescript-node, typescript-reduxquery) keep using "camelCase"

**Implementation notes**
- I decoupled `toVarName` from `toParamName`. Former respects the `modelPropertyNaming`, while latter keeps always using camelCase.
- Refactoring: use an enum instead of string for `modelPropertyNaming`. It now better resembles `enumPropertyNaming`.
- No samples have changed. It's because `petstore.yaml` has all model vars defined in camelCase - which means `original` naming (new) is effectively the same as `camelCase` (old).

### PR checklist

@TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @nicokoenig (2018/09) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -316,18 +316,6 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
         return name;
     }
 
-    /**
-     * Checks whether property names have breaking characters like ':', '-'.
-     * @param str string to check for breaking characters
-     * @return <code>true</code> if breaking characters are present and <code>false</code> if not
-     */
-    private boolean propertyHasBreakingCharacters(String str) {
-        final String regex = "^.*[+*:;,.()-]+.*$";
-        final Pattern pattern = Pattern.compile(regex);
-        final Matcher matcher = pattern.matcher(str);
-        return matcher.matches();
-    }
-
     @Override
     public String toModelName(final String name) {
         String fullModelName = name;
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -25,6 +25,7 @@ import io.swagger.v3.oas.models.parameters.Parameter;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.openapitools.codegen.CodegenConstants.ENUM_PROPERTY_NAMING_TYPE;
+import org.openapitools.codegen.CodegenConstants.MODEL_PROPERTY_NAMING_TYPE;
 import org.openapitools.codegen.*;
 import org.openapitools.codegen.meta.features.*;
 import org.openapitools.codegen.utils.ModelUtils;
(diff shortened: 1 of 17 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -274,39 +283,36 @@ public String modelFileFolder() {
...
     public String toParamName(String name) {
-        // sanitize name
         name = sanitizeName(name, "[^\\w$]");
 
         if ("_".equals(name)) {
             name = "_u";
         }
 
-        // if it's all uppper case, do nothing
-        if (name.matches("^[A-Z_]*$")) {
```
amakhrov: I carelessly removed this special case.
Please let me know if there is a particular reason to handle upper case differently from everything else!

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -274,39 +283,36 @@ public String modelFileFolder() {
...
     }
 
     @Override
     public String toVarName(String name) {
-        name = this.toParamName(name);
-        
-        // if the property name has any breaking characters such as :, ;, . etc.
-        // then wrap the name within single quotes.
-        // my:interface:property: string; => 'my:interface:property': string;
-        if (propertyHasBreakingCharacters(name)) {
```
amakhrov: this would never happen, because at this point `name` has been sanitized with even more restrictive rules

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptRxjsClientCodegen.java
```diff
@@ -89,8 +89,6 @@ public void setNpmRepository(String npmRepository) {
     @Override
     public void processOpts() {
         super.processOpts();
-        additionalProperties.put("isOriginalModelPropertyNaming", getModelPropertyNaming().equals("original"));
-        additionalProperties.put("modelPropertyNaming", getModelPropertyNaming());
```
amakhrov: This was not used either in the class or in the mustache templates. Most likely - copy/paste from another generator with runtime conversion support (like typescript-fetch)

wing328:

PR #5431, opened 2020-02-25
Base path name

topce:
add new option in cli to set different const name for BASE_PATH for angular-typescript client
default value is BASE_PATH to pass another base path name use option 
-p basePathName=MY_BASE_PATH
`export const MY_BASE_PATH = new InjectionToken<string>('myBasePath');`
instead of 
`export const BASE_PATH = new InjectionToken<string>('basePath');`

### PR checklist

@TiFu  @taxpon  @sebastianhaas  @kenisteward  @Vrolijkx  @macjohnny  @nicokoenig  @topce @akehir  @petejohansonxo @amakhrov


jimschubert:
Thanks for the PR. I'd like to understand why you would need to modify template variables in this way rather than customizing templates? I ask because this approach seems like a slippery slope to start down.

topce:
Hi @jimschubert and @amakhrov 

My use case: one Angular application 2 REST API that have different base paths
For both of them I need to generate code in order to call REST API from 
Angular App
 I solve problem with adding additional parameter basePathName.
Then I can generate different injection Tokens, 
so I do not need to modify  generated code in order to compile.

Obliviously It also works if Angular application need to call more then 2 REST APIs

**Do you have some better solution?
Sorry do not understand what you mean "slippery slope"  comment !?**

topce:
Hi @jimschubert and @amakhrov 

My use case: one Angular application 2 REST API that have different base paths
For both of them I need to generate code in order to call REST API from 
Angular App
 I solve problem with adding additional parameter basePathName.
Then I can generate different injection Tokens, 
so I do not need to modify  generated code in order to compile.

Obliviously It also works if Angular application need to call more then 2 REST APIs

**Do you have some better solution?
Sorry do not understand what you mean "slippery slope"  comment !?**

amakhrov:
@topce 
- **On "slipper slope"** :) : The fixed name of a single constant in the generated file is quite a minor detail. Adding separate options for such details will inevitable bloat the generator codebase and eventually make it harder to make future changes.
- **Do you have some better solution?** You got it! A suggested solution is to customize mustache templates for your use case and provide them to the generated via the --template-dir cli argument. You basically copy the corresponding .mustache file(s) from this repo to you local drive, make required tweaks and pass the dir name with the tweaked files via --template-dir. This generally allows much more flexible customizations than via config options.

topce:
Hi 
@jimschubert  and @amakhrov 
Thank you for fast response.
From your comments it look to me that you do not have many experience with Angular,
is this correct  ?
**What is "slippery slope" for you for me is elegant solution!**

Far superior than your suggested solution:
It is easier for end user to pass parameter than to provide for each 
different REST API templates that you even did not provided !

This constant is very important and typically is imported into app-module.ts from angular.
It also add good documentation option add user option to give good name to variable
(as you both know naming  of variable is important )
so we can call it instead of BASE_PATH for example PET_STORE_BASE_PATH
TWITTER_BASE_PATH etc

Regarding future 
I already PR support for angular 9 (probably angular 10 will also work ),
so future is bright ;-).
To me change does not look so complicated, just 4 files changed.


To summarize :
Use case: one angular app use multiple different REST APIs with 
my solution user just provide basePathName parameter,
if she does not provide it, it works like before no breaking compatibility!
She can use REST API: import this very important constant into angular app.

Your solution is IMHO worst because user need to modify templates 
at least 2 files (still we do not know how templates look like you did not provide it)
instead of just passing one additional parameter!

Originally request code review from  @wing328 and @macjohnny ,
but you two jump in!? 
So I asked them again what they think about this PR

For me it elegantly solve problem calling different REST API from 
same Angular application,
And you did not provide any solution , 
If you try to do it (create new templates for each REST API )
maybe you will more appreciate this PR ;-)

amakhrov:
@topce isn't BASE_PATH just one piece of the api configuration? The approach with renaming it will only work for it, while the rest of configuration (injected as `Configuration`) remains a problem. Frankly, i don't even know why basePath is injected as a separate token in the first place - Configuration contains basePath also.
So probably a right problem to solve is how to inject different configurations.

Alternative way for that is not to use provided in root (already configurable in generator). This way you just import your module with whatever configuration you want (using the forRoot pattern).

Any thoughts?

jimschubert:
@topce please realize that we're all trying to address the problem you're solving. As a core member of this project, it's important for me to review changes to make sure they are maintainable and that they're not confusing to our users. Consider, for example, if a user might think that `basePathName` might be misunderstood for the _actual_ basePath value. Consider, also, that we don't currently allow for dynamically modifying template variable names (at least nowhere that I'm aware of).

amakhrov:
@topce I just realized I still don't clearly understand the problem you're facing.
So you have 2 apis, and you generate two clients. Probably in two different directories (e.g. api1/, api2/). As a result you already have two unique injection tokens (in api1/variables.ts and api2/variables.ts). So again, what's is the issue with importing injection tokens from them and providing in your app?

topce:
@amakhrov 

> isn't BASE_PATH just one piece of the api configuration? 
In theory yes in practice no.
 You take a look to blog post 
https://medium.com/@balramchavan/generate-angular-ionic-client-code-from-openapi-swagger-rest-api-specifications-128a6899681a
You will see that he does not use Configuration at all 
Magic happened here
` provide: BASE_PATH, useValue: environment.basePath`

topce:
> @topce I just realized I still don't clearly understand the problem you're facing.
> So you have 2 apis, and you generate two clients. Probably in two different directories (e.g. api1/, api2/). As a result you already have two unique injection tokens (in api1/variables.ts and api2/variables.ts). So again, what's is the issue with importing injection tokens from them and providing in your app?

I think it will be some conflict there ;of course you can try on your side

topce:
> @topce please realize that we're all trying to address the problem you're solving. As a core member of this project, it's important for me to review changes to make sure they are maintainable and that they're not confusing to our users. Consider, for example, if a user might think that `basePathName` might be misunderstood for the _actual_ basePath value. Consider, also, that we don't currently allow for dynamically modifying template variable names (at least nowhere that I'm aware of).
thank you both for fast response!

no problem @jimschubert  and @amakhrov  maybe I overreacted as well

regarding basePathName  it could be renamed or refactor my code to use
basePathPrefix not to rename constant but to prefix it

Also for renaming variables I did not know for this rule but technically speaking
it is not variable it is **constant**  
`export const MY_BASE_PATH = new InjectionToken<string>('myBasePath');`

maybe just InjectionToken could be changed and keep BASE_PATH constant
I think @amakhrov  probably now understand problem better 
so  maybe he will find better solution
previous solution to use different templates was not good.
Maybe also to validate additional parameter.

I offered solution that is working for me
and it is easy to use for end users,
If you two find something better in sense of maintenance 
or want to refactor
my PR no problem just do it ;-)

macjohnny:
does https://github.com/OpenAPITools/openapi-generator/issues/4101#issuecomment-540455323 help in this discussion?
the `BASE_PATH` symbol can be renamed when importing
```
import { BASE_PATH as BASE_PATH_API_A } from './apis/api-a';
import { BASE_PATH as BASE_PATH_API_B } from './apis/api-b';
```
The argument value of the `InjectionToken` constructor is just a description and not relevant.

topce:
thank you @macjohnny 
indeed it look very relevant !

macjohnny:
@topce this is not related to api module prefix

topce:
As @macjohnny suggested not need to change constant name BASE_PATH and injection token
it is enough to to use type alias

Issue #5395, opened 2020-02-21
[BUG] generated code crashes with additionalParameters label:"Client: C"


michelealbano:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I used the following code, and the compiled program crashed (I didn't expect "Segmentation fault (core dumped)")

`#include "../api/DefaultAPI.h"

int main(int argc, char** argv) {

list_t* orchestration_flags = list_create();
list_addElement(orchestration_flags, keyValuePair_create("onlyPreferred", false));
list_addElement(orchestration_flags, keyValuePair_create("overrideStore", (void*)true));

orchestration_form_request_dto_t* request_form = orchestration_form_request_dto_create(orchestration_flags);

printf("request:\n%s\n",cJSON_Print(orchestration_form_request_dto_convertToJSON(request_form)));

return 0;
}
`

##### openapi-generator version

Current Master from today 21/02/2020

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.2
info:
  title: Additional parameters
  contact:
    name: Michele Albano, Arrowhead Consortia
    email: mialb@cs.aau.dk
  version: 4.1.3
paths:
  /orchestrator/orchestration:
    post:
      summary: Start Orchestration process.
      operationId: orchestrationProcessUsingPOST
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/OrchestrationFormRequestDTO"
        description: request
        required: true
      responses:
        "200":
          description: Returns possible providers of the specified service.
      deprecated: false
servers:
  - url: 'http://192.168.115.4:8441/'
    description: Test server
components:
  schemas:
    OrchestrationFormRequestDTO:
      type: object
      properties:
        orchestrationFlags:
          type: object
          additionalProperties:
            type: boolean
      title: OrchestrationFormRequestDTO
```

##### Command line used for generation

[...]

PR #5440, opened 2020-02-25
Support for additionalProperties in the C generator "Client: C" solves #5395

michelealbano:
I worked on the C-libcurl client. It was crashing when working on models with additionalProperties (see issue #5395). I found 2 problems:
- the generator was creating too many nodes one into the other
- when the additionalProperties were using native datatypes (booleans, etc), it was not allocating memory for them - but was trying to "free" them

I verified the json I produce now against json produced by the python client when I use integers as additionalProperties. They both produce this JSON now:
{
	"orchestrationFlags":	{
		"onlyPreferred":	0,
		"overrideStore":	1
	}
}

There is still a problem with booleans, but I will solve it later.

@zhemant
@ityuhui

### PR checklist


modules/openapi-generator/src/main/resources/C-libcurl/apiKey.c.mustache
```diff
@@ -9,6 +9,12 @@ keyValuePair_t *keyValuePair_create(char *key, void *value) {
     return keyValuePair;
 }
 
+keyValuePair_t* keyValuePair_create_allocate(char* key, double value) {
```
ityuhui: Hi @michelealbano, thank you for your great work. I have some questions here.

1. Is this function "keyValuePair_create_allocate" used now?

2. If the data type of value is not "double", this function cannot be used.

3. Form my view, user should manage(allocate and free) the memory.  If we (OpenAPI-Generator) provide the allocating function, we should also provide the free function.

modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
```diff
@@ -643,22 +642,24 @@ fail:
     keyValuePair_t *localMapKeyPair;
     cJSON_ArrayForEach({{{name}}}_local_map, {{{name}}})
     {
+		cJSON *localMapObject = {{{name}}}_local_map;
```
ityuhui: Hi

Why "{{{name}}}_local_map" is changed to "cJSON *localMapObject" ?

modules/openapi-generator/src/main/resources/C-libcurl/apiKey.c.mustache
michelealbano: It is to be used in the main.
Currently, the generated code for models says:
`if(cJSON_AddNumberToObject(localMapObject, localKeyValue->key, *(double *)localKeyValue->value) == NULL)`
thus it considers the value as a pointer. The keyValuePair_create limited itself to copying the value. Imagine I wanted to have a ("numberOfTeeth", 5), I had to do this in my main:
```
int number = 5; // since I need a lvalue
double* val1 = malloc(sizeof(number));
memcpy(val1, &number, sizeof(number));
keyValuePair_create("numberOfTeeth", val1);
```

I tried to put directly the number (5) in the value, casting int to void *, but I would have to change a great deal of generated code for models, for example the ones I showed at the beginning of this message.

We are actually providing free functions! For example, I have this in my model.c:
```
void orchestration_form_request_dto_free(orchestration_form_request_dto_t *orchestration_form_request_dto) {
    listEntry_t *listEntry;
    list_ForEach(listEntry, orchestration_form_request_dto->orchestration_flags) {
        keyValuePair_t *localKeyValue = (keyValuePair_t*) listEntry->data;
        free (localKeyValue->key);
        free (localKeyValue->value);
    }
    list_free(orchestration_form_request_dto->orchestration_flags);
    free(orchestration_form_request_dto);
}
```

modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
michelealbano: Later on, the code did:
`cJSON_AddItemToObject({{{name}}},"", localMapObject);`
meaning that there was an in item containing an item etc. The JSON for 
`{"fruit" : "apple"}`
 would become 
`{"" : {"fruit" : "apple"}}`
 (that's why I verified my JSON against the one produced by other generators - see the beginning of my PR message)

I needed to use `{{{name}}}_local_map` in the code, for example a few lines later where I do:
`localMapKeyPair = keyValuePair_create(strdup(localMapObject->string),strdup(localMapObject->valuestring));`
but since I was into an `{{#items}}`, the content of `{{{name}}}` was different.
Thus, I provided an alias for `{{{name}}}_local_map`.

modules/openapi-generator/src/main/resources/C-libcurl/apiKey.c.mustache
ityuhui: OK. I got it. No problems.

modules/openapi-generator/src/main/resources/C-libcurl/model-body.mustache
ityuhui: OK. I got it.

michelealbano:
What can I do to facilitate the merging of this PR?

michelealbano:
Dear maintainers
@zhemant
@ityuhui
Do you think that this PR is good enough to be merged?

ityuhui:
Hi @michelealbano

The PR looks good to me.

PR #5450, opened 2020-02-26
[kotlin][client] make base path globally configurable

4brunu:
My app needs to change the server that it's connected to, dev, staging, prod.

Currently the only way to do this is by passing the url to the API classes each time I instantiate them.

In this PR, I introduce a new class Configuration, that allows to easily change the global base path.
### PR checklist


wing328:
Looks like refactoring to me. Shall we include more properties/attributes in the configuration object similar to what we've done in other client generators (e.g. C#) and target this change against 5.0.x branch?

Ref: 
- https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/csharp/OpenAPIClient/src/Org.OpenAPITools/Client/Configuration.cs#L166
- https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/ruby/lib/petstore/configuration.rb#L129

4brunu:
Why targeting the 5.0.x?
This is not a breaking change, it will not affect anyone using the current version of kotlin client.
We can add some extra feature to it in the future, but now I would prefer merge this as is, and add more configuration as it's needed.
What do you think?

jimschubert:
This may not be a breaking change per se, but it is defining a new contract for consumers of the generated Kotlin client which we'd have to work around if we need to change it in the future.

Do you need to change this server at runtime or at compile time? If you're changing it at runtime, why can't you provide the new URL directly rather than setting the intermediate `Configuration.basePath`?

Rather than define this configuration helper object, would it be better to support 12-factor configuration via env or system property (if you're not switching at runtime)?

A major concern I have with this approach is that `Configuration.basePath` is not thread-safe. I agree it would probably be better to target this for 5.0 after more discussion. As an example, when I refactored the C# client to include a new concept of a Configuration object, that client was available for multiple minor versions as a beta/experimental refactoring. This gave the community plenty of time to provide feedback on the configuration structure before "mainlining" it.

4brunu:
> Do you need to change this server at runtime or at compile time?

The URL is defined at compile time.
I set the server URL only when my app is starting up, then I don't change it anymore.

> If you're changing it at runtime, why can't you provide the new URL directly rather than setting the intermediate Configuration.basePath?

I consume two swagger endpoints and it's error prone to set the right URL, because I have need to pass the server url every time I instantiate an Api class.

> Rather than define this configuration helper object, would it be better to support 12-factor configuration via env or system property (if you're not switching at runtime)?

I think I'm already doing this.
Let me give some code snippets to show my problem.

First of all, let me give some context.
I consume two different swagger endpoints.
Let's call them server one and server two.
Both server one and server two, have dev, stage and production.

Here is an example of my problem.

```
object MyApplication {
    val basePathServerOne: String
        get() {
            return System.getProperties().getProperty("basePathServerOne")
        }

    val basePathServerTwo: String
        get() {
            return System.getProperties().getProperty("basePathServerTwo")
        }
}

class Repository {
    fun deletePet() {
        val response = PetApi(MyApplication.basePathServerOne).deletePet(1000)
        // handle response
    }

    fun getStoreInventory() {
        val response = StoreApi(MyApplication.basePathServerTwo).getInventory()
        // handle response
    }
}
```

With the current implementation, I need to the basePath every time I instantiate the Api class.
And I need to decide if it's connected to server one and server two.

With the approach in this PR, I would only need to set at start up time:

```
object MyApplication {
    val basePathServerOne: String
        get() {
            return System.getProperties().getProperty("basePathServerOne")
        }

    val basePathServerTwo: String
        get() {
            return System.getProperties().getProperty("basePathServerTwo")
        }

    fun main() {
        com.example.serverone.Configuration.basePath = basePathServerOne
        com.example.servertwo.Configuration.basePath = basePathServerTwo
    }
}

class Repository {
    fun deletePet() {
        val response = PetApi().deletePet(1000)
        // handle response
    }

    fun getStoreInventory() {
        val response = StoreApi().getInventory()
        // handle response
    }
}
```

This way I would not have to pass the url every time I instantiate a Api class, because it's centralised.

I'm not sure if it was confusing or not.
But I'm trying to solve two problems:
- Centralize the basePath, so I don't have to pass it every time. (dev, staging, production)
- Centralize the basePath because I'm consuming two different swagger endpoints and it's error prone to set the right url every time.

I do see your points about the configuration class.
Do you have any suggestion on how to centralize the basePath that could be merged into master?

Maybe instead of putting the basePath into Configuration, setting it into the ApiClient.kt?

I'm asking this, because the 5.0.0 release is schedule only to May, that the reason why I'm trying to avoid that, to be able to use this before that date.

Thanks

jimschubert:
Thanks for sharing the use case. I wrote up a nice response, then accidentally navigated away and lost it. I hope I can convey my message without much broken thought.

By introducing `Configuration.basePath`, we're defining a contract for consuming clients who may or may not know about this. Whereas current code could easily be modified to change the behavior of the constructor, we'll now have another external file and additional things to consider.

Consider for instance if I'm a user who generates a Kotlin SDK from openapi-generator. I may not look closely at the generated code, expecting templates to "do the right thing" with my custom templates. I've hard-coded my production URL in my templates since my code is generated from a spec with an invalid URL (perhaps that URL is the hostname on my corporate machine). Externalizing this to a Configuration object now potentially exposes internal network structures to SDK users.

In addition to this, setting the single basePath in an object may make this harder to consume from Java, and may make templated servers more difficult to implement in the future without a breaking change. I haven't attempted interop on the Kotlin client, but I would expect objects with modifyable properties to be marked with `@JvmStatic`.

I'm wondering if this might make more sense and support your use case as well as others?

```
class PetApi(basePath: kotlin.String = defaultBasePath) : ApiClient(basePath) {
    companion object {
        @JvmStatic
        val defaultBasePath: String by lazy {
            System.getProperties().getProperty("server.baseUrl", "http://petstore.swagger.io/v2")
        }
    }
…
```

We could make `server.baseUrl` configurable, and even build it off the API name so like `server.petStore.baseUrl`. This way, you could generate with the appropriate configuration, it'll be set-once and thread-safe, and existing code would work exactly as it does now.

4brunu:
Hey, thanks for putting time into this.

Looks good to me.

But, could we instead of using `"server.baseUrl"` to get the server path, use something that has the identifier of the project in it?  

Something like `"projectname.server.baseUrl"`.

This way, I could set different base path for the two swagger endpoints that I'm consuming.

jimschubert:
I've tagged 4.3.0. I'd like to get this into that release but looking for one more set of eyes on the approach.

PR #5456, opened 2020-02-26
[Core] Only debug log what files are created

richardwhiuk:
Currently the generator is quite spammy about what files are being created, which can be a large number, which makes it difficult to see warnings emitted.

It feels like this isn't really worth an info level log - and could be a debug level log.

Thoughts?

### PR checklist


jimschubert:
As a user running a tool which is intended to output files, I would want the default output to show that file. An example of similar behavior would be tar or zip commands.

Some tools (like my IntelliJ plugin) really rely on this output to let users know what's being done. Same would go for our CI, where dropping the INFO logging might hide problems in generation.

Would it make sense to have a `--quite` type of option which would reduce the console noise?

I did add a Logger decorator to reduce some noise, and I think lots of other stuff could move to DEBUG, but I'd prefer file writes to stay INFO.

PR #5459, opened 2020-02-26
[kotlin][client] Add Jackson to interface properties and remove extra line feed

mtraynham:
Follow on to https://github.com/OpenAPITools/openapi-generator/pull/5236, this add Jackson annotations to interface variables as well. 

I also added the JsonFormat annotation to all of them for consistency and removed the extra line feeds from the data class var files to fix the output.

### PR checklist

@jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10) @4brunu (2019/11)

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
@@ -10,8 +10,7 @@
     {{/gson}}
     {{#jackson}}
     {{#isDateTime}}
-    @JsonFormat
-    (shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
     {{/isDateTime}}
     @JsonProperty("{{{vendorExtensions.x-base-name-literal}}}")
     {{/jackson}}
--- a/modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
@@ -9,9 +9,6 @@
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
     {{#jackson}}
-    {{#isDateTime}}
-    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
-    {{/isDateTime}}
     @JsonProperty("{{{vendorExtensions.x-base-name-literal}}}")
     {{/jackson}}
     {{/multiplatform}}
--- a/modules/openapi-generator/src/main/resources/kotlin-client/data_class_req_var.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/data_class_req_var.mustache
@@ -9,9 +9,6 @@
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
     {{#jackson}}
-    {{#isDateTime}}
-    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
-    {{/isDateTime}}
     @JsonProperty("{{{vendorExtensions.x-base-name-literal}}}")
     {{/jackson}}
     {{/multiplatform}}
--- a/modules/openapi-generator/src/main/resources/kotlin-client/data_class_req_var.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/data_class_req_var.mustache
@@ -9,6 +9,9 @@
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
     {{#jackson}}
+    {{#isDateTime}}
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
+    {{/isDateTime}}
     @JsonProperty("{{{vendorExtensions.x-base-name-literal}}}")
     {{/jackson}}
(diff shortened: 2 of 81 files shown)


mtraynham:
Also wanted to point out, it seems a bit heavy handed to use the date formatter.

```java
@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
```

It's probably better to have a custom serializer, as I don't think you can actually override `@JsonFormat` annotations.

modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
```diff
@@ -10,10 +10,9 @@
     {{/gson}}
     {{#jackson}}
     {{#isDateTime}}
-    @JsonFormat
-    (shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
```
4brunu: I think here we should use the iso8601 date format

modules/openapi-generator/src/main/resources/kotlin-client/data_class_req_var.mustache
```diff
@@ -9,7 +9,10 @@
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
     {{#jackson}}
+    {{#isDateTime}}
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
```
4brunu: The same here

modules/openapi-generator/src/main/resources/kotlin-client/interface_opt_var.mustache
```diff
@@ -8,5 +8,11 @@
     {{#gson}}
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
+    {{#jackson}}
+    {{#isDateTime}}
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
```
4brunu: And here

modules/openapi-generator/src/main/resources/kotlin-client/interface_req_var.mustache
```diff
@@ -8,5 +8,11 @@
     {{#gson}}
     @SerializedName("{{{vendorExtensions.x-base-name-literal}}}")
     {{/gson}}
+    {{#jackson}}
+    {{#isDateTime}}
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
```
4brunu: And here

modules/openapi-generator/src/main/resources/kotlin-client/data_class_opt_var.mustache
mtraynham: Hey @4brunu , thanks for the feedback.  I'm wondering if it would be better to just remove these `@JsonFormat` annotations in favor of a Jackson serializer/deserializer, similar to the other serialization libraries, such as Moshi, https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-client/jvm-common/infrastructure/OffsetDateTimeAdapter.kt.mustache.

I think moving to the serializer/deserializer would likely be better as I mentioned, once you have these `@JsonFormat` annotations, I don't think you can change them, so you'd be stuck with  whatever format you specified here, which I agree should be the iso8601 date format.

Maybe @herveDarritchon could also provide some feedback here as he added these annotations with #5236.

4brunu: Looks like a good idea to me 👍

4brunu: Looks like a good idea to me 👍

mtraynham:
@4brunu I've corrected the Date format issue you mentioned and applied some other fixes here.  I documented all the changes in the commit, but here that commit message is for reference:

> - Don't use JsonFormat for Date objects, this should be controlled via
>   a custom serializer/deserializer or a turning on and off serialization
>   features of Jackson.  I've updated the jacksonObjectMapper config to
>   write the dates as strings, which I think was intended in the original
>   commit.
>   https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_DATES_AS_TIMESTAMPS
>   https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/cfg/MapperConfig.html#getDateFormat--
> - Dont' use @JsonFormat(shape = JsonFormat.Shape.OBJECT) for enums.
>   This causes Enums to be formatted as objects with an internal "value"
>   field.  In reality, OpenAPI enums are just strings without properties
>   and should be treated as a string.
>   https://www.baeldung.com/jackson-serialize-enums#2-enum-as-json-object
> - Add's Kotlin use site annotation @get: to JsonProperty for parent interface
>   properties.  Otherwise Kotlin will warn:
>   "This annotation is not applicable to target 'member property without
>   backing field or delegate'"
> - Add's JsonTypeInfo annotations to interfaces for inheritance.  This
>   was copied verbatim from the kotlin-spring generator.
>   https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-spring/typeInfoAnnotation.mustache

Hopefully @herveDarritchon might have some feedback as he performed the initial PR (#5459), but all this works well for my own code base which is a bit more advanced than the Pet store examples...

wing328:
@mtraynham thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5464, opened 2020-02-27
[REQ][JAVA]: Add option to generate arrays with "uniqueItems" true as Sets rather than Lists

### LINKED ISSUE #5254: [REQ][JAVA]: Add option to generate arrays with "uniqueItems" true as Sets rather than Lists
### Description
It would be great if there was an option to generate properties declared as "array" with "uniqueItems": true as a Set/HashSet rather than a List/ArrayList.

### Proposed solution
This could be enabled with a flag option to avoid it beying a breaking change.

Januson:
### Summary
 - Update Java code generation to use sets instead of lists when uniqueItems is set to true
 - Add import resolution for sets
 - Add tests

In case you do not want this to be the default behavior it could be made opt-in via additionalProperties.

resolve #5254

### Java Technical Committee

 - @bbdouglas 
 - @sreeshas 
 - @jfiala
 - @lukoyanov 
 - @cbornet
 - @jeff9finger
 - @karismann
 - @Zomzog
 - @lwlee2608
 - @bkabrda

### PR checklist


wing328:
Should we target the fix to the current master instead as it's more a bug fix? (users can switch back to "old' incorrect behavior by removing `uniqueItems` in the spec)

Januson:
@wing328 Sure I can create a PR for master branch as well.

Januson:
Added PR targeting master. #5466

PR #5465, opened 2020-02-27
[typescript-rxjs] add support for raw response and progressSubscriber

denyo:
This PR adds support for an optional parameter on controller functions in order to retrieve the status code or pass a `progressSubscriber` to receive events for longer lasting calls like file uploads.

It also makes use of the new `??` operator to support a relative `baseUrl` like `""`.

### PR checklist


denyo:
@amakhrov thank you for your feedback.

The `raw` response approach definitely makes a lot of sense. 
Regarding `RequestOpts` and `AjaxRequest`, you are right. `RequestOpts` is basically a subset except for one or two props. I changed it to extend `AjaxRequest` with more explicit typing.

About the typescript version it is also brings the optional chaining operator `?`.
I guess most people use this generator for React or NodeJS projects. At least for React it is used since November in CRA.

Please have a look at the updated changes.

macjohnny:
since this PR contains breaking-changes without fallback, we would have to wait for the 5.0.0 release (https://github.com/OpenAPITools/openapi-generator/milestone/20)

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -35,19 +35,13 @@ export class Configuration {
     }
 
     get apiKey(): ((name: string) => string) | undefined {
-        const apiKey = this.configuration.apiKey;
-        if (!apiKey) {
-            return undefined;
-        }
-        return typeof apiKey === 'string' ? () => apiKey : apiKey;
+        const { apiKey } = this.configuration;
+        return apiKey ? (typeof apiKey === 'string' ? () => apiKey : apiKey) : undefined;
```
ermik, line 39: I don't think double ternary is appropriate there. It's obfuscating the conditional, it makes it less readable. I've addressed the nully coalescing in #5329, and I think webpack can take care of the compressing the code.

modules/openapi-generator/src/main/resources/typescript-rxjs/runtime.mustache
```diff
@@ -35,19 +35,13 @@ export class Configuration {
...
     }
 
     get accessToken(): ((name: string, scopes?: string[]) => string) | undefined {
-        const accessToken = this.configuration.accessToken;
-        if (!accessToken) {
-            return undefined;
-        }
-        return typeof accessToken === 'string' ? () => accessToken : accessToken;
+        const { accessToken } = this.configuration;
+        return accessToken ? (typeof accessToken === 'string' ? () => accessToken : accessToken) : undefined;
```
ermik, line 44: Double ternary is problematic, as I described above.

denyo:
I am thinking about adding support for a flag that will control weather or not the optional `progressSubscriber` will be part of the generated code.
What would be the best approach for this? Introducing an `--additional-properties` flag?
Any hints where I can see this? Or is looking at use of the `withInterfaces` flag enough even though it seems to be available in some generators but without any functionality like here.

macjohnny:
@denyo can you please resolve the conflicts?

ermik:
Do you think these changes are possible without losing readability? I've highlighted the issue with the double ternary above.

denyo:
> Do you think these changes are possible without losing readability? I've highlighted the issue with the double ternary above.

I could improve the formatting but I am against having multiple exit points within a function since it makes it more difficult to debug is against the one entry point / one exit point rule.

ermik:
`one entry point / one exit point rule` I don't think this applies here, since tenrary has two exit points (`true` & `false`) and the Nth ternary has 2ⁿ exits. The fact that all those exits happen on the same line of code, doesn't make it any easier to debug them, in fact, I would argue, it makes it harder.

@macjohnny I don't know if you have any strong opinions on this. Mine is certainly just that — an opinion — this shouldn't be held up over any sort of debate on best practices.

macjohnny:
@ermik I agree with you, in my opinion, the double ternaries are less readable than two separate exit points, which, given that the resulting function is very small, are easy to spot.

moreover, I would prefer to avoid the `??` operator, as it breaks compatibility with older typescript versions without a real necessity to do so.

however, since this would be part of version 5.0.0, and users could still use 4.3.1 for older typescript versions, I think we should merge this one, as it adds features and improves the code, and @denyo has put a lot of effort in creating and improving the typescript-rxjs generator.

ermik:
🎉  great job @denyo.

PR #5466, opened 2020-02-27
[M][REQ][JAVA]: Add option to generate arrays with "uniqueItems" true as Sets rather than Lists

### LINKED ISSUE #5254: [REQ][JAVA]: Add option to generate arrays with "uniqueItems" true as Sets rather than Lists

Januson:
### Summary
 - Update Java code generation to use sets instead of lists when uniqueItems is set to true
 - Add import resolution for sets
 - Add tests

resolve #5254

### Java Technical Committee
 - @bbdouglas
 - @sreeshas
 - @jfiala
 - @lukoyanov
 - @cbornet
 - @jeff9finger
 - @karismann
 - @Zomzog
 - @lwlee2608
 - @bkabrda

### PR checklist


bkabrda:
I'm wondering whether `HashSet` is the right Set implementation to use here. IIUC the marking a list with `uniqueItems: true` still means that it's a list, so ordering matters. I think `LinkedHashSet` would be a better implementation to use here.

Januson:
@bkabrda Switched Set implementation to LinkedHashSet.

Januson:
@bkabrda 
Fixed a couple issues where there was a missing import for Set/LinkedHashSet when uniqueItems was used for param/response.

Second issue was that some of the templates were expecting only list types, so I had to update those to support sets.

Now the generation of sets should work for models, parameters and responses.

samples/client/petstore/csharp-netcore/OpenAPIClient/src/Org.OpenAPITools/Api/FakeApi.cs
```diff
@@ -257,7 +257,7 @@ public interface IFakeApiSync : IApiAccessor
         /// <param name="enumQueryString">Query parameter enum test (string) (optional, default to -efg)</param>
         /// <param name="enumQueryInteger">Query parameter enum test (double) (optional)</param>
         /// <param name="enumQueryDouble">Query parameter enum test (double) (optional)</param>
-        /// <param name="enumFormStringArray">Form parameter enum test (string array) (optional, default to $)</param>
```
wing328, line 260: @Januson overall it's good but looks like it breaks the default value for array of enum (string). Can you please have a look?

Januson: @wing328 Done.

barnabycourt:
@wing328 Is there anything else that you would like from this fix? Thanks!

wing328:
I've resolved the merge conflicts via https://github.com/OpenAPITools/openapi-generator/pull/5466/commits/08931f30b87ad75af6979744d3bee5b8dc5714c1

If all tests pass and no further feedback from anyone, I'll merge this PR.

PR #5469, opened 2020-02-27
Change any declaration to declaration from axios

nibytes:
I think this will be better than any declaration. Samples will generate soon.


nibytes:
> 
> 
> Thanks for your PR @nibytes !
> Do you think it could be beneficial to specify a proper type in other similar places:
> 
>     * baseApi (RequestArgs)
> 
>     * apiInner (options)
> 
>     * anything else?

Yes, I added this, about another else I don't know. All places where user can to interact with types I was change.

nibytes:
> 
> 
> Looks good to me

Can you rerun CI?

PR #5477, opened 2020-02-28
PR to solve 2 open issues on enums: #5091 and #4293

### LINKED ISSUE #4293: [BUG][C] support for enumerators
#### Bug Report Checklist

- [ X ] Have you provided a full/minimal spec to reproduce the issue?
- [ X ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ X ] What's the version of OpenAPI Generator used?
- [ X ] Have you search for related issues/PRs?
- [ X ] What's the actual output vs expected output?

##### Description

I tried to compile the code generated from the first declaration I report below, and the result was:

```
In file included from /mnt/hgfs/shared/clientC/api/DefaultAPI.h:8,
                 from /mnt/hgfs/shared/clientC/api/DefaultAPI.c:4:
/mnt/hgfs/shared/clientC/api/../model/service_registry_response_dto.h:15:34: error: redeclaration of enumerator ‘NOT_SECURE’
                 typedef enum  {  NOT_SECURE, CERTIFICATE, TOKEN } secure_e;
                                  ^~~~~~~~~~
```

In C the namespace of the enums is global, thus I tried the second OpenAPI declaration I report, where I don't repeat enums' names. The result was:
```
In file included from /mnt/hgfs/shared/clientC/api/DefaultAPI.h:8,
                 from /mnt/hgfs/shared/clientC/api/DefaultAPI.c:4:
/mnt/hgfs/shared/clientC/api/../model/service_registry_response_dto.h:15:67: error: conflicting types for ‘secure_e’
                 typedef enum  {  NOT_SECURE, CERTIFICATE, TOKEN } secure_e;
                                                                   ^~~~~~~~
```

commit 4543c21 (grafted, HEAD -> master, origin/master, origin/HEAD)
Author: sullis github@seansullivan.com
Date: Sat Oct 26 21:10:55 2019 -0700

Repeated enum names:
https://gist.github.com/michelealbano/77f9795d84206c57676cd93c22cf8308

Multiple enums:
https://gist.github.com/michelealbano/60d912e034d2ec2307226e7ccc50c409

java -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g c -i cproblem2.json -o clientC --additional-properties=swaggerDocketConfig=true

##### Steps to reproduce

java -jar openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g c -i cproblems.json -o clientC --additional-properties=swaggerDocketConfig=true
cd clientC
mkdir build
cd build
cmake ..
make

##### Related issues/PRs

N/A

##### Suggest a fix

I suggest to decorate the enum names. See for example:
https://stackoverflow.com/questions/2161940/two-enums-have-some-elements-in-common-why-does-this-produce-an-error
In the case of my code, this:
```
                typedef enum  {  NOT_SECURE, CERTIFICATE, TOKEN } secure_e;
```
should become this:
```
                typedef enum  {  NOT_SECURE, CERTIFICATE, TOKEN } service_registry_response_dto_secure_e;
```

and this:
```
                typedef enum  {  NOT_SECURE, CERTIFICATE, TOKEN } service_registry_response_dto_secure_e;
```

### LINKED ISSUE #5091: [BUG] C Libcurl no enums created
Minimal spec
https://gist.github.com/axbusch/d7c0a34ff31167e052f4eed7ad5c744d

Version: 4.2.3-SNAPSHOT

cmd:
OPENAPI_GENERATOR_VERSION=4.2.3-SNAPSHOT ./openapi-generator-cli generate -i ~/bin/weather_minimal.yaml -g c  -o ~/openweather-c-minimal

#### Problem
The parameters are defined as enums in the API specification, so i expect i get some enums in my c-sources

#### Output
A default API is created with 
```
model_200_t* DefaultAPI_weatherGet(apiClient_t *apiClient ,units_e units ,mode_e mode);
```

But the enums for units_e & mode_e are not defined

michelealbano:
To fix #5091 that requires enums to be generated when the enum is in a param to a API call, instead than in a model. I did that by copying and adapting enum code from model-body.mustache to api-body.mustache, and from model-header.mustache to model-body.mustache.

To fix #4293 that needs to decorate enums, for when enum names or enum values are repeated over the yaml definition.
I decorated the values of the enums in API's params as {{enumName}}_{{{value}}}. The enum name did not need decoration.
For models, I decorated the values as {{classFilename}}_{{enumName}}_{{{value}}}. The enum names were decorated as {{classFilename}}_{{classname}}_e.

As with pre-PR version, string enums are well supported. If you provide me interesting YAML with other kinds of enums, I can add material to the mustache templates to support them too.

Take a look at this YAML:
[https://gist.github.com/michelealbano/88262eddc22966553055acc20fb6021f](url)
and to this main.c:
[https://gist.github.com/michelealbano/5ebb78590354af4785b345842891ae16](url)

@zhemant
@ityuhui

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java
@@ -623,11 +623,7 @@ public class CLibcurlClientCodegen extends DefaultCodegen implements CodegenConf
     }
 
     public void setProjectName(String projectName) {
-        System.out.println("michele was here setProjectName this1 "+this.projectName);
-        System.out.println("michele was here setProjectName no1 "+projectName);
         this.projectName = underscore(sanitizeName(projectName.toLowerCase(Locale.ROOT)));
-        System.out.println("michele was here setProjectName this2 "+this.projectName);
-        System.out.println("michele was here setProjectName no2 "+projectName);
     }
 
     @Override
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java
@@ -617,11 +617,17 @@ public class CLibcurlClientCodegen extends DefaultCodegen implements CodegenConf
         } else {
             setProjectName(defaultProjectName);
         }
+        System.out.println("michele was here preprocessOpenAPI "+projectName);
         additionalProperties.put(PROJECT_NAME, projectName);
+        System.out.println("michele was here preprocessOpenAPI after "+projectName);
     }
 
     public void setProjectName(String projectName) {
-        this.projectName = underscore(projectName.toLowerCase(Locale.ROOT));
+        System.out.println("michele was here setProjectName this1 "+this.projectName);
+        System.out.println("michele was here setProjectName no1 "+projectName);
+        this.projectName = underscore(sanitizeName(projectName.toLowerCase(Locale.ROOT)));
+        System.out.println("michele was here setProjectName this2 "+this.projectName);
+        System.out.println("michele was here setProjectName no2 "+projectName);
     }
 
     @Override
--- a/modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
@@ -13,6 +13,95 @@
 
 {{#operations}}
 {{#operation}}
+{{#allParams}}
+
+{{#isEnum}}
+// Functions for enum {{enumName}} for {{{classname}}}_{{{operationId}}}
+
+
+char* {{enumName}}_ToString({{baseName}}_e {{enumName}}){
+char *{{enumName}}Array[] =  { {{#allowableValues}}{{#enumVars}}"{{{value}}}"{{^-last}},{{/-last}}{{/enumVars}}{{/allowableValues}} };
+    return {{enumName}}Array[{{enumName}}];
+}
+
+{{baseName}}_e {{enumName}}_FromString(char* {{enumName}}){
+    int stringToReturn = 0;
+    char *{{enumName}}Array[] =  { {{#allowableValues}}{{#enumVars}}"{{{value}}}"{{^-last}},{{/-last}}{{/enumVars}}{{/allowableValues}} };
+    size_t sizeofArray = sizeof({{enumName}}Array) / sizeof({{enumName}}Array[0]);
+    while(stringToReturn < sizeofArray) {
+        if(strcmp({{enumName}}, {{enumName}}Array[stringToReturn]) == 0) {
+            return stringToReturn;
+        }
+        stringToReturn++;
+    }
+    return 0;
+}
+
+cJSON *{{enumName}}_convertToJSON({{baseName}}_e {{enumName}}) {
+cJSON *item = cJSON_CreateObject();
+{{#isString}}
+    if(cJSON_AddStringToObject(item, "{{{paramName}}}", {{{enumName}}}_ToString({{{enumName}}})) == NULL) {
+        goto fail;
+    }
+{{/isString}}
+{{#isNumeric}}
+    if(cJSON_AddNumberToObject(item, "{{{paramName}}}", {{{enumName}}}) == NULL) {
+        goto fail;
+    }
+{{/isNumeric}}
+    return item;
+    fail:
+    cJSON_Delete(item);
+    return NULL;
+}
+
+{{baseName}}_e {{enumName}}_parseFromJSON(cJSON *{{enumName}}JSON){
+
+{{baseName}}_e *{{enumName}} = NULL;
+
+{{#isEnum}}
+{{#isNumeric}}
+cJSON *{{{enumName}}}Var = cJSON_GetObjectItemCaseSensitive({{enumName}}JSON, "{{{paramName}}}");
+if(!cJSON_IsNumber({{{enumName}}}Var))
+{
+    goto end;
+}
+{{/isNumeric}}
+{{#isString}}
+{{{baseName}}}_e {{enumName}}Variable;
+cJSON *{{{enumName}}}Var = cJSON_GetObjectItemCaseSensitive({{enumName}}JSON, "{{{paramName}}}");
+if(!cJSON_IsString({{{enumName}}}Var) || ({{{enumName}}}Var->valuestring == NULL)){
+    goto end;
+}
+{{enumName}}Variable = {{enumName}}_FromString({{{enumName}}}Var->valuestring);
+{{/isString}}
+{{/isEnum}}
+return {{enumName}}Variable;
+end:
+return 0;
+}
+{{/isEnum}}
+
+{{/allParams}}
+{{/operation}}
+{{/operations}}
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+{{#operations}}
+{{#operation}}
 {{#summary}}
 // {{{summary}}}
 //
@@ -143,7 +232,8 @@
         {{^isListContainer}}
         keyQuery_{{{paramName}}} = strdup("{{{baseName}}}");
         valueQuery_{{{paramName}}} = {{#isString}}{{^isEnum}}strdup({{/isEnum}}{{/isString}}({{{paramName}}}){{#isString}}{{^isEnum}}){{/isEnum}}{{/isString}};
-        keyPairQuery_{{paramName}} = keyValuePair_create(keyQuery_{{{paramName}}}, {{#isEnum}}(void *){{/isEnum}}{{^isString}}&{{/isString}}valueQuery_{{{paramName}}});
+        keyPairQuery_{{paramName}} = keyValuePair_create(keyQuery_{{{paramName}}}, {{#isEnum}}(void *)strdup({{enumName}}_ToString(
+		{{/isEnum}}{{^isString}}&{{/isString}}valueQuery_{{{paramName}}}{{#isEnum}})){{/isEnum}});
         list_addElement(localVarQueryParameters,keyPairQuery_{{paramName}});
         {{/isListContainer}}
     }
@@ -334,10 +424,12 @@
         free(keyQuery_{{{paramName}}});
         keyQuery_{{{paramName}}} = NULL;
     }
+{{^isEnum}}
     if(valueQuery_{{{paramName}}}){
         free(valueQuery_{{{paramName}}});
         valueQuery_{{{paramName}}} = NULL;
     }
+{{/isEnum}}
     if(keyPairQuery_{{{paramName}}}){
         keyValuePair_free(keyPairQuery_{{{paramName}}});
         keyPairQuery_{{{paramName}}} = NULL;
--- a/modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
@@ -57,9 +57,7 @@ cJSON *item = cJSON_CreateObject();
 
 {{baseName}}_e {{enumName}}_parseFromJSON(cJSON *{{enumName}}JSON){
 
-{{baseName}}_e *{{enumName}} = NULL;
-
-{{#isEnum}}
+{{{baseName}}}_e {{enumName}}Variable = 0;
 {{#isNumeric}}
 cJSON *{{{enumName}}}Var = cJSON_GetObjectItemCaseSensitive({{enumName}}JSON, "{{{paramName}}}");
 if(!cJSON_IsNumber({{{enumName}}}Var))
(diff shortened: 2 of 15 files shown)


michelealbano:
@axbusch please verify that this works for your API

zhemant:
@michelealbano I would suggest making first value as NULL, this will take care of all the checks to be synchronized as enums will have value 0. And this will also take care when the enums are nullable.

michelealbano:
Hi @zhemant what do you mean with "making first value"?

zhemant:
Also If I am not wrong, even the models which are present in the paths are not supported. I have to check this did not try the lastest master branch yet.

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
```diff
@@ -11,6 +11,95 @@
...
+{{#operations}}
+{{#operation}}
+{{#allParams}}
+
+{{#isEnum}}
+// Functions for enum {{enumName}} for {{{classname}}}_{{{operationId}}}
+
+
+char* {{enumName}}_ToString({{baseName}}_e {{enumName}}){
+char *{{enumName}}Array[] =  { {{#allowableValues}}{{#enumVars}}"{{{value}}}"{{^-last}},{{/-last}}{{/enumVars}}{{/allowableValues}} };
```
zhemant: Here, the first value when the code is generated will be the actual enum value and its respective number will be 0, so if the enum is mandatory or optional in both cases the enum will be sent. But to avoid this we can make the array start with { NULL , and rest of values}. This will make NULL as 0, so when the enum is optional you can actually skip sending the enum value.

michelealbano: ok!

michelealbano: Same for model-body.mustache?

zhemant: yes, same for model-body and model-header. If you can push that also it would be great.

michelealbano: NULL is a token and cannot be used. I am worried about using NOTHING or UNDEFINED because they could be also values in the enum.
How do you suggest to call the first "NULL" element?

michelealbano: I could just decorate "NULL" and it could work.
On the other hand, the String corresponding to the NULL enum would be better be "", since a lot of user code does strcmp etc. Do you agree?

zhemant: These are my opinions and things which I have learned while using the code generator.

you could do something like this: 
```typedef enum  { {{projectName}}_{{classVarName}}_NULL = 0,{{#enumVars}} {{projectName}}_{{classVarName}}_{{{value}}} {{^-last}},{{/-last}}{{/enumVars}} } {{classname}}_e;```

In case of keywords, you have to escape them. So how it will be is like in ToString and FromString, the actual values will be present as these are sent to the destination, and these values are in string format so they will not create problems.  and in model-header, where the enum is defined, it will contain escaped/prefixed values then it will not conflict with C keywords or duplicate values.

Here is an model-body code snippet
```
char* {{classname}}_ToString({{classname}}_e {{classname}}){
char *{{classname}}Array[] =  { "NULL", {{#allowableValues}}{{#enumVars}}"{{{value}}}"{{^-last}},{{/-last}}{{/enumVars}}{{/allowableValues}} };
    return {{classname}}Array[{{classname}}];
}

{{classname}}_e {{classname}}_FromString(char* {{classname}}){
    int stringToReturn = 0;
    char *{{classname}}Array[] =  {"NULL",  {{#allowableValues}}{{#enumVars}}"{{{value}}}"{{^-last}},{{/-last}}{{/enumVars}}{{/allowableValues}} };
    size_t sizeofArray = sizeof({{classname}}Array) / sizeof({{classname}}Array[0]);
    while(stringToReturn < sizeofArray) {
        if(strcmp({{classname}}, {{classname}}Array[stringToReturn]) == 0) {
            return stringToReturn;
        }
        stringToReturn++;
    }
    return 0;
}
```

model-header snippet:
```
typedef enum  { {{projectName}}_{{classVarName}}_NULL = 0,{{#enumVars}} {{projectName}}_{{classVarName}}_{{{value}}} {{^-last}},{{/-last}}{{/enumVars}} } {{classname}}_e;
```

There is a reason I am using ```{{projectName}}_{{classVarName}}_{{{value}}}```, the codebase I have contains a lot of repeated enum values definitions and C keywords, to overcome all these issues I have prefixed everything thing(functions, enums, etc) with these prefixes.


Also for your second comment, string compare will not create problems as enums in the code works as int values and if there is strcmp used anywhere in the code except FromString function then it is not good.

File: samples/client/petstore/c/api/PetAPI.c [hunk 2]
```diff
@@ -10,6 +10,83 @@
...
+
+
+
+
+
+
+
+
+// Functions for enum STATUS for PetAPI_findPetsByStatus
+
```
ityuhui: Hello @michelealbano @zhemant 

Actually, for the test case Petstore, the function involve enum is:

```
list_t*
PetAPI_findPetsByStatus(apiClient_t *apiClient ,list_t * status);
```
The status is a list rather than an enum, so it does not need add the code in api-header and api-body.

yaml snippet of [petstore.yaml](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/petstore.yaml)
```
parameters:
        - name: status
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: array
          items:
            type: string
            enum:
              - available
              - pending
              - sold
```

Maybe we need another template code to control whether generating the enum related code in api header and body.

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
michelealbano: The second comment was about using NULL, "" or "NULL", and in fact with "NULL" everything is fine.

The approach
`typedef enum  { {{projectName}}_{{classVarName}}_NULL = 0,{{#enumVars}} {{projectName}}_{{classVarName}}_{{{value}}} {{^-last}},{{/-last}}{{/enumVars}} } {{classname}}_e;`
it does not work in the use case where the values are repeated among two enums used in the same model, e.g.:
```
  schemas:
    '200':
      title: Successful response
      type: object
      properties:
        weather:
          type: string
          description: (more info Weather condition codes)
        secure:
          type: string
          enum:
            - NOT_SECURE
            - CERTIFICATE
            - TOKEN
        secure2:
          type: string
          enum:
            - NOT_SECURE
            - CERTIFICATE
            - TOKEN
```
Thus, we need at least {{enumName}} in the decoration.
Then, if there is repetition between different models, and we need also {{classVarName}}.
If we put there also `{{projectName}}`, an example of enum name is now `openweathermap_api__200_SECURE_NOT_SECURE`. Isn't it a little heavy? Would it be better without {{projectName}}?

samples/client/petstore/c/api/PetAPI.c
zhemant: But this is only for enum, what if there is a struct. Even the structs are not generated from the path parameters. I remember I was told that other generators support creating models from path params, I think we should do same. So, in this case, a model will be generated for enum and it will be handled as a list of models in api-body . If we have two api and they have same struct defined in path params, then it should be generated as a model rather than putting model code in api-files

ityuhui: Hi @zhemant  

I agree with you that the enum in API path should be created in api-body for this case.  

But I have a little question:

Take the petstore as an example,  there is an enum "pet_status_e" created by model-body, it can be used for 
```
list_t*
PetAPI_findPetsByStatus(apiClient_t *apiClient ,list_t * status);
```

 User can use pet_status_e for the "status" of PetAPI_findPetsByStatus

There is no need to create 
```
typedef enum  {  STATUS_available, STATUS_pending, STATUS_sold } status_e;
```
in api-body.

It is duplicated but I admit it is not a big deal.

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
```diff
@@ -11,6 +11,91 @@
...
+return 0;
+}
+{{/isEnum}}
+
+{{/allParams}}
+{{/operation}}
+{{/operations}}
+
+
+
```
ityuhui: Hello @michelealbano

There are so many blank lines here. Could you please remove them ?

michelealbano: Done

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
michelealbano: It should be ok now. Anyway, I have not covered classes enums. I will work on that on a future PR, after somebody provides YAMLs to play against.

samples/client/petstore/c/api/PetAPI.c
michelealbano: Should I do something for this thread? Please try out the C generator as it is now, and tell me if you like the result.

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
michelealbano: @zhemant Is it ok the way I changed the code?

michelealbano:
Hi @zhemant @ityuhui @wing328 I think that this PR is ready for merging.
The massive decoration to the enums is ugly, but we have no more options with C, since it does not have objects and namespaces.

samples/client/petstore/c/api/PetAPI.c
ityuhui: Hello @michelealbano 

I don't understand why the generated functions (findPetsByStatus_STATUS_ToString, openapi_petstore_findPetsByStatus_status_e ) are "static" ?  (because I found they are not used in this source file, and users cannot use these functions in their source codes )

Actually, I agree with your commit first time, though it is duplicated for petstore sample, but it is acceptable. 

And why need add "{{#allowableValues}}" ?

michelealbano: Hi,
I made the functions "static" since I did not want to fill up the namespace too much. If they were not static, they would have to be heavily decorated. By the way, I see no use cases where we need *_convertToJSON and *_parseFromJSON for the enums, since the JSONification is done on the whole models. Should I remove enums' *_convertToJSON and *_parseFromJSON from the api-body, like I did with the model-body?

I added `{{#allowableValues}}` since the models' dump is:
```
      "allowableValues" : {
        "enumVars" : [ {
          "name" : "NOTSECURE",
          "isString" : false,
          "value" : "NOT_SECURE"
        }, {
          "name" : "CERTIFICATE",
          "isString" : false,
          "value" : "CERTIFICATE"
        }, {
          "name" : "TOKEN",
          "isString" : false,
          "value" : "TOKEN"
        } ],
        "values" : [ "NOT_SECURE", "CERTIFICATE", "TOKEN" ]
      }```
Thus, I had to go "inside" allowableValues to retrieve the possible values for the enums.

Comment (ityuhui) [reply samples/client/petstore/c/api/PetAPI.c]: Yes. I think you can if they are not needed for now.

> Should I remove enums' *_convertToJSON and *_parseFromJSON from the api-body, like I did with the model-body?

Comment (michelealbano) [reply samples/client/petstore/c/api/PetAPI.c]: Done

**Comment by michelealbano** [comment]
Hi @zhemant @ityuhui @wing328 I also added some code to sanitize the project title, to be able to compile for example the code from issue #2338 
I think that this PR is ready for merging.

File: modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java [hunk 4]
```diff
@@ -617,11 +617,17 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
...
+        System.out.println("michele was here preprocessOpenAPI after "+projectName);
     }
 
     public void setProjectName(String projectName) {
-        this.projectName = underscore(projectName.toLowerCase(Locale.ROOT));
+        System.out.println("michele was here setProjectName this1 "+this.projectName);
+        System.out.println("michele was here setProjectName no1 "+projectName);
+        this.projectName = underscore(sanitizeName(projectName.toLowerCase(Locale.ROOT)));
+        System.out.println("michele was here setProjectName this2 "+this.projectName);
+        System.out.println("michele was here setProjectName no2 "+projectName);
```
wing328: @michelealbano Shall we remove these `System.out.println`?

michelealbano: OUCH! I changed the .java to sanitize the input and I forgot to remove them!

michelealbano: Done

michelealbano:
The current error is on PHP:
```
./bin/openapi3/php-symfony-petstore.sh: line 22: cd: ./bin/openapi3/..: No such file or directory
3771
```

michelealbano:
Hi, I rebased and pushed, and it's not working anyway.
Let me know when an upstream/main passes all the tests, I will repeat the rebase/push.

axbusch:
thanks for your work 👍 

we switched from c generator to cpp so i did not test anymore, but hopefully will find time in the next few weeks to test the c generator again

we compile with clang and have sanitizers running and there were lots of leakage warnings in the generated soruces. i hope i can fix some of these

wing328:

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
zhemant: @michelealbano Yes it is okay. Sorry for late replies

wing328:
@michelealbano thanks for the PR, which has been included in the 4.3.0 release: https://twitter.com/oas_generator/status/1243455743937789952

PR #5484, opened 2020-02-29
[core] Sanitize/underscore/camelize cache expiry

### LINKED ISSUE #5473: Move cache string helper methods to Caffeine cache
This should be moved from a Map to caffeine cache to automatically expire records, otherwise OpenAPI Generator will be unusable in hosted environments (such as our online generator).

_Originally posted by @jimschubert in https://github.com/OpenAPITools/openapi-generator/pull/5152#issuecomment-592279263_

jimschubert:
The helper methods for sanitize/underscore/camelize were recently
updated to cache values in static maps. This would lead to a memory leak
in hosted environments as the maps had no cleanup/expiry.

This moves those cached entries to Caffeine caches with expiry based on
last access to allow the edge-case performance improvement gains on very
large documents, without the memory leak in hosted or embedded
environments.

fixes #5473

### PR checklist


jimschubert:
@sebastien-rosset could you test this against whatever large specification you were evaluating which drove your contribution in #5152? This PR avoids memory leak for hosted/embedded situations.

sebastien-rosset:
> @sebastien-rosset could you test this against whatever large specification you were evaluating which drove your contribution in #5152? This PR avoids memory leak for hosted/embedded situations.

sure, checking now....

sebastien-rosset:
There seems to be a 40% increase of the execution time.

```
  | Map | Caffeine
-- | -- | --
  | Real | User | Real | User
  | 15.41 | 38.229 | 25.897 | 60.931
  | 18.447 | 47.051 | 25.805 | 57.767
  | 22.026 | 55.234 | 26.661 | 60.774
  | 18.866 | 46.036 | 24.342 | 55.947
  | 17.024 | 41.245 | 29.303 | 65.984
Average | 18.3546 | 45.559 | 26.4016 | 60.2806
```

I will try increasing the caffeine max cache and retest.

sebastien-rosset:
Increasing the expiration does not have much impact, but maybe it would make sense to increase to 30 seconds by default?

sebastien-rosset:
I tried setting the max cache sizes to 5000, 2000 and 2000 (instead of 500, 200, 200). As a rough estimate, if the average size of a string in the cache is 50 bytes, the max cache size will be around 500MB.

Now the real time for the same experiment is around 21 seconds.

jimschubert:
I think the above benchmarks are acceptable without changes.  Caffeine will be thread safe and atomic, while the map is not thread-safe and had no locking or lock-free guarantees.

What lead to the recommendation to increase to 30 seconds? Is it an arbitrary ceiling based on your benchmark?

I'd rather not increase defaults here to allow for 3 caches to hold 500MB of data each. Do you think it would be acceptable to make these settings configurable via system properties as I did with `OnceLogger`?

sebastien-rosset:
That works for me as is. Maybe later if we have performance issues I'll do more analysis

jimschubert:
@wing328 @ackintosh I've requested your reviews to make sure I'm not missing anything. I consider this high priority for the upcoming 4.3.0 release.

PR #5485, opened 2020-02-29
[Kotlin] Remove kotlin-reflect dependency when not needed

adamkaplan:
Removes un-unsed kotlin-reflect dependency when the combination of network and serialization library do not require it.
- When not using `ok-http3` or `ok-http4`
- **and**
  - not using `moshi`
  - _or_ using `moshi` with `moshiCodeGen=true`

### PR checklist


wing328:
@adamkaplan thanks for the PR, which sounds like a backward compatible change to me. Shall we target this change to the current master instead?

wing328:
Please run `./bin/kotlin-client-all.sh` to update Petstore samples so that CIs can test and verify the change.

adamkaplan:
@wing328 Sounds good. I’ll resubmit to 4.3.0.

Also, I did run the test update script in my second commit 3eda0c4 but it generated VERSION files with 4.3.0 in them. Not clear why that was...

4brunu:
Why is this PR targeting 5.0.x?
I thinks this doesn't introduce a breaking change, so I think it can target master.
What do you think?

Another question is that this PR is committing a lot of temporary files under build directory.
Could you please remove them?

adamkaplan:
See #5502

PR #5486, opened 2020-02-29
[scala] [template] scala model property style

chameleon82:
Property model style supported on AbstractScala level, but was not implemented in all depended templates. 
This PR will fix it

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
```diff
@@ -262,11 +262,6 @@ public String toParamName(String name) {
         return formatIdentifier(name, false);
     }
 
-    @Override
-    public String toVarName(String name) {
-        return formatIdentifier(name, false);
```
wing328: Would it be an issue if formatIdentifier is no longer used in `toVarName`?

chameleon82: As of this change it is not backward compatible to akka-http, as `formatIdentifier` and `toVarName` has some difference when original name in UPPERCASE. 
`uPPERCASE` for `formatIdentifier` 
`UPPERCASE` for `toVarName` (do not apply naming rules for uppercase names)

I've added some tests to cover those cases.
I could revert formatIdentifier for default `camelCase` to make it backward compatible for `2.3.0` release, should I?

chameleon82: @wing328 i could add next code
```
    @Override
    @Deprecated//(since = "5.0.0", forRemoval = true)
    public String toVarName(String name) {
        if(CodegenConstants.ENUM_PROPERTY_NAMING_TYPE.camelCase.name().equals(this.getModelPropertyNaming())) {
            return formatIdentifier(name, false);
        } else {
           return super.toVarName(name);
        }
    }
```
with tests
```
        // Next cases will be UPPERCASE since 5.0.0
        Assert.assertEquals(scalaAkkaClientCodegen.toVarName("USERNAME"), "uSERNAME");
        Assert.assertEquals(scalaAkkaClientCodegen.toVarName("USER123NAME"), "uSER123NAME");
        Assert.assertEquals(scalaAkkaClientCodegen.toVarName("1AAAA"), "`1aAAA`");
        Assert.assertEquals(scalaAkkaClientCodegen.toVarName("1A"), "`1a`");
```
do we need this fallback?

wing328: As discussed let's consider this as a bug fix as it's fixing the inconsistent variable naming in different Scala generators.

chameleon82: https://github.com/OpenAPITools/openapi-generator/issues/5496

PR #5514, opened 2020-03-03
[BUG][scala][template] scala generate java.math.BigDecimal instead of scala type

chameleon82:
With option `--type-mappings=number=BigDecimal` generators produce models with imported `java.math.BigDecimal` type instead of `scala.math.BigDecimal` which is not required any import as has public alias `BigDecimal` 

This PR is to replace import from DefaultCodegen with proper scala related types import

### PR checklist

@clasnake (2017/07), @jimschubert (2017/09) ❤️, @shijinkui (2018/01), @ramzimaalej (2018/03), @chameleon82 (2020/03)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractScalaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractScalaCodegen.java
@@ -29,10 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static org.openapitools.codegen.utils.StringUtils.camelize;
 import static org.openapitools.codegen.utils.StringUtils.underscore;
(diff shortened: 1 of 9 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractScalaCodegen.java
```diff
@@ -105,11 +102,27 @@ public AbstractScalaCodegen() {
...
+        importMapping.put("UUID", "java.util.UUID");
+        importMapping.put("URI", "java.net.URI");
+        importMapping.put("File", "java.io.File");
+        importMapping.put("Timestamp", "java.sql.Timestamp");
+        importMapping.put("HashMap", "java.util.HashMap");
+        importMapping.put("Array", "java.util.List");
+        importMapping.put("ArrayList", "java.util.ArrayList");
+        // todo remove legacy date types
+        importMapping.put("Date", "java.util.Date");
+        importMapping.put("DateTime", "org.joda.time.*");
```
jimschubert: For this and others, we should use java.time rather than Joda. Non-standard libraries should be optional.

chameleon82: i've copied that from `DefaultCodegen` for backward compatibility.
Next PR https://github.com/OpenAPITools/openapi-generator/pull/5291 should solve java.time which should become default since 5.0.0, i think.

wing328:
Drone.io, Bitrise.io CI failure not related to this PR.

PR #4906, opened 2020-01-01
[CORE] Fixes composed schema discriminator map

spacether:
This PR improves how we define and handle discriminators.
We have added a feature flag legacyDiscriminatorBehavior
When it is set to false we:
- define the CodegenDiscriminator on all models that have a discriminator, including the ones that allOf inherit it from an ancestor
- fill in the discriminator MappedModels map with data from data from 6 sources (see them described below)
- validate the discriminator presence: validate that oneOf or anyOf models contain the required discriminator. If they do not contain it helpful errors are thrown.

I am doing this for the the below reasons:

Not all codegen models use extend to make composed models.
On generators that don't do that, they could lack the inherited discriminator.
Python experimental is an example where we don't use class inheritance to allOf inherit parent classes/models.
In that generator, discriminator is defined in each codegenModel, so we should not rely on java's composed schema inheritance solution to set a granparent's discriminator on a child model and a grandchild model that allOf inherit their ancestors.

We also need to do this if we want the values of the CodegenDiscriminator MappedModels to differ from grandparent to parent to child. When we include the below 1-6 sources of MappedModels, the MappedModels map will differ depending upon which model you are looking at (grandparent/parent/child)

My solution is to have the code set the discriminator on each model that contains a discriminator. 
Prior to this PR models that inherit a discriminator do not contain a discriminator. 
Then on that CodegenDiscriminator add the following to MappedModels:

1. any mappings from the current or inherited schema Discriminator Mapping
2. any `x-discriminator-value` mappings in:
   -  child models that allOf inherit from the the current schema
   - oneOf models in the current schema
   - anyOf models in the current schema
3. any child models that allOf inherit from the the current schema
4. any descendent models that allOf inherit from the above child models
5. oneOf models in the current schema
6. anyOf models in the current schema

### Why Feature Flag This?
- Some generators (javascript, probably others) implement discriminators differently and do not work if we set the flag to True. Cordoning this off behind the feature flag allows us to fix them later.
- This is a really big change
- I want to support legacy users who bring their old mustache files but use the latest generator code
- Even though I could get all tests to pass, that does not mean that all models are being tested
  - When the flag is true, Java models Cat and Dog now have discriminators with empty maps. I do not know if that is okay.

### Background Context:
When using oneOf or anyOf models with a discriminator in a parent composed schema, the discriminator map does not include the child oneOf models.
[...]

Issue #843, opened 2018-08-18
Abstract away from OpenAPI via transformation.


jimschubert:
##### Description

We have abstractions such as `CodegenModel`, `CodegenOperation`, `CodegenProperty`, etc. which represent abstractions away from OpenAPI for the sake of generation. This currently creates tight coupling between different versions of OpenAPI specifications and the underlying templating (examples: `CodegenOperation#vendorExtensions`, `CodegenOperation#externalDocs`)

##### Suggest a fix/enhancement

Create a representation of an API which is not tightly bound to OpenAPI specifications in general. Transform from OpenAPI specification to this data type via transform phase and inject this transform into the workflow (either `DefaultGenerator` or `Executor` (tbd #841).

This will unblock us to eventually allow configurable input specification parsers as well as configurable transforms (effectively opening us to code generation for all specification formats).

PR #5526, opened 2020-03-05
Adds parseFlattenSpec

spacether:
We automatically flatten inline models in specs in allOf and other composed schemas into new different Schemas. All of our testing was not including that flattening.
This PR adds the function parseFlattenSpec which includes the flattening and uses it in all tests except for the inline model resolver tests.

This PR was extracted from work in: https://github.com/OpenAPITools/openapi-generator/pull/4906/files and is blocking that PR

### PR checklist

Core Team Members
@wing328 (2015/07) 
@jimschubert (2016/05) 
@cbornet (2016/05)
@ackintosh (2018/02) 
@jmini (2018/04) 
@etherealjoy (2019/06)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -1134,8 +1134,8 @@ public class ModelUtils {
                     } else {
                         // not a parent since discriminator.propertyName is not set
                         hasAmbiguousParents = true;
-                        Boolean isNotExtractedInlineSchmea = !parentName.equals(schemaName+"_allOf");
-                        if (isNotExtractedInlineSchmea) {
+                        Boolean isNotExtractedInlineSchema = !parentName.equals(schemaName+"_allOf");
+                        if (isNotExtractedInlineSchema) {
                             refedWithoutDiscriminator.add(parentName);
                         }
                     }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -269,10 +269,10 @@ public class ModelUtils {
 
     /**
      * Invoke the specified visitor function for every schema that matches mimeType in the OpenAPI document.
-     * 
+     *
      * To avoid infinite recursion, referenced schemas are visited only once. When a referenced schema is visited,
      * it is added to visitedSchemas.
-     * 
+     *
      * @param openAPI the OpenAPI document that contains schema objects.
      * @param schema the root schema object to be visited.
      * @param mimeType the mime type. TODO: does not seem to be used in a meaningful way.
@@ -355,14 +355,14 @@ public class ModelUtils {
 
     /**
      * Return true if the specified schema is an object with a fixed number of properties.
-     * 
+     *
      * A ObjectSchema differs from an MapSchema in the following way:
      * - An ObjectSchema is not extensible, i.e. it has a fixed number of properties.
      * - A MapSchema is an object that can be extended with an arbitrary set of properties.
      *   The payload may include dynamic properties.
-     * 
+     *
      * For example, an OpenAPI schema is considered an ObjectSchema in the following scenarios:
-     * 
+     *
      *   type: object
      *   additionalProperties: false
      *   properties:
@@ -370,7 +370,7 @@ public class ModelUtils {
      *       type: string
      *     address:
      *       type: string
-     * 
+     *
      * @param schema the OAS schema
      * @return true if the specified schema is an Object schema.
      */
@@ -394,7 +394,7 @@ public class ModelUtils {
     /**
      * Return true if the specified schema is composed, i.e. if it uses
      * 'oneOf', 'anyOf' or 'allOf'.
-     * 
+     *
      * @param schema the OAS schema
      * @return true if the specified schema is a Composed schema.
      */
@@ -659,7 +659,7 @@ public class ModelUtils {
 
     /**
      * Check to see if the schema is a free form object.
-     * 
+     *
      * A free form object is an object (i.e. 'type: object' in a OAS document) that:
      * 1) Does not define properties, and
      * 2) Is not a composed schema (no anyOf, oneOf, allOf), and
@@ -737,7 +737,7 @@ public class ModelUtils {
      * Return a Map of the schemas defined under /components/schemas in the OAS document.
      * The returned Map only includes the direct children of /components/schemas in the OAS document; the Map
      * does not include inlined schemas.
-     * 
+     *
      * @param openAPI the OpenAPI document.
      * @return a map of schemas in the OAS document.
      */
@@ -767,7 +767,7 @@ public class ModelUtils {
         });
         return allSchemas;
     }
-    
+
     /**
      * If a RequestBody contains a reference to an other RequestBody with '$ref', returns the referenced RequestBody if it is found or the actual RequestBody in the other cases.
      *
@@ -906,10 +906,10 @@ public class ModelUtils {
 
     /**
      * Return the first Schema from a specified OAS 'content' section.
-     * 
+     *
      * For example, given the following OAS, this method returns the schema
      * for the 'application/json' content type because it is listed first in the OAS.
-     * 
+     *
      * responses:
      *   '200':
      *     content:
@@ -918,8 +918,8 @@ public class ModelUtils {
      *           $ref: '#/components/schemas/XYZ'
      *       application/xml:
      *          ...
-     *   
-     * @param content a 'content' section in the OAS specification. 
+     *
+     * @param content a 'content' section in the OAS specification.
      * @return the Schema.
      */
     private static Schema getSchemaFromContent(Content content) {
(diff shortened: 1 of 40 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1126,7 +1134,10 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
                     } else {
                         // not a parent since discriminator.propertyName is not set
                         hasAmbiguousParents = true;
-                        refedWithoutDiscriminator.add(parentName);
+                        Boolean isNotExtractedInlineSchmea = !parentName.equals(schemaName+"_allOf");
```
sebastien-rosset: Typo: Schmea.

jimschubert:
Why does this keep getting closed and reopened?

spacether:
> Why does this keep getting closed and reopened?

It was closed and reopened earlier because CI (including master branch) had failing tests. Closing and reopening a PR starts the CI tests over again.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: Thanks for the catch. This typo has been corrected.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1126,7 +1134,10 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
                     } else {
                         // not a parent since discriminator.propertyName is not set
                         hasAmbiguousParents = true;
-                        refedWithoutDiscriminator.add(parentName);
+                        Boolean isNotExtractedInlineSchema = !parentName.equals(schemaName+"_allOf");
```
TiFu: `boolean` instead of `Boolean`? Don't think we actually need an object here

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
```diff
@@ -28,6 +28,30 @@
...
 public class TestUtils {
 
+    /**
+     * Helper method for parsing specs as a generator would be presented at runtime (inline models resolved, flattened).
+     *
+     * @param specFilePath The path to the specification file
+     * @return A processed OpenAPI document
+     */
+    public static OpenAPI parseFlattenSpec(String specFilePath) {
+        OpenAPI openAPI = new OpenAPIParser().readLocation(specFilePath, null, new ParseOptions()).getOpenAPI();
```
TiFu: Call `TestUtils.parseSpec` instead of reimplementing parsing here?

spacether: The long term intention is to deprecate parseSpec. For that reason these functions should be independent of each other.

spacether: Is this request blocking? Why?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: This has been changed to type boolean
True we don't need an object but having it named conveys that we are checking that a schema isNotExtractedInlineSchema which helps with understanding.

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
TiFu: Sounds good.

jimschubert: > The long term intention is to deprecate, possibly remove, or change the location of parseSpec.

Not true. We will still need `parseSpec` for current and future classes which operate on "raw" spec files.

spacether: Sorry, I meant my long term intention, not ours. @jimschubert classes which operate on raw files aren't something that I know about. What are some examples? Are they not being tested right now?

jimschubert: InlineModelResolver, and as mentioned in your other PR, I will be reworking Spec processing using a transformer pipeline which will need this too.

spacether: Thank you for clarifying that. That other use case is newer and is not yet in the code path.
Per that conversation in the other PR this issue: https://github.com/OpenAPITools/openapi-generator/issues/843 documents the spec processing.

spacether: parseFlattenSpec has been updated to call parseSpec because we need to keep parseSpec as a public util

modules/openapi-generator/src/test/java/org/openapitools/codegen/utils/ModelUtilsTest.java
```diff
@@ -26,86 +26,84 @@
...
-        Assert.assertTrue(allUsedSchemas.contains("SomeHeader24"), "contains 'SomeHeader24'");
-        Assert.assertTrue(allUsedSchemas.contains("SomeObj25"), "contains 'SomeObj25'");
-        Assert.assertTrue(allUsedSchemas.contains("SomeObj26"), "contains 'SomeObj26'");
-        Assert.assertTrue(allUsedSchemas.contains("Param27"), "contains 'Param27'");
-        Assert.assertTrue(allUsedSchemas.contains("Param28"), "contains 'Param28'");
-        Assert.assertTrue(allUsedSchemas.contains("Parent30"), "contains 'Parent30'");
-        Assert.assertTrue(allUsedSchemas.contains("AChild30"), "contains 'AChild30'");
-        Assert.assertTrue(allUsedSchemas.contains("BChild30"), "contains 'BChild30'");
+        Assert.assertEquals(allUsedSchemas.size(), 42);
+        List<String> expectedallUsedSchemas = Arrays.asList(
```
jimschubert: Be careful when refactoring things like this. While YAML (as used in this method) may process schemas in a consistent order, JSON does not have a determinate order on object properties.

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
```diff
@@ -28,6 +28,30 @@
...
 
+    /**
+     * Helper method for parsing specs as a generator would be presented at runtime (inline models resolved, flattened).
+     *
+     * @param specFilePath The path to the specification file
+     * @return A processed OpenAPI document
+     */
+    public static OpenAPI parseFlattenSpec(String specFilePath) {
+        OpenAPI openAPI = parseSpec(specFilePath);
+        // resolve inline models
```
jimschubert: This comment is unnecessary.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1110,6 +1110,14 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
         int nullSchemaChildrenCount = 0;
         boolean hasAmbiguousParents = false;
         List<String> refedWithoutDiscriminator = new ArrayList<>();
+        String schemaName = "";
```
jimschubert: What is the point of this for loop? Why can't you just do `composedSchema.getName()` to get this schema's name? I think this needs a comment because it could be a pretty large performance impact on huge schema documents.

modules/openapi-generator/src/test/java/org/openapitools/codegen/utils/ModelUtilsTest.java
spacether: Thank you for sharing that json and yaml are processed differently. Do you want me to change this back to our old testing style so it will work if we swap in json in the future?

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
spacether: This comment has been removed

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: This for loop is needed because the Schema.name is null for all schemas in allSchemas.
Do you want me to edit the getSchemas method to set the names in the returned schemas?
Use composedSchema.getName() like you have described.

modules/openapi-generator/src/test/java/org/openapitools/codegen/utils/ModelUtilsTest.java
spacether: Or optionally I could cast them to sets and check equality of the sets.

spacether: This check has been changed to use List.containsAll to check presence rather than order.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
jimschubert: I see. No need to change, as I suspect ModelUtils will change pretty heavily by 5.0 anyway.

spacether:
Thanks for the review and merge!

wing328:
@spacether Looks like this breaks allOf in OpenAPI/Swagger 2.0 spec. For details, please refer to https://github.com/OpenAPITools/openapi-generator/issues/5876.

A workaround is to migrate the spec to 3.0.

Issue #5320, opened 2020-02-14
[BUG] Handlebars template engine doesn't generate SupportingFiles properly


lindgrenfredrik:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
       Any spec, any language using defaultGenerator that has supporting Files
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
      Issue related to Template engine and not schema
- [X] What's the version of OpenAPI Generator used?
          945ea14b2350e41bd14a846fd9f70cf5937822e8 from GIT
- [X] Have you search for related issues/PRs?
        Searched for Handlebars, found https://github.com/OpenAPITools/openapi-generator/issues/510
        Not match for specific issue but understand that Handlebars suppport may not be fully in place
- [X] What's the actual output vs expected output?
       Actual: Supporting files don't get compiled with template variables.
       Expected: Supporting files get compiled with template

##### Description

Have existing Templates for Handlebars that are used together with Swagger codegen and looking to migrate to openapi codegen instead

In generateSupportingFiles in DefaultGenerator.java it looks if files end with correct template extension before compiling them as templates otherwise just writing them as files.
This works fine when mustache files registered for the different languages end with correct mustache but when switching engine it expects handlebars extension which isn't whats registered by "language" generator
The name conversion is done later in current setup (that is why it works for i.e. model files), which never gets called for Supporting files.

Any ideas on how to solve this properly?
 - Allow mustache extension for any template engine?
 - Convert names earlier?
 - - Register template files in some other manner to append correct ending?
 - Run compile for all supporting files regardless of extension?

##### openapi-generator version

945ea14b2350e41bd14a846fd9f70cf5937822e8

##### Command line used for generation

generate -g aspnetcore -o tmp -i pet2.json -e handlebars -t aspnetcore3/2.1/

pet2.json is petstore swagger definition update to 2.0 format found here https://petstore.swagger.io/v2/swagger.json

templates for handlebars is existing mustache ones found in openapi repo but update for handlebars style (also available in my fork atm https://github.com/OpenAPITools/openapi-generator/compare/master...lindgrenfredrik:aspnetcore_handlebarsdiff-bd1652e990ffe227072a5c8908fd3054)

##### Steps to reproduce

run generate command above and check output in supporting files

##### Related issues/PRs

Non found

##### Suggest a fix

[...]

PR #5542, opened 2020-03-06
Move mustache template to new subfolder

### LINKED ISSUE #5541: [REQ][Handlebars] Move all mustache templates in resource folder to new mustache subfolder to better support multiple template Engines

Currently all templates are located in resources folder and all languages point to them there
If choosing to run different templating engine and provide those templates would today have to live in same folder as current mustache template.
This solution would make a clutter of files for every language that has supports multiple templating engines.

make a new substructure in resources folder for mustache templates matching identifier of the templating engine
Then new templates for other templating engines can be places in parallel folder structure related to their identifier (i.e. new folder for handlebars templates)

Add new info to Templating engine related to "folder offset" to find templates for that particular templating engine

lindgrenfredrik:
Move all mustache templates into subfolder called mustache (same as templateEngine identifier)
Updated so that all references to templates to point into new subfolder

fix #5541

### PR checklist

- [NA] If contributing template-only or documentation-only changes which will change sample output, [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) before.

bin/run-all-petstore reports [SUCCESS] 268 generators finished.

- [NA] Copy the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.


lindgrenfredrik:
@wing328 seems there was some timeout in ci/bitrise test job (which is linked below your user id), something I need to do to retrigger it?

jimschubert:
I don't follow why this has been done? We don't intend to support multiple template files for built-in templates, only Mustache. The structure of the templates under the resources directory has nothing to do with the support of other engines for custom templates.

lindgrenfredrik:
If you don't intend to support multiple template types for the same language being delivered as part of the openapi jar then this isn't needed.

My concern is related to if some language would like to setup support for both Mustache and Handlebars templates from inside the openapi jar and use the existing name conversion then templates ending with .mustache and .handlebars would have to be stored in the same folder structure.
This seemed messy and as such I saw that if the template engine name was part of the folder structure path then there would be a separation of where the different templates would be placed related to which template engine was chosen to executed

wing328:
@lindgrenfredrik first of all, thanks for the PR. I agreed with @jimschubert that it's unlikely we will support both mustache and handlebar. We may (tentatively) target cutting over to Handlebar in 6.0 release scheduled next year as the demand for handlebar isn't huge

lindgrenfredrik:
@wing328 thanks for feedback.
As mentioned in previous post I wasn't aware that the intention wasn't to support both template styles from inside the openapi jars, but rather only with external templates.

I'm working on a company internal project where we have already moved to handlebar templates with swagger, but we ran into other issues with their codegen not behaving correctly and those issues are not present with OpenApi codegen.

Since I saw that Handlebars was in Beta for OpenApi and got help with fix for #5320 I just assumed plan was to support both template engines in the future (from within the openapi jar) and that the separation of templates just hadn't occurred to anyone yet do to low usage of handlebars

Hence why I made this proposal.
But I'm fine with dropping it if that is not in the plan to handle this type of setup from within the openapi jars (will regardless have internal templates for the company project)

PR #5546, opened 2020-03-06
[Swift] fix URLSession file upload

4brunu:
The URLSession implementation was failing to upload files to the server.
The Alamofire implementation it's working.

In this PR the URLSession implementation is fixed.

### PR checklist


wing328:
@4brunu thanks for the fix. What about adding a test with the `uploadFile` in the Petstore API to ensure the file upload works moving forward?

4brunu:
Hey, I added some unit tests as you suggested 🙂

samples/client/petstore/swift5/alamofireLibrary/SwaggerClientTests/SwaggerClientTests/FileUtils.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  FileUtils.swift
+//  SwaggerClientTests
+//
+//  Created by Bruno Coelho on 11/03/2020.
+//  Copyright © 2020 Swagger. All rights reserved.
```
wing328: Please use `Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)` instead.

4brunu: Like this?

wing328: Yup

jerbob92:
@4brunu I'm not sure why this works for you, but the changes you did are not according to the multipart spec: https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

```
Note that the encapsulation boundary must occur at the beginning of a line, i.e., following a CRLF, and that that initial CRLF is considered to be part of the encapsulation boundary rather than part of the preceding part. The boundary must be followed immediately either by another CRLF and the header fields for the next part, or by two CRLFs, in which case there are no header fields for the next part (and it is therefore assumed to be of Content-Type text/plain).
NOTE: The CRLF preceding the encapsulation line is considered part of the boundary so that it is possible to have a part that does not end with a CRLF (line break). Body parts that must be considered to end with line breaks, therefore, should have two CRLFs preceding the encapsulation line, the first of which is part of the preceding body part, and the second of which is part of the encapsulation boundary.

The requirement that the encapsulation boundary begins with a CRLF implies that the body of a multipart entity must itself begin with a CRLF before the first encapsulation line -- that is, if the "preamble" area is not used, the entity headers must be followed by TWO CRLFs. This is indeed how such entities should be composed. A tolerant mail reading program, however, may interpret a body of type multipart that begins with an encapsulation line NOT initiated by a CRLF as also being an encapsulation boundary, but a compliant mail sending program must not generate such entities.

Encapsulation boundaries must not appear within the encapsulations, and must be no longer than 70 characters, not counting the two leading hyphens.
```
I read it like the format should be:

{--boundary} {newline} {header1} {newline} {header2} {newline} {newline} {body} {newline} {--boundary} {newline} {header1} {newline} {header2} {newline} {newline} {body} {newline} {--boundary--} {newline}

This is also what other http clients do that I reviewed. 
However, your change does this:

{--boundary} {newline} {header1} {newline} {newline} {header2} {newline} {newline} {body} {newline} {newline}  {--boundary} {newline} {header1} {newline} {newline} {header2} {newline} {newline} {body} {newline} {newline} {--boundary--}

The differences are:
 - One newline too much between headers
 - One newline too much after the body

The multipart form data parser in Go (our backend) does not parse your generated body. When I change it to mine it does work properly. 

The best way to compare is to use two generated clients, one with URLSession and one with Alamofire, and then post the form data to a RequestBin to review the differences.

If you want I can create a pull request.

4brunu:
@jerbob92 can you please create a PR so I can test them on my side?
From what I read in your comment, I think you are right.

jerbob92:
@4brunu since I would have to do some stuff before I can submit the PR, I didn't yet.
You can review the changes in my branch: https://github.com/jerbob92/openapi-generator/tree/feature/swift-urlsession-file-upload

If you think that's a good fix I will create a PR.

4brunu:
@jerbob92 I just tested your branch and it works.
Could you please submit a PR with that change please?
Thanks

Issue #5080, opened 2020-01-22
In php, date type query parameters formatted as date-time


paul-utley:
##### Description

When using a parameter with the format 'date', it is incorectly formatted as a date-time. In my situation, we're passing 2 dates in the query parameters, and our api is receiving the parameters formatted as date-time and rejecting the request.

##### openapi-generator version

Version 4.2.0

##### OpenAPI declaration file content or url
```json
"parameters": [{
                        "type": "string",
                        "format": "date",
                        "in": "query",
                        "name": "start_date"
                    },
                    {
                        "type": "string",
                        "format": "date",
                        "in": "query",
                        "name": "end_date"
                    }
                ]
```

##### Command line used for generation

openapi-generator generate -i test.json -g php -o .

##### Steps to reproduce
1. Use the parameters array from above to generate a simple api definition to do a get request.

If you look at where the query parameters are serialized, you can see that the object serializer formats the datetime object as a ISO8601 date regardless if the query parameter is defined as a date or datetime type in the api definition.

##### Related issues/PRs
This following issue in TypeScript is exactly what I am experiencing. I just wanted to see that it was reported in php as well.
https://github.com/OpenAPITools/openapi-generator/issues/4763
##### Suggest a fix/enhancement

Need a way to pass the format from the query parameter to the object serializer so it can format it correctly as a date i.e. 2020-01-01 vs a datetime i.e. 2020-01-01T00:00:00Z

jpswade:
I've not been able to figure out where this applies in this codebase.

Instead I found the generator templates over here, and raised a bug/PR to fix it:

https://github.com/swagger-api/swagger-codegen-generators/issues/621

jpswade:
Scrap that, found it here:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache

jpswade:
I raised a PR: https://github.com/OpenAPITools/openapi-generator/pull/5579

PR #5579, opened 2020-03-11
Add ability to pass the Data Format through to the Query formatter

jpswade:
This is so you can pass the data format through to the query value formatter method, toQueryValue.

See: https://github.com/OpenAPITools/openapi-generator/issues/5080

### PR checklist


jpswade:
I can't seem to add reviewers, so paging @jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) ❤️, @ybelenko (2018/07), @renepardon (2018/12)

ybelenko:
@jpswade I've rebased your PR on latest master but cannot push. Is this PR submitted with "Allow access to maintainers" checkmark?

jpswade:
@ybelenko 
Screenshot 2020-05-14 at 17 41 08

ybelenko:
Seems like failed unit test.
```console
$ vendor/bin/phpunit tests
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.........F....................................................... 65 / 69 ( 94%)
....                                                              69 / 69 (100%)

Time: 1.5 seconds, Memory: 10.00 MB

There was 1 failure:

1) OpenAPI\Client\DateTimeSerializerTest::testDateSanitazion
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'1973-04-30T17:05:00+02:00'
+'1973-04-30'

/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/php/OpenAPIClient-php/tests/DateTimeSerializerTest.php:38

FAILURES!
Tests: 69, Assertions: 201, Failures: 1.
```

@jpswade Can you take a look?

jpswade:
I've fixed the test but for some reason `ensure-up-to-date` does not pass. I don't understand enough about this codebase to understand why it's not working.

ybelenko:
> I've fixed the test but for some reason `ensure-up-to-date` does not pass. I don't understand enough about this codebase to understand why it's not working.

Sometimes `ensure-up-to-date` gives me incorrect output too. Maybe it somehow related to Java Runtime Machine version. Which is more weird `ensure-up-to-date` with `--batch` option sometimes gives different output. You should take seriously only uncommitted changes in folders of updated codegen(with `php` in path in current PR). Anyway, you'll understand if something wrong in CI logs attached to current PR.

jpswade:
Think I've cracked it. It's structured in quite a complicated way, so it was difficult to locate where the data types were being defined.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPhpCodegen.java
```diff
@@ -115,8 +115,8 @@ public AbstractPhpCodegen() {
         typeMapping.put("string", "string");
         typeMapping.put("byte", "int");
         typeMapping.put("boolean", "bool");
-        typeMapping.put("date", "\\DateTime");
-        typeMapping.put("Date", "\\DateTime");
+        typeMapping.put("date", "date");
```
ybelenko, line 118: That is a significant change as soon as it overrides all PHP codegens(php client, lumen, laravel, symfony, slim4, php-ze). I guess it can break many unit tests.

I don't know perfectly how php client works, so I should ask @ackintosh or @wing328 to review.

ybelenko: I would say that mapping format `date` and `date-time` to `\DateTime` PHP instance was a mistake at the first place, but I don't know how to solve it without breaking changes.

Problems with `\DateTime` auto converting that I can recall is:
- When client sends `date-time` value without timezone like `2010-01-01T00:00:00` it can be parsed with default server timezone and if it's different we can end up with different hours or even date.
- You don't know which date time format client used because generated code already converted it to `\DateTime` without ability to get source string.
- You always not sure that `DateTime` is equals to original request parameter because of different datetime formats and timezone(especially when TZ is omitted).

jpswade: The mapping of date-time to DateTime is OK, this is expected because it'll default to the expected format of DateTime::ATOM.

But for 'date', it needs to be 'date', otherwise the system will think that it is a DateTime and return the date and time.

This all started because the date it was returning was incorrect, it was the wrong format, it was returning a date time rather than the expected date.

The date format issue that you're describing with the timezone really seems to be more of an issue with the protocol than the generator.

The specification is very clear that the date-time format should be RFC3339 format, and the timezone issue is negated if you always use UTC.

ybelenko: Imagine Public API of booking service. User wants to create appointment at 9 o'clock Berlin time. Application configured to send remind notifications 2 hours before. You suggest to always convert date time to UTC value, yes, it's doable, but it will look confusing in the database and you cannot force all companies to use autogenerated client library, so there is no guarantee that developer sends date time value in RFC3339 format.

Anyway, it's just my personal view on date time issues. I can use `type: string` without `format: datetime` if I don't like autogenerated implementations.

jpswade: I think that’s an issue with the protocol itself.

Anyway, the problem here is not with datetimes, it’s with date.

A date should be a date only, not a date time.

ybelenko: Technically there is no `date` class in php, only `\DateTime`. I would map it to `string` scalar type to avoid syntax errors.

jpswade: You're right, I wasn't sure if it was a "php" type or some kind of virtual type that openapi would make up. For example, there's no such data type as "map".

I'll change it to string, as I think that makes the most sense here.

Issue #5381, opened 2020-02-20
[BUG] [spring][server] oneOf Polymorphism support for spring boot


dkirrane:
##### Description

Currently, openapi-generator doesn't support `oneOf`. It currently generates the following broken code, outputting `UNKNOWN_BASE_TYPE` rather than `Pet` for the example below.
The code is broken with or without the `discriminator` in the Spec.

```java
    default ResponseEntity<Void> _petsPost(@ApiParam(value = ""  )  @Valid @RequestBody(required = false) UNKNOWN_BASE_TYPE UNKNOWN_BASE_TYPE) {
        return petsPost(UNKNOWN_BASE_TYPE);
    }
```

##### openapi-generator version
`pom.xml`
```xml
<plugin>
    <groupId>org.openapitools</groupId>
    <artifactId>openapi-generator-maven-plugin</artifactId>
    <version>4.2.3</version>
    <executions>
        <execution>
            <id>generate</id>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>${project.basedir}/src/main/resources/api.yaml</inputSpec>
                <generatorName>spring</generatorName>
                <generateSupportingFiles>true</generateSupportingFiles>
                <configOptions>
                    <sourceFolder>src/gen/java/main</sourceFolder>
                    <library>spring-boot</library>
                    <java8>true</java8>
                    <interfaceOnly>true</interfaceOnly>
                    <skipDefaultInterface>false</skipDefaultInterface>
                    <delegatePattern>true</delegatePattern>
                </configOptions>
            </configuration>
        </execution>
    </executions>
</plugin>
```

##### OpenAPI Spec
`api.yaml`
```yaml
openapi: "3.0.2"
info:
  version: 1.0.0
  title: Polymorphism
paths:
  /pets:
    post:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
              discriminator:
                propertyName: pet_type
      responses:
        '200':
          description: created
components:
  schemas:
    # Parent
    Pet:
      type: object
      required:
        - pet_type      
      properties:
        pet_type:
          type: string
        eats:
          type: string
      discriminator:
        propertyName: pet_type
    # Child 1
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Dog`
          properties:
            bark:
[...]

PR #5596, opened 2020-03-15
[Spring] OneOf Polymorphism support for spring boot 

alexsuperdev:
Fix [5381](https://github.com/OpenAPITools/openapi-generator/issues/5381)
Implemented OneOf Polymorphism for java spring library.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
@@ -21,6 +21,12 @@ import com.samskivert.mustache.Mustache;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.PathItem;
+import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.ComposedSchema;
+import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.parameters.RequestBody;
+import io.swagger.v3.oas.models.responses.ApiResponse;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.openapitools.codegen.*;
 import org.openapitools.codegen.languages.features.BeanValidationFeatures;
@@ -29,6 +35,7 @@ import org.openapitools.codegen.languages.features.PerformBeanValidationFeatures
 import org.openapitools.codegen.meta.features.*;
 import org.openapitools.codegen.templating.mustache.SplitStringLambda;
 import org.openapitools.codegen.templating.mustache.TrimWhitespaceLambda;
+import org.openapitools.codegen.utils.ModelUtils;
 import org.openapitools.codegen.utils.URLPathUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -98,6 +105,8 @@ public class SpringCodegen extends AbstractJavaCodegen
     protected boolean hateoas = false;
     protected boolean returnSuccessCode = false;
     protected boolean unhandledException = false;
+    protected boolean useOneOfInterfaces = false;
+    protected List<CodegenModel> addOneOfInterfaces = new ArrayList<CodegenModel>();
 
     public SpringCodegen() {
         super();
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -579,6 +588,251 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
                 }
             }
         }
+
+        if (useOneOfInterfaces && openAPI.getComponents() != null){
```
bkabrda: Since this method calls `super.preprocessOpenAPI`, I don't think you should need to include this block at all. Is there a specific reason why you added it here?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -579,6 +588,251 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
...
+                    if (addProps != null && ModelUtils.isComposedSchema(addProps)) {
+                      addOneOfNameExtension(addProps, nOneOf);
+                      addOneOfInterfaceModel((ComposedSchema) addProps, nOneOf);
+                    }
+                }
+            }
+        }
+    }
+
+    public void addOneOfNameExtension(Schema s, String name) {
```
bkabrda: AFAICS, this is the same as `DefaultCodegen.addOneOfNameExtension` and so I think it can be safely omitted.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -579,6 +588,251 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
...
+        ComposedSchema cs = (ComposedSchema) s;
+        if (cs.getOneOf() != null && cs.getOneOf().size() > 0) {
+          cs.addExtension("x-oneOf-name", name);
+        }
+    }
+
+    public void addOneOfInterfaceModel(ComposedSchema cs, String name) {
```
bkabrda: Same as above - this is already implemented in `DefaultCodegen.addOneOfInterfaceModel`, so let's omit it here (unless there's a specific reason to override it).

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
```diff
@@ -579,6 +588,251 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
...
+            return getAlias(schemaName);
+          } else {
+            LOGGER.warn("Error obtaining the datatype from ref:" + unaliasSchema.get$ref() + ". Default to 'object'");
+            return "object";
+          }
+        }
+    return null;
+    }
+
+    private class OneOfImplementorAdditionalData {
```
bkabrda: Same as above, I don't think this is really necessary here.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/SpringCodegen.java
alexsuperdev: I added it because SpringCodegenTest.testAdditionalPropertiesPutForConfigValues and 
SpringCodegenTest.testInitialConfigValues fails with NPE  otherway.

alexsuperdev:
> 
> 
> Hey 👋 IIUC a lot of code added is a copy-paste from the old version of the `oneOf` implementation that I wrote for `JavaClientCodegen`. Since then, most of the functionality was moved to `DefaultCodegen`, so you really only need to make sure the overloading methods properly call their parent's implementation (specifically `postProcessOpenAPI` and `postProcessAllModels`). Then it should be matter of setting `useOneOfInterfaces = true` and perhaps overriding the `addImportsToOneOfInterface` method (see how this is done in `JavaClientCodegen`).
> 
> I hope this makes sense.



> 
> 
> Hey 👋 IIUC a lot of code added is a copy-paste from the old version of the `oneOf` implementation that I wrote for `JavaClientCodegen`. Since then, most of the functionality was moved to `DefaultCodegen`, so you really only need to make sure the overloading methods properly call their parent's implementation (specifically `postProcessOpenAPI` and `postProcessAllModels`). Then it should be matter of setting `useOneOfInterfaces = true` and perhaps overriding the `addImportsToOneOfInterface` method (see how this is done in `JavaClientCodegen`).
> 
> I hope this makes sense.

Hey @bkabrda! Thank you for your codereview of my pullrequest.
You are totally right, I copy-paste your code from JavaClientCodegen.
I founded your commits in JavaClientCodegen and didn't know that you implemented that already in DefaultCodegen class, from where i can extend this stuff in SpringCodegen class.
Do you think it will be okay if i cancel this pullrequest and make a new one to master, where i will reuse your methods ?

bkabrda:
>  Do you think it will be okay if i cancel this pullrequest and make a new one to master, where i will reuse your methods ?

Sure, I think that's perfectly fine.

> How your implementation works without overriding fromRequestBody?
> Without this one i had import of UNKNOWN_BASE_TYPE in the generated OneOf Model and
UNKNOWN_BASE_TYPE as Type Parameter in API class

TBH I don't have a good answer. I can definitely take a closer look when time permits, but I have tons of work right now and can't promise anything.

Issue #2398, opened 2019-03-14
[C#] aspnetcore / server: Property type binary generates Stream instead of IFormFile


jwieneke:
##### Description

I try to generate a ASP .NET Core server side library for file uploads.
All examples I've seen so far are using the C# type `IFormFile` to receive file content.
However if I declare an interface with content type `multipart/form-data` and a property of type `string` with format `binary` (see yaml snippet) openapi generates a method parameter of type `` `System.IO.Stream`.
If I try uploading a file using this interface I am getting an exception:
```
System.InvalidOperationException: Could not create an instance of type 'System.IO.Stream'. Model bound complex types must not be abstract or value types and must have a parameterless constructor. Alternatively, give the 'image' parameter a non-null default value.
   at Microsoft.AspNetCore.Mvc.ModelBinding.Binders.ComplexTypeModelBinder.CreateModel(ModelBindingContext bindingContext)
   at Microsoft.AspNetCore.Mvc.ModelBinding.Binders.ComplexTypeModelBinder.BindModelCoreAsync(ModelBindingContext bindingContext)
   at Microsoft.AspNetCore.Mvc.ModelBinding.ParameterBinder.BindModelAsync(ActionContext actionContext, IModelBinder modelBinder, IValueProvider valueProvider, ParameterDescriptor parameter, ModelMetadata metadata, Object value)
   at Microsoft.AspNetCore.Mvc.Internal.ControllerBinderDelegateProvider.<>c__DisplayClass0_0.<<CreateBinderDelegate>g__Bind|0>d.MoveNext()
```

Am I missing something or is there any kind of workaround (make `System.IO.Stream` working or forcing openapi to use `IFormFile`)?

##### openapi-generator version

3.3.4

##### OpenAPI declaration file content or url

[openapi.yaml](https://gist.github.com/jwieneke/036fb15a9a3a913b30021e16b8d40d99)

##### Command line used for generation

`java -jar openapi-generator-cli-3.3.4.jar generate -i openapi.yaml -g aspnetcore --model-name-suffix DTO`

##### Steps to reproduce

1. Use `openapi.yaml` to call command line.
2. Have a look at the generated file `FileUploadApiController`.
3. If you want to reproduce the exception in action: Create a small implementation an try posting a file to `http://127.0.0.1/api/upload`

##### Related issues/PRs

There is a [similar issue](https://github.com/OpenAPITools/openapi-generator/issues/1381) about problems with files on client side in C#.
[...]

PR #5636, opened 2020-03-19
[C#] [aspnetcore] Option to have files as IFormFile instead of Stream

lindgrenfredrik:
Added option to aspnetcore generator to have it generate file data type as IFormFile instead of Stream

fix: #2398

### PR checklist

- [NA] If contributing template-only or documentation-only changes which will change sample output, [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) before.
Added new script bin/aspnetcore-binary.sh

cc
@mandrean 
@jimschubert 
@frankyjuang 
@shibayan 

Question: Do I need to merge the sample output from the test script into the repo as well?

### DIFF
--- a/bin/aspnetcore-binary-iformfile.sh
+++ b/bin/aspnetcore-binary-iformfile.sh
@@ -27,6 +27,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate -g aspnetcore -i modules/openapi-generator/src/test/resources/3_0/form-multipart-binary-array.yaml -t modules/openapi-generator/src/main/resources/aspnetcore/2.1/ -o samples/server/binary/aspnetcore --additional-properties packageGuid={3C799344-F285-4669-8FD5-7ED9B795D5C5} --additional-properties useIFormFileInsteadOfStream=true $@"
+ags="generate -g aspnetcore -i modules/openapi-generator/src/test/resources/3_0/form-multipart-binary-array.yaml -t modules/openapi-generator/src/main/resources/aspnetcore/2.1/ -o samples/server/binary/aspnetcore-iformfile --additional-properties packageGuid={3C799344-F285-4669-8FD5-7ED9B795D5C5} --additional-properties useIFormFileInsteadOfStream=true $@"
 
 java $JAVA_OPTS -jar $executable $ags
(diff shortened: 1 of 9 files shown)


wing328:
@lindgrenfredrik thanks for the PR. What about using the `typeMapping` option directly instead?

lindgrenfredrik:
@wing328 if you mean the command line option like: --type-mappings file=Microsoft.AspNetCore.Http.IFormFile

First I wasn't aware of that command line option.
Second, it doesn't seem to work right since it looses the . in the mapping

lindgrenfredrik:
Adding both Type mapping and import mapping made it possible to generate without this change.

Command line options:
--type-mappings file=IFormFile --import-mappings IFormFile=Microsoft.AspNetCore.Http.IFormFile"

Hence this change is not needed as same result can be achieved with command line options.
Though perhaps documentation should be updated to include this as a example in some manner

jmnbo-iscteiulpt:
> Adding both Type mapping and import mapping made it possible to generate without this change.
> 
> Command line options:
> --type-mappings file=IFormFile --import-mappings IFormFile=Microsoft.AspNetCore.Http.IFormFile"
> 
> Hence this change is not needed as same result can be achieved with command line options.
> Though perhaps documentation should be updated to include this as a example in some manner

Won't this just make all files as IFormFile?
I tried this option but it changes both uploads and downloads to IFormFile and I believe there is no way to send a IFormFile from .net backend to a frontend

lindgrenfredrik:
Yes options make both upload and download be IFormFile
Code does that to, so there is no difference there.

If IFormFile is used as type in the SwaggerResponse/Produces Swagger UI reports correct schema type of string($binary)
While if type in SwaggerResponse/Produces is current Stream type return schema indicates:
System.IO.Stream{
canRead	boolean
readOnly: true
canSeek	boolean
readOnly: true
canTimeout	boolean
readOnly: true
canWrite	boolean
readOnly: true
length	integer($int64)
readOnly: true
position	integer($int64)
readTimeout	integer($int32)
writeTimeout	integer($int32)
}

Current template sets return value in Controllers to IActionResult so change between Stream and IFormFile only has impact on "documentation" part atm and as such doesn't limit from sending Valid return type to still allow download

Though if developers look at documentation in the generated code they will probably question the return value since  even if the return such a Type it fails.

So the correct solution would probably be to have different type listed for Results (i.e. FileResult as example given by SwashBuckle indicates https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/fefe65217c24a7194c0588a2d430a8952ed791b2/test/WebSites/Basic/Controllers/FilesController.cs) and update input parameter to be IFormFile instead since at least Kestrel setup of dotnet core can't handle file upload as Stream

wing328:
> Though perhaps documentation should be updated to include this as a example in some manner

Agreed with you the documentation can be improved.

PR #5651, opened 2020-03-20
[C++] [Qt5] fixed cpp-client-qt5 HttpRequestWorker requests crashing on timeout...

onze:
… when they have actually NOT timed out (were calling back into a deleted struct).

Old behavior:
- request *without* timeout on a responding url: get an answer back, works as expected.
- request *without* timeout on a non-responding url: get no answer back, no callback called, works as expected.
- request *with* timeout on a non-responding url: get no answer back, timeout gets triggered, works as expected.
- request *with* timeout on a responding url: get answer back, reply gets deleted, timeout gets triggered, callback writes on deleted data -> crash.

Patch content: replaced the timeout integer with a QTimer to handle and own the timeout callback. The timer is stopped whenever a call gets an answer back, or the owner object is deleted.

This has only been tested empirically since the patch has a small scope.

Copying the technical committee: @ravinikam @stkrwork @etherealjoy @martindelille @muttleyxd

Thank you

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
@@ -45,7 +45,7 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
 }
 
 {{prefix}}HttpRequestWorker::{{prefix}}HttpRequestWorker(QObject *parent)
-    : QObject(parent), manager(nullptr), _timeOut(0) {
+    : QObject(parent), manager(nullptr), timeOutTimer() {
     qsrand(QDateTime::currentDateTime().toTime_t());
     manager = new QNetworkAccessManager(this);
     workingDirectory = QDir::currentPath();
@@ -53,6 +53,7 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
 }
 
 {{prefix}}HttpRequestWorker::~{{prefix}}HttpRequestWorker() {
+    QObject::disconnect(&timeOutTimer, nullptr, nullptr, nullptr);
     for (const auto &item : multiPartFields) {
         if (item != nullptr) {
             delete item;
@@ -86,8 +87,11 @@ QByteArray *{{prefix}}HttpRequestWorker::getMultiPartField(const QString &fieldn
     return nullptr;
 }
 
-void {{prefix}}HttpRequestWorker::setTimeOut(int timeOut) {
-    _timeOut = timeOut;
+void {{prefix}}HttpRequestWorker::setTimeOut(int timeOutMs) {
+    timeOutTimer.setInterval(timeOutMs);
+    if(timeOutTimer.interval() == 0) {
+        QObject::disconnect(&timeOutTimer, nullptr, nullptr, nullptr);
+    }
 }
 
 void {{prefix}}HttpRequestWorker::setWorkingDirectory(const QString &path) {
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
```diff
@@ -45,14 +45,15 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
 }
 
 {{prefix}}HttpRequestWorker::{{prefix}}HttpRequestWorker(QObject *parent)
-    : QObject(parent), manager(nullptr), _timeOut(0) {
+    : QObject(parent), manager(nullptr), timeOutTimer() {
```
etherealjoy: I think you might want to set the parent for `timeOutTimer`

modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
```diff
@@ -86,8 +87,11 @@ QByteArray *{{prefix}}HttpRequestWorker::getMultiPartField(const QString &fieldn
     return nullptr;
 }
 
-void {{prefix}}HttpRequestWorker::setTimeOut(int timeOut) {
-    _timeOut = timeOut;
+void {{prefix}}HttpRequestWorker::setTimeOut(int timeOutMs) {
+    timeOutTimer.setInterval(timeOutMs);
+    if(timeOutTimer.interval() == 0) {
```
etherealjoy: is this needed?

modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
```diff
@@ -347,12 +351,14 @@ void {{prefix}}HttpRequestWorker::execute({{prefix}}HttpRequestInput *input) {
         buffer->setParent(reply);
 #endif
     }
-    if (_timeOut > 0) {
-        QTimer::singleShot(_timeOut, [=]() { on_manager_timeout(reply); });
+    if (timeOutTimer.interval() > 0) {
+        timeOutTimer.callOnTimeout([=]() { on_manager_timeout(reply); });
```
etherealjoy: This is available only on 5.12 and above. Is there a possibility to use a more backward compatible call?

modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
```diff
@@ -347,12 +351,14 @@ void {{prefix}}HttpRequestWorker::execute({{prefix}}HttpRequestInput *input) {
...
     }
 }
 
 void {{prefix}}HttpRequestWorker::on_manager_finished(QNetworkReply *reply) {
+    if(timeOutTimer.interval() > 0)
```
etherealjoy: Need to stop the timer here too.

etherealjoy:
@onze Thanks for the PR. I left some comments please have a look if it makes sense.

onze:
Thanks for the review.
I added the timer a a QObject child, reaplaced the call to callOnTimeout with something a bit more backward compatible, and added proper calls to QTimer::stop when disconnecting.
I testing this locally and the patch still fixes the issue, we should be good to go as far as I know.

etherealjoy:
@onze 
I guess you are running windows, so I have updated the generated files using the script.

onze:
Yeah it's windows at work here, sorry about that. I'll try to upload those files as well in future PR, thanks for taking care of this for me!

-- change of topic
I wanted to discuss the possibility of pushing larger changes, namely a solution to optional parameters in queries, which are simply not taken into account in this c++ client. This would be a larger PR/discussion, should I simply open a PR and we can talk about it there?

etherealjoy:
> I wanted to discuss the possibility of pushing larger changes, namely a solution to optional 
> parameters in queries (all parameters are considered required in this c++ client). This would be a 
> larger PR/discussion, should I simply open a PR and we can talk about it there?

You can open an issue and we can discuss there including your solution, then open a PR to fix the issue.

onze:
Will do in the coming days/weeks after this sprint then. Thanks!

wing328:
CI failures already addressed in the master

wing328:
@onze thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

PR #5679, opened 2020-03-23
[Python] Fix Python UTF-8 Encoding Issue

CrshOverride:
Currently, the Python client generator assumes that, in Python3, all responses are going to be UTF-8 encoded strings. This results in run-time errors when using APIs that return other content types (such as `image/gif`, etc). More details can be found in the following issue:

https://github.com/OpenAPITools/openapi-generator/issues/206

### PR checklist

@taxpon 
@frol
@mbohlool 
@cbornet
@kenjones-cisco
@tomplus
@Jyhess
@slash-arun
@spacether


spacether:
You can close and reopen the PR rather than doing a one line change to kick off CI

spacether:
Hello, thank you for writing this PR. Can you please add tests to python and python-experimental showing that receiving a bytes payload works? I think that I had already added a test of receiving a file in one of the python-experimental samples. Can you also add a test of sending a file? Should we add a test which sends form data including a file and non file object?

samples/client/petstore/python-asyncio/docs/FakeApi.md
```diff
@@ -490,7 +490,7 @@ with petstore_api.ApiClient(configuration) as api_client:
     number = 3.4 # float | None
 double = 3.4 # float | None
 pattern_without_delimiter = 'pattern_without_delimiter_example' # str | None
-byte = 'byte_example' # str | None
+byte = None # bytes | None
```
spacether: Why is this None? Can we add a byte example here?

samples/client/petstore/python-experimental/tests/test_api_exception.py
```diff
@@ -50,7 +51,8 @@ def test_404_error(self):
         except ApiException as e:
             self.assertEqual(e.status, 404)
             self.assertEqual(e.reason, "Not Found")
-            self.checkRegex(e.body, "Pet not found")
+            body = e.body.decode('utf-8') if six.PY3 else e.body
+            self.checkRegex(body, "Pet not found")
```
spacether: So we used to have e.body decoded here.
Your change as-is is breaking for the python generator because you are changing the type of e.body.
Sorry, I know that in the related issue I suggested putting it in the api client but I didn't know that it wouldbe a breaking change.
Can you move you decoding code into rest.mustache to make this a non-breaking change?

CrshOverride: @spacether I actually tried to put the decoding in `rest.py` but there's no access to `response_type` there (which is what is used to determine if something should be decoded or not). Would adding `response_type` as a parameter to `call_api` be preferable to this method (I'm assuming both are breaking changes so I'm not sure which is the better spot)?

CrshOverride:
@spacether For the tests, it looks like there's already a test for file upload using form encoding with and without other data: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/tests/test_pet_api.py#L244

What's the proper way to add methods to the test api that the tests use so that I can add a download method (which would return the bytes payload)?

spacether:
> @spacether For the tests, it looks like there's already a test for file upload using form encoding with and without other data: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/tests/test_pet_api.py#L244
> 
> What's the proper way to add methods to the test api that the tests use so that I can add a download method (which would return the bytes payload)?

Ah, thanks for pointing me to that uploading a file test.
For those tests at ests/test_pet_api.py:test_upload_file can you add this response.body checking to
- python-asyncio
- python-experimental

To add a test of file deserialization you can start from [this example](https://github.com/OpenAPITools/openapi-generator/blob/cfe3b86d70cd038c7e4e073d57c750b9d0189376/samples/client/petstore/python-experimental/tests/test_deserialization.py#L337)

Make sure that you are adding test methods in the tests/blah.py file name. Those files have been manually added by developers. The tests in test/blah.py are autogenerated.
Tests you write in either location will not be overwritten when samples are regenerated.

samples/client/petstore/python-experimental/tests/test_api_exception.py
spacether: Hmmm, what if we captured the raised exception in api_client:
- decoded the exception body
- re-raised the exception?
Then we have a nonbreaking change.

spacether:
@CrshOverride just checking in. Are you still interested in getting this merged?

CrshOverride:
@spacether Definitely! I've just been swamped with other issues at the moment. Hopefully will have the requested changes in the next couple of days.

spacether:
> @spacether Definitely! I've just been swamped with other issues at the moment. Hopefully will have the requested changes in the next couple of days.

Great, I am looking forward to getting it merged. This has been something that users have been asking for for a while. I'm here for you if a chat or review helps. Thanks!

CrshOverride:
@spacether I just pushed the breaking-change fix and a test for bytes -> json serialization. Let me know if that's on the right path (I'll fight CI in the mean time).

CrshOverride:
@spacether I'm having some issues after merging `master` into my fork. It looks like running `./bin/utils/ensure-all-updated.sh` is reverting the version number on files from 4.3.1 to 4.3.0 which results in a bunch of changes that I don't think belong.

samples/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -377,6 +380,36 @@ def get_header(name, default=None):
         finally:
             os.unlink(file_path)
 
+    def test_deserialize_binary(self):
```
spacether, line 383: Can you rename this deserialize_binary_to_str ?
The current name implies that we expect to receive binary data and process it as binary in downstream consumers.

spacether: Can you add a separate test where we keep the response data in bytes?
I'm not sure if we should return an open file object if bytes are received we chose to store them in a file object before because a file was sent. How about keeping non-file bytes in a value of type bytes?
If I were receiving bytes as a user I would not expect them to be stored in a file object.

samples/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -377,6 +380,36 @@ def get_header(name, default=None):
...
+            reason='OK',
+            data=json.dumps(data).encode("utf-8") if six.PY3 else json.dumps(data),
+            getheaders=get_headers,
+            getheader=get_header
+        )
+
+        mock_response = RESTResponse(http_response)
+
+        binary = self.deserialize(mock_response, response_types_mixed, True)
+        self.assertEqual(isinstance(binary, str), True)
```
spacether, line 410: Why would this sometimes be binary and sometime be str?
We are setting the response_types_mixed as str only.

samples/client/petstore/python-experimental/tests/test_deserialization.py
spacether: How do we know that sending bytes will work?

spacether:
@CrshOverride I don't want to add too much scope to this PR if it holds us back.
Would you rather add the bytes type here or separate it out into a separate PR?
This one could be capturing the exception and delaying the decoding until we need to.
The other one would be adding the bytes type and demonstrating that we can send and receive bytes.
It's also fine if you want to keep the bytes addition it here if you want.
What do you want to do?

CrshOverride:
@spacether I originally wrote that test to return a raw binary from `deserialize` but then I realized that the current logic seems to always return a file in those cases so I wasn't made sure to test `bytes` being returned raw. If possible, I'd like to get this in and then submit a separate PR for returning bytes as needed.

spacether:
> @spacether I originally wrote that test to return a raw binary from `deserialize` but then I realized that the current logic seems to always return a file in those cases so I wasn't made sure to test `bytes` being returned raw. If possible, I'd like to get this in and then submit a separate PR for returning bytes as needed.

Great, sounds good.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -119,6 +119,7 @@ public PythonClientCodegen() {
         languageSpecificPrimitives.add("object");
         // TODO file and binary is mapped as `file`
         languageSpecificPrimitives.add("file");
+        languageSpecificPrimitives.add("bytes");
```
spacether, line 122: No longer needed because we are only concerned with bytes in the python layer when grabbing the raw resonse?

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -186,22 +186,36 @@ class ApiClient(object):
...
+                _request_timeout=_request_timeout)
+
+            content_type = response_data.getheader('content-type')
+
+            self.last_response = response_data
+
+            return_data = response_data
+            if _preload_content:
+                if six.PY3:
+                    if response_type not in ["file", "bytes"]:
```
spacether, line 204: Should we do this for file-type only now?

CrshOverride: @spacether If possible, I'd like to leave the groundwork for the next PR if at all possible (if it's not hurting anything).

spacether: As long as we remove the bytes addition from the python generator in the java files then I am fine with that.
That would make it so we would never see this "bytes" type until your later PR.

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
```diff
@@ -176,26 +176,30 @@ class ApiClient(object):
...
+                if response_type:
+                    return_data = self.deserialize(
+                        response_data,
+                        response_type,
+                        _check_type
+                    )
+                else:
+                    return_data = None
+        except ApiException as e:
+            e.body = e.body.decode('utf-8') if six.PY3 else e.body
```
spacether, line 201: Style nit (not necessary to change):
```
if six.PY3:
  e.body = e.body.decode('utf-8')
```
For the else case we don't need to reassign the value of e.body

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
spacether: Adding the type here without handling it in python-experimental beans that we will have a new unhandled type.
Can you defer this to the next PR?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -134,10 +135,6 @@ public PythonClientCodegen() {
         typeMapping.put("DateTime", "datetime");
         typeMapping.put("object", "object");
         typeMapping.put("file", "file");
-        // TODO binary should be mapped to byte array
-        // mapped to String as a workaround
-        typeMapping.put("binary", "str");
-        typeMapping.put("ByteArray", "str");
```
spacether, line 140: Can we add these back in?

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -186,22 +186,36 @@ class ApiClient(object):
...
+                method, url, query_params=query_params, headers=header_params,
+                post_params=post_params, body=body,
+                _preload_content=_preload_content,
+                _request_timeout=_request_timeout)
+
+            content_type = response_data.getheader('content-type')
+
+            self.last_response = response_data
+
+            return_data = response_data
```
spacether, line 201: Do we need to handle the _preload_content==False case with something like this?
```
if not _preload_content:
  return response_data
# dedented remaining code without the if _preload_content block
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -25,6 +25,7 @@ if six.PY3:
     # these are needed for when other modules import str and int from here
     str = str
     int = int
+    bytes = bytes
```
spacether, line 28: Can you remove this and the bytes definition on line 37?

CrshOverride: @spacether I originall put it there for consistency with the rest of the definitions (str and int). What's the reasoning not to also include bytes (and wouldn't not including here break imports in other locations)?

spacether: We aren't using this bytes definition yet though, right?

samples/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -377,6 +377,36 @@ def get_header(name, default=None):
...
+            status=200,
+            reason='OK',
+            data=json.dumps(data).encode("utf-8") if six.PY3 else json.dumps(data),
+            getheaders=get_headers,
+            getheader=get_header
+        )
+
+        mock_response = RESTResponse(http_response)
+
+        binary = self.deserialize(mock_response, response_types_mixed, True)
```
spacether, line 406: The current variable name makes it sound like it should contain a binary value when it contains a string value.
How about changing the variable binary to response or result or deserialized?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
```diff
@@ -53,8 +53,20 @@
...
+            {{^isBinary}}
+                {{#isByteArray}}
+                'pattern': r'{{{vendorExtensions.x-regex}}}',  # noqa: E501{{#vendorExtensions.x-modifiers}}
+                {{#-first}}'flags': (re.{{.}}{{/-first}}{{^-first}}          re.{{.}}{{/-first}}{{^-last}} | {{/-last}}{{#-last}}){{/-last}}{{/vendorExtensions.x-modifiers}}
+                {{/isByteArray}}
+                {{^isByteArray}}
+                'pattern': r'{{{vendorExtensions.x-regex}}}',  # noqa: E501{{#vendorExtensions.x-modifiers}}
+                {{#-first}}'flags': (re.{{.}}{{/-first}}{{^-first}}          re.{{.}}{{/-first}}{{^-last}} | {{/-last}}{{#-last}}){{/-last}}{{/vendorExtensions.x-modifiers}}
+                {{/isByteArray}}
+            {{/isBinary}}
```
spacether, line 69: Is this working?
I am not seeing any samples have binary regex patterns.

CrshOverride: @spacether Without any of the other `bytes` stuff, this isn't actually needed anymore. It's another thing kept for the sake of the next PR (so the logic/need isn't forgotten).

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -186,22 +186,44 @@ class ApiClient(object):
...
+            if not _preload_content:
+                {{^tornado}}
+                return return_data
+                {{/tornado}}
+                {{#tornado}}
+                raise tornado.gen.Return(return_data)
+                {{/tornado}}
+
+            if six.PY3:
+                if response_type not in ["file", "bytes"]:
```
spacether, line 212: Can you combine these two if statements?
We don't need two levels of indentation here.
You can just use:
`if six.PY3 and response_type not in {"file", "bytes"}:`
Note: I also switched to set definition with {} for quicker lookups.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
spacether: When we separate pieces of code like this, I worry that we will forget to include tests or samples that demonstrate them working.

- If you have it here can you add a spec example which creates binary regex patterns in the python-experimental sample?
- Otherwise please remove this from here and include it in your next PR

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -186,22 +186,43 @@ class ApiClient(object):
...
-        return_data = response_data
-        if _preload_content:
             # deserialize response data
             if response_type:
                 return_data = self.deserialize(response_data, response_type)
             else:
                 return_data = None
+        except ApiException as e:
+            e.body = e.body.decode('utf-8') if six.PY3 else e.body
+            raise e
```
spacether, line 225: Is this only raised by the `self.request` call?
If so can we move this up to be right after that call?
That way all of the other code can be moved out of this try block because it doesn't need to be inside it.

samples/client/petstore/python/test/test_format_test.py
```diff
@@ -36,19 +36,18 @@ def make_instance(self, include_optional):
...
             return FormatTest(
-                integer = 1E+1, 
-                int32 = 2E+1, 
-                int64 = 56, 
-                number = 32.1, 
-                float = 54.3, 
-                double = 67.8, 
-                string = 'a', 
-                byte = 'YQ==', 
-                binary = bytes(b'blah'),
```
spacether, line 47: Why is the binary assignment gone?

CrshOverride: If we're removing returning binary, it shouldn't still be there should it?

modules/openapi-generator/src/main/resources/python/python-experimental/api_client.mustache
```diff
@@ -176,26 +176,39 @@ class ApiClient(object):
...
-                return_data = None
+
+        if not _preload_content:
+            {{^tornado}}
+            return (return_data)
+            {{/tornado}}
+            {{#tornado}}
+            raise tornado.gen.Return(return_data)
+            {{/tornado}}
+            return return_data
```
spacether, line 201: It looks like we need the decoding here also for python-experimental for non-file types.
What do you think?

samples/client/petstore/python/test/test_format_test.py
spacether: Let's still keep the field, but just give it a string value.
Per [here](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python/petstore_api/models/format_test.py#L43)
the bytes variable exists in the model and accepts a string value.
When you do your next PR we will see the type there change to byes, and we can update the sample value for this parameter here.

wing328:
@CrshOverride thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

PR #5699, opened 2020-03-24
tests  for models for C-libcurl generator

michelealbano:
This PR first of all adds supports for models that contain other models as properties. In fact, this yaml could not be used with the master:
https://gist.github.com/michelealbano/bd705d2bf5a1c36dde3e514a1c744f13

Moreover, I added support for the generation of tests for the generated models.
I had to mix template and C, since I have access to different data from the 2 environments.
From C I can build a "default" example, and/or decorate strings with double quotes.
To be able to support objects into objects (a property of an "external" object is an "internal" object itself), I had to play with `#ifdef` and `#include` clauses: 
- I scan the properties (vars) of the external object, and for each internal object I will preliminary add the prototype of the function `instantiate_{{classname}}` that can instantiate the object before the generating `instantiate_{{classname}}` of the external object. I will then call the `instantiate_{{classname}}` where I am instantiating the external object.
- I `#include` together all the test*.c files that I need, to build the a.out for the test. I will use `#ifdef` clauses to have just 1 main and not to repeat the function definition.

I had to change the name of the test*.c to play with `#include` of the files.

@ityuhui @zhemant 

By the way, I wrote a python script to generate, compile and test a yaml file:
https://gist.github.com/michelealbano/51e40ca2c403969bf98fb1ef7e2520c5
Example of usage: 
`python ../doAllTests.py /mnt/hgfs/shared/miaroba/*.yaml`

The PR is not complete - I still have to remove printf etc - but I wanted your comments.

### PR checklist

### DIFF
--- a/samples/client/petstore/c/unit-test/test_category.c
+++ b/samples/client/petstore/c/unit-test/test_category.c
@@ -1,10 +1,10 @@
-#ifndef _TEST
-#define _TEST
+#ifndef category_TEST
+#define category_TEST
 
 // the following is to include only the main from the first c file
 #ifndef TEST_MAIN
 #define TEST_MAIN
-#define _MAIN
+#define category_MAIN
 #endif // TEST_MAIN
 
 #include <stdlib.h>
@@ -14,12 +14,16 @@
 #include "../external/cJSON.h"
 
 
+
 #include "../model/category.h"
 category_t* instantiate_category(int include_optional);
 
 
+
 category_t* instantiate_category(int include_optional) {
-  category_t* category = category_create(
+  category_t* category = NULL;
+if (include_optional) {
+  category = category_create(
 56
 //primitive
 , // id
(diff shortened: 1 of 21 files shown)


samples/client/petstore/c/unit-test/test_category.c
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+category_t* instantiate_category(int include_optional) {
+  category_t* category = NULL;
+if (include_optional) {
+  category = category_create(
+56
+//primitive
+, // id
```
wing328: What about putting the comments in the same line as the example value (e.g. `56, // id (primitive)`) ?

michelealbano: This is the plan. I created the PR to receive comments while I developed the approach further.

wing328: 👌

michelealbano:
@wing328  Done, ready to merge

michelealbano:
@wing328  do you need any more work to merge this PR?

wing328:
@michelealbano thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

Issue #4690, opened 2019-12-03
[BUG] Ruby models order in inheritance case


gigaga:
##### Description

By generating a ruby client, the models order defined in `openapi_client.rb` is incorrect in case where there is some inheritance.

##### openapi-generator version

```bash
$ openapi-generator version
4.2.2
```

##### OpenAPI declaration file content or url

petstore.yaml

```yaml
openapi: 3.0.0
info:
  version: 1.0.0-SNAPSHOT
  title: petstore
  description: |
    This is the specification of Petstore with the following resources:

    - Pet
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      tags:
        - pets
      responses:
        "200":
          description: A paged array of pets
          content:
            application/json:
              schema:
                type: object
                description: 1.0.0
                required:
                  - pet_type
                properties:
                  pet_type:
                    type: string
                discriminator:
                  propertyName: pet_type
                  mapping:
                    dog: "#/components/schemas/Dog"
                    cat: "#/components/schemas/Cat"
                    lizard: "#/components/schemas/Lizard"
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
        startDate:
          type: string
          example: "2019-11-08"
        num:
          type: integer
          example: 1
      discriminator:
        propertyName: pet_type
        mapping:
          dog: "#/components/schemas/Dog"
          cat: "#/components/schemas/Cat"
          lizard: "#/components/schemas/Lizard"
    Dog:
      allOf:
        - $ref: "#/components/schemas/Pet"
        - type: object
          properties:
            bark:
              type: string
    Cat:
      allOf:
        - $ref: "#/components/schemas/Pet"
        - type: object
          properties:
            name:
              type: string
    Lizard:
      allOf:
        - $ref: "#/components/schemas/Pet"
        - type: object
          properties:
            lovesRocks:
              type: boolean

```

##### Command line used for generation

```bash
openapi-generator generate --generator-name ruby -i petstore.yaml
```

##### Steps to reproduce

1. After generating, build gem `gem build openapi_client.gemspec`
1. Install it `gem install openapi_client-1.0.0.gem`
1. Use it in your ruby code in `test.py` file for example: `require 'openapi_client'
1. Execute it `ruby test.py`

The following error has occured: `uninitialized constant OpenapiClient::Pet (NameError)`

[...]

PR #5706, opened 2020-03-25
[ruby] fix oneOf handling

### LINKED ISSUE #3630: [BUG][Ruby] Using oneOf generated code has unexpected results
#### Bug Report Checklist

##### Description

When a `oneOf` is used, the generated model only includes one of the possible classes. Using the client, objects for all class types are forced into the same class, even if they're a different class. 

In the below example, if the API returns a `AuthenticationMonitor` MonitorResult object, the generated client forces it into the `SmokeTestMonitor` class type. This is a problem as it's the incorrect class type and as a result the model attributes are all `nil`.

###### Current output for the `AuthenticationMonitor` MonitorResult object:
```
...
@items=[#<ClientExample::SmokeTestResult:0x00007fb45220eaa0>]
...
```

###### Expected output for the `AuthenticationMonitor` MonitorResult object:
```
...
@items=[#<ClientExample::AuthenticationMonitorItems:0x00007fb45220eaa0 @result=true>]
...
```

4.1.0, seems to occur in every 4.x version. Haven't checked previous versions.

<details>
  <summary> OpenAPI spec</summary>

`AuthenticationMonitor` and `SmokeTestMonitor` are the same but the property name inside `items` is different
```yaml
openapi: 3.0.2
info:
  title: Example
  version: v1
paths:
  /endpoint:
    get:
      summary: Get All Monitors
      responses:
        '200':
          description: Successful operation.
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Monitor'
components:
  schemas:
    Monitor:
      type: object
      properties:
        results:
          $ref: '#/components/schemas/MonitorResults'
    MonitorResults:
      oneOf:
        - $ref: '#/components/schemas/AuthenticationMonitor'
        - $ref: '#/components/schemas/SmokeTestMonitor'
    AuthenticationMonitor:
      type: object
      properties:
        items:
          type: array
          items:
            type: object
            properties:
              result:
                type: boolean
    SmokeTestMonitor:
      type: object
      properties:
        items:
          type: array
          description: Collection of smoke test result checks.
          items:
            type: object
            properties:
              passed:
                type: boolean
```
</details>

`openapi-generator generate -i <OPENAPI_SPEC> -g ruby -o <TEMP_DIR> -c <CONFIG_PATH>`

##### Steps to reproduce

1. Run the command to generate the Ruby client
2. Look at the 

##### Related issues/PRs

Maybe https://github.com/OpenAPITools/openapi-generator/issues/15 is related, but this issue is more specific to the Ruby client.

Another issue related to `oneOf` for the Ruby client is the generation of objects within a `response` schema. If you have the following OpenAPI spec, the class type gets generated as `ExampleClient::OneOfsimpleObjectcomplexObject`, and that class does not exist.

<details>
  <summary> OpenAPI spec snippet</summary>
Here is the spec:

```yaml
paths:
  /endpoint:
    get:
      summary: Get All Monitors
      responses:
        '200':
          description: Successful operation.
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/simpleObject'
                  - $ref: '#/components/schemas/complexObject'

components:
  schemas:
    simpleObject:
      type: object
      properties:
        simple_property1:
          type: string
    complexObject:
      type: object
      properties:
        complex_property1:
          type: string
        complex_property2:
          type: string
```
</details>

##### Suggest a fix

In https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/ruby-client/partial_model_generic.mustache#L49, the generated client only includes one option `Array<SmokeTestResult>`. I'm not sure about a specific fix, but has there been any Ruby-specific development into this area?

jfeltesse-mdsol:
Fixes #3630.

~I have yet to update the documentation related templates and update the samples but opening now to gather initial feedback on the direction.~ 👌 

~Basically, this builds up on the work from @bkabrda on the java and go experimental clients.
I didn't go "ruby-experiemental" because the support of oneOf in the current ruby client is broken anyway.~ _not doing this in the end._

How it works is, in the spirit of the "oneof as interfaces" this creates a **module**, not a class, for oneOf models and provides a set of class methods related to the functionality.
The core method is `build_from_hash` whose job in this case is to find the relevant oneOf item and transparently building and returning the correct model.
It does so by using the discriminator if present, mapping if present and "falls back" to the existing validation-based look up.

~I changed some test yaml files to include more oneOf examples.~ _no need in the end._

If I try to do a before/after using the modified `composed-one.yaml` file, it goes like (some parts trimmed for brevity):

`openapi-generator generate -i composed-oneof.yaml -g ruby --library faraday -o ./out/ruby_oneof`

### model files
```sh
# current master
custom_one_of_schema.rb
obj_a.rb
obj_b.rb
obj_c.rb
obj_d.rb

# this branch
custom_one_of_array_schema_one_of.rb
custom_one_of_schema.rb
obj_a.rb
obj_b.rb
obj_c.rb
obj_c_data_one_of.rb
obj_d.rb
```

### custom_one_of_schema.rb
```ruby
# current master
module OpenapiClient
  class CustomOneOfSchema
    attr_accessor :realtype

    attr_accessor :message

    attr_accessor :description

    attr_accessor :code

    # Attribute mapping from ruby-style variable name to JSON key.
    def self.attribute_map
      {
        :'realtype' => :'realtype',
        :'message' => :'message',
        :'description' => :'description',
        :'code' => :'code'
      }
    end

    # Attribute type mapping.
    def self.openapi_types
      {
        :'realtype' => :'String',
        :'message' => :'String',
        :'description' => :'String',
        :'code' => :'Integer'
      }
    end

    # List of attributes with nullable: true
    def self.openapi_nullable
      Set.new([
      ])
    end

    # List of class defined in oneOf (OpenAPI v3)
    def self.openapi_one_of
      [
      :'ObjA',
      :'ObjB'
      ]
    end

    # discriminator's property name in OpenAPI v3
    def self.openapi_discriminator_name
      :'realtype'
    end

    # Initializes the object
    # @param [Hash] attributes Model attributes in the form of hash
    def initialize(attributes = {})

# another 200 lines of model code...
```

```ruby
# this branch
module OpenapiClient
  module CustomOneOfSchema
    # List of class defined in oneOf (OpenAPI v3)
    def self.openapi_one_of
      [
        :'ObjA',
        :'ObjB'
      ]
    end

    # discriminator's property name in OpenAPI v3
    def self.openapi_discriminator_name
      :'realtype'
    end

    # discriminator's mapping in OpenAPI v3
    def self.openapi_discriminator_mapping
      {
        :'a-type' => :'ObjA',
        :'b-type' => :'ObjB'
      }
    end

    # Builds the object from hash
    # @param [Hash] attributes Model attributes in the form of hash
    # @return [Object] Returns the model itself
    def self.build_from_hash(attributes)
      discriminator_value = attributes[openapi_discriminator_name]
      return nil unless discriminator_value

      _class = openapi_discriminator_mapping[discriminator_value.to_sym]
      return nil unless _class

      OpenapiClient.const_get(_class).build_from_hash(attributes)
    end
  end

end
```

Noteworthy is in the obj_c.rb the nested oneOf property has its model generated, albeit with a sligthly different name:

```ruby
# current master
    def self.openapi_types
      {
        :'realtype' => :'String',
        :'data' => :'OneOfObjAObjB'  # <==== model not generated!
      }
    end

# this branch
    def self.openapi_types
      {
        :'realtype' => :'String',
        :'data' => :'ObjCDataOneOf'  # <==== "interface" module generated
      }
    end
```

### Limitations / bugs / food for thought

- Interestingly, a `custom_one_of_array_schema_one_of.rb` "interface" module is now generated but the oneOf is taking place inside the `items`. I'm not sure this solution is going to work here. In this case maybe a regular model should be generated alongside some `custom_one_of_array_schema_one_of_items.rb` "interface" module?
  @bkabrda how do the java and experimental go clients fare in this case?
- In java the interface pattern used in this PR makes the models referenced in a oneOf carry the "dynamic" props illustrated in ObjD. https://github.com/OpenAPITools/openapi-generator/pull/4695 sort of discourages that pattern but I guess if the tool can do it, why not. What could be improved in the case of ruby though is to make some "dynamicProps" module in the interface module and dynamically include the props at runtime upon deserializing. Performance may not be stellar so it's trade off between "pureness" of the models and speed I guess.

### PR checklist

@cliffano (2017/07) @zlx (2017/09) @autopp (2019/02)

### DIFF
--- a/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
@@ -2,9 +2,7 @@
     # @param [Hash] attributes Model attributes in the form of hash
     # @return [Object] Returns the model itself
     def self.build_from_hash(attributes)
-      new(attributes).build_from_hash(attributes)
-    rescue ArgumentError
-      nil
+      new.build_from_hash(attributes)
     end
 
     # Builds the object from hash
--- a/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
@@ -2,7 +2,9 @@
     # @param [Hash] attributes Model attributes in the form of hash
     # @return [Object] Returns the model itself
     def self.build_from_hash(attributes)
-      new.build_from_hash(attributes)
+      new(attributes).build_from_hash(attributes)
+    rescue ArgumentError
+      nil
     end
 
     # Builds the object from hash
@@ -93,7 +95,7 @@
           is_nullable = self.class.openapi_nullable.include?(attr)
           next if !is_nullable || (is_nullable && !instance_variable_defined?(:"@#{attr}"))
         end
-        
+
         hash[param] = _to_hash(value)
       end
       hash
--- a/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
@@ -69,7 +69,9 @@
           end
         end
       else # model
-        {{moduleName}}.const_get(type).build_from_hash(value)
+        # models (e.g. Pet) or oneOf
+        klass = {{moduleName}}.const_get(type)
+        klass.respond_to?(:openapi_one_of) ? klass.build(value) : klass.build_from_hash(value)
       end
     end
 
--- a/modules/openapi-generator/src/main/resources/ruby-client/model_test.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/model_test.mustache
@@ -11,19 +11,13 @@ require 'date'
 # Please update as you see appropriate
 {{#models}}
 {{#model}}
-describe '{{classname}}' do
-  before do
-    # run before each test
-    @instance = {{moduleName}}::{{classname}}.new
-  end
-
-  after do
-    # run after each test
-  end
+describe {{moduleName}}::{{classname}} do
+{{^oneOf}}
+  let(:instance) { {{moduleName}}::{{classname}}.new }
 
   describe 'test an instance of {{classname}}' do
     it 'should create an instance of {{classname}}' do
-      expect(@instance).to be_instance_of({{moduleName}}::{{classname}})
+      expect(instance).to be_instance_of({{moduleName}}::{{classname}})
     end
   end
 {{#vars}}
@@ -33,7 +27,7 @@ describe '{{classname}}' do
       # assertion here. ref: https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers
       # validator = Petstore::EnumTest::EnumAttributeValidator.new('{{{dataType}}}', [{{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}}, {{/-last}}{{/enumVars}}{{/allowableValues}}])
       # validator.allowable_values.each do |value|
-      #   expect { @instance.{{name}} = value }.not_to raise_error
+      #   expect { instance.{{name}} = value }.not_to raise_error
       # end
       {{/isEnum}}
       {{^isEnum}}
@@ -43,6 +37,42 @@ describe '{{classname}}' do
   end
 
 {{/vars}}
+{{/oneOf}}
+{{#oneOf}}
+{{#-first}}
+  describe '.openapi_one_of' do
+    it 'lists the models referenced in the oneOf array' do
+      expect(described_class.openapi_one_of).to_not be_empty
+      described_class.openapi_one_of.each { |klass| expect { {{moduleName}}.const_get(klass) }.to_not raise_error }
+    end
+  end
+
+  {{#discriminator}}
+  {{#propertyName}}
+  describe '.openapi_discriminator_name' do
+    it 'returns the value of the "discriminator" property' do
+      expect(described_class.openapi_discriminator_name).to_not be_empty
+    end
+  end
+
+  {{/propertyName}}
+  {{#mappedModels}}
+  {{#-first}}
+  describe '.openapi_discriminator_mapping' do
+    it 'returns the key/values of the "mapping" property' do
+      expect(described_class.openapi_discriminator_mapping.values.sort).to eq(described_class.openapi_one_of.sort)
+    end
+  end
+
+  {{/-first}}
+  {{/mappedModels}}
+  {{/discriminator}}
+  describe '.build' do
+    it 'returns the correct model' do
+    end
+  end
+{{/-first}}
+{{/oneOf}}
 end
 {{/model}}
 {{/models}}
--- a/modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
@@ -1,0 +1,130 @@
+  {{#description}}
+  # {{{description}}}
+  {{/description}}
+  module {{classname}}
+    class << self
+      {{#oneOf}}
+      {{#-first}}
+      # List of class defined in oneOf (OpenAPI v3)
+      def openapi_one_of
+        [
+      {{/-first}}
+          :'{{{.}}}'{{^-last}},{{/-last}}
+      {{#-last}}
+        ]
+      end
+
+      {{/-last}}
+      {{/oneOf}}
+      {{#discriminator}}
+      {{#propertyName}}
+      # Discriminator's property name (OpenAPI v3)
+      def openapi_discriminator_name
+        :'{{{.}}}'
+      end
+
+      {{/propertyName}}
+      {{#mappedModels}}
+      {{#-first}}
+      # Discriminator's mapping (OpenAPI v3)
+      def openapi_discriminator_mapping
+        {
+      {{/-first}}
+          :'{{{mappingName}}}' => :'{{{modelName}}}'{{^-last}},{{/-last}}
+      {{#-last}}
+        }
+      end
+
+      {{/-last}}
+      {{/mappedModels}}
+      {{/discriminator}}
+      # Builds the object
+      # @param [Mixed] Data to be matched against the list of oneOf items
+      # @return [Object] Returns the model or the data itself
+      def build(data)
+      {{#discriminator}}
+        discriminator_value = data[openapi_discriminator_name]
+        return nil unless discriminator_value
+      {{#mappedModels}}
+      {{#-first}}
+
+        klass = openapi_discriminator_mapping[discriminator_value.to_sym]
+        return nil unless klass
+
+        {{moduleName}}.const_get(klass).build_from_hash(data)
+      {{/-first}}
+      {{/mappedModels}}
+      {{^mappedModels}}
+        {{moduleName}}.const_get(discriminator_value).build_from_hash(data)
+      {{/mappedModels}}
+      {{/discriminator}}
+      {{^discriminator}}
+        # Go through the list of oneOf items and attempt to identify the appropriate one.
+        # Note:
+        # - We do not attempt to check whether exactly one item matches.
+        # - No advanced validation of types in some cases (e.g. "x: { type: string }" will happily match { x: 123 })
+        #   due to the way the deserialization is made in the base_object template (it just casts without verifying).
+        # - TODO: scalar values are defacto behaving as if they were nullable.
+        # - TODO: logging when debugging is set.
+        openapi_one_of.each do |klass|
+          begin
+            next if klass == :AnyType # "nullable: true"
+            typed_data = find_and_cast_into_type(klass, data)
+            return typed_data if typed_data
+          rescue # rescue all errors so we keep iterating even if the current item lookup raises
+          end
+        end
+
+        openapi_one_of.include?(:AnyType) ? data : nil
+      {{/discriminator}}
+      end
+      {{^discriminator}}
+
+      private
+
+      SchemaMismatchError = Class.new(StandardError)
+
+      # Note: 'File' is missing here because in the regular case we get the data _after_ a call to JSON.parse.
+      def find_and_cast_into_type(klass, data)
+        return if data.nil?
+
+        case klass.to_s
+        when 'Boolean'
+          return data if data.instance_of?(TrueClass) || data.instance_of?(FalseClass)
+        when 'Float'
+          return data if data.instance_of?(Float)
+        when 'Integer'
+          return data if data.instance_of?(Integer)
+        when 'Time'
+          return Time.parse(data)
+        when 'Date'
+          return Date.parse(data)
+        when 'String'
+          return data if data.instance_of?(String)
+        when 'Object' # "type: object"
+          return data if data.instance_of?(Hash)
+        when /\AArray<(?<sub_type>.+)>\z/ # "type: array"
+          if data.instance_of?(Array)
+            sub_type = Regexp.last_match[:sub_type]
+            return data.map { |item| find_and_cast_into_type(sub_type, item) }
+          end
+        when /\AHash<String, (?<sub_type>.+)>\z/ # "type: object" with "additionalProperties: { ... }"
+          if data.instance_of?(Hash) && data.keys.all? { |k| k.instance_of?(Symbol) || k.instance_of?(String) }
+            sub_type = Regexp.last_match[:sub_type]
+            return data.each_with_object({}) { |(k, v), hsh| hsh[k] = find_and_cast_into_type(sub_type, v) }
+          end
+        else # model
+          const = {{moduleName}}.const_get(klass)
+          if const
+            model = const.build_from_hash(data)
+            return model if model && model.valid?
+          end
+        end
+
+        raise # if no match by now, raise
+      rescue
+        raise SchemaMismatchError, "#{data} doesn't match the #{klass} type"
+      end
+      {{/discriminator}}
+    end
+  end
--- a/modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
@@ -116,8 +116,15 @@
         else # model
           const = {{moduleName}}.const_get(klass)
           if const
-            model = const.build_from_hash(data)
-            return model if model && model.valid?
+            if const.respond_to?(:openapi_one_of) # nested oneOf model
+              model = const.build(data)
+              return model if model
+            else
+              # raise if data contains keys that are not known to the model
+              raise unless (data.keys - const.acceptable_attributes).empty?
+              model = const.build_from_hash(data)
+              return model if model && model.valid?
+            end
           end
         end
 
--- a/bin/configs/ruby-faraday.yaml
+++ b/bin/configs/ruby-faraday.yaml
@@ -1,7 +1,7 @@
 generatorName: ruby
 outputDir: samples/client/petstore/ruby-faraday
 library: faraday
-inputSpec: modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
 templateDir: modules/openapi-generator/src/main/resources/ruby-client
 additionalProperties:
   gemVersion: 1.0.0
--- a/bin/configs/ruby-faraday.yaml
+++ b/bin/configs/ruby-faraday.yaml
@@ -1,7 +1,7 @@
 generatorName: ruby
 outputDir: samples/client/petstore/ruby-faraday
 library: faraday
-inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
 templateDir: modules/openapi-generator/src/main/resources/ruby-client
 additionalProperties:
   gemVersion: 1.0.0
--- a/bin/configs/ruby.yaml
+++ b/bin/configs/ruby.yaml
@@ -1,7 +1,7 @@
 generatorName: ruby
 outputDir: samples/client/petstore/ruby
 library: typhoeus
-inputSpec: modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
 templateDir: modules/openapi-generator/src/main/resources/ruby-client
 additionalProperties:
   gemVersion: 1.0.0
--- a/bin/configs/ruby.yaml
+++ b/bin/configs/ruby.yaml
@@ -1,7 +1,7 @@
 generatorName: ruby
 outputDir: samples/client/petstore/ruby
 library: typhoeus
-inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
 templateDir: modules/openapi-generator/src/main/resources/ruby-client
 additionalProperties:
   gemVersion: 1.0.0
--- a/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api_client.mustache
@@ -149,8 +149,9 @@ module {{moduleName}}
           data.each { |k, v| hash[k] = convert_to_type(v, sub_type) }
         end
       else
-        # models, e.g. Pet
-        {{moduleName}}.const_get(return_type).build_from_hash(data)
+        # models (e.g. Pet) or oneOf
+        klass = {{moduleName}}.const_get(return_type)
+        klass.respond_to?(:openapi_one_of) ? klass.build(data) : klass.build_from_hash(data)
       end
     end
 
--- a/modules/openapi-generator/src/main/resources/ruby-client/api_doc.mustache
+++ b/modules/openapi-generator/src/main/resources/ruby-client/api_doc.mustache
@@ -4,10 +4,13 @@
 
 All URIs are relative to *{{basePath}}*
 
-Method | HTTP request | Description
-------------- | ------------- | -------------
-{{#operations}}{{#operation}}[**{{operationId}}**]({{classname}}.md#{{operationId}}) | **{{httpMethod}}** {{path}} | {{#summary}}{{summary}}{{/summary}}
-{{/operation}}{{/operations}}
+| Method | HTTP request | Description |
+| ------ | ------------ | ----------- |
+{{#operations}}
+{{#operation}}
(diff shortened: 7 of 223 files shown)


gommo:
Does this PR address the below part in the original bug?

> Another issue related to oneOf for the Ruby client is the generation of objects within a response schema. If you have the following OpenAPI spec, the class type gets generated as ExampleClient::OneOfsimpleObjectcomplexObject, and that class does not exist.

jfeltesse-mdsol:
@gommo ah right, that's not addressed. Even the java generator with improved oneOf handling fails in this case. In both languages it refers to some "EndpointGET200" model that doesn't exist.

@bkabrda is this a known issue of the current implementation or should I open a new issue specifically for this?

wing328:
> is this a known issue of the current implementation or should I open a new issue specifically for this?

Better open an issue with details (e.g. mininal spec) to track the issue.

jfeltesse-mdsol:
@wing328 I just opened https://github.com/OpenAPITools/openapi-generator/issues/5903

wing328:
@jfeltesse-mdsol thanks for the PR. When you've time, can you please PM me via Slack to discuss this PR?

jfeltesse-mdsol:
any chance to get this reviewed/merged?

wing328:
Tested with a spec with an oneOf schema containing null:
```
    def self.openapi_one_of
      [
        :'Cat',
        :'DogDot',
        :'Null'
      ]
    end
```

Something that we need to address later (not this PR) to properly handle nullable oneOf schemas.

wing328:
I did a test with primary types in oneOf schemas and looks like your current approach assumes the schemas defined in oneOf are models (with properties). Something we need to handle/address in future PRs.

wing328:
If discriminator.propertyName is set to `petType`, I got the following code in the oneOf model:
```
    # discriminator's property name in OpenAPI v3
    def self.openapi_discriminator_name
      :'pet_type'
    end
```

I don't think this will work as the `pet_type` won't match the attribute/field `petType` in the JSON payload. Can you please take a look when you've time?

jfeltesse-mdsol:
> I did a test with primary types in oneOf schemas and looks like your current approach assumes the schemas defined in oneOf are models (with properties). Something we need to handle/address in future PRs.

Are you referring to the case where the oneOf members are other types like array or scalars? How are the other languages handling this?

wing328:
Example:
```
    One.Of.Primitive.Types:
      oneOf:
      - type: string
      - type: integer
```

Java (jersey2), Go, PowerShell generator should be able to handle this without issues.

wing328:
If I understand correctly, your current approach relies on `discriminator` (which is optional) so we will need to come up with the implementation to support `oneOf` without the `discriminator`.

jfeltesse-mdsol:
> If I understand correctly, your current approach relies on discriminator (which is optional) so we will need to come up with the implementation to support oneOf without the discriminator.

Actually it does work without a discriminator, the `build_from_hash` logic goes in this order:
- discriminator + mapped model
- discriminator alone (e.g. discriminator value is a model name)
- if no discriminator, for each oneOf item try to validate the model and return the first one that validates.

wing328:
That's good to know.

> if no discriminator, for each oneOf item try to validate the model and return the first one that validates.

Should it go through the rest to ensure there's only one match (to fulfill the definitions of `oneOf`)?

wing328:
For the spec used by the Ruby petstore samples listed below, can you please switch it to `modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml` instead which covers many more test cases for oneOf, anyOf models?

- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/configs/ruby.yaml#L4
- https://github.com/OpenAPITools/openapi-generator/blob/master/bin/configs/ruby-faraday.yaml#L4

jfeltesse-mdsol:
> Should it go through the rest to ensure there's only one match (to fulfill the definitions of oneOf)?

I don't think so, the logic I described is used when deserizalizing the response and doing validation at that stage brings no benefit. On the request side you wouldn't use the module directly but rather pass the relevant data directly. There, alongside the existing client side validations, checking the oneOf promise may make sense but it should be done in a separate PR to keep things bite-sized.

wing328:
> I don't think so, the logic I described is used when deserizalizing the response 

So `anyOf` and `oneOf` are essentially the same with respect to the logic you described?

jfeltesse-mdsol:
> > I don't think so, the logic I described is used when deserizalizing the response
> 
> So `anyOf` and `oneOf` are essentially the same with respect to the logic you described?

Correct. 

As a user though, I have a hard time picturing how validation when receiving can be useful. Imagine I'm using an autogenerated client to access an external API I have no control over. What good does it do to me that even though I received some data, the client throws an error (and prevents me from accessing the data?) because the API's openapi spec file incorrectly advertises oneOf when it really should be anyOf?
As long as the data I receive can be deserialized properly, I don't really care whether it happens to also match another schema. It's a bit arbitrary because you might expect the 2nd item in the oneOf list and get the 1st one (because it checks in order) but then the clients are not AI-powered, there's a minimum expectation that the openapi spec matches the server side implementation, isn't there?

Having said that though, it's not a huge challenge to validate that part. How the client should behave when encountering such cases though? Throw an error? Return null? Return the first valid item and log a warning? some other way??

wing328:
> What good does it do to me that even though I received some data, the client throws an error (and prevents me from accessing the data?) 

The user can then report the user to the API owner to look into the error as the response (payload) doesn't conform to the contract/spec (`oneOf` in this case). If multiple matches are expected, then anyOf should be used instead.

> there's a minimum expectation that the openapi spec matches the server side implementation

Sorry that it's not something I can agree with. Both client and server should conform to the contract (openapi spec).

> How the client should behave when encountering such cases though?

In Java, Go, PowerShell implementation, the client throws an exception or returns an error.

modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
```diff
@@ -0,0 +1,69 @@
...
+      {{moduleName}}.const_get(_class).build_from_hash(attributes)
+    {{/-first}}
+    {{/mappedModels}}
+    {{^mappedModels}}
+      {{moduleName}}.const_get(discriminator_value).build_from_hash(attributes)
+    {{/mappedModels}}
+    {{/discriminator}}
+    {{^discriminator}}
+      openapi_one_of.each do |_class|
+        model = {{moduleName}}.const_get(_class).build_from_hash(attributes)
```
zippolyte, line 62: This should be wrapped in a try/rescue. `build_from_hash` for enum values can raise. And when that's the case, the loop should continue to try the next model.

jfeltesse-mdsol: > build_from_hash for enum values can raise

can you provide an example?

jfeltesse-mdsol:
heads up here, I've not abandoned this PR, it's just that I got super busy at work. I might be able to go back to this in a couple of weeks.

modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
zippolyte: It's in here https://github.com/OpenAPITools/openapi-generator/blob/a0a7464786eb727d1ca4b483ad711bc79bc2bdf6/modules/openapi-generator/src/main/resources/ruby-client/partial_model_enum_class.mustache#L17

wing328:
@jfeltesse-mdsol please take your time. We appreciate your contributions to this.

modules/openapi-generator/src/main/resources/ruby-client/partial_oneof_module.mustache
zippolyte: Oh one more thing I noticed, as I'm trying out this PR on our spec: the `oneOf` types can be primitive types like `Float`, `String`, ... so we can't call `build_from_hash` on those. Maybe you could add some special cases, comparing the `openapi_one_of` value to primitive types, and try to directly parse the value in that case ?

jfeltesse-mdsol: Yes, this has been pointed out and it is something I plan to address https://github.com/OpenAPITools/openapi-generator/pull/5706#issuecomment-695869515

zippolyte: Ah right, sorry for the noise 🙈

jfeltesse-mdsol:
@wing328 About https://github.com/OpenAPITools/openapi-generator/pull/5706#issuecomment-696094795, I rebased my branch with latest master, re-built the generator and updated the inputSpec with

```diff
--- a/bin/configs/ruby-faraday.yaml
+++ b/bin/configs/ruby-faraday.yaml
@@ -1,7 +1,7 @@
 generatorName: ruby
 outputDir: samples/client/petstore/ruby-faraday
 library: faraday
-inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
 templateDir: modules/openapi-generator/src/main/resources/ruby-client
 additionalProperties:
   gemVersion: 1.0.0
```

But then running `./bin/generate-samples.sh bin/configs/ruby-faraday.yaml` fails with:
```
[main] WARN  o.o.c.languages.RubyClientCodegen - 200_response (model name starts with number) cannot be used as model name. Renamed to Model200Response
[main] WARN  o.o.c.languages.RubyClientCodegen - Return (reserved word) cannot be used as model name. Renamed to ModelReturn
[main] WARN  o.o.c.languages.RubyClientCodegen - 200_response (model name starts with number) cannot be used as model name. Renamed to Model200Response
[main] WARN  o.o.c.languages.RubyClientCodegen - Return (reserved word) cannot be used as model name. Renamed to ModelReturn
[main] WARN  o.o.c.languages.RubyClientCodegen - 200_response (model name starts with number) cannot be used as model name. Renamed to Model200Response
[main] INFO  o.o.codegen.DefaultGenerator - Model inline_object_3 (marked as unused due to form parameters) is generated due to the system property skipFormModel=false (default)
Exception in thread "main" java.lang.RuntimeException: Could not process model 'inline_object_3'.Please make sure that your schema is correct!
	at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:499)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:879)
Caused by: java.lang.ClassCastException: class java.time.OffsetDateTime cannot be cast to class java.lang.String (java.time.OffsetDateTime and java.lang.String are in module java.base of loader 'bootstrap')
	at org.openapitools.codegen.languages.AbstractRubyCodegen.toDefaultValue(AbstractRubyCodegen.java:150)
	at org.openapitools.codegen.DefaultCodegen.fromProperty(DefaultCodegen.java:3083)
	at org.openapitools.codegen.DefaultCodegen.addVars(DefaultCodegen.java:4853)
	at org.openapitools.codegen.DefaultCodegen.addVars(DefaultCodegen.java:4802)
	at org.openapitools.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:2550)
	at org.openapitools.codegen.DefaultGenerator.processModels(DefaultGenerator.java:1259)
	at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:494)
```

Any clue? The input file doesn't have any `inline_object_3`.

zippolyte:
The problem comes from this one: https://github.com/OpenAPITools/openapi-generator/blob/690be29f4b7d998fb58962b32fc0cac4bec8a363/modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml#L761-L761

The `toDefaultValue` function of the ruby codegen can't deal with dates it looks like.

wing328:
I'll try to come up with a fix this weekend.

Have a nice weekend everyone.

wing328:
I've filed https://github.com/OpenAPITools/openapi-generator/pull/7648 to fix the exception.

jfeltesse-mdsol:
confirmed the new configs work with the fix. I'll commit the updated samples last.

jfeltesse-mdsol:
@wing328 About the logic and ensuring the oneOf promise is fullfilled, how about verifying it _only when a discriminator is not set_? This is to conform to the openapi spec's explanation about discriminator: 
>  a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#discriminator-object

IMHO, if we always validate all the schemas then it makes the discriminator property pointless and undermines the work of the API spec author.

modules/openapi-generator/src/main/resources/ruby-client/model_test.mustache
```diff
@@ -43,6 +44,42 @@ describe '{{classname}}' do
   end
 
 {{/vars}}
+{{/oneOf}}
+{{#oneOf}}
```
zippolyte, line 48: probably missing a `{{#-first}}` here

jfeltesse-mdsol: Indeed. Thanks for spotting that!

jfeltesse-mdsol:
@wing328 So I had a look at validating the models that may be defined in the oneOf array so that when looping through those we can double check the oneOf promise. The take away is this is not as simple as it sounds and it should be done in a different PR.

Basically, a bare bones implementation may look like

```ruby
    {{^discriminator}}
      valid_models = []
      openapi_one_of.each do |_class|
        model = {{moduleName}}.const_get(_class).build_from_hash(attributes)
        valid_models << model if model.valid?
      end
      return valid_models[0] if valid_models.size == 1

      raise "{{classname}} is using oneOf but more than 1 schema validated #{attributes} => #{valid_models.inspect}"
    {{/discriminator}}
```

Looks good right? Except the `valid?` method does nothing most of the time...
The current code's validation only looks at required fields, min/max lengths, nullable, enums, patterns... but that's about it and if none of those are declared in the schema, it only ever returns `true`. Just with that we can't identify the models that may be breaking the oneOf promise, for that we'd need to actually do a proper JSON schema validation. A concrete example, with the code above, in the following spec all 3 schemas are considered valid if I pass say `{ "realtype" => "a", "code" => "b" }` to build_from_hash:

```json
    ObjA:
      type: object
      properties:
        realtype:
          type: string
        message:
          type: string
    ObjB:
      type: object
      properties:
        realtype:
          type: string
        code:
          type: integer
    ObjC:
      type: object
      properties:
        realtype:
          type: string
        code:
          type: string
    ObjAOrObjB:
      oneOf:
        - $ref: '#/components/schemas/ObjA'
        - $ref: '#/components/schemas/ObjB'
        - $ref: '#/components/schemas/ObjC'
```

In this case, that's because the `valid?` method doesn't check the types or the property names.

So yeah, not opening that pandora's box just now.

I'll try to complete the PR with support for scalar values though, stay tuned!

jfeltesse-mdsol:
Unless the CI reveals bug, I think this is feature complete. The most interesting bits are in [`d1c599d` (#5706)](https://github.com/OpenAPITools/openapi-generator/pull/5706/commits/d1c599d037535328a71f329bd73ab4f813818594#diff-1d8adf5df02e0bd3aae9bf9e7f2399b796a47760bb4381868d84d6d86ecb2e33).

Now oneOf supports scalars, arrays, hashes and the combinations you may think of. Maybe not perfect but should support most of the cases. And yeah, it doesn't try to check the oneOf promise. I got it to actually check whether the properties passed existed (see the changes in base_object, before it'd build_from_hash _without_ passing the data to the constructor, bypassing the property name checks...) but the values in base_object are cast without checking the type. If someone's pumped about fixing this, it should be the topic of a separate PR. I can't fix all the issues in one PR 😅

jfeltesse-mdsol:
@wing328 I tried to update the default ruby version that's incorrect here [`d1c599d` (#5706)](https://github.com/OpenAPITools/openapi-generator/pull/5706/commits/d1c599d037535328a71f329bd73ab4f813818594#diff-aa72b0bb63cdc14a2d17ebd641eb7b9f74f1d665552cf5f14ec3b7cbfe4aeea3R19) (by default the generator would use `>= 2.4`) but the ensure-up-to-date script reverts that change. How to update those docs?

wing328:
Travis CI reported the following errors:
```
An error occurred while loading spec_helper.
Failure/Error:
    class ChildCat < ParentPet
      attr_accessor :name
  
      attr_accessor :pet_type
  
      class EnumAttributeValidator
        attr_reader :datatype
        attr_reader :allowable_values
  
        def initialize(datatype, allowable_values)
NameError:
  uninitialized constant Petstore::ParentPet
# ./lib/petstore/models/child_cat.rb:17:in `<module:Petstore>'
# ./lib/petstore/models/child_cat.rb:16:in `<top (required)>'
# ./lib/petstore.rb:35:in `require'
# ./lib/petstore.rb:35:in `<top (required)>'
# ./spec/spec_helper.rb:14:in `require'
# ./spec/spec_helper.rb:14:in `<top (required)>'
No examples found.
Finished in 0.00003 seconds (files took 0.1754 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples

```

Ref: https://travis-ci.org/github/OpenAPITools/openapi-generator/builds/739218479

Please take a look when you've time.

wing328:
> but the ensure-up-to-date script reverts that change. How to update those docs?

CircleCI, which runs the ensure-up-to-date script, didn't report any failure so I assume it's no longer an issue, right?

jfeltesse-mdsol:
@wing328 I'll have a look at the Travis failure tomorrow. 

As for the Circle CI one, it's still a problem, I reverted the change so it passes, that's it. The doc [here](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/ruby.md) is still going to advertise the default ruby version for generated clients as ">= 1.9" when in fact it's ">= 2.4". How to update the doc and make the CI not change it back?

wing328:
> How to update the doc and make the CI not change it back?

One way is to run `./bin/utils/ensure-up-to-date` locally and commit the changes (in the doc).

jfeltesse-mdsol:
> One way is to run ./bin/utils/ensure-up-to-date locally and commit the changes (in the doc).

Not sure if I explain this correctly but locally the script will _always_ revert to the previous value.

I commit this:

```diff
commit ab351a70fb (HEAD -> ruby/fix_oneof)
Author: Julien Feltesse <jfeltesse@mdsol.com>
Date:   Wed Oct 28 15:34:55 2020 +0900

    use ruby 2.4 by default

diff --git a/docs/generators/ruby.md b/docs/generators/ruby.md
index 4c1cbca4e2..24b340eeec 100644
--- a/docs/generators/ruby.md
+++ b/docs/generators/ruby.md
@@ -16,7 +16,7 @@
 |gemHomepage|gem homepage. | |http://org.openapitools|
 |gemLicense|gem license. | |unlicense|
 |gemName|gem name (convention: underscore_case).| |openapi_client|
-|gemRequiredRubyVersion|gem required Ruby version. | |&gt;= 1.9|
+|gemRequiredRubyVersion|gem required Ruby version. | |&gt;= 2.4|
 |gemSummary|gem summary. | |A ruby wrapper for the REST APIs|
 |gemVersion|gem version.| |1.0.0|
 |hideGenerationTimestamp|Hides the generation timestamp when files are generated.| |true|
```

Run the script:
```
./bin/utils/ensure-up-to-date
```

And I get that diff:
```diff
diff --git a/docs/generators/ruby.md b/docs/generators/ruby.md
index 24b340eeec..4c1cbca4e2 100644
--- a/docs/generators/ruby.md
+++ b/docs/generators/ruby.md
@@ -16,7 +16,7 @@
 |gemHomepage|gem homepage. | |http://org.openapitools|
 |gemLicense|gem license. | |unlicense|
 |gemName|gem name (convention: underscore_case).| |openapi_client|
-|gemRequiredRubyVersion|gem required Ruby version. | |&gt;= 2.4|
+|gemRequiredRubyVersion|gem required Ruby version. | |&gt;= 1.9|
 |gemSummary|gem summary. | |A ruby wrapper for the REST APIs|
 |gemVersion|gem version.| |1.0.0|
 |hideGenerationTimestamp|Hides the generation timestamp when files are generated.| |true|
```


Where the heck the `1.9` value comes from I do not know. I even tried to modify the value in modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java, rebuild and re-run: same result. Giving up because I can't spend hours fighting that script.

wing328:
> Where the heck the 1.9 value comes from I do not know. I even tried to modify the value in modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RubyClientCodegen.java, rebuild and re-run: same result. Giving up because I can't spend hours fighting that script.

The doc is auto-generated. We'll take care of it in a separate PR instead.

jfeltesse-mdsol:
As mentioned in the last commit, I reverted to the old configs because the new ones would run into the issue described at https://github.com/OpenAPITools/openapi-generator/issues/4690
We actually have a fix for this one and we'll submit a PR soon.

modules/openapi-generator/src/main/resources/ruby-client/base_object.mustache
```diff
@@ -93,7 +95,7 @@
           is_nullable = self.class.openapi_nullable.include?(attr)
           next if !is_nullable || (is_nullable && !instance_variable_defined?(:"@#{attr}"))
         end
-        
+
```
zippolyte, line 98: Hmm I think you need to do the same check you did [here](https://github.com/OpenAPITools/openapi-generator/pull/5706/files?file-filters%5B%5D=.mustache#diff-488b370aef6f37384cf8676c79a431685e7f9ca1c6cd2796e4a746e9a0f7e91fR142-R144)  in `api_client.mustache` for `build` vs `build_from_hash` in the `_deserialize` method of `base_object` in case the oneOf is a property of a regular object.

jfeltesse-mdsol:
CI passes but please do not merge yet, thought more about my last fix and I think it's not going to work in all cases. Will post an update, probably tomorrow.

jfeltesse-mdsol:
fails on the .net things which I'm pretty sure aren't related to the PR here... 😑

jfeltesse-mdsol:
@wing328 This PR is ready to be merged but some CIs are failing randomly. I tried to force-push a couple of times but it's one CI failing once, another CI failing afterwards, and so on. On unrelated stuff.

zippolyte:
hey @wing328, any chance this can get merged in soon ? We're excited to start using this feature for generating our ruby client.

wing328:
We were busy with the v5.0.0-beta3 release: https://twitter.com/oas_generator/status/1329828788264321025. Let me try to review today or tomorrow.

Is it correct to say that you've tested the change locally and it works in your use cases?

zippolyte:
Yeah it works for the few use cases I tried

wing328:
@zippolyte thanks for testing it. I did a quick look and the PR looks good overall. I will merge this instead of being the bottleneck to get this PR merged. Will open an issue or file a PR if I find anything later.

zippolyte:
Just found a minor bug in the generated tests. They fail for oneOf containing primitive types because of the generated following code
```ruby
    it 'lists the models referenced in the oneOf array' do
      expect(described_class.openapi_one_of).to_not be_empty
      described_class.openapi_one_of.each { |klass| expect { DatadogAPIClient::V2.const_get(klass) }.to_not raise_error }
    end
```
Primitive types are not constants of the API client module.
I propose to just remove this test, WDYT ?

jfeltesse-mdsol:
ah, probably an oversight of mine when I refactored the code to accept scalar types and not just schemas. I think removing the 2nd line is ok since we wouldn't be able to make advanced tests in the generated code (lack of context). "Real" oneOf specs should be added in the samples but I didn't do that since the config need to be updated first.

Issue #5749, opened 2020-03-29
[BUG] [Erlang-Server] security definition context changes don't propagate to handler


ehansen31:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Erlang-Server generator doesn't apply context changes from security definitions to request handlers after validation is successful. In the spec provided there is an api-key validation that attempts to attach a field to the context to be passed into whatever request handler succeeds the call. The values are passed into the state of the open_api handler but not accessed when the request handler is called.

##### openapi-generator version

4.3.0

##### OpenAPI declaration file content or url

```yaml
"securityDefinitions": {
    "api_key": {
      "type": "apiKey",
      "name": "api_key",
      "in": "header"
    }
  },
```
```erlang
-spec authorize_api_key(OperationID ::
			    openapi_api:operation_id(),
			ApiKey :: binary()) -> {true, #{}}.
```
in the above setting anything in the map return value doesn't populate on a succeeding `handle_request/3` call.

##### Command line used for generation
compiled jar=>
swagger generate -i ./src/handler/swagger.json -g erlang-server -o ./apps/openapi

##### Steps to reproduce

1. add security definition to an existing endpoint
2. assign a value to the context map in the authorization handler.
3. attempt to access that value in the succeeding request handler.

##### Related issues/PRs
https://github.com/OpenAPITools/openapi-generator/pull/5751
created pr after filing issue, needs approved.

##### Suggest a fix
fix is map merging state context and params from openapi_api.erl. PR to follow shortly.

PR #5751, opened 2020-03-29
[Erlang-Server] security definition context changes don't propagate to handler

ehansen31:
erlang-server handler requires context and params to be merged before returned to user defined request_handler. Specifically if using a security definition and passing info into the context, it will not show up in the request handler succeeding the authorization method.

issue:
https://github.com/OpenAPITools/openapi-generator/issues/5749

### PR checklist

@tsloughter (2017/11) @jfacorro (2018/10) @robertoaloi (2018/10)


modules/openapi-generator/src/main/resources/erlang-server/handler.mustache
```diff
@@ -225,7 +225,7 @@ handle_request_json(
                 LogicHandler,
                 OperationID,
                 Req1,
-                Populated
+                maps:merge(State#state.context, Populated)
```
jfacorro: Just to confirm. Shouldn't this be merge in the reverse order in order to keep the request specific values?
```suggestion
                maps:merge(Populated, State#state.context)
```

ehansen31: https://erlang.org/doc/man/maps.html#merge-2

I was shooting for backwards compatibility, in such that a user is used to having only the parameters input into context. this change wouldn't overwrite any user input parameters and preserve any functionality where a user may accidentally overwrite with the addition of the context being passed from security definitions. So your suggestion would place user parameters in the context, then overwrite with the security definitions. Makes sense from a data transformation POV. I'll make the change later today.

wing328:
@ehansen31 thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

Issue #5781, opened 2020-04-01
[REQ] [typescript-angular] support for object query parameters as nested key/json string


dougal83:
### Is your feature request related to a problem? Please describe.

Cannot consume `. notation` query parameters objects in a current project.

## Describe the solution you'd like

Extend support to include option to configure how objects are encoded in query parameters.  To include options for `nested keys` and `JSON string`.

## Describe alternatives you've considered

Support  `. notation` in current project but that would not solve the issue for others.

## Additional context

Suggest change (as an option):

https://github.com/OpenAPITools/openapi-generator/blob/603709e17e318b22177f7dff193f65cf8527dbcf/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L181

```ts
key != null ? `${key}[${k}]` : k,
```

Also, add option to JSON stringify the object.

See: #4407

PR #5790, opened 2020-04-01
[typescript-angular] support for object query parameters as nested key/json string

dougal83:
fixes: #5781

Extend support to include an option to configure how objects are encoded in query parameters. To include options for nested keys and JSON string.

### PR checklist

@TiFu @taxpon  @sebastianhaas @kenisteward @Vrolijkx @macjohnny @nicokoenig @topce @akehir @petejohansonxo @amakhrov

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -27,6 +27,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.apache.commons.lang3.StringUtils.capitalize;
 import static org.openapitools.codegen.utils.StringUtils.*;
@@ -37,6 +39,8 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
     private static String CLASS_NAME_PREFIX_PATTERN = "^[a-zA-Z0-9]*$";
     private static String CLASS_NAME_SUFFIX_PATTERN = "^[a-zA-Z0-9]*$";
     private static String FILE_NAME_SUFFIX_PATTERN = "^[a-zA-Z0-9.-]*$";
+
+    public static enum QUERY_PARAM_OBJECT_FORMAT_TYPE {dot, json, key};
 
     private static final String DEFAULT_IMPORT_PREFIX = "./";
 
@@ -55,6 +59,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
     public static final String FILE_NAMING = "fileNaming";
     public static final String STRING_ENUMS = "stringEnums";
     public static final String STRING_ENUMS_DESC = "Generate string enums instead of objects for enum values.";
+    public static final String QUERY_PARAM_OBJECT_FORMAT = "queryParamObjectFormat";
 
     protected String ngVersion = "9.0.0";
     protected String npmRepository = null;
@@ -65,6 +70,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
     protected String modelFileSuffix = "";
     protected String fileNaming = "camelCase";
     protected Boolean stringEnums = false;
+    protected QUERY_PARAM_OBJECT_FORMAT_TYPE queryParamObjectFormat = QUERY_PARAM_OBJECT_FORMAT_TYPE.dot;
 
     private boolean taggedUnions = false;
 
@@ -107,6 +113,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         this.cliOptions.add(new CliOption(MODEL_FILE_SUFFIX, "The suffix of the file of the generated model (model<suffix>.ts)."));
         this.cliOptions.add(new CliOption(FILE_NAMING, "Naming convention for the output files: 'camelCase', 'kebab-case'.").defaultValue(this.fileNaming));
         this.cliOptions.add(new CliOption(STRING_ENUMS, STRING_ENUMS_DESC).defaultValue(String.valueOf(this.stringEnums)));
+        this.cliOptions.add(new CliOption(QUERY_PARAM_OBJECT_FORMAT, "The format for query param objects: 'dot', 'json', 'key'.").defaultValue(this.queryParamObjectFormat.name()));
     }
 
     @Override
@@ -223,6 +230,13 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
             this.setFileNaming(additionalProperties.get(FILE_NAMING).toString());
         }
 
+        if (additionalProperties.containsKey(QUERY_PARAM_OBJECT_FORMAT)) {
+            setQueryParamObjectFormat((String) additionalProperties.get(QUERY_PARAM_OBJECT_FORMAT));
+        }
+        additionalProperties.put("isQueryParamObjectFormatDot", getQueryParamObjectFormatDot());
+        additionalProperties.put("isQueryParamObjectFormatJson", getQueryParamObjectFormatJson());
+        additionalProperties.put("isQueryParamObjectFormatKey", getQueryParamObjectFormatKey());
+
     }
 
     private void addNpmPackageGeneration(SemVer ngVersion) {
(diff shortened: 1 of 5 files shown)


macjohnny:
thanks for your PR.

please re-generate the samples by running
```
mvn clean package
bin/typescript-angular-petstore-all.sh
```
and committing the changed samples.

dougal83:
> according to the CI (https://circleci.com/gh/OpenAPITools/openapi-generator/14068#tests/containers/2) the docs file needs to be changed

@macjohnny Thank you for the review.

I added the following line:
```md
|queryParamObjectFormat|Query parameter object value format: 'dot', 'json', 'key'.| |dot|
```
to [docs/generators/typescript-angular.md](https://github.com/OpenAPITools/openapi-generator/pull/5790/files#diff-1babc6e6a4ed2d3cde957d3495a6982bR34).  Is this the correct way to update the docs/web page for the parameters?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -55,6 +55,10 @@
     public static final String FILE_NAMING = "fileNaming";
     public static final String STRING_ENUMS = "stringEnums";
     public static final String STRING_ENUMS_DESC = "Generate string enums instead of objects for enum values.";
+    public static final String QUERY_PARAM_OBJECT_FORMAT = "queryParamObjectFormat";
+    public static final String QUERY_PARAM_OBJECT_FORMAT_DOT = "isQueryParamObjectFormatDot";
+    public static final String QUERY_PARAM_OBJECT_FORMAT_JSON = "isQueryParamObjectFormatJson";
+    public static final String QUERY_PARAM_OBJECT_FORMAT_KEY = "isQueryParamObjectFormatKey";
```
amakhrov, line 61: are these 3 constants used?

dougal83: Yes, open to an alternative if there is one for passing into mustache template:

https://github.com/OpenAPITools/openapi-generator/pull/5790/files#diff-f61af8a4c4eb2720bfb15259e0dca419R170-R187

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -360,6 +372,18 @@ public Boolean getStringEnums() {
         return stringEnums;
     }
 
+    public boolean getQueryParamObjectFormatDot() {
+        return queryParamObjectFormat != null && queryParamObjectFormat.equals("dot");
```
amakhrov, line 376: this can be simplified a bit but swapping operands:
```suggestion
        return dot.equals(queryParamObjectFormat);
```

however, switching from `string` to `enum` would eliminate the problem altogether

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
amakhrov: I mean the actual java constants: `QUERY_PARAM_OBJECT_FORMAT_KEY`, `QUERY_PARAM_OBJECT_FORMAT_JSON`, `QUERY_PARAM_OBJECT_FORMAT_DOT` - they are defined here, but I can't see any code that refers to them. Am I missing the point?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -65,6 +69,7 @@
     protected String modelFileSuffix = "";
     protected String fileNaming = "camelCase";
     protected Boolean stringEnums = false;
+    protected String queryParamObjectFormat = "dot";
```
amakhrov, line 72: Suggest using `enum` instead of hardcoded strings. See e.g. `MODEL_PROPERTY_NAMING_TYPE`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
dougal83: I think it's that I don't understand Java, I was just muddling through.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
dougal83: Would the enum constant go with `MODEL_PROPERTY_NAMING_TYPE` in the same global file or be held locally?  (I'll do the later for now)

amakhrov: I believe it should be defined locally, as it's very specific to the angular generator.

dougal83: Thanks.  One more query if you don't mind!  If `this.queryParamObjectFormat` is an enum then how do we approach the following?  It was a string but with enum we need a bit more, or shall I just omit the CLI option?

```
this.cliOptions.add(new CliOption( // ... )
.defaultValue(this.queryParamObjectFormat));
```

amakhrov: `.defaultValue(this.queryParamObjectFormat.name())`
you could refer to how `MODEL_PROPERTY_NAMING` and `ENUM_PROPERTY_NAMING` are used inside `AbstractTypeScriptClientCodegen`

dougal83: @amakhrov Thank you for the guidance.  Applied feedback.

amakhrov:
@dougal83 
After reviewing the spec, I also realized that OAS3 has a standard way to define the format of nested objects:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#style-values (see `deepObject` style)

However, it's not really supported by any of the existing generators. Only php generators do smth like that, but they do not respect the `style=deepObject` setting anyway.

I'm really curious if there is any "standard-ish" approach implemented by the majority of other generators for object params serialization that we can consider a de-facto standard?
@TiFu @macjohnny any insights on that?

dougal83:
> @dougal83
> After reviewing the spec, I also realized that OAS3 has a standard way to define the format of nested objects:
> https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#style-values (see `deepObject` style)

I'm interested in using openapi-generator with loopback, there is an issue over there [regarding supporting deep objects](https://github.com/OAI/OpenAPI-Specification/issues/1706).  Am interested in helping if I can.

amakhrov:
Looks like json serialization is supported by the spec, as mentioned in this comment: https://github.com/OAI/OpenAPI-Specification/issues/1706#issuecomment-523418975

should we support this in the typescript generator instead of a custom generator cli option?

dougal83:
> Looks like json serialization is supported by the spec, as mentioned in this comment: [OAI/OpenAPI-Specification#1706 (comment)](https://github.com/OAI/OpenAPI-Specification/issues/1706#issuecomment-523418975)
> 
> should we support this in the typescript generator instead of a custom generator cli option?

I'll leave this up to the tech. committee to decide.  Feel free to do as you see fit with this PR depending on the outcome of the discussion. 👍

dougal83:
> Looks like json serialization is supported by the spec, as mentioned in this comment: [OAI/OpenAPI-Specification#1706 (comment)](https://github.com/OAI/OpenAPI-Specification/issues/1706#issuecomment-523418975)
> 
> should we support this in the typescript generator instead of a custom generator cli option?

Perhaps we should close this PR in the light of missing the latest release and focus on the Typescript avenue for improvement.

macjohnny:
I will merge this once https://github.com/OpenAPITools/openapi-generator/pull/6360 is merged, since there will be conflicts

mrbatista:
@macjohnny any update?

macjohnny:
@dougal83 could you please merge the latest master into your branch, resolve the conflicts and re-generated the samples?

dougal83:
Performed simple rebase and resolved conflict.  Untested and samples not regenerated(don't think they are affected).  I need to install JDK & Maven... try to do shortly.

dougal83:
@macjohnny Tests pass. Sample regeneration run. 👍

dougal83:
Removed the last commit with sample regen as nothing material was added.  I hope this is ready to go after checks...

PR #5791, opened 2020-04-01
[Python-experimental] Documentation enhancement for oneOf schema and minor err msg improvement

sebastien-rosset:
* Improve documentation.
* Improve error message.

### PR checklist

### DIFF
--- a/samples/client/petstore/python-experimental/petstore_api/model_utils.py
+++ b/samples/client/petstore/python-experimental/petstore_api/model_utils.py
@@ -1201,13 +1201,14 @@ def get_allof_instances(self, model_args, constant_args):
             composed_instances.append(allof_instance)
         except Exception as ex:
             raise ApiValueError(
-                "Invalid inputs given to generate an instance of %s. The "
-                "input data was invalid for the allOf schema % in the composed "
-                "schema %s. Error=%s" %
-                allof_class.__class__.__name__,
-                allof_class.__class__.__name__,
-                self.__class__.__name__,
-                str(ex)
+                "Invalid inputs given to generate an instance of '%s'. The "
+                "input data was invalid for the allOf schema '%s' in the composed "
+                "schema '%s'. Error=%s" % (
+                    allof_class.__class__.__name__,
+                    allof_class.__class__.__name__,
+                    self.__class__.__name__,
+                    str(ex)
+                )
             )
     return composed_instances
 
--- a/samples/client/petstore/python-experimental/petstore_api/model_utils.py
+++ b/samples/client/petstore/python-experimental/petstore_api/model_utils.py
@@ -1371,10 +1371,18 @@ def get_unused_args(self, composed_instances, model_args):
 
 def validate_get_composed_info(constant_args, model_args, self):
     """
-    For composed schemas/classes, validates the classes to make sure that
-    they do not share any of the same parameters. If there is no collision
-    then composed model instances are created and returned to the calling
-    self model
+    For composed schemas, generate schema instances for
+    all schemas in the oneOf/anyOf/allOf definition. If additional
+    properties are allowed, also assign those properties on
+    all matched schemas that contain additionalProperties.
+    Openapi schemas are python classes.
+
+    Exceptions are raised if:
+    - no oneOf schema matches the model_args input data
+    - > 1 oneOf schema matches the model_args input data
+    - > 1 oneOf schema matches the model_args input data
+    - no anyOf schema matches the model_args input data
+    - any of the allOf schemas do not match the model_args input data
 
     Args:
         constant_args (dict): these are the args that every model requires
--- a/samples/client/petstore/python-experimental/petstore_api/model_utils.py
+++ b/samples/client/petstore/python-experimental/petstore_api/model_utils.py
@@ -1196,8 +1196,19 @@ def get_allof_instances(self, model_args, constant_args):
 
         # and use it to make the instance
         kwargs.update(constant_args)
-        allof_instance = allof_class(**kwargs)
-        composed_instances.append(allof_instance)
+        try:
+            allof_instance = allof_class(**kwargs)
+            composed_instances.append(allof_instance)
+        except Exception as ex:
+            raise ApiValueError(
+                "Invalid inputs given to generate an instance of %s. The "
+                "input data was invalid for the allOf schema % in the composed "
+                "schema %s. Error=%s" %
+                allof_class.__class__.__name__,
+                allof_class.__class__.__name__,
+                self.__class__.__name__,
+                str(ex)
+            )
     return composed_instances
 
 
@@ -1308,8 +1319,8 @@ def get_anyof_instances(self, model_args, constant_args):
             pass
     if len(anyof_instances) == 0:
         raise ApiValueError(
-            "Invalid inputs given to generate an instance of %s. Unable to "
-            "make any instances of the classes in anyOf definition." %
+            "Invalid inputs given to generate an instance of %s. None of the "
+            "anyOf schemas matched the inputs." %
             self.__class__.__name__
         )
     return anyof_instances
--- a/samples/client/petstore/python-experimental/petstore_api/model_utils.py
+++ b/samples/client/petstore/python-experimental/petstore_api/model_utils.py
@@ -720,7 +720,11 @@ def get_required_type_classes(required_types_mixed):
 def change_keys_js_to_python(input_dict, model_class):
     """
     Converts from javascript_key keys in the input_dict to python_keys in
-    the output dict using the mapping in model_class
+    the output dict using the mapping in model_class.
+    If the input_dict contains a key which does not declared in the model_class,
+    the key is added to the output dict as is. The assumption is the model_class
+    may have undeclared properties (additionalProperties attribute in the OAS
+    document).
     """
 
     output_dict = {}
@@ -1199,12 +1203,20 @@ def get_allof_instances(self, model_args, constant_args):
 
 def get_oneof_instance(self, model_args, constant_args):
     """
+    Find the oneOf schema that matches the input data (e.g. payload).
+    If exactly one schema matches the input data, an instance of that schema
+    is returned.
+    If zero or more than one schema match the input data, an exception is raised.
+    In OAS 3.x, the payload MUST, by validation, match exactly one of the
+    schemas described by oneOf.
     Args:
         self: the class we are handling
         model_args (dict): var_name to var_value
-            used to make instances
+            The input data, e.g. the payload that must match a oneOf schema
+            in the OpenAPI document.
         constant_args (dict): var_name to var_value
-            used to make instances
+            args that every model requires, including configuration, server
+            and path to item.
 
     Returns
         oneof_instance (instance/None)
@@ -1213,12 +1225,17 @@ def get_oneof_instance(self, model_args, constant_args):
         return None
 
     oneof_instances = []
+    # Iterate over each oneOf schema and determine if the input data
+    # matches the oneOf schemas.
     for oneof_class in self._composed_schemas()['oneOf']:
-        # transform js keys to python keys in fixed_model_args
+        # transform js keys from input data to python keys in fixed_model_args
         fixed_model_args = change_keys_js_to_python(
             model_args, oneof_class)
 
-        # extract a dict of only required keys from fixed_model_args
+        # Extract a dict with the properties that are declared in the oneOf schema.
+        # Undeclared properties (e.g. properties that are allowed because of the
+        # additionalProperties attribute in the OAS document) are not added to
+        # the dict.
         kwargs = {}
         var_names = set(oneof_class.openapi_types().keys())
         for var_name in var_names:
@@ -1238,14 +1255,14 @@ def get_oneof_instance(self, model_args, constant_args):
             pass
     if len(oneof_instances) == 0:
         raise ApiValueError(
-            "Invalid inputs given to generate an instance of %s. Unable to "
-            "make any instances of the classes in oneOf definition." %
+            "Invalid inputs given to generate an instance of %s. None "
+            "of the oneOf schemas matched the input data." %
             self.__class__.__name__
         )
     elif len(oneof_instances) > 1:
         raise ApiValueError(
             "Invalid inputs given to generate an instance of %s. Multiple "
-            "oneOf instances were generated when a max of one is allowed." %
+            "oneOf schemas matched the inputs, but a max of one is allowed." %
             self.__class__.__name__
         )
     return oneof_instances[0]
@@ -1345,7 +1362,7 @@ def validate_get_composed_info(constant_args, model_args, self):
     """
     For composed schemas/classes, validates the classes to make sure that
     they do not share any of the same parameters. If there is no collision
-    then composed model instances are created and returned tot the calling
+    then composed model instances are created and returned to the calling
     self model
 
     Args:
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
@@ -943,10 +943,16 @@ def get_allof_instances(self, model_args, constant_args):
 
 def get_oneof_instance(self, model_args, constant_args):
     """
+    Find the oneOf schema that matches the input data (e.g. payload).
+    If exactly one schema matches the input data, an instance of that schema
+    is returned.
+    If zero or more than one schema match the input data, an exception is raised.
+    In OAS 3.x, the payload MUST, by validation, match exactly one of the
+    schemas described by oneOf.
     Args:
         self: the class we are handling
         model_args (dict): var_name to var_value
-            used to make instances
+            the input data, e.g. the payload  used to make instances
         constant_args (dict): var_name to var_value
             used to make instances
 
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
@@ -965,8 +965,10 @@ def get_oneof_instance(self, model_args, constant_args):
         return None
 
     oneof_instances = []
+    # Iterate over each oneOf schema and determine if the input data
+    # matches the oneOf schemas.
     for oneof_class in self._composed_schemas()['oneOf']:
-        # transform js keys to python keys in fixed_model_args
+        # transform js keys from input data to python keys in fixed_model_args
         fixed_model_args = change_keys_js_to_python(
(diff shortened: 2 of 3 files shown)


sebastien-rosset:
From the OpenAPI specification:

> The discriminator object is legal only when using one of the composite keywords oneOf, anyOf, allOf.
In OAS 3.0, a response payload MAY be described to be exactly one of any number of types
> ...
> which means the payload MUST, by validation, match exactly one of the schemas described by Cat, Dog, or Lizard. In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. 
> ...
> The expectation now is that a property with name petType MUST be present in the response payload, and the value will correspond to the name of a schema defined in the OAS document.

spacether:
Here are my thoughts here. We have three use cases:
1. self schema with child schemas that allOf inherrit from self. self has a discriminator that maps to the children
  - when deserializing self and receiving a child do not validate self, instantiate the child. self's validation is included because the child instances has an instance of the parent in child.composed_instances. That means that parent's validations are run
2. self composed schema with only oneOf schemas OR self composed schema with only anyOf schemas
  - do not validate self because self contains no extra validation info, instead return the child instance
3. self composed schema with allOf AND (oneOf or anyOf defined)
  - do validate self because the oneOf or anyOf models lack the validation in self
  - if that validation passes try to make a oneOf or anyOf model with all of the inputs used for self. That may succeed or fail because in later versions of the openapi spec additionalProperties: {} is set by default
  - if we succeed in making that oneOf/anyOf instance using all of self's inputs then return it
  - if we fail then return self
    - Note: self contains a oneOf instance that uses a subset of self's input arguments

Our python-experimental code currently handles cases 1 + 2 but I don't think it handles case 3 yet

sebastien-rosset:
> 1. self composed schema with only oneOf schemas OR self composed schema with only anyOf schemas
> 
> * do not validate self because self contains no extra validation info, instead return the child instance

I think this is the example I provided. Currently, it seems the existing **get_oneof_instance** function does fail the validation (depending on the exact scenario). Are you suggesting we should change the validation logic? Or are you stating how it is supposed to be implemented today?

sebastien-rosset:
Ah, I think if I add a specific enum value for the discriminator in each of the "subclasses", this would fix the problem as a side effect, because the construction of the instance would fail. Let me try that.

spacether:
Can you provide examples of the data you are ingesting and the schema?
In cases 1 + 2 I am describing how our current code works which is how I think that it shoudl work.
For cases 3 and 4 I am describing how I think it should work, that code has not been written yet.
- That code addition would be small and it may make sense to add it in this PR

spacether:
>     organization.Organization.Relationship:
      oneOf:
      - type: 'null'
      - $ref: '#/components/schemas/mo.MoRef'
      - $ref: '#/components/schemas/organization.Organization'

For python-experimental this may not work. The code assumes that oneOf and anyOf are $refs to schemas not null.

spacether:
> As an example, suppose we have the following OAS document with schemas A, B, C. In the schema, none of these schemas have any required properties, except for the discriminator property. Also, the additionalProperties attribute is not specified, so by default B and C allow undeclared properties. When the payload is received, it seems the get_oneof_instance will always match the input payload to B and C because the only required property is common to both B and C (the discriminator property); and both B and C have additionalProperties: true (by default). Hence this would cause get_oneof_instance to always raise an exception.

That's right. The python code assumes that child oneOf/anyOf classes must have at least one property. If that property is only the discriminator then multiple schemas will match which is not allowed in oneOf.
If you need that to work, then you should use anyOf. But why have more than one schema with only a discriminator + additionalProperties?

sebastien-rosset:
> Can you provide examples of the data you are ingesting and the schema?
> In cases 1 + 2 I am describing how our current code works which is how I think that it shoudl work.
> For cases 3 and 4 I am describing how I think it should work, that code has not been written yet.
> 
> * That code addition would be small and it may make sense to add it in this PR

It is at the beginning of this message. I tried to summarize the relevant part of the OAS document, because the entire document is really big. 

I also just added one more comment which I think may be the crux of the issue. For each leaf class, there is only one possible value for the discriminator property. But the discriminator is defined in a parent class.  In the python generated __init__ function, there does not appear to be a check for the value of the discriminator.

sebastien-rosset:
> > As an example, suppose we have the following OAS document with schemas A, B, C. In the schema, none of these schemas have any required properties, except for the discriminator property. Also, the additionalProperties attribute is not specified, so by default B and C allow undeclared properties. When the payload is received, it seems the get_oneof_instance will always match the input payload to B and C because the only required property is common to both B and C (the discriminator property); and both B and C have additionalProperties: true (by default). Hence this would cause get_oneof_instance to always raise an exception.
> 
> That's right. The python code assumes that child oneOf/anyOf classes must have at least one property. If that property is only the discriminator then multiple schemas will match which is not allowed in oneOf.
> If you need that to work, then you should use anyOf. But why have more than one schema with only a discriminator + additionalProperties?

The schemas do have declared properties, but these properties are not specified as **required** in the OAS document. Hence the __init__ constructor ends up accepting all of these unknown properties as additional, undeclared properties, and no exception is raised.

sebastien-rosset:
As a variant, suppose we have the following spec:

```yaml
    A:
      oneOf:
      - $ref: '#/components/schemas/B'
      - $ref: '#/components/schemas/C'
    Base:
       type: object
        discriminator:
           propertyName: Class
        required:
        - Class
        properties:
           Class:
             type: string
    B:
        allOf:
        - $ref: '#/components/schemas/Base'
        - type: object
           properties:
              X:
    C:
        allOf:
        - $ref: '#/components/schemas/Base'
        - type: object
           properties:
              Y:
```
For the B class, I expect the following payload:
```
{ "Class": "B", "X": "abc" }
```

For the C class, I expect the following payload:
```
{ "Class": "C", "X": "abc" }
```

But I didn't explicitly specified the the value of B.Class must be "B" anywhere in the OAS doc. So it could be changed as follows:

```yaml
    B:
        allOf:
        - $ref: '#/components/schemas/Base'
        - type: object
           properties:
              Class:
                 type: string
                 enum:
                 - B
              X:
```

But I'm worried about the problems it will bring because that means the same property will be present twice in the same OAS schema, and that typically causes problems for the code generator.

I'm also wondering if it's wise to implement the strictest possible validation during unmarshaling. I think when OASTool is invoked with "validate --recommend", it would be good to be as strict as possible. But during unmarshaling, if there is no room for leniency, in practice there could be too many failures, especially when there are ambiguities in the OAS specification. If the API server is not perfectly compliant with your interpretation of the spec, then users will not be able to use the generated SDK. In my case, I think I can adjust the OAS spec, but I can see other people just giving up because the tool keeps being too strict. Another problem is if you have a tool that allows importing a OAS document from 3rd parties, you are not in control of the OAS doc.

sebastien-rosset:
It's arguable whether the current python-experimental validation implements this part of the OAS spec:

> a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.

spacether:
> > > As an example, suppose we have the following OAS document with schemas A, B, C. In the schema, none of these schemas have any required properties, except for the discriminator property. Also, the additionalProperties attribute is not specified, so by default B and C allow undeclared properties. When the payload is received, it seems the get_oneof_instance will always match the input payload to B and C because the only required property is common to both B and C (the discriminator property); and both B and C have additionalProperties: true (by default). Hence this would cause get_oneof_instance to always raise an exception.
> > 
> > 
> > That's right. The python code assumes that child oneOf/anyOf classes must have at least one property. If that property is only the discriminator then multiple schemas will match which is not allowed in oneOf.
> > If you need that to work, then you should use anyOf. But why have more than one schema with only a discriminator + additionalProperties?
> 
> The schemas do have declared properties, but these properties are not specified as **required** in the OAS document. Hence the **init** constructor ends up accepting all of these unknown properties as additional, undeclared properties, and no exception is raised.

That's not what's happening. If those properties are optional, then they exist in `oneof_class.openapi_types` The function get_oneof_instance looks through all of the oneOf classes and uses all of the properties defined in the oneOf class's openapi_types as inputs when making an instance of that class/schema. It does not depend upon whether or not the arguments are required or optional.

spacether:
> It's arguable whether the current python-experimental validation implements this part of the OAS spec:
> 
> > a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.

What information makes you think that this is arguable?
Use case 1 and 2 skip creation of the parent schema instance and instead return the allOf/oneOf/anyOf child. That is shortcutting the validation in the parent.
You can see this code [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L595)

sebastien-rosset:
What do you think about adding the following logic?

Given a schema (CodegenModel) with name “X”, get the associated CodegenDiscriminator. From there, iterate through all the mappings (both explicit and inherited) and get the subset of mappings that refer to schema “X”. I think in most cases it will be reduced to a set of size 1 (but not necessarily, especially with explicit mappings).

The subset of discriminator mappings could be used to validate a model object when it is constructed. For example, in the python-experimental  __init__ function of the model, this could be used to validate the value of the discriminator.

spacether:
> What do you think about adding the following logic?
> 
> Given a schema (CodegenModel) with name “X”, get the associated CodegenDiscriminator. From there, iterate through all the mappings (both explicit and inherited) and get the subset of mappings that refer to schema “X”. I think in most cases it will be reduced to a set of size 1 (but not necessarily, especially with explicit mappings).
> 
> The subset of discriminator mappings could be used to validate a model object when it is constructed. For example, in the python-experimental **init** function of the model, this could be used to validate the value of the discriminator.

Your proposal to iterate through all the mappings is exactly what we do here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L967
Have you read my comments above?
Why do we need to verify the value of the discriminator? We already chose it from the mapping at the parent level? If you want, you can add that leaf schema validation by defining the discriminator as having only one value option.

sebastien-rosset:
> > The schemas do have declared properties, but these properties are not specified as **required** in the OAS document. Hence the **init** constructor ends up accepting all of these unknown properties as additional, undeclared properties, and no exception is raised.
> 
> That's not what's happening. If those properties are optional, then they exist in `oneof_class.openapi_types` The function get_oneof_instance looks through all of the oneOf classes and uses all of the properties defined in the oneOf class's openapi_types as inputs when making an instance of that class/schema. It does not depend upon whether or not the arguments are required or optional.

Let me try to clarify. Suppose the kwargs argument built [here](https://github.com/OpenAPITools/openapi-generator/pull/5791/files#diff-cc72ac40056827e3a103f9b4f56059cfR983) has keys for properties p1, p2, p3, and none of these properties are declared in the schema. There [here](https://github.com/OpenAPITools/openapi-generator/blob/f56e18a8a8dc8289101757e93af35b61f2a2d730/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L996) the oneOf schema will be built without any exception.

spacether:
> As a variant, suppose we have the following spec:
> 
> ```yaml
>     A:
>       oneOf:
>       - $ref: '#/components/schemas/B'
>       - $ref: '#/components/schemas/C'
>     Base:
>        type: object
>         discriminator:
>            propertyName: Class
>         required:
>         - Class
>         properties:
>            Class:
>              type: string
>     B:
>         allOf:
>         - $ref: '#/components/schemas/Base'
>         - type: object
>            properties:
>               X:
>     C:
>         allOf:
>         - $ref: '#/components/schemas/Base'
>         - type: object
>            properties:
>               Y:
> ```
> 
> For the B class, I expect the following payload:
> 
> ```
> { "Class": "B", "X": "abc" }
> ```
> 
> For the C class, I expect the following payload:
> 
> ```
> { "Class": "C", "X": "abc" }
> ```
> 
> But I didn't explicitly specified the the value of B.Class must be "B" anywhere in the OAS doc. So it could be changed as follows:
> 
> ```yaml
>     B:
>         allOf:
>         - $ref: '#/components/schemas/Base'
>         - type: object
>            properties:
>               Class:
>                  type: string
>                  enum:
>                  - B
>               X:
> ```
> 
> In that case, I'm worried about the problems it will bring because that means the same property will be present twice in the same OAS schema, and that typically causes problems for the code generator.
> 
> I'm also wondering if it's wise to implement the strictest possible validation during unmarshaling. I think when OASTool is invoked with "validate --recommend", it would be good to be as strict as possible. But during unmarshaling, if there is no room for leniency, in practice there could be too many failures, especially when there are ambiguities in the OAS specification. If the API server is not perfectly compliant with your interpretation of the spec, then users will not be able to use the generated SDK. In my case, I think I can adjust the OAS spec, but I can see other people just giving up because the tool keeps being too strict. Another problem is if you have a tool that allows importing a OAS document from 3rd parties, you are not in control of the OAS doc.
> 
> A good example is postman. Postman support OAS v2 and v3 to some extent, and the tool tends to be lenient. That's because they know OAS documents are not perfect, and in fact it's tough to write a doc which is perfectly compliant since different people disagree on the meaning of the spec.

Sending { "Class": "C", "X": "abc" }
would not work because class C lacks the property X.

python-experimental allows the same properties to be present in multiple composed schemas. This is good because Openapi allows additive declaration of properties. We can define a property height and set a max on it it in one class E and a min on it in class F and then have a composed schema allOf inherit E + F. This is good. Many other generators do not support this but they should.

What leniency do you think that we should add? What would you like to allow?
The flag that you added about additionalProperties should allow a lot more leniency.

sebastien-rosset:
> > What do you think about adding the following logic?
> > Given a schema (CodegenModel) with name “X”, get the associated CodegenDiscriminator. From there, iterate through all the mappings (both explicit and inherited) and get the subset of mappings that refer to schema “X”. I think in most cases it will be reduced to a set of size 1 (but not necessarily, especially with explicit mappings).
> > The subset of discriminator mappings could be used to validate a model object when it is constructed. For example, in the python-experimental **init** function of the model, this could be used to validate the value of the discriminator.
> 
> Your proposal to iterate through all the mappings is exactly what we do here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L967
> Have you read my comments above?
> Why do we need to verify the value of the discriminator? We already chose it from the mapping at the parent level? If you want, you can add that leaf schema validation by defining the discriminator as having only one value option.

Yes I did read. 

> > What do you think about adding the following logic?
> > Given a schema (CodegenModel) with name “X”, get the associated CodegenDiscriminator. From there, iterate through all the mappings (both explicit and inherited) and get the subset of mappings that refer to schema “X”. I think in most cases it will be reduced to a set of size 1 (but not necessarily, especially with explicit mappings).
> > The subset of discriminator mappings could be used to validate a model object when it is constructed. For example, in the python-experimental **init** function of the model, this could be used to validate the value of the discriminator.
> 
> Your proposal to iterate through all the mappings is exactly what we do here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L967

The above line returns a list of all the properties specified in the model, it's not the same thing as what I was proposing.

> Have you read my comments above?

Yes.

> Why do we need to verify the value of the discriminator? We already chose it from the mapping at the parent level? If you want, you can add that leaf schema validation by defining the discriminator as having only one value option.

Well at this point I am trying to understand why the generated python-experimental code fails unmarshaling (an exception is raised), but it works for other languages or tools with the same OAS document. 

The payload I receive is as below, with  **_class** being the discriminator property.
```
{'_class': 'mo.MoRef',
'object_type': 'organization.Organization',
'moid': '5de82cea6972652d30b307bd',
'link': 'https://..../organization/Organizations/5de82cea6972652d30b307bd'
}
```

And that payload successfully unmarshals into the organization.Organization model, even though the discriminator value for the organization.Organization schema is supposed to be "organization.Organization", not "mo.MoRef". I was expecting that unmarshaling that payload into the the organization.Organization model would fail.

spacether:
I look forward to hearing why the payload is not deserializing into the child model correctly. We could add code to check that the discriminator propertyName value is the same as the model's class name. However, we have use cases where in the v3 spec users will supply mappings to go from small_cat to the class Cat. If they have another mapping to go from big_cat to Cat, then we can't do this check inside the Cat class. How do you propose addressing this use case and still checking the value of the discriminator propertyName?
Why do we not trust the propertyName value to Class mappings in the .discriminator() dict?

samples/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -1345,7 +1362,7 @@ def validate_get_composed_info(constant_args, model_args, self):
     """
     For composed schemas/classes, validates the classes to make sure that
     they do not share any of the same parameters. If there is no collision
-    then composed model instances are created and returned tot the calling
+    then composed model instances are created and returned to the calling
     self model
```
spacether, line 1366: This comment is not accurate. We do allow classes to use the same parameters. Can you update this to:
```
    For composed schemas, generate schema instances for
    all schemas in the oneOf/anyOf/allOf definition. If additional
    properties are allowed, also assign those properties on
    all matched schemas that contain additionalProperties.
    Openapi schemas are python classes.

    Exceptions are raised if:
    - no oneOf schema matches the model_args input data
    - > 1 oneOf schema matches the model_args input data
    - > 1 oneOf schema matches the model_args input data
    - no anyOf schema matches the model_args input data
    - any of the allOf schemas do not match the model_args input data
```

sebastien-rosset:
> This looks good. I have some small changes that I'd like here too:
> 
> * Please add your improved schema message to the exception in get_anyof_instances.

Done

> * Can you add the below `in the composed schema %s` suffix to all of these error messages so the user knows where this is coming from?

I think it only makes sense in the allOf case. In the other cases, it already prints class name. BTW, what would really help is a printout of the relevant section of the OAS doc

> * Can you also add something like this exception in get_allof_instances to be consistent?
> 
> ```
>         try:
>             allof_instance = allof_class(**kwargs)
>         except Exception as ex:
>             raise ApiValueError(
>                 "Invalid inputs given to generate an instance of %s. The "
>                 "input data was invalid for the allOf schema % in the composed "
>                 "schema %s. Error=%s" %
>                 allof_class.__class__.__name__,
>                 allof_class.__class__.__name__,
>                 self.__class__.__name__,
>                 str(ex)
>             )
> ```

done

samples/client/petstore/python-experimental/petstore_api/model_utils.py
sebastien-rosset: done

PR #5798, opened 2020-04-02
[codegen] Helper function for discriminator mapping allowed values

sebastien-rosset:
Add a helper function that returns a list of allowed discriminator mapping values for a schema object.

The list of all possible schema discriminator mapping values is obtained from explicit discriminator mapping values in the OpenAPI document, and from inherited discriminators through oneOf, allOf, anyOf.

For example, a discriminator may be defined in a 'Pet' schema as shown below.
The Dog and Cat schemas inherit the discriminator through the allOf reference.
In the 'Pet' schema, the supported discriminator mapping values for the 'objectType' properties are 'Dog' and 'Cat'.
     * The allowed discriminator mapping value for the Dog schema is 'Dog'.
     * The allowed discriminator mapping value for the Cat schema is 'Dog'.

Currently, I don't see any helper function that would return a list of allowed discriminator mapping values, such as "Dog" for the Dog schema.

```yaml 
Pet:
   type: object
   discriminator:
     propertyName: objectType
   required:
     - objectType
   properties:
     objectType:
     type: string
 Dog:
   allOf:
   - type: object
     properties:
       p1:
        type: string
 Cat:
   allOf:
   - type: object
     properties:
       p2:
         type: string
```

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenDiscriminator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenDiscriminator.java
@@ -11,6 +11,7 @@ public class CodegenDiscriminator {
     private String propertyName;
     private String propertyBaseName;
     private String propertyGetter;
+    // The implementation type for the language generator being executed.
     private String propertyType;
     private Map<String, String> mapping;
     private Set<MappedModel> mappedModels = new LinkedHashSet<>();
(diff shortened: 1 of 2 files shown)


sebastien-rosset:
@spacether , @wing328

spacether:
@sebastien-rosset per our offline discussion that list of validated discriminator mapping values exist in the discriminator mapping. We should not need to verify them because the java code already verified them before loading them in to the codegendiscriminator mapping. If a payload is given a prooertyName value that exists in that map then checking presence in the map's keys is verification.

PR #5809, opened 2020-04-02
[codegen][Python-experimental] Discriminator NPE fix, handle 'null' type, #4906 enhancements

### LINKED ISSUE #4912: [Python] Multilevel inheritance discriminator support

sebastien-rosset:
1. When processing discriminators of composed schemas, handle scenario when composed schema has 'null' type
1. Remove extraneous characters in comments
1. Add traversal of discriminator mappings in case oneOf child schema indirectly refers to discriminator.
1. There is no need to duplicate the `get_discriminator_class()` code for every generated class.
1. Add unit test when discriminator is not specified in payload and handle that scenario in the `get_discriminator_class()` function. IMO, another PR should be provided to provide a better error message when the input data does not contain the discriminator property.

This PR along these:
- [Add discriminator memoization + move discriminator logic to the new method and allow deserialization through 2+ discriminators](https://github.com/OpenAPITools/openapi-generator/pull/6124)
- [Improve composed schema discriminator map](https://github.com/OpenAPITools/openapi-generator/pull/4906)

will resolve #4912

### PR checklist


spacether:
@sebastien-rosset your 3rd point Fix NPE when at least one extension is defined but not x-discriminator-value is already addressed in my PR which will land before this one, right?

sebastien-rosset:
> @sebastien-rosset your 3rd point Fix NPE when at least one extension is defined but not x-discriminator-value is already addressed in my PR which will land before this one, right?

I found one more case where a NPE can be raised. There are two locations in your PR where you do a lookup of the extension.

spacether:
> > @sebastien-rosset your 3rd point Fix NPE when at least one extension is defined but not x-discriminator-value is already addressed in my PR which will land before this one, right?
> 
> I found one more case where a NPE can be raised. There are two locations in your PR where you do a lookup of the extension.

Right, but both of those locations now check that the x-discriminator-value key exists before getting it in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
- line 2584 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
- line 2639 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
Doesn't that eliminate the NPE?

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
```diff
@@ -203,6 +211,63 @@ public void setDiscriminator(CodegenDiscriminator discriminator) {
...
+     */
+    public List<String> getAllowedDiscriminatorMappingValues() {
+        List<String> mappingValues = new ArrayList<String>();
+        for (CodegenDiscriminator.MappedModel mm : discriminator.getMappedModels()) {
+            if (name.equals(mm.getModelName())) {
+                mappingValues.add(mm.getMappingName());
+            }
+        }     
+        return mappingValues;   
+    }
```
spacether, line 261: Where are we using this? I can't find a call site.
If we aren't using it, why include it?
Our users can already iterate through discriminator MappedModel.
If we want to provide an easier way to do this, why not put it in the CodegenDiscriminator class instead?

sebastien-rosset:
> > > @sebastien-rosset your 3rd point Fix NPE when at least one extension is defined but not x-discriminator-value is already addressed in my PR which will land before this one, right?
> > 
> > 
> > I found one more case where a NPE can be raised. There are two locations in your PR where you do a lookup of the extension.
> 
> Right, but both of those locations the code now checks that the x-discriminator-value key exists before getting it in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
> 
> * line 2584 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
> * line 2639 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
> 
> Doesn't that eliminate the NPE?

Ok. I didn't try to go deep in the code. If you run codegen against our openapi document, you will see there is a NPE, similar to the one I sent you a fix. I just added a check without looking into why the code is invoked.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
sebastien-rosset: I am planning to remove it, after your PR has been merged. I accidentally cherry picked some commits I didn't want to merge

spacether:
> > > > @sebastien-rosset your 3rd point Fix NPE when at least one extension is defined but not x-discriminator-value is already addressed in my PR which will land before this one, right?
> > > 
> > > 
> > > I found one more case where a NPE can be raised. There are two locations in your PR where you do a lookup of the extension.
> > 
> > 
> > Right, but both of those locations the code now checks that the x-discriminator-value key exists before getting it in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
> > 
> > * line 2584 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
> > * line 2639 `if (vendorExtensions != null && !vendorExtensions.isEmpty() && vendorExtensions.containsKey("x-discriminator-value")) {`
> > 
> > Doesn't that eliminate the NPE?
> 
> Ok. I didn't try to go deep in the code. If you run codegen against our openapi document, you will see there is a NPE, similar to the one I sent you a fix. I just added a check without looking into why the code is invoked.

Thanks for explaining that background. Just now I built the openapi generator and produced the client in python-experimental using the spec that you sent me. So it looks like the NPE bug has been resolved in my PR.

modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
```diff
@@ -240,6 +240,9 @@
     @Option(name = {"--generate-alias-as-model"}, title = "generate alias (array, map) as model", description = CodegenConstants.GENERATE_ALIAS_AS_MODEL_DESC)
     private Boolean generateAliasAsModel;
 
+    @Option(name = {"--discriminator-explicit-mapping-verbose"}, title = "explicitly define a discriminator on all models which have one, add verbose mapping to that discriminator, validate discriminators", description = CodegenConstants.DISCRIMINATOR_EXPLICIT_MAPPING_VERBOSE_DESC)
```
jimschubert, line 243: This option needs clarification since it's currently doing a lot.

1. Can we remove the "verbose" since that's not really relevant to the user?
2. The term `mapping` would indicate to users who are familiar with our options that this would take a string in the format `key=value`, which is not the case.
3. The option is triggering OpenAPI 3.x logic, but doesn't indicate that clearly here.

Might I recommend changing this to something more like:

```
@Option(name = {"--legacy-discriminators"}, title = "Support legacy logic for evaluating discriminators", description = CodegenConstants.DISCRIMINATOR_EXPLICIT_MAPPING_VERBOSE_DESC)
private Boolean legacyDiscriminators;
```

This value would need to default to `true` until the new logic becomes default (I don't recall if it's now default or not).

The value would then have to be applied to CodegenConfigurator, which is not done in `Generate.java` here. The same default would have to be done in the Maven and Gradle plugins, with a task to add the feature option to those. We also have a new SBT plugin (pending first release) where this would have to be added.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenDiscriminator.java
```diff
@@ -63,15 +90,36 @@ public void setMappedModels(Set<MappedModel> mappedModels) {
...
+            if (mappingName == null) {
+                throw new RuntimeException("Discriminator mapping name cannot be null for model '" + modelName + "'");
+            }
             this.mappingName = mappingName;
             this.modelName = modelName;
         }
 
+        @Override
+        public int compareTo(MappedModel other) {
+            return getMappingName().compareTo(other.getMappingName());
```
jimschubert, line 120: Potential NPE. You protect in the constructor, but not in the property setter.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -156,7 +156,7 @@ private Boolean getGeneratorPropertyDefaultSwitch(final String key, final Boolea
         return defaultValue;
     }
 
-    private void configureGeneratorProperties() {
+    protected void configureGeneratorProperties() {
```
jimschubert, line 159: This should not be marked as protected as it reduces maintainability. If necessary for tests, please remove `private` to make it package private.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -394,7 +394,7 @@ private void generateModel(List<File> files, Map<String, Object> models, String
     }
 
     @SuppressWarnings("unchecked")
-    private void generateModels(List<File> files, List<Object> allModels, List<String> unusedModels) {
+    protected void generateModels(List<File> files, List<Object> allModels, List<String> unusedModels) {
```
jimschubert, line 397: This should not be marked protected.

jimschubert:
Sorry, I didn't realize many of my comments were for the base PR rather than yours. I'll copy them over to the pre-req PR.

sebastien-rosset:
> Sorry, I didn't realize many of my comments were for the base PR rather than yours. I'll copy them over to the pre-req PR.

Sure, no worries.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
spacether: This was changed to package private

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
spacether: This was changed to package private

spacether: This was changed to package private

modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
sebastien-rosset: This was mistakenly commented in the wrong PR. I am marking it resolved.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2526,36 +2526,60 @@ private Discriminator recursiveGetDiscriminator(Schema sc, OpenAPI openAPI) {
...
                 // All oneOf definitions must contain the discriminator
                 Integer hasDiscriminatorCnt = 0;
+                Integer hasNullTypeCnt = 0;
                 Set<String> discriminatorsPropNames = new HashSet<>();
                 for  (Schema oneOf: composedSchema.getOneOf()) {
+                    if (ModelUtils.isNullType(oneOf)) {
+                        // The null type does not have a discriminator. Skip.
+                        hasNullTypeCnt++;
+                        continue;
+                    }
```
spacether, line 2536: @sebastien-rosset @wing328 @jimschubert what should we allow here, valid objects and empty state values? 
Null only? Null and empty dict?
Is this a language specific decision? If so we should separate it off as a separate method.
- For example for python values like empty dict {} and None are valid empty values here.
- For golang the zero value of a map is nil.

sebastien-rosset: The "null type" (supported in OAS >=3.1) is when the OAS schema is:
```yaml
type: 'null'
```

It means the value in the payload must be null. This can typically be list as one child of "oneOf".

spacether: So I am not asking if it is in the specification. I understand that it is. I am asking if we should also allow an empty object payload here as synonymous with null because that is a python zeto state like null/None.

sebastien-rosset: The change is to specifically cover the following scenario:
```yaml
oneOf:
  - type: null
  - $ref: abc   # possibly a list of inline types or references to types.
```
Without this PR, the count of oneOf children was not be equal to the number of discriminators. That's because the "null" value does not have a discriminator. As a result, the disc.setPropertyName() was not set, which is incorrect.

> I am asking if we should also allow an empty object payload

Can you give how this would be modeled in the OAS document? Here we are parsing OAS schemas, not payloads.

spacether: We are describing payloads here. An empty object represents an rmpty dict. Should this definition be valid?
```
ComposedSchema:
OneOf
- type: null
- type: object
- AnyType
- Pet
discriminator:
  PropertyName: petType

AnyType:
  description: I allow any type including null
```

spacether: I think that AnyType should not be allowed because list and other primitives are different data types than an intended dict/pojo.
My preference is to allow null and empty object but log a warning to the user that this definition is not complaint with the v3 specification bit we allow it.

sebastien-rosset: Thanks for the clarification. Here is one statement in the OAS 3.1 spec:

> When using the discriminator, inline schemas will not be considered.

The way I interpret is that if there is a mix of oneOf children, some through a $ref, others inline, then at runtime:
1. For the inline schemas, use JSON schema validation to determine if the payload matches the schema. Don't use the discriminator.
1. For $ref schemas, don't use JSON schema validation, and instead use the discriminator.

So if my understand is correct, the schema you have provided is a valid OAS schema. HOWEVER, at runtime it will fail validation in many situations, because of the presence of "AnyType". It will fail when the payload is "null", object and Pet. It would work if the payload is a primitive type or an array. So it would be much better to write the OAS spec differently.

I'd rather address these use cases in separate PRs. IMO, we would ideally add validation rules to the `--validate` option.

In the OAS 3.0.3 spec:

> When used, the discriminator will be the name of the property that decides which schema definition validates the structure of the model. As such, the discriminator field MUST be a required field.

spacether: Our users could just as easily write a spec that looks like this:
```
Animal:
oneOf:
- $ref: NullType
- $ref: Pet
discriminator:
  propertyName: animalType

NullType:
type: "null"
```
So I am okay addressing the object type question in the future.
You are proposing accepting nulls in a payload when a strict interpretation of the spec would require the presence of a pojo with a discriminator.propertyName at not null. So I feel like we're in uncharted territory here.

You mention that it would fail because it would double match null on AnyType and type: None. That wasn't the point that I was trying to make. The questions that I am trying to get your answers on are:
- Should AnyType be allowed in oneOf generally?
- Should empty object `type: object` be allowed in oneOf?
I think that we should allow `type: "null"` in oneOf but I do not think that that behavior is defined in the draft 3.1.0 spec yet.

spacether: When mixing types in oneOf, should we disallow the use of allOf + anyOf in the composed schema.
Otherwise you could have:
```
Creature:
allOf:
- $ref: Animal
oneOf:
- type: int
- $ref: Pet
```

sebastien-rosset: > Our users could just as easily write a spec that looks like this:
> 
> ```
> Animal:
> oneOf:
> - $ref: NullType
> - $ref: Pet
> discriminator:
>   propertyName: animalType
> 
> NullType:
> type: "null"
> ```
> 
> Should we handle that referenced null case? I think that we should.
> 

I don't think this matches what is written in the OAS 3.1 spec. If the type is a ref and there is a discriminator, the discriminator must be used. That's my understand of the spec. So your YAML is not equivalent.

> You are proposing accepting nulls in a payload when a strict interpretation of the spec would require the presence of a pojo with a discriminator.propertyName at not null. So I feel like we're in uncharted territory here.

Which part of the spec (3.1) makes you think that? It is indeed my understanding, reading the many threads over the null type that the OAS spec authors intend to replace "nullable" with `type: 'null'` as oneOf child.

> 
> In your comment you mention that it would fail because it would double match null on AnyType and type: None. That wasn't the point that I was trying to make. The questions that I am trying to get your answers on are:
> 
> * Should AnyType be allowed in oneOf generally?
>   

Yes, absolutely, for example below it's a sensible way to say the payload can be anything, including null. I think there is even one thread about this specific item in the draft discussions that led to 3.1, I will try to dig it up.

```yaml
oneOf:
- type: 'null'
- description: any type
```

>   * My vote is no. If they have this in the oneOf then it has to be the only entry. Also then they can't have any allOf definition. Which is equivalent to just using the AnyType directly.
> * Should empty object `type: object` be allowed in oneOf?
> 
> I think that we should allow `type: "null"` in oneOf with a discriminator but I do not think that that behavior is defined in the draft 3.1.0 spec yet.

sebastien-rosset: > When mixing types in oneOf, should we disallow the use of allOf + anyOf in the composed schema?

Sorry I'm not getting this part. Did you mean a schema like this where there is "anyOf"? In the example you provide it is allOf + oneOf, not anyOf.

```
Creature:
allOf:
- $ref: Animal
anyOf:
- type: int
- $ref: Pet
```

Also, can you help me understand how your questions are relevant to this PR? I'm just trying to handle the "null" case here, I'm certainly not claiming to cover all possible composed scenarios + "null". Is there something specific you see that I am breaking with the support of the "null" value? Or are you extending the discussion to other use cases?

> Otherwise you could have the invalid:
> 
> ```
> Creature:
> allOf:
> - $ref: Animal
> oneOf:
> - type: int
> - $ref: Pet
> ```

spacether: This is more warranted as a separate discussion.

spacether: > > Our users could just as easily write a spec that looks like this:
> > ```
> > Animal:
> > oneOf:
> > - $ref: NullType
> > - $ref: Pet
> > discriminator:
> >   propertyName: animalType
> > 
> > NullType:
> > type: "null"
> > ```
> > 
> > 
> > Should we handle that referenced null case? I think that we should.
> 
> I don't think this matches what is written in the OAS 3.1 spec. If the type is a ref and there is a discriminator, the discriminator must be used. That's my understand of the spec. So your YAML is not equivalent.
> 
> > You are proposing accepting nulls in a payload when a strict interpretation of the spec would require the presence of a pojo with a discriminator.propertyName at not null. So I feel like we're in uncharted territory here.
> 
> Which part of the spec (3.1) makes you think that? It is indeed my understanding, reading the many threads over the null type that the OAS spec authors intend to replace "nullable" with `type: 'null'` as oneOf child.
> 
> > In your comment you mention that it would fail because it would double match null on AnyType and type: None. That wasn't the point that I was trying to make. The questions that I am trying to get your answers on are:
> > 
> > * Should AnyType be allowed in oneOf generally?
> 
> Yes, absolutely, for example below it's a sensible way to say the payload can be anything, including null. I think there is even one thread about this specific item in the draft discussions that led to 3.1, I will try to dig it up.
> 
> ```yaml
> oneOf:
> - type: 'null'
> - description: any type
> ```
> 
> > * My vote is no. If they have this in the oneOf then it has to be the only entry. Also then they can't have any allOf definition. Which is equivalent to just using the AnyType directly.
> > * Should empty object `type: object` be allowed in oneOf?
> > 
> > I think that we should allow `type: "null"` in oneOf with a discriminator but I do not think that that behavior is defined in the draft 3.1.0 spec yet.
> 
> It is, though admittedly to comprehend the spec it helps to read the many discussion threads rather than the spec itself. One relevant statement of the spec is:
> 
> > When using the discriminator, inline schemas will not be considered.

When using the discriminator, inline schemas will not be considered.

I read this as don't deserialize inline schemas when a discriminator exists in a composed schema.

spacether:
> Yes, absolutely, for example below it's a sensible way to say the payload can be anything, including null. I think there is even one thread about this specific item in the draft discussions that led to 3.1, I will try to dig it up.
> 
> oneOf:
> - type: 'null'
> - description: any type

This is equivalent to just using the AnyType schema directly because it includes null.
Also, when you have that oneOf definition additional allOf or anyOf definitions will conflict with that oneOf schema and make validation always fail.

sebastien-rosset:
Can we take this discussion outside this PR? Maybe you can open an issue if you want to discuss more.

spacether:
> Can we take this discussion outside this PR? Maybe you can open an issue if you want to discuss more.

Earlier I opened an issue asking for clarification from the spec maintainers here: https://github.com/OAI/OpenAPI-Specification/issues/2215

Happy to discuss this offline or over Slack

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -567,6 +567,43 @@ def fix_model_input_data(model_data, model_class):
...
+            False if the data is from the client
+        model_data (list/dict): data to instantiate the model.
+        cls_visited (set): list of model classes that have been visited.
+            Used to determine the discriminator class without
+            visiting circular references indefinitely.
+
+    Returns:
+        A tuple containing the child class specified by the discriminator,
+        the name of the discriminator property, and the discriminator value
+        in the input data.
```
spacether, line 585: Can you add type descriptions here?
```
A tuple containing (used_model_class, discr_propertyname_js, discr_value)
used_model_class (class): the chosen class that will be used to deserialize the data, for example dog.Dog
discr_propertyname_js (string): the discriminator.propertyName from the opeanpi spec
discr_value (string): the class name in model_data which is the key that we use to return used_model_class from our discriminator, for example 'Dog'
```

sebastien-rosset: Sure, I am adding. We should use the term "OpenAPI document" not "OpenAPI specification" when referring to a YAML/JSON document describing an API. The "spec" is the standard.

samples/openapi3/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -148,7 +148,7 @@ def __new__(cls, *args, **kwargs):
         oneof_anyof_classes = []
         oneof_anyof_classes.extend(cls._composed_schemas.get('oneOf', ()))
         oneof_anyof_classes.extend(cls._composed_schemas.get('anyOf', ()))
-        new_cls = cls.get_discriminator_class(kwargs)
+        new_cls = get_discriminator_class(cls, kwargs, set())
```
spacether, line 151: We may be passing in _visited_composed_classes in kwargs here, so we need to include it in the inputs to get_discriminator_class.
Can you extract _visited_composed_classes from kwargs and pass it in to get_discriminator_class here?

samples/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -148,7 +148,7 @@ def __new__(cls, *args, **kwargs):
         oneof_anyof_classes = []
         oneof_anyof_classes.extend(cls._composed_schemas.get('oneOf', ()))
         oneof_anyof_classes.extend(cls._composed_schemas.get('anyOf', ()))
-        new_cls = cls.get_discriminator_class(kwargs)
+        new_cls = get_discriminator_class(cls, kwargs, set())
```
spacether, line 151: We may be passing in _visited_composed_classes in kwargs here, so we need to include it in the inputs to get_discriminator_class.
Can you extract _visited_composed_classes from kwargs and pass it in to get_discriminator_class here?

spacether: Never mind, I think that we are fine here

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -622,6 +622,46 @@ def deserialize_primitive(data, klass, path_to_item):
...
+        discr_value = model_data[discr_propertyname_js]
+    elif discr_propertyname_py in model_data:
+        discr_value = model_data[discr_propertyname_py]
+    else:
+        # The input data does not contain the discriminator property.
+        # The caller will raise an exception.
+        return None
+    class_name_to_discr_class = discriminator[discr_propertyname_py]
+    used_model_class = class_name_to_discr_class.get(discr_value)
+    if used_model_class is None:
```
spacether, line 655: Why are we looping over these values?
Do we want to loop over the ancestors here?
If we have:
allOf:
- Animal
oneOf:
- Cat
- Dog
we are looking at the discriminators in Cat and Dog.
Don't we instead want to look at the discriminator in Animal? Can you add comments here describing these different cases?

sebastien-rosset: > Why are we looping over these values?
> Do we want to loop over the ancestors here?

What is your exact definition of ancestor? Is it the allOf children?

> If we have:
> allOf:
> 
> * Animal
> 
> oneOf:
> 
> * Cat
> * Dog
> 
> we are looking at the discriminators in Cat and Dog.
> Don't we instead want to look at the discriminator in Animal? 

Sure yes in that example, but what if instead the schema is:
```yaml
oneOf:
- Animal
- Plant
```

In that case there are potentially many subclasses of animal and plant, so we would need to traverse through oneOf. Same thing for anyOf.

> Can you add comments here describing what we are doing?
> `# if we didn't find a discriminated class it may exist in an ancestor`
> Whatever you are adding, can you add a sample schema and test that verifies that this feature is working?
> Otherwise if it breaks in the future, we will have no way of seeing that it has broken.
> 
> For example [ParentPet](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/models/parent_pet.py) only discriminates to
> 
> * ChildCat
> * ChildDog
> * ChildLizard
> 
> But it allOf includes [GrandparentAnimal](https://github.com/OpenAPITools/openapi-generator/blob/30ee3f343eb681514ce5b33438de3b8f4fb1f50b/samples/client/petstore/python-experimental/petstore_api/models/grandparent_animal.py#L57).
> which includes those AND ParentPet.
> So if we deserialize a payload:
> {petType: "ParentPet"} where data_type = parent_pet.ParentPet, we should travel back through its GrandparentAnimal ancestor to find the correct class to deserialize to, as long as the discriminator is the same.
> 
> Should we look in the current discriminator, then look in descendants, then look in ancestors?
> It looks like yes per https://github.com/OpenAPITools/openapi-generator/pull/4913/files

spacether: We are agreed that we should loop over descendants then ancestors.
Descendants are the oneOf/anyOf children and are present in the class_name_to_discr_class.
Ancestors are in cls._composed_schemas['allOf']
Can you:
- add comments describing where we are looping over ancestors and descentents
- add code to loop over ancestors
- add a test case that demonstrates deserializing a descendent
- add a test case that demonstrates deserializing an ancestor. You should be able to do this with the ParentPet class deserializing a ParentPet
- update this function to not pass in model_data, instead pass in the value from the payload corresponding to the discriminator.propertyName. This lets us cache the results of function calls in the next item.
- cache the results of the function cals by adding a lru_cache decorator around the get_discriminator_class function. I am fine with us only adding this for python3. You can do this with pseudocode:
```
if six.PY3:
  get_discriminator_class = lru_cache(get_discriminator_class)
```

samples/client/petstore/python-experimental/petstore_api/model_utils.py
spacether: I mean:
```
_visited_composed_classes = kwargs.get('_visited_composed_classes', ())
new_cls = get_discriminator_class(cls, kwargs, _visited_composed_classes)
```
You need this because if we travel through two discriminators. The `__new__` call to the class with the second discriminator will be passing in _visited_composed_classes here.

samples/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -883,6 +883,46 @@ def deserialize_primitive(data, klass, path_to_item):
         )
 
 
+def get_discriminator_class(model_class, model_data, cls_visited):
+    """Returns the child class specified by the discriminator.
+
+    Args:
+        model_class (OpenApiModel): the model class
+        model_data (list/dict): data to instantiate the model.
+        cls_visited (set): list of model classes that have been visited.
```
spacether, line 892: cls_visited should be a tuple, not a set. Our classes are not hashable so we need to make sure to store them in a tuple not a set.

samples/client/petstore/python-experimental/petstore_api/model_utils.py
sebastien-rosset: hmm, ok, I am going through your code to try to understand what it does, because now there would be two parts of the code adding to the _visited_composed_classes tuple, with two separate tests to check if a class has already been visited or not.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -622,6 +622,46 @@ def deserialize_primitive(data, klass, path_to_item):
         )
 
 
+def get_discriminator_class(model_class, model_data, cls_visited):
+    """Returns the child class specified by the discriminator.
+
+    Args:
+        model_class (OpenApiModel): the model class
+        model_data (list/dict): data to instantiate the model.
+        cls_visited (set): list of model classes that have been visited.
```
spacether, line 631: cls_visited should be a tuple, not a set. Our classes are not hashable so we need to make sure to store them in a tuple not a set.

sebastien-rosset:
Before I proceed, can you make the following changes:

1. In the [new function](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L74), can you add comments to elaborate why *you are sure that you want to instantiate it this time*, with an example.
1. Can you point me to source code with a unit test that exercises the above scenario?
1. In the [new function](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L73), you state "we don't have a discriminator". I suggest rephrasing that to "The OpenAPI schema does not specify a discriminator for the cls class".
1. [Here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L81), can you explain why you are traversing the oneOf and anyOf children to lookup a discriminator?
1. Can you point me to a unit test that exercise the above item? I.e. 
1. I think [these lines](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L78-L80)  should be moved  closer to line 99 because that local variable is not used until later in the code.
1. Can you clarify this [comment](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L96). Avoid using "if we are coming". Why do you have this code?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -65,31 +65,53 @@ class OpenApiModel(object):
...
             cls.discriminator is None or
             cls in visited_composed_classes
         ):
-            # we don't have a discriminator
-            # or we have already visited this class before and are sure that we
-            # want to instantiate it this time
+            # There is no discriminator.
+            # Or we have already visited this class before and are sure that we
+            # want to instantiate it this time. For example, this happens when
+            # we deserialize Dog from Animal. Dog contains an Animal instance. We go from Animal to Dog back to Animal
```
spacether, line 84: For example if Animal has a discriminator
petType and we pass in "Dog", and the class Dog
allOf includes Animal, we move through Animal
once using the discriminator, and pick Dog.
Then in Dog, we will make an instance of the
Animal class but this time we won't travel
through its discriminator because we passed in
_visited_composed_classes = (Animal,)

spacether:
> Before I proceed, can you make the following changes:
> 
> 1. In the [new function](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L74), can you add comments to elaborate why _you are sure that you want to instantiate it this time_, with an example.
> 2. Can you point me to source code with a unit test that exercises the above scenario?

[This test](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/tests/test_deserialization.py#L104) deserializes a Dog given petstore_api.Animal as the data type, so that tests exercises the "sure you want to instantiate it" use case.

spacether:
> 3. In the [new function](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L73), you state "we don't have a discriminator". I suggest rephrasing that to "The OpenAPI schema does not specify a discriminator for the cls class".

That is fine, how about `the openapi spec document does not define a discriminator for this schema`

spacether:
> 4\. I think [these lines](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L78-L80)  should be moved  closer to line 99 because that local variable is not used until later in the code.

We need the oneof_anyof_classes where it is on line 69 so we can use it to check if None is in oneof_anyof_classes and that None was passed in as the first positional argument.
This check must be done before returning an instance of cls because we will never have the data that we need to return an instance of cls when we pass in None. Am I missing something here?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -65,31 +65,53 @@ class OpenApiModel(object):
...
+                discr_propertyname_js, kwargs.get(discr_propertyname_py))
             raise ApiValueError(
                 "Cannot deserialize input data due to invalid discriminator "
                 "value. The OpenAPI document has no mapping for discriminator "
                 "property '%s'='%s' at path: %s" %
-                (disc_prop_name_js, disc_prop_value, path_to_item)
+                (discr_propertyname_js, disc_prop_value, path_to_item)
             )
 
         if new_cls in visited_composed_classes:
```
spacether, line 117: # if we are making an instance of a composed schema Descendent
            # which allOf includes Ancestor, then Ancestor contains
            # a discriminator that includes Descendent.
            # So if we make an instance of Descendent, we have to make an
            # instance of Ancestor to hold the allOf properties.
            # This code detects that use case and makes the instance of Descendant
            # For example:
            # When making an instance of Dog, _visited_composed_classes = (Dog,)
            # then we make an instance of Animal to include in dog._composed_instances
            # so when we are here, cls is Animal
            # cls.discriminator != None
            # cls not in _visited_composed_classes
            # new_cls = Dog
            # but we know we know that we already have Dog
            # because it is in visited_composed_classes
            # so make Animal here

spacether:
> 5. Can you clarify this comment. Avoid using "if we are coming". Why do you have this code?
> 6. Can you point me to the source code with unit tests that exercise the above test?
[This unit test](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/test/test_dog.py#L29) uses that code path.

spacether:
> It would be helpful to explain (or have a reference to doc elsewhere) that explains how ancestors (allOf) are implemented. In other programming languages (such as Java), there is a class hierarchy. In go-experimental, it's going in the direction of using composition. In python, the the super class is not the OAS ancestor.

How about we put that in the ModelComposed class?
"""
_composed_schemas (dict) stores the anyOf/allOf/oneOf classes
  key (str): allOf/oneOf/anyOf
  value (list): the classes in the XOf definition.
    Note: none_type can be included when the openapi document version >= 3.1.0
_composed_instances (list): stores a list of instances of the composed schemas defined in _composed_schemas. When properties are accessed in the self instance, they are returned from the self._data_store or the data stores in the instances in self._composed_schemas
_var_name_to_model_instances (dict): maps between a variable name on self and the composed instances (self included) which contain that data
  key (str): property name
  value (list): list of class instances, self or instances in _composed_instances which contain the value that the key is referring to
```

File: modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache (hunk 14 above)
Comment (sebastien-rosset) [reply modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache:135]: I have copied this as is. Please submit a pull request if you want to make changes.

File: modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache (hunk 10 above)
Comment (sebastien-rosset) [reply modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache:718]: I will not do 6 in this PR. This increases the scope of the PR and should be handled in a different PR.

Comment (spacether) [reply modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache:721]: Thanks for letting me know about point 6. I can definitely add 6 in a different PR.

File: modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache [hunk 15]
```diff
@@ -622,6 +691,56 @@ def deserialize_primitive(data, klass, path_to_item):
...
+    """
+
+    if model_class in cls_visited:
+        # The class has already been visited and no suitable class was found.
+        return None
+    cls_visited.append(model_class)
+    used_model_class = None
+    if discr_name in model_class.discriminator:
+        class_name_to_discr_class = model_class.discriminator[discr_name]
+        used_model_class = class_name_to_discr_class.get(discr_value)
```
spacether, line 720: We are not yet handling the use case where discr_name is not in model_class.discriminator here.
If we hit that case, then the discriminator has changed and we did not find what we are looking for.
Here, can you add:
```
else:
    # the discr_name has changed and we did not find the class that we are looking for
    return None
```

Or maybe we should make it the clearer:
```
if discr_name not in model_class.discriminator:
    # the discr_name has changed and we did not find the class that we are looking for
    return None
class_name_to_discr_class = model_class.discriminator[discr_name]
used_model_class = class_name_to_discr_class.get(discr_value)
```

sebastien-rosset:
> Thank you for adding those sample tests.
> 
> The remaining ask here is to add tests that require deserialization through the discriminators in a descendent (oneOf/anyOf schemas) or in an ancestor (allOf ancestor of self composed schema)

There is a ancestor relationship with 'allOf' in chordate, primate, mammal, hominid and the unit test is exercising this, I don't understand why you are stating the tests do not include ancestors. Also see below some additional details. 

> Also, ideally we will use existing sample schemas rather than adding additional
> ones, because that helps keep our sample spec as small as possible.

Can you provide a specific example that you want? You are very opinionated about how things should be done, can you please provide a specific relationship using the existing spec? Otherwise I think it'll keep going for a long time.

> 
> Examining the added tests we are testing the following cases:
> 
> 1. using a discriminator map defined in a composed schema without using an ancestor or descendant's discriminator.
> 
> self is BiologyChordate and its discriminator contains Mammal, Primate, Reptile, and Hominid
> so se we are using self's discriminator without using an ancestor or descendent's discriminator:
> 
> * biology.Mammal from BiologyChordate
> * biology.Primate from BiologyChordate
> * biology.Reptile from BiologyChordate
> * biology.Hominid from BiologyChordate
>   Per:
> 
> ```
>     @cached_property
>     def discriminator():
>         val = {
>             'biology.Hominid': biology_hominid.BiologyHominid,
>             'biology.Mammal': biology_mammal.BiologyMammal,
>             'biology.Primate': biology_primate.BiologyPrimate,
>             'biology.Reptile': biology_reptile.BiologyReptile,
>         }
>         if not val:
>             return None
>         return {'class_name': val}
> ```
> 
> * biology.Hominid from BiologyMammal
>   Per:
> 
> ```
>     @cached_property
>     def discriminator():
>         val = {
>             'biology.Hominid': biology_hominid.BiologyHominid,
>             'biology.Primate': biology_primate.BiologyPrimate,
>         }
>         if not val:
>             return None
>         return {'class_name': val}
> ```
> 
> So these tests do not exercise new code and do not demonstrate ancestor or descedant deserialization.

I'm not sure why you are saying this. I modeled the schemas to address the missing recursion in the existing code when the discriminators are accessible through a traversal of the ancestors. The way I tested locally is I wrote the unit tests, wrote the code until the unit test passed, then I  disabled the newly-introduced recursion in `get_discriminator_class()`, and the unit test failed.
When I disable the recursion, the code is as it was before my proposed changes. So the fact that UT outcome changes based on the recursion is a proof that the new code is exercised. Also, during debugging I added some print statements and I could see the new code was executed. 

I will wait for you response before responding to your points below.

> They are functionally equivalent to
> 
> * test_deserialize_animal (only picks one class in cls.discriminator)
> * test_deserialize_mammal (only picks one class in cls.discriminator)
>   The use the code path in get_discriminator_class line 720 where we use 1 discriminaotr map and get the class:
>   `used_model_class = class_name_to_discr_class.get(discr_value)`
> 
> _Can you remove these tests and sample schemas because they are not exercising a new code path?_
> 
> ## Ancestor deserialization:
> * biology.Hominid from biology.Hominid
>   This test does deserialize using an ancestor's discriminator, but we can already do this using the
>   existing class/schema ParentPet. Can you:
> 
> _Change this test so it deserializes ParentPet from ParentPet? Remove the added biology.Hominid schema because we no longer need it?_
> 
> ## Descendant Deserialization:
> We have not yet demonstrated this yet, and we need to because we have added this feature.
> We can do this by deserializing from a grandparent through a oneof child into a oneof child of that child.
> So if we deserialize from a grandparent to a grandchild using oneOf relationships, we will demonstrate this.
> 
> _Can you add a new schema with this relationship and add a test deserializing it?_
> 
> * For example you could add a Pig schema, which is in mammal's oneOf class list
>   
>   * Pig can OneOf define BasquePig and DanishPig, all of with include the className property
>   * Then deserializing a BasquePig payload from a mammal data_type will demonstrate descendent deserialization because BasquePig is not in mammal's discriminator map. It is in Pig's discriminator map.

spacether:
Thank you for describing what is going on in those unit tests more. I was wrong in saying that those first examples didn't use ancestor or descendent inheritance. They do use ancestor inheritance. Sorry about that.
First they move through the single discriminator in Chordate to select new_cls. Then they check the discriminator in the chosen new_cls use the ancestor allOf relationship to decide to deserialize using new_cls.

All of these tests demonstrate ancestor inheritance:
- biology.Mammal from BiologyChordate
- biology.Primate from BiologyChordate
- biology.Reptile from BiologyChordate
- biology.Hominid from BiologyChordate
- biology.Hominid from BiologyMammal

Let's continue this discussion on our call later.
Here is the get_discriminator_class function with specific variables for descendant and ancestor classes which may help with our discussion:
```
def get_discriminator_class(model_class,
                            discr_name,
                            discr_value, cls_visited):
    """Returns the child class specified by the discriminator.

    Args:
        model_class (OpenApiModel): the model class.
        discr_name (string): the name of the discriminator property.
        discr_value (any): the discriminator value.
        cls_visited (list): list of model classes that have been visited.
            Used to determine the discriminator class without
            visiting circular references indefinitely.

    Returns:
        used_model_class (class/None): the chosen child class that will be used
            to deserialize the data, for example dog.Dog.
            If a class is not found, None is returned.
    """

    if model_class in cls_visited:
        # The class has already been visited and no suitable class was found.
        return None
    cls_visited.append(model_class)
    used_model_class = None
    if discr_name in model_class.discriminator:
        class_name_to_discr_class = model_class.discriminator[discr_name]
        used_model_class = class_name_to_discr_class.get(discr_value)
    if used_model_class is None:
        # We didn't find a discriminated class in class_name_to_discr_class.
        # The discriminator mapping may exist in a descendant (anyOf, oneOf)
        # or ancestor (allOf).
        # Ancestor example: in the "Dog -> Mammal -> Chordate -> Animal"
        #   hierarchy, the discriminator mappings may be defined at any level
        #   in the hieararchy.
        # Descendant example: a schema is oneOf[Plant, Mammal], and each
        #   oneOf child may itself be an allOf with some arbitrary hierarchy,
        #   and a graph traversal is required to find the discriminator.
        descendant_classes = (
            model_class._composed_schemas.get('oneOf', ()) +
            model_class._composed_schemas.get('anyOf', ())
        )
        ancestor_classes = model_class._composed_schemas.get('allOf', ())
        possible_classes = (
            descendant_classes +
            ancestor_classes
        )
        for cls in possible_classes:
            # Check if the schema has inherited discriminators.
            if cls.discriminator is not None:
                used_model_class = get_discriminator_class(
                                    cls, discr_name, discr_value, cls_visited)
                if used_model_class is not None:
                    return used_model_class
    return used_model_class
```

spacether:
Further work will be needed to resolve #4912. That work will be done at a later time.

PR #5810, opened 2020-04-02
[Python-experimental] Fix TypeError: unhashable type: 'list'

sebastien-rosset:
During unmarshaling, handle scenario when value is a list. The generated code was trying to insert the value into a set, and because a list is not hashable, a TypeError exception was raised.

### PR checklist


sebastien-rosset:
@spacether

sebastien-rosset:
@spacether , I think I need to do one more thing: the generate model classes must be hashable, because for this change to work, if the list is converted to a tuple, the elements in the list must be hashable. That means the generated model classes need to implement __hash__ and __eq__.

Thoughts?

spacether:
> @spacether , I think I need to do one more thing: the generate model classes must be hashable, because for this change to work, if the list is converted to a tuple, the elements in the list must be hashable. That means the generated model classes need to implement `__hash__` (and `__eq__` too but it's already implemented).
> 
> Thoughts?

Yup, we definitely need to implement class hashing.
It looks like by default user classes return the id of self. That may be a nice simple solution for us here. Otherwise we need to hash all the primitive types present in a class's values.

sebastien-rosset:
> We could use id(self) as the hashing function, that is what user classes normally user.
> But that would not give the same integer for two class instances that have the same data in them.
> So this is fast but a little inaccurate.
> 
> Another better solution would be to use:
> 
> ```
> def __hash__(self):
>   dict_with_simple_values = sanitize_for_serialization(self.to_dict())
>   return hash(json.dumps(dict_with_simple_values))
> ```
> 
> Which would be slower but more accurate.
> We have to make sure that we can also handle date values, file values, and None values.
> What do you prefer @sebastien-rosset ?

How about?
```
return hash(frozenset(self._data_store.items()))
```

samples/openapi3/client/petstore/python-experimental/petstore_api/model_utils.py
```diff
@@ -204,6 +208,10 @@ def to_str(self):
         """Returns the string representation of the model"""
         return pprint.pformat(self.to_dict())
 
+    def __hash__(self):
+        """Returns the hash value of this object."""
+        return hash(frozenset(self._data_store.items()))
+
     def __eq__(self, other):
         """Returns true if both objects are equal"""
         if not isinstance(other, self.__class__):
```
spacether, line 217: How about using our hash method in out `__eq__` method?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_tostr_eq_simple.mustache
```diff
@@ -2,6 +2,10 @@
         """Returns the string representation of the model"""
         return str(self.value)
 
+    def __hash__(self):
+        """Returns the hash value of this object."""
+        return hash(self._data_store['value'])
```
spacether, line 7: Can you change this to `hash(frozenset(self._data_store.items()))` so it is the same as the one in modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_todict_tostr_eq_shared.mustache ? That will reduce the chance of hash collisions.

spacether:
@sebastien-rosset can you include a model that tests your hashing?
Can your model include parameters of multiple type including:
- a property with another model value, like an instance of Cat or Dog
- a property with a list value
- a property with a file value
- a property with a date value
- a property with a datetime value
- a property with a dict value

My concern is that the above types may not be hashable even when in a frozenset.
My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.

sebastien-rosset:
> @sebastien-rosset can you include a model that tests your hashing?
> Can your model include parameters of multiple type including:
> 
> a property with another model value, like an instance of Cat or Dog
> a property with a list value
> a property with a file value
> a property with a date value
> a property with a datetime value
> a property with a dict value
> My concern is that the above types may not be hashable even when in a frozenset.
> My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.

Yes it does (all of them). Do you have a suggestion?

samples/openapi3/client/petstore/python-experimental/petstore_api/model_utils.py
sebastien-rosset: > How about using our hash method in out `__eq__` method?

What are you trying to achieve by doing this? hash can have collisions, so it cannot be used as is, we must do an actual equality comparison.

spacether:
> > @sebastien-rosset can you include a model that tests your hashing?
> > Can your model include parameters of multiple type including:
> > a property with another model value, like an instance of Cat or Dog
> > a property with a list value
> > a property with a file value
> > a property with a date value
> > a property with a datetime value
> > a property with a dict value
> > My concern is that the above types may not be hashable even when in a frozenset.
> > My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.
> 
> Yes it does (all of them). Do you have a suggestion?

So property values are stored as native python types. So when additionalproperties are set to true we can have:
```
self._data_store = {
  "list": [1, 2, 3],
  "dict": {"a": "a"}
}
```
Your code does not yet handle this use case. That's why I am asking for the above test.
That data cannot be hashed with: `hash(frozenset(self._data_store.items()))`

samples/openapi3/client/petstore/python-experimental/petstore_api/model_utils.py
spacether: My intent was to make our definition consistent with what I was reading in python docs about implementing `__hash__` and `__eq__`. If we have a collision chance, I agree that keeping this as-is is better.

sebastien-rosset:
> > > @sebastien-rosset can you include a model that tests your hashing?
> > > Can your model include parameters of multiple type including:
> > > a property with another model value, like an instance of Cat or Dog
> > > a property with a list value
> > > a property with a file value
> > > a property with a date value
> > > a property with a datetime value
> > > a property with a dict value
> > > My concern is that the above types may not be hashable even when in a frozenset.
> > > My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.
> > 
> > 
> > Yes it does (all of them). Do you have a suggestion?
> 
> So property values are stored as native python types. So when additionalproperties are set to true we can have:
> 
> ```
> self._data_store = {
>   "list": [1, 2, 3],
>   "dict": {"a": "a"}
> }
> ```
> 
> Your code does not yet handle this use case. That's why I am asking for the above test.
> That data cannot be hashed with: `hash(frozenset(self._data_store.items()))`

ok. If there are multiple levels of nested schemas, are you saying they are stored as plain dicts in python? Or each nested value is an instance of the python class implementing the schema? I thought that was the latter.

spacether:
> > > > @sebastien-rosset can you include a model that tests your hashing?
> > > > Can your model include parameters of multiple type including:
> > > > a property with another model value, like an instance of Cat or Dog
> > > > a property with a list value
> > > > a property with a file value
> > > > a property with a date value
> > > > a property with a datetime value
> > > > a property with a dict value
> > > > My concern is that the above types may not be hashable even when in a frozenset.
> > > > My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.
> > > 
> > > 
> > > Yes it does (all of them). Do you have a suggestion?
> > 
> > 
> > So property values are stored as native python types. So when additionalproperties are set to true we can have:
> > ```
> > self._data_store = {
> >   "list": [1, 2, 3],
> >   "dict": {"a": "a"}
> > }
> > ```
> > 
> > 
> > Your code does not yet handle this use case. That's why I am asking for the above test.
> > That data cannot be hashed with: `hash(frozenset(self._data_store.items()))`
> 
> ok. If there are multiple levels of nested schemas, are you saying they are stored as plain dicts in python? Or each nested value is an instance of the python class implementing the schema? I thought that was the latter.

It can be both. 
Lists are always stored as lists and you can define a list that contains many nested lists. 
AdditionalProperties definitely lets it be both.

sebastien-rosset:
> > > @sebastien-rosset can you include a model that tests your hashing?
> > > Can your model include parameters of multiple type including:
> > > a property with another model value, like an instance of Cat or Dog
> > > a property with a list value
> > > a property with a file value
> > > a property with a date value
> > > a property with a datetime value
> > > a property with a dict value
> > > My concern is that the above types may not be hashable even when in a frozenset.
> > > My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.
> > 
> > 
> > Yes it does (all of them). Do you have a suggestion?
> 
> So property values are stored as native python types. So when additionalproperties are set to true we can have:
> 
> ```
> self._data_store = {
>   "list": [1, 2, 3],
>   "dict": {"a": "a"}
> }
> ```
> 
> Your code does not yet handle this use case. That's why I am asking for the above test.
> That data cannot be hashed with: `hash(frozenset(self._data_store.items()))`

Aren't there already model cases for __eq__? Sorry I'm not very familiar with this code.

spacether:
> > > > @sebastien-rosset can you include a model that tests your hashing?
> > > > Can your model include parameters of multiple type including:
> > > > a property with another model value, like an instance of Cat or Dog
> > > > a property with a list value
> > > > a property with a file value
> > > > a property with a date value
> > > > a property with a datetime value
> > > > a property with a dict value
> > > > My concern is that the above types may not be hashable even when in a frozenset.
> > > > My testing shows that list will need to be converted to tuple and dicts will need to be converted via frozenset.
> > > 
> > > 
> > > Yes it does (all of them). Do you have a suggestion?
> > 
> > 
> > So property values are stored as native python types. So when additionalproperties are set to true we can have:
> > ```
> > self._data_store = {
> >   "list": [1, 2, 3],
> >   "dict": {"a": "a"}
> > }
> > ```
> > 
> > 
> > Your code does not yet handle this use case. That's why I am asking for the above test.
> > That data cannot be hashed with: `hash(frozenset(self._data_store.items()))`
> 
> Aren't there already model cases for **eq**? Sorry I'm not very familiar with this code.

Please check out the classes that we have in our sample spec. There are probably ones in there which already contain or allow list and nested data fields.
Like here:
- Many types of properties https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/python-client-experimental/petstore-with-fake-endpoints-models-for-testing.yaml#L1477
- Any type https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/python-client-experimental/petstore-with-fake-endpoints-models-for-testing.yaml#L1572
- Nested arrays https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/2_0/python-client-experimental/petstore-with-fake-endpoints-models-for-testing.yaml#L1668

spacether:
We have an equals test here: 
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/samples/client/petstore/python-experimental/tests/test_enum_arrays.py
but it isn't testing much.

sebastien-rosset:
> We have an equals test here:
> https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/tests/test_enum_arrays.py#L141
> but it isn't testing much.

Let's step back a bit. Is it really needed to have `values = set()` in the __getattr__ function? Is there another way? I am trying to understand the logic in the getattr, it looks like this part of the code is handling undeclared properties from additionalproperties ?

spacether:
> > We have an equals test here:
> > https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/tests/test_enum_arrays.py#L141
> > but it isn't testing much.
> 
> Let's step back a bit. Is it really needed to have `values = set()` in the **getattr** function? Is there another way? I am trying to understand the logic in the getattr, it looks like this part of the code is handling undeclared and optional properties?
> 
> Can you explain what each `model_instance` represent in the `for model_instance in model_instances:` loop at line 46?

Sure, each composed model stores child or parent models in itself under the hood. Those instances are stored in the model instances. Because a property can exist in multiple schemas/classes/instances. One possibility is that the properties stored in the instances get out of sync with one another. To prevent that we get all instances that store a given property with
_var_name_to_model_instances, we loop through those instances and extract our property values, and add them to a set. That set's length should only be 1 because there should only be one value that we have stored. That's why we have that loop.

sebastien-rosset:
> > > We have an equals test here:
> > > https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/python-experimental/tests/test_enum_arrays.py#L141
> > > but it isn't testing much.
> > 
> > 
> > Let's step back a bit. Is it really needed to have `values = set()` in the **getattr** function? Is there another way? I am trying to understand the logic in the getattr, it looks like this part of the code is handling undeclared and optional properties?
> > Can you explain what each `model_instance` represent in the `for model_instance in model_instances:` loop at line 46?
> 
> Sure, each composed model stores child or parent models in itself under the hood. Those instances are stored in the model instances. Because a property can exist in multiple schemas/classes/instances. One possibility is that the properties stored in the instances get out of sync with one another. To prevent that we get all instances that store a given property with
> _var_name_to_model_instances, we loop through those instances and extract our property values, and add them to a set. That set's length should only be 1 because there should only be one value that we have stored. That's why we have that loop.

Thanks. What if instead of `values = set()`, we have a list? `values = []` Would your logic still work? Or is there a situation when there are two values that are equal, so even though the value is added twice, the size of the set is still one?

spacether:
We could use a list. But we should only add items to it if they are not present in the list (which is set behavior).

sebastien-rosset:
How about something like this? https://github.com/OpenAPITools/openapi-generator/pull/5810/commits/de4432412a8be75d3fe205d2c5280f06acc3fe01

I'm trying to understand what's the likelihood of having a very large list.

spacether:
> How about something like this? [de44324](https://github.com/OpenAPITools/openapi-generator/commit/de4432412a8be75d3fe205d2c5280f06acc3fe01)
> 
> I'm trying to understand what's the likelihood of having a very large list.

That should work fine
The only way I see us having large lists here is if you have a composed schema class with aproper being used in many allof or anyof classes and for the anyof class case you would have to not have a discriminator. It seems possible but unlikely.

sebastien-rosset:
> > How about something like this? [de44324](https://github.com/OpenAPITools/openapi-generator/commit/de4432412a8be75d3fe205d2c5280f06acc3fe01)
> > I'm trying to understand what's the likelihood of having a very large list.
> 
> That should work fine
> The only way I see us having large lists here is if you have a composed schema class with a property being used in many allof or anyof classes and for the anyof class case you would have to not have a discriminator in our composed schema. It seems possible but unlikely.

Right. And what is the likelihood of having two different values for the same property in two different child model? Does this happen when the application code  is somehow changing the data directly in self._var_name_to_model_instances? Would there be input data (e.g. JSON document) that would cause inconsistencies across two child models?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_tostr_eq_simple.mustache
sebastien-rosset: Based on the discussion, I think I can mark this resolved.

spacether:
> > > How about something like this? [de44324](https://github.com/OpenAPITools/openapi-generator/commit/de4432412a8be75d3fe205d2c5280f06acc3fe01)
> > > I'm trying to understand what's the likelihood of having a very large list.
> > 
> > 
> > That should work fine
> > The only way I see us having large lists here is if you have a composed schema class with a property being used in many allof or anyof classes and for the anyof class case you would have to not have a discriminator in our composed schema. It seems possible but unlikely.
> 
> Right. And what is the likelihood of having two different values for the same property in two different child model? Does this happen when the application code is somehow changing the data directly in self._var_name_to_model_instances? Would there be input data (e.g. JSON document) that would cause inconsistencies across two child models?

An input payload is converted into a composed schema model. So the value starts out the same. The values can only differ when a malicious actor manually changes the value in one of the composed instances. There's already a python test that shows that an exception is raised if the code does that. It's in dog, cat, or animal tests I think.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_tostr_eq_simple.mustache
spacether: Yup, no need for this method anymore

PR #5820, opened 2020-04-03
[codegen] change x-oneOf-name to x-one-of-name. Consistency with naming conventions and x-all-of-name

sebastien-rosset:
### PR checklist


sebastien-rosset:
@wing328

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -1840,7 +1840,13 @@ public String toAnyOfName(List<String> names, ComposedSchema composedSchema) {
     }
 
     /**
-     * Return the name of the oneOf schema
+     * Return the name of the oneOf schema.
+     *
+     * This name is used to set the value of CodegenProperty.openApiType.
```
sebastien-rosset, line 1845: @wing328 ,  it would be good to explain  how this field is used. In other words, the 'x-one-of-name' extension exists for a reason. Why should an OAS author set this extension? What is the value of setting that extension?

wing328: > Is it ok to set the value of x-one-of-name to the same value as the container? 

Let's step back for a second. Why not update LivingOrganism if the users want to have complete control of the naming?

In other words, x-one-of-name seems to be redundant if we can simply use the container (schema) name.

sebastien-rosset: > > Is it ok to set the value of x-one-of-name to the same value as the container?
> 
> Let's step back for a second. Why not update LivingOrganism if the users want to have complete control of the naming?
> 
> In other words, x-one-of-name seems to be redundant if we can simply use the container (schema) name.

Yes, I totally agree the the container name would be the best default name. I was wondering why that was not the case. I suspect this is because in ModelUtils, the functions typically don't have access to anything other than the local schema passed as an argument, which makes it harder to access elements like the container schema. 
Here the use of x-one-of-name  is just a workaround.

wing328: > Without x-one-of-name, the generated oneOf name for the schema below would be OneOf<Animal,Plant,Fungus,Bacteria>

I think it's because the default `toOneOfName` is used: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java

I'll do some tests to try to figure out the "correct" way to override this method in the language generator so that we don't need to use x-one-of-name if possible.

wing328:
In the future, we may come up with a better way to name the oneOf schema without using the extension and we will remove the extension with further notice (so it's not considered as a breaking change)

PR #5828, opened 2020-04-05
[C][Client]Fix data lost when libcurl write-data callback function is called multiple times

ityuhui:
When using libcurl to get the http response, the application program needs to configure the write-data callback function. 
```c
       curl_easy_setopt(handle,
                         CURLOPT_WRITEFUNCTION,
                         writeDataCallback);
       curl_easy_setopt(handle,
                         CURLOPT_WRITEDATA,
                         &apiClient->dataReceived);
```
When libcurl receives data(http response), it will call the callback function (defined in openapi-generator/c-libcurl)
```c
size_t writeDataCallback(void *buffer, size_t size, size_t nmemb, void *userp) {
...
}
```
to put the data from ```void *buffer``` to user's memory location ```void *userp``` (&apiClient->dataReceived).

But sometimes (commonly the response data is big) for a http request,  the response data will be divided to multiple parts and sent by multiple times, so the callback function ```writeDataCallback``` will be called multiple times.

The current writeDataCallback will lose the data the last time received:
```c
size_t writeDataCallback(void *buffer, size_t size, size_t nmemb, void *userp) {
    *(char **) userp = strdup(buffer);    <=== The old userp will be discarded, and userp points to a new memory.
    return size * nmemb;
}
```
So I changed the apiClient_t and writeDataCallback
```c
typedef struct apiClient_t {
...
    long dataReceivedLen;
...
} apiClient_t;

       curl_easy_setopt(handle,
                         CURLOPT_WRITEDATA,
                         apiClient);

size_t writeDataCallback(void *buffer, size_t size, size_t nmemb, void *userp) {
    size_t size_this_time = nmemb * size;
    apiClient_t *apiClient = (apiClient_t *)userp;
    apiClient->dataReceived = (char *)realloc( apiClient->dataReceived, apiClient->dataReceivedLen + size_this_time + 1);   <== now the previous data will be reserved.
    memcpy(apiClient->dataReceived + apiClient->dataReceivedLen, buffer, size_this_time);
    apiClient->dataReceivedLen += size_this_time;
    return size_this_time;
}
```
Reference:
https://stackoverflow.com/questions/3573146/problem-using-libcurl-it-does-not-appear-to-get-the-entire-page

### PR checklist

@wing328 @zhemant @michelealbano


ityuhui:
Hi @wing328 @zhemant @michelealbano

Could you please review the code change ?  Thanks.

modules/openapi-generator/src/main/resources/C-libcurl/api-body.mustache
```diff
@@ -373,7 +373,9 @@ end:
     {{/returnTypeIsPrimitive}}
     if (apiClient->dataReceived) {
         free(apiClient->dataReceived);
+        apiClient->dataReceived = NULL;
     }
+    apiClient->dataReceivedLen = 0;
```
zhemant, line 378: Is there any specific reason to keep this outside of the if check? or can it be moved inside also? as it will be already 0 if the if statement is false. So the line becomes obsolete.

ityuhui: Yes. it can be moved inside. I will do it.

ityuhui:
Hi @wing328 

Because no more comment is received from reviewers, I think this PR is ready to merge.

wing328:
@ityuhui thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

Issue #5836, opened 2020-04-06
[BUG] C client generator cannot generate petstore label:"Client: C" 


michelealbano:
#### Bug Report Checklist

- [ X ] Have you provided a full/minimal spec to reproduce the issue?
- [ X ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ X ] What's the version of OpenAPI Generator used?
- [ X ] Have you search for related issues/PRs?
- [ X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

Current C client generator cannot generate correctly code for petstore.yaml (modules/openapi-generator/src/test/resources/3_0/petstore.yaml)

##### Description

Generate code for modules/openapi-generator/src/test/resources/3_0/petstore.yaml cannot compile. When tweaked manually, the test code does not compile.

##### openapi-generator version

Current master.

##### OpenAPI declaration file content or url

modules/openapi-generator/src/test/resources/3_0/petstore.yaml

##### Command line used for generation

java -jar ../openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g c -i ../openapi-generator/modules/openapi-generator/src/test/resources/3_0/petstore.yaml

##### Steps to reproduce

mkdir build
cd build
cmake ..
make

##### Related issues/PRs

maybe issue #5735 

##### Suggest a fix

I found a number of issues with current code. I am providing a PR.

PR #5837, opened 2020-04-06
C client generator improvements to support petstore. Solves #5836

michelealbano:
Improvement to the C client generator, driven by:
openapi-generator/modules/openapi-generator/src/test/resources/3_0/petstore.yaml

In particular:
- A .h and a .c to manage binary data, like it's done for lists
- Correct allocation of data for binary data in the model body
- Encode/decode of binary data when OPENSSL is not in use
- Correct creation of tests for binary data
- Added an "example" for patterns
- Decoration of apiKeys in apiClient, needed when a yaml has got both auth_cookie and api_key

Now the generator can support that API definition, comprising creation of test files for the models.

@ityuhui @zhemant @wing328 

### PR checklist


modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
```diff
@@ -632,7 +632,12 @@ char *base64encode (const void *b64_encode_this, int encode_this_many_bytes){
...
     (*mem_bio_mem_ptr).data[(*mem_bio_mem_ptr).length] = '\0';  //Adds null-terminator to tail.
     return (*mem_bio_mem_ptr).data; //Returns base-64 encoded data. (See: "buf_mem_st" struct).
-#endif
+#else // OPENSSL
+    char* ret = malloc(encode_this_many_bytes);
+    memcpy(ret, b64_encode_this, encode_this_many_bytes-1);
+    ret[encode_this_many_bytes] = 0;
+    return ret;
+#endif // OPENSSL
 }
```
ityuhui, line 641: Why this fake encode/decode is needed ?  If uses does not define ```-DOPENSSL```, but use the function ```base64decode``` or ```base64encode``` ,  they will not get the result that they want.

I think if the users want to use the function ```base64decode``` or ```base64encode``` defined by openapi-generator/c-libcurl , they must define -DOPENSSL in CMakefile.

modules/openapi-generator/src/main/resources/C-libcurl/model-header.mustache
```diff
@@ -11,6 +11,11 @@
 #include "../external/cJSON.h"
 #include "../include/list.h"
 #include "../include/keyValuePair.h"
+#include "../include/binary.h"
+
+char *base64encode(const void *b64_encode_this, int encode_this_many_bytes);
+
+char *base64decode(const void *b64_decode_this, int decode_this_many_bytes, int *decoded_bytes);
```
ityuhui, line 19: The two functions ```base64encode``` or ```base64decode``` are declarated in apiClient.h
So why they are needed here ?

modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
michelealbano: I didn't get it. What if `-DOPENSSL` is not defined? What do you intend to happen in that case? I didn't have it defined, and the function "returned" without issuing a `return` statement.

modules/openapi-generator/src/main/resources/C-libcurl/model-header.mustache
michelealbano: Because I don't `#include apiClient.h` everywhere. I am actually considering to move base64* to binary.h / binary.c

ityuhui: OK. I got it.
But users may want to use these 2 functions , but they do not want to include "binary.h" 
So I suggest keeping the definition and decalration of the base64* functions in apiClient.h|c, but add the declarations to binary.h

modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
ityuhui: e.g. 
If uses do not define ```-DOPENSSL```, but use the function ```base64encode```, the returned ```char *``` is still equal to the original string, not a base64 encoded string.

michelealbano: I thought that was idea, and that's the code I implemented. The original code did nothing (not even `return` a buffer).

modules/openapi-generator/src/main/resources/C-libcurl/model-header.mustache
michelealbano: ok

ityuhui: Moving to binary.h/binary.c is also acceptable for me.

zhemant: @michelealbano you dont need to include apiClient.h any where. You just need to import the main api's api.h this will import all the models required and supporting files for that API

michelealbano: Thinking about it, I will move them to binary.h/binary.c. I `#include "../include/binary.h"` in apiClient.h, thus the functions are available to whoever includes apiClient.h.

modules/openapi-generator/src/main/resources/C-libcurl/apiClient.c.mustache
ityuhui: Yes. I know.
So I suggest printing a warning message here if this fake code is needed:
e.g.
```c
#else // OPENSSL
    fprintf(stderr, "The data will not be decoded. If you want to use this function, please define \"-DOPENSSL\" when building.\n");
...
    return ret;
#endif // OPENSSL
```

michelealbano: ok, will do that

ityuhui: Message proposal: 

The data will not be encoded. If you want to use this function \"base64encode\", please define \"-DOPENSSL\" when building the library.
The base64 data will not be decoded. If you want to use this function, \"base64decode\", please define \"-DOPENSSL\" when building the library.

zhemant: I would suggest returning NULL so the whole API call exits with an error.  this is because we use strlen to get the length of the data when we add data to curl handle. This will break if the data has binary(may include null termination). So for a safer side, I would suggest to return NULL or mandate OpenSSL installation as a prerequisite.

michelealbano: What if I don't have OpenSSL, and I just want to generate unencrypted APIs?

zhemant: If you don't have OpenSSL then you cannot send or receive binary data. 

Also, I didn't get what you mean by unencrypted API? Because OpenSSL is used only for encoding binary data. It is not used for encryption.

michelealbano: Ok.

michelealbano: I updated also `CMakeLists.txt`, but I am not an expert on `cmake`.

michelealbano:
Ready to merge @wing328

PR #5892, opened 2020-04-10
Fixes to javascript-apollo generator for proper module exports, type annotations

erithmetic:


modules/openapi-generator/src/main/resources/Javascript-Apollo/api_doc.mustache
```diff
@@ -1,112 +0,0 @@
-# {{moduleName}}.{{classname}}{{#description}}
```
wing328, line 1: @erithmetic is this file removed intentionally?

PR #5894, opened 2020-04-10
[Java] Discriminator lookup should not be case insensitive by default

sebastien-rosset:
1. Case sensitivity

The OpenAPI specification explicitly states field names are case sensitive:
> All field names in the specification are case sensitive. This includes all fields that are used as keys in a map, except where explicitly noted that keys are case insensitive.

That includes discriminator values. The discriminator value lookup must be case sensitive by default. Adding a property to support case insensitive lookup.

2. Incorrect class name in generated discriminator lookup.

In the generated JSON.java file, there is a lookup function that takes a string as input and returns the name of the Java class. The map incorrectly uses the name of the Java class to register the mapping in classByDiscriminatorValue. This isn't a problem if the OAS document has a name that does not require sanitization (e.g. "Animal"), but for example "animal" would be converted to "Animal" when creating a Java class.

### PR checklist

### DIFF
--- a/samples/client/petstore/java/feign/src/main/java/org/openapitools/client/model/Animal.java
+++ b/samples/client/petstore/java/feign/src/main/java/org/openapitools/client/model/Animal.java
@@ -43,7 +43,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 
 public class Animal {
   public static final String JSON_PROPERTY_CLASS_NAME = "className";
-  private String className;
+  protected String className;
 
   public static final String JSON_PROPERTY_COLOR = "color";
   private String color = "red";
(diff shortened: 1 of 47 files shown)


sebastien-rosset:
@wing328

samples/client/petstore/java/feign/src/main/java/org/openapitools/client/model/Animal.java
```diff
@@ -43,7 +43,7 @@
 
 public class Animal {
   public static final String JSON_PROPERTY_CLASS_NAME = "className";
-  private String className;
+  protected String className;
```
sebastien-rosset, line 46: @wing328 , what causes the change from private to protected?

sebastien-rosset:
@bbdouglas  @sreeshas @jfiala @lukoyanov  @cbornet @jeff9finger  @karismann  @Zomzog @lwlee2608  @bkabrda

wing328:
@sebastien-rosset thanks for the PR. I would consider this a bug fix so eventually, we may want to remove this option (maybe in 5.x) to ensure the correct behavior moving forward.

sebastien-rosset:
> @sebastien-rosset thanks for the PR. I would consider this a bug fix so eventually, we may want to remove this option (maybe in 5.x) to ensure the correct behavior moving forward.

Agreed.

PR #5898, opened 2020-04-11
[maven] mark Mojo threadSafe=true + fix concurrency issue in CodegenConfigurator

### LINKED ISSUE #5862: [BUG] openapi-generator-maven-plugin is not marked as threadSafe
##### Description

Building a Maven multimodule project in parallel (e.g. `mvn -T2 ...`) that uses `openapi-generator-maven-plugin` yields the following warning for each plugin execution:
```
[2020-04-08T13:59:47.119Z] [WARNING] *****************************************************************
[2020-04-08T13:59:47.119Z] [WARNING] * Your build is requesting parallel execution, but project      *
[2020-04-08T13:59:47.119Z] [WARNING] * contains the following plugin(s) that have goals not marked   *
[2020-04-08T13:59:47.119Z] [WARNING] * as @threadSafe to support parallel building.                  *
[2020-04-08T13:59:47.119Z] [WARNING] * While this /may/ work fine, please look for plugin updates    *
[2020-04-08T13:59:47.119Z] [WARNING] * and/or request plugins be made thread-safe.                   *
[2020-04-08T13:59:47.119Z] [WARNING] * If reporting an issue, report it against the plugin in        *
[2020-04-08T13:59:47.119Z] [WARNING] * question, not against maven-core                              *
[2020-04-08T13:59:47.119Z] [WARNING] *****************************************************************
[2020-04-08T13:59:47.119Z] [WARNING] The following plugins are not marked @threadSafe in some-module:
[2020-04-08T13:59:47.119Z] [WARNING] org.openapitools:openapi-generator-maven-plugin:4.2.3
[2020-04-08T13:59:47.119Z] [WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[2020-04-08T13:59:47.119Z] [WARNING] *****************************************************************
```

4.2.3

not relevant

`mvn clean install -T2` (example)

##### Steps to reproduce

Build a multimodule project taht is using `openapi-generator-maven-plugin` in parallel.

##### Related issues/PRs

n/a

##### Suggest a fix

Add `threadSafe=true` to `@Mojo`: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java#L73
This _assumes_ that the plugin does handle concurrent invocations (form different modules) without concurrency issues!

famod:
Fixes #5862.

### PR checklist

---

The change to `CodegenConfigurator` looks more than it really is. I just pulled out everything that is related to actually _parse_ the file into a `synchronized` method to resolve the following concurrency issue: https://github.com/OpenAPITools/openapi-generator/issues/5862#issuecomment-611023874

I must say though, that the existing approach to mutate a singleton `ObjectMapper` seems rather dodgy.

Before my change, I was able to provoke the issue in my current project (which I am not allowed to share) after max. 30 mins via:
```
time while mvn -T2 openapi-generator:generate@generate-code -pl moduleA,moduleB -Dopenapi-generator-maven-plugin.version=4.3.1-SNAPSHOT; do :; done
```
After my change I stopped the loop after 2 hours.

### DIFF
--- a/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
+++ b/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
@@ -70,7 +70,7 @@ import com.google.common.io.Files;
  * Goal which generates client/server code from a OpenAPI json/yaml definition.
  */
 @SuppressWarnings({"unused", "MismatchedQueryAndUpdateOfCollection"})
-@Mojo(name = "generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES)
+@Mojo(name = "generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, threadSafe = true)
 public class CodeGenMojo extends AbstractMojo {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CodeGenMojo.class);
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
@@ -79,67 +79,72 @@ public class CodegenConfigurator {
     public static CodegenConfigurator fromFile(String configFile, Module... modules) {
 
         if (isNotEmpty(configFile)) {
-            ObjectMapper mapper;
+            DynamicSettings settings = readDynamicSettings(configFile, modules);
 
-            if (FilenameUtils.isExtension(configFile, new String[]{"yml", "yaml"})) {
-                mapper = Yaml.mapper();
-            } else {
-                mapper = Json.mapper();
+            CodegenConfigurator configurator = new CodegenConfigurator();
+
+            GeneratorSettings generatorSettings = settings.getGeneratorSettings();
+            WorkflowSettings workflowSettings = settings.getWorkflowSettings();
+
+            // We copy "cached" properties into configurator so it is appropriately configured with all settings in external files.
+            // FIXME: target is to eventually move away from CodegenConfigurator properties except gen/workflow settings.
+            configurator.generatorName = generatorSettings.getGeneratorName();
+            configurator.inputSpec = workflowSettings.getInputSpec();
+            configurator.templatingEngineName = workflowSettings.getTemplatingEngineName();
+            if (workflowSettings.getSystemProperties() != null) {
+                configurator.systemProperties.putAll(workflowSettings.getSystemProperties());
+            }
+            if(generatorSettings.getInstantiationTypes() != null) {
+                configurator.instantiationTypes.putAll(generatorSettings.getInstantiationTypes());
+            }
+            if(generatorSettings.getTypeMappings() != null) {
+                configurator.typeMappings.putAll(generatorSettings.getTypeMappings());
+            }
+            if(generatorSettings.getAdditionalProperties() != null) {
+                configurator.additionalProperties.putAll(generatorSettings.getAdditionalProperties());
+            }
+            if(generatorSettings.getImportMappings() != null) {
+                configurator.importMappings.putAll(generatorSettings.getImportMappings());
+            }
+            if(generatorSettings.getLanguageSpecificPrimitives() != null) {
+                configurator.languageSpecificPrimitives.addAll(generatorSettings.getLanguageSpecificPrimitives());
+            }
+            if(generatorSettings.getReservedWordMappings() != null) {
+                configurator.reservedWordMappings.putAll(generatorSettings.getReservedWordMappings());
+            }
+            if(generatorSettings.getServerVariables() != null) {
+                configurator.serverVariables.putAll(generatorSettings.getServerVariables());
             }
 
-            if (modules != null && modules.length > 0) {
-                mapper.registerModules(modules);
-            }
+            configurator.generatorSettingsBuilder = GeneratorSettings.newBuilder(generatorSettings);
+            configurator.workflowSettingsBuilder = WorkflowSettings.newBuilder(workflowSettings);
 
-            mapper.registerModule(new GuavaModule());
-
-            try {
-                DynamicSettings settings = mapper.readValue(new File(configFile), DynamicSettings.class);
-                CodegenConfigurator configurator = new CodegenConfigurator();
-
-                GeneratorSettings generatorSettings = settings.getGeneratorSettings();
-                WorkflowSettings workflowSettings = settings.getWorkflowSettings();
-
-                // We copy "cached" properties into configurator so it is appropriately configured with all settings in external files.
-                // FIXME: target is to eventually move away from CodegenConfigurator properties except gen/workflow settings.
-                configurator.generatorName = generatorSettings.getGeneratorName();
-                configurator.inputSpec = workflowSettings.getInputSpec();
-                configurator.templatingEngineName = workflowSettings.getTemplatingEngineName();
-                if (workflowSettings.getSystemProperties() != null) {
-                    configurator.systemProperties.putAll(workflowSettings.getSystemProperties());
-                }
-                if(generatorSettings.getInstantiationTypes() != null) {
-                    configurator.instantiationTypes.putAll(generatorSettings.getInstantiationTypes());
-                }
-                if(generatorSettings.getTypeMappings() != null) {
-                    configurator.typeMappings.putAll(generatorSettings.getTypeMappings());
-                }
-                if(generatorSettings.getAdditionalProperties() != null) {
-                    configurator.additionalProperties.putAll(generatorSettings.getAdditionalProperties());
-                }
-                if(generatorSettings.getImportMappings() != null) {
-                    configurator.importMappings.putAll(generatorSettings.getImportMappings());
-                }
-                if(generatorSettings.getLanguageSpecificPrimitives() != null) {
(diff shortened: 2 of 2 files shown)


famod:
Sorry, wrong user in commit message. Fixed via force push.

jimschubert:
Thanks for contributing this fix.

Unfortunately, I don't think this is a solution for the bug you've linked from. It looks like you're generating from one module (moduleA) and then in the same build you're trying to compile the generated code (moduleB). The reason the code in your PR would work is because you're locking on a static call and blocking your `generate` phase. This doesn't address thread-safety, really, but rather makes this operation blocking and kinda defeats the existing usage in multi-threaded applications.

`CodegenConfigurator` is already thread-safe, as `ObjectMapper` instances are guaranteed to be thread-safe if configured before use, and no state on this `fromFile` is shared between threads. `CodegenConfigurator` has been used in our [`batch` command](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/GenerateBatch.java#L143) since around October or November 2019.

Can you revert the change to synchronize within CodegenConfigurator and provide a full example to better evaluate whether CodegenMojo can be made to be thread-safe for your use case?

famod:
@jimschubert 

> It looks like you're generating from one module (moduleA) and then in the same build you're trying to compile the generated code (moduleB).

I should have added more context:
- moduleA is generating _and compiling_ code from its own api.json (let's call it apiA.json)
- moduleB is generating _and compiling_ code from its own api.json (let's call it apiB.json)
- _both_ modules use a JSON `configFile` from yet another module (where exactly that file resides is not really important, IMO), I even guess each module could use its _own_ JSON `configFile` and the problem would still arise

So, both module have their individual generation and compilation. Compilation does not actually contribute to the problem because the bug can be triggered (sooner or later) by just executing the openapi-plugin (concurrenty in the two modules). This does not involve compilation at all.

> `CodegenConfigurator` is already thread-safe, as `ObjectMapper` instances are guaranteed to be thread-safe if configured before use, and no state on this `fromFile` is shared between threads. 

AFAICS, yes and no.
Yes, `ObjectMapper` instances can and should be used concurrently for reading/parsing, _once such an instance is fully configured!_ See also JavaDoc of `ObjectMapper`:

>  * Mapper instances are fully thread-safe provided that ALL configuration of the
>   * instance occurs before ANY read or write calls. If configuration of a mapper instance
>   * is modified after first usage, changes may or may not take effect, and configuration
>   * calls themselves may fail.

But `CodegenConfigurator` _is_ (potentially) mutating the configuration of the singleton `ObjectMapper` it gets from `Json.mapper()` by registering modules.
`CodegenConfigurator` bit:
```java
            if (FilenameUtils.isExtension(configFile, new String[]{"yml", "yaml"})) {
                mapper = Yaml.mapper();
            } else {
                mapper = Json.mapper();
            }

            if (modules != null && modules.length > 0) {
                mapper.registerModules(modules);
            }

            mapper.registerModule(new GuavaModule());
```
`io.swagger.v3.core.util.Json` bit:
```java
    private static ObjectMapper mapper;

    public static ObjectMapper mapper() {
        if (mapper == null) {
            mapper = ObjectMapperFactory.createJson();
        }
        return mapper;
    }
```

This can happen concurrently, or to be more specific: In my case thw slower thread seems to modify the mapper while the first, faster thread already tries to read the file.

> provide a full example to better evaluate whether CodegenMojo can be made to be thread-safe for your use case?

I can try to come up with a sample project.
In the beginning I tried to reproduce the problem with a unit test but failed. Concurrency problems are often hard to provoke.

jimschubert:
I see, modifying the mapper is not ideal here. It would be better to use the copy operation and create a new instance of the desired mapper, especially since this method allows passing multiple additional modules.

If you'd like, I can make that change so we can avoid the thread lock. But, I'd still be interested in seeing a minimal full repro. I've run the batch operation on all generators hundreds of times with my configuration at 24 threads. I'd like to profile and understand why you can repro this with 2 threads through the Maven plugin while others and myself haven't had issues by other means to rule out any other issues with Maven being multi-threaded. For instance, we should see this issue in the Gradle Plugin with parallel builds if it's causing issue in Gradle. Also, although the SBT plugin is new, I've run it numerous times without issue, but I'm not sure if my SBT configuration is threaded... it's also only a single project.

famod:
I was able to reproduce the problem with a unit test:
```java
package org.openapitools.codegen.config;

import java.util.stream.IntStream;

import org.openapitools.codegen.config.CodegenConfigurator;
import org.testng.annotations.Test;

public class CodegenConfiguratorParallelTest {

    @Test
    public void test() {
        final String file = "src/test/resources/sampleConfig.json";

        IntStream.range(0, 10).parallel()
            .forEach(i -> {
                System.out.println(i + ": " + Thread.currentThread().getId());
                CodegenConfigurator.fromFile(file);
            });
    }
}
```
Just add this test to `modules/openapi-generator/src/test/java/org/openapitools/codegen/config` and run it until the error strikes:
```
while mvn jacoco:prepare-agent@jacoco-initialize surefire:test@default-test -Dtest=CodegenConfiguratorParallelTest -Djacoco.skip; do :; done
```
This optimized command just executes surefire (and jacoco because I could not figure out how to set `@{argLine}` to an empty string).
You can add `-f modules/openapi-generator` if calling running from root.

After some time the loop should break and the error should be logged. This may take more than 10 minutes!

jimschubert:
@famod thanks, I'll see if I can repro with that test as well.

Here's a diff which _should_ resolve without any locks and also fixes an additional bug where non-lowercase filenames would result in yaml being parsed as JSON (sorry, saw this while looking at this code with a critical eye).

```diff
diff --git a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
index f7e17bd8e2..840790ed4c 100644
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
@@ -124,13 +124,13 @@ public class CodegenConfigurator {
         return null;
     }
 
-    private static synchronized DynamicSettings readDynamicSettings(String configFile, Module... modules) {
+    private static DynamicSettings readDynamicSettings(String configFile, Module... modules) {
         ObjectMapper mapper;
 
-        if (FilenameUtils.isExtension(configFile, new String[]{"yml", "yaml"})) {
-            mapper = Yaml.mapper();
+        if (FilenameUtils.isExtension(configFile.toLowerCase(Locale.ROOT), new String[]{"yml", "yaml"})) {
+            mapper = Yaml.mapper().copy();
         } else {
-            mapper = Json.mapper();
+            mapper = Json.mapper().copy();
         }
 
         if (modules != null && modules.length > 0) {

```

Would you mind trying the above diff on your branch to see if it resolves?

famod:
Error after > 11 mins:
image

famod:
> Would you mind trying the above diff on your branch to see if it resolves?

Sure! First I actually wanted to use `copy()` myself but I was under the impression that this would not suffice, but I cannot remember why.

Btw, the above screenshot is from a Windows 10 machine with Git Bash but the problem first occured on our Jenkins running on CentOS7.

jimschubert:
I appreciate it. I bumped the parallel count to 50 and I ran for 18 minutes with my diff and T24 (which shouldn't make a difference here, as it should just do parallel work in the test), and had no issue. My machine completes the full build in ~5s, so that's quite a few executions of parallel runs. I'll rerun without the diff and see how long it takes to repro.

jimschubert:
ok I've run my diff for 45 minutes and your code for about 30 minutes, and neither failed. I bumped the max thread from 10 to 100 in your test and removed synchronized on your branch and the build fails immediately.

Your concerns about `.copy()` were probably the additional memory. Most folks are likely running similar scenarios to yours; up to a handful of one-off compilations and then the process is terminated. However, there are a few other places that openapi-generator is being used (APIs, SaaS, IDE plugins). For those cases, I'd rather increase the memory a little and trigger GC than lock one or more threads with synchronization.

If you're willing to apply the diff and verify the multi-threaded functionality in your build, I'd like to get this into 4.3.1.

I really appreciate you not only discovering this issue, but digging in to find a fix, and for being open to suggestions and testing.

famod:
You are welcome!

I'll try `copy()` later today.

If you have no objections, I'd keep the new private method (without `synchronized`) as this improves readability, IMO.

PS: Besides more memory consumption the copy-solution will probably spend a few more CPU cycles initially:

> Although configuration is shared, cached serializers and deserializers are NOT shared, which means that the new instance may be re-configured before use; meaning that it behaves the same way as if an instance was constructed from scratch.

But I do think `copy()` is the best option.

famod:
I pushed the copy-solution. I aborted the test loop after one hour.

famod:
Travis CI build is green (the status here is outdated) and Bitrise hit a timeout. No idea why.

famod:
@jimschubert is there anything left to do for me?

jimschubert:
@famod fixes should be good. I won't have chance to evaluate until the weekend, though. Been too busy at work this week to allow for much extracurricular open source, unfortunately.

jimschubert:
cc @OpenAPITools/generator-core-team This looks good to me. If there are no objections to this approach, I'll merge either tonight or tomorrow.

wing328:
@famod thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

PR #5914, opened 2020-04-13
[Go]: Interface definitions for api functions

### LINKED ISSUE #2294: [REQ][Golang][client] interface definitions for api functions

When using an external api in testing it is common to use a golang interface to introduce a mock or fake implementation. 

Automatically generating these interfaces along side the api definitions would make a consistent place for referencing these interfaces to facilitate unit testing for applications that would use this client.

I could create my own package that has this implementation, but the openapi spec already includes all the information I need. Furthermore, the generated client code includes the request input and output types, so it makes the most sense for the interface to be included in the same package.
 
## Additional context

A quick-and-dirty stab at implementation would look something like this, though I would be up to discuss alternatives:

```
diff --git a/modules/openapi-generator/src/main/resources/go/api.mustache b/modules/openapi-generator/src/main/resources/go/api.mustache
index fe72ace87e..c8a374575e 100644
--- a/modules/openapi-generator/src/main/resources/go/api.mustache
+++ b/modules/openapi-generator/src/main/resources/go/api.mustache
@@ -17,6 +17,35 @@ var (
        _ context.Context
 )

+type {{classname}} interface {
+{{#operation}}
+    /*
+    {{{classname}}}Service{{#summary}} {{{.}}}{{/summary}}
+    {{#notes}}
+    {{notes}}
+    {{/notes}}
+    * @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+    {{#allParams}}
+    {{#required}}
+    * @param {{paramName}}{{#description}} {{{.}}}{{/description}}
+    {{/required}}
+    {{/allParams}}
+    {{#hasOptionalParams}}
+    * @param optional nil or *{{{nickname}}}Opts - Optional Parameters:
+    {{#allParams}}
+    {{^required}}
+    * @param "{{vendorExtensions.x-exportParamName}}" ({{#isPrimitiveType}}{{^isBinary}}optional.{{vendorExtensions.x-optionalDataType}}{{/isBinary}}{{#isBinary}}optional.Interface of {{dataType}}{{/isBinary}}{{/isPrimitiveType}}{{^isPrimitiveType}}optional.Interface of {{dataType}}{{/isPrimitiveType}}) - {{#description}} {{{.}}}{{/description}}
+    {{/required}}
+    {{/allParams}}
+    {{/hasOptionalParams}}
+    {{#returnType}}
+    @return {{{returnType}}}
+    {{/returnType}}
+    */
+    {{{nickname}}}(ctx context.Context{{#hasParams}}, {{/hasParams}}{{#allParams}}{{#required}}{{paramName}} {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/required}}{{/allParams}}{{#hasOptionalParams}}localVarOptionals *{{{nickname}}}Opts{{/hasOptionalParams}}) ({{#returnType}}{{{returnType}}}, {{/returnType}}*http.Response, error)
+{{/operation}}
+}
+
 type {{classname}}Service service
 {{#operation}}
```

arvindth:
* Add golang interface definitions for api functions
  * Fix #2294 with code suggested by @partkyle
* Add new go generator option `generateInterfaces`, defaulted to false.
* Slight updates to method comments in `api.mustache` to prettify and align better.
* Update `go-petstore`, `go-petstore-withXml`, and `openapi3-go-petstore` samples.
* Verified that the following pass:
  * `mvn clean install`
  * `mvn -Pstatic-analysis clean install`
  * `mvn integration-test -rf :GoPetstore` in `samples/client/petstore/go`, both for the default `go-petstore` sample generation, and when generated with `generateInterfaces=true`.

Questions I had:
* I see that `go-petstore-withXml` has `withGoCodegenComment=true` as an additional option, so it looked like that might be where alternate options were being tested, so I've added `generateInterfaces=true` to those samples. Let me know if that should get reverted.
* It looks like the mustache template for the `go-experimental` generator differs fairly significantly from the `go` generator, and it's not clear if I should also update `go-experimental` templates/samples as well.

### PR checklist

cc go technical committee: @antihax @bvwells @grokify @kemokemo @bkabrda


modules/openapi-generator/src/test/java/org/openapitools/codegen/go/GoClientOptionsTest.java
```diff
@@ -45,10 +45,11 @@ protected void verifyOptions() {
         verify(clientCodegen).setPackageName(GoClientOptionsProvider.PACKAGE_NAME_VALUE);
         verify(clientCodegen).setWithGoCodegenComment(GoClientOptionsProvider.WITH_GO_CODEGEN_COMMENT_VALUE);
         verify(clientCodegen).setWithXml(GoClientOptionsProvider.WITH_XML_VALUE);
-        verify(clientCodegen).setWithXml(GoClientOptionsProvider.ENUM_CLASS_PREFIX_VALUE);
+        verify(clientCodegen).setEnumClassPrefix(GoClientOptionsProvider.ENUM_CLASS_PREFIX_VALUE);
```
arvindth, line 48: I didn't see any reason this repeated the `setWithXml` from the previous line, so I think this was a bug.

arvindth:
Looks like circleci failed with the assumption that `go-experimental` also needs to support the same command line options, so I'll try to update it.

bpicolo:
@arvindth: Just a spectator, but thanks for taking this on! Hit this in attempts to do some testing with openapi this week.

arvindth:
Hi @antihax @bvwells @grokify @kemokemo @bkabrda
Apologies for the many debug commits the last week while I tried to fix my circleCI failures. I should have held off on cc'ing you until after a green build.

The changes to go-petstore were fairly simple based on the original code snippet from #2294. However, the recent change to go-experimental (#4787 by @zippolyte) that added a builder pattern made it pretty difficult to enable interfaces. I had to make some breaking changes to enable `generateInterfaces` support.

Summary of changes to go-experimental:
* I've broken apart the request builder and the execute method and moved the execute method from the request to the service.
* I think that the builder pattern should only concern itself with building the object, and not acting on it, so I feel my change makes sense.
* In addition, now we only have to generate an interface for the service, which only has 2 methods: `getRequest` and `execute`. Since the goal of adding interfaces is mainly to allow mocking, it makes more sense to me to only mock these 2 methods rather than the entire request.

bkabrda:
@arvindth thanks for sending the PR. I'm not immediately opposed to this change, but I'd like you to provide more context. You state:

> I think that the builder pattern should only concern itself with building the object, and not acting on it, so I feel my change makes sense.

But why do you think that? Have you seen this as a pattern elsewhere? The builder pattern similar to what go has right now is used in the Java clients, so changing this would introduce inconsistency across languages, which I don't believe to be beneficial (unless there's a good language-based reason for it).

Additionally, your approach makes using the generated client less practical (2 lines for each API call instead of one). Could you summarize the benefits it brings? Maybe we could satisfy your goals in a different way while keeping the current approach.

arvindth:
Hi @bkabrda thanks for the review. I suppose I really had 2 reasons for this specific change choice:

1. The builder pattern I've seen pretty much everywhere looks something like:
```
var obj = new ObjectType().arg1(arg1Val).arg2(arg2Val).build()
// do something with the newly constructed obj
```
The chained request methods in the generated code are acting on the Request object, not on the Service. But `Execute`, though, at least to me, feels like it should be on the service object and not the request object, to separate out the service from the data it acts on. Just feels like the right separation of concerns.

2. As for benefits, the main reason to embark on this change was to allow for easy mocking of the call to the backend. If we leave it as it was, then it looks like the entire request would need to be mocked, which doesn't really make sense to me. Some requests have several setter methods, but only the `Execute` method calls the backend. I didn't see the need to require mocking the Request methods when it's really only `Execute` that needs to be mocked in order to unit test applications that use the generated client library.

I do see your point about diverging from the Java client, and I wouldn't be opposed to putting up a PR at some point to do the same thing there to keep it aligned.

I also see your point about needing to use 2 lines for the API instead of one, but if feels like that's the way builders should work, to separate the building from the usage. That said, I'm definitely open to any suggestions. I'm new'ish to go, so it's very possible I'm overlooking a way to achieve #2 without having to do #1.

PS. We currently use a custom fork with this interface support for the go (not experimental) generator. I'm also ok with reverting the go-experimental changes completely and having the interface support only on the main `go` one. However, I'm not sure if that's the right thing to do, especially if the intention is for go-experimental to become the main generator in the future.

arvindth:
@bkabrda any comments on my response above? Please let me know if you have any additional questions, concerns or suggestions.

CI/circle_parallel.sh
```diff
@@ -33,6 +33,12 @@ elif [ "$NODE_INDEX" = "2" ]; then
   if [[ $project_version == *"-SNAPSHOT" ]]; then
     echo "Running node $NODE_INDEX to test ensure-up-to-date"
     java -version
+    # Install golang version 1.14
+    go version
+    sudo mkdir /usr/local/go1.14
+    wget -c https://dl.google.com/go/go1.14.linux-amd64.tar.gz -O - | sudo tar -xz -C /usr/local/go1.14
```
jimschubert, line 39: Why is all this necessary?

bin/go-experimental-petstore.sh
```diff
@@ -34,6 +34,17 @@ rm -rf $STUB_DIR
...
+# Generate mocks for testing the samples
+cd samples/client/petstore/go-experimental/go-petstore/
+echo "which go = $(which go)"
+go version
+gopath=$(go env GOPATH)
+GO111MODULE=on go get github.com/golang/mock/mockgen@v1.4.3
+mockgencmd="${gopath%%:*}/bin/mockgen -source=api_pet.go -destination=mock_api_pet.go -package petstore"
+echo "Running $mockgencmd"
+$mockgencmd
+cd -
```
jimschubert, line 50: This does not belong in the generation script.

CI/circle_parallel.sh
agilob: it's probably to run checks against latest version of go, but this looks far out of scope for this projects and IMO shouldn't be here, that's another thing to remember and maintain.

jimschubert:
@arvindth I see you're still working on this. Please just ping whenever it's ready for another review.

arvindth:
Hi @jimschubert, do you think you may be able to look at this PR soon?

jimschubert:
Yeah, I'll try to take a look in the next few days. I didn't see your last ping, but the past week has been pretty full.

modules/openapi-generator/src/test/java/org/openapitools/codegen/go/GoClientOptionsTest.java
jimschubert: Yeah, apparently this was just previously running this twice:

```
verify(clientCodegen).setWithXml(true);
```

jimschubert:
I think this looks good.

It does expose the request object publicly when the new option is false, but as `Execute` is called indirectly through the request object, I think that's fine.

We'll probably want to have one or two of the samples generate with the interface option set to false, but I'll do that via another PR.

jimschubert:
> We'll probably want to have one or two of the samples generate with the interface option set to false, but I'll do that via another PR.

Oh, just realized as I went to change this that `bin/configs/go-experimental-go-petstore-oas2.yaml` is already doing this. :)

arvindth:
Thanks @jimschubert

PR #5924, opened 2020-04-14
[Python-experimental] Use DER encoding for ECDSA signatures, add parameter to configure hash algorithm

sebastien-rosset:
ECDSA signatures can be encoded using raw binary or ASN.1 encoding. This PR uses the same ASN1 encoding as the go-experimental implementation.

Add hash algorithm parameter in the configuration.py for HTTP signatures.

The IETF is still actively working on the draft and I opened an issue with the IETF HTTP working group to clarify the encoding of ECDSA signatures.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/signing.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/signing.mustache
@@ -340,7 +340,9 @@ class HttpSigningConfiguration(object):
             if sig_alg is None:
                 sig_alg = ALGORITHM_ECDSA_MODE_FIPS_186_3
             if sig_alg in ALGORITHM_ECDSA_KEY_SIGNING_ALGORITHMS:
-                signature = DSS.new(self.private_key, sig_alg).sign(digest)
+                # draft-ietf-httpbis-message-signatures-00 does not specify the ECDSA encoding.
+                # Issue: https://github.com/w3c-ccg/http-signatures/issues/107
+                signature = DSS.new(key=self.private_key, mode=sig_alg, encoding='der').sign(digest)
             else:
                 raise Exception("Unsupported signature algorithm: {0}".format(sig_alg))
         else:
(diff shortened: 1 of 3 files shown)


sebastien-rosset:
@spacether

modules/openapi-generator/src/main/resources/python/python-experimental/signing.mustache
```diff
@@ -118,6 +130,24 @@ class HttpSigningConfiguration(object):
         self.private_key_path = private_key_path
         self.private_key_passphrase = private_key_passphrase
         self.signing_algorithm = signing_algorithm
+        self.hash_algorithm = hash_algorithm
+        if signing_scheme == SCHEME_RSA_SHA256:
+            if self.hash_algorithm == None:
```
spacether, line 135: The standard python practice is to use `is None` when comparing to None.
Can you change this and the below lines to that pattern?
- 141
- 147
If a user passes in an instance of a custom class then using `self.hash_algorithm == None` could be True when we don't expect it to be per https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or

samples/openapi3/client/petstore/python-experimental/petstore_api/signing.py
```diff
@@ -126,6 +138,24 @@ def __init__(self, key_id, signing_scheme, private_key_path,
         self.private_key_path = private_key_path
         self.private_key_passphrase = private_key_passphrase
         self.signing_algorithm = signing_algorithm
+        self.hash_algorithm = hash_algorithm
```
spacether, line 141: Where is self.hash_algorithm used other than these exception raising checks?
If it is only used for exception raising, why assign it as the self.hash_algorithm instance variable?

sebastien-rosset: It's used while creating the digest.

spacether: Ah, thanks I missed that

spacether:
This looks good, waiting on the CI tests

PR #3237, opened 2019-06-28
Use io.swagger.util.Yaml instead of io.swagger.v3.core.util.Yaml to add  quotes around string value and fix compatibility with parser using yaml 1.1.

GuillaumeSmaha:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. If contributing template-only or documentation-only changes which will change sample output, be sure to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) first.
- [X] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `4.1.x`, `5.0.x`. Default: `master`.
- [X] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR
Fix #3196

In spec 1.1, some words without quotes can be implicitly converted to
 a boolean like "yes" or "on" (See https://yaml.org/type/bool.html)
In v3.core.util.Yaml with MINIMIZE_QUOTES enabled, the formatter created
 a YAML but forget to protect some words.

PRs are created to fix this in the formatter:
 - https://github.com/FasterXML/jackson-dataformats-text/pull/137
 - https://github.com/FasterXML/jackson-dataformats-text/pull/138

Until the PRs will be merged and released, this patch fix it by
re-adding quotes.

Issue #4365, opened 2019-11-04
[BUG][Openapi-yaml] Single quotes are deleted in examples


jorgerod:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

I use the generator *openapi-yaml* for merge serveral yaml files into one. The bug is that the generator remove de simple quotes in examples.

##### openapi-generator version

4.1.3, 4.2.0

##### OpenAPI declaration file content or url
Simplified input file:

```yaml
openapi: 3.0.1
info:
  title: ping test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/'
paths:
  /search:
    get:
      operationId: search
      parameters:
        - name: date
          in: query
          description: The date of the document.
          example: '2019-08-06T13:40:43.449Z'
          required: true
          schema:
            type: string
            format: date-time
      responses:
        '201':
          description: OK
```

##### Command line used for generation

 java -jar openapi-generator-cli.jar generate -g openapi-yaml -i openapi.yml

##### Steps to reproduce

Run the above command
Expected output:
```yaml
...
     parameters:
        - name: date
          in: query
          description: The date of the document.
          example: '2019-08-06T13:40:43.449Z'
...
```

Actual output:
```yaml
...
     parameters:
        - name: date
          in: query
          description: The date of the document.
          example: 2019-08-06T13:40:43.449Z
...
```

jimklo:
This may be a duplicate of #3196
Either way it appears related as the stripping of the quotes from the original YAML is causing issues.

Issue #3196, opened 2019-06-21
[BUG] OpenAPI.yaml generated returns strings without double quotes and "no" string is implicitly interpreted as boolean


GuillaumeSmaha:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

    

##### Description

openapi.yaml file generated in the project with `{{{openapi-yaml}}}` like https://github.com/OpenAPITools/openapi-generator/blob/20b8eff6e3dcec5dfe955b5dda1e4b84d8bdce44/modules/openapi-generator/src/main/resources/python-flask/openapi.mustache doesn't use double quotes around string values. 
Because in YAML specification, it seems value true/false, yes/no and on/off can be implicitly convert to boolean unless the value have double quotes.
Here the code in PyYAML: https://github.com/yaml/pyyaml/blob/4c2e993321ad29a02a61c4818f3cef9229219003/lib3/yaml/resolver.py#L170-L175

Example: The yaml generated should return `- "no"` instead of `- no` to avoid implicit convertion.

##### openapi-generator version
4.0.3-beta
It can be a regression because it was not present in swagger-codegen

##### OpenAPI declaration file content or url & Command line used for generation

I am using the online version to generate python-flask server with the following JSON [openapi.quote.json.zip](https://github.com/OpenAPITools/openapi-generator/files/3315720/openapi.quote.json.zip)

```yaml
{
  "host": "HOST",
  "info": {
    "description": "Car example",
    "title": "Car API",
    "version": "1.0.0"
  },
  "definitions": {
    "car": {
      "properties": {
        "name": {
          "description": "The name of the car.",
          "maxLength": 255,
          "type": "string"
        },
        "test": {
          "description": "Buggy enum: no becomes False",
          "enum": [
            "no",
            "great",
            "other"
          ],
          "type": "string"
        }
      },
      "required": [
        "name",
        "test"
      ]
    }
  },
  "paths": {
    "/v1/cars": {
      "post": {
[...]

PR #5933, opened 2020-04-15
[feat] Allow configuration of yaml minimize quotes

jimschubert:
There are cases where minimizing quotes results in invalid YAML. For example, an input YAML with string "1234_1234" will be converted to YAML value 1234_1234 which is an int in YAML 1.1 (https://yaml.org/type/int.html)

The only option in these cases is to either:

* Revert the option completely to always quote values
* Provide a user-customization to disable quotes minimization

This applies the latter with the assumption that this is an edge case and users who are unaffected will default to the "prettier" version.

An alternative would be to write a custom serializer for strings, and if they are in the format of of any of the valid formats defined in YAML:

```
[-+]?0b[0-1_]+ # (base 2)
|[-+]?0[0-7_]+ # (base 8)
|[-+]?(0|[1-9][0-9_]*) # (base 10)
|[-+]?0x[0-9a-fA-F_]+ # (base 16)
|[-+]?[1-9][0-9_]*(:[0-5]?[0-9])+ # (base 60)
```

Then wrap the result in quotes. That approach was not taken because of the potential for significant performance impact on very large specs, which our users are often tasked with transforming.

This is related to #3237, which wouldn't be a viable solution because it would remove support for OpenAPI 3.0 types and remove a handful of default configurations in the new YAML factory.

Users who are affected by this edge case or others resulting in the removal of quotes will be able to generate with the system property:

```
-Dorg.openapitools.codegen.utils.yaml.minimize.quotes=false
```

This will result in quote behavior similar to the Swagger YAML 1.x/2.x implementation without removing the new expectations for OpenAPI 3.x types.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
@@ -594,6 +594,7 @@ public class PythonAbstractConnexionServerCodegen extends DefaultCodegen impleme
             }
         }
 
+        generateJSONSpecFile(objs);
         generateYAMLSpecFile(objs);
 
         for (Map<String, Object> operations : getOperations(objs)) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/serializer/SerializerUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/serializer/SerializerUtils.java
@@ -2,25 +2,38 @@ package org.openapitools.codegen.serializer;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
-import io.swagger.v3.core.util.Yaml;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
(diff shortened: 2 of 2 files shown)


jimschubert:
I've removed python client/server labels and changed to both a Bug and general enhancement because multiple languages/frameworks would be affected by this edge case and the PR would unblock those. The source branch refers to python because that's where the discussion was reported.

jimschubert:
This should provide a user-facing workaround for #4365 and #3196

jimklo:
I've verified that this PR works in my use case. Thanks @jimschubert!

jimschubert:
@OpenAPITools/generator-core-team any objections to this approach?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
```diff
@@ -594,6 +594,7 @@ private void addSecurityExtensions(OpenAPI openAPI) {
             }
         }
 
+        generateJSONSpecFile(objs);
```
wing328, line 597: @jimschubert is line 598 `generateYAMLSpecFile(objs);` because we're now generating the spec in JSON?

jimschubert: I think they should both be included for consistency across all generators, tbh. We're limiting what people can extend or customize when we don't have a consistent model bound to templates.

wing328: As discussed, this is more like a proactive enhancement that the user didn't ask for. I'm ok with that and let's see if there are users asking for something similar in other generators.

PR #5945, opened 2020-04-16
[python-experimental] Minor doc update, code comments and exception handling

sebastien-rosset:
This PR is not ready, I am opening to start discussion. I'm not sure yet what needs to be fixed.

An OAS document may have a schema that does not specify the "type" attribute.
```yaml
Pet:
  type: object
  properties:
    name:
      type: string
    annotations:
      type: Annotations

Annotations:
  description: an arbitrary element that can represent any JSON schema,
     including the "null" type, an object, or an array.
     Notice the "type" attribute has not been set.
  nullable: true
```
 In that case, the generated Python `openapi_types()` function will have entries such as:
```
'name': (str,),  # noqa: E501
# without nullable: true
'annotations': (object,),  # noqa: E501
# When nullable: true is specified:
'annotations': (object, none_type,),  # noqa: E501     
```
Without this PR change, in model_utils.py, COERCION_INDEX_BY_TYPE, when the order_response_types function is invoked, it will call COERCION_INDEX_BY_TYPE[class_or_instance] where class_or_instance is 'object', and because 'object' is not in COERCION_INDEX_BY_TYPE, an exception is raised. Also, because there was not check to determine if the key exists in the COERCION_INDEX_BY_TYPE dict, the raised exception was not user-friendly.

### PR checklist


sebastien-rosset:
@spacether

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -84,7 +84,7 @@ COERCION_INDEX_BY_TYPE = {
     date: 10,
     str: 11,
     file_type: 12,   # 'file_type' is an alias for the built-in 'file' or 'io.IOBase' type.
-    object: 13,      # Any type, the 'type' attribute is not specified in a schema in a OAS document.
+    object: 13,      # The 'type' attribute is not specified in the OAS schema.
```
spacether, line 87: @sebastien-rosset can we discuss this over Skype?
I am concerned with doing it this way as many python classes inherit from the object type so if we did it this way it would allow in a ton of other class instances that we shouldn't like set, tuple, and arbitrary other classes.
Swagger/Openapi only allows specific types (string, int, float, dict, list, bytes, datetime, date, file_type, none_type)
My preference would be to instead convert these any type classes in python into a composed schema where it oneOf allows any type. As an added benefit, this could maybe be used to verify that we can have oneOf models with simple or complex types, which I'm not sure we can do.
Things like:
oneOf:
- string
- Pet
What do you think?

sebastien-rosset: > @sebastien-rosset can we discuss this over Skype?
> I am concerned with doing it this way because many python classes inherit from the object type. So if we did it this way, it would allow in a ton of other class instances that we shouldn't like set, tuple, and arbitrary other classes.
> Swagger/Openapi only allows specific types (string, int, float, dict, list, bytes, datetime, date, file_type, none_type)
> My preference would be to instead convert these any type classes in python into a composed schema where it oneOf allows any type. As an added benefit, this could maybe be used to verify that we can have oneOf models with simple or complex types, which I'm not sure we can do.
> Things like:
> oneOf:
> 
> * string
> * Pet
> 
> What do you think?

Yes I think that makes sense. Currently, when the type is "any", the generated openapi_types() function has a dict of properties to set of allowed types, and the allowed type is "object".  That currently causes a runtime exception. So instead of having "object" in the list, we would enumerate all possible types: str, bool, date, datetime, dict, float, int, list

spacether:
@sebastien-rosset what do you think about this proposal? When the generator sees an any type schema, replace it with a composed schema model with oneOf definition of every type.

sebastien-rosset:
> @sebastien-rosset what do you think about this proposal? When the generator sees an any type schema, replace it with a composed schema model with oneOf definition of every type.

Do you mean?
1. (str, bool, date, datetime, dict, float, int, list)
2. Every type defined in the OAS document
3. 1 and 2?

spacether:
> > @sebastien-rosset what do you think about this proposal? When the generator sees an any type schema, replace it with a composed schema model with oneOf definition of every type.
> 
> Do you mean?
> 
> 1. (str, bool, date, datetime, dict, float, int, list)
> 2. Every type defined in the OAS document
> 3. 1 and 2?
> 
> If 3, I don't see how at run time you will figure out what the actual type is. Since the value can be an arbitrary JSON document, it could match zero, some or all schemas in the OAS doc. There may or may not be any common discriminator. I think the best you can do is to map to primitive types, a dict or a list. In that case there may not be any schema (at least nothing explicitly defined in the OAS document).

Yes there is no way to know the type because any type is accepted.
Hmm I see your point about also accepting the other Schema types.
If we do this processing after processing all models, then we know the allowed oneOf options for this Schema, why not be more explicit here because:
- we know what is acceptable
- that list of types/classes is smaller than the full list of classes that languages contain

sebastien-rosset:
> > > @sebastien-rosset what do you think about this proposal? When the generator sees an any type schema, replace it with a composed schema model with oneOf definition of every type.
> > 
> > 
> > Do you mean?
> > 
> > 1. (str, bool, date, datetime, dict, float, int, list)
> > 2. Every type defined in the OAS document
> > 3. 1 and 2?
> > 
> > If 3, I don't see how at run time you will figure out what the actual type is. Since the value can be an arbitrary JSON document, it could match zero, some or all schemas in the OAS doc. There may or may not be any common discriminator. I think the best you can do is to map to primitive types, a dict or a list. In that case there may not be any schema (at least nothing explicitly defined in the OAS document).
> 
> Yes there is no way to know the specific type because multiple types are accepted.
> Hmm I see your point about also accepting the other Schema types. 1 + 2 could be the answer but
> but I have mixed feelings about this. A dict can be a Pet, but should that Pet payload automatically be converted to a Pet instance if it is received in any type? To me the any type feels more like Case 1 where it accepts anything but has no special enum/validation checking.
> 
> Even if we do think that the answer is (1+2) if we do this processing after processing all models, then we know the allowed oneOf options for this Schema, why not be more explicit here because:
> 
> * we know what is acceptable
> * that list of types/classes is smaller than the full list of classes that languages contain

Yes I think it should be 1. (or some variant of it); the code should not attempt to create a schema object. The payload should not be converted to a Pet just because it looks like a Pet instance.

sebastien-rosset:
Suppose the "request" property is "any type". Today the following code is generated and that causes a runtime exception:

```python
    @staticmethod
    def openapi_types():
        return {
            "event": (str,),  # noqa: E501
            "request": (object, none_type,),  # noqa: E501
        }
```

So instead we would generate the following code?
```python
    @staticmethod
    def openapi_types():
        return {
            "event": (str,),  # noqa: E501
            "request": (str, bool, date, datetime, dict, float, int, list, none_type,),  # noqa: E501
        }
```

spacether:
Yes, that would be the simplest way to do it and it's consistent with how python-experimental works.y point is that this any type definition is part of the general case of composed oneof schema with all primitive types. If we build it that way the solution will work for many generators. If we build it your suggested way, then it only works for Python-experimental. Which way do you prefer? My preference is the general way but either works.

sebastien-rosset:
> Yes, that would be the simplest way to do it and it's consistent with how python-experimental works. My point is that this any type definition is part of the general case of composed oneof schema with all primitive types. If we build it that way the solution will work for many generators. If we build it your suggested way, then it only works for Python-experimental. Which way do you prefer? My preference is the general way but either works.

I agree, it would be nice if codegen converts "any type" to oneOf (string, boolean, map, number, integer, array), and optionally "null". That way the language generator would treat it the same way as if the OAS schema had oneOf. @wing328 , what do you think?

sebastien-rosset:
I am renaming the PR, and I think we should open a separate PR in codegen for
"Adding support for OAS schema that does not have "type" attribute specified"

PR #5992, opened 2020-04-20
Fix inline schema in allOf without object type

wing328:
When an inline schema in allOf is defined without "type: object", then the inline schema will be handled differently.

This PR will fix the incorrect behavior and generate the inline schema as a separate model (the same as how an inline scheme in allOf with "type: object" is defined.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -332,11 +332,11 @@ public class InlineModelResolver {
      * Flattens properties of inline object schemas that belong to a composed schema into a
      * single flat list of properties. This is useful to generate a single or multiple
      * inheritance model.
-     * 
+     *
      * In the example below, codegen may generate a 'Dog' class that extends from the
      * generated 'Animal' class. 'Dog' has additional properties 'name', 'age' and 'breed' that
      * are flattened as a single list of properties.
-     * 
+     *
      * Dog:
      *   allOf:
      *     - $ref: '#/components/schemas/Animal'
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -390,6 +392,8 @@ private void flattenComposedChildren(OpenAPI openAPI, String key, List<Schema> c
                         listIterator.set(schema);
                     }
                 }
+            } else {
+                // likely a reference to schema (not inline schema)
```
jimschubert, line 396: Maybe we should log something here?

Have you tested with an object without type or properties defined (additionalProperties only)?

wing328: I tried to log something here - it's just a reference to a model. (e.g. Foo), which means it's not an inline schema.

wing328: > Have you tested with an object without type

Yes of course. Let me add a test case tomorrow so that we'll be alerted in case the behavior has changed.

wing328:
> Do you consider this a bug fix? It may change this behavior for people who rely on the behavior.

Yes, my reason is that with or without "type:object" should generate the same code.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
wing328: Will add test cases in a separate PR.

wing328: Added a test via https://github.com/OpenAPITools/openapi-generator/pull/6042

PR #5995, opened 2020-04-20
[typescript-fetch] Fix map API return type

haraldF:
Fix APIs that return a map.

given the following schema (yaml) as the `application/json` body of an API:

```yaml
    Foo:
      type: object
      additionalProperties:
        $ref: '#/components/schemas/Bar'
```

The generator generated the wrong code that wouldn't transpile. This PR fixes that by handling maps the same way as we handle maps in other places.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -290,7 +290,12 @@ export class {{classname}} extends runtime.BaseAPI {
         return new runtime.JSONApiResponse(response, (jsonValue) => jsonValue.map({{returnBaseType}}FromJSON));
         {{/isListContainer}}
         {{^isListContainer}}
+        {{#isMapContainer}}
+        return new runtime.JSONApiResponse(response, (jsonValue) => runtime.mapValues(jsonValue, {{returnBaseType}}FromJSON));
+        {{/isMapContainer}}
+        {{^isMapContainer}}
         return new runtime.JSONApiResponse(response, (jsonValue) => {{returnBaseType}}FromJSON(jsonValue));
+        {{/isMapContainer}}
         {{/isListContainer}}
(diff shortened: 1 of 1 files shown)


haraldF:
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov review please?

amakhrov:
@haraldF what is a problem that you're fixing with this PR? Is there a related issue with the details? If not - could you add details in the PR?

haraldF:
> @haraldF what is a problem that you're fixing with this PR? Is there a related issue with the details? If not - could you add details in the PR?

@amakhrov thanks for the quick follow up, I've added a more verbose explanation. Hope it's clear now :)

modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
```diff
@@ -290,7 +290,12 @@ export class {{classname}} extends runtime.BaseAPI {
         return new runtime.JSONApiResponse(response, (jsonValue) => jsonValue.map({{returnBaseType}}FromJSON));
         {{/isListContainer}}
         {{^isListContainer}}
+        {{#isMapContainer}}
+        return new runtime.JSONApiResponse<any>(response);
```
amakhrov, line 294: Without any `transfromer` (missing second argument for the constructor), how would it process nested objects in this map?

All other types are (recursively) processed via `<someBaseType>FromJSON()>`. This processing includes mapping between raw property names and camelCase / snake_case / whatever is configued in `modelPropertyNaming`. Also this processing includes serializing / deserializing dates

haraldF: Thank you very much for the review! Indeed I tried to take the easy way out :)

I found `runtime.mapValues` which should do exactly what we need here - construct a free-form key-value object while running a transformer over the values.

PR #6004, opened 2020-04-21
typescript-fetch: Escape operation-id if clashing with model names

### LINKED ISSUE #6002: [BUG][typescript-fetch] Name clash between models and operations
##### Description

Imagine a Model:

```yaml
    SubscribeRequest:
      type: object
      properties:
        # ...
```

and an operation called subscribe:

```yaml
paths:
  /subscriptions:
    post:
      summary: Subscribe
      operationId: subscribe
      parameters:
          # ...
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/SubscribeRequest'
        required: true
```

The generated TypeScript operation would import `SubscribeRequest` but also generate a class called `SubscribeRequest` in the same file, which leads to a compilation error:

```js
import {
    SubscribeRequest,
    SubscribeRequestFromJSON,
    SubscribeRequestToJSON,
} from '../models';

// ...

// ERROR - SubscribeRequest name already taken
export interface SubscribeRequest {
    // ...
}
```

Today's master (but also happens with latest released version)

##### Suggest a fix

One "naive" fix would be to simply escape all models that end with `Request`:

```java
    @Override
    public String toModelName(final String name) {
        String modelName = super.toModelName(name);
        if (modelName.endsWith("Request") && !modelName.startsWith("Model")) {
            modelName = "Model" + modelName;
        }
        return modelName;
    }
```

haraldF:
A model import might clash with an operation when badly named. Append a "Operation" to the name to disambiguate and make the code compil.e

Closes #6002

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
@@ -235,7 +235,22 @@ public class TypeScriptFetchClientCodegen extends AbstractTypeScriptClientCodege
         this.updateOperationParameterEnumInformation(operations);
         this.addOperationObjectResponseInformation(operations);
         this.addOperationPrefixParameterInterfacesInformation(operations);
+        this.escapeOperationIds(operations);
         return operations;
+    }
+
+    private void escapeOperationIds(Map<String, Object> operations) {
+        Map<String, Object> _operations = (Map<String, Object>) operations.get("operations");
+        List<CodegenOperation> operationList = (List<CodegenOperation>) _operations.get("operation");
+        for (CodegenOperation op : operationList) {
+            String param = op.operationIdCamelCase + "Request";
+            if (op.imports.contains(param)) {
+                // we import a model with the same name as the generated operation, escape it
+                op.operationIdCamelCase += "Operation";
+                op.operationIdLowerCase += "operation";
+                op.operationIdSnakeCase += "_operation";
+            }
+        }
     }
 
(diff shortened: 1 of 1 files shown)


haraldF:
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov review please?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptFetchClientCodegen.java
```diff
@@ -235,9 +235,24 @@ private void addNpmPackageGeneration() {
...
     }
 
+    private void escapeOperationIds(Map<String, Object> operations) {
+        Map<String, Object> _operations = (Map<String, Object>) operations.get("operations");
+        List<CodegenOperation> operationList = (List<CodegenOperation>) _operations.get("operation");
+        for (CodegenOperation op : operationList) {
+            String param = op.operationIdCamelCase + "Request";
+            if (op.imports.contains(param)) {
+                // we import a model with the same name as the generated operation, escape it
+                op.operationIdCamelCase += "Operation";
```
macjohnny, line 249: @wing328 is changing the operation id a behavior a user would expect? or does a user expect the operation id to remain unaltered?

wing328: For example, if the operationId is a reserved word (e.g. return), the output won't work/compile so we need to escape it.

We're doing the same thing in other generators to ensure the output compiles without issues by escaping operationId, property name, parameter name and more.

PR #6021, opened 2020-04-22
[C++] [Qt5] Export response code to worker

### LINKED ISSUE #6018: [REQ] [C++] [Qt5] [Client] Provide the http status code
It would be great if the generated API exposed the http status code received.
Currently only the `QNetworkReply::NetworkError` error is available which doesn't always represent the actual http status code.

This change only relates to C++ Qt5 client generator (cpp-qt5-client).

etherealjoy:
Fixes #6018
@ravinikam @stkrwork @MartinDelille @muttleyxd 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
@@ -45,7 +45,7 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
 }
 
 {{prefix}}HttpRequestWorker::{{prefix}}HttpRequestWorker(QObject *parent)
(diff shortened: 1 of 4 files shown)


ChrisAfaik:
Thanks for the quick reaction to #6018.

One little concern about the PR:
The constructor initializes the http response code to -1, which indicates that it has not been received/determined yet. However, on_manager_finished sets it to 0 in the same case (the QVariant returned is invalid and thus toInt returns 0). I suggest setting -1 here as well.

etherealjoy:
@ChrisAfaik 

> One little concern:
> The constructor initializes the http response code to -1, which indicates that it has not been received/determined yet. However, on_manager_finished sets it to 0 in the same case (the QVariant returned is invalid and thus toInt returns 0). I suggest setting -1 here as well.

I added a check on the *toInt* call

image

ChrisAfaik:
Thanks @etherealjoy ! It's looking good.

I wouldn't have used the `auto` type because it's `int` by [definition](https://doc.qt.io/qt-5/qvariant.html#toInt). But that's ok.

Thanks again and looking forward for the 4.3.1 release 👍

PR #6027, opened 2020-04-23
Typescript refactor fixes

bodograumann:
Fixes the issues identified in https://github.com/OpenAPITools/openapi-generator/issues/802#issuecomment-617262139

I'll also try to add binary file download to the fetch http client class.

CC @TiFu


bodograumann:
This is also ready to be merged from my point of view.
I know there are lots of changes, but most are small and should require no discussion or have already been discussed.
Please have a look if there is something you object to @TiFu .
The test failure is unrelated to the change. I think it comes from a temporary error in the petstore server.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -783,40 +783,40 @@ private void generateSupportingFiles(List<File> files, Map<String, Object> bundl
...
-                        }
-                    }
-
-                } else {
+                if (!ignoreProcessor.allowsFile(new File(outputFilename))) {
                     if (dryRun) {
                         dryRunStatusMap.put(outputFilename, new DryRunStatus(java.nio.file.Paths.get(outputFilename), DryRunStatus.State.Ignored));
                     }
                     LOGGER.info("Skipped generation of {} due to rule in .openapi-generator-ignore", outputFilename);
+                    continue;
```
TiFu, line 791: Does this change anything except inverting the condition and adding `continue`?

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
```diff
@@ -112,7 +112,7 @@ export class RequestContext {
     	return this.headers;
     }
     
-    public getBody(): string | FormData { 
+    public getBody(): string | FormData | undefined{
```
TiFu, line 115: Would it make sense to define a type for this? `type Body = string | FormData | undefined` or something similar?

modules/openapi-generator/src/main/resources/typescript/generators/fetch-api.mustache
```diff
@@ -1,7 +0,0 @@
-export * from './http/http';
```
TiFu, line 1: Should `generators/jquery.mustache` also be removed?

TiFu:
Thanks! Just a couple questions :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
bodograumann: No, as noted in the commit message this only improves the code style to make it easier to understand.

modules/openapi-generator/src/main/resources/typescript/generators/fetch-api.mustache
bodograumann: Ah yes definitely, guess that got lost during rebasing.

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
bodograumann: I'll have to think about that ;-)

bodograumann:
Sure, I'm always happy to see progress being made and rebasing has become second nature for me.

modules/openapi-generator/src/main/resources/typescript/generators/fetch-api.mustache
bodograumann: Ok, removed it now.

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
bodograumann: I have added a type `RequestBody`. `setBody` can also be used with it, so setting the body to `undefined` removes it.

TiFu:
Great, I finally got around to merging #6139. 

It looks like that PR might not cause any issues. I would just like to make sure that we don't revert any of the changes in #6138 (that's the pr that adds top-level exports to `fetch-api.mustache` and `jquery.mustache`, not  #6139).

Other than resolving that one conflict, is there anything else that needs to be done before this PR can be merged?

bodograumann:
I rebased and everything seems fine.
The top-level exports are kept, afaics.

I think this is ready to be merged then. Can we still include this in #6341?

TiFu:
Yeah that should be fine. Let's see what the CI says and then I'll merge this PR into `ts-refactor-master`.

bodograumann:
The test failures seem to be the same as in #6341. So I guess they are unrelated to my changes.
I can't even look at the output of circlci without giving them rw access to all my repos!?

TiFu:
Alright, both CI failures are unrelated to this PR. Merging and then I'll have a look at #6177 after the CI re-run.

Thanks a lot for all the work you put into this generator!

Issue #6044, opened 2020-04-24
[REQ] [dart-dio] Security like api key should be handled


jaumard:
##### Description
Jaguar generator is handling security in header and path automatically by exposing some methods on the main Api class to set token and api key. 

Would be really useful if dart dio was able to handle this too.

The idea is to keep security name and value as a map under the main class, then have an interceptor checking if secure is needed and if needed set the correct header/query params based on the definition.

Issue #6026, opened 2020-04-23
[BUG] [dart-dio] generated code doesn't work, deserialization issue


jaumard:
##### Description
The generation works fine but once I use the generated code I'm getting: 

```
Deserializing '{"token":"8745a41de84e7a03150da5a36a8e6a17ae50b2de5f425efd6317dd95b10492fe5e3...' to 'LoginResponse' failed due to: type 'String' is not a subtype of type 'Iterable<dynamic>' in type cast
#0      BuiltJsonSerializers._deserialize (package:built_value/src/built_json_serializers.dart:153:11)
#1      BuiltJsonSerializers.deserialize (package:built_value/src/built_json_serializers.dart:104:18)
#2      BuiltJsonSerializers.deserializeWith (package:built_value/src/built_json_serializers.dart:35:12)
#3      AuthApi.login.<anonymous closure> (package:dalk_dashboard_api/api/auth_api.dart:52:33)
#4      _RootZone.runUnary (dart:async/zone.dart:1439:54)
#5      _FutureListener.handleValue (dart:async/future_impl.dart:141:18)
#6      Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:682:45)
```

Look like the String is given to the deserializer this it expect an object not a String.

##### openapi-generator version
4.3.0

##### OpenAPI declaration file content or url
```yaml
openapi: 3.0.0
info:
  title: Dalk API v1
  description: Dalk API v1
  contact:
    email:
  license:
    name: MIT License
  version: v1
security:
  - Bearer:
      []
paths:
  /v1/auth/login:
    post:
      operationId: login
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/LoginRequest'
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/LoginResponse'
          description: "Logged successfully, use the token to your next calls"
      tags:
        - auth
      security: []
  /v1/auth/logout:
    post:
      operationId: logout
      responses:
        204:
          description: "logged successfully"
      tags:
        - auth
  /v1/auth/register:
    post:
      operationId: register
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/RegistrationData'
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/LoginResponse'
          description: "Successfully registered, use the token for your next calls"
      security: []
      tags:
        - auth

servers:
[...]

PR #6047, opened 2020-04-24
dart-dio improvements 

jaumard:
fix https://github.com/OpenAPITools/openapi-generator/issues/6044
fix https://github.com/OpenAPITools/openapi-generator/issues/6026

### PR checklist

cc @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @nickmeinhold (2019/09) @athornz (2019/12) @amondnet (2019/12)

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
@@ -88,12 +88,12 @@ class {{classname}} {
 
         {{#isListContainer}}
                 final FullType type = const FullType(BuiltList, const [const FullType({{returnBaseType}})]);
-                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(jsonDecode(response.data), specifiedType: type);
+                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(response.data is String ? jsonDecode(response.data) : response.data, specifiedType: type);
                 var data = dataList.toList();
         {{/isListContainer}}
         {{^isListContainer}}
         var serializer = _serializers.serializerForType({{{returnType}}});
-        var data = _serializers.deserializeWith<{{{returnType}}}>(serializer, jsonDecode(response.data));
+        var data = _serializers.deserializeWith<{{{returnType}}}>(serializer, response.data is String ? jsonDecode(response.data) : response.data);
         {{/isListContainer}}
 
             return Response<{{{returnType}}}>(
--- a/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
@@ -78,6 +78,9 @@ class {{classname}} {
             options: Options(
             method: '{{httpMethod}}'.toUpperCase(),
             headers: headerParams,
+            extra: {
+                'secure': [{{#authMethods}} {"type": "{{type}}", "name": "{{name}}"{{#isApiKey}}, "keyName": "{{keyParamName}}", "where": "{{#isKeyInQuery}}query{{/isKeyInQuery}}{{#isKeyInHeader}}header{{/isKeyInHeader}}"{{/isApiKey}} }{{#hasMore}}, {{/hasMore}}{{/authMethods}}],
+            },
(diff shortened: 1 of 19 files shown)


modules/openapi-generator/src/main/resources/dart-dio/api.mustache
```diff
@@ -78,19 +78,22 @@ class {{classname}} {
...
         {{#isListContainer}}
                 final FullType type = const FullType(BuiltList, const [const FullType({{returnBaseType}})]);
-                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(response.data, specifiedType: type);
+                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(jsonDecode(response.data), specifiedType: type);
                 var data = dataList.toList();
         {{/isListContainer}}
         {{^isListContainer}}
         var serializer = _serializers.serializerForType({{{returnType}}});
-        var data = _serializers.deserializeWith<{{{returnType}}}>(serializer, response.data);
+        var data = _serializers.deserializeWith<{{{returnType}}}>(serializer, jsonDecode(response.data));
```
athornz, line 96: What was the issue here? I've been using clients generated from this template without any deserialization issues.

jaumard: The issue was that out of the box nothing was working for me serializer was expecting a map and it wasn't happy with the string check #6026 for details. maybe I've done something wrong...

athornz: Ah, I think I know why now. I use a Dio Transformer in all my clients: https://github.com/flutterchina/dio_flutter_transformer

I think a better approach would be to create default transformer for the generated client that parses the json. Then it can be overidden by a custom transformer.

We probably need to check what type the response is before deserializing and throw an error if its not a map.

jaumard: I'm super new to dio and don't know transformer ^^ but from what I read yes that's why you don't have any issue, I'll change my code to use a transformer then :)

jaumard: but from the look of the doc transformers are not always used :/ 
```
/// [transformer] allows changes to the request/response data before it is sent/received to/from the server
  /// This is only applicable for request methods 'PUT', 'POST', and 'PATCH'.
  Transformer transformer;
```

jaumard: why not check if response is string to use jsonDecode on it ?

athornz: I think those docs are wrong because my GET requests are being transformed. 

Checking the response type is an ok approach as well.

jaumard: done

wing328:

wing328:
@jaumard thanks for the PR, which has been included in the v4.3.1 release: https://twitter.com/oas_generator/status/1258057660974329861

modules/openapi-generator/src/main/resources/dart-dio/api.mustache
```diff
@@ -88,12 +88,12 @@ class {{classname}} {
 
         {{#isListContainer}}
                 final FullType type = const FullType(BuiltList, const [const FullType({{returnBaseType}})]);
-                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(jsonDecode(response.data), specifiedType: type);
+                BuiltList<{{returnBaseType}}> dataList = _serializers.deserialize(response.data is String ? jsonDecode(response.data) : response.data, specifiedType: type);
```
haritonstefan, line 91: After this change, you cannot have a 'text/plain' response, as it tries to apply jsonDecode on it.

:(

jaumard: With  'text/plain'  response you shouldn't arrived here as you don't need to deserialize at all

haritonstefan: Indeed, but it was trying to deserialize a text/plain response.

PR #6048, opened 2020-04-24
Readonly properties for Swift 5

### LINKED ISSUE #6046: [REQ][SWIFT5] Readonly properties on struct models

Structs support immutability after initialization. Immutable data types, particularly value types like structs, encourage functional programming practices. By reducing shared mutable state inherit in objects with mutable properties, using immutable value types makes it less likely that UI appears out of sync with an application's data. 

Swift4/5 structs, when built without Objective-C compatibility, rather than declaring their properties with the `var` keyword, can use `private(set) var` keywords to make them only be mutated by the struct's initializer.

This solution could be enabled through an additional property that defaults to false.

The `let` keyword for properties on structs cannot be combined with default values. So, to keep the syntax consistent between properties that have default values and those that don't, use `private(set) var` instead.

iliaskarim:
Resolves #6046 

### PR checklist

CC: @jgavris @ehyche @Edubits @jaz-ah @4brunu


4brunu:
Thanks for submitting this PR.
Looks good to me 👍

4brunu:
Just one question, is there any reason why you choose ‘private(set) var’ instead of ‘let’?

iliaskarim:
> Just one question, is there any reason why you choose ‘private(set) var’ instead of ‘let’?

Good question, it's because `private(set) var` works for properties with default values whereas `let` does not, it will generate a build error.

4brunu:
You are right, good catch, thanks for explaining it.
For me this can be merged 👍

PR #6051, opened 2020-04-25
[Codegen] Convert "any type" to oneOf model

sebastien-rosset:
**Issue 1:**
In the JSON schema and OAS specifications, the array "items" attribute is NOT required. However I found out some language generators assume incorrectly that "items" is required. I am fixing the issue for:
* go-experimental
* python-experimental

https://json-schema.org/understanding-json-schema/reference/array.html

**Issue 2**
An OAS document may have a schema that does not specify the "type" attribute. For example, below notice the "type" attribute has not been set in the "annotation" property.

```yaml
Pet:
  type: object
  properties:
    name:
      type: string
    annotations:
      description: an arbitrary element that can represent any JSON schema,
         The value may be null, integer, boolean, string, number, object, or array.
```

This is equivalent to writing:
```yaml
anyOf:
  - type: 'null'
  - type: integer
  - type: string
  - type: number
  - type: object
  - type: array
```

To help with the code generation, I am proposing that **codegen** converts the "any type" to a **anyOf** composed schema. The alternative is every language generator has its own way to handle "any type", but it seems more generic to do this conversion. Potentially this could be controlled by a codegen runtime property.

### PR checklist


sebastien-rosset:
@wing328 , @jimschubert, what do you think? What is the proper way to handle "any type"? This is just a draft PR to get feedback before I go any further.

wing328:
Thanks for the PR. Just want to clarify a little bit, oneOf means only (and exactly) one type/schema is matched. If the value is 2 for example, it will match both integer and number so it will fail the match.

anyOf allows matching more than one schema defined in the anyOf schema list.

Ref: https://swagger.io/docs/specification/data-models/data-types/ => "Any Type"

sebastien-rosset:
> If the value is 2 for example, it will match both integer and number so it will fail the match.

Good point, I am changing to anyOf.

sebastien-rosset:
I don't see anything in JSON schema and OAS that states the "items" attribute is required when defining an array. However OpenAPITools does require the presence of "items", imo that's a bug.

https://json-schema.org/understanding-json-schema/reference/array.html

sebastien-rosset:
@spacether for Python-experimental bug fix.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -910,7 +910,14 @@ public String getTypeString(Schema p, String prefix, String suffix) {
         } else if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return prefix + "[" + getTypeString(inner, "", "") + "]" + fullSuffix;
+            if (inner == null) {
+                // Per JSON schema specification, the array "items" attribute is optional.
+                // When "items" is not specified, the elements of the array
+                // may be anything at all.
+                return prefix + "[bool, date, datetime, dict, float, int, list, str, none_type]" + fullSuffix;
```
sebastien-rosset, line 917: @spacether , can you help review this?

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2100,6 +2106,38 @@ public String toModelName(final String name) {
...
+            .addAnyOfItem(new IntegerSchema())
+            .addAnyOfItem(new NumberSchema())
+            .name(anyTypeName);
+        // The map keys must be strings and the values can be anything.
+        cs.addAnyOfItem(new MapSchema().additionalProperties(true));
+        // The array items can be anything.
+        cs.addAnyOfItem(new ArraySchema());
+        if (schema != null) {
+            cs.setTitle(schema.getTitle());
+            cs.setDescription(schema.getDescription());
```
sebastien-rosset, line 2136: Initially I thought I should create a single "AnyType" schema which is shared by everyone, but in reality there may be corner cases where constraints other than "type" have been specified.

spacether: But aren't those constraints specific to types? If you had minItems it would only apply to dict and array, right? My take is if they have any of those constraints then they should be fully explicit and list all types. We are just trying to cover this one super general yoy said it could be anything case. What do you think?

spacether: That way that constraint question is the headache of specific generators. I agree with you about adding this model/schema once and then using it multiple places if the writers used it multiple places.

sebastien-rosset: > But aren't those constraints specific to types? If you had minItems it would only apply to dict and array, right? My take is if they have any of those constraints then they should be fully explicit and list all types. We are just trying to cover this one super general yoy said it could be anything case. What do you think?

I am planning to handle these edge cases later. Right now just trying to make the simple case work, i.e. no OAS attribute whatsoever is defined in the OAS schema, it's really any type. Even this simple case is not so simple.

spacether:
> > If the value is 2 for example, it will match both integer and number so it will fail the match.
> 
> Good point, I am changing to anyOf.

SHouldn't it match on oneOf? Why should it be anyOf? null can never be a Dict or List.

sebastien-rosset:
> > > If the value is 2 for example, it will match both integer and number so it will fail the match.
> > 
> > 
> > Good point, I am changing to anyOf.
> 
> Shouldn't it be oneOf? Why should it be anyOf? null can never be a Dict or List.

Because 2 can be both Integer and Number.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2121,6 +2191,11 @@ public CodegenModel fromModel(String name, Schema schema) {
             return null;
         }
 
+        if (ModelUtils.isAnyTypeSchema(schema)) {
+            // "Any type" means the payload can be any type, e.g. integer, number, object, array...
+            return getAnyTypeModel(name, schema);
+        }
```
spacether, line 2197: What if we have AnyTypeSchema in:
- a Model Field
- a Model
- Inline in a response definition?
How about instead:
- pre-processing the the openapi document to check for the locations that need to be fixed
- then insert qty 1 anyTypeModel:
```
AnyType:
  anyOf:
  - type: "null"
  - type: integer
  - type: array ...
```
- link all references that require it to that $ref
Then we don't need any super specific logic like you are adding in python-experimental.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: So this will only happen if we are in an array, right?
Don't we instead need to cover the ModelUtils.isAnyTypeSchema(schema) case?
How about higher up at line 907 use:
```
Schema anyType = new Schema();
if (ModelUtils.isAnyTypeSchema(schema)):
  return prefix + "bool, date, datetime, dict, float, int, list, str, none_type" + fullSufix;
} else if ((ModelUtils.isMapSchema(p) || "object".equals(p.getType())) && 
  ...
} else if (ModelUtils.isArraySchema(p)) 
  Schema inner = ap.getItems();
  if (inner == null) {
    inner = anyType;
  }
  return prefix + "[" + getTypeString(inner, "", "") + "]" + fullSuffix;
...

```

spacether: We also need the `if (ModelUtils.isAnyTypeSchema(schema)):` in case we travel through an object(dict) to get there. In that case, our return would look like:
`{str: (bool, date, datetime, dict, float, int, list, str, none_type)}`

wing328:
> In the JSON schema and OAS specifications, the array "items" attribute is NOT required. However I found out some language generators assume incorrectly that "items" is required. I am fixing the issue for:

Just want to clarify a bit. Currently, `items` is still required:

> items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. items MUST be present if the type is array.

Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#properties

Next version of OpenAPI Spec should align better with JSON schema: https://github.com/OAI/OpenAPI-Specification/pull/1977

To model array of any type, please use the following instead:
```yaml
        array_any_value:
          items: {}
        array_any_value_with_desc:
          items: 
            description: inline any value
        array_any_value_nullable:
          items:
            nullable: true
            description: inline any value nullable
```

(I've a pending PR for better support of any type: https://github.com/OpenAPITools/openapi-generator/pull/6091)

sebastien-rosset:
> > In the JSON schema and OAS specifications, the array "items" attribute is NOT required. However I found out some language generators assume incorrectly that "items" is required. I am fixing the issue for:
> 
> Just want to clarify a bit. Currently, `items` is still required:
> 
> > items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. items MUST be present if the type is array.

Doesn't this state that if "items" is present, then items must be an object? It does not actually state that items must be present.

> 
> Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#properties
> 
> Next version of OpenAPI Spec should align better with JSON schema: [OAI/OpenAPI-Specification#1977](https://github.com/OAI/OpenAPI-Specification/pull/1977)
> 
> To model array of any type, please use the following instead:
> 
> ```yaml
>         array_any_value:
>           items: {}
>         array_any_value_with_desc:
>           items: 
>             description: inline any value
>         array_any_value_nullable:
>           items:
>             nullable: true
>             description: inline any value nullable
> ```
> 
> (I've a pending PR for better support of any type: #6091)

spacether:
But it says:
items MUST be present if the type is array.
Which means items is required for arrays.

sebastien-rosset:
> But it says:
> items MUST be present if the type is array.
> Which means items is required for arrays.

Sorry, good catch, I missed that. So to expand "any type", it would have to be something like this:

```yaml
AnyType:
  anyOf:
  - type: 'null'
  - type: integer
  - type: number
  - type: string
  - type: object
  - type: array
     items:
       $ref: '#components/schema/AnyType'
```

spacether:
Yes, I think that that will work.

sebastien-rosset:
> > But it says:
> > items MUST be present if the type is array.
> > Which means items is required for arrays.
> 
> Sorry, good catch, I missed that. So to expand "any type", it would have to be something like this:
> 
> ```yaml
> AnyType:
>   anyOf:
>   - type: 'null'
>   - type: integer
>   - type: number
>   - type: string
>   - type: object
>   - type: array
>      items:
>        $ref: '#components/schema/AnyType'
> ```

Actually, that statement was removed in the 3.1 specification, to be more inline with JSON schema. So it is no longer required to have `items` under `arrray`.

wing328:
cc @OpenAPITools/generator-core-team as it impacts default codegen

PR #6053, opened 2020-04-25
[cpp][Qt5] Add the ability to pass QNetworkAccessManager as a parameter 

alfredotg:
Problem:
Every api call created new connection to server. It happens because _OAIHttpRequestWorker_ creates new _QNetworkAccessManager_ every time.

Solution:
Passing _QNetworkAccessManager_ as parameter. It allows QT reusing an opened early connection if it possible (if server support keep alive connections and if there is no active request).
This change is optional and should not affect existing code.
Also I change connect to signal _QNetworkReply::finished_ instead of _QNetworkAccessManager::finished_. 

@ravinikam  @stkrwork @etherealjoy  @martindelille  @muttleyxd 

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-qt5-client/HttpRequest.cpp.mustache
@@ -44,12 +44,13 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
     files.append(file);
 }
 
-{{prefix}}HttpRequestWorker::{{prefix}}HttpRequestWorker(QObject *parent)
-    : QObject(parent), manager(nullptr), timeOutTimer(this), isResponseCompressionEnabled(false), isRequestCompressionEnabled(false), httpResponseCode(-1) {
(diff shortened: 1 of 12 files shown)


etherealjoy:
@alfredotg 
Many thanks for the PR,
I am wondering about the chance of a crash when the api object is deleted but the reply object is still alive because now it is managed by the manager and it could still call the slot.

alfredotg:
> @alfredotg
> Many thanks for the PR,
> I am wondering about the chance of a crash when the api object is deleted but the reply object is still alive because now it is managed by the manager and it could still call the slot.

Reply object is child of **OAIHttpRequestWorker** so it will be deleted with **OAIHttpRequestWorker** which will be deleted along with api object.

> ```c++
> void {{prefix}}HttpRequestWorker::on_manager_timeout(QNetworkReply *reply)
> ```
> 
> Need to **disconnect** in the function from the **reply*

Thanks. I will fix it.

MartinDelille:
@etherealjoy I think we could improve the code memory management thanks to [QSharedPointer](https://doc.qt.io/qt-5/qsharedpointer.html) and friends.

alfredotg:
> @etherealjoy I think we could improve the code memory management thanks to [QSharedPointer](https://doc.qt.io/qt-5/qsharedpointer.html) and friends.

Why and where we need shared point?  I thought parent-child relationships were the main way to manage memory in Qt.

PR #6060, opened 2020-04-27
Remove deprecations 5.0

### LINKED ISSUE #4976: [REQ] Standardize vendor extension naming conventions

Vendor extensions are not intuitive, and are case sensitive. They're also not fully documented on the wiki at https://github.com/OpenAPITools/openapi-generator/wiki/Vendor-Extensions

* We should standardize this to avoid confusion, probably kebab lower case (`x-objc-operationid` rather than `x-objc-operationId`).
* We may want to require that _all_ words are separated by hyphen (`x-objc-operation-id` rather than `x-objc-operationid`).

none

## Additional context

The issue is apparent in a vendor extension such as `x-mysqlSchema`. If someone mistakenly defines this as `x-mysql-schema` or  `x-mysqlschema`, the configuration will not be accepted. However, if we normalize from the current `x-mysqlSchema` to kebab-cased then lower-case this, the result is `x-mysql-schema`.

Test examples:

```
@Test
public void testVendorExtensionConventionConversion(){
    String input = "x-mysqlSchema";
    CaseFormat formatter = CaseFormat.LOWER_CAMEL;
    String output = formatter.converterTo(CaseFormat.LOWER_HYPHEN).convert(input);
    Assert.assertEquals(output, "x-mysql-schema");
}
@Test
public void testVendorExtensionConventionNoop(){
    String input = "x-mysql-schema";
    CaseFormat formatter = CaseFormat.LOWER_CAMEL;
    String output = formatter.converterTo(CaseFormat.LOWER_HYPHEN).convert(input);
    Assert.assertEquals(output, "x-mysql-schema");
}
@Test
public void testVendorExtensionUnknown(){
    String input = "x-mySQLSchema";
    List<String> supportedVendorExtensions = Arrays.asList("x-mysql-schema");
    CaseFormat formatter = CaseFormat.LOWER_CAMEL;
    String output = formatter.converterTo(CaseFormat.LOWER_HYPHEN).convert(input);
    // In implementation, we'd want to log a warning.
    Assert.assertFalse(supportedVendorExtensions.contains(output), "Invalid vendor extension should fail.");
}
```

We could easily make the change in 5.0 and maintain currently documented extensions. We'd want to wait for 5.0, though, because templates will need to be updated to lower cased vendor extensions.

jimschubert:
This removes the normalized/duplicated vendor extensions and warnings prepping users for the removal in 5.0. This will be a breaking change for any users with custom templates targeting any incorrectly cased vendor extensions.

This also changes from `systemProperties` to `globalProperties` and updates docs explaining which properties are available and the ways in which a user could apply these properties. This will be a breaking change for users passing `-D` after the generate command in the CLI, and will require Gradle users to rename the variable from `systemProperties` to `globalProperties`. Maven users defining `systemProperties` within the `configuration` node will also have to rename to `globalProperties`.

closes #4976 

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
```diff
@@ -29,10 +29,10 @@
 
 import static org.openapitools.codegen.utils.StringUtils.underscore;
 
+@SuppressWarnings("unchecked")
```
ybelenko, line 32: Don't know what this _suppress_ does, but hope it's for good reasons.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
```diff
@@ -271,12 +271,11 @@ public void processOpts() {
                 modelDescription = (modelDescription == null || modelDescription.isEmpty()) ? commentExtra : modelDescription + ". " + commentExtra;
             }
 
-            if (modelVendorExtensions.containsKey(CODEGEN_VENDOR_EXTENSION_KEY)) {
+            if (modelVendorExtensions.containsKey(VENDOR_EXTENSION_MYSQL_SCHEMA)) {
```
ybelenko, line 274: `CODEGEN_VENDOR_EXTENSION_KEY` vs `VENDOR_EXTENSION_MYSQL_SCHEMA`?
What difference does it make? :smile:
Forget it, already done.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
```diff
@@ -335,15 +334,15 @@ public void processIntegerTypeProperty(CodegenModel model, CodegenProperty prope
         String description = property.getDescription();
         String minimum = property.getMinimum();
         String maximum = property.getMaximum();
-        Boolean exclusiveMinimum = property.getExclusiveMinimum();
-        Boolean exclusiveMaximum = property.getIExclusiveMaximum();
+        boolean exclusiveMinimum = property.getExclusiveMinimum();
```
ybelenko, line 337: Didn't know that Java scalars shouldn't start with capital. Maybe you need to describe it somewhere in contributor guide. When created my first PR I used `AbstractPhpCodegen` as reference and it filled up with `String` declarations. Or maybe we need some Java linter to avoid this style changes in future.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
```diff
@@ -464,8 +461,8 @@ public void processDecimalTypeProperty(CodegenModel model, CodegenProperty prope
         } else {
             Float min = (minimum != null) ? Float.valueOf(minimum) : null;
             Float max = (maximum != null) ? Float.valueOf(maximum) : null;
-            if (exclusiveMinimum == true && min != null) min += 1;
-            if (exclusiveMaximum == true && max != null) max -= 1;
+            if (exclusiveMinimum && min != null) min += 1;
```
ybelenko, line 464: I used long declaration for readability, but it's ok. Don't like shortcuts in `conditions` because it's hard to read long `if` statements like `if (!isPropValid() && !countErrors() && !$someVar){}`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
```diff
@@ -1139,7 +1131,7 @@ public String escapeUnsafeCharacters(String input) {
      */
     public void setDefaultDatabaseName(String databaseName) {
         String escapedName = toDatabaseName(databaseName);
-        if (escapedName.equals(databaseName) == false) {
+        if (!escapedName.equals(databaseName)) {
```
ybelenko, line 1134: Same done for readability here, but it's ok.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
jimschubert: This is called boxing in Java. It'll be caught by Sonar, we just have low stats on Sonar at the moment and don't make it known that we track there.

In the past couple of months, we've also introduced some quality checks in build which would warn on boxing. I think reviewers just need to be diligent on quality review.

That's not to say this is wrong. A `Boolean` in Java is an object that wraps a primitive `boolean`. Its usually frowned upon in performance critical paths, while in others it makes the most sense because of the true/false/null triple states.  It should flag as a warning in most IDEs for a line like this one as unnecessary boxing, because the method you're calling returns `boolean` and there's no possibility of that third null state which would represent "unset".

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
jimschubert: In Java, generic types are "erased" at runtime, so if you cast like a `Map<String, Object>` to `Map<String, CodegenModel>`, some IDEs like IntelliJ will warn that the cast is unchecked. You should also see these warnings at compilation time.  You'll see these unchecked suppressions throughout our code because much of our ADT is Object typed and we extract into the concrete types. It's safe throughout because we manage the types privately, although it could fail if someone were to set a different type in an overridden method (but that would break everything anyway).

I'm hoping to improve our ADT and template binding models for 5.0 to be typed better.

jimschubert:
> What about -Dcolor and -Dlog.level? Shouldn't they be moved to --global-property?

These are not Generator properties, they're logger properties. The only thing that goes in `--global-property` are key/value pairs that maintain generation functionality, where the color options are not "generation logic".

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -734,28 +731,20 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
...
         op.vendorExtensions.put("x-operation-id", underscoredOperationId);
-        op.vendorExtensions.put("uppercase_operation_id", underscoredOperationId.toUpperCase(Locale.ROOT)); // TODO: 5.0 Remove
         op.vendorExtensions.put("x-uppercase-operation-id", underscoredOperationId.toUpperCase(Locale.ROOT));
         String vendorExtensionPath = op.path.replace("{", ":").replace("}", "");
-        op.vendorExtensions.put("path", vendorExtensionPath);  // TODO: 5.0 Remove
         op.vendorExtensions.put("x-path",vendorExtensionPath);
-        op.vendorExtensions.put("PATH_ID", pathId); // TODO: 5.0 Remove
         op.vendorExtensions.put("x-path-id", pathId);
-        op.vendorExtensions.put("hasPathParams", !op.pathParams.isEmpty()); // TODO: 5.0 Remove
         op.vendorExtensions.put("x-has-path-params", !op.pathParams.isEmpty());
```
richardwhiuk, line 739: This needs to be:

```suggestion
        op.vendorExtensions.put("x-has-path-params", hasPathParams);
```

Before we were adding hasPathParams twice, once with `!op.pathParams.isEmpty()`, and then again with `hasPathParams`. This was probably caused by a mistaken merge.

This is causing the difference in samples

wing328:
I've updated this PR to target the master instead as the master is now 5.0.0-SNAPSHOT

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/MysqlSchemaCodegen.java
jimschubert: As a static field, this property has to make sense to consumers not only within this project but for those extending the artifact.

`MysqlSchemaCodegen.CODEGEN_VENDOR_EXTENSION_KEY` doesn't explain what this key is. Although it's technically accurate, you're able to have any number of vendor extensions. The new name removes the redundant `CODEGEN_` prefix and changes `KEY` which provides no details about the public field to `MYSQL_SCHEMA` which represents what the vendor extension is (x-mysql-schema).

jimschubert:
Pinging @OpenAPITools/generator-core-team again for review. I'd like to get this into the first 5.0.0 beta release.

wing328:
Can you please resolve the merge conflicts when you've time? I will review over the weekend. Thanks.

jimschubert:
> Can you please resolve the merge conflicts when you've time? I will review over the weekend. Thanks.

done

Issue #6028, opened 2020-04-23
[BUG] [JAVA][MICROPROFILE] Error during client generation due to wrong import  addition


FuoriDiTesta:
##### Description

Generating a client  for language Java, library Microprofile, in a maven project, the compilation failed due to additional packages that are not included in microprofile dependencies. To be specific:

1) Date library are always JodaTime, even if something different is specified
2) javax.xml.bind are always imported, even if xml is set to false
3) package org.apache.cxf.jaxrs.ext.multipart  is always declared as import

Point 2 and 3 are packages that are not used in the code.

It's possible to fix these behavior, to avoid to have to add not necessary dependencies to the project? 

##### openapi-generator version
openapi-generator 4.3.0 used

##### OpenAPI declaration file content or url

To try to make it work at least locally, this is what i found in the code:

1) JodaTime:
	in modules\openapi-generator\src\main\java\org\openapitools\codegen\languages\JavaClientCodegen.java:
	* comment line 422 (importMapping.put("LocalDate", "org.joda.time.LocalDate");)
	* comment line 202-203 
		        } else if (MICROPROFILE.equals(getLibrary())) {
		            dateLibrary = "legacy";
     to disable the forcing of JodaLibraries as api, and it works perfectly, for JAVA8. Not tested for other date approach
2) XML package import always present:
In modules\openapi-generator\src\main\resources\Java\libraries\microprofile\pojo.mustache i replaced
			import javax.xml.bind.annotation.XmlElement;
			import javax.xml.bind.annotation.XmlRootElement;
			import javax.xml.bind.annotation.XmlAccessType;
			import javax.xml.bind.annotation.XmlAccessorType;
			import javax.xml.bind.annotation.XmlType;
			import javax.xml.bind.annotation.XmlEnum;
			import javax.xml.bind.annotation.XmlEnumValue;
			
	with
			{{#withXml}}	 
			import javax.xml.bind.annotation.XmlElement;
			import javax.xml.bind.annotation.XmlRootElement;
			import javax.xml.bind.annotation.XmlAccessType;
			import javax.xml.bind.annotation.XmlAccessorType;
			import javax.xml.bind.annotation.XmlType;
			import javax.xml.bind.annotation.XmlEnum;
			import javax.xml.bind.annotation.XmlEnumValue; 
			{{/withXml}}
			
3) For cxf multipart:
In modules\openapi-generator\src\main\resources\Java\libraries\microprofile\api.mustache removed 
[...]

PR #6062, opened 2020-04-27
Fix issue: Microprofile dateLibrary java8 not working

Ravisankar-Challa:
This fixes the issue reported here https://github.com/OpenAPITools/openapi-generator/issues/6028
setting dateLibrary to java8 is generating joda time library imports.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
@@ -199,7 +199,7 @@ public class JavaClientCodegen extends AbstractJavaCodegen
     public void processOpts() {
         if ((WEBCLIENT.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) || NATIVE.equals(getLibrary())) {
             dateLibrary = "java8";
-        } else if (MICROPROFILE.equals(getLibrary())) {
+        } else if (MICROPROFILE.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) {
             dateLibrary = "legacy";
         }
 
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
```diff
@@ -199,7 +199,7 @@ public void addOperationToGroup(String tag, String resourcePath, Operation opera
     public void processOpts() {
         if ((WEBCLIENT.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) || NATIVE.equals(getLibrary())) {
             dateLibrary = "java8";
-        } else if (MICROPROFILE.equals(getLibrary())) {
+        } else if (MICROPROFILE.equals(getLibrary()) && "threetenbp".equals(dateLibrary)) {
```
wing328, line 202: Is it correct to say that MICROPROFILE only supports `legacy` datetime library?

If yes, what about throwing a warning using LOGGER.warn to inform the user on that.

Ravisankar-Challa: The reported issue is setting dateLibrary=java8 or dateLibrary=native microprofile rest client is generating class files with Joda time because at the line 424 it has been hard coded to org.joda.time.LocalDate. So I removed line 424. Next coming to line number 202 if the user doesn't provide any dateLibrary we need to default it to native which is java.util.Date for microprofile rest client. If the user doesn't select dateLibrary open api default to threetenbp. So we are making a check at line 202 any forcing dateLibrary to native if it is microprofile & threetenbp. If user selects dateLibrary='java8' or 'joda' or 'native' this condition at line 202 will not pass and what ever user defined dateLibrary is used.

wing328: Tested with these and the results are good.

PR #6065, opened 2020-04-27
Create python abstract codegen

mwilby:
### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
@@ -1,0 +1,453 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openapitools.codegen.languages;
+
+import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.Schema;
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.openapitools.codegen.*;
+import org.openapitools.codegen.utils.ModelUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.*;
+
+import static org.openapitools.codegen.utils.StringUtils.camelize;
+import static org.openapitools.codegen.utils.StringUtils.underscore;
+
+public class PythonAbstractCodegen extends DefaultCodegen implements CodegenConfig {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PythonAbstractCodegen.class);
+
+    // nose is a python testing framework, we use pytest if USE_NOSE is unset
+    public static final String USE_NOSE = "useNose";
+
+    protected String packageName;
+    protected String packageVersion;
+    
+    protected Map<Character, String> regexModifiers;
+    protected boolean useNose = Boolean.FALSE;
+    
+    protected String pythonSrcRoot;
+
+    public PythonAbstractCodegen(String templateDirectory) {
+        super();
+
+        modelPackage = "models";
+
+        languageSpecificPrimitives.clear();
+        languageSpecificPrimitives.add("int");
+        languageSpecificPrimitives.add("float");
+        languageSpecificPrimitives.add("list");
+        languageSpecificPrimitives.add("dict");
+        languageSpecificPrimitives.add("bool");
+        languageSpecificPrimitives.add("str");
+        languageSpecificPrimitives.add("datetime");
+        languageSpecificPrimitives.add("date");
+        languageSpecificPrimitives.add("file");
+        languageSpecificPrimitives.add("object");
+        languageSpecificPrimitives.add("byte");
+        languageSpecificPrimitives.add("bytearray");
+
+        typeMapping.clear();
+        typeMapping.put("integer", "int");
+        typeMapping.put("float", "float");
+        typeMapping.put("number", "float");
+        typeMapping.put("long", "int");
+        typeMapping.put("double", "float");
+        typeMapping.put("array", "List");
+        typeMapping.put("map", "Dict");
+        typeMapping.put("boolean", "bool");
+        typeMapping.put("string", "str");
+        typeMapping.put("date", "date");
+        typeMapping.put("DateTime", "datetime");
+        typeMapping.put("object", "object");
+        typeMapping.put("file", "file");
+        typeMapping.put("byte", "bytearray");
+        typeMapping.put("ByteArray", "bytearray");
+        // map uuid to string for the time being
+        typeMapping.put("UUID", "str");
+        typeMapping.put("URI", "str");
+
+
+        // clear import mapping (from default generator) as it does not contain any
+        // Python specific mappings.
+        importMapping.clear();
+        //
+        // Add the type mappings
+        importMapping.put("List", "from typing import List");
+        importMapping.put("Dict", "from typing import Dict");
+        importMapping.put("date", "from datetime import date");
+        importMapping.put("datetime", "from datetime import datetime");
+
+        // from https://docs.python.org/3/reference/lexical_analysis.html#keywords
+        setReservedWordsLowerCase(
+                Arrays.asList(
+                        // @property
+                        "property",
+                        // python Keywords
+                        "and", "del", "from", "not", "while", "as", "elif", "global", "or",
+                        "with","assert", "else", "if", "pass", "yield", "break", "except",
+                        "import", "class", "in", "raise", "continue", "finally", "is",
+                        "return", "def", "for", "lambda", "try", "None", "True", "False",
+                        "nonlocal", "async", "await",
+                        // python reserved words
+                        "print", "exec", "self"));
+
+        modelTemplateFiles.put("model.mustache", ".py");
+
+        /*
+         * Template Location.  This is addImportthe location which templates will be read from.  The generator
+         * will use the resource stream to attempt to read the templates.
+         */
+        embeddedTemplateDir = templateDir = templateDirectory;
+
+        regexModifiers = new HashMap<Character, String>();
+        regexModifiers.put('i', "IGNORECASE");
+        regexModifiers.put('l', "LOCALE");
+        regexModifiers.put('m', "MULTILINE");
+        regexModifiers.put('s', "DOTALL");
+        regexModifiers.put('u', "UNICODE");
+        regexModifiers.put('x', "VERBOSE");
+    }
+
+    public void setUseNose(String val) {
+        this.useNose = Boolean.valueOf(val);
+    }
+
+    public String packagePath() {
+        return packageToPath(packageName);
+    }
+
+    protected static String packageToPath(String pkg) {
+        return pkg.replace('.', File.separatorChar);
+    }
+
+    protected static String dropDots(String str) {
+        return str.replaceAll("\\.", "_");
+    }
+
+    @Override
+    public String toApiTestFilename(String name) {
+        return "test_" + toApiFilename(name);
+    }
+
+    /**
+     * Escapes a reserved word as defined in the `reservedWords` array. Handle escaping
+     * those terms here.  This logic is only called if a variable matches the reserved words
+     *
+     * @return the escaped term
+     */
+    @Override
+    public String escapeReservedWord(String name) {
+        if (this.reservedWordsMappings().containsKey(name)) {
+            return this.reservedWordsMappings().get(name);
+        }
+        return "_" + name; // add an underscore to the name
+    }
+
+    @Override
+    public String getTypeDeclaration(Schema p) {
+        if (ModelUtils.isArraySchema(p)) {
+            ArraySchema ap = (ArraySchema) p;
+            Schema inner = ap.getItems();
+            return getSchemaType(p) + "[" + getTypeDeclaration(inner) + "]";
+        } else if (ModelUtils.isMapSchema(p)) {
+            Schema inner = ModelUtils.getAdditionalProperties(p);
+            return getSchemaType(p) + "[str, " + getTypeDeclaration(inner) + "]";
+        }
+        return super.getTypeDeclaration(p);
+    }
+
+    @Override
+    public String getSchemaType(Schema p) {
+        String schemaType = super.getSchemaType(p);
+        String type = null;
+        if (typeMapping.containsKey(schemaType)) {
+            type = typeMapping.get(schemaType);
+            if (languageSpecificPrimitives.contains(type)) {
+                return type;
+            }
+        } else {
+            type = toModelName(schemaType);
+        }
+        return type;
+    }
+
+    @Override
+    public String toVarName(String name) {
+        // sanitize name
+        name = sanitizeName(name); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.
+
+        // remove dollar sign
+        name = name.replaceAll("$", "");
+
+        // if it's all uppper case, convert to lower case
+        if (name.matches("^[A-Z_]*$")) {
+            name = name.toLowerCase(Locale.ROOT);
+        }
+
+        // underscore the variable name
+        // petId => pet_id
+        name = underscore(name);
+
+        // remove leading underscore
+        name = name.replaceAll("^_*", "");
+
+        // for reserved word or word starting with number, append _
+        if (isReservedWord(name) || name.matches("^\\d.*")) {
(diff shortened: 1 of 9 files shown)


wing328:

mwilby:
Thanks. I think I fixed it, but I am not interested in a video score, so I am not worried about it.

What does worry me is the failed checks.
One was I violated your "no tabs" policy, which I think I fixed.
But there is still an error

```
#!/bin/bash --login
docker pull swaggerapi/petstore

Using default tag: latest
latest: Pulling from swaggerapi/petstore

error pulling image configuration: received unexpected HTTP status: 502 Bad Gateway

Exited with code exit status 1
```
 Some sort of server error and I have no idea how to fix this.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
```diff
@@ -0,0 +1,454 @@
...
+        typeMapping.put("URI", "str");
+
+
+        // clear import mapping (from default generator) as it does not contain any
+        // Python specific mappings.
+        importMapping.clear();
+        //
+        // Add the type mappings
+        importMapping.put("List", "from typing import List");
+        importMapping.put("Dict", "from typing import Dict");
```
spacether, line 95: How does this work for python 2.7?
Don't we currently support both 2.7 and 3+? in the near future there is a plan to drop 2.7
What are your thoughts here @wing328 @jimschubert should 5.0.0 drop python 2.7 support?

The Python and python-experimental generators support python 2.7
How about defining these python 3 imports only in generators which exclusively produce python3 code?

mwilby:
Strictly speaking python 2 is supposed to die this year. I know it wont,
but, I can hope.

However, I am not that mean and I was not planning to abandon it.
According to

https://mypy.readthedocs.io/en/stable/python2.html

in theory it works for python 2

"To run your program, you must have the |typing| module in your Python 2
module search path. Use |pip install typing| to install the module. This
also works for Python 3 versions prior to 3.5 that don’t include
|typing| <https://docs.python.org/3/library/typing.html#module-typing>
in the standard library."

I say in theory because, I admit, I have not tested it.

This approach to typing has been in the server versions of the python
generators and nobody seems to be complaining about it. Its possible
nobody is using python2, so its possible its not indicative of it
working, but its also likely someone is.

Having said all this, its not my call, its just a suggestion. I took
this part of the design from the abstract server code, not the python
generator. This was done because in the python generator , the types are
wrapped in quotes, so its safe to assume the change will not effect any
of the current users and it was an approach that had been well tested. I
agree there is an issue when we take the quotes off, (which I have
tested for python3).

spacether:
> Strictly speaking python 2 is supposed to die this year. I know it wont, but, I can hope. However, I am not that mean and I was not planning to abandon it. According to https://mypy.readthedocs.io/en/stable/python2.html in theory it works for python 2 "To run your program, you must have the |typing| module in your Python 2 module search path. Use |pip install typing| to install the module. This also works for Python 3 versions prior to 3.5 that don’t include |typing| <https://docs.python.org/3/library/typing.html#module-typing> in the standard library." I say in theory because, I admit, I have not tested it. This approach to typing has been in the server versions of the python generators and nobody seems to be complaining about it. Its possible nobody is using python2, so its possible its not indicative of it working, but its also likely someone is. Having said all this, its not my call, its just a suggestion. I took this part of the design from the abstract server code, not the python generator. This was done because in the python generator , the types are wrapped in quotes, so its safe to assume the change will not effect any of the current users and it was an approach that had been well tested. I agree there is an issue when we take the quotes off, (which I have tested for python3).

Ah, thank you for sharing that background info. My mistake, I didn't realize that the typing was included in Python 2.7.6. So that solves the issue for List and Dict, etc.
If we needed it for earlier versions of python 2.7 we could add typing to setup.py/requirements.txt
What about python3 only bytes and bytearray?
Should these be overridden in modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
where we handle those types as str?

mwilby:
Forgot to mention I switch the  `bytearray` and `byte` back to strings in `PythonClientCodegen`, so it should work as before.

mwilby:
You could build a fag into python-experimental generator to switch the type defs and libraries between version specific cases. I personally would not back port it into the python generator, unless it turns out you need it in a earlier release. Anyone using it currently is expecting a specific behaviour regardless of python version. Changing the types is safe because they don't really get used. It will only become an issue on the next major release, when I am assuming you plan to switch to the experimental version..

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
spacether: This works in python2.7 per our conversation

spacether: This work in python python 2.7.18 and later only
@mwilby can you add an import that works for python 2.7 versions less than 18?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -392,8 +401,13 @@ private void fixModelImports(Set<String> imports) {
...
 
-                // fix the imports that each model has, change them to absolute
-                fixModelImports(cm.imports);
+                /*
+                 * Moved this functionality into toModelImport.
+                 * 
+                 * Now the imports are one level higher, so the template has also been altered.
+                 */
+                //// fix the imports that each model has, change them to absolute
+                //fixModelImports(cm.imports);
```
spacether, line 410: Can you delete this comment. We don't need it going forward.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -286,10 +286,19 @@ public String toDefaultValue(Schema p) {
...
+            modelImport = name;
+        } else {
+            String[] moduleStruct = name.split("\\.");
+            // https://exceptionshub.com/circular-or-cyclic-imports-in-python.html
+            modelImport = "    from " + modelPackage() + " import " + moduleStruct[0];
+            modelImport = "try:\n" +
+                                modelImport +
+                          "\nexcept ImportError:\n    " +
+                                moduleStruct[0] + " = sys.modules[\n        '" + modelPackage() + "." + moduleStruct[0] + "']";
+        }
```
spacether, line 300: Can we replace this code with robustImport?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
mwilby: Obviously deleting it is not a problem. But I left it in on purpose. I tried my best not to interfere with the experimental client, as I consider its not my place to interfere. I left this comment in to make it clear to the participants of that sub project, the nature of a speculative change that could effect them. This type definition here is an experimental proposal on top of an experiments design. It was not clear to me what the long term strategy was, nor if this change should be included here, or not.
You could could still choose for the new generator to go in a different direction. For me the fact that this PR effects the experimental generator make the decision to keep it, or not ambiguous.
So, I will leave this for you to confirm, or not.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
mwilby: I did it this way so I changed as little of the original code as possible, but I agree it would be best to use `robustImport` if it was going to be kept.
I an not certain now, but I think I thought at the time I had to modify `robustImport` to make it work and I didn't want to alter any of the code (see response to your previous comment.
Again, your call. Tell me what you want.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: The experimental generator can have breaking changes in it.
If we added comments like this when we made breaking changes we would have too many comments in the code base. I appreciate the thought, but feel that it adds unnecessary documentation to the code base. Please delete it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: It looks like robustImport may now be redundant with your update.
Let's remove robustImport and just use toModelImport at all locations if we can.
If we can do that, let's delete robustImport and fixModelImports if we can.
Or if you want to go the minimal changes route, how about not changing toModelImport?
My preference is to do the suggested clean up because it looks like we may be able to remove robustImport and fixModelImports.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
mwilby: OK, on the list

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
mwilby: OK, I will look through the experimental module and clean up the imports properly. I am a bit stretched at the moment, so it wont be for a couple of days, sorry.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
mwilby: This would be slightly problematic and a little bit strange. As far as I know the typing backport goes back at lest to 2016, so it should predate 2.7.18.

If not, in theory we would need a flag on the python version, but I don't think its necessary in the current python generator. In this case the types are represented as strings and the import is not included in the template, so the generated code will still be valid.

However, it is still an issue for the experimental version. I will look into it when I have a chance.

spacether: When I tested it locally, typing does not exist in python 2.7.15 per:
```
Python 2.7.15 (default, Oct  2 2018, 11:47:18) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import typing
  File "<stdin>", line 1, in <module>
ImportError: No module named typing
```
Thank you!

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: No worries. We can wait. Of if you want to go the quicker route you could revert the change to this function.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractCodegen.java
mwilby: OK. I have cross checked and currently the python generator only represents the types as strings. The associated templates do not use the import strings. This means the code related to types will work in earlier version of 2.7, without the typing library. Of course this does not help with the experimental version.
I found documentation going as far back a 2.7.6 for using the typing library, but that just means it did exist at one time. The problem with unsupported systems, is that they decay exponentially and this may be one of those cases. The typing library was not shipped with the core system for 2.7 (I think it may be in 2.7.18, but I am not sure), it has to be pip installed and if the support libraries are not there any more, then its not going to work.
Being very pragmatic, I would be inclined to say support in 2.7.18 is sufficient. Any software that is still working on a earlier version is likely to fail because of other libraries, so it becomes a legitimate question of how much effort goes into supporting something that is likely to fail for other reasons.
I agree by saying this I am trying to weasel out of the problem, but it doesn't make it any the less true.

wing328:
@mwilby thanks for the PR. Can you please resolve the merge conflicts when you've time?

mwilby:
Probably, its best to abandon the PR and start again at some point. The
conflicts are fairly extensive now and to be certain things are OK, its
best to start the refactoring process from scratch.

There are a mix of things in this proposed change. A commitment to a
certain view of Type specification. A restructuring of the import
assignment to the data bundle sent to the template engine. And finally,
a refactoring of of the python codgen module merged with the abstract
server module to generate a common base definition for python based
generators. The problem being it is related to a large number of
generators, including your experimental-python project. Only the last 2
are linked and all elements effect a broad range of systems, so the
change is not easily accepted and must be done very quickly, to avoid
the conflicts getting out of hand as they have now.

If you were to attempt this again, it requires a tiny bit of project
planning.

The type change is a very simple change and can be put in as a separate PR.

The import library is two parallel changes, one in the python codgen
module and one in the abstract server module, but essentially the same
code in both cases and in addition could lead to a change in the import
process in experimental-python, so that change needs to be monitored by
them.

Both of these need doing before the refactoring is done.

I have also realised the refactoring into a common base class is not
really something someone like me can do. It has to be someone inside the
project with write access and can make the changes and commit them
almost immediately.

To make this work, it needs a degree of common planning and commitment.
The tasks need to be agreed before hand and assigned and then confirmed
in order. So its not really the work of outside contributor.

wing328:
@mwilby thanks for starting this. I'll see if I can find the time later to submit a PR later to create the AbstractPythonCodegen class. 

Closing this one for the time being.

wing328:
@mwilby FYI. I've filed https://github.com/OpenAPITools/openapi-generator/pull/8817 to create the abstract python base class.

mwilby:
I had a quick look and I have one comment that matters to us and a one
that is irrelevant.

The important comment is that although you have not removed the python
`typing` module definitions of type from
`AbstractPythonConnexionServerCodegen.java`. Your in line comment
suggests you are considering it. I would urge you not to. It probably
wont matter in the long run as a work around can always be found, but
typing is becoming more of an issue with python. The reason we use
python, is that most of the code we import into our servers comes from
machine learning projects, which at the moment are predominantly python.
I am fairly sure we are not uncommon in this respect. The interaction of
the language with the base libraries, which are usually statically typed
is and always was an issue. I have no idea how its going to pan out, but
the python language community as a whole is going to come up with a
solution based on an evolution of its existing strategy of Type
Annotations and type definitions. Using anything else is creating a
potential problem that is not necessary.

The python server generators have used the typing module for as long as
we have been using the system, so why change it? The python client code
did not really have type information in it, so when we finally had to
build a client in python we discovered it didn't work for us, so we
didn't use it and still don't. Its fairly safe to assume that anyone who
has used it in the past does not care about the type definition. If it
was my choice I would work with the solution proposed by the language,
for the reasons I stated above, but I don't see any reason  why you
can't have your own approach when effectively starting from scratch. But
I reiterate the server libraries are not starting from scratch.

The irrelevant comment. The `DefaultCodegen` has a really elegant
solution for mapping composite types to import libraries. So why not use it?

For example in the constructor of
`AbstractPythonConnexionServerCodegen.java` you could add

```python
        importMapping.clear();
        //
        // Add the type mappings
        importMapping.put("List", "from typing import List");
        importMapping.put("Dict", "from typing import Dict");
        importMapping.put("date", "from datetime import date");
        importMapping.put("datetime", "from datetime import datetime");
```

If I remember correctly you have to make sure that
`postProcessAllModels` actually does put the libraries in the right
place, because the base class adds them to "imports", not "pyImports"

wing328:
Thanks for the review and feedback.

> The python client code
did not really have type information in it, so when we finally had to
build a client in python we discovered it didn't work for us, so we
didn't use it and still don't.

My understanding is that we want to add typing supporting in the python client. Just that no one has found the time to make the contribution yet.

> The irrelevant comment. The `DefaultCodegen` has a really elegant
solution for mapping composite types to import libraries. So why not use it?

I didn't know either. I wasn't the one who created it. We welcome PR to refactoring to leverage the mapping between  composite types and import libraries in the Default Codegen.

mwilby:
Sorry, I think I gave you the wrong impression. Let me clarify why I
made the comment, so you can better understand the reason for our
request. The new Python client code does have typing in it, although I
am not certain of the details, as I have not looked at it since its days
as an experimental platform.

However, the new typing does not use the python typing module. This, I
think, is because it was decided to support Python 2.7.x. (I think this
was not the right policy, but there are arguments either way.)

The old Python client didn't really support types. We use them because
we have our own serialization protocol that's designed to support
communications over low level hardware links to micro-controllers, such
as UART, I2C, SPI, etc. and we use the schema modules that are generated
to act as protocol converters. We could modify the tools source code,
but its much more stable to just modify the template. So we ended
building a work around to use the server generators to build the schema
modules and use the tool as you distribute it, But this use case is
probably unique and I seriously doubt if anybody else even noticed the
issue about types. Admittedly, this is used mostly in Java, but we have
some Python systems that use it.

So no, there is nothing wrong with the current Python client, but
because it does not use the typing module definitions its no use to us.

Hence, our request that the server versions continue to use the typing
module.

As for the irrelevant comment, that is a small and orthogonal change,
that should not effect any other ongoing work, or require extra testing,
so I think its something we could do. However, the change adds more
functionality in the  the direction of using the typing module for the
python server generators. This is a policy decision, if the team
managing the Python development want to go in different direction, then
there is no point in doing it. If you decide that this is the direction
you want to go in, let me know and we can put together a PR for you.

One last comment, I don't know who wrote DefaultCodegen either, but it
has some very well thought out processes in it. It would be good to have
a summary document of all the good feature as well as the bad ones, so
language developers know whats available to them.

PR #6121, opened 2020-04-30
[Python-experimental] JSON schema 'null' type should be modeled as 'none_type'

sebastien-rosset:
It looks like recently something changed because the JSON schema `null` type is now causing problems. I'm not sure what specific commit caused this problem. The `getTypeString` function in PythonClientExperimentalCodegen.java is incorrectly returning `null` when the input Schema is the Null type.

For example, with the following schema, the generated python-experimental code has syntax errors. 

```yaml
Foo:
  oneOf:
    - type: 'null'
    - '#/components/schemas/Bar'
```

Without this PR, the following import is incorrectly generated:
```python
try:
    from openapi_client.models import null
except ImportError:
    null = sys.modules[
        'openapi_client.models.null']
```

### PR checklist


sebastien-rosset:
@spacether , I'm not sure if I should handle the OAS `null` type in the `getTypeString` function (as proposed in this PR), or by adding the "null" type in the `typeMapping` field.

spacether:
I think that we need to add the type mapping but I have not used that before. For Python-experimental we should point it to our already defined none_type.

sebastien-rosset:
> I think that we need to add the type mapping but I have not used that before. For Python-experimental we should point it to our already defined none_type.

ok, sure, let me try with typeMappings.

samples/openapi3/client/petstore/python-experimental/petstore_api/models/fruit_req.py
```diff
@@ -39,6 +39,11 @@
 except ImportError:
     banana_req = sys.modules[
         'petstore_api.models.banana_req']
+try:
```
sebastien-rosset, line 42: I need to fix this such that this import is not generated.

samples/openapi3/client/petstore/go-experimental/go-petstore/api/openapi.yaml
```diff
@@ -1973,6 +1973,7 @@ components:
           type: string
     fruitReq:
       oneOf:
+      - type: "null"
```
sebastien-rosset, line 1976: I've just added this "null" type here to have one scenario with the "null" type in OneOf. @spacether , are you ok with this?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -388,7 +388,9 @@ private void fixModelImports(Set<String> imports) {
                 composedSchemaSets.add(cm.oneOf);
                 for (Set<String> importSet : composedSchemaSets) {
                     for (String otherModelName : importSet) {
-                        cm.imports.add(otherModelName);
+                        if (!"none_type".equals(otherModelName)) {
```
sebastien-rosset, line 391: @spacether , this is one way to ensure the import below is not generated, but it's brittle. Is there a better way to skip these special case imports?

```python
try:
    from openapi_client.models import null
except ImportError:
    null = sys.modules[
        'openapi_client.models.null']
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -388,7 +388,9 @@ private void fixModelImports(Set<String> imports) {
                 composedSchemaSets.add(cm.oneOf);
                 for (Set<String> importSet : composedSchemaSets) {
                     for (String otherModelName : importSet) {
-                        cm.imports.add(otherModelName);
+                        if (!languageSpecificPrimitives.contains(otherModelName)) {
```
sebastien-rosset, line 391: This is to avoid creating an import for "none_type".

samples/openapi3/client/petstore/go-experimental/go-petstore/api/openapi.yaml
spacether: Yup, I am okay adding this too our sample spec

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1043,6 +1050,13 @@ def get_anyof_instances(self, model_args, constant_args):
         return anyof_instances
 
     for anyof_class in self._composed_schemas()['anyOf']:
+        if anyof_class is none_type:
+            # The only way the input data can match the 'None' type
+            # is if the input data is the 'null' value.
+            if model_args is None:
+                # The input data matches the 'None' type.
+                oneof_instances.append(model_args)
```
spacether, line 1058: Did you mean anyof_instances.append..?
Also, I don't think that we should add model_args to anyof_instances.
anyof_instances is meant to store ModelSimple or ModelComposed instances, not None.
Let's just delete this if clause and continue.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1043,6 +1050,13 @@ def get_anyof_instances(self, model_args, constant_args):
         return anyof_instances
 
     for anyof_class in self._composed_schemas()['anyOf']:
+        if anyof_class is none_type:
+            # The only way the input data can match the 'None' type
+            # is if the input data is the 'null' value.
+            if model_args is None:
```
spacether, line 1056: If model_args can be None please update our docstrings to show that
- here +
- get_oneof_instance
- any calling function that passes it in

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
sebastien-rosset: Yes, good catch, I copy/pasted.

sebastien-rosset: > Did you mean anyof_instances.append..?
> Also, I don't think that we should add model_args to anyof_instances.
> anyof_instances is meant to store ModelSimple or ModelComposed instances, not None.
> Let's just delete this if clause and continue.

But then potentially `if len(anyof_instances) == 0:` would be true, if the input data is the `null` value and the OAS schema has `anyOf: [type: 'null', a, b, c]`

spacether: Yes, return None. The spec says that None is a valid value. We raise exceptions if deserialization fails.

spacether: We never get to counting oneOf or anyOf instance because we returned None in the `ComposedSchema.__new__` method and `ComposedSchema.__init__` is never called.

spacether:
```
a = FruitReq(None)
a == None -> True
isisnstance(a, FruitReq) -> False
isisnstance(a, none_type) -> True
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
sebastien-rosset: done

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache

sebastien-rosset:
> Please add a python test showing deserializing a None working correctly.

Added unit test

modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -1855,6 +1855,7 @@ components:
           - $ref: '#/components/schemas/banana'
     fruitReq:
        oneOf:
+          - type: 'null'
```
spacether, line 1858: Can you add this to the v3 python-experimental spec also?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1024,9 +1024,10 @@ def get_oneof_instance(self, model_args, constant_args):
     schemas described by oneOf.
     Args:
         self: the class we are handling
-        model_args (dict): var_name to var_value
+        model_args (dict/None): var_name to var_value
```
spacether, line 1027: With the neswest code we will never reach here when deserializing None.
Can you remove the None additions from this docstirng?
- this also applies to get_anyof_instances

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1041,6 +1042,13 @@ def get_oneof_instance(self, model_args, constant_args):
     # Iterate over each oneOf schema and determine if the input data
     # matches the oneOf schemas.
     for oneof_class in self._composed_schemas['oneOf']:
+        if oneof_class is none_type:
+            # The only way the input data can match the 'None' type
+            # is if the input data is the 'null' value.
+            if model_args is None:
+                # The input data matches the 'None' type.
+                oneof_instances.append(model_args)
```
spacether, line 1050: Can you delete this `if model_args is None:` block?
We will still go through this code path for deserializing pojo payloads but we should continue through the None like you are doing.
- this also applies to get_anyof_instances

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -125,4 +125,16 @@ def test_deserialize_mammal(self):
...
+        """
+        deserialize fruit
+        fruitReq is a oneOf composed schema model with discriminator, including 'null' type.
+        """
+
+        # Unmarshal 'null' value
+        data = None
+        response = MockResponse(data=json.dumps(data))
+        deserialized = self.deserialize(response, (petstore_api.FruitReq, type(None)), True)
+        self.assertEqual(type(deserialized), type(None))
```
spacether, line 140: I don't think that this will work anymore.
How about [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache#L66)
adding:
```
oneof_anyof_classes = cls._composed_schemas.get('oneOf', ()) + cls._composed_schemas.get('anyOf', ())
if oneof_anyof_classes and none_type in oneof_anyof_classes and args[0] is None:
  return None
```
Then you can also delete the oneof_anyof_classes definition from lines 78-80.

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache

modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
spacether: Was this done; where? In the PR I don't see any code added in the __new__ method.

sebastien-rosset: > Where was this done? In the PR I don't see any code added in the `__new__` method.

Where was what done?

spacether: My above suggestion asked that you add code in the new method. So I am asking where the requested code was added.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2986,6 +2993,12 @@ public CodegenProperty fromProperty(String name, Schema p) {
         } else if (ModelUtils.isFreeFormObject(p)) {
             property.isFreeFormObject = true;
         } else if (ModelUtils.isAnyTypeSchema(p)) {
+            // The 'null' value is allowed when the OAS schema is 'any type'.
+            // See https://github.com/OAI/OpenAPI-Specification/issues/1389
+            if (Boolean.FALSE.equals(p.getNullable())) {
+                LOGGER.error("Schema '{}' is any type, which includes the 'null' value. 'nullable' cannot be set to 'false'", p.getName());
```
wing328, line 2999: I would use LOGGER.warn instead and make it clear it's for the upcoming OAS v3.1

(after OAS v3.1 release, I would expect the parser (swagger-parser) to catch this kind of issue)

sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
sebastien-rosset: ok, I missed that proposed change.

sebastien-rosset: Is it expected that your proposed change has no check for "allOf"?

spacether: Can you describe a payload and schema that would need this allOf check where the check is missing?

when deserializing None where none is in OneOf or AnyOf, any allOf definition of object property fields can never be met by a None payload because None is not a pojo.

Further down in the new function we cover the use caas where we have allof definitionsl, verify that the payload conforms to it then deserialize to an anyof or oneof pojo payload.

sebastien-rosset: That was a question for you. You wrote this code, so I am asking you.

spacether: Does my above answer meet your needs? I don't understand when we would need to check allof but we miss that check. In the answer I describe both cases that do and don't check allof and why.

sebastien-rosset: Thanks. I just copied your code as is without checking the correctness.

spacether:
@sebastien-rosset we need the CI tests to pass before we can merge this.
Travis says:
```
E     File "/home/travis/build/OpenAPITools/openapi-generator/samples/client/petstore/python-experimental/petstore_api/model_utils.py", line 137
E       oneof_anyof_classes = cls._composed_schemas.get('oneOf', ()) +
```
It might need parens around that statement because it is multiline, like:
```
oneof_anyof_classes = (first_part +
  second_part)
```
Can you fix it?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1041,6 +1043,16 @@ def get_oneof_instance(self, model_args, constant_args):
     # Iterate over each oneOf schema and determine if the input data
     # matches the oneOf schemas.
     for oneof_class in self._composed_schemas['oneOf']:
+        # The composed oneOf schema allows the 'null' type and the input data
+        # is the null value. This is a OAS >= 3.1 feature.
+        if oneof_class is none_type:
```
spacether, line 1048: We will only reach this code if we are deserializing a dict payload.
This is because the `__new__` method is run before the `__init__` method. And in `__new__` we return None rather than an instance of the composed class when we pass in None and oneOf contains None type.
So in that case, get_oneof_instance is never called. So in this loop, all we need to do is skip over the none_types.
Can you change this if block to be only this?
```
if oneof_class is none_type:
  # skip none_types because we are deserializing dict data. none_type deserialization is handled in the __new__ method
  continue
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -1098,6 +1112,16 @@ def get_anyof_instances(self, model_args, constant_args):
         return anyof_instances
 
     for anyof_class in self._composed_schemas['anyOf']:
+        # The composed oneOf schema allows the 'null' type and the input data
+        # is the null value. This is a OAS >= 3.1 feature.
+        if anyof_class is none_type:
```
spacether, line 1117: We will only reach this code if we are deserializing a dict payload.
This is because the `__new__` method is run before the `__init__` method. And in `__new__` we return None rather than an instance of the composed class when we pass in None and oneOf contains None type.
So in that case, get_anyof_instances is never called. So in this loop, all we need to do is skip over the none_types.
Can you change this if block to be only this?
```
if anyof_class is none_type:
  # skip none_types because we are deserializing dict data. none_type deserialization is handled in the __new__ method
  continue
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
sebastien-rosset: Actually, I just added this code because otherwise the existing unit tests fail. This code IS executed when the unit tests are executed.

sebastien-rosset: Thank you for the change. I just tested it. It works.

spacether:
@sebastien-rosset to fix the CI-error you need to update this line:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_simple.mustache#L14
to:
```
_composed_schemas = {}
```

sebastien-rosset:
> @sebastien-rosset to fix this next CI-error you need to update this line:
> https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_simple.mustache#L14
> to:
> 
> ```
> _composed_schemas = {}
> ```
> 
> or change the oneof_anyof_classes assignment in `__new__` to only assign a value if bool(cls._composed_schemas) == True which is equivalent to:
> 
> ```
> oneof_anyof_classes = ()
> if cls._composed_schemas:
>   oneof_anyof_classes = (cls._composed_schemas.get('oneOf', ()) +
>                                 cls._composed_schemas.get('anyOf', ()))
> ```

I have some something like that here: https://github.com/OpenAPITools/openapi-generator/pull/5809/files#diff-cc72ac40056827e3a103f9b4f56059cfR70

I mentioned #5809 should be merged before this PR.

PR #6133, opened 2020-05-01
[python-experimental] Make the array items attribute optional

sebastien-rosset:
In OAS 3.0.x, the array "items" attribute is required.
In OAS >= 3.1, the array "items" attribute is optional such that the OAS specification is aligned with the JSON schema specification. When "items" is not specified, the elements of the array may be anything at all.

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -910,7 +913,15 @@ public String getTypeString(Schema p, String prefix, String suffix) {
         } else if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return prefix + "[" + getTypeString(inner, "", "") + "]" + fullSuffix;
+            if (inner == null) {
+                // In OAS 3.0.x, the array "items" attribute is required.
+                // In OAS >= 3.1, the array "items" attribute is optional such that the OAS
+                // specification is aligned with the JSON schema specification.
+                // When "items" is not specified, the elements of the array may be anything at all.
+                return prefix + "[bool, date, datetime, dict, float, int, list, str, none_type]" + fullSuffix;
```
wing328, line 921: @sebastien-rosset I don't think we need prefix and fullSuffix here. These are for models only (e.g. Pet => MyPetV3). Same for line 901.

sebastien-rosset: makes sense, I am fixing.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -897,6 +897,9 @@ public String getTypeString(Schema p, String prefix, String suffix) {
         if (")".equals(suffix)) {
             fullSuffix = "," + suffix;
         }
+        if (ModelUtils.isAnyTypeSchema(p)) {
+            return prefix + "bool, date, datetime, dict, float, int, list, str, none_type" + suffix;
```
wing328, line 901: @sebastien-rosset Shall we remove prefix, suffix from this line as well?

sebastien-rosset: done

sebastien-rosset:
@spacether

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -897,6 +899,9 @@ public String getTypeString(Schema p, String prefix, String suffix) {
         if (")".equals(suffix)) {
             fullSuffix = "," + suffix;
         }
+        if (ModelUtils.isAnyTypeSchema(p)) {
+            return "bool, date, datetime, dict, float, int, list, str, none_type";
```
spacether, line 903: We also need to include the prefix before these types and the fullSuffix after them.
Can you update it to this?
```
return prefix + "bool, date, datetime, dict, float, int, list, str, none_type" + fullSuffix;
```
Or you could include ModelUtils.isAnyTypeSchema(p) in the conditionals on lines:
- 906
- 909
And then we wouldn't need this additional if statement because we will correctly return in line 910.
Up to you which way you want to do it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -910,7 +915,15 @@ public String getTypeString(Schema p, String prefix, String suffix) {
         } else if (ModelUtils.isArraySchema(p)) {
             ArraySchema ap = (ArraySchema) p;
             Schema inner = ap.getItems();
-            return prefix + "[" + getTypeString(inner, "", "") + "]" + fullSuffix;
+            if (inner == null) {
+                // In OAS 3.0.x, the array "items" attribute is required.
+                // In OAS >= 3.1, the array "items" attribute is optional such that the OAS
+                // specification is aligned with the JSON schema specification.
+                // When "items" is not specified, the elements of the array may be anything at all.
+                return "[bool, date, datetime, dict, float, int, list, str, none_type]";
```
spacether, line 923: This is the 3rd location in this function that w have defined this list of all types.
Instead of defining this same info 3 times, how about doing one more recursive call and routing us to the return case on line 903 or 910? That keeps this simpler.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: I specifically removed the prefix based on Wing's feedback. prefix/suffix is only used for model, this is done on purpose.

sebastien-rosset: > We also need to include the prefix before these types and the fullSuffix after them.
> Can you update it to this?
> 
> ```
> return prefix + "bool, date, datetime, dict, float, int, list, str, none_type" + fullSuffix;
> ```
> 
> Or you could include `ModelUtils.isAnyTypeSchema(p)` in the conditionals on lines:
> 
> * 906
> * 909
> 
> And then we wouldn't need this additional if statement because we will correctly return in line 910.
> Up to you which way you want to do it.

 I think this would make the code more complicate to understand and maintain, I don't think it's worth adding recursion.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: I'm not convinced by your argument. I think recursion will make the code more complicated to understand and troubleshoot. Plus here the inner schema is null, so getTypeString() would have to be invoked with either a null value, or create/use a new AnyType schema

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: So reading through the java doc we have:
```
    /**
     * Return a string representation of the Python types for the specified schema.
     * Primitive types in the OAS specification are implemented in Python using the corresponding
     * Python primitive types.
     * Composed types (e.g. allAll, oneOf, anyOf) are represented in Python using list of types.
     *
     * @param p The OAS schema.
     * @param prefix prepended to the returned value.
     * @param suffix appended to the returned value.
     * @return a string representation of the Python types
     */
```
So we see that the prefix and suffix are put on the beginning and the end of the string.
Looking at the call site's javadoc we see that:
```
    /**
     * Output the type declaration of a given name
     *
     * @param p property schema
     * @return a string presentation of the type
     */
    @Override
        // this is used to set dataType, which defines a python tuple of classes
        // in Python we will wrap this in () to make it a tuple but here we
        // will omit the parens so the generated documentation will not include
        // them
        return getTypeString(p, "", "");
    }
```
So we see that at the outermost layer we omit setting the prefix and suffix so the documentation works correctly. The prefix and suffix are used for wrapping inner types in the required [], {}, or () characters.

Can't AnyType schemas be inside array items and object properties?
If so, then we need to add the prefix +blah + fullSuffix to include the required [], {}, or () container characters.

sebastien-rosset: > So reading through the java doc we have:
> 
> ```
>     /**
>      * Return a string representation of the Python types for the specified schema.
>      * Primitive types in the OAS specification are implemented in Python using the corresponding
>      * Python primitive types.
>      * Composed types (e.g. allAll, oneOf, anyOf) are represented in Python using list of types.
>      *
>      * @param p The OAS schema.
>      * @param prefix prepended to the returned value.
>      * @param suffix appended to the returned value.
>      * @return a string representation of the Python types
>      */
> ```

I've read that but that doc does not actually explain anything. It's not usable documentation that provides actionable information. Good doc provides contextual information. Without context, the prefix could be any arbitrary string. What value can the prefix get? Since the doc does not say anything, you have to assume it can be anything. Is it even ASCII? UTF-8? Is it a python type? Or maybe a package name? or 200 other possibilities.
Just to be clear, I'm adding documentation whenever I can to improve.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: Yup, my thought was that one would invoke it with the null Schema input.
- line 925 does this same [] wrapping logic
- The returned data is the line 903 anytype data, all we would need to add is a null check in line 902
- The rest of the function uses recursion
But if this is clearer to you we can keep it this way.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
spacether: That doc explains that prefix goes in front, suffix goes behind, and what types they are. I am hearing that you want an explanation of the valid values. Reading the function and my explanation shows that the values can be:
prefix "({[" and empty string
suffix "({[" and empty string
They allow us to enclose the python data types in containing characters which define tuples, dicts, and lists. Does that help?
Thank you for adding the documentation when it helps.
Man I have strived to add a lot of documentation and tests to python-experimental in many many places. When your comments start from a place of this is not a good doc, it is discouraging. Can you give me the benefit of the doubt? I am trying to make improvements; the goal is to inform people and provide a useful tool. This generator has a lot more documentation, features, and tests than a lot of others. I am trying hard here.

sebastien-rosset: I think this function should be private or protected. The fact that it's public means it can be invoked from anywhere, which means by contract the caller can pass anything, not just "({["

I understand you mean well, and we are all striving to improve the doc and the code, but also when a particular doc sentence is not helpful, I need to say it so we can improve it. Here are some guidelines for good docs: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

The doc states this:
@param prefix prepended to the returned value.

From that sentence and the type, you can see the prefix is a string, and the value is just prepended to the return value. But it does not have information about what the prefix is supposed to mean.

spacether: Sure, I get that.
-This is unclear to me,
-I think that this is missing import information about what characters are allowed
-In the future when you write docs, can you follow this best practice

are all sentences that convey the same intent with starting with this is not good.
I promise you that me and others will respond better with these types of responses. Let's try to treat each other well?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: I have added code comments and used recursion. Thanks for your input, that helps to improve the doc.
I have put some additional code comments for recursion because I think the recursion complicates readability and maintainability.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: Sorry, I was just trying to point out that this particular doc was not helpful. You can see from several of my other PRs that I keep improving the doc, so I don't complain, I just improve the doc.

PR #6138, opened 2020-05-02
[TS-Refactor] Top-level exports for fetch & jquery

TiFu:
As discussed in #802, this PR moves all exported classes and objects of the refactored ts-generator to the top-level.

### PR checklist

@bodograumann Did I miss anything?

### DIFF
--- a/docs/generators.md
+++ b/docs/generators.md
@@ -62,7 +62,7 @@ The following generators are available:
 * [swift3-deprecated (deprecated)](generators/swift3-deprecated.md)  
 * [swift4](generators/swift4.md)  
 * [swift5 (beta)](generators/swift5.md)  
-* [typescript](generators/typescript.md) 
+* [typescript (experimental)](generators/typescript.md)  
 * [typescript-angular](generators/typescript-angular.md)  
 * [typescript-angularjs](generators/typescript-angularjs.md)  
 * [typescript-aurelia](generators/typescript-aurelia.md)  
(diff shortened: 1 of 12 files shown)


bodograumann:
I have a few more changes, but they might require more discussion.
When you merge this commit I will need to rebase and then I'll create another PR.

The CI is still complaining, because we forgot to run `/bin/utils/export_generators_readme.sh`.
I don't think it is necessary to update all commits in `typescript-refactor-master` with the generated readme at that point in history, but if you could add the changes here, that would be great.

TiFu:
>  When you merge this commit I will need to rebase and then I'll create another PR.

Related to #5521 or #6027 ? 

Actually, how do you want to proceed with #5521 ? Do you want to rebase it onto `ts-refactor-master` or should I force push `ts-refactor-master` to `ts-refactor` and essentially rename `ts-refactor-master` that way?

bodograumann:
Related to both actually. I have taken the clean up commits out of #5521 and would propose them in a separat PR.

As for the branches, I would be fine with continuing to work on `typescript-refactor-master`, as I have already rebased on it.
My hope is that we can merge an experimental version into master sooner than later and than continue working there. Would you be able to merge into master at that point?

TiFu:
> Would you be able to merge into master at that point?

I would be fine with merging it into `master` - After we are done merging everything, I'll file a PR with the relevant changes and CC everyone required. We'll probably have to check with some of the core team members too because there are changes that touch the core system (like `DefaultCodegen`).

Issue #4513, opened 2019-11-17
[BUG] PHPSlim4 - Want only 1 authentication method out of multiple


Sroose:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used? 4.2.1
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
Actual: adds all security schemes in middlewares that all get executed. 
Expected: only one security scheme should match.

##### Description

The generator does not take into account the fact that multiple security schemes are defined as 'OR'. In other words, the generated PHP Slim code will always try to perform all authentication methods simultaneously.

They are added in the middewares like this:
```
$this->addRoute(..
                $middlewares
            )->setName($operation['operationId']);
```

##### openapi-generator version

4.2.1

##### OpenAPI declaration file content or url

```yaml
        get:
            ...
            security:
                -
                    AdminToken: []
                -
                    UserJWT: []
                -
                    ExternalToken: []
```

##### Steps to reproduce

Create an OpenApi operation with multiple possible security schemes.

##### Related issues/PRs

Looks similar to https://github.com/OpenAPITools/openapi-generator/issues/3844 for Python. Also https://github.com/OpenAPITools/openapi-generator/issues/797 seems relavant.

ybelenko:
@Sroose Thanks for feedback.

I'll add another layer of middleware that checks if request has been authorized by any auth schema.

ybelenko:
@wing328 or @jimschubert help me figure out specification. Spec says about [Security Requirement Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-requirement-object):

> Security Requirement Objects that contain multiple schemes require that all schemes MUST be satisfied for a request to be authorized.
This enables support for scenarios where multiple query parameters or HTTP headers are required to convey security information.

> When a list of Security Requirement Objects is defined on the [OpenAPI Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#oasObject) or [Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operationObject), only one of the Security Requirement Objects in the list needs to be satisfied to authorize the request.

[...]

Issue #3545, opened 2019-08-02
[REQ] Mock Server


ybelenko:
### Is your feature request related to a problem? Please describe.

Is there any server generator with mocking feature already? We briefly discussed it in #969. I'm really interested to make first one with PHP Slim framework. It's complicated topic, so we need to discuss how mock server should work.

## Describe the solution you'd like

Imho, the biggest problem is to distinguish mock server from real one. Client should be fully aware to which server he's calling. It can cause problems in both ways. When you want to call mock server and accidentally call real one you can change/delete real data. And when you call mock server instead of real one by mistake, these changes doesn't really happens even if response is right.

What options do we have, I have two in my mind.

### 1. Detached hostname for mock server

If main hostname is `my-api.tech` then mock server hostname would be `mock.my-api.tech` as subdomain.

#### Pros
* Pretty visual and straightforward. 
* Easy to call endpoints, you can even change url manually in Chrome if you need.

#### Cons
* Need dedicated server and full control over hosting and subdomains.
* SSL certificate for subdomains can cost extra.
* You still can call wrong url by accident. It's easy to misplace hostnames in your client sdk installation.
* You can't say from pure response whether it's mock response or not.

### 2. Specific HTTP Headers

Respond with mock data only when specific HTTP headers provided. Then client should send header `X-MOCK-SERVER: ping` and response will contain header `X-MOCK-SERVER: pong`.

#### Pros
* No need of dedicated server and full control over subdomains.
* Pretty hard to call mock server accidentally, almost impossible.
* With HTTP Headers we can add more mock options. Like call specific response(4xx - error response, 404 - not found etc.)
* You can say that response contains fake data when you see headers related to mock server.
* When response contains `X-MOCK-SERVER: pong` we know that server understood request correct.

#### Cons
* Harder to implement from client side.
* Can't call mock server with Chrome address bar.
* Need to check HTTP headers to understand that response is fake.

## Additional context

It may be future standard of mock server or "Best Practices with Public API" topic, I think we need full attention before proceed. I would prefer HTTP Headers implementation and if community finds it convenient too, then we need to declare these HTTP headers and server behaviour.

jimschubert:
I think the generation of this per language and per framework is beyond the scope of the project.

I also think that generating a production server which can auto-switch to a mock server (which is how I've read one of the suggestions) isn't something that anyone critical of security on a production system would want.

[...]

PR #6153, opened 2020-05-03
[core] Add type and format properties to model of inline response

### LINKED ISSUE #6150: [BUG][Core] No "type" property in {{modelJson}} of inline response
#### Bug Report Checklist

##### Description

Variable `{{modelJson}}` isn't complete for inline responses right now(maybe even for all inline schemas). When I launch any generator with `-DdebugModels` option the output is:
```console
[main] INFO  o.o.codegen.DefaultGenerator - ############ Model info ############
[ {
  "importPath" : "InlineResponse200",
  "model" : {
    "anyOf" : [ ],
    "oneOf" : [ ],
    "allOf" : [ ],
    "name" : "inline_response_200",
    "classname" : "InlineResponse200",
    "classVarName" : "inlineResponse200",
    "modelJson" : "{\n  \"properties\" : {\n    \"foo\" : {\n      \"type\" : \"string\"\n    },\n    \"bar\" : {\n      \"type\" : \"string\"\n    }\n  }\n}",
    "dataType" : "Object",
    "classFilename" : "InlineResponse200",
```
It's wrong, because ~~produced schema is invalid(`type` is required property) and~~ I've set `type` property in original spec file. Spec file provided below.

4.3.1-SNAPSHOT

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Test inline response model
  description: Test inline response model.
  license:
    name: MIT
paths:
  /foobar:
    get:
      summary:  No type property in modelJson of InlineResponse200.
      operationId: testOperation
      description: No type property in modelJson of InlineResponse200
      responses:
        200:
          description: InlineResponse200 itself.
          content:
            application/json:
              schema:
                type: object
                properties:
                  foo:
                    type: string
                  bar:
                    type: string
```

Any generator with `-DdebugModels` option.

##### Steps to reproduce

Run any generator with option above.

##### Suggest a fix

`{{modelJson}}` variable of inline response model should contain `type` property, because it exists in source spec file.

ybelenko:
I tried to copy original `Schema` instance, but failed I guess. Need a small help from a core team.

Closes #6150 

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2129,6 +2129,25 @@ public CodegenModel fromModel(String name, Schema schema) {
             return null;
         }
 
+        String modelJson = Json.pretty(schema);
+        if (schema.getType() == null ) {
+            Schema copy = ModelUtils.unaliasSchema(this.openAPI, schema, importMapping);
```
ybelenko, line 2134: I wonder how to clone `schema` instance here. I need to change result of `Json.pretty(schema);` but keep original instance not modified.

ybelenko:
After a few experiments with schema:
```yaml
definitions:
  Test:
    properties:
      propertyA:
        type: string
      proeprtyB:
        type: string
```
- Swagger Editor doesn't show any warnings.
- Builtin `validate --recommend` method doesn't show any warnings.

Well, let's consider output `modelJson` isn't broken. But I still insist that it's a bug, because input schema contains `type: object` while output doesn't.

Since it's a core bug I subscribe @OpenAPITools/generator-core-team

ybelenko:
Thanks for the review, Jim!

> Add a `type: object` to the output when it exists in the input makes sense, but consider that someone might think this is now a bug when they don't define `type: object` for input but it's now in the output.

Is it possible to check that input definition contains `type` property? I tried to solve it, but it seems that my fix doesn't work, because there are files in samples with additional `type` field eg. [openapi.yaml in go petstore](https://github.com/ybelenko/openapi-generator/blob/8698e0e05132922cba5ff4c8516d8ec8f0c8318e/samples/client/petstore/go-experimental/go-petstore/api/openapi.yaml). Can you briefly check the test I've made [DefaultCodegenTest.java](https://github.com/ybelenko/openapi-generator/blob/8698e0e05132922cba5ff4c8516d8ec8f0c8318e/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java)?

I've read OAS 3.0.3 spec yesterday and can't recall anything like "consider schema without `type` as `object`". Can you provide a ref from specification?

ybelenko:
@jimschubert The reason why I ask about parsing schemas without `type` is because I have different answer from @sebastien-rosset. Maybe I'm wrong, but he said that parser should **guess** schema type from additional properties. Here is what he said in Slack conversation:
> **Sebastian Rosset**
> Your question is more about JSON schema, since OAI uses JSON schema. The relevant sections are https://tools.ietf.org/html/draft-handrews-json-schema-02#section-4.2.1, https://tools.ietf.org/html/draft-handrews-json-schema-02#section-9.3.2, https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.1. Since JSON schema is constraint based and has a principle to achieve keyword independance, I believe the constraints are applied whenever applicable. If you take a simpler example:
> ```yml
> foo:
>   # type: string commented out
>   maxLength: 10
> ```
> then what this means is 1) there is no constraint on the type (because `type` is not specified), i.e. the type can be null, string, number, object, boolean, array. 2) when the value is a string, then the maximum length of the string is 10.
I think it's the same for `properties` and `type` although that would be extremely unusual.

And maybe it's not relevant to current issue directly, but @richardwhiuk said:
> **Richard Whitehouse**
> Yes, but, note, the following is valid:
> ```yml
> anyOf:
> - type: string
> - type: boolean
> ```
> is valid.
> However,  the following isn't really valid:
> ```yml
> allOf:
> - type: string
> - type: boolean
> ```
> because there's no valid options.
> Also, note, that while something might be valid OpenAPI specification, the OpenAPI specification is vast, and thus not all validators / generators may support it.

sebastien-rosset:
> I don't think this is a bug necessarily, since any schema without a type defined is an object implicitly. 

I don't think a schema without type means it is an object implicitly. Where do you see that specified? I think no type means the type can be anything. For example below the value of `any_prop` can be anything because the `type` keyword has not been specified.

```yaml
foo:
  type: object
  properties:
    int_prop:
      type: integer
    any_prop:
      description: a property whose value can be an integer, boolean, number, array or object.
```
> Add a `type: object` to the output when it exists in the input makes sense, but consider that someone might think this is now a bug when they don't define `type: object` for input but it's now in the output.
> 
> I do think it makes sense to have this, however. OAS 3.1 will have a `type: null` and I wouldn't be surprised if that became the default.

samples/client/petstore/go-experimental/go-petstore/api/openapi.yaml
```diff
@@ -2090,10 +2093,12 @@ components:
       properties:
         breed:
           type: string
+      type: object
```
sebastien-rosset, line 2096: Here I agree `type: object` was missing. It it there in the original document (modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml), it makes sense to add it.

samples/client/petstore/go-experimental/go-petstore/api/openapi.yaml
```diff
@@ -1422,14 +1422,17 @@ components:
       allOf:
       - $ref: '#/components/schemas/Animal'
       - $ref: '#/components/schemas/Dog_allOf'
+      type: object
```
sebastien-rosset, line 1425: In JSON schema, allOf means a compliant validator must validate the payload against all of the nested schemas under `allOf`. Since Animal is already specified as a `type: object`, it's redundant and verbose to add `type: object` here.

sebastien-rosset:
IMO, `type: object` should be present when the schema is a non-composed object type, i.e. a type with declared properties such as `Animal` in the example. But if the schema is a composed schema (anyOf/oneOf/allOf) where the nested schemas are already of type object, I think adding `type: object` should not be done, it unnecessarily adds verbosity.

sebastien-rosset:
In #6150, you state that *type is required property*, but that is not correct. The type keyword is not required.

ybelenko:
> In #6150, you state that _type is required property_, but that is not correct. The type keyword is not required.

Yeah, my mistake. Current OAS 3.0 Specification doesn't declare `type` as required property, but it doesn't say that `type` can be omitted either. At least two validators(Swagger Editor, builtin `validate` function) confirm your point which is enough for me.

For me(I need to parse OAS in my own tool) it means that I shouldn't throw exceptions when schema doesn't contain `type` prop. And I should **guess** which type is presented based on other properties.

sebastien-rosset:
> > In #6150, you state that _type is required property_, but that is not correct. The type keyword is not required.
> 
> Yeah, my mistake. Current OAS 3.0 Specification doesn't declare `type` as required property, but it doesn't say that `type` can be omitted either. At least two validators(Swagger Editor, builtin `validate` function) confirm your point which is enough for me.
> 
> For me(I need to parse OAS in my own tool) it means that I shouldn't throw exceptions when schema doesn't contain `type` prop. And I should **guess** which type is presented based on other properties.

Just FYI, the relevant sections of the JSON schema are:

> JSON Schema uses keywords to assert constraints on JSON instances

and

> JSON Schema keywords fall into several general behavior categories. Assertions validate that an instance satisfies constraints, producing a boolean result. [...] Evaluating an instance against a schema involves processing all of the keywords in the schema against the appropriate locations within the instance. Typically, applicator keywords are processed until a schema object with no applicators (and therefore no subschemas) is reached.

So if a keyword is not present, there is no constraint assertion. There is nothing to guess.

Also here I think  it's better to use the word *keyword* rather than *property* because "property" means something else in JSON schema.

ybelenko:
> So if a keyword is not present, there is no constraint assertion. There is nothing to guess.

My case is a bit different. I'm working on a data mocker. When mock tool gets input schema without `type` constraint should it generate `null` or free  object?

sebastien-rosset:
First, let me point out we make a distinction between "any object" versus free form object.
A free-form object is an *object* (i.e. a dictionary) schema that does not have any declared property AND it accepts arbitrary undeclared properties.
For example `{ "foo": "bar", "p1": true, "p2": { "a": "b" } }` is a free-form object. However `true`, `[ { "a": "b"}, true, "foo" ]` and `"abc"` are not free-form objects because their value is not a dictionary.

The "any type" schema is even more generic than free-form. It's not necessarily an object. It could be an object, but also an array, the null value, an integer, number, string or boolean.

Are you saying the purpose of the data mocker is to automatically generate mock data based on the OpenAPI schema?

ybelenko:
> Are you saying the purpose of the data mocker is to automatically generate mock data based on the OpenAPI schema?

That's how I see it. The main advantage of OAS is auto generation of application parts. OAS perfectly describes how request and response looks like. I've spent so many time on writing server unit tests which can be autogenerated. Usually I need to write fixtures of valid and invalid requests. And that is the place where mock or faker engine fits perfectly.

There are developers that already shows interest in mock server. This feature helps to get demo server as soon as possible. #3545

Here is the tool I make btw - [OpenAPI Data mocker](https://github.com/ybelenko/openapi-data-mocker).

sebastien-rosset:
> > Are you saying the purpose of the data mocker is to automatically generate mock data based on the OpenAPI schema?
> 
> That's how I see it. The main advantage of OAS is auto generation of application parts. OAS perfectly describes how request and response looks like. I've spent so many time on writing server unit tests which can be autogenerated. Usually I need to write fixtures of valid and invalid requests. And that is the place where mock or faker engine fits perfectly.
> 
> There are developers that already shows interest in mock server. This feature helps to get demo server as soon as possible. #3545
> 
> Here is the tool I make btw - [OpenAPI Data mocker](https://github.com/ybelenko/openapi-data-mocker).

I would consider the following scenarios that don't have the `type` keyword directly under the schema.

```yaml
components:
  freeFormSchema:
    description: free form object (must be a dictionary). It would be very unusual to write the schema that way, but since the schema must match ALL of the allOf nested schema, that means the top-level schema must be an object.
    allOf:
      - type: object
    additionalProperties: true
 anyTypeSchema:
   description: any type, the payload can be anything (boolean, integer, null, number, string, object, array). Ideally the mock server would be able to generate all of these values.
 unusualSchemaWithConstraintsAndNoType:
   description: this is (imo) a very unusual way to specify a schema. The type can be anything. But when the value is a number, it cannot be more than 10. If the value is a string, the length of the string cannot be more than 20. Any boolean value is ok. If the value is an array, it cannot have more than 30 items.
   maximum: 10
   maxLength: 20
   maxItems: 30
 nothingSchema:
  description: no JSON payload can ever match this schema
   allOf:
   - type: string
   - type: boolean
 composedSchemaAllOf:
  description: The type of this schema depends on the type of the nested schemas. The nested schemas must be of the same type, except the scenario when A is an integer and B is a number.
   allOf:
   - $ref: '#/components/schemas/A'
   - $ref: '#/components/schemas/B'
 composedSchemaOneOf:
  description: The type of this schema depends on the type of the nested schemas
   oneOf:
   - $ref: '#/components/schemas/A'
   - $ref: '#/components/schemas/B'
 composedSchema:
  description: The type of this schema depends on the type of the nested schemas. The nested schemas may have different types.
   anyOf:
   - $ref: '#/components/schemas/A'
   - $ref: '#/components/schemas/B'
```

ybelenko:
> ```yaml
> nothingSchema:
>   description: no JSON payload can ever match this schema
>    allOf:
>    - type: string
>    - type: boolean
> ```

I don't think so. As far as I remember encoding `JSON.stringify("hello world")` returns `"hello world"`. So string, boolean or any scalar can be JSON theoretically. Or maybe you mean something else here.

sebastien-rosset:
> > ```yaml
> > nothingSchema:
> >   description: no JSON payload can ever match this schema
> >    allOf:
> >    - type: string
> >    - type: boolean
> > ```
> 
> I don't think so. As far as I remember encoding `JSON.stringify("hello world")` returns `"hello world"`. So string, boolean or any scalar can be JSON theoretically. Or maybe you mean something else here.

What I mean is suppose the payload is:
```json
"hello"
}

A compliant validator must validate the payload against all `allOf` schemas. So it may start with the first nested schema, it is `type: string` so validation passes. Then it validates the payload against the next nested allOf schema, which is `type: boolean`. The payload does not match that type because it is a string. So overall validation fails.

**Comment by ybelenko** [comment]
Ohhh, now I see. Thanks for so many test cases. I asked community for weeks with no response at all. Thanks again Sebastien!

**Comment by sebastien-rosset** [comment]
> So, in a conclusion I need to fix behaviour with 4 keywords(`allOf`, `anyOf`, `oneOf`, `not`). When `type` keyword is missed in inline schema which contains `allOf/anyOf/oneOf/not` keyword, I need to search keyword in subschemas.

This can get tricky because it recursive, so you can't just look at the direct subschema. Example:

```yaml
 A:
   allOf:
   - $ref: '#/components/schemas/B'
 B:
   allOf:
   - $ref: '#/components/schemas/C'
 C:
   allOf:
   - $ref: '#/components/schemas/D'
```

Also, if you are looking for corner cases, here is one that you probably won't see very often :-)
```yaml
Model:
  anyOf:
    - allOf:
      - type: string
         maxLength: 15
   - type: boolean
   - $ref: '#/components/schemas/D'
  allOf:
  - type: string
    minLength: 20
  - pattern: [a-z]{3,50}
  oneOf:
  - type: string
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -621,6 +621,9 @@ private Schema modelFromProperty(OpenAPI openAPI, Schema object, String path) {
         XML xml = object.getXml();
         Map<String, Schema> properties = object.getProperties();
         Schema model = new Schema();
+        if (object.getType() != null) {
```
ybelenko, line 624: I've figured out Java debugger finally and found that InlineModelResolver strips data type here.

dang-gyg:
thanks for the PR! we are running into this same issue. any chances to get this reviewed?

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
ybelenko: @wing328 Yeas, I think so. It's not so critical as `type`, but it's weird that `InlineModelResolver` can cut off that field completely. I'll test it today, maybe `format` isn't missing.

ybelenko:
Small offtopic.

This PR should be merged as temporarily solution from my point of view. I think that we'll end up with parsing spec file directly anyway. It's hard to store all spec data in codegen variables if it's even possible(I don't know how to store deeply nested values in mustache variables). Maybe server framework should be able to parse full spec(`yaml|json`) and process it directly. There might be issues with referenced schemas if spec divided into many files, but it's doable.

For instance we have stale issue #4513 which cannot be solved because of codegen `{{variables}}`. These variables cannot show difference between `OR` and `AND` logic which presented in original spec.

ybelenko:
@wing328 or @jimschubert could you check and merge? It's 3 months already.

jimschubert:
@ybelenko this has been open for 3 months, but you have commits up to 15 days ago. Realize when you make additional changes and force-push, things have to be reviewed again.

If @sebastien-rosset is content with your recent changes, I'm fine to have this merged.

ybelenko:
> @ybelenko this has been open for 3 months, but you have commits up to 15 days ago. Realize when you make additional changes and force-push, things have to be reviewed again.

Sure, but I didn't drop any commit. I rebase and force push to:
1. Fix CI errors
2. Test on latest master.

I've requested your review again at June 13th.

jimschubert:
@ybelenko just curious, why was rebase necessary rather than merging master up to your branch? The rebase and force-push will often make inline review comments and their resolutions no longer accessible. It also makes it so that contributors can't tell whether this was just a master sync or if there was new work that requires another review.

Reviewing each of your 4 force pushes, I now see there were no changes since Jun 9, but it's unlikely that any contributors are going to dig into force push history when re-reviewing a pull request.

My question still remains that if your last true change was June 9, and concerns were still voiced on June 10, whether or not those concerns are blockers before merge.

ybelenko:
> @ybelenko just curious, why was rebase necessary rather than merging master up to your branch? The rebase and force-push will often make inline review comments and their resolutions no longer accessible. It also makes it so that contributors can't tell whether this was just a master sync or if there was new work that requires another review.

I've been in a project where `rebase` was the major requirement to all feature branches. Team Lead said: "Keep clean commit history and always rebase to latest master/develop and run tests before merge." and he also said "Never merge master/develop to feature branch". In our project I rebase branch because of samples commit. At first, it's much easier to resolve merge conflicts in samples if I drop `Refresh samples` commit and generate them again. Also if I don't rebase I have commit history like:
1. Fix 1
2. Fix 2
3. Refresh samples
4. Fix 3
5. Refresh samples

Beside, it was the first time I've changed my solution completely in the middle of PR. I agree that changes was hard to review. My bad.

> My question still remains that if your last true change was June 9, and concerns were still voiced on June 10, whether or not those concerns are blockers before merge.
I don't know whether my fix solves @dang-gyg issue because there is no details. I think that @sebastien-rosset concerns solved because I decided to fix it inside `InlineModelResolver` and not modify `{{modelJson}}` property directly. So `type` property saved in model only when it presented in original inline schema.

modules/openapi-generator/src/test/resources/3_0/6150_model_json_inline.yaml
```diff
@@ -0,0 +1,77 @@
...
+      responses:
+        400:
+          description: InlineResponse200 itself.
+          content:
+            application/json:
+              schema:
+                title: FooBarObject
+                type: object
+                # wrong format just to make sure it won't be stripped
+                format: date-time
```
sebastien-rosset, line 39: Can you explain the test you are exercising by adding the `format` keyword at the object level? In 3.0.3, the `format` keyword can be specified for primitive types, not for object types. See http://spec.openapis.org/oas/v3.0.3#dataTypeFormat:

> Primitives have an optional modifier property: format.

Or did you mean to add the `format` keyword for the foo property?
```yaml
properties:
  foo:
    format: date-time
    type: string
```

ybelenko: Yeah, I know that `format` property is for primitives only, but my goal was to keep `format` and `type` property even in broken schemas. In other words if user declares `format` property in `type: object` by some reason he will likely expect it won't be stripped of by generator.

modules/openapi-generator/src/test/resources/3_0/6150_model_json_inline.yaml
```diff
@@ -0,0 +1,77 @@
...
+      description: No type property in modelJson of InlineResponse200
+      responses:
+        200:
+          description: InlineResponse200 itself.
+          content:
+            application/json:
+              schema:
+                type: object
+                # OAS3 spec allows to use any value in format field
+                format: unknown
```
sebastien-rosset, line 21: Same question/comment as what I wrote for line 39 below. What is the purpose of specifying the `format` keyword for an object type in your test?

ybelenko: Just to check that `format` property saved in broken schemas. On purpose.

modules/openapi-generator/src/test/resources/3_0/6150_model_json_inline.yaml
sebastien-rosset: Then I recommend that you rewrite the comment. Writing "Wrong format ..." conveys that the value of `format` is incorrect, it should not be `date-time` but some other value. That's not what you were doing here.

```yaml
# In OAS 3.0.3, the 'format' keyword can be used for primitive types, not for object types, as
# specified in http://spec.openapis.org/oas/v3.0.3#dataTypeFormat.
# The `format` keyword is added in this test object type to show the generator retains the format property
# even in broken schemas.
```

sebastien-rosset: Note in the JSON schema specification, the `format` keyword can be used for any instance, not just primitive types. http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7. In the draft 3.1, the `format` section still indicates the `format` keyword applies to primitive types.

The JSON schema spec does not say much about what the `format` keyword would mean for an object type, all defined formats apply to string instances.

modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -621,6 +621,12 @@ private Schema modelFromProperty(OpenAPI openAPI, Schema object, String path) {
...
+            model.setType(object.getType());
+        }
+        if (object.getFormat() != null) {
+            model.setFormat(object.getFormat());
```
sebastien-rosset, line 628: IMO, there may be a contradiction in the OAS spec (both 3.0.3 and 3.1) versus the JSON schema specification.
I have opened https://github.com/OAI/OpenAPI-Specification/issues/2301. It would be interesting to have @wing328  and @jimschubert  opinions.
In any case, I suggest adding the code comments below because later it can be hard for maintainers to understand the context and why this was done.
```suggestion
            // Even though the `format` keyword typically applies to primitive types only,
            // the JSON schema specification states `format` can be used for any model type instance.
            model.setFormat(object.getFormat());
```

modules/openapi-generator/src/test/resources/3_0/6150_model_json_inline.yaml
ybelenko:
```yaml
title: FooBarObject
type: object
# object type cannot contain format, but it still should be presented in resolved model
format: date-time
```
Something like that?

sebastien-rosset: No because the JSON schema specification explicitly supports the `format` keyword for any model instance type. See https://github.com/OAI/OpenAPI-Specification/issues/2301 I just opened related to this issue.
I think the OAS should be slightly rephrased to clarify `format` applies to all types (including object), not just primitive types. Mike Ralphson seems to agree.

```yaml
# It is legal to use the `format` keyword for object types. The JSON schema specification explicitly allows this.
# Even if in practice most OAS authors use `format` for primitive types, it should still be allowed to use format for object types.
```

sebastien-rosset:
> Re-approving but awaiting feedback from others whether their concerns have been addressed.

The changes look good. I did make a couple of suggestions to add code comments. This is to avoid the element of surprise/confusion when future code maintainers will read the source code. It's not immediately obvious the `format` keyword is legal for object types.

ybelenko:
Thanks @sebastien-rosset. I'll add those comments and push again. Then I'll call everybody to check again :smile:

ybelenko:
Obviously CI failure isn't related to changes, because I modified comments only.

sebastien-rosset:
thanks! lgtm

jimschubert:
CircleCI was a timeout I have the type/format change and 20 other attributes were weren't copying in #7106, so I'll merge this and retain your comments when I merge it up into my PR.

PR #6177, opened 2020-05-05
Support media types other than json

bodograumann:
Currently the new typescript generator only supports receiving http responses with `Content-Type: application/json`. This is an approach to support other media types as well.
In particular the first goal is to allow parsing arbitrary binary files as `HttpFile`.
Later other media types, like xml could be added (possibly togglable with an config option to not always pull in the relevant parser library)

Please have a look, when you find the time @TiFu. The file download is not quite finished though.

Cf. #802

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
@@ -21,6 +21,12 @@ let primitives = [
                     "number",
                     "any"
                  ];
+
+const supportedMediaTypes: { [mediaType: string]: number } = {
+  "application/json": Infinity,
+  "application/octet-stream": 0
+}
+
                  
 let enumsMap: Set<string> = new Set<string>([
     {{#models}}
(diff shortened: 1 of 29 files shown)


modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
```diff
@@ -162,4 +168,73 @@ export class ObjectSerializer {
...
+     * Normalize media type
+     *
+     * We currently do not handle any media types attributes, i.e. anything
+     * after a semicolon. All content is assumed to be UTF-8 compatible.
+     */
+    public static normalizeMediaType(mediaType: string | undefined): string | undefined {
+        if (mediaType === undefined) {
+            return undefined;
+        }
+        return mediaType.split(";")[0].trim().toLowerCase();
```
TiFu, line 183: Is `;` required at the end of a mediaType/http header field if no additional types follow?

TiFu:
Great addition! :) Looking good so far. 

Adding an additional switch for the expected content type might be a neat addition later on.

modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
bodograumann: `;` separates the parameters from the main media type. The function itself only handles one media type. So multiples would have to be split beforehand.
If there is no `;` that is ok, because then `.split` returns an array with one element that is the whole media type.

bodograumann:
Just a little update, @TiFu:
An earlier suspicion of mine was confirmed. Using fetch api's `Response.text()` interprets the response body as UTF-8. So if it is actually binary data, it will mangle the data.
That means we need the http library to always return binary data and later do the decoding.

bodograumann:
Downloading files works! :tada: 
This now still contains the changes of #6027, so when that is merged, I can rebase again. @TiFu

I did not go for `TextDecoder`, because it only works for `ArrayBuffer` and not for `Blob`. Instead I use immediately `FileReader`. If there are any performance problems please tell me.

`isomorphic-fetch` seems unmaintained, so I directly imported `node-fetch` or `whatwg-fetch`.

TiFu:
Amazing :confetti_ball: 

I'll review the CI failures in #6027 (likely unrelated, saw those two fail in other PRs too) and then merge. If we are lucky, there will be no conflicts and I can directly merge this PR once CI finishes.

bodograumann:
Ok, this should do the trick. The conflicts were only because of the duplicate commits.

TiFu:
Great! :)  Thanks for the PR which has been merged into `ts-refactor-master`

PR #6186, opened 2020-05-06
add use-site target field for JsonProperty annotation in kotlin data classes

### LINKED ISSUE #6185: [BUG] Kotlin generators should have use-site target field for @JsonProperty annotation
In the generated Kotlin data classes there is no use-site target specified on the @JsonProperty annotation. Since constructor paramters are used, Jackson does not recognize this annotation because it is only scanning for field and getter/setter annotations. 
In my case im having problems with a property which should be serialized as `eMailAddress`, but with the default bean naming strategy of Jackson, the serialized property is called `emailAddress`. 
##### OpenAPI declaration file
```yaml
openapi: 3.0.1
info:
  title: Test API
  version: v1
paths:
  /test:
    post:
      tags:
        - User
      operationId: createUser
      requestBody:
        description: The data for the account creation.
        content:
          application/json;charset=utf-8:
            schema:
              $ref: '#/components/schemas/UserInfo'

      responses:
        '201':
          description: The account was created.
components:
  schemas:
    UserInfo:
      required:
        - eMailAddress
      type: object
      properties:
        eMailAddress:
          maxLength: 256
          type: string
          description: The email address of the user.
      description: request object

```
##### Actual generated data class
```kotlin
data class UserInfo(
    @get:NotNull @get:Size(max=256) 
    @JsonProperty("eMailAddress") val eMailAddress: kotlin.String
) {}
```
##### Expected generated data class
```kotlin
data class UserInfo(
    @get:NotNull @get:Size(max=256) 
    @field:JsonProperty("eMailAddress") val eMailAddress: kotlin.String
) {}
```

I'm using the current version 4.3.0 of the generator.

m-kay:
fixes #6185


m-kay:
@jimschubert, @dr4ke616 Could you please have a look at this? This should be for the next 4.3.x version but there was no 4.3.x branch, therefore I filed this PR against master.

jimschubert:
cc rest of kotlin technical committee @karismann @Zomzog @andrewemery @4brunu @yutaka0m 

I think it looks good, and if there aren't any further concerns in the next couple days I'll go ahead and merge.

jimschubert:
Looks like there aren't concerns, so I'll merge and we can see how the change is received.

m-kay:
@jimschubert Will this make it into a 4.3.2 release or are you going to release 5.0.0 soon?

jimschubert:
We plan to release 5.0.0 in stages as we do with major releases. The first stage should be in the next few weeks.

Issue #6206, opened 2020-05-07
[BUG] docker-entrypoint.sh fails with error if no parameters passed


digininja:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

If no command line argument is passed to docker-entrypoint.sh, when it tries to use `$1` on lines 18 and 22 it fails with the error:

```
./docker-entrypoint.sh: line 22: $1: unbound variable
```

##### openapi-generator version

Latest master from Git

##### OpenAPI declaration file content or url

n/a

##### Command line used for generation

n/a

##### Steps to reproduce

```
$ ./run-in-docker.sh 
++ dirname ./run-in-docker.sh
+ cd .
+ maven_cache_repo=/home/robin/.m2/repository
+ mkdir -p /home/robin/.m2/repository
++ id -u
++ id -g
+ docker run --rm -it -w /gen -e GEN_DIR=/gen -e MAVEN_CONFIG=/var/maven/.m2 -e 'MAVEN_OPTS=-Dhttps.protocols=TLSv1.2 -Dmaven.repo.local=/var/maven/.m2/repository -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=WARN -Dorg.slf4j.simpleLogger.showDateTime=true -Djava.awt.headless=true' -u 1000:1000 -v /tmp/openapi-generator-4.3.1x:/gen -v /tmp/openapi-generator-4.3.1x/CI/run-in-docker-settings.xml:/var/maven/.m2/settings.xml -v /home/robin/.m2/repository:/var/maven/.m2/repository --entrypoint /gen/docker-entrypoint.sh maven:3-jdk-8
/gen/docker-entrypoint.sh: line 18: $1: unbound variable
/gen/docker-entrypoint.sh: line 22: $1: unbound variable
```

##### Related issues/PRs

Could not find any

##### Suggest a fix

[...]

PR #6207, opened 2020-05-07
[Docker] Added a check to make sure an argument is passed

digininja:
PR to go with issue https://github.com/OpenAPITools/openapi-generator/issues/6206 to stop `unbound variable` errors when running docker-entrypoint.sh without an argument.

### PR checklist

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -12,10 +12,14 @@ codegen="${cli}/target/openapi-generator-cli.jar"
 # We code in a list of commands here as source processing is potentially buggy (requires undocumented conventional use of annotations).
 # A list of known commands helps us determine if we should compile CLI. There's an edge-case where a new command not added to this
 # list won't be considered a "real" command. We can get around that a bit by checking CLI completions beforehand if it exists.
-commands="list,generate,meta,help,config-help,validate,version"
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -14,6 +14,11 @@ codegen="${cli}/target/openapi-generator-cli.jar"
 # list won't be considered a "real" command. We can get around that a bit by checking CLI completions beforehand if it exists.
 commands="list,generate,meta,help,config-help,validate,version"
 
+if [ $# == 0 ]; then
+	echo "Missing a command"
```
wing328, line 18: @digininja thanks for the PR. Can we also list the commands in the message so that users know what commands to provide?

digininja:
Yes, I can add that.

digininja:
I've done that, and just for OCD value, put the commands into alphabetical order.

My only worry with the change I've made is that the final else on the original if passes through any commands the script doesn't know about to exec. Should this be able to be reached without an argument? If it should, then my update will break that functionality as I exit early if no command is passed. If this has to have at least one argument, then my change is OK.

wing328:
I think we can go with what you've right now and submit PRs later if there are feedbacks from other users on the usage.

PR #6224, opened 2020-05-08
[Typescript-node] Use strict equality (===) instead of (==) to better match rest of code

sinaa:
In few minor cases on the `models.ts` file generated for the typescript-node client, strict equality was not used. This leads to tripping our linting requirements.

This is a minor change to use strict equality (`===` ) across the generated codebase (it is already using strict equality on the majority of the generated code).

### PR checklist

  - @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)

### DIFF
--- a/samples/client/petstore/typescript-node/npm/model/models.ts
+++ b/samples/client/petstore/typescript-node/npm/model/models.ts
@@ -42,7 +42,7 @@ let typeMap: {[index: string]: any} = {
 
 export class ObjectSerializer {
     public static findCorrectType(data: any, expectedType: string) {
-        if (data == undefined) {
+        if (data === undefined) {
             return expectedType;
         } else if (primitives.indexOf(expectedType.toLowerCase()) !== -1) {
             return expectedType;
@@ -59,7 +59,7 @@ export class ObjectSerializer {
 
             // Check the discriminator
             let discriminatorProperty = typeMap[expectedType].discriminator;
-            if (discriminatorProperty == null) {
+            if (discriminatorProperty === null) {
                 return expectedType; // the type does not have a discriminator. use it.
             } else {
                 if (data[discriminatorProperty]) {
@@ -77,7 +77,7 @@ export class ObjectSerializer {
     }
 
     public static serialize(data: any, type: string) {
-        if (data == undefined) {
+        if (data === undefined) {
             return data;
         } else if (primitives.indexOf(type.toLowerCase()) !== -1) {
             return data;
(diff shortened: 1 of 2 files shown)


samples/client/petstore/typescript-node/npm/model/models.ts
```diff
@@ -42,7 +42,7 @@ let typeMap: {[index: string]: any} = {
 
 export class ObjectSerializer {
     public static findCorrectType(data: any, expectedType: string) {
-        if (data == undefined) {
+        if (data === undefined) {
```
amakhrov, line 45: code below assumes `data` is non-null and tries to access its property: 
```
if (data[discriminatorProperty])
```

original code (`== undefined`) would check for both undefined and null. Now `null` remains unchecked, and can lead to NPE at runtime

samples/client/petstore/typescript-node/npm/model/models.ts
```diff
@@ -59,7 +59,7 @@ export class ObjectSerializer {
 
             // Check the discriminator
             let discriminatorProperty = typeMap[expectedType].discriminator;
-            if (discriminatorProperty == null) {
+            if (discriminatorProperty === null) {
```
amakhrov, line 62: `discriminator` can be undefined.
hence, `== null` is a more robust check

samples/client/petstore/typescript-node/npm/model/models.ts
```diff
@@ -117,7 +117,7 @@ export class ObjectSerializer {
     public static deserialize(data: any, type: string) {
         // polymorphism may change the actual type.
         type = ObjectSerializer.findCorrectType(data, type);
-        if (data == undefined) {
+        if (data === undefined) {
```
amakhrov, line 120: samewise, code below tries to access properties of `data`, which can still be null (it it's a nullable type as defined in OpenAPI Spec v3)

samples/client/petstore/typescript-node/npm/model/models.ts
sinaa: @amakhrov many thanks for the review.

I think the assumption here is rather implicit rather than explicit, so would you agree if I change this PR to `if (data === undefined || data === null)`?

amakhrov: yeah that would make sense!

macjohnny:
@sinaa could you please update the code to handle both, `undefined` and `null`?

PR #6229, opened 2020-05-09
[python-experimental] Raise builtin AttributeError instead of custom ApiKeyError

sebastien-rosset:
Currently when invoking the getattr, setattr, hasattr functions, the generated **ApiKeyError** exception is raised. This causes non-standard, non-compliant behavior.

This PR change the implementation (use **AttributeError**) to ensure compliance with the Python documentation:

1. The builtin [getattr](https://docs.python.org/3/library/functions.html#getattr) function has the following contract: If the named attribute does not exist, default is returned if provided, otherwise **AttributeError** is raised.
1. The builtin [hasattr](https://docs.python.org/3/library/functions.html#hasattr) function returns True if the string is the name of one of the object’s attributes, False if not. (This is implemented by calling getattr(object, name) and seeing whether it raises an **AttributeError** or not.)

Similarly for the [setattr](https://docs.python.org/3/library/functions.html#setattr) function 

### PR checklist


sebastien-rosset:
@spacether , @wing328

spacether:
@sebastien-rosset thanks for making those updates, they look good. There are some remaining errors in Travis and CircleCi about too many blank lines and uncommitted changes
Can you fix those?

samples/client/petstore/python-experimental/test/test_child.py
```diff
@@ -64,19 +64,24 @@ def testChild(self):
 
         # setting a value that doesn't exist raises an exception
         # with a key
-        with self.assertRaises(petstore_api.ApiKeyError):
+        with self.assertRaises(AttributeError):
```
spacether, line 67: nit:
We should be checking against our specific petstore_api.ApiAttributeErrors here
not the general python AttributeError

sebastien-rosset: I thought about with but actually I think it's better to check for AttributeError because that's exactly what hasattr/getattr/setattr expect. A slight improvement would be to assert for bot exceptions.

spacether: If we asserted only the AttributeError then we are not proving that we are raising an api.Api AttributeError. It would be more specific to check both or api.Api AttributeError because that is the specific exception that we expect our package to raise.

PR #6259, opened 2020-05-11
[dart-dio][dart-jaguar][client] Update type key from `oauth` to `oauth2`

adg-mh:
The interceptor for dart-dio was checking for a auth method type key of `oauth` but `oauth2` is the actual key being passed through in `authMethods`. A similar issue existed in dart-jaguar (which is what the dio code seemed to be based on). This fix is for both clients.

### PR checklist

@ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @nickmeinhold (2019/09) @athornz (2019/12) @amondnet (2019/12)

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart-dio/auth/oauth.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/auth/oauth.mustache
@@ -7,7 +7,7 @@ class OAuthInterceptor extends AuthInterceptor {
 
     @override
     Future onRequest(RequestOptions options) {
-        final authInfo = getAuthInfo(options, "oauth");
+        final authInfo = getAuthInfo(options, "oauth2");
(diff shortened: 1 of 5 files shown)


jaumard:
So the right way to change this would be that change, but also change the interceptor code to really do an oauth2 support

adg-mh:
I understand... I saw that the interceptor sent the `Bearer` token (which is a similar method with the OAuth2 password flow) and made the wrong conclusion. I'll use my own interceptor to add the `Bearer` header for now. Since Full OAuth2 is going to be more involved I'll close this for now and leave that for a separate PR.

ayushin:
would be good to have it documented

dalewking:
I ran into this too. I believe the current code for OAuth is now dead code because it is checking for "oauth" which the generator will never generate.

PR #6262, opened 2020-05-12
[python-experimental] Support schema property which has $ref to 'oneOf' schema

sebastien-rosset:
This PR adds support for use cases when a schema property is referencing another schema which is itself a oneOf.

### PR checklist


sebastien-rosset:
@spacether , I don't think we currently have any sample or unit test that exercises the following use case: a schema property is referencing a `oneOf` schema.

Based on my initial testing, currently this use case breaks, there is a problem when attempting to deserialize the type, deserialization fails.

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -161,3 +161,26 @@ def test_deserialize_mammal(self):
         self.assertTrue(isinstance(deserialized, petstore_api.Zebra))
         self.assertEqual(deserialized.type, zebra_type)
         self.assertEqual(deserialized.class_name, class_name)
+
+    def test_deserialize_oneof_reference(self):
```
spacether, line 165: Thanks for adding this test
SO the pattern that we have been using is to have this test_deserialization.py file verify deserializing json payloads into python classes.
We use the files like test_drawing.py to demonstrate that we can instantiate model instances correctly.

This test is helpful because it demonstrates that there is a bug here and drawings are not able to accept Shape/Triangle instances because type checking is failing.

In order to follow our patterns on where tests are how about
- moving this test into samples/openapi3/client/petstore/python-experimental/test/test_drawing.py ?

If you want to add a test here can you update it to demonstrate deserializing a Drawing json payload?

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -161,3 +161,26 @@ def test_deserialize_mammal(self):
...
+                    shape_type="Triangle",
+                    triangle_type="IsoscelesTriangle"
+                ),
+                petstore_api.Shape(
+                    shape_type="Quadrilateral",
+                    quadrilateral_type="ComplexQuadrilateral"
+                ),
+            ],
+        )
+        assert isinstance(inst, petstore_api.BiologyHominid)
```
spacether, line 186: This line is hard for me to understand. Why are we expecting the instance type to be BiologyHominid. Should it be Drawing?

Also, because Drawing is an object type schema (ModelNormal subclass) in python, and doesn't have a discriminator, we will never pick a different class when moving through the new method, so we are sure that instances that we make from it will always be of type Drawing.

spacether:
@sebastien-rosset please take a look at [this branch](https://github.com/spacether/openapi-generator/tree/python-oneof-schema-reference_jblack) which I made from this PR.
In it I add the function is_valid_type to check if input models are present in the discriminator in the required data type.
So if we expect Shape and we pass in ScaleneTrianagle, it checks the discriminator in Shape to see if it can find ScaleneTrianagle.

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
sebastien-rosset: done

samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
sebastien-rosset: > This line is hard for me to understand. Why are we expecting the instance type to be BiologyHominid. Should it be petstore_api .Drawing?

This is a typo.

> 
> Also, because Drawing is an object type schema (ModelNormal subclass) in python, and doesn't have a discriminator, we will never pick a different class when moving through the new method, so we are sure that instances that we make from it will always be of type Drawing.

Sure, it's just that I was planning to add more assertions, and it does not hurt to add an assertion.

spacether: Nope it doesn't hurt

sebastien-rosset:
> @sebastien-rosset please take a look at [this branch](https://github.com/spacether/openapi-generator/tree/python-oneof-schema-reference_jblack) which I made from this PR.
> In it I add the function is_valid_type to check if input models are present in the discriminator in the required data type.
> So if we expect Shape and we pass in ScaleneTrianagle, it checks the discriminator in Shape to see if it can find ScaleneTrianagle.
> 
> My branch's update makes it so we can instantiate Drawings with oneOf class inputs for Shape and Triangle types.
> How about including that fix here?

That's awesome, let me check.

sebastien-rosset:
I have tested your proposed changes, this seems to work, which is great.
I did have to make a small [change](https://github.com/OpenAPITools/openapi-generator/pull/6262/files#diff-4cfee576d2b374a4aacb7f5796a33a98R1156) to convert the dict_keys to a list before accessing the item at index 0.

samples/openapi3/client/petstore/python-experimental/test/test_drawing.py
```diff
@@ -0,0 +1,58 @@
...
+            # 'main_shape' has type 'Shape', which is a oneOf [triangle, quadrilateral]
+            # composed schema. So we should be able to assign a petstore_api.Triangle
+            # to a 'main_shape'.
+            main_shape=isosceles_triangle,
+            shapes=[
+                petstore_api.Shape(
+                    shape_type="Triangle",
+                    triangle_type="EquilateralTriangle"
+                ),
+                petstore_api.Shape(
```
spacether, line 45: Can we keep one of these as a Triangle, or a specific shape like EquilateralTriangle?
That requires that our solution depends upon shape's discriminator and not on instance._visited_composed_classes containing Shape.

sebastien-rosset: I've improved the unit tests.

sebastien-rosset:
There is a dependency on #6121 being merged before we can handle the 'null' type in oneOf. If we have a `ShapeOrNull` schema which is oneOf [type: 'null', Triangle, Quadrilateral] and in Drawing we have a `secondary_shape` property, then we should be able to assign `None` to `Drawing.secondary_shape`, whereas assigning `None` to `Drawing.main_shape` should raise an exception.

PR #6275, opened 2020-05-13
[python-experimental] Support schema property which has $ref to 'oneOf' schema and child schema is null type

sebastien-rosset:
This is a follow-up to #6262. It adds support for the scenario when a oneOf/anyOf child schema is the 'null' type.  
This should be merged AFTER #6262.

### PR checklist


sebastien-rosset:
@spacether

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -872,6 +872,65 @@ def attempt_convert_item(input_value, valid_classes, path_to_item,
...
+        bool
+    """
+    valid_type = input_class_simple in valid_classes
+    if not valid_type and (
+            issubclass(input_class_simple, OpenApiModel) or
+            input_class_simple is none_type):
+        for valid_class in valid_classes:
+            if not valid_class.discriminator:
+                continue
+            if input_class_simple is none_type and is_type_nullable(valid_class):
```
sebastien-rosset, line 921: @spacether , I think this test is needed to handle the scenario when 1) the input value is None, AND 2) the oneOf/anyOf schema has a child schema which is the 'null' type. Is there a better way to handle this scenario?

spacether: In my opinion this is the right way to test for this use case.
We can have a schema with:
oneOf:
 - null
 - integer
 - string
If we have this case with no discriminator, then we would never return True, because we continued on line 920
How about moving the none check to go before the discriminator continue block?

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
```diff
@@ -872,6 +872,65 @@ def attempt_convert_item(input_value, valid_classes, path_to_item,
...
+        bool
+    """
+    if input_type is none_type:
+        return True
+    # TODO: how do we handle the 'nullable: true' scenario?
+    if issubclass(input_type, ModelComposed):
+        # If oneOf/anyOf, check if the 'null' type is one of the allowed types.            
+        for t in input_type._composed_schemas.get('oneOf', ()):
+            if t is none_type:
+                return True
```
spacether, line 897: t could also be a ModelComposed subclass.
How about also recursively calling is_type_nullable on t here?
- this also applies in the below anyOf block

modules/openapi-generator/src/main/resources/python/python-experimental/model_utils.mustache
sebastien-rosset: done

Issue #1378, opened 2018-11-05
[Elm] import 'Data.List_' not used, but causing an error


Stealthmate:
##### Description

I'm generating an API from haskell's servant-swagger, where an endpoint returns a type `QueryResponse` which is an object with two array fields.
The output file for this a module `Data.QueryResponse`, which has the import line
```
import Data.List_ as List_ exposing (List_)
```
however, `List_` is not used anywhere in the file. On top of that, Elm doesn't recognize any module like that, and from what I've found there isn't one.

##### openapi-generator version
`3.3.2`, built from source from current `master`

##### OpenAPI declaration file content or url
[gist](https://gist.github.com/Stealthmate/e8756ef17f9cb3267081feff171c467c#file-swagger-json-L341-L365)

##### Command line used for generation
`openapi-generator-cli.sh generate -i "http://localhost:8080/swagger.json" -l elm -o ./api`

##### Steps to reproduce
Just execute the above command on a host that serves the file

##### Related issues/PRs
I tried looking for similar issues, but didn't find anything.

##### Suggest a fix/enhancement
I'm not really sure, but I think just simply not importing List_ would do the trick.

Stealthmate:
I noticed some other problems.

- The API returns `[[string]]`, however the template generated is for `List String` (the `data` field)
- The encoder/decoder functions use `String.encoder` / `String.decoder` respectively, however those don't exist.

Here's a [gist](https://gist.github.com/Stealthmate/afa501109255f4b700930bcdb524255f) of the code as well.

I think nested list generation may not be working properly. I'll try to do some more research during the week if I have time.

trenneman:
Thanks for your report @Stealthmate. `List_` is indeed a bug as nested Lists/Dicts were not yet supported. See the PR for a fix for both.

Stealthmate:
@trenneman Thank you! I'll take a look!

Issue #1374, opened 2018-11-05
[JAVA] Untyped request bodies not handled


tsiq-karold:
##### Description

Generating code (both client and server) for a model where a resource has a body parameter with either no type or of type `object` generates a warning and code which does not compile.

Warning this generates is similar to:
```
[main] WARN  o.o.codegen.DefaultCodegen - Unknown type found in the schema: object
[main] WARN  o.o.codegen.DefaultCodegen - The following schema has undefined (null) baseType. It could be due to form parameter defined in OpenAPI v2 spec with incorrect consumes. A correct 'consumes' for form parameters should be 'application/x-www-form-urlencoded' or 'multipart/form-data'
[main] WARN  o.o.codegen.DefaultCodegen - schema: class ObjectSchema {
    class Schema {
        title: null
        multipleOf: null
        maximum: null
        exclusiveMaximum: null
        minimum: null
        exclusiveMinimum: null
        maxLength: null
        minLength: null
        pattern: null
        maxItems: null
        minItems: null
        uniqueItems: null
        maxProperties: null
        minProperties: null
        required: null
        type: null
        not: null
        properties: null
        additionalProperties: null
        description: null
        format: null
        $ref: null
        nullable: null
        readOnly: null
        writeOnly: null
        example: null
        externalDocs: null
        deprecated: null
        discriminator: null
        xml: null
    }
    type: object
    defaultObject: null
}
[main] WARN  o.o.codegen.DefaultCodegen - codegenModel is null. Default to UNKNOWN_BASE_TYPE
```

Generated client code in `DefaultApi`:
```java
import org.openapitools.client.model.UNKNOWN_BASE_TYPE;

...

public void thePost(UNKNOWN_BASE_TYPE UNKNOWN_BASE_TYPE) throws ApiException {

    thePostWithHttpInfo(UNKNOWN_BASE_TYPE);
}
```

No such model is generated though even if it was, it wouldn't be terribly useful.  The server generated code has a very similar problem.

I have tried this with both `type: object` and no type specified.

##### openapi-generator version

I've tried with 3.3.2 and master.

##### OpenAPI declaration file content or url

```json
{
  "swagger": "2.0",
  "info": {
    "version": "0.0.1",
    "title": "Test"
  },
  "basePath": "/",
  "paths": {
    "/": {
      "post": {
        "operationId": "thePost",
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "required": "false",
            "schema": {
[...]

Issue #1369, opened 2018-11-04
[rust client] Client does not generate enums. Bug: oneOf schemas are not correct


toiletspider:
##### Description

Enums are not been generated (there is a TODO though).
The [oneOf](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/) schemas are not generated correctly. See my [pr](https://github.com/toiletspider/binance_api/commit/d67384fa40cbe672ca1219f941c757a7b059b0c6) where I changed the generated code so that it could compile. 

##### openapi-generator version

version 3.3.2 and 4.0.0

##### OpenAPI declaration file content or url

This is the [schema](https://github.com/toiletspider/binance_api/blob/master/binance.yml) used. See below what caused the bug.
```yaml
        200:
          description: Ticker/s for the last 24hr
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/ticker24hr'
                  - $ref: '#/components/schemas/ticker24hrlist'
```
##### Command line used for generation

'java -jar openapi-generator-cli-3.3.2.jar generate -i binance.yml -o test/ -g rust'

##### Suggest a fix/enhancement

I am new to rust and are still learning so are a bit unsure if the enums changes are correct if someone can confirm, Ill do a pr to add the enums.

The oneOf bug though, I am not sure how to fix.

toiletspider:
fixed in #2244 :+1:

PR #6290, opened 2020-05-13
[codegen][python-experimental] Composed schema with additionalProperties 

sebastien-rosset:
# Background

As per OAS specification:
1. Undeclared properties should be allowed when the `additionalProperties` keyword is not specified, when `additionalProperties: true` or when `additionalProperties: {}`. The `additionalProperties` keyword can also be used to specify a schema for the undeclared properties.
1. A composed schema may include the `additionalProperties` keyword, both as a sibling of allOf/oneOf/anyOf, and in the child schemas.
1. The value of undeclared properties may be the null value.

# Changes

1. Add `disallowAdditionalPropertiesIfNotPresent` CLI option to control the interpretation of the `additionalProperties` keyword. See below for more details.
1. Change the interpretation of the `additionalProperties` keyword to comply with the OAS 2.0 and 3.x specifications.
   1. Specifically, the following behavior changes have been made and are necessary to comply with the OAS spec:
      1. When the `additionalProperties` keyword is not present in the OAS document, that should be interpreted as if `additionalProperties` had been set to `true`.
       1. The additional properties should be nullable, i.e. the null value should be allowed.
   1. Some people may have built their OAS documents with the understanding that by default, additional properties are not allowed, because they incorrectly assumed the OpenAPITools implementation was correct. To retain the legacy behavior you can set the `disallowAdditionalPropertiesIfNotPresent` CLI option to `true`, which is the default for now.
   1. Limitation: due to issue [swagger-parser #1369](https://github.com/swagger-api/swagger-parser/issues/1369), achieving OAS compliance is not possible when the input document uses the OAS 2.0 schema. For OAS 2.0 documents, the legacy behavior should be used by setting the `disallowAdditionalPropertiesIfNotPresent` CLI option to `true`.
1. By default the CLI option is set to `true` (i.e. non-compliant behavior, as originally implemented). Initially I wanted to change the default value to `false` for all code generators, but it's too much work. For now I am only enabling the *compliant* mode for `python-experimental`.
1. Codegen and mustache templates were not properly processing composed schemas that have the `additionalProperties` keyword set. The issue is specifically resolved for python-experimental.
1. codegen: add equals and hashCode methods for SemVer class.
1. python-experimental: add unit test to validate a composed schema with additionalProperties and other use cases related to the `additionalProperties` keyword.
1. python-experimental: correctly handle the OAS `nullable` keyword. When a schema is nullable, the client should accept the `null` value in the input data.
1. python-experimental: correctly handle `additionalProperties` which is referencing a schema with a `$ref`. Generate imports
1. python-experimental: correctly generate the data type when `additionalProperties` references a schema
1. Add code comments in codegen.

Note: breaking changes are introduced in this PR, but they are necessary to be compliant with the OAS specification.

For example:
```yaml
    Dog:
      allOf:
        - type: object
          properties:
            breed:
              type: string
```

In the above example, the `additionalProperties` keyword is not specified, which means by default undeclared properties should be allowed. This equivalent to:

```yaml
    Dog:
      additionalProperties: true
      allOf:
        - type: object
          properties:
            breed:
              type: string
```

# Related Issues and future work

While working on this PR, the following related issues have been identified:
1. [swagger-parser #1378](https://github.com/swagger-api/swagger-parser/issues/1378) swagger-v2-converter incorrectly converts 2.0 schemas that do not have the 'type' keyword
1. [swagger-parser #1369](https://github.com/swagger-api/swagger-parser/issues/1369): additionalProperties not set correctly if omitted
1. Once the above issues have been fixed, some of the code in this PR can be disabled. Specifically it will no longer be necessary to differentiate between OAS 2.0 docs and OAS 3.0 docs.
1. [swagger-parser #1374](https://github.com/swagger-api/swagger-parser/pull/1374): when a OAS 2.0 doc is converted, it would be helpful to retain the "2.0" version in a vendor extension, that would make it possible to know the original OAS doc was 2.0, not 3.x. That can help to drive some of the processing logic.
1. Support for composed schemas that use the `additionalProperties` keyword is limited to python-experimental for now. Significant changes may have to be done in other languages. For example, in the Java clients, single inheritance is used, that causes a conflict between inheriting the `allOf` parent and the `HashMap` parent.
1. Some of the names in samples OAS documents should be changed to better reflect the intent and behavior. See discussion in this PR for more details.
1. Issue #6400

### PR checklist


samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py
```diff
@@ -176,3 +176,27 @@ def test_deserialize_fruit_null_value(self):
...
+        # The OAS document for Dog does not specify the 'additionalProperties' keyword. 
+        # The additionalProperties keyword is used to control the handling of extra stuff,
+        # that is, properties whose names are not listed in the properties keyword.
+        # By default any additional properties are allowed.
+        data = {
+            'className': 'Dog',
+            'color': 'brown',
+            'breed': 'golden retriever',
+            # Below are additional, undeclared properties
+            'group': 'Terrier Group',
```
sebastien-rosset, line 195: @spacether , currently this unit test raises the following exception:

`petstore_api.exceptions.ApiValueError: Invalid input arguments input when making an instance of class Dog. Not all inputs were used. The unused input data is {'group': 'Terrier Group', 'size': 'medium'}`

spacether: Does the java code set additionalProperties to true? I think that we should fix that on the java side, then the python will work.

spacether: The oas document spec version should determine this additionalProperties behavior

sebastien-rosset: > Does the java code set additionalProperties to true? I think that we should fix that on the java side, then the python will work.

ok, let me check the Java side.

sebastien-rosset: There is a good discussion at https://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties

sebastien-rosset: Yes, this is a problem in Java codegen: https://github.com/OpenAPITools/openapi-generator/pull/6290/files#diff-57d7532cf464a8d7c24aab4b22ceb993R1102

sebastien-rosset:
@wing328 , I think [addAdditionPropertiesToCodeGenModel](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2365) should be called for composed schemas as well, not just map schemas. AFAIK, the `additionalProperties` keyword can be used for `type: object` but also for composed schemas. What do you think?

sebastien-rosset:
@jimschubert

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -970,13 +972,17 @@ public String toInstantiationType(Schema property) {
...
-            codegenModel.additionalPropertiesType = typeString;
-        } else {
-            addParentContainer(codegenModel, codegenModel.name, schema);
+        if (addProps != null) {
+            if (addProps.get$ref() == null) {
+                // if AdditionalProperties exists and is an inline definition, get its datatype and
+                // store it in codegenModel.additionalPropertiesType.
+                codegenModel.additionalPropertiesType = getTypeDeclaration(addProps);;
+            } else {
+                addParentContainer(codegenModel, codegenModel.name, schema);
```
sebastien-rosset, line 981: @spacether , can you comment on the purpose of calling addParentContainer() for python? I don't understand the use case.
Also, previously if addProps was null, we were calling addParentContainer(). Why? addProps is null when the OAS schema has `additionalProperties: false`.

spacether: So I'm not sure what benefit we get from calling addParentContainer. It looks like I copied and pasted that code from DefaultCodegen at:
https://github.com/OpenAPITools/openapi-generator/blob/dc1bdac82073093913b3066fca51b730f2f8cedb/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2796

1.
The way I read that code is that if additional properties exist, we call the if block.
If they don't call the else block. And the else block code is consistent with what the DefaultCodegen does in its own addAdditionPropertiesToCodeGenModel method.

2. Per my #1 interpretation addParentContainer would be called if additionalProperties is false. It would also be called if additionalProperties is null.

3. I don't think that it should matter if additionalProperties is inline or a $ref. Doesn't the code that we have work for both cases?

spacether: If additionalProperties is null and the spec version is 3.0.0 or 3.0.03 or higher, shouldn't additionalProperties default to true before we hande it here?
v3.0.3 explicitly says that it defaults to true here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md
v3.0.0 points us to the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md JSON schema definition, and does not tell us a default value.

sebastien-rosset: I think the default is always specified regardless of the version, because of the following text in the OAS specs:

> The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification.
> additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema.

Note OAS 2.0 is not exactly the same language as OAS 3.0, but they both mean the `additionalProperties` keyword is specified in the JSON schema validation draft.

Hence whatever is specified in JSON schema validation is applicable to OAS. Each OAS version is referring to a different version of the JSON drafts, but they all seem to have the same default behavior.

Specifically from https://tools.ietf.org/html/draft-wright-json-schema-validation-00#page-10:

> If "additionalProperties" is absent, it may be considered present with an empty schema as a value.

Which means the values can be anything.

sebastien-rosset: BTW, this sentence in 3.0.3 is superfluous: "Consistent with JSON Schema, additionalProperties defaults to true."  It's already defined in the JSON core/validation specs, hence it should not be repeated in OAS.

In fact in OAS 3.1, it was completely removed because the goal of the technical committee is for OAS to be a superset of JSON schema, i.e. supporting all JSON schema constructs as is, without the need to rephrase the JSON specification inside the OAS spec.

sebastien-rosset: > 
> The way I read that code is that if additional properties exist, we call the if block.
> If they don't call the else block. And the else block code is consistent with what the DefaultCodegen does in its own addAdditionPropertiesToCodeGenModel method.
> 

ok, let me try to refactor that code a bit so that it works the same whether it's a reference or not. I will add a unit test too to exercise the code.

samples/openapi3/client/petstore/python-experimental/petstore_api/models/drawing.py
```diff
@@ -69,7 +69,7 @@ class Drawing(ModelNormal):
     validations = {
     }
 
-    additional_properties_type = None
+    additional_properties_type = (one_ofapplebanana.OneOfapplebanana,)  # noqa: E501
```
sebastien-rosset, line 72: @spacether , here this is not correctly generated, it should be a reference to the `Fruit` class defined in ./petstore_api/models/fruit.py.

spacether: This issue could be handled in that function or it could be handled in getTypeString.

My preference is to handle it in getTypeString because that is our function that handles conversions from Schemas to python types.

How about in get_type string we add an in block that handles the case where schema is of type composed schema? If that is the case we need to convert the referenced model name into a python module style model name, so Pet should become pet.Pet with:
```
#get the model name form the reference
String modelName = ModelUtils.getSimpleRef(ref);
modelName = toModelName(modelName);
```
Are we also handling this case for:
- object type models in additionalproperties
- primitive models (non-object and non-array type) with validations and enums in additionalproperties?
  - should we handle those in a separate PR?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: Unfortunately, I found out the OpenAPI parser for a V2 document sets `Schema.additionalProperties` to the null value even when the OAS document has `additionalProperties: true|false`
So I need to add some logic that if the doc is version 2, then assume the old behavior, because we don't know if additionalProperties should be true or false. This seems to be a bug in the parser.

spacether: What if you added the CLI flag supportsAdditionalProperties
And if the generator has it turned on, postprocess the spec to correctly set additionalProperties to {} if the input is null
Or should that not be done, or be done in another PR?

samples/openapi3/client/petstore/python-experimental/petstore_api/models/drawing.py
sebastien-rosset: > This issue could be handled in that function or it could be handled in getTypeString.
> 
> My preference is to handle it in getTypeString because that is our function that handles conversions from Schemas to python types.

Thanks for the suggestion. Done

> 
> Are we also handling this case for:
> 
> * object type models in additionalproperties
> * array type models in additionalproperties
> * primitive models (non-object and non-array type) WITHOUT validations and enums in additionalproperties? We do not produce models for these so we should return the simple type from the ref model
> * primitive models (non-object and non-array type) with validations and enums in additionalproperties? We do produce these models so we should return a type like model_simple.ModelSimple
>   
>   * should we handle those in a separate PR?

This will be handled in a separate PR.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1080,12 +1080,55 @@ public static Schema unaliasSchema(OpenAPI openAPI,
...
+                // Note openAPI.getOpenapi() is always set to 3.x even when the document
+                // is converted from a OAS/Swagger 2.0 document.
+                // https://github.com/swagger-api/swagger-parser/pull/1374
+                Object ext = extensions.get("x-original-swagger-version");
+                if (ext instanceof String) {
+                    SemVer version = new SemVer((String)ext);
+                    if (version.major == 2) {
+                        // The OAS version 2 parser sets Schema.additionalProperties to the null value
+                        // even if the OAS document has additionalProperties: true|false
+                        // So we are unable to determine if additional properties are allowed or not.
```
spacether, line 1117: Should we handle this in a preprocessor before a v2 spec is converted to v3?

samples/client/petstore/python-experimental/petstore_api/models/additional_properties_any_type.py
```diff
@@ -64,7 +64,7 @@ class AdditionalPropertiesAnyType(ModelNormal):
     validations = {
     }
 
-    additional_properties_type = (bool, date, datetime, dict, float, int, list, str,)  # noqa: E501
+    additional_properties_type = ({str: (bool, date, datetime, dict, float, int, list, str, none_type)},)  # noqa: E501
```
spacether, line 67: So in python all properties in an object type model are stored in a dictionary named _data_store.
You can see our assignment to it [here](https://github.com/spacether/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/model_utils.py#L98).

We require that key variable names are strings [here](https://github.com/spacether/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/model_utils.py#L68).
The intention in defining these additional_properties_type is describing the value types that may be assigned.
So if the object spec does not include a property "cat",
we can assign the property "cat" which is the key. That assignment is handed in the [`__setattr__`](https://github.com/spacether/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/model_utils.py#L174) and [`set_attribute`](https://github.com/spacether/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/model_utils.py#L49) methods, which require that "cat" be a string.
We use additional_properties_type to describe the type values only.
Because we already have the key string portion taken care of [here](https://github.com/spacether/openapi-generator/blob/master/samples/client/petstore/python-experimental/petstore_api/model_utils.py#L68)
additional_properties_type should only have the value `(bool, date, datetime, dict, float, int, list, str,)`
Can you update your code to do that?
If you have a different interpretation of what should be happening here can you explain it?

spacether: Should we handle this differently if {} anyType is defining additionalProperties on object type schema vs being in a object type schema's CodegenProperty?
1. in an object type schema, the place that we store variables already exists and is `_data_store`, we are only defining value types here, so I think that we should use:
`(bool, date, datetime, dict, float, int, list, str,)`
2. in a property, we are assigning a value. Should this data type be:
  a. `(bool, date, datetime, dict, float, int, list, str,)` or
  b `{str: (bool, date, datetime, dict, float, int, list, str,)}`
It sounds like for the CodegenProperty case we should be using definition b, is that correct?

spacether: Another possible solution here is to keep:

`additional_properties_type = ({str: (bool, date, datetime, dict, float, int, list, str, none_type)},)`
as-is in both additional_properties_type and model properties.
If we did that we would need to:

- update model docstring [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/docstring_openapi_validations.mustache#L6) to describe the updated additional_properties_type and how to use it
- update the set_attribute method [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_set_attribute.mustache#L18) to ignore the `str` key in additional_properties_type and use the dict values only which are `(bool, date, datetime, dict, float, int, list, str, none_type)`

If we didd that, our documentation [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_doc.mustache#L22):
Would produce descriptions like:
```
# additional_properties_any_type.AdditionalPropertiesAnyType

## Properties
Name | Type | Description | Notes
------------ | ------------- | ------------- | -------------
**name** | **str** |  | [optional] 
**any string name** | **{str: (bool, date, datetime, dict, float, int, list, str, none_type)}** | any string name can be used but the value must be the correct type | [optional]
```
That is concerning to me because we are listing that an additionalProperty can have a name of any string, but it looks like the value should be a dict. This is incorrect, the value should be `(bool, date, datetime, dict, float, int, list, str, none_type)`
But because this is rendering `{{additionalPropertiesType}}` we would see the dict definition here which is confusing as this information is describing value types.
How would you fix this?

What do you want to do here?
```

File: modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java (hunk 4 above)
Comment (spacether) [reply modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java:1117]: This is a fine fix for our current use case.
It looks like you are working on a 2.0 fix [here](https://github.com/swagger-api/swagger-parser/pull/1370)

File: modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java (hunk 2 above)
Comment (sebastien-rosset) [reply modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java]: I am adding a legacyAdditionalPropertiesBehavior CLI flag.

File: samples/client/petstore/python-experimental/petstore_api/models/additional_properties_any_type.py (hunk 5 above)
Comment (sebastien-rosset) [reply samples/client/petstore/python-experimental/petstore_api/models/additional_properties_any_type.py:67]: > How about changing the model name AdditionalPropertiesArray to AdditionalPropertiesArrayOfMapAnytype

The values of the undeclared (additional) properties must be an array, and the items in the array must be an object, they cannot be any type. So technically it's not "any type".
I'd rather not change the names in this PR because it's easier to make comparisons of the generated code for the issue raised in this PR.

> anytype_2: I think that this should this result in the data type `(bool, date, datetime, dict, float, int, list, str,)`.

I am going to try to understand the code a bit more.

> Why is this data type the same as anytype_1 + anytype_3?

It was already that way before, so from this perspective my PR is not disturbing whatever was implemented before. Let me go deeper into the code.

Comment (sebastien-rosset) [reply samples/client/petstore/python-experimental/petstore_api/models/additional_properties_any_type.py:67]: The OAS schema for `anytype_1` is as below. That means the data type must be a dictionary; there are no declared properties, and since the `additionalProperties` keyword is not specified, there can be any undeclared properties and their value can be any type.
```yaml
      anytype_1:
        type: object
```

Before this PR, the generated data type was:
```
'anytype_1': (bool, date, datetime, dict, float, int, list, str,),
```

Isn't that incorrect? I.e. doesn't that mean that the value of `anytype_1` can be a boolean, a date, float, etc? But that's not correct, `anytype_1` cannot be a boolean, it must be a dict where the key is a string and the value can be anything.

spacether: > How about changing the model name AdditionalPropertiesArray to AdditionalPropertiesArrayOfMapAnytype
> The values of the undeclared (additional) properties must be an array, and the items in the array must be an object, they cannot be any type. So technically it's not "any type".

I agree with you that's why I am suggesting a rename of:
AdditionalPropertiesArrayOfMapAnytype
Map is a synonym for dict/object. With the suffix `MapAnytype` I mean Map of AnyType, which means Map where the values can be any type. Is AdditionalPropertiesArrayOfMapOfAnytype clearer?

So I added these variables in a prior PR to give examples of AnyType and type: object usage. But my understanding at the time was incorrect. Now that we have a better  correct understanding here, why not make the values stored consistent with the variable names that describe those types?

Otherwise we are giving our users examples which are confusing and where the type described by the variable is different than what the actual type is.
Another solution here is to rename the current variables to be consistent with the updated types that are accepted. But those names get very long.
AdditionalPropertiesClass.map_array_anytype would be AdditionalPropertiesClass.map_map_array_anytype which feels less helpful to our users understanding because there are so many layers of types (4).

> anytype_1 is a bit of a misnomer
Yup, I agree, that's why I am asking that we rename those variables to:
- anytype_1 -> map_anytype1
- anytype_3 -> map_anytype1
Because we are storing dicts where the values can be any types.

The only case where the stored data type looks incorrect to me is anytype_2. All the other ones look like they were produced correcty.

sebastien-rosset: > So I added these variables in a prior PR to give examples of AnyType and type: object usage. But my understanding at the time was incorrect. Now that we have a better correct understanding here, why not make the values stored consistent with the variable names that describe those types?

Yes, I got that part from your earlier comment. I was saying I'd rather not make these changes in this PR because by keeping the names it's a lot easier to compare the changes introduced by this PR. The PR already has a lot of changes, it's getting tricky to compare all the changes.

> 
> Otherwise we are giving our users examples which are confusing and where the type described by the variable is different than what the actual type is.

Yes, I understand, but I'd rather not make these changes here. It's a separate improvement.

> Another solution here is to rename the current variables to be consistent with the updated types that are accepted. But those names get very long.
> AdditionalPropertiesClass.map_array_anytype would be AdditionalPropertiesClass.map_map_array_anytype which feels less helpful to our users understanding because there are so many layers of types (4).
> 
> > anytype_1 is a bit of a misnomer
> 
> Yup, I agree, that's why I am asking that we rename those variables to:
> 
> * anytype_1 -> map_anytype1
> * anytype_3 -> map_anytype2
>   Because we are storing dicts where the values can be any types.
> 
> The only case where the stored data type looks incorrect to me is anytype_2. All the other ones look like they were produced correcty.

Ok, let me work on fixing the generated data type for anytype_2, independant of the renaming.

spacether: Okay, sounds good

sebastien-rosset: Ah, the problem with `anytype_2` is because the Schema.java instantiation is ObjectSchema with `"type: object"`, so it is incorrectly considered a property of type object. I am digging.

sebastien-rosset: Marking as resolved.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -907,18 +909,27 @@ private String getTypeString(Schema p, String prefix, String suffix) {
...
             fullSuffix = "," + suffix;
         }
-        if (ModelUtils.isAnyTypeSchema(p)) {
+        if (StringUtils.isNotEmpty(p.get$ref())) {
+            // The input schema is a reference. If the resolved schema is
+            // a composed schema, convert the name to a Python class.
+            Schema s = ModelUtils.getReferencedSchema(this.openAPI, p);
+            if (s instanceof ComposedSchema) {
+                String modelName = ModelUtils.getSimpleRef(p.get$ref());
+                return prefix + toModelName(modelName) + fullSuffix;
```
sebastien-rosset, line 918: @spacether , here the type may end up being used in the `codegenModel.additionalPropertiesType` property, which is set by `addAdditionPropertiesToCodeGenModel()`. I need to add that type to the imports.

sebastien-rosset: Fixed in f57338d2110e0694dec7c0e07c9966c914c74359

samples/openapi3/client/petstore/python-experimental/petstore_api/models/shape_or_null.py
```diff
@@ -74,7 +74,7 @@ class ShapeOrNull(ModelComposed):
     validations = {
     }
 
-    additional_properties_type = None
+    additional_properties_type = (bool, date, datetime, dict, float, int, list, str, none_type,)  # noqa: E501
 
     @cached_property
     def openapi_types():
```
sebastien-rosset, line 80: Note to self: the generated openapi_types() is wrong, it should include none_type.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java
```diff
@@ -359,4 +359,18 @@
...
 
+    // The reason this parameter exists is because there is a dependency
+    // on swagger-api/swagger-parser issue https://github.com/swagger-api/swagger-parser/issues/1369.
+    // When that issue is resolved, this parameter should be removed.
+    public static final String LEGACY_ADDITIONAL_PROPERTIES_BEHAVIOR = "legacyAdditionalPropertiesBehavior";
+    public static final String LEGACY_ADDITIONAL_PROPERTIES_BEHAVIOR_DESC =
+        "If true, the 'additionalProperties' keyword is implemented as specified in the OAS and JSON schema specifications. " +
+        "Full compliance currently works with OAS 3.0 documents only. " +
+        "It is not supported for 2.0 documents until issues #1369 and #1371 have been resolved in the dependent swagger-parser project. " +
+        "If false, codegen uses a legacy, non-compliant interpretation of the 'additionalProperties' keyword. " +
```
spacether, line 370: This description is confusing to me. We are describing it as legacyAdditionalPropertiesBehavior.
So it sounds like True mean s use legacy behavior.
But then in our definition we say true is spec compliant behavior
False is legacy non-compliant behavior here.
How about instead naming this something like:
- NULLABLE_ADDITIONAL_PROPERTIES_BEHAVIOR
- COMPLIANT_ADDITIONAL_PROPERTIES_BEHAVIOR
or keeping it:
- LEGACY_ADDITIONAL_PROPERTIES_BEHAVIOR
but have True be non-compliant legacy behavior, and False be compliant behavior

Also, what generator sets this to True?
I don't see that happening outside fo the tests. Did I miss it?

sebastien-rosset: Good catch, I inadvertently swapped the values.

sebastien-rosset: Improved documentation. I don't think it's necessary to change the name based on the improved documentation.

sebastien-rosset: The default value is set at https://github.com/OpenAPITools/openapi-generator/pull/6290/files#diff-7cb46fa53f89a458a7b7cb201d2214a8R1511

wing328:
cc @OpenAPITools/generator-core-team  as the changes cover default codegen.

sebastien-rosset:
> cc @OpenAPITools/generator-core-team as the changes cover default codegen.

Thank you William. After thinking about this a bit more, I expect in practice we might have the following needs:
1. Many server-side developers want to disallow additional (undeclared) properties by default. Hence they would want to have `additionalProperties: false` behavior.
1. Many client-side developers want to allow additional (undeclared) properties by default. Hence they would want to have `additionalProperties: true` behavior.

My rationale is this:
1. On the server side, the developers define OAS schemas with a specific set of static properties. If the client sends undeclared properties the request should be rejected because that property simply does not exist. When developers upgrade server-side software, they may also update the OAS schema and add new static properties. They will release the new version of the server, then release the client software. It would be unusual for the client to be at a higher software version than the server.
1. On the client side, I expect it is common to have a client at a lower software version than the server version. That's because engineers may download the generated OAS client, build an application and deploy it. Then as long as the OAS schema maintains backward-compatibility guarantees, the client does not need to be upgraded whenever the server is upgraded. That means the client may start receiving "undeclared" properties (from the perspective of the client) when the server is upgraded, and the client may not care about these properties.

Thoughts?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -969,14 +1007,20 @@ public String toInstantiationType(Schema property) {
...
+        Schema addProps = getAdditionalProperties(schema);
+        if (addProps != null) {
+            // if AdditionalProperties exists, get its datatype and
+            // store it in codegenModel.additionalPropertiesType.
+            // The 'addProps' may be a reference, getTypeDeclaration will resolve
+            // the reference.
+            List<String> referencedModelNames = new ArrayList<String>();
+            codegenModel.additionalPropertiesType = getTypeString(addProps, "", "", referencedModelNames);
+            if (referencedModelNames.size() != 0) {
+                codegenModel.vendorExtensions.put(referencedModelNamesExtension, referencedModelNames);
```
spacether, line 1019: Should we also do `codegenModel.imports.addAll(referencedModelNames);`
Do we have an example that uses this code? Where?

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
```diff
@@ -289,9 +289,8 @@ public void mapModelTest() {
         Assert.assertEquals(cm.classname, "sample.Sample");
         Assert.assertEquals(cm.description, "a map model");
         Assert.assertEquals(cm.vars.size(), 0);
-        Assert.assertEquals(cm.parent, "dict");
-        Assert.assertEquals(cm.imports.size(), 1);
-        Assert.assertEquals(Sets.intersection(cm.imports, Sets.newHashSet("children.Children")).size(), 1);
```
spacether, line 294: Why did we remove the imports check here?

spacether:
@sebastien-rosset do we have to land https://github.com/OpenAPITools/openapi-generator/commit/f57338d2110e0694dec7c0e07c9966c914c74359 before this one?
Can you add a comment in you description listing that blocker.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
sebastien-rosset: > Should we also do `codegenModel.imports.addAll(referencedModelNames);`

Can you explain why? I can do it if you think this is required, I would like to understand why.

> Do we have an example that uses this code? Where?

spacether:
@sebastien-rosset this PR is very big.
Can you break it up into the following pieces to make it easier for me to understand and review?

PR 1: introduce the CLI option disallowAdditionalPropertiesIfNotPresent
- set all generators to disallowAdditionalPropertiesIfNotPresent = True
- update getSchemaAdditionalProperties in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
PR 2
- turn on disallowAdditionalPropertiesIfNotPresent = False for python-experimental
- add python-experimental specific changes

modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-and-additional-properties-for-testing.yaml
```diff
@@ -0,0 +1,2009 @@
+swagger: '2.0'
```
spacether, line 1: Why make this separate yaml file?
To test disallowAdditionalPropertiesIfNotPresent=True/False?
If that is the case, can you make this spec much smaller, give it a file name like:
`issues_1369_1371.yaml` or
`additional_properties_spec.yaml`
and use it in DefaultCodegenTest.java?
We want to minimize the spec additions that we make.

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
sebastien-rosset: > Why did we remove the imports check here?

Because now that composed schemas support the `additionalProperties` keyword, the additional properties are no longer in the parent. The other unit tests cover the scenario of composed schema with additional properties.

sebastien-rosset:
> @sebastien-rosset do we have to land [f57338d](https://github.com/OpenAPITools/openapi-generator/commit/f57338d2110e0694dec7c0e07c9966c914c74359) before this one?
> Can you add a comment in you description listing that blocker.

Before this PR? What do you mean? The commit you are referencing is part of this PR.

modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-and-additional-properties-for-testing.yaml
sebastien-rosset: > Why make this separate yaml file?
> To test disallowAdditionalPropertiesIfNotPresent=True/False?
> If that is the case, can you make this spec much smaller, give it a file name like:
> `issues_1369_1371.yaml` or
> `additional_properties_spec.yaml`
> and use it in DefaultCodegenTest.java?
> We want to minimize the spec additions that we make.

Because initially I modified the existing PR, but it caused a huge number of changes because there are many other languages using the original file. Sure I can make this file smaller.

spacether:
> f57338d

Ah, my mistake. I thought that the commit hash was a PR link.

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
spacether: Okay, but  what should we expect to see in cm.imports here?
An empty list?
If it a list fo length 1, what should it be?
Can you put the assertions back in verifying that the length and value of cm.imports is correct?

sebastien-rosset: It is there: https://github.com/OpenAPITools/openapi-generator/pull/6290/files/89117b9b54879154459c08fe790586cf6e03fbb8#diff-81a58419c148eb8bb7ddfa31ed3f0d2cR292

modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-and-additional-properties-for-testing.yaml
sebastien-rosset: I reduced the size of the file.

sebastien-rosset:
> @sebastien-rosset this PR is very big.
> Can you break it up into the following pieces to make it easier for me to understand and review?
> 
> PR 1: introduce the CLI option disallowAdditionalPropertiesIfNotPresent
> 
> * set all generators to disallowAdditionalPropertiesIfNotPresent = True
> * update getSchemaAdditionalProperties in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
> * any other general changes
> 
> PR 2
> 
> * turn on disallowAdditionalPropertiesIfNotPresent = False for python-experimental
> * add python-experimental specific changes

I am going to push back a bit here, how is this any different from what you have done in your own large PRs?

spacether:
> > @sebastien-rosset this PR is very big.
> > Can you break it up into the following pieces to make it easier for me to understand and review?
> > PR 1: introduce the CLI option disallowAdditionalPropertiesIfNotPresent
> > 
> > * set all generators to disallowAdditionalPropertiesIfNotPresent = True
> > * update getSchemaAdditionalProperties in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
> > * any other general changes
> > 
> > PR 2
> > 
> > * turn on disallowAdditionalPropertiesIfNotPresent = False for python-experimental
> > * add python-experimental specific changes
> 
> I am going to push back a bit here, how is this any different from what you have done in your own large PRs?

Yup, when asked I have broken out separate PRs here.
https://github.com/OpenAPITools/openapi-generator/pull/2049#issuecomment-473761917
But I am not sure that that is relevant.
The difficulty here is that this PR is 362 files and +3,583 lines
This is very hard for me to understand all at one time. This is a big change.
If you need this to be reviewed all at one time it is possible but it may take longer because we will be working on more updates at one time.

So would you prefer to keep it in this one PR knowing that there may be:
- more back and forth
- and more time spent working on it

or would you prefer to split it out to keep the scope of the changes smaller, easier to understand, and easier to review?

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
```diff
@@ -52,11 +53,19 @@ public static OpenAPI parseFlattenSpec(String specFilePath) {
      * @return A "raw" OpenAPI document
      */
     public static OpenAPI parseSpec(String specFilePath) {
-        return new OpenAPIParser().readLocation(specFilePath, null, new ParseOptions()).getOpenAPI();
+        OpenAPI openAPI = new OpenAPIParser().readLocation(specFilePath, null, new ParseOptions()).getOpenAPI();
+        // Invoke helper function to get the original swagger version.
+        // See https://github.com/swagger-api/swagger-parser/pull/1374
+        // Also see https://github.com/swagger-api/swagger-parser/issues/1369.
+        ModelUtils.getOpenApiVersion(openAPI, specFilePath, null);
```
spacether, line 60: Great addition here. This is very helpful, thanks!

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -1539,6 +1543,22 @@ components:
...
+        map_with_additional_properties:
+          type: object
+          additionalProperties: true
+        map_without_additional_properties:
+          type: object
+          additionalProperties: false
+        map_string:
+          type: object
+          additionalProperties:
+            type: string
```
spacether, line 1561: These property names are being added, but their values disagree with their descriptions.
Can you change them to be consistent so our users aren't confused?

- anytype_1 results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` how about naming it map_of_anytype1
- anytype_2 results in `bool, date, datetime, dict, float, int, list, str, none_type` so how about naming it anytype1
- anytype_3 results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` so how about naming it map_of_anytype2
- map_with_additional_properties results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` so how about naming it map_of_anytype3
- map_without_additional_properties results in `bool, date, datetime, dict, float, int, list, str` so how about renaming it anytype2

map_without_additional_properties lacks none_type in its resultant data_type? Is that correct or incorrect?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
```diff
@@ -100,6 +100,8 @@
 
     additional_properties_type = {{#additionalPropertiesType}}({{{additionalPropertiesType}}},)  # noqa: E501{{/additionalPropertiesType}}{{^additionalPropertiesType}}None{{/additionalPropertiesType}}
 
+    _nullable = {{#isNullable}}True{{/isNullable}}{{^isNullable}}False{{/isNullable}}
```
spacether, line 103: Do we have a test where we test this new feature with `assert None == NullableClass(None)`?
Where?
If we don't have it, can you add it?

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
sebastien-rosset: > These property names are being added, but their values disagree with their descriptions.
> Can you change them to be consistent so our users aren't confused?
> 
> * anytype_1 results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` how about naming it map_of_anytype_1

done

> * anytype_2 results in `bool, date, datetime, dict, float, int, list, str, none_type` so how about naming it anytype_1

done

> * anytype_3 results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` so how about naming it map_of_anytype_2

done

> * map_with_additional_properties results in `{str: (bool, date, datetime, dict, float, int, list, str, none_type)}` so how about naming it map_of_anytype_3

done

> * map_without_additional_properties results in `bool, date, datetime, dict, float, int, list, str` so how about renaming it anytype_2

I think it should be name `empty_map`. Since `additionalProperties` is False, undeclared properties are not allowed; and there are no declared properties, hence no properties are allowed. I think the allowed types are set to `bool, date, datetime, dict, float, int, list, str` because there are no restrictions on the type, but since there are no declared and undeclared properties, nothing will be allowed.

> 
> map_without_additional_properties lacks none_type in its resultant data_type? Is that correct or incorrect? Isn't that incorrect?
> 
> * Shouldn't map_without_additional_properties be a map where only certain specific key property names are allowed because additionalProperties is False?

No, I wanted to have an empty map.

> * Do we need to convert that into a a new separate object type CodegenModel which becomes a Python ModelNormal to enforce that behavior?

What is `that`?

sebastien-rosset:
> > > @sebastien-rosset this PR is very big.
> > > Can you break it up into the following pieces to make it easier for me to understand and review?
> > > PR 1: introduce the CLI option disallowAdditionalPropertiesIfNotPresent
> > > 
> > > * set all generators to disallowAdditionalPropertiesIfNotPresent = True
> > > * update getSchemaAdditionalProperties in modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
> > > * any other general changes
> > > 
> > > PR 2
> > > 
> > > * turn on disallowAdditionalPropertiesIfNotPresent = False for python-experimental
> > > * add python-experimental specific changes
> > 
> > 
> > I am going to push back a bit here, how is this any different from what you have done in your own large PRs?
> 
> Yup, when asked I have broken out separate PRs here.
> [#2049 (comment)](https://github.com/OpenAPITools/openapi-generator/pull/2049#issuecomment-473761917)
> But I am not sure that that is relevant.
> The difficulty here is that this PR is 362 files and +3,583 lines
> This is very hard for me to understand all at one time. This is a big change.
> If you need this to be reviewed all at one time it is possible but it may take longer because we will be working on more updates at one time.
> 
> So would you prefer to keep it in this one PR knowing that there may include:
> 
> * more back and forth comments
> * and more time spent working on it
> 
> or would you prefer to split it out to keep the scope of the changes smaller, easier to understand, and easier to review?

It's too much work to split, I will not do that.

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
spacether: I have some more questions here:

map_with_additional_properties results in {str: (bool, date, datetime, dict, float, int, list, str, none_type)} so how about naming it map_of_anytype_3
done.
Update: actually I don't think it's a good suggestion. We would lose the fact that the map allows undeclared properties using multiple variants. I will name it map_with_undeclared_properties_anytype_1/2/3.
There are three properties allowing undeclared properties.

1. What does multiple variants mean?
2. What is the difference between undeclared properties and additionalProperties: True?

The example that you want to name empty_map.
I am okay naming this empty_map, but the python code is not meeting your intention of only allowing an empty map here.

3. How do you propose having the python code only allow empty map for this property?
There are 2 solutions that I see:
a) when this condition is seen (property schema of type object with no properties and additionalProperties: False) in Java we need to set the maxItems for this property to 0. That will only allow empty dictionaries here.
b) create a new CodegenModel form the inline schema with no properties and additionalProperties: False, this is a complicated solution

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
sebastien-rosset: I'm going to cover the composed model in this PR. BTW, this is only relevant to the oneOf/anyOf children, for other cases it's already covered in the existing code and unit tests.

sebastien-rosset: See 8fdbd278fdbdf1da674206cf1ca8d337a7c26d80

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
sebastien-rosset: Separate issue #6400 is tracking this. Marking resolved.

samples/openapi3/client/petstore/python-experimental/tests/test_discard_unknown_properties.py
```diff
@@ -36,51 +36,86 @@
...
+            'Exception message: {0}'.format(str(cm.exception)))
+
+
+    def test_deserialize_isosceles_triangle_do_not_discard_unknown_properties(self):
+        """
+        deserialize IsoscelesTriangle with unknown properties.
+        Strict validation is enabled.
+        Composed schema scenario.
+        """
+        config = Configuration(discard_unknown_keys=False)
```
spacether, line 69: Are unknown_keys the same as additional properties?
How about changing this variable name to discard_additional_properties?

sebastien-rosset: No "unknown key" means the client is sending something unknown (garbage data) to the server. I.e. the property is unknown because it's not explicitly declared in the OAS document. I have added code comments to help. I think it's better handled by having code comments, one other possibility would have been to use the word "garbage" to show it's not supposed to be there, but now with the code comments it should be pretty clear.

sebastien-rosset:
This PR build fails because of unrelated build issue in Java-jersey2. That build issue should be fixed with #6413

samples/openapi3/client/petstore/python-experimental/test/test_fruit.py
```diff
@@ -189,5 +189,24 @@ def testFruit(self):
             fruit._additional_properties_model_instances, []
         )
 
+    def testFruitNullValue(self):
+        # Since 'apple' is nullable, validate we can create an apple with the 'null' value.
+        apple = petstore_api.Apple(None)
+        self.assertIsNone(apple)
+
+        # But 'banana' is not nullable.
+        banana = petstore_api.Banana(None)
```
spacether, line 198: @sebastien-rosset 
please take a look at a branch that I made [here](https://github.com/spacether/openapi-generator/tree/all-of-additional-properties_jblack) which updates `banana = petstore_api.Banana(None)` to raise an exception and has this test assert that the exception is raised.
Which do you prefer to get this working?
- merging all-of-additional-properties_jblack into this PR or
- having me build a separate PR which implements _nullable?

If we make a separate _nullable PR, then I would prefer that we
- land the _nullable PR before this one if this contains _nullable code or
- remove the _nullable code from this one and land them separately with no requirement on which lands first

sebastien-rosset: Merged.

spacether: Thank you for the merge!

PR #6303, opened 2020-05-14
A single oneOf-referred schema is not a parent

boxdot:
A schema referring another single schema in oneOf is not its child,
and the latter is not the parent of the former. This assumption
prevented population of `schema.vars` from the referred schema.
And this in turn yielded invalid generated model for the oneOf
schema.

Also see the example provided as a test.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -1212,8 +1212,9 @@ public class ModelUtils {
             }
         }
 
-        // parent name only makes sense when there is a single obvious parent
-        if (refedWithoutDiscriminator.size() == 1) {
+        // parent name only makes sense when there is a single obvious parent in not 'oneOf' case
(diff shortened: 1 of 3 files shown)


wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1212,8 +1212,9 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
             }
         }
 
-        // parent name only makes sense when there is a single obvious parent
-        if (refedWithoutDiscriminator.size() == 1) {
+        // parent name only makes sense when there is a single obvious parent in not 'oneOf' case
+        boolean isOneOf = composedSchema.getOneOf() != null && !composedSchema.getOneOf().isEmpty();
+        if (!isOneOf && refedWithoutDiscriminator.size() == 1) {
```
spacether, line 1217: What about if our schema is length one eith anyOf only?

PR #6308, opened 2020-05-15
[typescript] Remove "v4-compat" value of enumSuffix

amakhrov:
In https://github.com/OpenAPITools/openapi-generator/pull/5138 , a new `enumSuffix` option was introduced. Before that the logic for generating enum names was inconsistent across generators / different modes. So in order to keep backward compatibility, a special `v4-compat` value for this option was introduced (and used by default) to simulate the original behavior as of pre v4.2.3 .

Now, for 5.0 release we can get rid of this v4-compat mode. Consumers can still get the original 
behavior by setting `enumSuffix` to the value they need in their project.

The new default value of `enumSuffix` is `"Enum"`.
It results in minimal changes in the generated samples - most of the sample projects were already configured in a way that emited the same enum names as with the new default.

### PR checklist


macjohnny:
@amakhrov thanks for the PR.
could you please update the `docs/generators/javascript-flowtyped.md`?

amakhrov:
@macjohnny - done!

PR #6315, opened 2020-05-15
[go-experimental] Ensure enum deserialization checks enum values

bkabrda:
This new unmarshalling method for enums ensures that unmarshalling fails if the value is not in the enumeration choices. As the enum type is just a type alias in go, technically any value of the base type can actually be assigned to it - this patch prevents that at least on unmarshalling.

This allows to properly deserialize `oneOf` members that have a common field that has specific distinct choices for each of the members. For example:

* `A` and `B` are members of `oneOf` and both have a `foo`
* For `A`, `foo` is an enum with allowed choices `a`, `A`
* For `B`, `foo` is an enum with allowed choices `b`, `B`
* The rest of the fields is potentially the same.

Having this new unmarshalling method ensures that entity with `foo: b` can't get deserialized as `A`, but only as `B`, hence the `oneOf` deserialization works properly.

### PR checklist

### DIFF
--- a/samples/client/petstore/go-experimental/go-petstore/model_enum_class.go
+++ b/samples/client/petstore/go-experimental/go-petstore/model_enum_class.go
@@ -13,6 +13,10 @@ import (
 	"encoding/json"
 )
 
+import (
+    "fmt"
+)
+
 // EnumClass the model 'EnumClass'
 type EnumClass string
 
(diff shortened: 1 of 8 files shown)


samples/client/petstore/go-experimental/go-petstore/model_enum_class.go
```diff
@@ -13,6 +13,10 @@ import (
 	"encoding/json"
 )
 
+import (
+    "fmt"
+)
+
```
wing328, line 19: The change looks good to me. I'll try to consolidate the multiple `import` later with another PR.

PR #6322, opened 2020-05-15
[Java][jersey2] Allow unknown properties returned by server.

bkabrda:
This PR makes it possible to have server return properties that are unknown/unspecced. I find this very useful, as having this disabled effectively makes a client library break whenever there's a new feature/field added on server side.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/JSON.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/jersey2/JSON.mustache
@@ -27,7 +27,7 @@ public class JSON implements ContextResolver<ObjectMapper> {
   public JSON() {
     mapper = new ObjectMapper();
     mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, true);
+    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     mapper.configure(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE, true);
     mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
     mapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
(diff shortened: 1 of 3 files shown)


wing328:
> This PR makes it possible to have server return properties that are unknown/unspecced. I find this very useful, as having this disabled effectively makes a client library break whenever there's a new feature/field added on server side.

Agreed this could be useful but at the same time, I think some users prefer an exception if the payload contains unknown properties so it looks like we need another option to cater to both cases? (another way is to customize the template instead to avoid adding another option...)

Also I think the current logic to handle oneOf/anyOf needs to be revised as it relies on exceptions to determine whether the deserialization is successful or not.

bkabrda:
> Also I think the current logic to handle oneOf/anyOf needs to be revised as it relies on exceptions to determine whether the deserialization is successful or not.

Why is that a problem? There's an exception raised right now if any entity returned by the API is not deserialized successfully, so why would we want to treat `oneOf/anyOf` in a different way?

wing328:
> Do you mean the fact that we basically try/catch deserialization of every oneOf choice to determine which one succeeds?

Yes. `discriminator` is optional so approach doesn't rely on it.

wing328:
Just ran some tests locally with a spec containing `oneOf` and got the errors:
```
[ERROR] org.openapitools.client.api.DefaultApiTest.oneofdogGetTest  Time elapsed: 1.569 s  <<< ERROR!
org.openapitools.client.ApiException: Response body is invalid as it matches more than one schema (Cat, Dog) defined in the oneOf model: org.openapitools.client.model.OneOfResponse
	at org.openapitools.client.api.DefaultApiTest.oneofdogGetTest(DefaultApiTest.java:66)

[ERROR] org.openapitools.client.api.DefaultApiTest.oneofcatGetTest  Time elapsed: 1.56 s  <<< ERROR!
org.openapitools.client.ApiException: Response body is invalid as it matches more than one schema (Cat, Dog) defined in the oneOf model: org.openapitools.client.model.OneOfResponse
	at org.openapitools.client.api.DefaultApiTest.oneofcatGetTest(DefaultApiTest.java:55)
```
which is expected as it's now matching both Cat and Dog in the oneOf's list of schemas.

wing328:
@bkabrda can you please give another try with the latest master, which has better support for `additionalProperties`?

Happy to reopen this PR if it's still needed.

PR #6331, opened 2020-05-16
[PHP] ObjectSerializer fix for array of objects

alxnegrila:
Array of objects translate to **map[string,object][]** and they fail to deserialize. This is because the deserialization does not parse the mapping string correctly. A quick fix is moving array deserialization before object deserialization.

```
    ObjectExample:
      type: object
      properties:
        data:
          type: array
          items:
            type: object
            additionalProperties: true
```

@jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @ybelenko @renepardon

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -250,6 +250,14 @@ class ObjectSerializer
     {
         if (null === $data) {
             return null;
+        } elseif (strcasecmp(substr($class, -2), '[]') === 0) {
+            $data = is_string($data) ? json_decode($data) : $data;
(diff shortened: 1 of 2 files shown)


wing328:
Please run `./bin/php-petstore.sh` or `bin\windows\php-petstore.bat` to update the PHP samples so that CIs can test and verify the change.

alxnegrila:
> Please run `./bin/php-petstore.sh` or `bin\windows\php-petstore.bat` to update the PHP samples so that CIs can test and verify the change.

updated the samples

wing328:
The fix looks good to me. I'll update samples/openapi3/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php in a separate commit.

Issue #5171, opened 2020-01-31
[BUG][typescript-angular] allOf without discriminator results in unnecessary imports


amakhrov:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?

##### Description

When a model uses `allOf`, where some of the referenced models do not have a discriminator (hence, do not denote any hierarchy), the target interface just gets all properties copied from the referenced models, no interface inheritance/extending involved.

However, the base interfaces are still imported into the file with the target model - despite being unused.

##### openapi-generator version

Latest

##### OpenAPI declaration file content or url
```yaml
definitions:
  ExtraFields:
    type: object
    properties:
      extra_field:
        type: string

  ResponseOnePlus:
    type: object
    allOf:
      - $ref: '#/definitions/ExtraFields'
      - type: object
        properties:
          own_field:
            type: string
```

##### Command line used for generation

```
docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate \
    -i /local/spec.yaml \
    -g typescript-angular \
    -o /local/generated
```

##### Steps to reproduce

- Run a command above, providing a spec file with allOf
- Check `model/responseOnePlus.ts`.It will looks like this:
```
import { ExtraFields } from './extraFields';
import { ResponseOnePlusAllOf } from './responseOnePlusAllOf';

export interface ResponseOnePlus { 
    extraField?: string;
    ownField?: string;
}
```
Both imports here are not used.

##### Related issues/PRs

Similar issues:
- `typescript-axios`: https://github.com/OpenAPITools/openapi-generator/issues/4804
- `typescript-rxjs`: https://github.com/OpenAPITools/openapi-generator/issues/4629

##### Suggest a fix

The fixes for linked axios and rxjs generators above change the approach for generating a target interface: instead of copying all properties from the source interfaces into the target one, the target type is simply defined as type intersection.

Perhaps we should consider doing the same for typescript-angular (and probably other typescript generators that emit each model in a separate file with imports)

PR #6335, opened 2020-05-16
typescript-axios anytype is not defined

codeserk:
Proposed solution for this issue:
[bug typescript `AnyType` is not defined #6332](https://github.com/OpenAPITools/openapi-generator/issues/6332)

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -845,6 +845,11 @@ public String toAllOfName(List<String> names, ComposedSchema composedSchema) {
             }
             return schemaType;
         }).distinct().collect(Collectors.toList());
-        return String.join(" & ", types);
+
+        if (types.size() > 1 && types.contains("any")) {
```
amakhrov, line 849: nit: it might be simpler to add `.filter()` call to the stream instead of post-processing the resulting list

codeserk:
@amakhrov

🤔 actually I think the root of the problem is in the methods that get types for `allOf` / `oneOf` / `anyOf`, since they return `AnyType` when they don't find any valid type (`properties`, `ref` or `type` are not set, only `description` in this case), while I think they should ignore those lines.
I'll dig a bit more into that and I'll also check why so many tests failed!

Also, I'll check what happens when one of the types is a `file` or anything that is translated to `any` in typescript, that might lead to some trouble.

I'll keep you updated, but maybe is safer to first just add the translation for `AnyType` in a first iteration and investigate the root problem further. 
With that fix the generated type would be `Adult & any` which means you probably can't get type information in the IDE (at least not in VSCode), but there won't be any type error, which was the main issue :D

codeserk:
@amakhrov

I've pushed some more changes: Now I'm filtering out schemas that are translated to `AnyType` in the parent, instead of all the types that are `any` in typescript (because those include types like `filter`, `map`, which we don't want to ignore).

I've also extended the sample, to include `allOf`, `anyOf` and `oneOf`. It seems to work as expected: `any` is only used for `file`, but not when fields like `description` are part of the composed schema.
Example:
image
^ Left: new branch, Right: without changes in `AbstractTypeScriptClientCodegen.java` - notice that `AnyType` is not necessary here (and it's also causing type errors)

amakhrov:
@codeserk could you please explain the reasoning behind filtering out AnyType from combined schemas?

codeserk:
@amakhrov yes, I tried to explain in the previous comment 😅 

The method that obtains types from the schemas is unable to determine the right type when this is not present (there is no `type` nor `ref` present in the schema definition). This happens in `composed schemas` where some lines just add `description`, `example`, etc. They add complementary information but not type:
- [Example for `allOf`: adding description to `ref` and `object` properties](https://github.com/OpenAPITools/openapi-generator/pull/6335/files#diff-1a56b6da00296924fc8b999247dbdcf0R70) - The resulting type included `& AnyType` before the changes, caused by the line that adds description.
- [Example for `allOf`: adding description to `file` type](https://github.com/OpenAPITools/openapi-generator/pull/6335/files#diff-1a56b6da00296924fc8b999247dbdcf0R56) - The type for the file was `file?: any | AnyTpe` before the changes, caused by the line that adds description.
- [Example for `anyOf`: adding description when you have multiple `refs`](https://github.com/OpenAPITools/openapi-generator/pull/6335/files#diff-1a56b6da00296924fc8b999247dbdcf0R39) - The parameter was `petByAgePetByTypeAnyType?: PetByAge | PetByType | AnyType` before the changes, caused by the line that adds description.
- [Example for `oneOf`: adding `description` to a list of `refs`](https://github.com/OpenAPITools/openapi-generator/pull/6335/files#diff-1a56b6da00296924fc8b999247dbdcf0R22)  - The parameter was `catDogAnyType?: Cat | Dog | AnyType` before the changes , caused by the line that adds description.

These are all valid OpenAPI v3 definitions, and the result was wrong in all 4 because `AnyType` should not be present, that line in the schema was indicating a description, not adding another type to the composed schema.

----
Notice that:
- `any` was not simply removed: There are some types that end in `any` and they are preserved (there's an example for `file` type that ends in `any`, and the same would happen with `map` and some other types.
- `AnyType` is not removed if that's the only type extracted (I guess in composed schemas without any type information)
- There is an active bug related to `AnyType` not being translated to `any` so I think `AnyType` is only used in the cases where the type cannot be determined (which should be few cases (?)).

Hope this clarifies it a bit! 
and thanks for your time reviewing :)

amakhrov:
Sounds good, thanks for clarifying!

I feel having indeterminate model in api spec, that results in AnyType, as a part of a combined schema is not something common. Or, frankly, something good at all :).

samples/client/petstore/typescript-axios/builds/composed-schemas/api.ts
```diff
@@ -0,0 +1,420 @@
...
+     * @memberof Cat
+     */
+    age?: number;
+}
+/**
+ * 
+ * @export
+ * @interface CatAllOf
+ */
+export interface CatAllOf {
```
macjohnny, line 46: where is this interface used?

codeserk: I think it's not used anywhere, 
I've also noticed that, but testing without the changes in this branch generates the same unused interfaces - so I think it's another bug unrelated to this branch

amakhrov: It's probably a manifestation of this old bug: https://github.com/OpenAPITools/openapi-generator/issues/5171

codeserk:
@amakhrov @macjohnny 
Hello, what are the next steps? Do you still see problems in this PR? I so, please let me know so I can fix them :)
Thanks!

PR #6341, opened 2020-05-17
[TypeScript] Rewritten TypeScript client generator supporting fetch & jquery

### LINKED ISSUE #802: Rewrite of TypeScript Generators

TiFu:
This PR introduces a new experimental generator for TypeScript. This generator is based on the rewrite started in #802 and separates the framework-independent code from the framework-specific code.  
fixes #802

At this time, the generator only supports the fetch API and JQuery. We hope that finally merging this generator will accelerate the integration of additional frameworks like axios, etc. and further integration of additional features.

Credits:
@bodograumann Thank you a lot for all your help and all the work you put into this generator!

@Core Team Members:
Could you please provide some feedback on changes to the core java classes in this PR?

### PR checklist

CC TS Committee @taxpon (2017/07) @sebastianhaas  (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny  (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11)  @amakhrov (2020/02)

CC Core Members  @wing328 (2015/07) @jimschubert (2016/05) @cbornet (2016/05) @ackintosh (2018/02) @jmini (2018/04) @etherealjoy (2019/06) @spacether (2020/05)


wing328:
@TiFu 👍 I'll try to review and give feedbacks by coming Friday.

TiFu:
Awesome, thanks! :)

amakhrov:
@TiFu I'm curious what is the motivation for creating a brand new java class (which seems to largely duplicate AbstractTypescriptClient) instead of extending latter?

amakhrov:
And one more, ver different thought:
Historically, typescript generators emitted code for publishing stuff to npm. Since you're introducing a brand new generator with no need for backward compatibility, perhaps it's a good chance to revise that approach.
Npm scripts generally do not depend on the api spec. It feels like a whole separate thing to care about in the generator, in addition to generating valid TS that conforms to the api spec.

TiFu:
> @TiFu I'm curious what is the motivation for creating a brand new java class (which seems to largely duplicate AbstractTypescriptClient) instead of extending latter?

The main motivation was being completely independent of what we already use for the other generators - ideally simplifying the generator class. This also allows us to make any changes we deem necessary without having to worry about causing conflicts with the old generators. 




> And one more, very different thought:
> Historically, typescript generators emitted code for publishing stuff to npm. Since you're introducing a brand new generator with no need for backward compatibility, perhaps it's a good chance to revise that approach.
> Npm scripts generally do not depend on the API spec. It feels like a whole separate thing to care about in the generator, in addition to generating valid TS that conforms to the API spec.
> 
> Would love to learn what you think about it

I personally would like to keep it for two reasons
a) It makes consuming the API simpler - you can just generate the API, push it to the registry (be it private or public), and use it. Additionally, you get properly tagged versions without any additional work.
b) This might be subjective, but from my view maintaining these scripts wasn't a whole lot of additional work. (Please correct me if I'm wrong)

I don't know if we currently support this, but we could add a flag that disables (or enables) the generation of the npm-related code.

bodograumann:
One advantage of generating an npm package is that you get the necessary dependencies right away.
Iirc when I started using the typescript-inversify generator, I did not know that it could generate a package, given the correct config, and it took me a while to figure out the correct dependency setup.
So having a switch seems fine to me, as long as we document the dependencies properly when it is disabled.

Btw. there already seem to be some conflicts between my changes and the master branch (even though we recently rebased on master). Hope it's not too hard to resolve.

TiFu:
Yay conflicts :confetti_ball: I'll merge #6177 first and then fix the conflicts tomorrow or so.

TiFu:
`drone.io` failure is unrelated to this PR.

wing328:
> @wing328 Is there any general rule in regards to supporting OAI v3 vs v2? Is OAI v3 completely supported now?

OAS v3 has better support for oneOf, anyOf, inheritance, model compositions, etc so it's what we advise users to use.

In other generators, we've not yet encountered use cases in which OAS v2 doesn't work but the converted OAS 3 spec works. Let us know if you guys encounter such use cases and I'll take a deeper look before suggesting how to deal with it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3799,6 +3800,7 @@ public CodegenResponse fromResponse(String responseCode, ApiResponse response) {
                 default:
                     throw new RuntimeException("Invalid response code " + responseCode);
             }
+            r.isSuccessCode = r.code.startsWith("2");
```
wing328, line 3803: Heads-up: later we may switch it to something like r.is2XXCode as 2 or 3 other generators need is4XXCode and is5XXCode as well (and probably is3XXCode)

TiFu: Good catch! That's actually already the case - see the `switch` statement right above this line. I'll fix it later today.

TiFu:
@bodograumann What do you think?

We currently test only against OAIv2 - Should we support both or just make a switch to testing against v3?

bodograumann:
Testing with the v3 spec of petstore sounds like a good idea.
I don't really know how much we gain in coverage, as the petstore api is still limited, but the switch should not be difficult.
Personally I only use v3 api specs so there is that :-)

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
```diff
@@ -0,0 +1,246 @@
+{{#platforms}}
+{{#node}}
+// TODO: evaluate if we can easily get rid of this library
+import * as FormData from "form-data";
+{{/node}}
+{{/platforms}}
+// typings of url-parse are incorrect...
+// @ts-ignore 
+import * as URLParse from "url-parse";
```
macjohnny, line 9: is this really necessary in the browser?

modules/openapi-generator/src/main/resources/typescript/auth/auth.mustache
```diff
@@ -0,0 +1,152 @@
+import {RequestContext} from '../http/http';
+// typings for btoa are incorrect
+//@ts-ignore
+import * as btoa from "btoa";
```
macjohnny, line 4: is this necessary in the browser?

macjohnny:
@TiFu thanks a lot for your effort!

modules/openapi-generator/src/main/resources/typescript/api/api.mustache
```diff
@@ -0,0 +1,212 @@
...
+     * {{&notes}}
+     {{/notes}}
+     {{#summary}}
+     * {{&summary}}
+     {{/summary}}
+     {{#allParams}}
+     * @param {{paramName}} {{description}}
+     {{/allParams}}
+     */
+    public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}options?: Configuration): RequestContext {
```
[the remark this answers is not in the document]
bodograumann: I get your point, but personally I would prefer the current structure.
Ordering or naming problems where never a problem in my experience.
Also, calling with an object would make usage slightly more complicated, especially with few arguments.

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
bodograumann: In most browsers we probably could use URL.searchParams, except for IE.
This way we can use the same for all environments.
Not sure if there is a URL.searchParams polyfill for both IE and node.

modules/openapi-generator/src/main/resources/typescript/auth/auth.mustache
bodograumann: The polyfill would be needed in IE ≤9 and Opera ≤ 10.1.
I would be fine with only including it in node context.
What do you think, @TiFu ?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
```diff
@@ -0,0 +1,865 @@
...
+        for (String option: TypeScriptClientCodegen.PLATFORMS) {
+            // TODO: improve description?
+            platformOption.addEnum(option, option);
+        }
+        platformOption.defaultValue(PLATFORMS[0]);
+
+        cliOptions.add(platformOption);
+
+
+        // TODO: gen package.json?
```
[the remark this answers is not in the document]
bodograumann: I would say no, we can already generate the package.json.

modules/openapi-generator/src/main/resources/typescript/api/api.mustache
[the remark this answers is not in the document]
macjohnny: combining the params into one single object can be very helpful when e.g. the api is refactored or extended, so that the compiler will immediately spot naming mismatches, even if the types would match.
I guess it would be best to provide an option for that, see e.g. 
https://github.com/OpenAPITools/openapi-generator/pull/6288
https://github.com/OpenAPITools/openapi-generator/pull/4479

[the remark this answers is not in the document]
bodograumann: That sounds good.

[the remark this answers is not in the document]
TiFu: @bodograumann Are you already working on this? If not I'll give it a shot tomorrow.

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
TiFu: There's `https://www.npmjs.com/package/url-search-params-polyfill` which works for both IE and node, has no dependencies, and actually checks if `URL.searchParams` is available before using its own code. 

Alternatively, we could also just expect people to provide their own polyfill for IE, i.e. we only use URLParse in node and assume that URLParse is available in the browser

Which option would you prefer?

modules/openapi-generator/src/main/resources/typescript/auth/auth.mustache
TiFu: Agreed - I'll exclude it

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
TiFu: Edit: It would be a giant pain to re-implement this without a polyfill - Given that `url-parse` has exactly 2 dependencies (including recursive dependencies), I'd propose to just use url-parse.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
```diff
@@ -0,0 +1,865 @@
...
+    private String getReturnType(List<CodegenResponse> responses) {
+        StringBuilder returnType = new StringBuilder();
+        boolean firstReturnType = true;
+        boolean atLeastOneSuccess = false;
+        boolean addVoid = false;
+        for (CodegenResponse response: responses) {
+            // TODO: we should probably catch an exception here
+            if (response.is2xx) {
+                if (response.dataType != null) {
+                    if (!firstReturnType) {
```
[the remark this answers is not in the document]
TiFu: Replaced with a `Set` and simplified the code

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
```diff
@@ -0,0 +1,865 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ * Copyright 2018 SmartBear Software
```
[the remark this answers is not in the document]
TiFu: Most of this file is copied from `AbstractClientCodegen` - so I believe we would have to at least maintain the SmartBearSoftware notice. 

I have updated the year for OAI-Gen Contributors though. 

Is there anything else which I should change?

modules/openapi-generator/src/main/resources/typescript/api/api.mustache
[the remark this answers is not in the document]
bodograumann: No, I'm not working on this. Please go ahead :-)

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
bodograumann: I think it is fine for now, we can always improve this later, by using `url-serach-params-polyfill` instead and maybe add a configuration option to enable polyfills or not.

TiFu:
@macjohnny Thanks for the feedback!

macjohnny:
@TiFu do you think we should merge it and see how it goes?

TiFu:
I'd be fine with that. @bodograumann 's inversify PR is still open but we can merge that later without any issues.

bodograumann:
@TiFu Thank you for merging my PR.
I think with the Merge master afterwards, you lost my change to `CONTRIBUTING.md` though.

I feel adding this notice is important, because until I found out how to regenerate the samples without recompiling the java packages, I lost many increments of 3min for compiling during development. It was especially annoying when I had to go back and forth between templates and samples multiple times until I go things right.

Cf. https://github.com/OpenAPITools/openapi-generator/pull/6489/commits/abe3aa7b65a883c6792f25dd770510f211000e65 and https://github.com/OpenAPITools/openapi-generator/pull/6489/commits/41f5e3e6713fe2c1db03c6169805f2027e60d6d0.

TiFu:
Ah, sorry for that! I re-added the changes.

TiFu:
@macjohnny I think we are good and should be able to merge into master without problems. Do you want to do the merge or should I do it this Friday?

TiFu:
@rgomezp See #10283 for an initial draft. Happy to get your feedback and understand a bit more in what form the documentation would be most helpful.

PR #6351, opened 2020-05-18
Fix #6350 by removing if statement with arbitrary condition

### LINKED ISSUE #6350: [BUG] [haskell-http-client] Using a schema reference for an optional form parameter of a request body results in broken code
#### Bug Report Checklist

- [x] Provided a broken spec.

##### Description

Using a schema reference for an optional form parameter of a request body results in broken code.

I expected compiling code.

From Commit 6c5621d97e65a1b635eee5c0830017fa54a68265 (HEAD of master at the moment)

```yaml
openapi: 3.0.2
info:
  title: Example
  description: Example to demonstrate issues with optional parameters for haskell-http-client
  contact:
    name: example
    url: https://www.example.example
    email: example@test.test
  license:
    name: whateveryouneed
    url: https://www.example.example
  version: 0.0.1
servers: []
paths:
  /example:
    post:
      tags:
        - example
      description: Example
      operationId: exampleOp
      requestBody:
        description: example
        required: true
        content:
          'application/x-www-form-urlencoded':
            schema:
              type: object
              properties:
                example:
                  $ref: '#/components/schemas/Example'
      responses:
        '204':
          description: Success
components:
  schemas:
    "Example":
      title: ExampleEnum
      type: string
      enum:
        - "Value1"
        - "Value2"
        - "Value3"
      description: ''
```

    openapi-generator-cli generate -o test -i example.yaml -g haskell-http-client

##### Steps to reproduce

    cd test
    cabal build

##### Related issues/PRs

none to my knowledge

##### Suggest a fix

pr incoming. There is a completely arbitrary if condition in the code, with no else branch
and the else case leaves some vendor extensions thingies uninitialised.

typetetris:
else branch was also missing. Not doing something in the else
branch simply resulted in empty strings put in the templates,
which resulted in broken haskell code which did not compile.

Fixes #6350
### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/HaskellHttpClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/HaskellHttpClientCodegen.java
@@ -757,21 +757,16 @@ public class HaskellHttpClientCodegen extends DefaultCodegen implements CodegenC
                 op.vendorExtensions.put(VENDOR_EXTENSION_X_HAS_OPTIONAL_PARAMS, true);
             }
 
-            if (typeMapping.containsKey(param.dataType)
-                    || param.isMapContainer || param.isListContainer
-                    || param.isPrimitiveType || param.isFile || param.isEnum) {
+            String dataType = genEnums && param.isEnum ? param.datatypeWithEnum : param.dataType;
 
-                String dataType = genEnums && param.isEnum ? param.datatypeWithEnum : param.dataType;
+            String paramNameType = toDedupedModelName(toTypeName("Param", param.paramName), dataType, !param.isEnum);
+            param.vendorExtensions.put(X_PARAM_NAME_TYPE, paramNameType); // TODO: 5.0 Remove
+            param.vendorExtensions.put(VENDOR_EXTENSION_X_PARAM_NAME_TYPE, paramNameType);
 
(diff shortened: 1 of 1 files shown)


wing328:
@typetetris thanks for the PR

cc @jonschoning

jonschoning:
I added this condition when including support for enums; though I'm not sure i remember exactly the cases when it was necessary - it might be something around aliases.  I'll do some testing on this.

thanks for the pr

jonschoning:
removing this changes the behavior somewhat, unfortunately. I'll see if I can find a fix for your problem though.

https://github.com/jonschoning/openapi-generator/commit/eab52227ea993df8cb286dd03201430bd2ead20b

jonschoning:
@typetetris What the if statement is doing is deciding if a newtype should be generated for the parameter or not.  The mustache templates need newtypes if it's a primitive or not a body parameter, but the assumption is that we won't generate a newtype for absolutely every thing, notably required body parameters that are model objects.

There were a few issues here:

1) The core code was is telling the haskell generator that this parameter is not an enum, when it clearly is an enum. 
We see that `param.isEnum == false` but also `param.allowableValues != null` which doesn't make sense, because there should only be allowableValues when it is an Enum.  cc @wing328

2) the mustache templates made too many assumptions about what was a newtype and what wasn't, so i updated that.

I added a PR which fixes fixes 6350 @ https://github.com/OpenAPITools/openapi-generator/pull/6383/files

jonschoning:
fixed in https://github.com/OpenAPITools/openapi-generator/pull/6383

Issue #841, opened 2018-08-18
Decouple generation workflow from DefaultGenerator and CodegenConfig types


jimschubert:
##### Description

Some discussion around this exists in #503. I'm copying here for context.

---

Conceptually, we have the following concerns:

* Generation workflow
* Templating engine
* Global Options
* Framework generator
* Language definition
* Generator Options
* Spec loading
* Spec parsing
* API definition ADT (e.g. CodegenModel)
* Template object ADT (currently, a map)
* potentially others I can't think of at the moment

Everything above is rolled into `DefaultGenerator` and `DefaultCodegen` (or derived class). There's very little reusability. Where we have reusability at the language level, it is done as an abstract base class. This is problematic because changes in the base class for a target framework (you may be asking why framework logic exists in a language base class…) inadvertently introduces a bug in other generators sharing the same base class.

Our current design makes a refactor a little more difficult because the entry point into the process is DefaultGenerator, which holds a reference to the generator (CodegenConfig/DefaultCodegen instance). We use a form of delegation from the generator to the "config" component at almost every step of the workflow. This is intimidating for new contributors because it's not only undocumented, but is hard to follow even while debugging.

---

There are cases where our workflow could be simplified by sharing the _process_ but offloading language-specific or framework-specific modifications off to the `CodegenConfig` classes.

In other cases, users may not want the processing workflow to be done in the way as it is in `DefaultGenerator`. There's no clean way to provide any generator _other_ than the default, so it seems like the code diverged at some point from its originally intended design.

##### Suggest a fix/enhancement

This will be a major breaking change. But I recommend that we do the following:

## Part 1 Renames

* Rename `Generator` to `Workflow`
* Rename `AbstractGenerator` to `AbstractWorkflow`
* Rename `DefaultGenerator` to `DefaultWorkflow`
* Rename `CodegenConfig` to `Generator`
* Rename `DeafultConfig` to `GeneratorBase`
* Rename `ClientOpts` to `WorkflowOptions`
* Rename or remove `ClientOptInput` (type is not really needed, see "New Functionality" below)

## Part 2 Refactors

* Add a property to `GeneratorBase` (prev `DefaultConfig`) which exposes a `Worfkflow` (prev `Generator`).
  - This allows each generator to have control over its desired workflow.
  - This allows users to extend existing or new generators with their own new or derived workflows
[...]

PR #6357, opened 2020-05-18
[core] Refactor templating management

jimschubert:
This refactors template management to get logic out of DefaultGenerator
and to provide a cleaner API to template search and read/compile.

Deprecates MockDefaultGenerator, which is not a mock and causes
in-memory retention of file contents. Maintainers should prefer
executing a "dryRun" with new DefaultGenerator(true) or do true
mock/spies if evaluating template intermediaries is truly necessary.
Tests may read written files with lower overhead than the in-process
retention of those bytes.

This attempts to maintain some compatibility with existing templating
adapter interfaces. Any breaking change here would be unintentional but
minimal effort to retarget the new interface.

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/AbstractGenerator.java
```diff
@@ -1,213 +0,0 @@
...
-import org.slf4j.LoggerFactory;
-
-import java.io.*;
-import java.nio.file.Paths;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Scanner;
-import java.util.regex.Pattern;
-
-public abstract class AbstractGenerator implements TemplatingGenerator {
```
jimschubert, line 36: FYI breaking change, but we should really try sticking to interfaces and move away from abstract base types.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -163,7 +163,6 @@ apiTemplateFiles are for API outputs only (controllers/handlers).
     protected Map<String, String> reservedWordsMappings = new HashMap<String, String>();
     protected String templateDir;
     protected String embeddedTemplateDir;
-    protected String commonTemplateDir = "_common";
```
jimschubert, line 166: this is now handled by template loading functionality, as that is not the responsibility of DefaultCodegen.

modules/openapi-generator/src/test/java/org/openapitools/codegen/TestUtils.java
```diff
@@ -83,31 +88,32 @@ public static OpenAPI createOpenAPIWithOneSchema(String name, Schema schema) {
...
 
+    /**
+     * Extract file from {@link MockDefaultGenerator}
+     *
+     * @param generator Generator
+     * @param root root path
+     * @param filename filename under root
+     *
+     * @return a {@link WrittenTemplateBasedFile}
+     * @deprecated Since 5.0. Please avoid this method and usage of {@link MockDefaultGenerator}, prefer {@link DefaultGenerator#DefaultGenerator(Boolean)} with dryRun=true.
```
jimschubert, line 99: `MockDefaultGenerator` duplicates all written bytes. `getTemplateBasedFile` is used to "inspect" data passed to templates, but this should be tested differently in the future.

jimschubert:
Note on breaking change without fallback… removal of `AbstractGenerator` has no fallback abstract class and requires template management construction similar to what is done in this PR.

jimschubert:

jimschubert:
@spacether would you like to take a look before merge?

cc @Fjolnir-Dvorak (who's waiting on this work)

jimschubert:
@wing328 I'd created a GitHub Workflow to cross-compile in linux and Windows, then run a smoke test for the gradle build of the go client which was failing on AppVeyor. This allowed me to finish up more quickly as I was seeing 1.5 hours in queue and 30-45 minutes on AppVeyor. We may want to consider a cross-compile job as part of our PRs for quicker feedback.

Here's the workflow:

```
name: Build
on:
  push:
    branches:  
      - master
  pull_request:
    branches:
      - master

jobs:
  build:
    name: Build on JDK ${{ matrix.java }} and ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        java: [8, 11]
        os: [ubuntu-latest, windows-latest]
        
    steps:
      - name: Check out code
        uses: actions/checkout@v2

      - name: Set up JDK ${{ matrix.java }}
        uses: actions/setup-java@v1 
        with:
          java-version: ${{ matrix.java }}
          
      - uses: actions/cache@v1
        with:
          path: ~/.m2/repository
          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
          restore-keys: |
            ${{ runner.os }}-maven-
        
      - name: Build with Maven
        shell: bash
        run: mvn -nsu -B --quiet -Djacoco.skip=true -Dorg.slf4j.simpleLogger.defaultLogLevel=error --no-transfer-progress clean install --file pom.xml ${{ matrix.flags }}
        
      - name: Test gradle
        shell: bash
        run: gradle -b modules/openapi-generator-gradle-plugin/samples/local-spec/build.gradle buildGoSdk --stacktrace
```

image

bin/ci/java-feign.json
```diff
@@ -3,7 +3,7 @@
   "generatorName": "java",
   "inputSpec": "modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml",
   "outputDir": "samples/client/petstore/java/feign",
-  "templateDir": "modules/openapi-generator/src/main/resources/Java/libraries/feign",
```
jimschubert, line 6: Note these removals of `libraries/<library name>` in the Java sample configs and scripts.

While this syntax works if the entirety of the library code is in the named folder, this would cause fallbacks to fail if anyone were to remove something like `model.mustache` intending to fallback to `resources/Java/model.mustache`. We should avoid this possible confusion especially when these subdirectories are structured to follow our library lookup logic:

* User customized library path (e.g. `custom_template/libraries/feign/model.mustache`)
* User customized generator top-level path (e.g. `custom_template/model.mustache`)
* Embedded library path (e.g. `resources/Java/libraries/feign/model.mustache`)
* Embedded top-level path (e.g. `resources/Java/libraries/feign/model.mustache`)
* Common embedded path (e.g. `resources/_common/model.mustache`)
* throw `TemplateNotFoundException`

appveyor.yml
```diff
@@ -50,7 +50,7 @@ build_script:
   # install openapi-generator locally
   - mvn clean install --quiet -Dorg.slf4j.simpleLogger.defaultLogLevel=error
   # run the locally installed openapi-generator-gradle-plugin
-  - gradle -b modules\openapi-generator-gradle-plugin\samples\local-spec\build.gradle buildGoSdk --info
+  - gradle -b modules\openapi-generator-gradle-plugin\samples\local-spec\build.gradle buildGoSdk --stacktrace
```
jimschubert, line 53: I'd like to leave `--stacktrace` here rather than `--info`. It was taking ~2 hours to get the feedback from AppVeyor after commit and I was unable to repro the issue in Windows locally.

modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Meta.java
```diff
@@ -136,52 +141,38 @@ public void execute() {
...
+                        new TemplateManagerOptions(false, false),
+                        new MustacheEngineAdapter(),
+                        new TemplatePathLocator[]{ new CommonTemplateContentLocator("codegen") }
+                );
+
+                String template = templateProcessor.readTemplate(new File(TEMPLATE_DIR_CLASSPATH, support.templateFile).getPath());
+
+                String formatted = template;
+
+                Mustache.TemplateLoader loader = name -> templateProcessor.getTemplateReader(name.concat(MUSTACHE_EXTENSION));
```
jimschubert, line 160: Notice regarding the design… template processor no longer requires that we pass that logic through a `DefaultGenerator` instance, which holds logic for OpenAPI input generation to templated output.

We can now generate without tight coupling to OpenAPI documents. This is work directed toward #841 and somewhat related to #843.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPhpCodegen.java
```diff
@@ -218,7 +218,7 @@ public void processOpts() {
         // supportingFiles.add(new SupportingFile("LICENSE", "", "LICENSE"));
 
         // all PHP codegens requires Composer, it means that we need to exclude from SVN at least vendor folder
-        supportingFiles.add(new SupportingFile(".gitignore", "", ".gitignore"));
```
jimschubert, line 221: I found that `.gitignore` files sometimes don't load via `this.getClass().getClassLoader().getResource(path)`.

For example, this works: `this.getClass().getClassLoader().getResource("_common/.openapi-generator-ignore")` while this did not: `this.getClass().getClassLoader().getResource("graphql-nodejs-express-server/.gitignore")`. This only seems to affect PHP and the GraphQL Express servers as everything else is either `gitignore` or `gitignore.mustache` (the formats documented as acceptable on Java documentation).

modules/openapi-generator/src/main/java/org/openapitools/codegen/templating/HandlebarsEngineAdapter.java
```diff
@@ -83,14 +82,22 @@ public TemplateSource sourceAt(String location) {
...
+        for (String file : possibilities) {
             try {
                 return new StringTemplateSource(file, generator.getFullTemplateContents(file));
             } catch (Exception ignored) {
             }
         }
-        throw new TemplateNotFoundException(templateFile);
+
+        // allow lookup of files without extension modification (such as .openapi-generator-ignore, README.md, etc)
+        try {
```
jimschubert, line 95: This is required because filenames without extensions would become `.hbs` or `.handlebars` in the above loop. A file which isn't intended to be handled like `REAMDE.md` would become `README.hbs` or `README.handlebars` and would result in template not found without this as-is attempt.

samples/client/petstore/kotlin-multiplatform/.openapi-generator-ignore
```diff
@@ -1,23 +1,5 @@
...
-# You can recursively match patterns against a directory, file or extension with a double asterisk (**):
-#foo/**/qux
-# This matches foo/bar/qux, foo/baz/qux, and foo/bar/baz/qux
-
-# You can also negate patterns with an exclamation (!).
-# For example, you can ignore all files in a docs folder with the file extension .md:
-#docs/*.md
-# Then explicitly reverse the ignore rule for a single file:
-#!docs/README.md
+gradle/wrapper/gradle-wrapper.jar
```
jimschubert, line 2: I had issues executing this test locally with Bash 5.0. Had to update gradle in the example so the bash sourcing would succeed. I've only updated the generated output rather than the embedded files so we're not forcing users to update to a potentially invalid Gradle version as they regenerate the kotlin multiplatform generate.

jimschubert:
This PR completes my targeted refactor work for the "Template Engine" component defined in #845

image

jimschubert:
I'll merge this tomorrow or Friday if there's no additional feedback.

modules/openapi-generator/src/main/java/org/openapitools/codegen/templating/GeneratorTemplateContentLocator.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public String getFullTemplatePath(String relativeTemplateFile) {
+        CodegenConfig config = this.codegenConfig;
+
+        //check the supplied template library folder for the file
+        final String library = config.getLibrary();
+        if (StringUtils.isNotEmpty(library)) {
+            //look for the file in the library subfolder of the supplied template
+            final String libTemplateFile = buildLibraryFilePath(config.templateDir(), library, relativeTemplateFile);
+            if (new File(libTemplateFile).exists() || this.getClass().getClassLoader().getResource(libTemplateFile) != null) {
```
DenisKnoepfle, line 63: JAR classpath resource path requires '/' as file separator always, must not be OS dependent, e.g. '\\' under Windows won't work!
Suggestion: use `TemplateManager.getCPResourcePath(libTemplateFile)`, may extract to similar method like `embeddedTemplateExists(String)`

modules/openapi-generator/src/main/java/org/openapitools/codegen/templating/GeneratorTemplateContentLocator.java
```diff
@@ -0,0 +1,91 @@
...
+                return libTemplateFile;
+            }
+        }
+
+        //check the supplied template main folder for the file
+        final String template = config.templateDir() + File.separator + relativeTemplateFile;
+        if (new File(template).exists() || this.getClass().getClassLoader().getResource(template) != null) {
```
DenisKnoepfle, line 70: Same issue with '\\' in resource path under windows, see [line 63](https://github.com/OpenAPITools/openapi-generator/pull/6357/commits/636c4dd2d6681d67a11c5c374814034257ef7771#r449851055).

modules/openapi-generator/src/main/java/org/openapitools/codegen/templating/GeneratorTemplateContentLocator.java
jimschubert: @DenisKnoepfle are you suggesting this based on code quality or based on an error you've experienced?

This method falls back to embedded templates in the last two conditions, which allows for Windows paths evaluation using exactly what you've mentioned above.

The methods as written here allow us to check for the embedded file in operating systems without windows style paths without incurring the performance overhead of doing that adjustment.

Maybe this method's comment would be better to document it as this to avoid confusion?

```
     * 1) (template dir)/libraries/(library) [Windows + *nix]
     * 2) (embedded template dir)/libraries/(library) [*nix]
     * 2) (template dir) [Windows + *nix]
     * 3) (embedded template dir) [*nix]
     * 4) (embedded template dir)/libraries/(library) [Windows only]
     * 4) (embedded template dir) [Windows only]
```

And since we use File.separator in `buildLibraryFilePath`, the file check works for both Windows and non-Windows separators consistently throughout.

The alternative that you suggest would require transforming via the regular expression in `TemplateManager.getCPResourcePath` for all operating systems, and we really shouldn't have to pay that price on non-Windows.

DenisKnoepfle: Sorry for the late response, I had been on holidays.
@jimschubert I'm suggesting this based on an error I've experienced under Windows using the template path feature (resolve from classpath).

> And since we use File.separator in buildLibraryFilePath, the file check works for both Windows and non-Windows separators consistently throughout.

This is what's breaking it. You _must not use_ backslashes ('\\') in a java classpath-path, it won't work, not even under Windows. Thus it must always be a forward slash ('/') when forming a classpath-path.
Is it clear now how I meant it? Did you test it under Windows trying to find a resource on a classpath that was not in the root of a jar, e.g. _some.jar/some/deep/folder/JavaSpring/pojo.mustache_? In my case it always resorted to the embedded dir instead of using my custom mustache template.

jimschubert: @DenisKnoepfle thanks for the detailed response. The issue was indeed that I was testing  only with a directory structure which mirrors the `resources` directory in our project. However, I was also testing under Git Bash which I assume is not using the same file separators. I've never done Java development under windows, so I'm not entirely sure how to attach a remote debugger to find out.

If you don't mind, I'd like to tag you in the PR I'll open if you're able to test.

DenisKnoepfle: Sorry, I couldn't test it sooner, but I confirmed that it's working properly now for me under Windows (v5.0.0-beta3). Thank you!

PR #6360, opened 2020-05-19
[typescript-angular] drop support of angular below 6.0.0

macjohnny:
Remove support for old angular versions below 6.0.0, allowing to clean up the code base.

### PR checklist


TiFu:
This raises a philosophical question: Do we want to fix the shortcomings of OAS2 with an implementation that tries to be smart? 

In the case that you described, what happens if a new enum value is introduced only for `Query2SortEnum` (let's say `random`)? 
Does TS still allow what you are trying to do or would this potentially cause errors totally unrelated to the actual change in the spec?

amakhrov:
> Does TS still allow what you are trying to do or would this potentially cause errors totally unrelated to the actual change in the spec?

You will be able to assign `Query1SortEnum` values to variables of type `Query2SortEnum` but not vice versa.

>  Do we want to fix the shortcomings of OAS2

I'm not a big proponent of that approach. I mostly wanted to point out that the change in the PR can have backward-incompatible outcome without a fallback even for angular 8/9 users.

On the other hand, e.g. axios generator has never had such `stringEnum` option - it just always emits enums. We can stick to this approach for consistency across typescript generators.

wing328:
If it's a limitation only with OAS2, then I would suggest asking users to migrate to OAS3 spec instead.

macjohnny:
In order to minimize the breaking changes, I decided to revert the changes related to the string enums, since the functionality is already implemented and not too much of a burden.

PR #6386, opened 2020-05-21
[Python-experimental] Should accept float value serialized without decimal point

sebastien-rosset:
This fixes an issue when a floating point number is serialized without a decimal point. For example, both `12345` and `100.1234` are valid representation of a float value in a serialized JSON document.

```json
{
   "float_value": 12345,
   "another_float_value": 100.1234
}
```

### PR checklist


sebastien-rosset:
@spacether

spacether:
Floating point numbers being serialized without a decimal point means that they are sending integers not floats.
We have three possible solutions here.
1. do type conversion like you have suggested
2. the spec author can define the spec as oneOf: - integer, - float
3. leave it to the end users to edit upconversion_type_pairs to add int to float conversion like you did

Option 1: If a user wants integers to not work as floats, the user will have to call the endpoint with _preload_content. If they do that, then type checking and validation are skipped.
Option 2: This solution meets the needs of a user that:
- wants only integers
- wants a value that can be an integer or a float
Option 3:
This meets the needs of a user that want integers accepted in all float locations and does not force that solution on all users

I'm concerned that by picking option 1, we force all users into that solution and don't allow for situations where we want only an integer for integer values.
What do you think of the above 3 options?

sebastien-rosset:
> Floating point numbers being serialized without a decimal point means that they are sending integers not floats.

I think you are assuming a floating point must be serialized with a decimal point, but that's not the case. It's legal to serialize the floating point number 1.0 as "1.0", "1" and "1.0000". Actually there is an infinite number of representations.  In JSON, there is just a "number" data type. 

OAS uses JSON schema, which uses the types specified in RFC8259.

Here is a an example of golang marshaling a floating point without decimal: https://play.golang.org/p/rGaLUV_fsqj
Several libraries that serialize floating points using the most condensed format possible.

PR #6390, opened 2020-05-22
[Python-experimental] Rename from_server variable to json_variable_naming

sebastien-rosset:
Add unit test validating the creation of an object from a locally available JSON dict

### PR checklist


sebastien-rosset:
@spacether

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_composed.mustache
```diff
@@ -1,7 +1,7 @@
     required_properties = set([
         '_data_store',
         '_check_type',
-        '_from_server',
+        '_json_variable_naming',
```
wing328, line 4: What about just `_variable_naming` as later users may want the support of other formats such as XML?

(only Java, Go client support XML payload at the moment)

spacether: @wing328 I am seeing many different acceptable xml variable naming styles.
What are you thinking that people will pass in, something like this?
variable names:
```
python: first_name
json: firstName
xml: <firstname> or <firstName> or <FIRSTNAME> or <first_name> etc
```

sebastien-rosset: If it's XML, this should be the name as specified in the XML object at https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#xmlObject

sebastien-rosset: Maybe it should be oai_variable_naming?

spacether: _spec_variable_naming?
oai or oas may not mean anything to the users.
We referred to json and yaml files fir endpoints as spec documents.

sebastien-rosset: spec_naming_convention? spec_naming_syntax?

sebastien-rosset: It just occurred to me "variable" is a bit of a misnomer. When you pass the arguments, typically I think it's going to be a dict, and the naming convention flag applies to the keys of the dictionary, not the variable names. I understand they do get converted to variables because of `**`, but still it's not very natural. In the OAS doc, they are "properties"

sebastien-rosset: serialized_naming_format?

spacether: _spec_property_naming?

sebastien-rosset: ok, I will go with that.

wing328:
Just want to add that XML is one example. Protocol buffer is also something the generators can support (Dart client generator already supports it)

sebastien-rosset:
> Just want to add that XML is one example. Protocol buffer is also something the generators can support (Dart client generator already supports it)

Good point, I think the new name is agnostic of the serialization scheme.

PR #6469, opened 2020-05-28
[feature][python] Support aliasing of API keys

jirikuncar:
Adds support for API key aliasing as it's done in Go and Java:

- https://github.com/OpenAPITools/openapi-generator/pull/4940
- https://github.com/OpenAPITools/openapi-generator/pull/4966

### PR checklist


modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -466,7 +466,7 @@ conf = {{{packageName}}}.Configuration(
             auth['{{name}}'] = {
                 'type': 'api_key',
                 'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
-                'key': '{{keyParamName}}',
+                'key': '{{#vendorExtensions.x-auth-id-alias}}{{.}}{{/vendorExtensions.x-auth-id-alias}}{{^vendorExtensions.x-auth-id-alias}}{{name}}{{/vendorExtensions.x-auth-id-alias}}',
```
spacether, line 469: Why aren't we using {{keyParamName}} if there is no vendor extension set?

spacether:
Can you please add a test that sets configuration values using a shared vendor extension key?
When we add features without any verification tests we have no way of verifying that it is working now, and we have no automatic way of telling if and when it breaks in the future.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: > using `keyParamName` is wrong since there can be a collision when using API key in header and query argument under same name: `curl -H 'my-key: header-key' https://example.com/endpoint` vs. `curl https://example.com/endpoint?my-key=query-key`

https://github.com/OpenAPITools/openapi-generator/pull/4940#discussion_r381172053

jirikuncar:
@spacether can you point me to an example test?

spacether:
> @spacether can you point me to an example test?

Sure, [this file](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py#L36) instantiates the Configuration class

If you want to call api endpoints then you can add those test to a file like:
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/test/test_pet_api.py
Looking in the docs [here](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/docs/PetApi.md#get_pet_by_id) I see that get_pet_by_id uses api_key auth so maybe you can add a test of that endpoint

jirikuncar:
@spacether I have added tests. It should work together with prefixes too. I had to modify the lookup method.

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -572,6 +572,9 @@ paths:
       summary: To test "client" model
       description: To test "client" model
       operationId: testClientModel
+      security:
+        - api_key_query: []
+          api_key: []
```
spacether, line 577: Is this missing a `-`?

samples/openapi3/client/petstore/python-experimental/petstore_api/configuration.py
```diff
@@ -396,15 +396,15 @@ def logger_format(self, value):
         self.__logger_format = value
         self.logger_formatter = logging.Formatter(self.__logger_format)
 
-    def get_api_key_with_prefix(self, identifier):
+    def get_api_key_with_prefix(self, identifier, alias=None):
```
spacether, line 399: How about changing alias to `identifier_alias` to help convey that this is related to the identifier?

For the key assignment how about something like this easier to understand snippet?
```
key = self.api_key.get(identifier)
if key is None and identifier_alias is not None:
  key = self.api_key.get(alias)
```

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
```diff
@@ -43,6 +44,31 @@ def test_configuration(self):
...
+            assert 'PREFIX SECRET_VALUE' == kwargs['headers']['api_key']
+            raise RuntimeError("passed")
+
+        with petstore_api.ApiClient(configuration) as client:
+            api = FakeApi(client)
+            client.request = request
+            try:
+                api.test_client_model(client_client={})
+            except RuntimeError as e:
+                assert "passed" == str(e)
```
spacether, line 69: Thank you for adding this example, I understand how this will be used now

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
jirikuncar: No, I'm checking usage of `Multiple API Keys`.

samples/openapi3/client/petstore/python-experimental/petstore_api/configuration.py
jirikuncar: I would really love to remove this internal function from public API.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -423,15 +423,15 @@ conf = {{{packageName}}}.Configuration(
         self.__logger_format = value
         self.logger_formatter = logging.Formatter(self.__logger_format)
 
-    def get_api_key_with_prefix(self, identifier):
+    def get_api_key_with_prefix(self, identifier, alias=None):
         """Gets API key (with prefix if set).
 
         :param identifier: The identifier of apiKey.
```
jirikuncar, line 429:
```suggestion
        :param identifier: The identifier of apiKey.
        :param alias: The alias of identifier to fetch value from.
```

samples/openapi3/client/petstore/python-experimental/petstore_api/configuration.py
jirikuncar: I will add the docstring, but `identifier_alias` seems unnecessary too long.

jirikuncar:
@spacether I have updated the spec to show different values for api key ID and name. I hope the test is clear now.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -462,12 +463,15 @@ conf = {{{packageName}}}.Configuration(
...
-        if '{{keyParamName}}' in self.api_key:
+        if '{{name}}' in self.api_key{{#vendorExtensions.x-auth-id-alias}} or '{{.}}' in self.api_key{{/vendorExtensions.x-auth-id-alias}}:
             auth['{{name}}'] = {
                 'type': 'api_key',
                 'in': {{#isKeyInCookie}}'cookie'{{/isKeyInCookie}}{{#isKeyInHeader}}'header'{{/isKeyInHeader}}{{#isKeyInQuery}}'query'{{/isKeyInQuery}},
                 'key': '{{keyParamName}}',
-                'value': self.get_api_key_with_prefix('{{keyParamName}}')
+                'value': self.get_api_key_with_prefix(
+                    '{{name}}',{{#vendorExtensions.x-auth-id-alias}}
+                    alias='{{.}}',{{/vendorExtensions.x-auth-id-alias}}
```
jirikuncar, line 473: @spacether please note that the `alias`argument is only generated for `x-auth-id-alias` now.

modules/openapi-generator/src/main/resources/python/python_doc_auth_partial.mustache
```diff
@@ -95,11 +95,11 @@ configuration = {{{packageName}}}.Configuration(
 configuration = {{{packageName}}}.Configuration(
     host = "{{{basePath}}}",
     api_key = {
-        '{{{keyParamName}}}': 'YOUR_API_KEY'
+        '{{#vendorExtensions.x-auth-id-alias}}{{{.}}}{{/vendorExtensions.x-auth-id-alias}}{{^vendorExtensions.x-auth-id-alias}}{{{name}}}{{/vendorExtensions.x-auth-id-alias}}': 'YOUR_API_KEY'
     }
 )
 # Uncomment below to setup prefix (e.g. Bearer) for API key, if needed
-# configuration.api_key_prefix['{{{keyParamName}}}'] = 'Bearer'
+# configuration.api_key_prefix['{{#vendorExtensions.x-auth-id-alias}}{{{.}}}{{/vendorExtensions.x-auth-id-alias}}{{^vendorExtensions.x-auth-id-alias}}{{{name}}}{{/vendorExtensions.x-auth-id-alias}}'] = 'Bearer'
```
jirikuncar, line 102:
```suggestion
# configuration.api_key_prefix['{{name}}'] = 'Bearer'
```

modules/openapi-generator/src/main/resources/python/python_doc_auth_partial.mustache
```diff
@@ -95,11 +95,11 @@ configuration = {{{packageName}}}.Configuration(
 configuration = {{{packageName}}}.Configuration(
     host = "{{{basePath}}}",
     api_key = {
-        '{{{keyParamName}}}': 'YOUR_API_KEY'
+        '{{#vendorExtensions.x-auth-id-alias}}{{{.}}}{{/vendorExtensions.x-auth-id-alias}}{{^vendorExtensions.x-auth-id-alias}}{{{name}}}{{/vendorExtensions.x-auth-id-alias}}': 'YOUR_API_KEY'
```
jirikuncar, line 98:
```suggestion
        '{{name}}': 'YOUR_API_KEY'
```

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -1159,11 +1162,13 @@ components:
             'read:pets': read your pets
     api_key:
       type: apiKey
-      name: api_key
+      name: X-Api-Key
```
jimschubert, line 1165: It may not be obvious but some generators samples have "live" tests which run pre-written tests against compiled code (see jersey2-java8). The tests aren't overwritten on generation (as a rule within DefaultGenerator).

This is documented in the description at the top of the file:

> This spec is mainly for testing Petstore server and contains fake endpoints, models. Please do not use this for any other purpose.

If you update this api key, it would require changing those other "live tests".

In this case, I'd recommend creating a separate smaller spec or adding another auth type outside of api_key which allows you to test the aliasing. That way you wouldn't have the conflict in the test.

spacether: Jim does this point apply to this cordoned off spec file for python-experimental?
Is there another tag where we could ad this security scheme on to an use it there?
Should we use a new api tag?

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Why not get rid of the alias parameter and pass in the alias value as the identifier value? When we have the alias value don't we never use the identified value?

jirikuncar: I am trying to keep the code backward compatible. In general, it's easier to keep backward compatibility with keyword arguments especially when they have default value.

jimschubert: I agree that the approach in the PR would be my preference as well.

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
jimschubert: @spacether I don't know why a spec for python-only would be used by Java. My comment above is about changing a spec which is shared elsewhere and would require changes to hand-written Java tests.

I don't really follow your other questions. My recommendation was to add to the spec to avoid breaking those Java tests, or to create a standalone spec which demonstrates the aliasing behavior. If it was me, I'd do a separate minimal spec which demonstrates the behavior in unit tests then add the new alias parts to any existing spec to verify samples generation.

An issue with modifying existing specs that target swagger petstore (as does the one modified here) is that swagger petstore defines the header as `api_key` and someone wouldn't be able to use this spec to generate a valid client against that live API any longer. If we instead have another spec which doesn't target a real system's design (search for `pony`, `fruits`, and minimal "ping" specs), we don't risk breaking existing examples.

Eventually, I think we need to better organize and document the test specs so we know each are "fakes", which are intended to be exemplar full APIs, and which ones are only meant to evaluate full feature support.

jirikuncar: I will create a separate spec to test API key aliasing feature and generate new Python client with tests for it.

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
```diff
@@ -21,6 +21,7 @@
 from collections import namedtuple
 
 import petstore_api
+from petstore_api.api import FakeApi
```
jirikuncar, line 24:
```suggestion
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
spacether: Understood, thanks

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
spacether: So this spec is used in 2 locations to generate samples:
- bin/openapi3/java-petstore-jersey2-java8.sh
- bin/openapi3/python-experimental-petstore.sh
So I agree with Jim's assessment that we need a new spec for this case because I do not see a PR adding support for api key aliasing in java-jersey2.
If this spec were only used by python-experimental then we could just:
- add new security schemes
- use them in a non-like api like the `fake` tag api

bin/openapi3/python-experimental-petstore.sh
```diff
@@ -32,3 +32,8 @@ yaml="modules/openapi-generator/src/test/resources/3_0/python-experimental/petst
 ags="generate -t modules/openapi-generator/src/main/resources/python -i $yaml -g python-experimental -o samples/openapi3/client/petstore/python-experimental/ --additional-properties packageName=petstore_api $@"
 
 java $JAVA_OPTS -jar $executable $ags
+
+yaml="modules/openapi-generator/src/test/resources/3_0/extensions/x-auth-id-alias.yaml"
+ags="generate -t modules/openapi-generator/src/main/resources/python -i $yaml -g python-experimental -o samples/openapi3/client/extensions/x-auth-id-alias/python-experimental/ --additional-properties packageName=x_auth_id_alias $@"
```
spacether, line 37: If we decide to keep this separate sample can you put in a python-experimental folder so our uses see both python-experimental samples at the same place?
Something like
- `samples/openapi3/client/petstore/python-experimental/general_sample`
- `samples/openapi3/client/petstore/python-experimental/auith_id_alias_sample`

Also, how about [this solution](https://github.com/OpenAPITools/openapi-generator/pull/6469#discussion_r435337114) of separating our existing spec into a python-experimental only version and a java version. That will let us add this feature in a python-experimental only spec, have no impact on other generators, not impact live server testing, and will keep it clear to users what sample to look at.

jirikuncar: I would like to generate other clients for this feature. I don't think it's `python-experimental` specific.

jirikuncar, line 37: I could change the folder name where the client is generated:
```suggestion
ags="generate -t modules/openapi-generator/src/main/resources/python -i $yaml -g python-experimental -o samples/openapi3/client/python-experimental/extensions/x-auth-id-alias/ --additional-properties packageName=x_auth_id_alias $@"
```

jirikuncar:
@spacether @jimschubert the whole idea behind having a separate spec for each feature is to have a support matrix (with test status):

```
language \ extensions | x-auth-id-alias | ...
----------------------------------------------
python-experimental   | x               | ...
java/jersey2          | ?               | ...
go-experimental       | ?               | ...

x - tested
? - supported (not tested)
```

jimschubert:
@jirikuncar I completely agree. It seems like there might be some misunderstanding in this PR so I'll reframe.

We have two needs for specs:

* testing
* samples

What you've done with the custom spec to be testable is correct. We also follow the same practice elsewhere to generate feature specific or option specific samples.

It sounds like what Justin is aiming for is the "golden" example, which we really don't currently define or practice. By "golden", I mean a spec that represents all features and can be used across the mature generators. I feel like we're a little way from being able to do that consistently. My project (number 5) aims to get us there.

I don't see any issue with the approaches in this PR.

bin/openapi3/python-experimental-petstore.sh
spacether: Ah, I understand. That's fine then

spacether: No need, where you have it makes more sense given that this is a feature prove out and will apply across generators like you said

spacether:
@jirikuncar can you run `bin/utils/ensure-up-to-date`?
The PR looks good, it is only blocked by these CI errors:
```

	modified:   samples/client/petstore/python-asyncio/petstore_api/configuration.py
	modified:   samples/client/petstore/python-tornado/petstore_api/configuration.py
	modified:   samples/client/petstore/python/petstore_api/configuration.py
	modified:   samples/openapi3/client/extensions/x-auth-id-alias/python-experimental/.openapi-generator/FILES
	modified:   samples/openapi3/client/petstore/python-experimental/.openapi-generator/FILES
	modified:   samples/openapi3/client/petstore/python-experimental/README.md
	modified:   samples/openapi3/client/petstore/python-experimental/petstore_api/models/__init__.py
	modified:   samples/openapi3/client/petstore/python/petstore_api/configuration.py
```
Running ensure-up-to-date and committing the changes will allow CI to pass.

jirikuncar:
@jimschubert @spacether I have merged `master` to my feature branch and re-run the ./bin/*/python*.sh` scripts.

jirikuncar:
@spacether can you please re-review?

PR #6477, opened 2020-05-28
[REQ][typescript-axios] useSingleRequestParameter should mark parameter optional if all properties are optional

### LINKED ISSUE #6440: [REQ] typescript-axios useSingleRequestParameter should mark parameter optional if all properties are optional
Follow-up to #6288.

If the `useSingleRequestParameter` flag is passed to typescript axios, then an object is a required parameter of all request functions, even if all object properties are optional, so I end up with code like this:

`myApiCall({})`.

The generator should be able to check if any of the params are required. If not, we can specify the whole object parameter optional with `?`.

codeserk:
This PR uses a default value to `requestParameters` (`= {}`) if all the parameters are optional.

fixes #6440

### PR checklist

@benwiley4000 I think this is not exactly the solution you had in mind, but it should work alike:
If all the parameters are optional, you can use it without parameters (`myApiCall()`). Using default parameter instead of making it optional makes it a bit less complex.


benwiley4000:
Indeed it seems to work the same, since TypeScript automatically marks a param with a default value as optional:

Screen Shot 2020-05-28 at 7 09 35 PM

Issue #6178, opened 2020-05-05
[Rust] Issues with client code generation.


seunlanlege:
##### Description
The generated client code for rust has a few problems:

 -  It uses the deprecated futures v0.1, when it should use the newer futures v0.3
 - The return types of the api methods aren't strongly typed:
```rust
Box<dyn Future<Item = ::std::collections::HashMap<String, serde_json::Value>, Error = Error<serde_json::Value>>>
```

The models are properly generated but the return type of the methods don't reference the generated models at all.

 - Some models aren't generated at all when defined as such in the schema:

![Screenshot from 2020-05-05 13-27-18](https://user-images.githubusercontent.com/16494314/81065867-2c8efa00-8ed4-11ea-83aa-0eb55e132fb2.png)

##### openapi-generator version

```
openapi-generator-cli 5.0.0-SNAPSHOT
  commit : 46216cd
  built  : 2020-04-28T13:03:54Z
  source : https://github.com/openapitools/openapi-generator
  docs   : https://openapi-generator.tech/

```
##### OpenAPI declaration file content or url

 https://api.slack.com/specs/openapi/v2/slack_web.json

##### Command line used for generation

`java -jar ~/Downloads/openapi.jar generate -g rust -i https://api.slack.com/specs/openapi/v2/slack_web.json -o rust`

##### Steps to reproduce

run the command above

##### Suggest a fix/enhancement

- [ ] The generated client code should use the `reqwest` crate with async/await.
- [ ] Change the api for methods from:

```rust
pub trait UsersApi {
    fn users_conversations(&self, params: UsersConversationsParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error>;
    fn users_delete_photo(&self, params: UsersDeletePhotoParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error>;
    fn users_get_presence(&self, params: UsersGetPresenceParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error>;
    fn users_identity(&self, params: UsersIdentityParams) -> Result<serde_json::Value, Error>;
    fn users_info(&self, params: UsersInfoParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error>;
    fn users_list(&self, params: UsersListParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error>;
[...]

PR #6481, opened 2020-05-29
[Rust][Client][Reqwest] Better http error handling

bcourtine:
When the server response is not OK (code >= 300), return error is not a generic "Reqwest::Error" but a custom "ResponseError", containing the status, the raw body (String), and a deserialization as "serde_json::Value" when it is possible.

Related issue: https://github.com/OpenAPITools/openapi-generator/issues/6178

### PR checklist

cc @frol (2017/07) @farcaller (2017/08) @bjgill (2017/12) @richardwhiuk (2019/07) @paladinzh (2020/05) @wing328


samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
```diff
@@ -47,7 +47,16 @@ impl DefaultApi for DefaultApiClient {
...
+        let mut resp = client.execute(req)?;
+        if resp.status().is_success() {
+            Ok(resp.json()?)
+        } else {
+            let status = resp.status();
+            let content = resp.text()?;
+            let entity: Option<serde_json::Value> = serde_json::from_str(&content).ok();
+            let error = crate::apis::ResponseErrorContent { status, content, entity };
+            Err(Error::ResponseError(error))
+        }
```
seunlanlege, line 59: hey :raising_hand_man:  great attempt on this issue, but this isn't complete yet.

The openapi spec allows for defining a schema for every possible response status: eg

```json
"responses": {
    "200": {},
    "202": {},
    "400": {},
    "default": {}
}
```

so instead of just checking for a success status, you'll need to match over the response status.

and deserialize based on the defined schema for that status as well as the fallback(default) schema.

I imagine this means that the return type of methods would become an enum, 

```rust
pub enum OperationIdResponse {
    Status200(Status200Schema),
    Status202(Status202Schema),
    Status400(Status400Schema),
    Unspecified(DefaultSchema)
}
```

bcourtine: Hi @seunlanlege,

Thanks for your comment. I know this is just a first attempt which is incomplete.

I had the same idea of an enum as response, but I don't like it. I think we should not return `Ok` when the return status is not ok…

bcourtine: I am working on something like:

```rust
#[derive(Debug, Clone)]
pub struct ResponseErrorContent<T> {
    pub status: reqwest::StatusCode,
    pub content: String,
    pub entity: Option<T>,
}

#[derive(Debug)]
pub enum Error<T> {
    Reqwest(reqwest::Error),
    Serde(serde_json::Error),
    Io(std::io::Error),
    ResponseError(ResponseErrorContent<T>),
}
```

where T could be an enum of error types corresponding to error codes.

seunlanlege: also bear in mind that there could be more than one success response in the `2xx` range, So success types would potentially be an enum

bcourtine:
@seunlanlege Still not complete, but the last version should be much closer from what you want.

samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
bcourtine: With an enum, multiple success codes can be handled, but this case is not frequent. So I don't know if it is worth complicating the generated code for this.

bcourtine:
Feature seems almost complete, with some limitations:

- deserialization will match the error structure. So if 4XX and 5XX errors have the same structure, a 500 response may result in a `Status400` enum value.
- for some APIs (Slack), default response is the error case. For others (petstore), it is the ok response. I let the default response in the error enum, but it depends…

samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
seunlanlege: > multiple success codes can be handled, but this case is not frequent. 
unfortunately it's actually very common :sweat_smile: 

jira/sendgrid's openapi specifications have multiple 2xx success schemas

seunlanlege:
:raising_hand: @bcourtine need a hand with this?

bcourtine:
> raising_hand @bcourtine need a hand with this?

Hi @seunlanlege,

Thanks for the proposition.

I think current code is the best compromise between features and complexity of generated code. Here are my thoughts about a track I did not implement (matching the return code to force the error enum value):

```rust
let error_entity = match response.status().as_u16() {
    400 => OperationErrors::Status400(response.json().await?)
    401 => OperationErrors::Status401(response.json().await?)
    402 => OperationErrors::Status402(response.json().await?)
    // etc.
}
```

I think this code is more complicated and useless, since `serde` can choose automatically a value from the response body structure. If exact return code is needed by client code, it is not lost since the `ResponseErrorContent` struct contains it.

As a user, when I generate such code, I correct generated code to group responses that have the same structure (with for example `Status4XX` instead of `Status400` + `Status401` + …). Such modification is much easier if there is not an explicit code match in the operation.

I may miss some quick improvements. If you find simple ways to improve the code, feel free to propose suggestions/evolutions.

If not, I didn't have plans to work more on this issue, and am waiting for the merge approval (/cc @wing328 ).

seunlanlege:
hey @bcourtine so I tested this with the slack api and this PR doesn't actually address the original issue. Which was, the return types of the methods should be strongly-typed.

Here's a method signature from one of slack's apis.

```rust
/// method signature shouldn't return a hashmap, it should be strongly typed with the provided schema in the spec
 pub async fn delete_photo(configuration: &configuration::Configuration, params: DeletePhotoParams) -> Result<::std::collections::HashMap<String, serde_json::Value>, Error<DeletePhotoErrors>>;

/// struct for typed errors of method `delete_photo`
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(untagged)]
pub enum DeletePhotoErrors {
    /// this also shouldn't be hashmap, this should be strongly typed as well, there's a schema in the spec for it
    DefaultResponse(::std::collections::HashMap<String, serde_json::Value>),
    /// this is unnecessary
    UnknownList(Vec<serde_json::Value>),
    /// as well as this.
    UnknownValue(serde_json::Value),
}
```

Also on the topic of enum success types, this is taken from [jira's openapi specification](https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json?_ga=2.249667201.627335544.1591098091-168643997.1585732852), as you can see there's a definition for `200` and `201` responses.

Screenshot from 2020-06-02 12-43-17

if you are unable to continue work on this that's fine please let @wing328 know so he can assiggn it to someone else.

wing328:
About the HashMap response type (in the default response), I've performed some tests with other generators as well and at this stage, I think it's a bug with the default codegen so I'll take a look and hopefully file a fix by the end of this week.

wing328:
The issue with slack_web.json has to do with Swagger Parser incorrectly parsing the inline schema when additioanlProperties is set to false.

I've removed it here: https://gist.github.com/wing328/61225b4ad2d838447340a70494e09333

Can you please this modified spec a try to see if the output looks good to you?

I did a test and no longer see the Map/Dict response and the inline model is created as a model correctly.

seunlanlege:
Hey @wing328 I've tested this and i noticed that the `DefaultResponse` variant of the error contains the success return data.

```rust
/// struct for typed errors of method `close`
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(untagged)]
pub enum CloseErrors {
    DefaultResponse(crate::models::ImCloseSchema),
    UnknownList(Vec<serde_json::Value>),
    UnknownValue(serde_json::Value),
}
// for the method:
pub async fn close(configuration: &configuration::Configuration, params: CloseParams) -> Result<crate::models::ImCloseSchema, Error<CloseErrors>>;
```

bcourtine:
Hi @seunlanlege,

Some APIs use default response to describe API "success", others use it as a "global error catch".

Since both usages exist, it is not possible to guess if default response must be added in the error enum or not. So it is always added. If you don't need it (because default describes a success instead of an error), you can to remove it.

seunlanlege:
> Since both usages exist, it is not possible to guess if default response must be added in the error enum or not. So it is always added. If you don't need it (because default describes a success instead of an error), you can to remove it.

what if we added a CLI argument that tells the generator how to handle the default case? as an error or success.

bcourtine:
Possible, but I don't know if CLI should be used for such micro-tunning of code generation.
What do you think about it, @wing328 ?

wing328:
I've filed https://github.com/OpenAPITools/openapi-generator/pull/6625 to fix the default response.

When testing with this PR, I got the following changes (partial):
```diff
@@ -248,7 +248,7 @@ pub enum ChannelsMarkError {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ChannelsRenameError {
-    DefaultResponse(crate::models::ChannelsRenameSchema),
+    DefaultResponse(crate::models::ChannelsRenameErrorSchema),
     UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
@@ -257,7 +257,7 @@ pub enum ChannelsRenameError {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ChannelsRepliesError {
-    DefaultResponse(crate::models::ChannelsRepliesSchema),
+    DefaultResponse(crate::models::ChannelsRepliesErrorSchema),
     UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
@@ -266,7 +266,7 @@ pub enum ChannelsRepliesError {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ChannelsSetPurposeError {
-    DefaultResponse(crate::models::ChannelsSetPurposeSchema),
+    DefaultResponse(crate::models::ChannelsSetPurposeErrorSchema),
     UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
```

I think this confirms the default response is now correctly using the error models defined

wing328:
@bcourtine #6625 has been meged into master. Can you please merge the latest master into this branch and regenerate the samples? Thank you.

seunlanlege:
just works™

wing328:
@seunlanlege thanks for the confirmation. I'll merge once all the tests pass.

wing328:
Travis CI tests passed: https://travis-ci.org/github/OpenAPITools/openapi-generator/builds/697621225?utm_source=github_status&utm_medium=notification but status not sent back to Github.

PR #6489, opened 2020-05-29
[Typescript] Support InversifyJS

bodograumann:
To be able to use the new typescript generator as a replacement for the existing typescript-inversify generator, we need to add a few things. This PR does just that.

This is not a 1:1 replacement. Instead I have tried to incorporate my experience with using typescript-inversify and not pollute the templates too much.

The option for enabling this functionality is called `useInversify`.
I had to change the way the SecurityAuthentication objects are generated in order to allow injecting parameters into them.
Normally we should depend on abstractions (interfaces), not implementations, so I added some interfaces. For `<service>RequestFactory` and `<service>ResponseProcessor` this seemed like overkill though.
Abstract classes are used as service identifiers. This has the advantage that during binding the types can be checked automatically and they are compiled into just an empty object.

It would be nice to have tests here, though I'm not quite sure where to put them. Just create a new sample with useInversify enabled and copy the default test setup? @TiFu 

My previous PR #6425, is still included here, so if that is merged I can rebase.

### PR checklist

CC @TiFu
CC TS Committee @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)

### Possible next steps
- Bind to provider and dynamic import
- More convenience method on the ApiServiceBinder


macjohnny:
cc @gualtierim

gualtierim:
@bodograumann the idea of make this feature optionally it's great to not create a breaking change. In my experience the configuration never required any parameter, but it is a possible scenario. I approve these changes.

TiFu:
> It would be nice to have tests here, though I'm not quite sure where to put them. Just create a new sample with useInversify enabled and copy the default test setup? @TiFu

Yeah just create a new sample

* Modify `bin/typescript.sh`, generate sample into a new subfolder
* Copy files from `samples/oai3/client/petstore/typescript/builds/default/*` that are not generated by `bin/typescript.sh to the new folder`
* Create the test cases in `samples/oai3/client/petstore/typescript/tests/[subfolder] by copying whatever you need from `tests/default`, change the project name in the `pom.xml` in the test directory
* Add the test directory to the root `pom.xml`

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -333,7 +333,8 @@ public void processOpts() {
                 .put("lowercase", new LowercaseLambda().generator(this))
                 .put("uppercase", new UppercaseLambda())
                 .put("titlecase", new TitlecaseLambda())
-                .put("camelcase", new CamelCaseLambda().generator(this))
```
TiFu, line 336: Is this a change from master?

modules/openapi-generator/src/main/resources/typescript/http/servers.mustache
```diff
@@ -55,3 +51,7 @@ export class ServerConfiguration<T> {
 {{#servers}}
 export const server{{-index}} = new ServerConfiguration<{ {{#variables}} "{{name}}": {{#enumValues}}"{{.}}"{{^-last}} | {{/-last}}{{/enumValues}}{{^enumValues}}string{{/enumValues}}{{^-last}},{{/-last}} {{/variables}} }>("{{url}}", { {{#variables}} "{{name}}": "{{defaultValue}}" {{^-last}},{{/-last}}{{/variables}} })
 {{/servers}}
+
+{{#useInversify}}
+export const servers = [{{#servers}}server{{-index}}{{^-last}}, {{/-last}}{{/servers}}];
```
TiFu, line 56: Is this something we can just provide in any generator? Doesn't seem to be inversify specific.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
bodograumann: No, I added this to allow the default `camelize`behaviour, i.e. uppercasing the first letter, to be available as lambda in the templates.

modules/openapi-generator/src/main/resources/typescript/http/servers.mustache
bodograumann: Sure, if you think it is useful, I will add it.

TiFu:
@bodograumann I just integrated the feedback from macjohnny into `ts-refactor-master` (cmp. #6341 ). Could you please update this PR? 

FYI: I added a new type called `ObjectParamAPI` that accepts objects as function parameters and uses either `PromiseAPI` or `ObservableAPI` internally depending on the cli args. This probably affects this PR.

bodograumann:
Thank you for the notification @Tifu. When I get back to this task I will take those changes into account.

modules/openapi-generator/src/main/resources/typescript/http/servers.mustache
bodograumann: Always including it now.

bodograumann:
I have rebased and add the small change. Still need to add suitable tests and add inversify support to the new object_params api.

bodograumann:
I think this should be ready.
Any idea what is going wrong with the tests, @TiFu?

TiFu:
Travis complains about `Project 'org.openapitools:TypeScriptDefaultPetstoreClientTests:1.0-SNAPSHOT' is duplicated in the reactor` 

Solution: Rename TypeScriptDefaultPetstoreClientTests in pom.xml in one of the test directories under samples.

CircleCI complains about some changes to `docs/contributing.md` - probably auto-generated from one of the files in `bin/utils/*`

bodograumann:
Yeah, no more failed checks :tada: 
Feel free to merge, @TiFu :wink:

TiFu:
Yay :confetti_ball: Great work!

Thanks for the PR which was merged into `ts-refactor-master`

PR #6493, opened 2020-05-29
[BUG][python] Support named arrays

jirikuncar:
**The model for named array is not generated.**

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
+++ b/modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
@@ -1097,6 +1097,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HealthCheckResult'
+  /fake/array-of-enums:
+    get:
+      tags:
+        - fake
+      summary: Array of Enums
+      operationId: getArrayOfEnums
+      responses:
+        200:
+          description: Got named array of enums
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ArrayOfEnums'
 servers:
   - url: 'http://{server}.swagger.io:{port}/v2'
     description: petstore server
@@ -1443,7 +1456,7 @@ components:
           maximum: 543.2
           minimum: 32.1
           type: number
-          multipleOf: 32.5 
+          multipleOf: 32.5
         float:
           type: number
           format: float
@@ -1703,6 +1716,10 @@ components:
         - placed
         - approved
         - delivered
+    ArrayOfEnums:
+      type: array
+      items:
+        $ref: '#/components/schemas/OuterEnum'
     OuterEnumInteger:
       type: integer
       enum:
--- a/samples/openapi3/client/petstore/python-experimental/petstore_api/api/fake_api.py
+++ b/samples/openapi3/client/petstore/python-experimental/petstore_api/api/fake_api.py
@@ -36,6 +36,7 @@ from petstore_api.model_utils import (  # noqa: F401
 )
 from petstore_api.models import health_check_result
 from petstore_api.models import outer_composite
+from petstore_api.models import array_of_enums
 from petstore_api.models import file_schema_test_class
 from petstore_api.models import user
 from petstore_api.models import client
@@ -598,6 +599,109 @@ class FakeApi(object):
             },
             api_client=api_client,
             callable=__fake_outer_string_serialize
+        )
+
+        def __get_array_of_enums(
+            self,
+            **kwargs
+        ):
+            """Array of Enums  # noqa: E501
+
+            This method makes a synchronous HTTP request by default. To make an
+            asynchronous HTTP request, please pass async_req=True
+            >>> thread = api.get_array_of_enums(async_req=True)
+            >>> result = thread.get()
+
+
+            Keyword Args:
+                _return_http_data_only (bool): response data without head status
+                    code and headers. Default is True.
+                _preload_content (bool): if False, the urllib3.HTTPResponse object
+                    will be returned without reading/decoding response data.
+                    Default is True.
+                _request_timeout (float/tuple): timeout setting for this request. If one
+                    number provided, it will be total request timeout. It can also
+                    be a pair (tuple) of (connection, read) timeouts.
+                    Default is None.
+                _check_input_type (bool): specifies if type checking
+                    should be done one the data sent to the server.
+                    Default is True.
+                _check_return_type (bool): specifies if type checking
+                    should be done one the data received from the server.
+                    Default is True.
+                _host_index (int): specifies the index of the server
+                    that we want to use.
+                    Default is 0.
+                async_req (bool): execute request asynchronously
+
+            Returns:
+                array_of_enums.ArrayOfEnums
+                    If the method is called asynchronously, returns the request
+                    thread.
+            """
+            kwargs['async_req'] = kwargs.get(
+                'async_req', False
+            )
+            kwargs['_return_http_data_only'] = kwargs.get(
+                '_return_http_data_only', True
+            )
+            kwargs['_preload_content'] = kwargs.get(
+                '_preload_content', True
+            )
+            kwargs['_request_timeout'] = kwargs.get(
+                '_request_timeout', None
+            )
+            kwargs['_check_input_type'] = kwargs.get(
+                '_check_input_type', True
+            )
+            kwargs['_check_return_type'] = kwargs.get(
+                '_check_return_type', True
+            )
+            kwargs['_host_index'] = kwargs.get('_host_index', 0)
+            return self.call_with_http_info(**kwargs)
+
+        self.get_array_of_enums = Endpoint(
+            settings={
+                'response_type': (array_of_enums.ArrayOfEnums,),
+                'auth': [],
+                'endpoint_path': '/fake/array-of-enums',
+                'operation_id': 'get_array_of_enums',
+                'http_method': 'GET',
+                'servers': [],
+            },
+            params_map={
+                'all': [
+                ],
+                'required': [],
+                'nullable': [
+                ],
+                'enum': [
+                ],
+                'validation': [
+                ]
+            },
+            root_map={
+                'validations': {
+                },
+                'allowed_values': {
+                },
+                'openapi_types': {
+                },
+                'attribute_map': {
+                },
+                'location_map': {
+                },
+                'collection_format_map': {
+                }
+            },
+            headers_map={
+                'accept': [
+                    'application/json'
+                ],
+                'content_type': [],
+            },
+            api_client=api_client,
+            callable=__get_array_of_enums
         )
 
         def __test_body_with_file_schema(
--- a/samples/openapi3/client/petstore/python-experimental/README.md
+++ b/samples/openapi3/client/petstore/python-experimental/README.md
@@ -87,6 +87,7 @@ Class | Method | HTTP request | Description
 *FakeApi* | [**fake_outer_composite_serialize**](docs/FakeApi.md#fake_outer_composite_serialize) | **POST** /fake/outer/composite | 
 *FakeApi* | [**fake_outer_number_serialize**](docs/FakeApi.md#fake_outer_number_serialize) | **POST** /fake/outer/number | 
 *FakeApi* | [**fake_outer_string_serialize**](docs/FakeApi.md#fake_outer_string_serialize) | **POST** /fake/outer/string | 
+*FakeApi* | [**get_array_of_enums**](docs/FakeApi.md#get_array_of_enums) | **GET** /fake/array-of-enums | Array of Enums
 *FakeApi* | [**test_body_with_file_schema**](docs/FakeApi.md#test_body_with_file_schema) | **PUT** /fake/body-with-file-schema | 
 *FakeApi* | [**test_body_with_query_params**](docs/FakeApi.md#test_body_with_query_params) | **PUT** /fake/body-with-query-params | 
 *FakeApi* | [**test_client_model**](docs/FakeApi.md#test_client_model) | **PATCH** /fake | To test \&quot;client\&quot; model
--- a/samples/openapi3/client/petstore/python-experimental/docs/FakeApi.md
+++ b/samples/openapi3/client/petstore/python-experimental/docs/FakeApi.md
@@ -9,6 +9,7 @@ Method | HTTP request | Description
 [**fake_outer_composite_serialize**](FakeApi.md#fake_outer_composite_serialize) | **POST** /fake/outer/composite | 
 [**fake_outer_number_serialize**](FakeApi.md#fake_outer_number_serialize) | **POST** /fake/outer/number | 
 [**fake_outer_string_serialize**](FakeApi.md#fake_outer_string_serialize) | **POST** /fake/outer/string | 
+[**get_array_of_enums**](FakeApi.md#get_array_of_enums) | **GET** /fake/array-of-enums | Array of Enums
 [**test_body_with_file_schema**](FakeApi.md#test_body_with_file_schema) | **PUT** /fake/body-with-file-schema | 
 [**test_body_with_query_params**](FakeApi.md#test_body_with_query_params) | **PUT** /fake/body-with-query-params | 
 [**test_client_model**](FakeApi.md#test_client_model) | **PATCH** /fake | To test \&quot;client\&quot; model
@@ -321,6 +322,62 @@ No authorization required
 | Status code | Description | Response headers |
 |-------------|-------------|------------------|
 **200** | Output string |  -  |
+
+[[Back to top]](#) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to Model list]](../README.md#documentation-for-models) [[Back to README]](../README.md)
+
+# **get_array_of_enums**
+> array_of_enums.ArrayOfEnums get_array_of_enums()
+
+Array of Enums
+
+### Example
+
+```python
+from __future__ import print_function
+import time
+import petstore_api
+from pprint import pprint
+# Defining the host is optional and defaults to http://petstore.swagger.io:80/v2
+# See configuration.py for a list of all supported configuration parameters.
+configuration = petstore_api.Configuration(
+    host = "http://petstore.swagger.io:80/v2"
+)
+
+
+# Enter a context with an instance of the API client
+with petstore_api.ApiClient() as api_client:
+    # Create an instance of the API class
+    api_instance = petstore_api.FakeApi(api_client)
+    
+    # example, this endpoint has no required or optional parameters
+    try:
+        # Array of Enums
+        api_response = api_instance.get_array_of_enums()
+        pprint(api_response)
+    except petstore_api.ApiException as e:
+        print("Exception when calling FakeApi->get_array_of_enums: %s\n" % e)
+```
+
+### Parameters
+This endpoint does not need any parameter.
+
+### Return type
+
+[**array_of_enums.ArrayOfEnums**](ArrayOfEnums.md)
+
+### Authorization
+
+No authorization required
+
+### HTTP request headers
+
+ - **Content-Type**: Not defined
+ - **Accept**: application/json
+
+### HTTP response details
+| Status code | Description | Response headers |
+|-------------|-------------|------------------|
+**200** | Got named array of enums |  -  |
 
 [[Back to top]](#) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to Model list]](../README.md#documentation-for-models) [[Back to README]](../README.md)
 


samples/openapi3/client/petstore/python-experimental/petstore_api/api/fake_api.py
```diff
@@ -600,6 +601,109 @@ def __fake_outer_string_serialize(
...
+            )
+            kwargs['_check_return_type'] = kwargs.get(
+                '_check_return_type', True
+            )
+            kwargs['_host_index'] = kwargs.get('_host_index', 0)
+            return self.call_with_http_info(**kwargs)
+
+        self.get_array_of_enums = Endpoint(
+            settings={
+                'response_type': (array_of_enums.ArrayOfEnums,),
```
jirikuncar, line 665: ❗

samples/openapi3/client/petstore/python-experimental/petstore_api/api/fake_api.py
```diff
@@ -36,6 +36,7 @@
 )
 from petstore_api.models import health_check_result
 from petstore_api.models import outer_composite
+from petstore_api.models import array_of_enums
```
jirikuncar, line 39: ❗

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -1703,6 +1716,10 @@ components:
         - placed
         - approved
         - delivered
+    ArrayOfEnums:
+      type: array
+      items:
+        $ref: '#/components/schemas/OuterEnum'
```
spacether, line 1722: Does the model AnimalFarm already demonstrate your use case?

spacether:
My investigation shows that this line:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L420
Is preventing these models from being generated
because the array model does not have any validations or enums.
A couple of solutions that come to mind here are:
1. define the data_type in this case as `[outer_enum. OuterEnum]`. This is probably the simpler solution to get working.
2. uncomment the suppression of these models and ensure that these models are generated correctly. When I do this is see that the openapi_types data in the produced models is incorrect. Also the produced models are subclases of the ModelNormal class when they should be subclasses of the ModelSimple class

What do you think about each of these possible solutions?

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
jirikuncar: I haven't seen it used as top level object. I can use it here instead. WDYT?

spacether: What do you mean by top level object? Do you mean used in an endpoint?
Testing shows that the same code path is hit with that model and this one so if you can use that schema and avoid adding this schema it would be preferred because it will keep our sample spec smaller.

jirikuncar: As I was not sure about the purpose, I have initially created new model. I am using the `AnimalFarm` now.

jirikuncar:
I have added `FarmApi` with `/farm/animals` endpoint returning array of `Animal`s.

samples/openapi3/client/petstore/python-experimental/docs/AnimalFarm.md
```diff
@@ -0,0 +1,9 @@
+# animal_farm.AnimalFarm
+
+## Properties
+Name | Type | Description | Notes
+------------ | ------------- | ------------- | -------------
```
spacether, line 5: Can you fix the doc template so it lists the information for the AnimalFarm model here?
Name: N/A
Type: [animal.Animal]

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
```diff
@@ -1097,6 +1097,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HealthCheckResult'
+  /farm/animals:
+    get:
+      tags:
+        - farm
```
spacether, line 1103: This tag has generated a new api which adds another file.
Can you change it to one of the existing tags like `pet` or `fake`
That way a new file will not be generated

samples/openapi3/client/petstore/python-experimental/petstore_api/api/farm_api.py
```diff
@@ -0,0 +1,373 @@
+# coding: utf-8
```
spacether, line 1: Can you delete this file after you switch the api tag to put the farm endpoint into the fake api?

samples/openapi3/client/petstore/python-experimental/test/test_animal_farm.py
```diff
@@ -0,0 +1,43 @@
...
+        pass
+
+    def tearDown(self):
+        pass
+
+    def testAnimalFarm(self):
+        """Test AnimalFarm"""
+        # FIXME: construct object with mandatory attributes with example values
+        # model = AnimalFarm()  # noqa: E501
+        pass
```
spacether, line 39: How about adding a test here like
animal = [animal.Animal(blah)]
model = AnimalFarm(animal)
assert model[0] === animal

- How about adding a test in test/test_fake_api.py where we send an instance of AnimalFarm to the endpoint that accepts it?

- How about adding a deserialization test like we have [here](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/tests/test_deserialization.py#L86) to show that our code can receive back array models and deserialize them correctly?

samples/openapi3/client/petstore/python-experimental/test/test_drawing.py
```diff
@@ -42,130 +47,12 @@ def setUp(self):
...
-        Validate the scenario when the type of a OAS property is 'oneOf', and the 'oneOf'
-        schema is specified as a reference ($ref), not an inline 'oneOf' schema.
-        Further, the 'oneOf' schema has the 'nullable' attribute (as introduced in
-        OpenAPI 3.0 and deprecated in 3.1).
-        """
+    def testDrawing(self):
+        """Test Drawing"""
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Drawing()  # noqa: E501
+        pass
```
spacether, line 54: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_fruit.py
```diff
@@ -39,184 +39,10 @@ def tearDown(self):
 
     def testFruit(self):
         """Test Fruit"""
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Fruit()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_fruit_req.py
```diff
@@ -39,142 +39,10 @@ def tearDown(self):
 
     def testFruitReq(self):
         """Test FruitReq"""
+        # FIXME: construct object with mandatory attributes with example values
+        # model = FruitReq()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_gm_fruit.py
```diff
@@ -39,193 +39,10 @@ def tearDown(self):
 
     def testGmFruit(self):
         """Test GmFruit"""
+        # FIXME: construct object with mandatory attributes with example values
+        # model = GmFruit()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_mammal.py
```diff
@@ -39,11 +44,9 @@ def tearDown(self):
...
     def testMammal(self):
         """Test Mammal"""
-
-        # tests that we can make a BasquePig by traveling through descendant discriminator in Pig
-        model = Mammal(class_name="BasquePig")
-        from petstore_api.model import basque_pig
-        assert isinstance(model, basque_pig.BasquePig)
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Mammal()  # noqa: E501
+        pass
```
spacether, line 49: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_outer_enum.py
```diff
@@ -29,16 +29,9 @@ def tearDown(self):
...
-        self.assertIsNone(inst)
-
-        inst = OuterEnum('approved')
-        assert isinstance(inst, OuterEnum)
-
-        with self.assertRaises(petstore_api.ApiValueError):
-            OuterEnum('garbage')
+        # FIXME: construct object with mandatory attributes with example values
+        # model = OuterEnum()  # noqa: E501
+        pass
```
spacether, line 34: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_parent_pet.py
```diff
@@ -39,12 +39,9 @@ def tearDown(self):
...
         """Test ParentPet"""
-
-        # test that we can make a ParentPet from a ParentPet
-        # which requires that we travel back through ParentPet's allOf descendant
-        # GrandparentAnimal, and we use the descendant's discriminator to make ParentPet
-        model = ParentPet(pet_type="ParentPet")
-        assert isinstance(model, ParentPet)
+        # FIXME: construct object with mandatory attributes with example values
+        # model = ParentPet()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_quadrilateral.py
```diff
@@ -39,10 +39,9 @@ def tearDown(self):
 
     def testQuadrilateral(self):
         """Test Quadrilateral"""
-        instance = Quadrilateral(shape_type="Quadrilateral", quadrilateral_type="ComplexQuadrilateral")
-        assert isinstance(instance, complex_quadrilateral.ComplexQuadrilateral)
-        instance = Quadrilateral(shape_type="Quadrilateral", quadrilateral_type="SimpleQuadrilateral")
-        assert isinstance(instance, simple_quadrilateral.SimpleQuadrilateral)
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Quadrilateral()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_shape.py
```diff
@@ -39,71 +39,10 @@ def tearDown(self):
...
-                   "property '{}'='{}' at path: ()"
-                   )
-        with self.assertRaisesRegexp(
-                petstore_api.ApiValueError,
-                err_msg.format("shapeType", "Circle")
-        ):
-            Shape(shape_type="Circle")
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Shape()  # noqa: E501
+        pass
```
spacether, line 44: Can you revert this file to the master branch version?

samples/openapi3/client/petstore/python-experimental/test/test_triangle.py
```diff
@@ -44,12 +44,9 @@ def tearDown(self):
...
         """Test Triangle"""
-        tri = Triangle(shape_type="Triangle", triangle_type="EquilateralTriangle")
-        assert isinstance(tri, equilateral_triangle.EquilateralTriangle)
-        tri = Triangle(shape_type="Triangle", triangle_type="IsoscelesTriangle")
-        assert isinstance(tri, isosceles_triangle.IsoscelesTriangle)
-        tri = Triangle(shape_type="Triangle", triangle_type="ScaleneTriangle")
-        assert isinstance(tri, scalene_triangle.ScaleneTriangle)
+        # FIXME: construct object with mandatory attributes with example values
+        # model = Triangle()  # noqa: E501
+        pass
```
spacether, line 49: Can you revert this file to the master branch version?

modules/openapi-generator/src/test/resources/3_0/python-experimental/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
jirikuncar: I would prefer to have it in separate API.

samples/openapi3/client/petstore/python-experimental/test/test_animal_farm.py
jirikuncar: @spacether I will go back to my fake example. There are too many issues with `Animal` models.

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
```diff
@@ -42,6 +43,12 @@ def test_configuration(self):
             config.disabled_client_side_validations = 'foo'
         config.disabled_client_side_validations = ""
 
+    def test_array_of_enums(self):
+        data = [
+            "placed", None
+        ]
+        response = MockResponse(data=json.dumps(data))
+        deserialized = self.api_client.deserialize(response, (array_of_enums.ArrayOfEnums, ), True)
```
jirikuncar, line 51: @spacether this does not work. I am going with following hack meantime:
```suggestion
        deserialized = self.api_client.deserialize(response, (list, ), True)
```

jirikuncar:
@spacether what is the intended use for `Model(Simple|Normal)` with lists?

samples/openapi3/client/petstore/python-experimental/docs/AnimalFarm.md
jirikuncar: The problem is that `{{#vars}}` is empty for arrays.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -843,7 +841,20 @@ public CodegenModel fromModel(String name, Schema schema) {
                         postProcessModelProperty(result, prop);
                     }
                 }
+            } else if (result.isArrayModel) {
+                LOGGER.warn("\n\n{}\n\n", modelSchema);
+                LOGGER.warn("\n\n{}\n\n", modelProperty);
```
jirikuncar, line 847:
```suggestion
```

samples/openapi3/client/petstore/python-experimental/docs/AnimalFarm.md
jirikuncar: @spacether solved - see https://github.com/OpenAPITools/openapi-generator/pull/6493/files#diff-ec679c7ef213544fd9e5fdad21ffd6eeR6

jirikuncar:
@spacether it should all be fixed now. I have run the sample generator locally but it is failing on CI. I don't know why 😞

jimschubert:
@spacether looks like all your comments have been addressed here, is this ok to merge?
We can fix merge conflicts and regenerate via separate PR if you have no further concerns.

spacether:
I will need a couple of days to dig in to this. It looks like the PR broke data types of unrelated samples so no I don't think that we should merge yet.

jirikuncar:
> PR broke data types of unrelated samples

@spacether can you be more specific? The changes for additionalProperties seem to come from master. I will try to regenerate samples again.

spacether:
The changes that I mentioned are the additionalProperties ones. If they came from master wouldn't rebasing on master make them disappear?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -829,13 +827,13 @@ public CodegenModel fromModel(String name, Schema schema) {
...
         // make non-object type models have one property so we can use it to store enums and validations
-        if (result.isAlias || result.isEnum) {
+        if (result.isAlias || result.isEnum || result.isArrayModel) {
             Schema modelSchema = ModelUtils.getSchema(this.openAPI, result.name);
             CodegenProperty modelProperty = fromProperty("value", modelSchema);
-            if (modelProperty.isEnum == true || modelProperty.hasValidation == true) {
+            if (modelProperty.isEnum == true || modelProperty.hasValidation == true || result.isArrayModel) {
                 // these models are non-object models with enums and/or validations
                 // add a single property to the model so we can have a way to access validations
-                result.isAlias = true;
```
spacether, line 838: Array models are making a model that aliases the array type, so I think that this should be true.
Long term my goal is to have the ModelSimple classes inherrit from their base classes like str int etc.
Why do we need to set this to false for Arraymodels? Historically generateAliasAsModel applied only to Map and Array models in most generators. In python-experimental we apply it to those and all primitives only if they have validations and ebums. Standard processing in the generators was to describe all these primitive models as their simple types like str, list. But if those models had validations and enums that data was lost in that process. So I added code to preserve that information and make object type models which preserve the validation information which would have been lost.

After I did that CodegenModel was updated to include validations which previously had been missing. So in summary this code here is an older hack to preserve information. If we can use the existing ArrayModels as is that is preferable. It should define its datatype somewhere inside the CodegebModel and that should be used to set the ModelSimple data type in the python class for our made up calue parameter.

We can do this by doing a check of the model type when we assign the openapi types in ModelSimple.
We can check the data that is available by running the generator with a debugmodels flag.

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
```diff
@@ -270,7 +270,7 @@ public void arrayModelTest() {
         Assert.assertEquals(cm.classname, "sample.Sample");
         Assert.assertEquals(cm.classVarName, "sample");
         Assert.assertEquals(cm.description, "an array model");
-        Assert.assertEquals(cm.vars.size(), 0);
+        Assert.assertEquals(cm.vars.size(), 1);  // there is one value for Childer definition
```
spacether, line 273: What is the dataType for this sample?
It having vars of length 1 shows that my older hack is being used and that we have hoisted model information into a new variable.
How about we keep the model as is an use the information from
CodegenModel.dataType or CodegenModel.arrayModelType?
We can see all of the properties that are available by looking at that class [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L29)

We can debug the CodegenModels and see the information in them by running the generate command with the `-DdebugModels=true` as [shown here](https://github.com/openapitools/openapi-generator/wiki/FAQ#how-to-debug-openapi-generator)

How about in the classvars template here
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache#L115
Doing a check to see if the model is an arrayModel
and setting the datatype accordingly?
It would look like:
```
{{#isArrayModel}}
            'value': ({{{dataType}}},),  # noqa: E501
{{/isArrayModel}}
{{^isArrayModel}}
{{#requiredVars}}
            '{{name}}': ({{{dataType}}},),  # noqa: E501
{{/requiredVars}}
{{#optionalVars}}
            '{{name}}': ({{{dataType}}},),  # noqa: E501
{{/optionalVars}}
{{/isArrayModel}}
```
We would also need to do a similar change in 
[method_init_shared](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_shared.mustache#L2)
To:
- not loop over required an optional args if we have an array model, if we are that case, just set value as a required positional arg
- else use our existing looping over variables
- in the docstring if we are an arraymodel set value as a required positional arg

@jirikuncar 
What do you think about making these changes?
If you don't want to make them, we can keep using the existing code/(my older solution) where we shovel the model data into a value variable.

jirikuncar:
> The changes that I mentioned are the additionalProperties ones. If they came from master wouldn't rebasing on master make them disappear?

It should but there is some problem with the sample generation. I will continue updating my branch until it resolves.

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
jirikuncar: I would prefer to keep the current code. It looks like too many changes and `{{#isArrayModel}}` guards are required for templates.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
jirikuncar: I am not aware of all the corner cases in the code. I was trying to make it work for my use-case and not break existing tests. Feel free to propose code changes.

jirikuncar: I hope I have found a sufficient fix in b2ca2f78e295ce1d94b42f6ebc55d9d5a64f1918.

jirikuncar:
@jimschubert I have run the `./bin/openapi3/python-experimental-petstore.sh` and also tried `./bin/utils/ensure-up-to-date --batch` but the generated code for additional properties is still there.

jirikuncar:
@spacether can we go with this solution for a time being?
@jimschubert the CI is passing 🎉

modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java
spacether: Ah, understood. That's fine.

PR #6495, opened 2020-05-30
[java][jersey2] Add support for discriminator, fix nullable typo and nullable deserialization

sebastien-rosset:
1. Fix `nullable` typo, the mustache template had `isNulalble`
1. Add support for discriminator in jersey2 library.
1. The deserialization code was incorrectly accepting `null` as input data for non-nullable anyOf/oneOf schemas. If the input data is `null` and the anyOf/oneOf does not accept null (either because nullable: true or `type: 'null'` is a child anyOf/oneOf, then a deserialization exception should be raised. Note: it's likely this problem exists for other schema types too.
1. When there is more than one level of `anyOf`/`oneOf`, the deserialization was previously failing. This is shown with the existing sample OAS 3.0 doc that works with the python and go clients.

Newly discovered issues:
1. The existing jersey2 implementation of oneOf/anyOf is incomplete. It is supposed to validate the input data against the JSON schema, but it does not validate any constraint (enum, min, max...)
1. Other issues listed with TODO comments in the Unit test file.

### PR checklist


CI/samples.ci/client/petstore/java/test-manual/jersey2-java8/JSONComposedSchemaTest.java
```diff
@@ -46,23 +38,126 @@ public void testOneOfSchemaWithDiscriminator() throws Exception {
...
+        // It is not valid to deserialize a equilateral triangle into a quadrilateral.
+        Exception exception = assertThrows(JsonMappingException.class, () -> {
+            json.getContext(null).readValue(str, Quadrilateral.class);
+        });
+        assertTrue(exception.getMessage().contains("Failed deserialization for Quadrilateral: 0 classes match result"));
+    }
+
+    @Test
+    public void testOneOfNestedComposedSchema() throws Exception {
+        /*
```
sebastien-rosset, line 115: @wing328 , I have temporarily commented out this code. If I enable it, then deserialization fails. It looks like there is a problem when a compose schema (Drawing) has the `additionalProperties` keyword, in which case the generated Drawing.class extends from HashMap<String, X>

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/anyof_model.mustache
```diff
@@ -48,6 +52,19 @@ public class {{classname}} extends AbstractOpenApiSchema{{#vendorExtensions.x-im
...
+        /**
+         * Handle deserialization of the 'null' value.
+         */
+        @Override
+        public {{classname}} getNullValue(DeserializationContext ctxt) throws JsonMappingException {
+        {{#isNullable}}
+            return null;
+        {{/isNullable}}
+        {{^isNullable}}
+            throw new JsonMappingException("{{classname}} cannot be null");
```
sebastien-rosset, line 77: @wing328 , is there a better way to fail deserialization of the `null` value that would work across all OAS schema types?

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/anyof_model.mustache
```diff
@@ -69,6 +86,7 @@ public class {{classname}} extends AbstractOpenApiSchema{{#vendorExtensions.x-im
         schemas.put("{{{.}}}", new GenericType<{{{.}}}>() {
         });
         {{/anyOf}}
+        JSON.registerDescendants({{classname}}.class, Collections.unmodifiableMap(schemas));
```
sebastien-rosset, line 101: @wing328 , it's a bit hacky for now, I'd rather not use a public method, but it's needed to have proper deserialization when there are multiple levels of oneOf/anyOf.

wing328:
The CircleCI failed with the following errors (partial):
```
[ERROR] location: class JSON
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/JSON.java:[256,45] error: cannot find symbol
[ERROR] symbol:   class IsoscelesTriangle
[ERROR] location: class JSON
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/JSON.java:[257,43] error: cannot find symbol
[ERROR] symbol:   class ScaleneTriangle
[ERROR] location: class JSON
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/JSON.java:[258,36] error: cannot find symbol
[ERROR] symbol:   class Triangle
[ERROR] location: class JSON
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/JSON.java:[259,30] error: cannot find symbol
[ERROR] symbol:   class Triangle
[ERROR] location: class JSON
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] 
 ```

I'll merge it into a branch first and further improve it.

sebastien-rosset:
For additionalProperties, I propose to use a Java field of type Map<String, X> instead of currently using class inheritance. The reason is that 1) `additionalProperties` keyword can be used with `allOf` composed schema, and 2) it makes more sense to use class inheritance for `allOf` rather than additionalProperties.

This would require a custom java deserializer.

Issue #6497, opened 2020-05-30
[BUG] [Javascript] Modify parseDate in ApiClient to support epoch time in addition to ISO-8601


tray2100:
##### Description

The APIs from the Javascript client generated seem to receive Date values in epoch millis instead of ISO-8601. However, it seems that the APIs (specifically the constructFromObject function) doesn't handle converting from epoch millis and results in 'Invalid Date'. This is happening because the long value is being converted to a string before being passed to the Date object constructor.

The solution here would be to just check if the incoming object type and if it's numeric then just pass it on to Date without wrapping it in a string first.

##### openapi-generator version

4.3.1

##### Suggest a fix
Modify [this portion](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache#L548-L550) of the mustache file to look like this:

```javascript
  exports.parseDate = function(str) {
    if (!isNaN(str)) {
      return new Date(+str);
    }
    return new Date(str.replace(/T/i, ' '));
  };
```

PR #6504, opened 2020-05-31
adding handling for epoch dates in javascript ApiClient mustache file…

tray2100:
… (#6497)

Added simple branching logic where a number is sent directly to the Date constructor or execute the previous logic.

### PR checklist

cc @CodeNinjai @frol @cliffano

### DIFF
--- a/modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache
@@ -541,12 +541,15 @@
 {{/usePromises}}  };
 
 {{#emitJSDoc}}  /**
-   * Parses an ISO-8601 string representation of a date value.
+   * Parses an ISO-8601 string representation or epoch representation of a date value.
    * @param {String} str The date value as a string.
    * @returns {Date} The parsed date object.
    */
 {{/emitJSDoc}}  exports.parseDate = function(str) {
-    return new Date(str.replace(/T/i, ' '));
+    if (isNaN(str)) {
+      return new Date(str.replace(/T/i, ' '));
+    }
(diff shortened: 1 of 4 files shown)


tray2100:
Pushed the recent upstream changes to this PR...

tray2100:
cc @wing328

modules/openapi-generator/src/main/resources/Javascript/ApiClient.mustache
```diff
@@ -541,12 +541,15 @@
...
 
 {{#emitJSDoc}}  /**
-   * Parses an ISO-8601 string representation of a date value.
+   * Parses an ISO-8601 string representation or epoch representation of a date value.
    * @param {String} str The date value as a string.
    * @returns {Date} The parsed date object.
    */
 {{/emitJSDoc}}  exports.parseDate = function(str) {
-    return new Date(str.replace(/T/i, ' '));
+    if (isNaN(str)) {
```
wing328, line 549: > The Number.isNaN() method determines whether a value is NaN (Not-A-Number).

Ref: https://www.w3schools.com/jsref/jsref_isnan_number.asp

Did you mean to check if str is a number?

tray2100: No I meant to check if str is **not** a number. If it's not a number then we'll default to the existing logic otherwise it's a number that we can just pass into Date.

wing328: OK. After reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN, I think we can go with your fix.

Issue #6333, opened 2020-05-16
[REQ] Feature Request - speed up + simplify sample generation


spacether:
### Is your feature request related to a problem? Please describe.

- Local sample generation with `./bin/utils/ensure-up-to-date` without batch invocation takes a while, ~15 minutes.
- When users update [.sh](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/ada-petstore.sh) and [.bat](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/windows/android-petstore.bat) files, sometimes they forget to update the [json config files](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/csharp-petstore-net-40.json)
- the .sh/.bat/.json config files offer three ways to do the same thing (provide generator inputs). If we use only one way, that will force our users to update the correct files

## Describe the solution you'd like

Desired changes:
- always use batch_mode=True [here](https://github.com/OpenAPITools/openapi-generator/blob/master/bin/utils/ensure-up-to-date#L15)
- convert all the .sh and .bat files to store their inputs in json or yaml files like https://github.com/OpenAPITools/openapi-generator/blob/master/bin/csharp-petstore-net-40.json
- eliminate the generator specific .sh and .bat files
- perhaps allow passing in a single generator name into ensure-up-to-date to allow generation of individual generators, like the old .sh and .bat files

## Describe alternatives you've considered

N/A

## Additional context

This ticket was created from a discussion with @jimschubert on Slack

jimschubert:
Thanks for raising this issue…

I think most of the work for this would just be migrating from `*.sh` and `*.bat` files to JSON or YAML configs. Many of our shell scripts already take JSON config inputs in the `bin` directory.

Two things we'll need to consider:

* A few scripts use custom logic such as `bin/java-petstore-feign.sh`
* There may be cases where we'd want to support the `*POST_PROCESS_FILE` environment variables, but I think those should be avoided in general if we're requiring the community to also regenerate these samples

If we move to config-based generation via the `batch` command we may see only trivial performance gains on CI where we may only have 1-2 CPUs, but the target here would be improving the process for all contributors.

Going this route, I wonder also if it might make sense to sync config scripts with generator names. For example, have structures such as:

```
bin/samples/scala-akka/*.json
bin/samples/csharp-netcore/*.json
```

[...]

PR #6509, opened 2020-06-01
[ci][cli] Moving ensures script to config-based batch generation of samples

jimschubert:
All samples scripts are now config based with a single bash entry point: `bin/generate-samples.sh`.

File cleanup using the `FILES` file will be done later.

Elm is currently skipped as it requires that we skip validation of the spec, so will need to be configured.

Running against the 179 configurations in `bin/configs` I see the following stats:

```
[SUCCESS] Batch generation completed successfully.
./bin/generate-samples.sh  54.20s user 6.64s system 243% cpu 24.954 total
```

This should address conversation in #6333.

### PR checklist


spacether:
What is the `other` folder meant to tell us or mean?
https://github.com/OpenAPITools/openapi-generator/tree/ensures/bin/configs/other

spacether:
How about changing the name of the configs folder to something like `samples_configs` to convey that these are the config files for sample generation? configs is very generic and `samples_configs` helps people understand that this folder exists to configure our sample generation.

jimschubert:
I'll write up more when this isn't a wip. But "other" are those that aren't in the ensure-up-to-date script in current master.

As for the suggested rename of the folder, I don't think that's necessary. When I'm done and I've updated docs, it'll be obvious enough what these configs are, especially with the naming of each yaml file.

jimschubert:
Notice the lines in appveyor.yml:

```
### TODO: Execute all generators via powershell or other
  # generate all petstore clients
  # - .\bin\windows\run-all-petstore.cmd
  # generate all petstore clients (openapi3)
  # - .\bin\openapi3\windows\run-all-petstore.cmd
```

GitHub Workflows has bash available on Windows, so I'd actually like to move the above commands to GitHub Workflows after this merge and reuse the single script at `bin/generate-samples.sh`.

jimschubert:
Just a heads up… bitrise will fail because it is running off of the master configuration file.

Here's a summary of what I've done.

I've moved us away from managing multiple os-dependent scripts with occasional accompanying json config files which sometimes differed by OAS version to configuration-only generation. This allows us to execute samples in batch and reduces sample generation time from around 19 minutes on my machine to under 1 min. I have seen stats from other contributors where this generation takes 30-40 minutes.

I'd previously attempted to migrate all "mature" generators (i.e. those in ensure-up-to-date) to configuration based under `bin/ci` to allow for batch generation by contributors. The problem is that we'd have updates in the bin/*.sh script and not in the bin/ci/*.json as well. We also saw some shell scripts which pointed to external configs would have new parameters added within the shell script rather than the external config.

This enforces a convention and will allow us to provide a `--clean` option to the batch generation which reads in the `FILES` cached list of generated files and deletes pre-existing files. This would address an issue where we sometimes have generator changes which introduce or remove files and results in CI failures due to those stale files.

Here's performance on my machine generating _all_ samples:

```
time ./bin/generate-samples.sh bin/{configs,configs/other}/*.yaml
…
[SUCCESS] Batch generation finished 276 generators successfully.

real	0m35.528s
user	1m16.225s
sys	0m9.098s
```

To put this into perspective, generating all samples prior to this change took 45 minutes or more.

In this PR, I have the generators which were previously included in the `ensure-up-to-date` file under `bin/configs`. All other generators (bash, windows, openapi3) which were not in that golden list are under `bin/configs/other`. I think with the performance improvement we could move these all under the single directory.

This pull request includes samples generation for the "golden" samples and not for those under `bin/configs/other`.

TODO:
- [ ] Either add back in sample regeneration in AppVeyor or create a GitHub Workflow where bash is available on Windows build machines
- [ ] Verify swift generations after merge since bitrise pulls old locations from master

jimschubert:
@spacether This is good to review now if you want to give it another pass.

CI/.drone.yml
```diff
@@ -38,6 +38,6 @@ steps:
   # generate test scripts
   - /bin/bash bin/tests/run-all-test
   # generate all petstore samples (client, servers, doc)
```
spacether, line 40: Are these for v2.X samples only? Can you include that in your comment on line 40?

jimschubert: I don't understand how that provides clarity to the comment... "All petstore samples" means everything.

jimschubert: btw something you might not be aware of is that all samples are being converted to OAS 3.

jimschubert:
Regarding the two TODO items... swift checks will have to be post-merge. Also for appveyor, we may consider using the following GitHub workflow so we have consistency in how we execute the script via bash:

```
name: Build
on:
  push:
    branches:  
      - master
  pull_request:
    branches:
      - master

jobs:
  build:
    name: Build on JDK ${{ matrix.java }} and ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        java: [8, 11]
        os: [ubuntu-latest, windows-latest]
        
    steps:
      - name: Check out code
        uses: actions/checkout@v2

      - name: Set up JDK ${{ matrix.java }}
        uses: actions/setup-java@v1 
        with:
          java-version: ${{ matrix.java }}
          
      - uses: actions/cache@v1
        with:
          path: ~/.m2/repository
          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
          restore-keys: |
            ${{ runner.os }}-maven-
        
      - name: Build with Maven
        shell: bash
        run: mvn -nsu -B --quiet -Djacoco.skip=true -Dorg.slf4j.simpleLogger.defaultLogLevel=error --no-transfer-progress clean install --file pom.xml ${{ matrix.flags }}
        
      - name: Test gradle
        shell: bash
        run: gradle -b modules/openapi-generator-gradle-plugin/samples/local-spec/build.gradle buildGoSdk --stacktrace

      - name: Upload Maven build artifact
        uses: actions/upload-artifact@v1
        if: matrix.java == '8'
        with:
          name: artifact
          path: modules/openapi-generator-cli/target/openapi-generator-cli.jar

  verify:
    name: Verifies integrity of the commit on ${{ matrix.os }}
    needs: build
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest]
    steps:
      - name: Check out code
        uses: actions/checkout@v2
      - name: Download build artifact
        uses: actions/download-artifact@v1
        with:
          name: artifact
      - name: Run Ensures Script
        run: |
          mkdir -p modules/openapi-generator-cli/target/
          mv artifact/openapi-generator-cli.jar modules/openapi-generator-cli/target/
          ./bin/utils/ensure-up-to-date
```

I've tested this in a private repository against this branch and it takes about 12 minutes to:

* Build Java 8 (Linux)
  - smoke test generation via gradle
  - (for Java 8 only) Cache the jar for use in the verification step
* Build Java 8 (Windows)
  - smoke test generation via gradle
* Build Java 11 (Linux)
  - smoke test generation via gradle
* Build Java 11 (Windows)
  - smoke test generation via gradle
* Run ensure-up-to-date (Linux)
* Run ensure-up-to-date (Windows)

The 4 build jobs run in parallel, once completed the verify job runs Linux+Windows in parallel.

As a separate verification, we could potentially move a couple sample tests here such as Go and Swift.

CI/.drone.yml
spacether: Ah, I wasn't aware of that

CI/.drone.yml
```diff
@@ -38,6 +38,6 @@ steps:
...
-  - /bin/bash bin/run-all-petstore
+  - /bin/bash bin/generate-samples.sh
   # generate all petstore samples (openapi3)
-  - /bin/bash bin/openapi3/run-all-petstore
+  - /bin/bash bin/generate-samples.sh bin/configs/other/*.yaml
```
spacether, line 43: So this other folder naming is still confusing to me
This comment says that ll samples in the other folder are v3 samples
looking in them I see v3 samples. Is this comment incorrect?
Are these v2 or v3?
Are they non-petstore samples?
Can you clarify this in the comment or the folder name?

jimschubert: Everything will move to openapi3 shortly, so this comment will only be stale for a brief time.

spacether: So these are v2 samples only then?

CI/bitrise.yml
```diff
@@ -38,7 +38,7 @@ workflows:
 
             set -e
 
-            sh bin/swift5-all.sh
+            sh bin/generate-samples.sh bin/configs/swift5-*.yaml
```
spacether, line 41: Aren't these samples also run in Drone when we invoke `/bin/bash bin/generate-samples.sh` why run them twice?

CI/samples.ci/client/petstore/java/test-manual/jersey2/ApiClientTest.java
```diff
@@ -1,249 +0,0 @@
...
-            assertEquals(1, pairs.size());
-
-            String delimiter = collectionFormatMap.get(collectionFormat);
-            String[] pairValueSplit = pairs.get(0).getValue().split(delimiter);
-
-            // must equal input values
-            assertEquals(values.size(), pairValueSplit.length);
-        }
-    }
-}
```
spacether, line 249: Note: this still exists at https://github.com/OpenAPITools/openapi-generator/blob/ensures/samples/client/petstore/java/jersey2-java8/src/test/java/org/openapitools/client/ApiClientTest.java

CI/samples.ci/client/petstore/java/test-manual/okhttp-gson/JSONTest.java
```diff
@@ -1,202 +0,0 @@
...
-        TimeZone tz = TimeZone.getDefault();
-        Calendar cal = GregorianCalendar.getInstance(tz, Locale.ROOT);
-        int offsetInMillis = tz.getOffset(cal.getTimeInMillis());
-
-        String offset = String.format(Locale.ROOT,"%02d:%02d", Math.abs(offsetInMillis / 3600000), Math.abs((offsetInMillis / 60000) % 60));
-        offset = (offsetInMillis >= 0 ? "+" : "-") + offset;
-
-        return offset;
-    }
-}
```
spacether, line 202: Note: this still exists at https://github.com/OpenAPITools/openapi-generator/blob/ensures/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/JSONTest.java

CI/samples.ci/client/petstore/java/test-manual/okhttp-gson/api/PetApiTest.java
```diff
@@ -1,564 +0,0 @@
...
-            this.contentLength = contentLength;
-            this.done = done;
-        }
-
-        @Override
-        public String toString() {
-            return "<Progress " + bytes + " " + contentLength + " " + done + ">";
-        }
-    }
-}
```
spacether, line 564: Note: this still exists at https://github.com/OpenAPITools/openapi-generator/blob/ensures/samples/client/petstore/java/okhttp-gson/src/test/java/org/openapitools/client/api/PetApiTest.java

CI/bitrise.yml
jimschubert: Different CI run in parallel and don't have a way to know this has been run.

See my comment in this PR about potentially using GitHub Workflow

CI/.drone.yml
jimschubert: No. As mentioned in the PR, these are the samples which were not previously included in our ensure script.

Any v2 samples in this directory won't be v2 for long, so the incorrectness of the comment is only temporary.

We don't really need to be flipping back and forth between words in comments to have them be correct across 20 or so commits (however many it'll take for conversion). There's a ton more that we can focus our energies on.

Plus, if CI is already running or queued, making a change to a few comments wastes the time of authors and reviewers. It's also bad for the environment if we're churning out unnecessary builds over and over.

CI/samples.ci/client/petstore/java/test-manual/jersey2/ApiClientTest.java
jimschubert: What are there notes for, and who are they directed at?

spacether: They are for me and other reviewers wondering if these tests still exist

CI/.drone.yml
spacether: I understand that changing a comment kicks off a new CI run, and that doing a ton of them is not good. One of the points that you highlighted was that before, people were updating incorrect locations for generators. Some reasons that they were doing this was that multiple locations were possible and also because they misunderstood what to do. My question is looking to clear up any misunderstanding that users have when trying to understand what the other folder is for. In my opinion kicking off one more CI run is a very low cost to clearing up confusion. But it's fine if you want to keep it as is.

CI/samples.ci/client/petstore/java/test-manual/jersey2/ApiClientTest.java
jimschubert: I see. It wasn't clear if you were making yourself a note, suggesting to me that I missed a file delete, or what.

Some of our generators were deleting generated sample directories, then copying these files into the generated test directories. These files always existed in the targeted output directory, which is why GitHub doesn't show a move operation here. The delete + copy is just legacy practice which isn't really needed.

There's no reason to delete and copy these test files anymore because OpenAPI Generator doesn't overwrite existing tests and we support and honor the ignore file in the targeted directory.

Those generators which previously deleted entire samples directories and regenerated into them were probably trying to address the issue of spec files changing (either changing models, adding or removing types, etc) which could leave stale files. We've long since moved toward separating specs according to generators and only making additive changes, so much of that original sample scripting is no longer necessary.

Also as mentioned in the PR description, cleanup using the `FILES` listing will be done later. At that point, we no longer need deletes or shuffling, we can very easily clean existing samples according to what files were generated by the previous run. We may end up with a stale file occasionally causing compilation failure but I'd take that over managing 600+ files separated by OS, OAS version, and shuffling around files rather than using the ignore file functionality.

CI/.drone.yml
jimschubert: I reviewed and I think I only mentioned this in chat, so I'll reproduce it here…

I'd personally like to include all these yaml in a single directory. It only takes about a minute on my machine to generate all generators.

CI/bitrise.yml
wing328: Right. In other words, the jobs run independently on different CI providers.

@spacether if you've further questions on the CI, PM me via Slack as I setup most of them.

CI/samples.ci/client/petstore/java/test-manual/jersey2/ApiClientTest.java
wing328: > There's no reason to delete and copy these test files anymore because OpenAPI Generator doesn't overwrite existing tests and we support and honor the ignore file in the targeted directory.

Just a bit more background on this (come up by @jmini if my memory serves me well). The solution aims to solve the problems that sometimes users manually empty the target output folder even though the test files are not manually overwritten during code generation. We encountered this occasionally from time to time (and yes we missed it in the PR which updates a lot of files and the test case removal becomes hard to spot) and restoring the test files did help us to ensure some tests were not removed as part of the PR.

But as I pointed out before, one drawback of this method is that it's a bit counter-intuitive to the contributors that they need to update some files under the `CI` folder in order to add more tests or adjust the tests as part of the change.

I'm ok this is dropped as part of this PR and I may be able to come up with something later for us to better spot test cases removal as part of the PR review.

wing328:
Overall it looks good 👍 

Did some tests locally (mac, windows) and here are 3 minor feedback:
- In https://github.com/OpenAPITools/openapi-generator/pull/6509/files#diff-195a635ad245ded9076330e31134bd80R8, I would suggest adding a line or 2 to ask Windows users install [Git BASH](https://gitforwindows.org/) as all the batch files are gone (Git bash works for me under Windows to run the generate script under ./bin/ to update samples) so we need put down some instructions to guide Windows users on how to update the sample.
- personally I prefer the generate script shows what files have been generated and also some generators will show useful information such as file post-processing, which will be helpful to new users/contributors.
- the script will wait for 5 seconds no matter what so I'll see if I can file a PR later to skip the 5-second wait time if there's at least one argument provided (so not updating all samples) as I update the samples a lot :)

CI/.drone.yml
spacether: A single directory is clearer longer term. Keeping it separate for now is fine. This PR has been in review for a while and this item isn't a blocking concern to me.

CI/bitrise.yml
spacether: Thank you. Yup I was aware that they run in parallel independently. What I didn't understand was why we need each of them. Jim pointed me to https://github.com/OpenAPITools/openapi-generator/wiki/Integration-Tests
Which shows that we use bitrise to test swift 4/5 so it looks like the reason that we generate them in both locations is that:
- in drone we are testing JDK11 test, OCaml client, etc and need the samples generated for that testing
- in bitrise we are specifically testing switft4/5 so we need to run sample generation prior to that

wing328:
I'll try to address the bitrise.io CI failure tomorrow and merge if no further questions/feedback on this PR.

wing328:
I'll file follow-up PRs for what I mentioned above and also for some recent changes (e.g. rust client petstore)

jimschubert:
@wing328 regarding

> personally I prefer the generate script shows what files have been generated and also some generators will show useful information such as file post-processing, which will be helpful to new users/contributors.

I guess if the input to the script is a single yaml, we could revert to calling `generate` rather than `batch`. I think that information is too verbose when running more than one generator at a time.

PR #6518, opened 2020-06-02
[Java][Jersey2] various improvements

wing328:
- add the option "oneOfDiscriminatorLookup" to speed up the oneOf schema matching process
- add additionalProperties support
- add serializer and deserializer to oneof/anyof schemas
- remove oneof/anyof logic from ApiClient

### PR checklist


samples/openapi3/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/model/Pig.java
```diff
@@ -60,12 +88,30 @@ public PigDeserializer(Class<?> vc) {
...
             JsonNode tree = jp.readValueAsTree();
+            Object deserialized = null;
+            Pig newPig = new Pig();
+            Map<String,Object> result2 = tree.traverse(jp.getCodec()).readValueAs(new TypeReference<Map<String, Object>>() {});
+            String discriminatorValue = (String)result2.get("className");
+            switch (discriminatorValue) {
+                case "BasquePig":
+                    deserialized = tree.traverse(jp.getCodec()).readValueAs(BasquePig.class);
+                    newPig.setActualInstance(deserialized);
+                    return newPig;
```
spacether, line 99: So it looks like this sets this.instance = to the deserialized instance of BasquePig where this is a Pig instance? Is that right?
Why include the BasquePig instances in this Pig instance vs just returning the instance of BasquePig?
You mentioned a use case when a client may want to change the type of an instance from
BasquePig to DanishPig.
That is possible if one knows about the ancestor class Pig, which the client would know because that is the data type that is being deserialized.
My understanding is that you want a client to be able to change a model instance class type(BasquePig/DanishPig) given only the deserialized instance and NOT having to know about the ingested data type, is that right?
Can you describe a more realistic use case than converting pig types?

wing328: > So it looks like this sets this.instance = to the deserialized instance of BasquePig where this is a Pig instance? Is that right?

Yes

>  Why include the BasquePig instances in this Pig instance vs just returning the instance of BasquePig?

Because Pig is an `oneOf`. Returning BasqueBig means the client will lose the information the return type is oneOf and DanishPig is another possibility.

>  That is possible if one knows about the ancestor class Pig, which the client would know because that is the data type that is being deserialized.

You mean DanishPig stores the oneOf schema information (Pig in this case)? What if DanishPig happens to be the oneOf schema node in "Creature" or "Animal" oneOf schema?

> My understanding is that you want a client to be able to change a model instance class type(BasquePig/DanishPig) given only the deserialized instance and NOT having to know about the ingested data type, is that right?

One can change the model (DanishPig or BasqueBig) in the client side independent of how "Pig" is created - whether it's deserialized from a payload returned by the server or it's created in the client side from scratch.

>  Can you describe a more realistic use case than converting pig types?

- The client calls the `petPig` and gets a `Pig` (which happens to be a DanishPig in this example)
- The client changes the instance from DanishPig to Snake and got an error (which is the expected behaviour according to the spec)
- The client changes the instance form DanishPig to BasqueBig and got no error
- The client calls `updatePet` with the `Pig` object (originally obtained from `getPig`) with the underlying instance set to BasqueBig in the above step
- The server checks the payload (BasqueBig) and accepts it since it conform to the `Pig` definition (oneOf)

I would suggest you step back from actual use case as OpenAPI spec is a contract and the implementation should work with respect to the contract. (this is not to say real use cases are bad)

modules/openapi-generator/src/main/resources/Java/libraries/jersey2/model.mustache
```diff
@@ -39,9 +51,16 @@ import javax.validation.Valid;
 {{#performBeanValidation}}
 import org.hibernate.validator.constraints.*;
 {{/performBeanValidation}}
+import {{invokerPackage}}.JSON;
```
symphony-youri, line 54: Hello @wing328 ,
Would you know why this import is required? I'm trying to generate code without the supporting files and it does not compile because of this import

PR #6539, opened 2020-06-04
[C++][Ue4] various bus fixes

Kahncode:
More fixes to UE4 code generator. Tested compilation under various project setups, including linux compilation. Added support for string and enum type definitions. Better handling of variable names in edge cases, should no longer produce empty variable names.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppUE4ClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppUE4ClientCodegen.java
@@ -497,13 +497,20 @@ public class CppUE4ClientCodegen extends AbstractCppCodegen {
             name = name.toLowerCase(Locale.ROOT);
         }
 
+        //Unreal variable names are CamelCase
+        String camelCaseName = camelize(name, false);
+
+        //Avoid empty variable name at all costs
+        if(!camelCaseName.isEmpty()) {
+            name = camelCaseName;
+        }
+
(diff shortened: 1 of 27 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CppUE4ClientCodegen.java
```diff
@@ -497,13 +497,20 @@ public String toVarName(String name) {
...
+
+        //Avoid empty variable name at all costs
+        if(!camelCaseName.isEmpty()) {
+            name = camelCaseName;
+        }
+
         // for reserved word or word starting with number, append _
         if (isReservedWord(name) || name.matches("^\\d.*")) {
             name = escapeReservedWord(name);
         }
```
wing328, line 511: I think the special handling of the reserved words may not work here (before and after this PR).

I'll file another PR to address it later.

wing328:
>  Better handling of variable names in edge cases

Do you have an example of the edge cases? We may add it to the test spec.

Kahncode:
Hi @wing328 The edge case encountered was "_". Because of camelize, this generates an empty variable name. Also in the previous logic, it used underscores to escape unique names and language reserved keywords, but before camelizing, so it would end up removing the underscores in the camelize step, and create invalid variable names.

wing328:
OK. Likely generators have similar issues handling just "_". Will add it to the test spec later and update other generators accordingly

PR #6557, opened 2020-06-05
[Python] Support for per-operation servers

jirikuncar:
Support `servers:` definition.

See:
- `go-experimental` https://github.com/OpenAPITools/openapi-generator/pull/4635
- `java` https://github.com/OpenAPITools/openapi-generator/pull/4998

### PR checklist


modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -155,20 +155,30 @@ conf = {{{packageName}}}.Configuration(
...
+        self._base_path = "{{{basePath}}}" if host is None else host
         """Default Base url
         """
+        self.server_index = 0 if server_index is None and host is None else server_index
+        self.server_operation_index = server_operation_index or {}
+        """Default server index
+        """
+        self.server_variables = server_variables or {}
+        self.server_operation_variables = server_operation_variables or {}
+        """Default sever variables
```
therve, line 180:
```suggestion
        """Default server variables
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -561,14 +571,14 @@ conf = {{{packageName}}}.Configuration(
             {{/servers}}
         ]
 
-    def get_host_from_settings(self, index, variables=None):
+    def get_host_from_settings(self, index, variables=None, servers=None):
         """Gets host URL based on the index and variables
         :param index: array index of the host settings
         :param variables: hash of variable and the corresponding value
```
therve, line 577: Missing servers params doc.

modules/openapi-generator/src/main/resources/python/python-experimental/api.mustache
```diff
@@ -131,7 +131,7 @@ class {{classname}}(object):
             kwargs['_check_return_type'] = kwargs.get(
                 '_check_return_type', True
             )
-            kwargs['_host_index'] = kwargs.get('_host_index', 0)
+            kwargs['_host_index'] = kwargs.get('_host_index', None)
```
therve, line 134:
```suggestion
            kwargs['_host_index'] = kwargs.get('_host_index')
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar, line 577:
```suggestion
        :param variables: hash of variable and the corresponding value
        :param servers: host settings for the endpoint
```

jirikuncar:
Can anyone please review this PR? Thank you!

jirikuncar:
@jimschubert @spacether how can I get more info on failed Travis job?
```
Loading composer repositories with package information
Updating dependencies (including require-dev)
                                                                              
  [Composer\Downloader\TransportException]                                    
  Content-Length mismatch, received 151799 bytes out of the expected 1479855  
                                                                              
install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (bundle-install) on project Slim4PetstoreServerTests: Command execution failed. Process exited with an error: 1 (Exit value: 1) -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :Slim4PetstoreServerTests
```

samples/client/petstore/python-asyncio/petstore_api/configuration.py
```diff
@@ -466,3 +480,17 @@ def get_host_from_settings(self, index, variables=None):
             url = url.replace("{" + variable_name + "}", used_value)
 
         return url
+
+    @property
+    def host(self):
+        """Return generated host."""
+        if self.server_index is None:
+            return self._base_path
```
spacether, line 488: How about removing this if block because it already exists in get_host_from_settings?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -595,3 +609,17 @@ conf = {{{packageName}}}.Configuration(
             url = url.replace("{" + variable_name + "}", used_value)
 
         return url
+
+    @property
+    def host(self):
+        """Return generated host."""
+        if self.server_index is None:
+            return self._base_path
```
spacether, line 617: How about removing this if block because it already exists in get_host_from_settings?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -155,20 +155,30 @@ conf = {{{packageName}}}.Configuration(
...
 {{#hasHttpSignatureMethods}}
                  signing_info=None,
 {{/hasHttpSignatureMethods}}
+                 server_index=None, server_variables=None,
+                 server_operation_index=None, server_operation_variables=None,
                  ):
         """Constructor
         """
-        self.host = host
+        self._base_path = "{{{basePath}}}" if host is None else host
```
spacether, line 171: How about naming this variable `_host`, which more clearly conveys that this is the default host that we use?

modules/openapi-generator/src/main/resources/python/python-experimental/api.mustache
```diff
@@ -105,7 +105,7 @@ class {{classname}}(object):
                     Default is True.
                 _host_index (int): specifies the index of the server
                     that we want to use.
-                    Default is 0.
+                    Default is read from the configuration.
```
spacether, line 108: Can you update the docstring here to show that _host_index can now be type `int/None`?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -561,14 +571,18 @@ conf = {{{packageName}}}.Configuration(
             {{/servers}}
         ]
 
-    def get_host_from_settings(self, index, variables=None):
+    def get_host_from_settings(self, index, variables=None, servers=None):
         """Gets host URL based on the index and variables
         :param index: array index of the host settings
         :param variables: hash of variable and the corresponding value
+        :param servers: host settings for the endpoint
```
spacether, line 578: Can you describe that servers should is of type (list/None) here?

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -155,20 +155,30 @@ conf = {{{packageName}}}.Configuration(
...
+    def __init__(self, host=None,
                  api_key=None, api_key_prefix=None,
                  username=None, password=None,
                  discard_unknown_keys=False,
                  disabled_client_side_validations="",
 {{#hasHttpSignatureMethods}}
                  signing_info=None,
 {{/hasHttpSignatureMethods}}
+                 server_index=None, server_variables=None,
+                 server_operation_index=None, server_operation_variables=None,
```
spacether, line 167: This method already has 8 inputs not counting self.
Adding 4 more variables here increases out input count to 12 which is a lot.
It is clearer if related variables are grouped together in dicts.
Can you restructure these 4 new additions into two dicts?
How about something like:
```
server_info=None
# which is a dict like: {index:1, variables={}}
operation_server_info=None:
# which is a dict like {index_by_operation_id: {'get_pet': 0}, variables_by_operation_id: {'get_pet': blah}}
```

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
```diff
@@ -44,6 +46,27 @@ def test_configuration(self):
...
+
+        def request(expected_url, method, url, **kwargs):
+            assert expected_url == url
+            raise RuntimeError('pass')
+
+        api_client.request = functools.partial(request, 'http://path-server-test.petstore.local/v2/pet')
+        try:
+            api.add_pet({'name': 'pet', 'photo_urls': []})
+        except RuntimeError as e:
+            assert "pass" == str(e)
```
spacether, line 62: Possible improvement:
This works and a more standard python way to do it is to use
[assert_called_with](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with)

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
```diff
@@ -44,6 +46,27 @@ def test_configuration(self):
...
+
+        api_client.request = functools.partial(request, 'https://localhost:8080/v1/pet/123456789')
+        try:
+            api.delete_pet(123456789)
+        except RuntimeError as e:
+            assert "pass" == str(e)
```
spacether, line 68: Please also add tests of server_operation_index and server_variables.
That code is added in this PR and does not yet have any test coverage.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: I am trying to keep the usage similar with other languages. I don't see this as an issue. I would argue that configuration should be a mapping itself.

jirikuncar: I will add the docstrings.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: Technically it's a base path since `host` would mean only host part of a URL.

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar: This method should be private in first place. I might create a private version of this method instead.

modules/openapi-generator/src/main/resources/python/configuration.mustache
```diff
@@ -595,3 +609,17 @@ conf = {{{packageName}}}.Configuration(
             url = url.replace("{" + variable_name + "}", used_value)
 
         return url
+
+    @property
+    def host(self):
+        """Return generated host."""
+        if self.server_index is None:
+            return self._base_path
+
```
jirikuncar, line 618:
```suggestion
```

modules/openapi-generator/src/main/resources/python/python-experimental/api.mustache
```diff
@@ -105,7 +105,7 @@ class {{classname}}(object):
                     Default is True.
                 _host_index (int): specifies the index of the server
```
jirikuncar, line 106:
```suggestion
                _host_index (int/None): specifies the index of the server
```

modules/openapi-generator/src/main/resources/python/configuration.mustache
jirikuncar, line 578:
```suggestion
        :param servers: an array of host settings or None
```

jirikuncar: @spacether I am going to use the same documentation as it is used for return value of `get_host_settings`.

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
jirikuncar: I have created a separate package with a detailed testsuite.

modules/openapi-generator/src/test/resources/3_0/features/dynamic-servers.yaml
```diff
@@ -0,0 +1,92 @@
+openapi: 3.0.0
+info:
+  description: This specification shows how to use dynamic servers.
```
jirikuncar, line 3: @jimschubert I have tried to create a minimal specification on which we can showcase the dynamic server configuration to avoid modifications of default "Petstore" example.

jirikuncar:
All tests are passing. cc @jimschubert @spacether

spacether:
The changes requested point that I brought up has not been addressed. Can you please address it?

jirikuncar:
@spacether which one?

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
jirikuncar: If we remove support for Python 2, then yes.

jirikuncar: It's still declared in https://github.com/OpenAPITools/openapi-generator/blob/294d164ab033df0f7c60f38885c9b57c1fb711c3/samples/openapi3/client/petstore/python-experimental/setup.py#L33. I would be happy to change it in the future for Python 3.6+ only.

jirikuncar:
The configuration object could be a mapping with defined schema. I am trying to keep the configuration similar between different clients Go/Java/Python/... I don't see why passing `server_info={"index": 1}` is better than `server_index=1` which is the common use-case or `server_variables={"region": "eu"}` for customers from EU.

spacether:
The unaddressed point is `Our configuration class init method is growing to accept too many arguments, please reduce the number of inputs that you use here to 2 or 1`. It is better because the init signature is growing to include too many parameters.
I am fine with many configuration_instance.property_name access to data inside the configuration, but our init method should not contain every parameter we need as a variable input.
If we do that, the method signature gets too large and much harder to understand.
- This is consistent with what the python style tool pylint suggests that we do through its `too many arguments` warning
- This is what we have done in the past with the signing_info which added one input rather than 6
- This is consistent with what the go generator does where only only servers input is added in the configuration Struct
Will it help if we discuss this? I can be available over Slack if you want to chat.

jirikuncar:
The problem with Python configuration object is that it acts both as Go configuration and context at the same time. We will have the same problem in the future.

spacether:
> The problem with Python configuration object is that it acts both as Go configuration and context at the same time. We will have the same problem in the future.

So in my mind that's not the problem. Here's my thought process. New users want to use the configuration instance. They go to read its init method, if they only need to use one feature, they should only have to pass in one argument. If I am a new user and I want to use one feature, and I have to read through 20+ variables to try to understand which specific input I should be passing in when those 20+ variables are inputs for 10 features, then we have made using this class unnecessarily complex and hard to understand. Many users will not be using the configurable servers feature. It is easier for them if they can skip over the 1 or 2 inputs because they are not relevant vs reading these 4 variables and thinking hey do I need this input for feature X.

jirikuncar:
@spacether it's more complex than you describe. It's not really a single feature. One could break it into:

1. Support for servers
1. Support for servers per endpoint
1. Support for templated servers (x2)

Based on usage in other libraries:

- Most users might change only a single variable (e.g. region);
- Some users might change the server index;
- In rare cases users want to change a configuration of a specific endpoint;

Nesting is not more user friendly! Moreover we have the same pattern in other languages and I want to make sure that it's consistent. All parameters have reasonable defaults so users do not have to specify them if they are not planning to use these features.

samples/openapi3/client/petstore/python-experimental/tests/test_api_validation.py
wing328: I think we can drop Python 2.x support in the upcoming 5.x release.

spacether:
Can you resolve the conflicts then I can merge it in?

PR #6598, opened 2020-06-09
[core] Templating: limit compilation to supported extensions and constrain target paths

jimschubert:
I refactored template management in #6357. @wing328 found that an issue I had with one of the rust files was because the file was being processed by Mustache where it hadn't been before. This PR updates to process only files with an ending supported by the targeted templating engine.

This also fixes some Sonar recommendations, one of which may be a breaking change; template and output directories now may not include `..` for directory traversals.

### PR checklist

- [x] Run the shell script `./bin/generate-samples.sh`to update all Petstore samples related to your fix. This is important, as CI jobs will verify _all_ generator outputs of your HEAD commit as it would merge with master. These must match the expectations made by your contribution. You may regenerate an individual generator by passing the relevant config(s) as an argument to the script, for example `./bin/generate-samples.sh bin/config/java*`.


modules/openapi-generator-core/src/main/java/org/openapitools/codegen/api/TemplatingEngineAdapter.java
```diff
@@ -65,9 +75,10 @@ String compileTemplate(TemplatingExecutor executor, Map<String, Object> bundle,
      * @param templateFile The original target filename
      * @return True if the template is available in the template search path, false if it can not be found
      */
+    @SuppressWarnings({"java:S2093"}) // ignore java:S2093 because we have double-assignment to the closeable
     default boolean templateExists(TemplatingExecutor generator, String templateFile) {
         return Arrays.stream(getFileExtensions()).anyMatch(ext -> {
-            int idx = templateFile.lastIndexOf(".");
+            int idx = templateFile.lastIndexOf('.');
```
jimschubert, line 81: lastIndexOf taking character is more performant

modules/openapi-generator/src/main/java/org/openapitools/codegen/TemplateManager.java
```diff
@@ -145,18 +158,32 @@ public Reader getTemplateReader(String name) {
...
+        if (this.engineAdapter.handlesFile(template)) {
+            // Only pass files with valid endings through template engine
+            String templateContent = this.engineAdapter.compileTemplate(this, data, template);
+            return writeToFile(target.getPath(), templateContent);
+        } else {
+            // Do a straight copy of the file if not listed as supported by the template engine.
+            InputStream is;
+            try {
+                // look up the file using the same template resolution logic the adapters would use.
+                String fullTemplatePath = getFullTemplateFile(template);
```
jimschubert, line 170: This fixes what I believe may have been a bug prior to this change… supporting files may not have resolved using the libraries path rules appropriately.

modules/openapi-generator/src/main/java/org/openapitools/codegen/TemplatingEngineLoader.java
```diff
@@ -37,7 +42,7 @@ public static TemplatingEngineAdapter byIdentifier(String id) {
             // Attempt to load skipping SPI
             return (TemplatingEngineAdapter) Class.forName(id).getDeclaredConstructor().newInstance();
         } catch (Exception e) {
-            throw new RuntimeException(String.format(Locale.ROOT, "Couldn't load template engine adapter %s. Available options: \n%s", id, sb.toString()), e);
+            throw new RuntimeException(String.format(Locale.ROOT, "Couldn't load template engine adapter %s. Available options: %n%s", id, sb.toString()), e);
```
jimschubert, line 45: `%n` uses system-specific newline rather than newline only.

modules/openapi-generator/src/main/resources/rust/request.rs
```diff
@@ -113,7 +113,7 @@ impl Request {
         let mut path = self.path;
         for (k, v) in self.path_params {
             // replace {id} with the value of the id path param
-            {{=<% %>=}}path = path.replace(&format!("{{{}}}", k), &v);<%={{ }}=%>
```
jimschubert, line 116: This was added in the template management refactor because the file was being processed by Mustache engine (note file is `request.rs` not `request.mustache` and I missed this during the refactor)

jimschubert:
Notice that php-lumen changes will be considered a breaking change in 5.0.0. This is the only generator which had mustache templates without mustache file extension.

ybelenko:
@jimschubert Do they still need `.mustache` extension if they are source files from `Lumen` framework which shouldn't be modified?

jimschubert:
@ybelenko no, anything without mustache extension will be copied as-is. Everything in my change will be processed through the mustache engine. Most of these files seemed to only have a mustache include for a header partial.

ybelenko:
@jimschubert So, if I want to copy these files "as is" I should add header part inside and don't use mustache syntax, right?

ybelenko (approved):
I don't know why 5e01c65 commit contains `samples/openapi3/client/petstore/java/jersey2-java8-special-characters/gradle/wrapper/gradle-wrapper.jar` file, beside that everything related to PHP Lumen looks good to me.

PR #6634, opened 2020-06-12
[java][jersey2] Differentiate request with no body vs request that contains the null value

sebastien-rosset:
Two scenarios need to be differentiated:
1. HTTP requests may have an empty body, e.g. HTTP GET requests.
1. HTTP requests may also have a body with the 'null' value.

We need to be able to differentiate these two cases when serializing the body. In particular, for HTTP GET request which have an empty body, the input object was set to null, and it was serialized as the 'null' value instead of the empty string.

This leads to incorrect message digest being calculated.

### PR checklist

### DIFF
--- a/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/ApiClient.java
+++ b/samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/ApiClient.java
@@ -25,6 +25,13 @@ import java.io.IOException;
 import java.io.InputStream;
 
 import java.net.URI;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import java.security.cert.X509Certificate;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import org.glassfish.jersey.logging.LoggingFeature;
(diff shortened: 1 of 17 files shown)


samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/ApiClient.java
```diff
@@ -787,7 +794,7 @@ public String escapeString(String str) {
    * @return String
    * @throws ApiException API exception
    */
-  public String serializeToString(Object obj, Map<String, Object> formParams, String contentType) throws ApiException {
+  public String serializeToString(BodyHolder obj, Map<String, Object> formParams, String contentType) throws ApiException {
```
sebastien-rosset, line 797: @wing328 , I'm a bit confused why `updateParamsForAuth` invokes serializeToString, but the request body is serialized using a different function. That means an authorization scheme that processes the message body may provide incorrect authorization data.
For example, the message digest may be different.

sebastien-rosset:

samples/client/petstore/java/jersey2-java8/src/main/java/org/openapitools/client/ApiClient.java
wing328: I agree ideally they should be using the same function to come up with the request body.

sebastien-rosset: thanks. I propose that we tackle this in a follow-up PR, what do you think? I see some discrepancies between serializeToString and serialize.

Issue #6646, opened 2020-06-12
[Java][jersey2] The jersey2 library does not support additional undeclared properties


sebastien-rosset:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

The jersey2 library fails deserialization when the input data contains additional (undeclared) properties. For example, suppose the input data is the following JSON document:

```json
{
      "Name": "Bob",
      "CreateTime": "2020-04-13T21:57:56.552Z",
      "UserUniqueIdentifier": "2aff0200-3d2e-4ba3-823d-c9ea8877556d"
}
```

The schema explicitly declares `Name` and `CreateTime` but not `UserUniqueIdentifier`:

When the jersey2 library attempts to deserialize the payload, an error occurs:
```
INFO: Input data does not match schema 'XYZ'
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException:
Unrecognized field "UserUniqueIdentifier" (class XYZ),
not marked as ignorable (31 known properties: ""CreateTime", "Name", ...])
 at [Source: UNKNOWN; line: -1, column: -1]
(through reference chain: XYZ["Results"]->java.util.ArrayList[0]->XYZ["UserUniqueIdentifier"])
```

##### openapi-generator version

master June 12th

##### OpenAPI declaration file content or url

```yaml
User:
  type: object
  properties:
    Name:
      type: string
    CreateTime:
      type: string
  additionalProperties: true
```

PR #6647, opened 2020-06-12
[java][jersey2] Fix processing of additional, undeclared properties

sebastien-rosset:
This is a fix for #6646.

1. Add more unit tests (in JSONComposedSchemaTest.java) to exercise various use cases for `additionalProperties`, such as:
   1. within composed schema
   1. non composed schemas
   1. additionalProperties with boolean value
   1. additionalProperties with schema
1. Fix processing of `additionalProperties`

### PR checklist

### DIFF
--- a/bin/configs/java-jersey2-8.yaml
+++ b/bin/configs/java-jersey2-8.yaml
@@ -8,3 +8,4 @@ additionalProperties:
   serverPort: "8082"
   dateLibrary: java8
   useOneOfDiscriminatorLookup: true
+  disallowAdditionalPropertiesIfNotPresent: false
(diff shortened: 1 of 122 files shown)


sebastien-rosset:

bin/configs/java-jersey2-8.yaml
```diff
@@ -8,3 +8,4 @@ additionalProperties:
   serverPort: "8082"
   dateLibrary: java8
   useOneOfDiscriminatorLookup: true
+  disallowAdditionalPropertiesIfNotPresent: false
```
sebastien-rosset, line 11: @wing328 , I think at some point we should set `disallowAdditionalPropertiesIfNotPresent` to false by default, maybe starting with jersey2.
Historically, the behavior has been to disallow additional properties by default, but that's not what the spec states.

wing328: I think we should let the user choose at this point instead of forcing the behavior per generator/library.

We'll introduce this option to the community as part of the v5.0.0 official release and consider changing the default value later if more users are familiar with the change.

sebastien-rosset: ok, sure. This is how it is currently implemented (i.e. non-compliant). I have set disallowAdditionalPropertiesIfNotPresent to false for the example only.

Issue #6648, opened 2020-06-13
[BUG][Javascript] Constructors don't handle required fields with default values well


tray2100:
##### Description

Currently the generator creates constructors where required fields are set to the parameters in the constructor. This is fine except when the parameters are null and there is a default value specified in the spec. 

##### openapi-generator version

4.3.1

##### Suggest a fix

This can be fixed by just having the constructor assign either the incoming value or the default value if the incoming value is undefined. I'll submit a PR for this.

PR #6649, opened 2020-06-13
[Javascript] Fix for constructors not handling required fields with default values well #6648

tray2100:
This PR is for issue #6648. I verified that the output was as expected where for a spec that contains a required field with a default value will now look like

```javascript
    _this['field'] = field || 'defaultValue';
```

instead of just

```javascript
    _this['field'] = field;
```

where it forces the required field to be undefined even though we have a default value. I also noticed that the samples were out of date and there wasn't a generate script for ES5 so I've added those here. I can squash the commits if needed before merging.

### PR checklist

cc: @CodeNinjai @frol @cliffano


wing328:
@tray2100 thanks for the PR. If a field is required, there shouldn't be any default value since the value must be provided.

In other words, the default value is for the optional field, right?

tray2100:
> @tray2100 thanks for the PR. If a field is required, there shouldn't be any default value since the value must be provided.
> 
> In other words, the default value is for the optional field, right?

Hey @wing328

I probably should have shared more of my use case. I've defined a oneOf entity and made the discriminator required in the schemas referenced in it. It feels super awkward right now when using the `constructFromObject` mechanism because you'll have to manually add in the discriminator after the object is constructed - even if you used the entity created by the generator in the first place. It's even more awkward if there is a chain of `constructFromObject` calls and the entity with the discriminator is a few levels down the object graph. 

Setting the field to the default value if nothing is specified solves this problem in a minimalist way. 

This is essentially the spec:

```
 OneOfDogCat:
      oneOf:
      - $ref: "#/components/schemas/Dog"
      - $ref: "#/components/schemas/Cat"
      type: "object"
      discriminator:
        propertyName: "entityType"
 Dog:
      type: "object"
      properties:
        entityType:
          default: "dog"
          type: "string"
        name:
          type: "string"
      required:
      - "entityType"
 Cat:
      type: "object"
      properties:
        entityType:
          default: "cat"
          type: "string"
        name:
          type: "string"
      required:
      - "entityType"
```

wing328:
@tray2100 as discussed we can use `{{#oneOf}} .. {{/oneOf}}` to further enhance this fix for oneOf models only. Let's see if the community can help out with this.

PR #6672, opened 2020-06-15
[go-experimental] Fix usage of discriminator for oneOf deserialization

bkabrda:
This PR partially fixes the new style usage of discriminator for better deserialization of `oneOf` composed schemas. Changes:

* Enables usage of string schemas referenced by `$ref` as type of the discriminator property (see `OneType` in example below).
* Removes some unnecessary (AFAICS) logic for unmarshalling - we don't need to check `if string(json{{{modelName}}}) == "{}"`, as we know that `json{{{modelName}}}` contains at least the discriminator property.

There's one more issue that I'm not sure to address. Consider this example:

```yaml
openapi: 3.0.0
info:
  title: test
  version: 1.0.0
paths:
  /test:
    patch:
      responses:
        '200':
          description: all good
      requestBody:
        description: request
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Something'
components:
  schemas:
    Something:
      type: object
      oneOf:
        - $ref: '#/components/schemas/One'
        - $ref: '#/components/schemas/Two'
      discriminator:
         propertyName: foobar
         mapping:
           fooone: '#/components/schemas/One'
           footwo: '#/components/schemas/Two'

    OneType:
      type: string
      enum:
        - fooone
    One:
      type: object
      required: [foobar]
      properties:
        foobar:
          $ref: '#/components/schemas/OneType'
        one:
          type: string

    Two:
      type: object
      required: [foobar]
      properties:
        foobar:
          type: string
        two:
          type: string
```

When I run `openapi-generator generate -i min.yaml -g go-experimental -o go --additional-properties useOneOfDiscriminatorLookup=true`

I get unmarshaling code like this (even without my change for working with the referenced schema):

```go
// Unmarshal JSON data into one of the pointers in the struct
func (dst *Something) UnmarshalJSON(data []byte) error {
	var err error
	// use discriminator value to speed up the lookup
	var jsonDict map[string]interface{}
	err = json.Unmarshal(data, &jsonDict)
	if err != nil {
		return fmt.Errorf("Failed to unmarshal JSON into map for the discrimintor lookup.")
	}

	// check if the discriminator value is 'One'
	if jsonDict["foobar"] == "One" {
		// try to unmarshal JSON data into One
		err = json.Unmarshal(data, &dst.One)
		if err == nil {
			return nil // data stored in dst.One, return on the first match
		} else {
			dst.One = nil
			return fmt.Errorf("Failed to unmarshal Something as One: %s", err.Error())
		}
	}

	// check if the discriminator value is 'Two'
        // <MORE-CODE>

	// check if the discriminator value is 'fooone'
	if jsonDict["foobar"] == "fooone" {
		// try to unmarshal JSON data into One
		err = json.Unmarshal(data, &dst.One)
		if err == nil {
			return nil // data stored in dst.One, return on the first match
		} else {
			dst.One = nil
			return fmt.Errorf("Failed to unmarshal Something as One: %s", err.Error())
		}
	}

	// check if the discriminator value is 'footwo'
        // <MORE-CODE>

	return nil
}
```

As you can see, the deserialization logic is generated twice, once for `One` and once for `fooone`. The `if jsonDict["foobar"] == "One" {` line is obviously wrong. I tracked it down to being added at [this line](https://github.com/OpenAPITools/openapi-generator/blob/ff68128c15363ece22d7b6376952eb4eabc1f0bd/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L2862). I would like to fix that, but I honestly don't understand the whole purpose of the `getOneOfAnyOfDescendants` function which returns these results. AFAICS it's redundant and isn't necessary at all, but maybe I'm missing some context. Any help would be appreciated.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2539,7 +2539,7 @@ public class DefaultCodegen implements CodegenConfig {
     private CodegenProperty discriminatorFound(String composedSchemaName, Schema sc, String discPropName, OpenAPI openAPI) {
         Schema refSchema = ModelUtils.getReferencedSchema(openAPI, sc);
         if (refSchema.getProperties() != null && refSchema.getProperties().get(discPropName) != null) {
-            Schema discSchema = (Schema) refSchema.getProperties().get(discPropName);
+            Schema discSchema = ModelUtils.getReferencedSchema(openAPI, (Schema) refSchema.getProperties().get(discPropName));
             CodegenProperty cp = new CodegenProperty();
             if (ModelUtils.isStringSchema(discSchema)) {
                 cp.isString = true;
--- a/modules/openapi-generator/src/main/resources/go-experimental/model_oneof.mustache
+++ b/modules/openapi-generator/src/main/resources/go-experimental/model_oneof.mustache
@@ -16,7 +16,6 @@ func {{{.}}}As{{classname}}(v *{{{.}}}) {{classname}} {
 // Unmarshal JSON data into one of the pointers in the struct
 func (dst *{{classname}}) UnmarshalJSON(data []byte) error {
 	var err error
-	match := 0
 	{{#isNullable}}
 	// this object is nullable so check if the payload is null or empty string
 	if string(data) == "" || string(data) == "{}" {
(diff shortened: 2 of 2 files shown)


wing328:
> Removes some unnecessary (AFAICS) logic for unmarshalling - we don't need to check if string(json{{{modelName}}}) == "{}", as we know that json{{{modelName}}} contains at least the discriminator property.

That's for nullable oneOf schema.

wing328:
@bkabrda for the issue you mentioned above, can you please open an issue to track it separately if you've not already done so?

PR #6673, opened 2020-06-15
[Rust][Client] Multiple returns becomes optional (fixes #6650).

### LINKED ISSUE #6650: [REQ][Rust Client] New Success types lead to exessive code when using the API

Before #6481 has been merged this morning, my test code for using the 5.0.0 SNAPSHOT with `supportAsync: true` and `useSingleRequestParameter: true` looked something like this:
```rust
    let servers = servers_api::list_servers(&configuration, params)
        .await
        .map_err(|err| format!("API call to list_servers failed: {:?}", err))?
        .servers;
```
For the same behaviour, I now have to write:
```rust
    let success = servers_api::list_servers(&configuration, params)
        .await
        .map_err(|err| format!("API call to list_servers failed: {:?}", err))?
        .entity
        .ok_or("API call to list_servers failed: Response could not be parsed.".to_string())?;
    let servers = match success {
        servers_api::ListServersSuccess::Status200(response) => Ok(response.servers),
        _ => Err("API call to list_servers failed: Unexpected success status code.".to_string()),
    }?;
```
In my oppinion, the code has become unnecessarily complex here.

I am not sure if I overlooked a way for a simpler approach, but it would be nice to only deal with one layer of error handling when using the API.

As this is only related to template code, I can of course change this back to the previous behaviour using a customized templates, but I would like to stick to the default templates as far as possible.

## Additional context

I am now using SNAPSHOT version 5.0.0-20200613.071037-272.

Thanks for looking into this! If I can give a hand finding a solution, I would be glad to do so.

bcourtine:
By default, rust client (using Reqwest) returns a simple return type. Handling multiple returns (several 2xx codes is optional).

### PR checklist

cc @frol (2017/07) @farcaller (2017/08) @bjgill (2017/12) @richardwhiuk (2019/07) @paladinzh (2020/05) @wing328

### DIFF
--- a/samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
+++ b/samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
@@ -31,31 +31,21 @@ impl DefaultApiClient {
 }
 
 
-/// struct for typed successes of method `fileresponsetest`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum FileresponsetestSuccess {
-    Status200(std::path::PathBuf),
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
 /// struct for typed errors of method `fileresponsetest`
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum FileresponsetestError {
-    DefaultResponse(std::path::PathBuf),
     UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
 
 pub trait DefaultApi {
-    fn fileresponsetest(&self, ) -> Result<ResponseContent<FileresponsetestSuccess>, Error<FileresponsetestError>>;
+    fn fileresponsetest(&self, ) -> Result<std::path::PathBuf, Error<FileresponsetestError>>;
 }
 
 impl DefaultApi for DefaultApiClient {
-    fn fileresponsetest(&self, ) -> Result<ResponseContent<FileresponsetestSuccess>, Error<FileresponsetestError>> {
+    fn fileresponsetest(&self, ) -> Result<std::path::PathBuf, Error<FileresponsetestError>> {
         let configuration: &configuration::Configuration = self.configuration.borrow();
         let client = &configuration.client;
 
@@ -73,9 +63,7 @@ impl DefaultApi for DefaultApiClient {
         let content = resp.text()?;
 
         if status.is_success() {
-            let entity: Option<FileresponsetestSuccess> = serde_json::from_str(&content).ok();
-            let result = ResponseContent { status, content, entity };
-            Ok(result)
+            serde_json::from_str(&content).map_err(Error::from)
         } else {
             let entity: Option<FileresponsetestError> = serde_json::from_str(&content).ok();
             let error = ResponseContent { status, content, entity };
--- a/samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
+++ b/samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
@@ -35,7 +35,6 @@ impl DefaultApiClient {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum FileresponsetestError {
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
--- a/samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
+++ b/samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
@@ -36,7 +36,6 @@ impl PetApiClient {
 #[serde(untagged)]
 pub enum AddPetError {
     Status405(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -45,7 +44,6 @@ pub enum AddPetError {
 #[serde(untagged)]
 pub enum DeletePetError {
     Status400(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -54,7 +52,6 @@ pub enum DeletePetError {
 #[serde(untagged)]
 pub enum FindPetsByStatusError {
     Status400(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -63,7 +60,6 @@ pub enum FindPetsByStatusError {
 #[serde(untagged)]
 pub enum FindPetsByTagsError {
     Status400(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -73,7 +69,6 @@ pub enum FindPetsByTagsError {
 pub enum GetPetByIdError {
     Status400(),
     Status404(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -84,7 +79,6 @@ pub enum UpdatePetError {
     Status400(),
     Status404(),
     Status405(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -93,7 +87,6 @@ pub enum UpdatePetError {
 #[serde(untagged)]
 pub enum UpdatePetWithFormError {
     Status405(),
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
@@ -101,7 +94,6 @@ pub enum UpdatePetWithFormError {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum UploadFileError {
-    UnknownList(Vec<serde_json::Value>),
     UnknownValue(serde_json::Value),
 }
 
--- a/samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
+++ b/samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
@@ -31,74 +31,6 @@ impl PetApiClient {
 }
 
 
-/// struct for typed successes of method `add_pet`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum AddPetSuccess {
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
-/// struct for typed successes of method `delete_pet`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum DeletePetSuccess {
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
-/// struct for typed successes of method `find_pets_by_status`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum FindPetsByStatusSuccess {
-    Status200(Vec<crate::models::Pet>),
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
-/// struct for typed successes of method `find_pets_by_tags`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum FindPetsByTagsSuccess {
-    Status200(Vec<crate::models::Pet>),
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
-/// struct for typed successes of method `get_pet_by_id`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum GetPetByIdSuccess {
-    Status200(crate::models::Pet),
-    UnknownList(Vec<serde_json::Value>),
-    UnknownValue(serde_json::Value),
-}
-
-/// struct for typed successes of method `update_pet`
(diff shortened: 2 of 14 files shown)


samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
```diff
@@ -452,9 +370,7 @@ impl PetApi for PetApiClient {
         let content = resp.text()?;
 
         if status.is_success() {
-            let entity: Option<UploadFileSuccess> = serde_json::from_str(&content).ok();
-            let result = ResponseContent { status, content, entity };
-            Ok(result)
+            serde_json::from_str(&content).map_err(Error::from)
```
seunlanlege, line 373: missing `Ok()` also i tried testing locally, didn't work for me:

`java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g rust -i slack_web_modified.json -o slack --skip-validate-spec --remove-operation-id-prefix --library reqwest --additional-properties useSingleRequestParameter=true,supportAsync=true,supportMultipleReturns=true`

bcourtine: Hi,

I generated a slack client from official Github JSON with the same parameters as yours:

```
java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g rust -i https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json -o /tmp/slack --skip-validate-spec --remove-operation-id-prefix --library reqwest --additional-properties useSingleRequestParameter=true,supportAsync=true,supportMultipleReturns=true
```

The client compiles and works fine: can you provide some information to reproduce the problem?

PS: in issue #6650, @HenningHolmDE tested the commit with and without the new param and didn't notice a problem.

bcourtine: In the code `serde_json::from_str(&content).map_err(Error::from)` you quoted, `Ok()` is not missing, since this line produces the expected return type `Result<Entity, Error>`.

seunlanlege:
```
supportMultipleReturns=false
```

this was the only thing that worked for me

bcourtine: Do you have an example OAS to reproduce your problem?

I have a compiling client (with Rust 1.44.0 and both values for `supportMultipleReturns`).

seunlanlege:
I think the last thing here is to remove the unknown value/list variants from the error type. The api specification defines all possible response schemas. So those variants are redundant.

HenningHolmDE:
> 
> 
> I think the last thing here is to remove the unknown value/list variants from the error type. The api specification defines all possible response schemas. So those variants are redundant.

Wouldn't it be useful to be able to access the data if the response type is unexpected? (This is of course an error, but we are talking error handling here.)
As for the `UnknownList`, this could already be covered by `UnknownValue` as `serde_json::value::Value` can represent this as `Array(Vec<Value>)`?

(Just my two cents as I was mentioned erlier and got the notfication for the comment.)

samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs
wing328: @bcourtine I think @seunlanlege is using https://gist.githubusercontent.com/wing328/61225b4ad2d838447340a70494e09333/raw/b42fa7cf01df2a9269b910d19393a96081a246ff/slack_web_modified.json

HenningHolmDE: Hm, using that spec file also works for me (rust stable as well as nightly):
```
java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g rust -i https://gist.githubusercontent.com/wing328/61225b4ad2d838447340a70494e09333/raw/b42fa7cf01df2a9269b910d19393a96081a246ff/slack_web_modified.json -o slack --skip-validate-spec --remove-operation-id-prefix --library reqwest --additional-properties useSingleRequestParameter=true,supportAsync=true,supportMultipleReturns=true
cd slack
cargo check
> Finished dev [unoptimized + debuginfo] target(s) in 0.14s
```
Or is the problem related to function instead of compilation?

wing328:
> Wouldn't it be useful to be able to access the data if the response type is unexpected?

That's what we're doing at the moment in other clients to include the response payload in the error response object (e.g. ApiException class) so the developers can find out more clues on why the error occurs from the response payload.

seunlanlege:
> Wouldn't it be useful to be able to access the data if the response type is unexpected?

yeah, I understand your reasoning. 

@bcourtine let's limit it to just `Unknown(serde_json::Value)` for now then like you pointed out.

samples/client/petstore/rust/reqwest/fileResponseTest/src/apis/default_api.rs
```diff
@@ -73,9 +63,7 @@ impl DefaultApi for DefaultApiClient {
         let content = resp.text()?;
 
         if status.is_success() {
-            let entity: Option<FileresponsetestSuccess> = serde_json::from_str(&content).ok();
-            let result = ResponseContent { status, content, entity };
-            Ok(result)
+            serde_json::from_str(&content).map_err(Error::from)
         } else {
             let entity: Option<FileresponsetestError> = serde_json::from_str(&content).ok();
```
seunlanlege, line 68: in light of [this](https://github.com/OpenAPITools/openapi-generator/pull/6673#issuecomment-645357769), I don't think it's possible for `serde_json::from_str()` to fail because you've already covered `serde_json::Value`

`ResponseContent` should become:

```rust
struct ResponseContent {
    entity: T,
    status: StatusCode,
}
```

HenningHolmDE: As you pointed out, the call to `serde_json::from_str()` won't fail due to unexpected Types as we have covered `serde_json::Value` - but it might fail when `content` does not contain valid JSON.

seunlanlege: hmm, then we should use the `response.json()` method.

HenningHolmDE: I'm afraid this will not help, as in case of invalid JSON [`reqwest::Response.json()`](https://docs.rs/reqwest/0.10.6/reqwest/struct.Response.html#method.json) will also fail:
> `pub async fn json<T: DeserializeOwned>(self) -> Result<T>`
> Try to deserialize the response body as JSON.
> [...]
> Errors
> This method fails whenever the response body is not in JSON format or it cannot be properly deserialized to target type `T`. For more details please see `serde_json::from_reader`.

seunlanlege: yes but then we already have an error for reqwest defined.

```rust
#[derive(Debug)]
pub enum Error<T> {
    Reqwest(reqwest::Error),
    Serde(serde_json::Error),
    Io(std::io::Error),
    ResponseError(ResponseContent<T>),
}
```

so code becomes

```rust
let status = resp.status();
let content = resp.text().await?;

if status.is_success() {
    // would recommend using derive_more::From for the error type
    Ok(response.json().await.map_err(Error::Reqwest)?)
} else {
        let entity: CloseError = response.json().await.map_err(Error::Reqwest)?;
        let error = ResponseContent { status, entity };
        Err(Error::ResponseError(error))
}
```

HenningHolmDE: Good idea! That way we don't have to deal with `serde_json::error::Error` explicitily. As a downside, the raw content would no longer be available in the result. There was probably a reason for adding this in the first place, maybe @bcourtine can shed some light on this.

As for your remark on `derive_more::From`:
`impl <T> From<reqwest::Error> for Error<T>` is already present in `api_mod.mustache`. The `map_err()` should therefore not be necessary. Or do I overlook something here?

seunlanlege: > As a downside, the raw content would no longer be available in the result.

since the response content is malformed json, the only thing useful to do with it is log to stdout/err or how would you try to handle it?

HenningHolmDE: > 
> 
> > As a downside, the raw content would no longer be available in the result.
> 
> since the response content is malformed JSON, the only thing useful to do with it is log to stdout/err or how would you try to handle it?

I meant in case of success: The current implementation always provides the raw content through `ResponseContent.content` (which requires using `response.text()` and `serde_json::from_str()`). Your suggestion would remove this. Sorry for the confusion!

seunlanlege: hmm, I'm not sure how this happens because the specification defines the response content-type.

```json
"produces": [
    "application/json"
]
```

and so if the success content isn't valid JSON shouldn't that be an error?

HenningHolmDE: My remark about dropping the raw content wasn't related to error handling - sorry for being too vague here. I was referring to your suggestion of changing the return value from
```rust
let status = resp.status();
let content = resp.text().await?;
if status.is_success() {
    let entity: Option<UsersSetRegularSuccess> = serde_json::from_str(&content).ok();
    let result = ResponseContent { status, content, entity };
    Ok(result)
} ...
```
to this:
```rust
let status = resp.status();
if status.is_success() {
   Ok(resp.json().await?) // .map_err(Error::Reqwest) should not be necessary
} ...
```
With this, `status` and `content` will no longer be provided to the API user in case of success. I do not know why this might be necessary or useful - however, it is provided by the current version of the code for some reason.

seunlanlege: > My remark about dropping the raw content wasn't related to error handling - sorry for being too vague here. I was referring to your suggestion of changing the return value from
> ```rust
> let status = resp.status();
> let content = resp.text().await?;
> if status.is_success() {
>     let entity: Option<UsersSetRegularSuccess> = serde_json::from_str(&content).ok();
>     let result = ResponseContent { status, content, entity };
>     Ok(result)
> } ...
> ```
> to this:
> ```rust
> let status = resp.status();
> if status.is_success() {
>    Ok(resp.json().await?) // .map_err(Error::Reqwest) should not be necessary
> } ...
> ```
> With this, `status` and `content` will no longer be provided to the API user in case of success. I do not know why this might be necessary or useful - however, it is provided by the current version of the code for some reason.

This pr removes use of `ResponseContent` in the success case, actually.

What I'm advocating for is removing the content field in `ResponseContent`

bcourtine: > This pr removes use of ResponseContent in the success case, actually.

ResponseContent is not dropped: it becomes optional. If you use `supportMultipleResponses=true`, you will still get a `ResponseContent` in success case.

I totally dropped ResponseContent with `supportMultipleResponses=false` because the PR aimed at providing the simplest possible signature, like in 4.x.x version.

`content` field exists mostly for error case: it allows to debug an error (malformed JSON which cannot be converted to entity). I let this field in success case to avoid multiplicating structs.

First, I will simplify enum values dropping `UnknowList` (included in `UnknownValue`).

For the rest, I am a little lost in this debate… do you have a consensus on the best possible solution (for both `supportMultipleResponses` values)?

seunlanlege: hopefully, @bcourtine's comment clarifies things @HenningHolmDE

to clarify further:

> What I'm advocating for is removing the `content` field in `ResponseContent`

because:

> since the response content is malformed JSON, the only thing useful to do with it is: log to stdout/err

wing328: > since the response content is malformed JSON, the only thing useful to do with it is: log to stdout/err

It may or may not be malformed. We've seen issues in which a different type of object was returned (e.g. expecting array of Dog but get a single Dog, expecting a Cat but get a Dog due to invalid spec, expecting JSON but got XML as XML is the default when `accept` is not specified in the request header in that particular case) so my recommendation is still to include it in the Error response object to make developer's life easier in troubleshooting what's wrong.

seunlanlege: I was under the impression that logging the reqwest Error had the response text, but after testing i just confirmed it doesn't so this is fine.

> include it in the Error response object to make developer's life easier in troubleshooting what's wrong.

I think that's the last thing, we can finally get this merged? :confetti_ball: :partying_face:

HenningHolmDE: > I think that's the last thing, we can finally get this merged? :confetti_ball: :partying_face:

No objections from my side. :+1:

wing328: 👍 👍 👍

wing328:
```
 |packageVersion|Rust package version.| |1.0.0|
 |supportAsync|If set, generate async function call instead. This option is for 'reqwest' library only| |true|
+|supportMultipleResponses|If set, return type wraps an enum of all possible 2xx schemas. This option is for 'reqwest' library only| |false|
 |useSingleRequestParameter|Setting this property to true will generate functions with a single argument containing all API endpoint parameters instead of one argument per parameter.| |false|
 
 ## IMPORT MAPPING
Perform git status
On branch pull/6673

	modified:   docs/generators/rust.md
```

@bcourtine Please update the doc as well when you've time.

seunlanlege:
Any chance we can get this merged today?

wing328:
Just pushed https://github.com/OpenAPITools/openapi-generator/pull/6673/commits/ae1a050273a156f017fe460d2a3133b3735cb9fb to resolve the CI failure. Will merge if all tests pass.

PR #6674, opened 2020-06-15
[Python][test] oneOf with overlapping types

jirikuncar:
**tl;dr** Shows a serialization problem of `oneOf` models in `python-experimental` with overlapping types.

### Minimal specification

Two models with overlapping field definition of named enum type.

```yaml
openapi: 3.0.1
info:
  title: fruity
  version: 0.0.1
paths:
  /:
    get:
      responses:
        "200":
          description: desc
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/fruit"
components:
  schemas:
    fruit:
      title: fruit
      oneOf:
        - $ref: "#/components/schemas/apple"
        - $ref: "#/components/schemas/banana"
    appleColor:
      type: string
      enum:
        - yellow
        - green
        - red
    apple:
      title: apple
      type: object
      properties:
        kind:
          type: string
        color:
          $ref: '#/components/schemas/appleColor'
    bananaColor:
      type: string
      enum:
        - yellow
        - green
    banana:
      title: banana
      type: object
      properties:
        count:
          type: number
        color:
          $ref: '#/components/schemas/bananaColor'
```

### Problem

The generated model contains all fields and the overlapping field has wrong type.

```python
    @cached_property
    def openapi_types():
        """..."""
        return {
            'kind': (str,),  # noqa: E501
            'color': (banana_color.BananaColor,),  # noqa: E501
            'count': (float,),  # noqa: E501
        }

     # ...

     attribute_map = {
        'kind': 'kind',  # noqa: E501
        'color': 'color',  # noqa: E501
        'count': 'count',  # noqa: E501
    }
```

This prevents deserialization of [`{"kind": "sweet", "color": "red"}` response](https://github.com/OpenAPITools/openapi-generator/pull/6674/files#diff-4b1b7af6f0b61eca441265dfa63b8962R40).

### How do I get expected behavior?

Simply remove definition of [`openapi_types`](https://github.com/OpenAPITools/openapi-generator/pull/6674/files#r440278942) and [`attribute_map`](https://github.com/OpenAPITools/openapi-generator/pull/6674/files#r440279315) and let the `oneOf` models do the job.

### PR checklist

cc @sebastien-rosset


modules/openapi-generator/src/test/resources/3_0/oneOfOverlapingTypes.yaml
```diff
@@ -0,0 +1,49 @@
...
+      oneOf:
+        - $ref: "#/components/schemas/apple"
+        - $ref: "#/components/schemas/banana"
+    appleColor:
+      type: string
+      enum:
+        - yellow
+        - green
+        - red
+
```
jirikuncar, line 28:
```suggestion
```

samples/openapi3/client/features/python-experimental/one_of/one_of/model/fruit.py
```diff
@@ -0,0 +1,241 @@
...
+
+        Returns
+            openapi_types (dict): The key is attribute name
+                and the value is attribute type.
+        """
+        return {
+            'kind': (str,),  # noqa: E501
+            'color': (banana_color.BananaColor,),  # noqa: E501
+            'count': (float,),  # noqa: E501
+        }
```
jirikuncar, line 101: This should be empty as all fields are defined in `oneOf` models.

```suggestion
        return {}
```

samples/openapi3/client/features/python-experimental/one_of/one_of/model/fruit.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @cached_property
+    def discriminator():
+        return None
+
+    attribute_map = {
+        'kind': 'kind',  # noqa: E501
+        'color': 'color',  # noqa: E501
+        'count': 'count',  # noqa: E501
+    }
```
jirikuncar, line 111: ❗ 
```suggestion
    attribute_map = {}
```

samples/openapi3/client/features/python-experimental/one_of/one_of/model/fruit.py
```diff
@@ -0,0 +1,241 @@
...
+try:
+    from one_of.model import banana
+except ImportError:
+    banana = sys.modules[
+        'one_of.model.banana']
+try:
+    from one_of.model import banana_color
+except ImportError:
+    banana_color = sys.modules[
+        'one_of.model.banana_color']
```
jirikuncar, line 50: ❗ 
```suggestion
```

samples/openapi3/client/features/one_of/python-experimental/one_of/model/fruit.py
```diff
@@ -0,0 +1,241 @@
...
+
+        Returns
+            openapi_types (dict): The key is attribute name
+                and the value is attribute type.
+        """
+        return {
+            'kind': (str,),  # noqa: E501
+            'color': (banana_color.BananaColor,),  # noqa: E501
+            'count': (float,),  # noqa: E501
+        }
```
jirikuncar, line 101:
```suggestion
        return {}
```

samples/openapi3/client/features/one_of/python-experimental/one_of/model/fruit.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @cached_property
+    def discriminator():
+        return None
+
+    attribute_map = {
+        'kind': 'kind',  # noqa: E501
+        'color': 'color',  # noqa: E501
+        'count': 'count',  # noqa: E501
+    }
```
jirikuncar, line 111:
```suggestion
    attribute_map = {}
```

sebastien-rosset:
Thank you for the explanation. This is similar to a problem that was also identified in Java jersey2. AFAIK, in case of composed schemas, the following use cases are not currently handled:
1. Same property name appears in two different schemas with different types. 
1. Same property name, different constraints, e.g. maxItems, minItems, maximum, minimum, pattern...

The problem exists for `anyOf`, `oneOf` and `allOf`, not just `oneOf`. A schema can have a combination of `anyOf`, `oneOf` and `allOf`, and conceivable the same property could appear multiple times across all three.

When two properties with the same name are specified, afaik the "last one wins", i.e. the last referenced schema in the OAS document wins.

 the "last" property wins

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
```diff
@@ -15,6 +15,12 @@ class {{unescapedDescription}}(ModelComposed):
 
 {{> python-experimental/model_templates/classvars }}
 
+{{#oneOf}}
+{{#-first}}
+    attribute_map = {}
```
jirikuncar, line 20: The correct solution would include all properties defined directly on the model and not in any `oneOf` one.

jirikuncar:
I have a quick update for templates that *works for me* ™️. I could not find out how to get directly defined properties on the model.

spacether:
> Thank you for the explanation. This is similar to a problem that was also identified in Java jersey2. AFAIK, in case of composed schemas, the following use cases are not currently handled:
> 
>     1. Same property name appears in two different schemas with different types.
> 
>     2. Same property name, different constraints, e.g. maxItems, minItems, maximum, minimum, pattern...
> 
> 
> The problem exists for `anyOf`, `oneOf` and `allOf`, not just `oneOf`. A schema can have a combination of `anyOf`, `oneOf` and `allOf`, and conceivable the same property could appear multiple times across all three.
> 
> When two properties with the same name are specified, afaik the "last one wins", i.e. the last referenced schema in the OAS document wins.
> 
> the "last" property wins

My understanding is that this is the cause. In java the codegenmodel pools all variable inputs from all oneOf anyOf allOf midels as inputs in the composed model. So we can have collisions. Maybe we should store plural types for CodegenParameter and CodegenProperty. What do you think @wing328 @jimschubert

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
spacether: I am a little concerned that this implementation turns off type checking in the composed schema. This is probably okay, because when we assign the variables to the oneOf/anyOf class instances, the type is also checked there.

Can't this happen for all of these cases for required + optional variables:
1. oneOf schemas have a variable with conflicting type
2. anyOf schemas have a variable of conflicting type
3. oneOf + anyOf schemas have a variable of conflicting type?

You have also revealed that the docstring for this function is incorrect for these 1-3 use cases.
Python-experimental has previously addressed issues with composed schema inputs and tweaking them in the past with the method [addNullDefaultToOneOfAnyOfReqProps](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L738). It goes through and sets composed schema required inputs to have a default null value in the composed schema init signature. This makes those inputs named arguments and makes them optional for users to pass them in. This is because:
we can have 2 oneOf schemas, schemaA, and schemaB each with 1 required variable, varA, varB. If we are only making an instance of schemaA, then varB is not required. So we set them all to null, and when we make an instance of schemaA, it requires that we pass in varA.

How about we do some similar tweaking at the composed schema variable level here?
We could
- look for variable type collisions in the oneOf/anyOf variables of a composed schema
- update their data types at the composed schema level to be a tuple of the allowed types
- this lets us keep using the mustache template as-is
- this keeps our documentation up to date

What do you think?

sebastien-rosset:
As a side note, the schema may be brittle if two `oneOf` child schemas define a appleColor and bananaColor that share common values such as "yellow", unless there are other required distinguishing constraints. The issue raised here is still valid.

spacether:
> 2\. Same property name, different constraints, e.g. maxItems, minItems, maximum, minimum, pattern...

For the record python-experimental handles this use case.
We would store the property in two model instances and differing validation would be checked in each model.

sebastien-rosset:
> > 2. Same property name, different constraints, e.g. maxItems, minItems, maximum, minimum, pattern...
> 
> For the record python-experimental handles this use case.
> We would store the property in two model instances and differing validation would be checked in each model when the composed model is instantiated.

Below is `oneOf` schema where the breed property can be either:
1. A string of 1 to 10 [a-z] characters, or
1. A string of 11 to 20 [a-z] characters.

I know, there is a much simpler way to write the schema, this is just an example to test the edge cases.
```yaml
    TestOneOf:
      oneOf:
        - type: object
          properties:
            breed:
              type: string
              pattern: '[a-z]{1,10}'
        - type: object
          properties:
            breed:
              type: string
              pattern: '[a-z]{11,20}'
```

Here is the generated python code.
```python
class TestOneOf(ModelComposed):
    validations = {
        ('breed',): {
            'regex': {
                # NOTE The last one wins. Doesn't that mean the generated code will always match the input data against this pattern?
                'pattern': r'[a-z]{11,20}',  # noqa: E501
            },
        },
    }
```

I'm not saying this breaks my use cases.

spacether:
> > > 1. Same property name, different constraints, e.g. maxItems, minItems, maximum, minimum, pattern...
> > 
> > 
> > For the record python-experimental handles this use case.
> > We would store the property in two model instances and differing validation would be checked in each model when the composed model is instantiated.
> 
> Below is `oneOf` schema where the breed property can be either:
> 
>     1. A string of 1 to 10 [a-z] characters, or
> 
>     2. A string of 11 to 20 [a-z] characters.
> 
> 
> I know, there is a much simpler way to write the schema, this is just an example to test the edge cases.
> 
> ```yaml
>     TestOneOf:
>       oneOf:
>         - type: object
>           properties:
>             breed:
>               type: string
>               pattern: '[a-z]{1,10}'
>         - type: object
>           properties:
>             breed:
>               type: string
>               pattern: '[a-z]{11,20}'
> ```
> 
> Here is the generated python code.
> 
> ```python
> class TestOneOf(ModelComposed):
>     validations = {
>         ('breed',): {
>             'regex': {
>                 # NOTE The last one wins. Doesn't that mean the generated code will always match the input data against this pattern?
>                 'pattern': r'[a-z]{11,20}',  # noqa: E501
>             },
>         },
>     }
> ```
> 
> I'm not saying this breaks my use cases.

Ah, my mistake python-experimental does not have this working yet. It looks like I need to combine or strip out oneOf and anyOf validations at the composed schema level. Then is should work. Thank you for pointing that out.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
jirikuncar: > 1. oneOf schemas have a variable with conflicting type

This can happen and we should support it.

> 2. anyOf schemas have a variable of conflicting type

It can happen. I am not sure how it should be resolved when a value matches multiple conflicting types.

> 3. Does our generator allow type collisions between variables in (allOf + oneOf) or (allOf + anyOf)? It should not, that violates the openapi spec.

You are right that *It should not*.

I am not sure about the solution though. I would completely remove validation on base schema for properties defined in *X*Of models.

---
I agree that we should not merge this mustache template patch. I was looking for a quick solution that would unblock me so I can at least generate code that can deserialize server response without raising an exception.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
```diff
@@ -112,6 +112,12 @@
             openapi_types (dict): The key is attribute name
                 and the value is attribute type.
         """
+{{#oneOf}}
+{{#-first}}
+        return {}
+{{/-first}}
+{{/oneOf}}
+{{^oneOf}}
```
jirikuncar, line 120: This should be removed after we find a solution for https://github.com/OpenAPITools/openapi-generator/pull/6674/files#r440966593.
```suggestion
```

samples/openapi3/client/features/one-of/python-experimental/.travis.yml
```diff
@@ -0,0 +1,17 @@
+# ref: https://docs.travis-ci.com/user/languages/python
+language: python
+python:
+  - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
```
jirikuncar, line 8: 🙏 
```suggestion
```

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
sebastien-rosset: > Does our generator allow type collisions between variables in (allOf + oneOf) or (allOf + anyOf)? It should not, that violates the openapi spec.

One can intentionally or accidentally write a OAS document which is 1) syntactically compliant with the JSON schema specification and 2) will not match *any* input payload. It's an interesting question whether it's even possible to analyze a schema and conclude whether there is at least one matching input data. 
IMO this should be handled with the `--validation` argument, and the code generator should still accept the OAS document. I.e. you want to know if a OAS doc is syntactically valid.

Consider `pattern`. There is an infinite number of regex that will not match any input string, including the empty string. Here is one simple example:
```yaml
type: string
pattern: .*
```

samples/openapi3/client/features/one-of/python-experimental/.travis.yml
jimschubert: @spacether do we need all these pythons?

🐍

jimschubert:
I scanned through this, and I think the approach makes sense. Were there additional requests to be made before merge (@spacether and @sebastien-rosset)?

There are a handful of proposed changes, were you wanting to change those before merge @jirikuncar ?

samples/openapi3/client/features/one-of/python-experimental/.travis.yml
spacether: Can we keep 3.5?

jimschubert: Hm. I do see:

```
$ cat samples/client/petstore/python-experimental/.travis.yml
# ref: https://docs.travis-ci.com/user/languages/python
language: python
python:
  - "3.5"
  - "3.6"
  - "3.7"
  - "3.8"
# command to install dependencies
install:
  - "pip install -r requirements.txt"
  - "pip install -r test-requirements.txt"
# command to run tests
script: pytest --cov=petstore_api
```

Maybe this will be resolved if master gets integrated up into this branch.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
spacether: So this omits variables that are defined in:
```
components:
  schemas:
    ComposedModelWithProperties:
      properties:
        - color:
        type: string
      oneOf:
        - Cat
        - Dog
```
Can you add an example like I have shown above and show that color is present in ComposedModelWithProperties.openapi_types?

jirikuncar:
@jimschubert all changes has been resolved by upgrading to latest master.

@spacether it looks like you have doing a big refactoring in #7274. I will wait until it's merged before adding more changes to this PR.

spacether:
> @spacether it looks like you have doing a big refactoring in #7274. I will wait until it's merged before adding more changes to this PR.

Thank you for your patience. 7274 was just merged so you can continue work here.

jimschubert:
@spacether You'd requested changes on this PR. Do you have any remaining concerns? I'd like to get it in before addition conflicts emerge.

spacether:
Yes, my change requested comment still applies.
Why not iterate over vars to only define types core parameters in a composed schema that are defined using the composed schema key of properties?
The new code that we have now:
- uses special handling when we have oneOf schemas
- if we have oneOf + properties in a composed schema that breaks that use case
- what about i we only have only anyOf schemas?
- what about if we have oneOf + allOf?
Iterating over vars should handle all of these above use case and prevent type conflicts.
If we iterate over vars, then the data types for conflicting oneOf types will be defined in the oneOf schemas, which should work. The problem that we have today is that a parameter has its type defined in the composed schema AND schemas oneOfA and oneOfB. With my suggestion, this conflicting parameter will be an additionalProperty in the composed schema and will have individual type definition in oneOfA + oneOfB.

To do this, in modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/model_composed.mustache
we need to iterate over vars when defining openapi_type and validations. And perhaps when defining attribute_map. Not sure about attribute_map. That map just goes from python name to spec var name.

spacether:
I am working on fixing this in two PRS:
1. fix additionalProperties in models #8802
right now some models have additionalPropertiesType being defined as None when it should have a real value
2. make the CodegenModel/CodegenProperty/CodegenParameter/CodegenResponse.vars only contain the properties defined in that schema #8816

spacether:
Closing due to inactivity

PR #6684, opened 2020-06-16
Improve sttpOpenApiClient generator

ghostbuster91:
We'd like to propose the following changes to the scala-sttp generator. While the existing version works quite well (thanks to all of the contributors!), in a couple of places it diverges from the way sttp is most often used in the wild, and from the original design goals. 

List of changes together with reasons behind them:
- allow libraries' versions customization - although it is not hard to change them by hand we think that it plays nicely with the rest of the cli options and improves the overal user's experience
- return erros by default in a separate error channel - being explicit in terms of  returned error types is the way to go in scala (there is a switch which allows to restore the old behavior)
- some of the helper functions were incorporated into sttp directly
- some of the helper functions were removed as they were redundant/unnecessary
- auth parameters are passed explicitly rather than implicitly - we belive that the rule of tumb to follow when choosing between imlicit vs explicit arguments is whether they are business or techincal one. Auth parameters belongs to the first category.
- because credentials were no longer passed implicitly there was no need to keep custom auth releated classes
- json processing related instances are now imported via single import statement
- fixed multipart requests
- removed ApiModel class - it was artifical and redundant
- added circe as another json library - circe is as popular as json4s if not more
- removed sttpClientCodegen inheritance from akkaClientCodegen
- extract classes to deal with cli properties - Dealing with them is quite cumbersome and what is worse it is error-prone. It is easy to make a mistake, especially when dealing with properties which mutually exlude themselfs. Having another layer of abstraction on top of them makes things much easier.
- build.sbt was reduced to minimum

Despite breaking changes I opened this PR against master branch since the sttp generator is marked as beta. But let me know if that needs to be changed.

### PR checklist

Summoning technical committee: 
@clasnake  @jimschubert , @shijinkui , @ramzimaalej , @chameleon82 , @Bouillie


modules/openapi-generator/src/main/resources/scala-sttp/api.mustache
```diff
@@ -4,42 +4,35 @@ package {{package}}
...
-        {{>paramFormCreation}},{{/formParams}}
-      )){{/formParams.0}}{{#bodyParam}}
+        {{>paramFormCreation}}{{#hasMore}}, {{/hasMore}}{{/formParams}}
+      )){{/isMultipart}}{{#isMultipart}}
+      .multipartBody(Seq({{#formParams}}
+        {{>paramMultipartCreation}}{{#hasMore}}, {{/hasMore}}{{/formParams}}
+      ).flatten){{/isMultipart}}{{/formParams.0}}{{#bodyParam}}
       .body({{paramName}}){{/bodyParam}}
-      .response(asJson[{{>operationReturnType}}])
+      .response({{#separateErrorChannel}}asJson{{/separateErrorChannel}}{{^separateErrorChannel}}asJsonAlwaysUnsafe{{/separateErrorChannel}}[{{>operationReturnType}}])
```
chameleon82, line 34: Looks good. Could you help to separate api errors as exceptions as described here https://github.com/softwaremill/sttp-model/issues/5 ? The main idea is to split application exceptions (ideally with models) from transport level exceptions by using expected status codes.
Thanks

modules/openapi-generator/src/main/resources/scala-sttp/build.sbt.mustache
```diff
@@ -2,26 +2,27 @@ version := "{{artifactVersion}}"
 name := "{{artifactId}}"
 organization := "{{groupId}}"
 
-scalaVersion := "2.13.0"
-
-crossScalaVersions := Seq(scalaVersion.value, "2.12.10",  "2.11.12")
+scalaVersion := "2.13.2"
```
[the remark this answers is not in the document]
adamw: Our goal here was to create an sbt project with a minimal number of settings. 

Cross-compilation as an option might be a good compromise - though as a vast majority of applications (not libraries) target a single Scala version, I'd opt for no-cross-compile as a reasonable default. What do you think @ghosbuster91?

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/api/PetApi.scala
```diff
@@ -14,20 +14,11 @@ package org.openapitools.client.api
...
-object PetApi {
-
-  def apply(baseUrl: String = "http://petstore.swagger.io/v2")(implicit serializer: SttpSerializer) = new PetApi(baseUrl)
-}
-
-class PetApi(baseUrl: String)(implicit serializer: SttpSerializer) {
-
-  import Helpers._
-  import serializer._
+class PetApi(baseUrl: String = "http://petstore.swagger.io/v2") {
```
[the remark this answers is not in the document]
adamw: I'm not sure what do you mean here? Do you mean an `object PetApi`? What would be its contents?

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/core/ApiInvoker.scala
```diff
@@ -1,60 +0,0 @@
...
-
-  /**
-    * Allows request execution without calling apiInvoker.execute(request)
-    * request.result can be used to get a monad wrapped content.
-    *
-    * @param request the apiRequest to be executed
-    */
-  implicit class ApiRequestImprovements[R[_], T](request: RequestT[Identity, Either[ResponseError[Exception], T], Nothing]) {
-
-    def result(implicit backend: SttpBackend[R, Nothing, Nothing]): R[T] = {
```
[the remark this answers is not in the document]
adamw: The way sttp-client is normally used (at least, what we've seen on gitter, issue reports etc.) is that people simply call `.send()` with an implicit backend in scope. What kind of functionality would the helper add?

modules/openapi-generator/src/main/resources/scala-sttp/api.mustache
adamw: Transport exceptions are already modeled using `SttpClientException`. These exceptions are thrown when read or connection exceptions occur, and result in a failed effect (failed `Future` or `IO`).

Though you are right that there's a piece missing here, if the OpenAPI description contains an e.g. JSON representation of an error? Then yes, we will need a custom hierarchy (as opposed to the standard `ResponseError`), covering three possibilities:

* `DeserializationError[T](body: String, error: T)` - if deserializing the success or error json fails
* `UnknownHttpError(body: String)` - for unmapped status codes
* `HttpError[T](body: T)` - for mapped status codes

Then, the overall request type would be: `Request[Either[ResponseError[T, U], V]]`, where:

* `T` is the type of deserialization errors, e.g. `io.circe.Error`
* `U` is the type of the http errors - but as there might be a couple of these, probably this needs to be a sealed trait, which is a implemented by all possible errors for this endpoint? This could be tricky, as each endpoint might have different error cases (e.g. `NotFound`).
* `V` is the type of the deserialized http success value (e.g. `User`)

Of course we'd need better naming for these :)

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/api/PetApi.scala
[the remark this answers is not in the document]
chameleon82: just a constructor like
```
object PetApi {
  def apply(baseUrl: String = "http://petstore.swagger.io/v2") = new PetApi(baseUrl)
}
```
is it make sense? next 5.0 release is not backward compatible, so it's optional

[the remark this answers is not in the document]
adamw: ah yes sure, make sense :)

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/core/ApiInvoker.scala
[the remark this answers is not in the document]
chameleon82: I think that helper can be on the app side as error transformations can be different from app to app.
The main idea of that helper is to get successful result directly. For example for `Try` framework it could be something like 
```
val pet: Pet = new PetApi("http://swagger").getPetById(1).result()
```

instead of code like next for every method
```
 new PetApi("http://swagger").getPetById(1).send()
   .flatMap { 
      case Right(pet) => pet 
            doSomething(pet)
      case ...
   }
```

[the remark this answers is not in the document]
adamw: For that purpose we've introduced the `separateErrorChannel` option: using sttp's mechanism (response specifications), requests can be modeled to return full error information (so the response will be roughly an `Either[Error, Success]`), or the errors can be represented as failed effects (as the errors are `Throwables`).

Of course people might want to generate code using full error information and then flatten the result anyway. Though here, as you write, it's quite an application-specific thing. I think if I wanted to flatten the results, I'd write an extension method, which would convert `F[Either[Error, Success]]` into a `F[Success]`, e.g.:

```
new PetApi("http://swagger").getPetById(1).send().flattenErrors
```

where

```
implicit class TryOps[E <: Throwable, S](r: Try[Either[E, S]]) {
  def flattenErrors: Try[S] = r.flatMap {
    case Left(e) => Failed(e)
    case Right(s) => Success(s)
  }
}
```

but I think this would belong to the specific app - not in the generated code. Otherwise we should have this in sttp directly :)

chameleon82:
@ghostbuster91 I've tested compiled code, looks very good. 
As a suggestion it might be good to have simple example of simple app for beginners in README file.
Please also check comments above and re-trigger CI's

modules/openapi-generator/src/main/resources/scala-sttp/build.sbt.mustache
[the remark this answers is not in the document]
ghostbuster91: Let's restore cross-compilation to scala 2.12 until it reaches EOL.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
```diff
@@ -87,4 +207,290 @@ public CodegenOperation fromOperation(String path,
...
+    private static class MainPackageProperty extends StringProperty {
+
+        private MainPackageProperty() {
+            super("mainPackage", "Top-level package name, which defines 'apiPackage', 'modelPackage', " +
+                    "'invokerPackage'", DEFAULT_PACKAGE_NAME);
+        }
+
+        @Override
+        public void updateAdditionalProperties(Map<String, Object> additionalProperties) {
+            if (additionalProperties.containsKey(name)) {
```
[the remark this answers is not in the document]
ghostbuster91: Hmm, this is definitively a good idea to cover it with some tests

ghostbuster91:
@chameleon82 Thanks for the review and your suggestions, I'm going to start working on it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
```diff
@@ -16,50 +16,170 @@
...
+        additionalProperties.put("fnCapitalize", new CapitalizeLambda());
+        additionalProperties.put("fnCamelize", new CamelizeLambda(false));
+        additionalProperties.put("fnEnumEntry", new EnumEntryLambda());
+
+        importMapping.remove("Seq");
+        importMapping.remove("List");
+        importMapping.remove("Set");
+        importMapping.remove("Map");
+
+        typeMapping = new HashMap<>();
```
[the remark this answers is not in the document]
ghostbuster91: Actually there are not the same. In `AbstractScalaCodegen`only `importMappings` are defined, while here we remove some of the and redefine them as `typeMappings` or `instantiationTypes`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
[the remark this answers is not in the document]
ghostbuster91: It isn't an error since the `mainPackage` property isn't used anywhere. All packages are specified by `apiPackage`, `modelPackage` and `invokerPackage` properties

modules/openapi-generator/src/main/resources/scala-sttp/project/build.properties.mustache
```diff
@@ -0,0 +1 @@
+sbt.version=1.2.4
```
chameleon82, line 1: This file/version might be the cause of CI build failed with next error:

https://app.circleci.com/pipelines/github/OpenAPITools/openapi-generator/3736/workflows/1d16cd64-dd72-4f2b-8a68-92adfbdb29b7/jobs/17104/steps

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.scala-sbt#compiler-bridge_2.13;1.2.3: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[error] ## Exception when compiling 14 sources to /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/scala-sttp/target/scala-2.13/classes
[error] The compiler bridge sources org.scala-sbt:compiler-bridge_2.13:1.2.3:compile could not be retrieved.
[error]  
[error] 	Note: Unresolved dependencies path:
[error] 		org.scala-sbt:compiler-bridge_2.13:1.2.3
[error] 		  +- org.scala-sbt.temp:temp-module-52f238aede5a1d8d7a484c2bec27a144810e2ac0:1.2.3
 ``` 
Could you try sbt.version=1.2.8? Should we even specify this version explicitly?

Comment (adamw) [reply modules/openapi-generator/src/main/resources/scala-sttp/project/build.properties.mustache:1]: Yes, the sbt version should be specified in `project/build.properties` so that the version is fixed and not platform-dependent. However, we should probably use `1.3.12`, which is the newest one.

File: samples/client/petstore/scala-sttp/build.sbt [hunk 8]
```diff
@@ -2,23 +2,17 @@ version := "1.0.0"
 name := "scala-sttp-petstore"
 organization := "org.openapitools"
 
-scalaVersion := "2.13.0"
-
-crossScalaVersions := Seq(scalaVersion.value, "2.12.10",  "2.11.12")
+scalaVersion := "2.13.2"
+crossScalaVersions := Seq(scalaVersion.value, "2.12.10")
```
eugeniyk, line 6:
```suggestion
crossScalaVersions := Seq(scalaVersion.value, "2.12.11")
```

modules/openapi-generator/src/main/resources/scala-sttp/api.mustache
adamw: That's always returned in the `Response` object which wraps the body. So the type of the result of a call is e.g. `Response[Either[ResponseError, V]]`, where `V` is the type of success

ghostbuster91:
@chameleon82 I think that I fixed/reviewed all your suggestions apart from translating application's errors from openapi to sttp. 
As @adamw mentioned this might not be a trivial thing to do as each endpoint can have its own errors' hierarchy.

Still we think that it would be a great thing to have, so we would like to implement it, but to not extend this PR too much, we would like to do it in a separate one. Also to make this task a little bit easier we will first introduce some modification to the core of sttp (see https://github.com/softwaremill/sttp/issues/618).

Having said that, let me know if there is still anything which should be done before we merge this PR?

ghostbuster91:
> @ghostbuster91 I've tested compiled code, looks very good.
> As a suggestion it might be good to have simple example of simple app for beginners in README file.
> Please also check comments above and re-trigger CI's

Is there any other backend which has this simple example app in the README so I can look at it?

chameleon82:
> Still we think that it would be a great thing to have, so we would like to implement it, but to not extend this PR too much, we would like to do it in a separate one. Also to make this task a little bit easier we will first introduce some modification to the core of sttp (see softwaremill/sttp#618).

Yeah, that definitely better to create separate PR for this improvement

> Is there any other backend which has such simple example app in the README so I can look at it?

Not sure we have good example. 

@wing328 how do you think about having some "basic example" in generated README file or to have specific client documentation somewhere? It might be not easy for newbies to start using generated client properly

> Having said that, let me know if there is still anything which should be done before we merge this PR?

Imo looks pretty good. Approved. 

@jimschubert could you help with review and merge?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
```diff
@@ -87,4 +191,300 @@ public CodegenOperation fromOperation(String path,
...
+        public String getModelPackage(Map<String, Object> additionalProperties) {
+            return additionalProperties.getOrDefault(CodegenConstants.MODEL_PACKAGE, DEFAULT_PACKAGE_NAME + ".model").toString();
+        }
+
+        public String getInvokerPackage(Map<String, Object> additionalProperties) {
+            return additionalProperties.getOrDefault(CodegenConstants.INVOKER_PACKAGE, DEFAULT_PACKAGE_NAME + ".core").toString();
+        }
+    }
+
+    private static abstract class CustomLambda implements Mustache.Lambda {
```
jimschubert, line 438: Note that at we have well-tested lambdas in the `org.openapitools.codegen.templating.mustache` package which handle camel case, title case, uppercase, lowercase, etc. We should probably move these to use the tested lambdas at some point.

modules/openapi-generator/src/main/resources/scala-sttp/requests.mustache
```diff
@@ -1,48 +0,0 @@
-{{>licenseInfo}}
```
jimschubert, line 1: removal of these built-in templates is a breaking change, so this needs to get into 5.0

samples/client/petstore/scala-sttp/build.sbt
```diff
@@ -2,23 +2,17 @@ version := "1.0.0"
...
+  "org.json4s" %% "json4s-jackson" % "3.6.8"
 )
 
 scalacOptions := Seq(
   "-unchecked",
   "-deprecation",
   "-feature"
 )
-
-publishArtifact in (Compile, packageDoc) := false
```
jimschubert, line 24: We should consider adding this back. Default behavior should be to avoid accidentally publishing content. Users can then configure publishing and ignore the build file, or enable published at invocation.

jimschubert:
I left a few minor comments. We can address them at any point in the future, like next time we modify this generator.

samples/client/petstore/scala-sttp/build.sbt
adamw: Right, well our goal here was to create a minimal working project, keeping things as simple as possible, usable for an application developer (who doesn't normally publish artifacts, or has a custom publishing pipeline anyway). But maybe our approach here is incorrect - are there some guidelines as to what should be the functionality of the generated build? 

Thanks!

adamw:
Thanks for merging!

As a next step, as @ghostbuster91 mentioned, we'd like to work on improving the typed-errors in sttp-client itself, and then building on that, improving the generator.

Another direction is creating an sbt plugin which would generate the client as part of an existing project's build (so not a self-contained build, but only the sources, much like e.g. the gRPC sbt plugin works). This is what we described in https://github.com/OpenAPITools/openapi-generator/issues/6685. Is this something that we could work on as part of OpenAPITools, and if so, should this be scoped only to this generator, or maybe it would be usable by other generators as well?

modules/openapi-generator/src/main/resources/scala-sttp/requests.mustache
ghostbuster91: Is compatibility maintained between generators which aren't stable? (sttp generator is marked as beta)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
ghostbuster91: I agree. I just copy-pasted it from scala-akka generator AFAIR. We should change it there as well.

PR #6688, opened 2020-06-16
[Java] Test simple oneOf model

jirikuncar:
Extracts a test for `oneOf` model for Java Jersey2 from https://github.com/OpenAPITools/openapi-generator/pull/6674.

### PR checklist


wing328:
What about adding the spec (oneOfOverlapingTypes.yaml) to the existing one used by Java (jersey2) client generator?

modules/openapi-generator/src/test/resources/3_0/oneOfOverlapingTypes.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - green
+        - red
+    apple:
+      title: apple
+      type: object
+      properties:
+        kind:
+          type: string
+        color:
+          $ref: '#/components/schemas/appleColor'
```
spacether, line 35: Can you make the color property required in both apple and banana?
Some languages treat required and optional variables differently, and making it required makes it required at the composed schema level also.

wing328: Just to add that the `Fruit` definition seems to be invalid with respect to how `oneOf` is defined.

Consider `{"color":"green"}` which will match both Apple and Banana and result in the violation of the `oneOf` definition (one and only one match)

jirikuncar: @wing328 `{"color":"green"}` this is the whole point! There is an overlap in a single field but you need other fields to determine the right model.

wing328: Then those other fields need to be required instead. Otherwise, `{"color":"green"}` will match both.

modules/openapi-generator/src/test/resources/3_0/oneOfOverlapingTypes.yaml
```diff
@@ -0,0 +1,48 @@
...
+    bananaColor:
+      type: string
+      enum:
+        - yellow
+        - green
+    banana:
+      title: banana
+      type: object
```
jirikuncar, line 43:
```suggestion
      type: object
      required:
        - color
```

modules/openapi-generator/src/test/resources/3_0/oneOfOverlapingTypes.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - $ref: "#/components/schemas/banana"
+    appleColor:
+      type: string
+      enum:
+        - yellow
+        - green
+        - red
+    apple:
+      title: apple
+      type: object
```
jirikuncar, line 30:
```suggestion
      type: object
      required:
        - color
```

modules/openapi-generator/src/test/resources/3_0/oneOfOverlapingTypes.yaml
spacether: So is the point of adding these tests to show differences in enum values or in the data type of the property?
My understanding was that the data type of the property was the bigger issue.
For example having:
- color be string in one model
- color be integer in another

Issue #5594, opened 2020-03-15
[REQ] [rust] Support XML objects in requests and responses


sverch:
### Is your feature request related to a problem? Please describe.

I'm trying to generate a client that works with these [AWS OpenAPI specs](https://github.com/APIs-guru/openapi-directory/tree/master/APIs/amazonaws.com/), but the AWS API is XML based, and the specs there all have XML definitions.

I don't know for sure how to tell whether a generator supports this, but if I understand correctly based on [this comment](https://github.com/OpenAPITools/openapi-generator/blob/ea55968737fa80a3e306d1b08032ff4d911774c9/modules/openapi-generator-core/src/main/java/org/openapitools/codegen/meta/features/GlobalFeature.java#L108) that says "this option indicates whether XML structures can be defined by spec document and honored by the caller", the `XMLStructureDefinitions` is what says whether a given generator supports XML objects in the requests and responses.

However, it doesn't appear that any clients actually support this feature:

```
$ git grep "|XMLStructureDefinitions|✓|"
docs/generators/dynamic-html.md:|XMLStructureDefinitions|✓|OAS2,OAS3
docs/generators/html.md:|XMLStructureDefinitions|✓|OAS2,OAS3
docs/generators/html2.md:|XMLStructureDefinitions|✓|OAS2,OAS3
docs/generators/openapi-yaml.md:|XMLStructureDefinitions|✓|OAS2,OAS3
docs/generators/openapi.md:|XMLStructureDefinitions|✓|OAS2,OAS3
docs/generators/spring.md:|XMLStructureDefinitions|✓|OAS2,OAS3
```

Those are all either documentation or server generators.

## Describe the solution you'd like

I would like the generated rust code to support XML based service definitions.

I managed to get the generated code working on a basic level (single call), so here's some idea of what I think is needed for this:

- Use [this XML parsing library](https://github.com/RReverser/serde-xml-rs) to replace all uses of the [JSON parser](https://github.com/serde-rs/json) that is used currently.
- Honor the [XML only features](https://swagger.io/docs/specification/data-models/representing-xml/) in the specification.

## Describe alternatives you've considered

I considered using a different language client, but it doesn't seem like I have another option.

However, I also filed this because I'm not sure if I'm actually understanding this correctly, so any corrections are very welcome!

## Additional context

[...]

PR #6690, opened 2020-06-16
[Core, Rust Server] anyOf / oneOf support for Rust Server

richardwhiuk:
This adds anyOf/oneOf support for the Rust Server generator using the support added in https://github.com/Metaswitch/swagger-rs/pull/112

There's a couple of minor core changes to InlineModelResolver - @OpenAPITools/generator-core-team:

- If we get a model named `12345AnyOfObject`, rename it to `_12345AnyOfObject`, not `_2345AnyOfObject`.

  This means for Rust Server, it'll end up as Model12345AnyOfObject, instead of Model2345AnyOfObject, which is nonsensical.

  ⚠️ I think this is technically breaking, but it's clearly bugged at the moment.

- Perform inline model resolution for enums, and pass the type through when performing inline model resolution.

There's a couple of minor fixups in Rust Server:

- Some minor template tweaking, including putting the IntoHeader implementation below the model definition.

- Update the supported features - we've support XML Structure Definition for a while - but the global feature flag is wrong - see user feedback in https://github.com/OpenAPITools/openapi-generator/issues/5594.

- Handle FromStr properly, which breaks for the oneOf model otherwise.

Optimistically, I'd like to merge this before 5.0 gets cut.

This change is by @mirw and myself on behalf of @Metaswitch.

## Rust Server Technical Committee

- @frol
- @farcaller
- @paladinzh

### PR checklist


modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
```diff
@@ -362,38 +362,33 @@ private void flattenComposedChildren(OpenAPI openAPI, String key, List<Schema> c
...
-                        addGenerated(innerModelName, innerModel);
-                        Schema schema = new Schema().$ref(innerModelName);
-                        schema.setRequired(op.getRequired());
-                        listIterator.set(schema);
-                    } else {
-                        Schema schema = new Schema().$ref(existing);
-                        schema.setRequired(op.getRequired());
-                        listIterator.set(schema);
-                    }
+            if ((component.get$ref() == null) &&
```
etherealjoy, line 365: cc @OpenAPITools/generator-core-team for review

jimschubert: This needs the `component != null` check that's part of the removed check condition because, unfortunately, people using the generator code programmatically could provide a component array with null as an array item. As an alternative, I guess we could convert from direct iterator usage to a stream and filter by `Object::nonNull`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -1191,8 +1189,13 @@ public String toInstantiationType(Schema p) {
 
     @Override
     public CodegenModel fromModel(String name, Schema model) {
+        LOGGER.debug("fromModel (schema): " + model);
```
jimschubert, line 1192: As a user with debug enabled, I'd have three concerns:

1. The use of debug calls in this method will cause a ton of noise as it'll be invoked with somewhat large object dumps every time for every model, potentially more than once per model
2. It's not clear from the log message what the difference is between `fromModel (schema):`, `fromModel (base end):`, and `fromModel (end):`.
3. The types being dumped are all different (Schema input, ADT before mutation, and ADT after mutation), which could be confusing especially if they have similar properties.

As a small point, the logging framework will interpolate content based on whether or not this log level is enabled using the `{}` formatter. Without this, a new string and model dump (which may or may not just be member reference string in this case, depends on the toString of the types) will be concatenated every time this line is hit regardless of whether or not debug is enabled.

```suggestion
        LOGGER.debug("fromModel (schema): {}", model);
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -1191,8 +1189,13 @@ public String toInstantiationType(Schema p) {
...
+
         Map<String, Schema> allDefinitions = ModelUtils.getSchemas(this.openAPI);
         CodegenModel mdl = super.fromModel(name, model);
+
+        LOGGER.debug("fromModel (base end): " + mdl);
```
jimschubert, line 1197:
```suggestion
        LOGGER.debug("fromModel (base end): {}", mdl);
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -1245,6 +1250,8 @@ public CodegenModel fromModel(String name, Schema model) {
             mdl.additionalPropertiesType = getTypeDeclaration(additionalProperties);
         }
 
+        LOGGER.debug("fromModel (end): " + mdl);
```
jimschubert, line 1253:
```suggestion
        LOGGER.debug("fromModel (end): {}", mdl);
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
```diff
@@ -1529,6 +1558,8 @@ private String matchingIntType(boolean unsigned, Long inclusiveMin, Long inclusi
             Map<String, Object> mo = (Map<String, Object>) _mo;
             CodegenModel cm = (CodegenModel) mo.get("model");
 
+            LOGGER.debug("postProcessModels: " + cm);
```
jimschubert, line 1561:
```suggestion
            LOGGER.debug("postProcessModels: {}", cm);
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustServerCodegen.java
richardwhiuk: I wanted to leave these in as they were very helpful debugging issues, so to address your feedback:

1. I've downgraded these to trace
2 + 3. I've clarified the logs to make it clearer that the first is the schema, and the last is the model. I've eliminated the middle line.

richardwhiuk:
CircleCI passsed, but seems to have failed to sync with GitHub - https://app.circleci.com/pipelines/github/OpenAPITools/openapi-generator/8310/workflows/54e31e58-6aab-45e7-83f9-4b41106f8790

PR #6699, opened 2020-06-17
Make ApiClient in retrofit2 be able to use own OkHttpClient

tgerth:
### Description
If I want to set a customized okHttpClient to the adapterBuilder like this:
```
val apiClient = ApiClient()
apiClient.adapterBuilder
      .baseUrl(baseUrl)
      .client(customizedOkHttpClient)
```
the client gets overwritten by this:
```
public <S> S createService(Class<S> serviceClass) {
        return this.adapterBuilder.client(this.okBuilder.build()).build().create(serviceClass);
    }
```
and unfortunately this method:
```
public void configureFromOkclient(OkHttpClient okClient) {
        this.okBuilder = okClient.newBuilder();
        this.addAuthsToOkBuilder(this.okBuilder);
    }
```
is of no use to me because the okBuilder would just get the clients fields but I need/want to have the client itself respectively its instance. So I added a boolean field "okBuilderUsed" which checks if the okBuilder is used and if it's used then createService will return its client and if not then createService will just return the adapterBuilder:
```
public <S> S createService(Class<S> serviceClass) {
        return this.okBuilderUsed ? this.adapterBuilder.client(this.okBuilder.build()).build().create(serviceClass) : this.adapterBuilder.build().create(serviceClass);
    }
```

### PR checklist


wing328:
cc @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04) @lwlee2608 (2019/10) @bkabrda (2020/01)

wing328:
Please update the samples by running `./bin/generate-samples.sh ./bin/configs/java-retrofit2*`

https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-can-i-update-the-samples-i-couldnt-find-the-script-under-bin-or-windows-batch-files-under-binwindows

cbornet:
I don't really like the `okBuilderUsed` variable. IMO it would be better to have an immutable ApiClient with a builder and a `setClient` method. This would be a breaking change but maybe it can be done for v5 ?
Meanwhile, can't you call adapterBuilder.build().createService() yourself ?

wing328:
> This would be a breaking change but maybe it can be done for v5 ?

Yes we can include in the v5.0.0 release (release date not yet finalized)

tgerth:
> I don't really like the `okBuilderUsed` variable. IMO it would be better to have an immutable ApiClient with a builder and a `setClient` method. This would be a breaking change but maybe it can be done for v5 ?
> Meanwhile, can't you call adapterBuilder.build().createService() yourself ?

I actually kind of expected that this is not a beautiful solution but I thought why not give a try. Could we also do your proposed solution for a CallFactory like ```setCallFactory```? 
And yeah you are right haha I don't know why I didn't see it, thanks.

tgerth:
@cbornet I'm not sure what exactly your thought was, but I would suggest something like this
```
  public <S> S createService(Class<S> serviceClass) {
    return adapterBuilder.build().create(serviceClass);
  }
  
  public void setOwnClient(OkHttpClient client) {
    adapterBuilder.client(client)
  }
  
  public void setOwnCallFactory(Call.Factory callFactory) {
    adapterBuilder.callFactory(callFactory)
  }
  
  public void setOkBuilderClient() {
    adapterBuilder.client(okBuilder.build())
  }
```
so the change for the current user would be that they would have to make a call to setOkBuilderClient but I think that would be a reasonable change(?)

jonnii:
I did a similar change for the regular okhttp client https://github.com/OpenAPITools/openapi-generator/pull/6401. Would be good if we can make this the same for both?

jonnii:
@cbornet what about taking the okhttp client as a constructor param? this removes any uncertainty about which order things need to be called but more importantly _doesn't_ instantiate an `OkHttpClient` which isn't free.

tgerth:
@jonnii Taking the okhttp client as a constructor param instead of the setClient method? Or how do you mean it?

jonnii:
@tgerth if you look at `createDefaultAdapter()` (which is called from the default constructor) it creates an instance of the okhttp client. if you're supplying your own okhttp client then this operation is redundant and should be avoided because instantiating a client isnt' free. my suggestion is to do something like

```
... = new ApiClient(okHttpClient)
```

instead of `setClient` because then you have the option of skipping creating the default client and using the one you supply.

cbornet:
> if you look at createDefaultAdapter() (which is called from the default constructor) it creates an instance of the okhttp client.

No it creates a builder not the actual client. A client is created only when calling createService.

jonnii:
Sorry, i meant a builder - that's also not free:

https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt#L470

it instantiates a dispatcher, a connection pool etc...

cbornet:
Are those allocations really that heavy compared to the rest ?
That said, it's not hard to do the constructor so why not. If you don't instantiate the okhttp builder, beware to send a proper exception in the methods that would need it.

jonnii:
I think it comes down to your usage pattern. If you can create a single `ApiClient` and re-use it then tbh the overhead of creating a building you don't use isn't a big deal. Unfortunately for my system we need to generate a new `jwt` for _every request_ which requires first hitting our auth server before sending our request out. We used to do this by generating a new `ApiClient` for each request but noticed that we were adding ~20-30ms to _each request_ by doing this - instead we now reuse the same client and some okhttp interceptor magic so we can share the okhttp client across `ApiClient`. We didn't see the full benefit of this until we were passing in the client via the ctor eliding the creating of the builder.

tgerth:
so maybe something like this:
```
private Map<String, Interceptor> apiAuthorizations;
private OkHttpClient.Builder okBuilder;
private Retrofit.Builder adapterBuilder;
private JSON json;
private OkHttpClient okHttpClient;

public ApiClient() {
    apiAuthorizations = new LinkedHashMap<String, Interceptor>();
    createDefaultAdapter();
    okBuilder = new OkHttpClient.Builder();
}

public ApiClient(OkHttpClient client){
    apiAuthorizations = new LinkedHashMap<String, Interceptor>();
    createDefaultAdapter();
    okHttpClient = client;
}

public void createDefaultAdapter() {
    json = new JSON();

    String baseUrl = "{{{basePath}}}";
    if (!baseUrl.endsWith("/"))
      baseUrl = baseUrl + "/";

    adapterBuilder = new Retrofit
      .Builder()
      .baseUrl(baseUrl)
      {{#useRxJava}}
      .addCallAdapterFactory(RxJavaCallAdapterFactory.create())
      {{/useRxJava}}
      {{#useRxJava2}}
      .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
      {{/useRxJava2}}
      {{#useRxJava3}}
      .addCallAdapterFactory(RxJava3CallAdapterFactory.create())
      {{/useRxJava3}}
      .addConverterFactory(ScalarsConverterFactory.create())
      .addConverterFactory(GsonCustomConverterFactory.create(json.getGson()));
}

public <S> S createService(Class<S> serviceClass) {
    if (okHttpClient != null) {
        return adapterBuilder.client(okHttpClient).build().create(serviceClass);
    else {
        return adapterBuilder.client(okBuilder.build()).build().create(serviceClass);
    }
}


```

cbornet:
Yes. And null checks on okBuilder where needed.

wing328:
CircleCI reports some failure in building the Java client, e.g.
```
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java:[352,10] illegal start of expression
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java:[352,36] ';' expected
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java:[352,58] ';' expected
[ERROR] /home/circleci/OpenAPITools/openapi-generator/samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java:[411,2] reached end of file while parsing
[ERROR] -> [Help 1]
```
Please refer to https://app.circleci.com/pipelines/github/OpenAPITools/openapi-generator/3722/workflows/3fe3d6fa-0df5-4e7e-bf30-7c5e135c23a7/jobs/17090/steps for the details.

modules/openapi-generator/src/main/resources/Java/libraries/retrofit2/ApiClient.mustache
```diff
@@ -357,7 +365,9 @@ public class ApiClient {
       throw new RuntimeException("auth name \"" + authName + "\" already in api authorizations");
     }
     apiAuthorizations.put(authName, authorization);
-    okBuilder.addInterceptor(authorization);
+    if(okBuilder != null){
```
cbornet, line 368: May be better to throw an Exception with an explicit message if okBuilder is null ?

tgerth:
sry I forgot a semicolon😂😅

samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java
```diff
@@ -137,7 +137,7 @@ public void createDefaultAdapter() {
   public <S> S createService(Class<S> serviceClass) {
     if (okHttpClient != null) {
         return adapterBuilder.client(okHttpClient).build().create(serviceClass);
-    else {
+    }else {
```
cbornet, line 140: space missing

samples/client/petstore/java/retrofit2/src/main/java/org/openapitools/client/ApiClient.java
```diff
@@ -311,9 +311,11 @@ public ApiClient addAuthorization(String authName, Interceptor authorization) {
       throw new RuntimeException("auth name \"" + authName + "\" already in api authorizations");
     }
     apiAuthorizations.put(authName, authorization);
-    if(okBuilder != null){
-      okBuilder.addInterceptor(authorization);
+    if(okBuilder == null){
+    	throw new Exception("okBuilder is null")
```
cbornet: Use RuntimeException ? (or even better : create an ApiClientException)

cbornet: The message could be more explicit. Something like "the ApiClient was created with a built okClient so it's not possible to add an authorization interceptor to it".

tgerth: What is the advantage of creating an own ApiClientException comparing to the RuntimeException here in this case?

cbornet: See https://rules.sonarsource.com/java/RSPEC-112 . But we already have a bunch of RuntimeException so...

tgerth: Ah ok I see but as you said we already have many RuntimeExceptions so I will use a RuntimeException here if that's ok and then maybe in future you/we can change them to own Exceptions.

PR #6704, opened 2020-06-17
[Java][Native] added WithHttpInfo method for native-sync

bgong-mdsol:
### PR checklist

### Description of the PR
This PR added  **WithHttpInfo** method for **native** only, it includes the changes below:
- updated JavaClientCodegen
- updated the template of api
- updated the template of api-doc
- regenerated Petstore samples for native/asyncNative

@wing328 @cbornet @jeff9finger @bbdouglas @sreeshas @jfiala @lukoyanov @karismann @Zomzog @lwlee2608 @bkabrda @jfeltesse-mdsol @jcarres-mdsol


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaClientCodegen.java
```diff
@@ -404,6 +404,9 @@ public void processOpts() {
         } else if (NATIVE.equals(getLibrary())) {
             setJava8Mode(true);
             additionalProperties.put("java8", "true");
+            if (!asyncNative) {
+                supportingFiles.add(new SupportingFile("ApiResponse.mustache", invokerFolder, "ApiResponse.java"));
```
wing328, line 408: Would it be possible to also use ApiResponse in async operations to make it consistent?

bgong-mdsol: @wing328 This `ApiResponse` is used by `WithHttpInfo` method which is implemented for `sync` operation only. Currently `async` operation doesn't need `ApiResponse`. I can include `ApiResponse` in `async` for future use, does anyone need `WithHttpInfo` method  for `async` operation?

wing328: > does anyone need WithHttpInfo method for async operation?

For consistency, I suggest we implement it as well if it's not a lot of work (we're implemented the same in other client generators)

jfeltesse-mdsol: My concern is, do we know whether people are actually using it? I'd hate we spend time on this when we might have a case of YAGNI.
If it is indeed needed, would it be ok to do that in a separate PR to facilitate the scheduling of work?

wing328: Valid concern. That's why I asked if it's a lot of work.

Feel free to do it later with another PR instead.

wing328: And one more thing to add is that usually the `sync` operations in other clients simply call the `async` version of the operation and wait for completion.

PR #6706, opened 2020-06-18
[typescript-fetch] fix double parsing deep object parameters

Vrtak-CZ:
Fixing double parsing query parameters for deep object.

Currently it generates something like this:

```
...&filter=filter%255Bstatus%255D%3Denabled%26
```

It's because there is double parsing one of it is what i removing and another one here: https://github.com/OpenAPITools/openapi-generator/blob/ca3a23bfa8d2d46612d9370db830ec6751566d99/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L209-L226 but deep object can be handled by this one: https://github.com/OpenAPITools/openapi-generator/blob/ca3a23bfa8d2d46612d9370db830ec6751566d99/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L219-L221

### PR checklist

@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -122,12 +122,7 @@ export class {{classname}} extends runtime.BaseAPI {
             queryParameters['{{baseName}}'] = (requestParameters.{{paramName}} as any).toISOString().substr(0,10);
             {{/isDate}}
             {{^isDate}}
-            {{#vendorExtensions.x-codegen-isDeepObject}}
-            queryParameters['{{baseName}}'] = runtime.querystring(requestParameters.{{paramName}} as unknown as runtime.HTTPQuery, '{{baseName}}');
-            {{/vendorExtensions.x-codegen-isDeepObject}}
-            {{^vendorExtensions.x-codegen-isDeepObject}}
             queryParameters['{{baseName}}'] = requestParameters.{{paramName}};
-            {{/vendorExtensions.x-codegen-isDeepObject}}
             {{/isDate}}
             {{/isDateTime}}
         }


macjohnny:
this would revert https://github.com/OpenAPITools/openapi-generator/pull/6075
@haraldF can you help here?

haraldF:
@macjohnny thanks for the heads-up, I'll write some unit tests to make sure the feature doesn't regress, or break because of other side effects

haraldF:
@Vrtak-CZ thanks for the patch, but doesn't it mean that a deepobject parameter is now _never_ escaped after your change?

haraldF:
Looks like my change was done in 4.x and something must have changed in 5.x which changed behavior. I can reproduce the double escaping with today's master branch.

@macjohnny is there any way to test the _generated_ TypeScript code? E.g. do a `tsc --noEmit` or something?

So here's what I got so far, a dummy yaml:

```yaml
openapi: 3.0.0
info:
  title: Sample API
  description: Optional multiline or single-line description in [CommonMark](http://commonmark.org/help/) or HTML.
  version: 0.1.9
servers:
  - url: http://api.example.com/v1
    description: Optional server description, e.g. Main (production) server
  - url: http://staging-api.example.com
    description: Optional server description, e.g. Internal staging server for testing
paths:
  /users:
    get:
      summary: Returns a list of users.
      description: Optional extended description in CommonMark or HTML.
      parameters:
        - in: query
          name: someDeepObject
          style: deepObject
          required: true
          schema:
            type: object
      responses:
        '200':    # status code
          description: A JSON array of user names
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
```

Unfortunately, when I apply @Vrtak-CZ 's patch, I'm getting:

```
apis/DefaultApi.ts:48:13 - error TS2322: Type 'Foo' is not assignable to type 'string | number | boolean | HTTPQuery | (string | number | boolean | null)[] | null'.
  Type 'Foo' is not assignable to type 'HTTPQuery'.
    Index signature is missing in type 'Foo'.

48             queryParameters['someDeepObject'] = requestParameters.someDeepObject;
```

e.g. we cannot just assign an object to `queryParameters`. When I 'hack' it to convert to any, it works.

So let's gather some opinions - fix it by extending `runtime.HttpQuery` or convert deep objects to `any`?

Vrtak-CZ:
@haraldF thanks for testing.

> e.g. we cannot just assign an object to `queryParameters`. When I 'hack' it to convert to any, it works.

that's what we do too because of this. But I do not know better solution and also there is already handler for object inside of `querystring` function.

haraldF:
> @haraldF thanks for testing.
> 
> > e.g. we cannot just assign an object to `queryParameters`. When I 'hack' it to convert to any, it works.
> 
> that's what we do too because of this. But I do not know better solution and also there is already handler for object inside of `querystring` function.

Thanks for the info and sorry for the delay. I did some more digging and the `as any` is already fixed in d9a6f4d7261

haraldF:
@Vrtak-CZ I believe it's better to revert the entire change, also removing the unnecessary extra flag from the model. I've created a new PR: https://github.com/OpenAPITools/openapi-generator/pull/6860

macjohnny:
closing this one in favor of https://github.com/OpenAPITools/openapi-generator/pull/6860

Issue #6711, opened 2020-06-18
[BUG] [Server: Python] Incorrect parameter example used for requestBody parameters 


pavlot:
#### Bug Report Checklist

- [ x ] Have you provided a full/minimal spec to reproduce the issue?
- [ x ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ x ] Have you search for related issues/PRs?
- [ x ] What's the actual output vs expected output?

##### Description

As mentioned incorrect example taken for requestBody parameters when generated code by `python-flask` generator

##### openapi-generator version

v4.3.1

##### OpenAPI declaration file content or url

```yaml
---
openapi: 3.0.2
info:
  title: OpenApiGeneratorTest
  version: 1.0.0
paths:
  /example:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: array
              items:
                $ref: '#/components/schemas/ExampleComplexType'
            examples:
              Random request data:
                value:
                - id: 19
                  text: some text
                - id: 63
                  text: some text
      responses:
        "200":
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/ExampleComplexType'
              examples:
                Random results:
                  value:
                  - id: 50
                    text: some text
                  - id: 3
                    text: some text
          description: Responce description text
components:
  schemas:
    ExampleComplexType:
      title: Root Type for ExampleComplexType
      description: ""
      type: object
      properties:
        id:
          format: int32
          type: integer
        text:
          type: string
      example:
        id: 0
        text: Some text
```
**expected**: python test must contain example like:
```json
[{"id":19, "text":"some text"},{"id":63, "text":"some text"}]
```
**Actual**
```json
{"id":0, "text":"some text"}
```

##### Command line used for generation

openapi-generator generate -g python-flask -i OpenApiGeneratorTest.yaml

##### Steps to reproduce

1. Generate python-flask code using given commandline
2. Check test_default_controller.py for example used in `test_example_post` test
3. Rut `tox` - it will be failed with message
```
E   AssertionError: 400 != 200 : Response body is : {
E     "detail": "{'id': 0, 'text': 'Some text'} is not of type 'array'",
E     "status": 400,
E     "title": "Bad Request",
E     "type": "about:blank"
E   }
```
##### Related issues/PRs

Not found

##### Suggest a fix

[...]

PR #6720, opened 2020-06-20
Fix for incorrect examples used by Flask codegenerator. Issue #6711

pavlot:
Guys, can you please review proposed changes
@taxpon  @frol  @mbohlool  @cbornet  @kenjones-cisco  @tomplus  @Jyhess  @arun-nalla  @spacether 

Fix is relatively simple: just to avoid re-setting parameter example in case if it is already set in 
modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java:966.

The rest of PR - test which reproduces issue and proof the fix

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
@@ -963,7 +963,9 @@ public class PythonAbstractConnexionServerCodegen extends DefaultCodegen impleme
             }
             if (operation.requestBodyExamples != null) {
                 for (Map<String, String> example : operation.requestBodyExamples) {
-                    if (example.get("contentType") != null && example.get("contentType").equals("application/json")) {
+                    if (operation.bodyParam.example == null
+                    		&& example.get("contentType") != null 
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonAbstractConnexionServerCodegen.java
```diff
@@ -963,7 +963,9 @@ public void postProcessParameter(CodegenParameter parameter) {
             }
             if (operation.requestBodyExamples != null) {
                 for (Map<String, String> example : operation.requestBodyExamples) {
-                    if (example.get("contentType") != null && example.get("contentType").equals("application/json")) {
+                    if (operation.bodyParam.example == null
+                    		&& example.get("contentType") != null 
+                    		&& example.get("contentType").equals("application/json")) {
```
wing328, line 968: FYI. You can do `"application/json".equals(example.get("contentType"))` instead to avoid the null check.

pavlot: Good point.
It was a part of code done here before me, I just add line breaks.

My changes related to the issue - check whether `operation.bodyParam.example` is not set, and do nothing if it is already there

PR #6730, opened 2020-06-20
[dart] fix toJson does not handle complex type

### LINKED ISSUE #4562: [BUG] Dart mustache model toJson doesn't handle the situation which the property is object or array
https://github.com/OpenAPITools/openapi-generator/blob/96bbab98a47c702c9b05f7f2bf06f485071dbd23/modules/openapi-generator/src/main/resources/dart2/class.mustache#L86

If there is a property is an object or an array, it dose not handle this situation.

agilob:
Rebase of #5189 that fixes #4562

### PR checklist

@ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @nickmeinhold (2019/09) @athornz (2019/12) @amondnet (2019/12)


sbu-WBT:
Hi, I tested this PR with this https://github.com/geoDavey/osrm-openapi/blob/master/osrm-openapi.yaml OpenAPI specification.
The thing is, I can't find any change between before and after the PR.

For Example:
`List<Waypoint> destinations = [];` is defined in the class `TableResponse`.

But in the `toJson()` method you can still find this:
```dart
    if (destinations != null)
      json['destinations'] = destinations;
```

The same problem exists also with objects without lists, if you look inside the annotation.dart.

It would be very nice if this could get fixed, because this is a very big problem for serialization. Thank you

modules/openapi-generator/src/main/resources/dart/class.mustache
```diff
@@ -58,14 +58,24 @@ class {{classname}} {
...
+          {{#isListContainer}}
+            '{{baseName}}': {{name}} == null ? null : {{name}}.map((item) => item.toJson()).toList(){{^-last}},{{/-last}}
+          {{/isListContainer}}
+          {{^isListContainer}}
+            '{{baseName}}': {{name}} == null ? null : {{name}}.toJson(){{^-last}},{{/-last}}
+          {{/isListContainer}}
+        {{/complexType}}
+        {{^complexType}}
+          '{{baseName}}': {{name}}{{^-last}},{{/-last}}
+        {{/complexType}}
```
wing328, line 78: @agilob Can we keep the indentation as 2-space? It would make the code review a lot easier.

agilob: This MR will cause a bunch of conflicts with 2 other MR I have in progress and it's not my changes, I resurrected commit from 
@arndt-s. I will take care of it as the last thing if you dont mind?

agilob:
>It would be very nice if this could get fixed, because this is a very big problem for serialization. Thank you

Totally agree, but as per comment above, I think this should be solved after more fundamental problems are solved in dart generation, like basic enums, non-nullable fields. Please also note these aren't my changes. I picked old MR, fixed git conflict and reopened it.

modules/openapi-generator/src/main/resources/dart/class.mustache
wing328: Agreed with taking care of the code indentation in another PR.

PR #6751, opened 2020-06-23
Dart - generate constructor with named params

agilob:
### PR checklist

@ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @athornz (2019/12) @amondnet (2019/12)

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart2/class.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/class.mustache
@@ -6,7 +6,12 @@ class {{classname}} {
   {{#min}} // range from {{min}} to {{max}}{{/min}}//{{^min}}enum {{name}}Enum { {{#values}} {{.}}, {{/values}} };{{/min}}{
   {{/allowableValues}}
   {{/vars}}
-  {{classname}}();
+
+  {{classname}}({
+  {{#vars}}
+    this.{{name}},
+  {{/vars}}
+  });
 
   @override
   String toString() {
--- a/samples/client/petstore/dart2/petstore/test/pet_test.dart
+++ b/samples/client/petstore/dart2/petstore/test/pet_test.dart
@@ -25,11 +25,12 @@ void main() {
         ..id = 124321
         ..name = 'Jose'
     ];
-    return Pet()
-      ..id = id
-      ..category = category
-      ..tags = tags
-      ..name = name
+    return Pet(
+      id : id,
+      category: category,
+      tags: tags,
+      name: name,
+    )
       ..status = status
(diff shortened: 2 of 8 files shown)


modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -6,7 +6,12 @@ class {{classname}} {
   {{#min}} // range from {{min}} to {{max}}{{/min}}//{{^min}}enum {{name}}Enum { {{#values}} {{.}}, {{/values}} };{{/min}}{
   {{/allowableValues}}
   {{/vars}}
-  {{classname}}();
```
wing328, line 9: @agilob I'm no expert in Dart but shall we keep the default constructor (no argument) for backward compatibility?

(upcoming 5.0.0 allows breaking changes without fallbacks but it's still worth considering backward compatibility)

agilob: Sure, happy to explain this feature. Dart has flexible syntax on method params, default values, and mandatory params.

Method/constructor params can be mandatory (as not named) or named (optional) or named AND mandatory. This change introduces constructors that have named optional parameters. Such parameters have input params inside `{}` brackets. Since all fields in these classes are public anyway (dart discourages use of setters and getters "just because"  like in java), so we have this syntax, of all valid constructors.

Constructors with all optional params are backwards compatible, it's really just programmers' preference
image


This MR introduces optional named params in constructors, while cascade operators is already supported, it's fully backwards compatible.
https://dart.dev/guides/language/language-tour#optional-parameters
https://dart.dev/guides/language/language-tour#cascade-notation-

>(upcoming 5.0.0 allows breaking changes without fallbacks but it's still worth considering backward compatibility)

I am myself a guardian of backwards compatibility ;)

This MR 6729 breaks backwards compatibility when creating enums (which now are just string without any validation), so I found it necessary.

wing328: > I am myself a guardian of backwards compatibility ;)

👍 

Thanks for the explanation. The change looks good to me. 

If no further feedback or question, I'll merge it over the weekend.

Mike278:
> Constructors with all optional params are backwards compatible

not for properties with default values

for example:
```
      "Foo": {
        "required": [
          "bar"
        ],
        "properties": {
          "bar": {
            "type": "string",
            "default": "some default value"
          }
        }
      },
```
generates
```dart

class Foo {
  
  String bar = "some default value";

  Foo({
    this.bar,
  });

  // ... 
}

print(Foo().bar); // prints null, used to print 'some default value'
```

agilob:
@Mike278 this is fixed now on master

PR #6753, opened 2020-06-23
[Rust][Client] Unify sync/async client structure

bcourtine:
Unify structure:

- Remove xxxApiClient structures and traits.
- Configuration is now passed as function argument (instead of being part of the client struct).

This evolution makes much easier the sync/async switch since structure and signatures are now much closer.

cc @frol (2017/07) @farcaller (2017/08) @richardwhiuk (2019/07) @paladinzh (2020/05) @wing328

Also cc @HenningHolmDE @seunlanlege since you are currently the main rust generator users.


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
```diff
@@ -301,9 +301,6 @@ public void processOpts() {
         if (HYPER_LIBRARY.equals(getLibrary())) {
             supportingFiles.add(new SupportingFile("request.rs", apiFolder, "request.rs"));
         }
```
HenningHolmDE, line 303: I'm afraid your change breaks `library: hyper`, as can easily be checked like this:
```
java -jar ./modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -g rust -i https://github.com/slackapi/slack-api-specs/raw/master/web-api/slack_web_openapi_v2.json -o slack --skip-validate-spec --remove-operation-id-prefix --library hyper
cargo check --manifest-path slack/Cargo.toml
> ...
> error[E0583]: file not found for module `client`
>    --> src\apis\mod.rs:162:1
>     |
> 162 | pub mod client;
>     | ^^^^^^^^^^^^^^^
>     |
>     = help: to create the module `client`, create file "src\apis\client.rs"
```
I would suggest something like this as an alternative for removing the line completely:

```suggestion
        if (HYPER_LIBRARY.equals(getLibrary())) {
            supportingFiles.add(new SupportingFile("request.rs", apiFolder, "request.rs"));
            supportingFiles.add(new SupportingFile(getLibrary() + "/client.mustache", apiFolder, "client.rs"));
        }
```

Maybe a test using `library: hyper` would be helpful to prevent something like this in the future?

HenningHolmDE:
@bcourtine I really like the unification! It makes changing between sync and async depending on your needs much easier. :+1: 

I tested a few generator settings (sync/async, single/multi param) with my project and besides from `library: hyper` (as mentioned above) everything seems to work fine for me.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
bcourtine: Good catch! I didn't think that client file was shared between hyper and reqwest. Will fix that. :+1:

seunlanlege:
is it possible to add the schema documentation to the methods?

bcourtine:
@seunlanlege I think this is not related to current PR…

It can eventually be done in another PR, but I don't understand what you mean by "add the schema to the methods". Would you have an example?

seunlanlege:
oh I meant the API documentation for the methods.

bcourtine:
@seunlanlege Since it was not a big deal, I added API method comments in current PR. You can try it out.

seunlanlege:
tested and works as advertised :pray:

PR #6760, opened 2020-06-24
[PHP] Better handling of invalid data (array)

Herrick19:
If the $data is an invalid Json (for example, add an extra comma after the last property) or if data is not an array, php gives error:

Invalid argument supplied for foreach() at line 257 (Now line is 262)

This is my fist git pull request so please pardon me if I did something wrong.

Best Regards

### DIFF
--- a/samples/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php
+++ b/samples/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php
@@ -262,6 +262,11 @@ class ObjectSerializer
             return null;
         } elseif (strcasecmp(substr($class, -2), '[]') === 0) {
             $data = is_string($data) ? json_decode($data) : $data;
+            
+            if (!is_array($data)) {
+                throw new \InvalidArgumentException("Invalid array '$class'");
+            }
+            
             $subClass = substr($class, 0, -2);
             $values = [];
             foreach ($data as $key => $value) {
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -252,6 +252,11 @@ class ObjectSerializer
             return null;
         } elseif (strcasecmp(substr($class, -2), '[]') === 0) {
             $data = is_string($data) ? json_decode($data) : $data;
+            
+            if (!is_array($data)) {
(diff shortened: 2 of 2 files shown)


ybelenko:
cc PHP technical committee @jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @renepardon

wing328:
@Herrick19 thanks for the PR. I've updated the samples via https://github.com/OpenAPITools/openapi-generator/pull/6760/commits/ede2150903f9aacaf8b6a1391492e8e7179fb163. Will merge if all the tests pass.

samples/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php
```diff
@@ -262,6 +262,11 @@ public static function deserialize($data, $class, $httpHeaders = null)
             return null;
         } elseif (strcasecmp(substr($class, -2), '[]') === 0) {
             $data = is_string($data) ? json_decode($data) : $data;
+            
+            if (!is_array($data)) {
```
ybelenko, line 266: I guess that it won't work in 99% cases because `json_decode` without second argument `$assoc = true` returns `StdClass` which isn't array. And my brief test confirms that:
```php
$data = '{"a":1,"b":2,"c":3,"d":4,"e":5}';
$data = is_string($data) ? json_decode($data) : $data;

if (!is_array($data)) {
    throw new \InvalidArgumentException("Invalid array '$class'");
}
// throws exception because StdClass isn't array
```

Herrick19: Hi,

You are right about the fact that json_decode would return a StdClass in that scenario. Since this code is only executed when you have declared your model as an array using "[]", if such case would happen, it would be ok to send an InvalidArgumentException in my opinion. Since you have declared your model as an array and the result is an object, it's just normal to receive an InvalidArgumentException... From my understanding, there is no such things as an associative array in JSON, it's just an object. So someone declaring the value as an array with "[]" and use an object object should receive this error ?

Perhaps I'm missing something in the evaluation.

ybelenko: Oh, now I see. Data should look like `["foo", "bar", "hello world"]` for `StringModel[]` class definition to pass this part of code without exception.

Issue #6522, opened 2020-06-02
PHP 7.1 end of life


ybelenko:
PHP 7.1 reached end of life at 1 Dec 2019

> End of life - A release that is no longer supported. Users of this release should upgrade as soon as possible, as they may be exposed to unpatched security vulnerabilities.

### [PHP: Supported Versions](https://www.php.net/supported-versions.php):
- 7.2 until 30 Nov 2020
- 7.3 until 6 Dec 2021
- 7.4 until 28 Nov 2022

At seems that we need to change minimum PHP version to 7.2 in all php generators.

### Current Requirements:
| Generator | PHP version |
|:---------:|:------------:|
| php client | >= 7.1                |
| Laravel server | >= 7.1.3      |
| Lumen server | >= 7.1.3        |
| Slim4 server | >= 7.1             |
| Symfony server | >= 7.1.3     |
| Ze-ph server | >= 7.1             |

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @renepardon

ybelenko:
@Articus I've seen your good work at #1902

Can you help with upgrade `php-ze-ph` server?
I would do it myself, but there are no unit tests in `php-ze-ph` generator.
We need to set PHP version to `^7.2` in `composer.json`:
https://github.com/OpenAPITools/openapi-generator/blob/a4e42e1651b9f0f70ee56d40051803a73ac3c6fa/modules/openapi-generator/src/main/resources/php-ze-ph/composer.json.mustache#L8

and also here https://github.com/OpenAPITools/openapi-generator/blob/a4e42e1651b9f0f70ee56d40051803a73ac3c6fa/modules/openapi-generator/src/main/resources/php-ze-ph/README.md.mustache#L8

Packagist says that `zendframework/zend-expressive` package and few others are abandoned. We need to replace them with suggested packages.

Articus:
Sure, but it would be better to move `zendframework/zend-expressive` replacement to another issue.

It is really simple to bump minimal PHP version to 7.2 - I am like 99%-sure that everything is still working, but it will be dutiful to run tests against current minor version of PHP 7.2. Can do that before next weekend.

Replacement of `zendframework/zend-expressive` is way more complex problem. All packages in `zendframework` namespace were renamed - https://getlaminas.org/blog/2020-03-09-transferring-zf-to-laminas.html (for example, Zend Expressive is now Laminas Mezzio). So it will require:
[...]

PR #6763, opened 2020-06-24
[php-ze-ph] Set required PHP version to ^7.2

ybelenko:
* Composer required PHP version set to ^7.2
* Readme PHP version set to >=7.2
* Input spec for samples generation changed to OAS3

It's weird that `samples/server/petstore/php-ze-ph/src/App/ErrorMiddleware.php` file has not been regenerated. This file isn't presented in generator templates folder `modules/openapi-generator/src/main/resources/php-ze-ph`.
Related #6522 

### PR checklist

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @renepardon
thanks to @Articus for help with testing


bin/configs/php-ze-ph.yaml
```diff
@@ -1,4 +1,4 @@
 generatorName: php-ze-ph
 outputDir: samples/server/petstore/php-ze-ph
-inputSpec: modules/openapi-generator/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
+inputSpec: modules/openapi-generator/src/test/resources/3_0/petstore.yaml
```
wing328, line 3: I assume fake petstore cover edge cases not yet handled by this generator, which is ok.

In the medium, we plan to switch from petstore to something else.

ybelenko: Yeah, I wanted to ask which 3.0 spec we should use from now, because I noticed links to different 3.0 files across generators.

ybelenko:
I would like to ask a review from authors of Zend Expressive via Github issues, but I can't find current main repo.
@Articus I guess you know this framework better, can you help me a bit? Was it renamed to Laminas? If so why it's suggest to use [mezzio/mezzio](https://github.com/mezzio/mezzio) package as substitution?

Articus:
You are correct - [Laminas project](https://getlaminas.org) is a new name for all packages previously known as parts of Zend Framework. [Here](https://framework.zend.com/blog/2020-01-24-laminas-launch) is full list of renamed organizations for packages. I am not sure about exact reasoning behind those renames though. They could use Laminas Expressive, but ended up with new name Mezzio instead.

weierophinney:
@Articus and @ybelenko —

The equivalent to Zend Expressive is indeed [Mezzio](https://docs.mezzio.dev). The reason for the renames:

- We transferred Zend Framework to the Linux Foundation. Since the company Zend still exists, the project as a whole needed a rename.
- Apigility and Expressive had existing trademarks, and the logistics and costs of transferring those were quite onerous for all parties. As such, those subprojects needed to be renamed as well. We chose to rename Apigility to Laminas API Tools as we didn't want to register a unique name for it when its future is a bit uncertain (due to the move to do API tooling within Expressive/Mezzio). We chose Mezzio as the new name for Expressive as it (a) is a word that does not exist in any other language, (b) has an affinity to the Italian "Mezzo", which in turn (c) has a meaning of "middle" or "between", which is reminiscent of Middleware.

I'd be happy to help in updating the php-ze-ph generator to Mezzio; could somebody point me to the relevant areas of code?

Articus:
@weierophinney, nice to see you here!
The situation with migration to Mezzio is a bit tricky, its summary is [in this issue](https://github.com/OpenAPITools/openapi-generator/issues/6522#issuecomment-643822732). 

The hardest step is complete - latest versions of [DataTransfer](https://github.com/Articus/DataTransfer/releases/tag/0.4) and [PathHandler](https://github.com/Articus/PathHandler/releases/tag/0.6) are using Laminas packages. And for last two weeks I was slowly updating `php-ze-ph` generator itself. 

> could somebody point me to the relevant areas of code?

The main bits are [generator class](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PhpZendExpressivePathHandlerServerCodegen.java) and [generator templates](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/php-ze-ph). You may also want to check [a sample of generated code](https://github.com/OpenAPITools/openapi-generator/tree/master/samples/server/petstore/php-ze-ph). It is quite different from standard Zend Expressive/Mezzio skeleton. Your review will be greatly appreciated :)

PR #6769, opened 2020-06-24
Update ObjectSerializer.mustache

Herrick19:
When we deserialize a json structure, only the first level of validation is done and accessible with "listInvalidProperties"

To validate all sub objects, everyone would need to loop again on the objects and follow all properties, check if they are objects and call listInvalidProperties on each of them.

A better approach would be to have the "listInvalidProperties" shared between parent and childs, but this would be a major structural change so this easy fix seems like a nice compromise.


ybelenko:
@Herrick19 Why did you decide to close this PR?

Herrick19:
Hi,

I Changed my mind and thought a separate function would be better.

So if you check this pr: https://github.com/OpenAPITools/openapi-generator/pull/6770

It does the same thing but in a better way IMO

Best Regards

Issue #5467, opened 2020-02-27
[BUG][typescript-angular] generator generates invalid TypeScript for polymorphic types


t03apt:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

[typescript-angular] generator generates invalid TypeScript for polymorphic types.
For example:
```[TypeScript]
import { Circle } from '../model/models';
import { Rectangle } from '../model/models';
import { Rectangle | Circle } from '../model/models';
```

Expected:
```[TypeScript]
import { Rectangle, Circle } from '../model/models';
```

##### openapi-generator version

4.2.3

##### How to reproduce
> openapi-generator generate -i https://gist.githubusercontent.com/t03apt/71391e24045ef08dcad11bfae12037ae/raw/9c83bd41be70063d1436f84c842253018a6c8c46/Swashbuckle.AspNetCore.Test.WebSites.Basic.swagger.json -o ./generated -g typescript-angular --skip-validate-spec

macjohnny:
related: https://github.com/OpenAPITools/openapi-generator/issues/4130#issuecomment-591458888

szilagyinorbi:
any update on this?

**AbstractTypeScriptClientCodegen.java**
```
private Set<String> parseImports(CodegenModel cm) {
   Set<String> newImports = new HashSet<String>();
   if (cm.imports.size() > 0) {
      for (String name : cm.imports) {
         if (name.indexOf(" | ") >= 0) {
            String[] parts = name.split(" \\| ");
            for (String s : parts) {
               newImports.add(s);
            }
         } else {
               newImports.add(name);
         }
      }
   }
   return newImports;
}

private List<Map<String, String>> toTsImports(CodegenModel cm, Set<String> imports) {
   List<Map<String, String>> tsImports = new ArrayList<>();
   for (String im : imports) {
      if (!im.equals(cm.classname)) {
         HashMap<String, String> tsImport = new HashMap<>();
[...]

PR #6789, opened 2020-06-26
Fix Union Types Import Issue

FrankEssenberger:
This is an attempt to fix the bug reported here: #5467 
The problem was that the union types in typescrit e.g. `TypeA | TypeB` represented by the `oneOf` keyword in the service definition lead to not valid code:
```
import{TypeA | TypeB} from '../models'
```
you can either use a comma in the import or split into separate imports. I did the latter one because I thought this could be more solid for models in different files and paths potentially. If they are not all exported from the same path.

I adjusted it for three typescript generators:
- axios (this I where I was effected)
- node
- angular

Then I stopped and thought it is a good point to start the discussion before I adjust it for all the other Typescript generators. 

Since I am new I would highly appreciate some help regarding the following issues:
- [ ] How to run the code formatter. When I run `mvn formatter:format` I get hundreds of code changes.
- [ ] How to run the tests. Is `mvn test` fine? 
- [ ] Not sure about the locations where I put the tests and so on. Perhaps this kind of test has to go somewhere else.

### PR checklist

~~If contributing template-only or documentation-only changes which will change sample output, [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) beforehand.~~


FrankEssenberger:

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -221,6 +224,65 @@ public void processOpts() {
         }
     }
 
+    @Override
+    public String toModelImport( String name){
```
macjohnny, line 228: please add some docs to describe what this does, with an example

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -221,6 +224,65 @@ public void processOpts() {
...
+            List<String> names = Arrays.asList(name.split("\\|"));
+           return names.stream()
+                   .map(withSpace->withSpace.replaceAll(" ",""))
+                   .map(noSpace->super.toModelImport(noSpace))
+                   .collect(Collectors.joining("|"));
+        }
+        return super.toModelImport(name);
+    }
+
+    protected String toModelImportForUnionTypes(String name, Function<String,String> toModelImportSingle){
```
macjohnny, line 239: docs

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -221,6 +224,65 @@ public void processOpts() {
...
+    public Map<String, Object> postProcessOperationsWithModels(final Map<String, Object> objs, final List<Object> allModels) {
+        Map<String, Object> objsUnderProcess=  super.postProcessOperationsWithModels(objs, allModels);
+
+        if(Boolean.valueOf(objsUnderProcess.get("hasImport").toString())){
+            objsUnderProcess.put("imports", splitUniontypeImports((List<Map<String,String>>) objsUnderProcess.get("imports")));
+        }
+        return objsUnderProcess;
+    }
+
+    protected List<Object> splitUniontypeImports(List<Map<String,String>> imports){
```
macjohnny, line 260: @amakhrov @wing328  do you think such a complicated logic is necessary, or could this be solved elsewhere?

amakhrov: It looks quite similar to the existing logic in TypeScriptAngularClientCodegen (see `postProcessAllModels` and `parseImports`)

Maybe we should lift the logic from AngularClient to AbstractTypescriptClient instead of pretty much repeating it?

The root problem, as I see it, is that `toModelImport` takes a string as an argument rather than a model and returns a string instead of a list. It makes it very limited in terms of handling composite models. But changing that would require a change in Generator and CodegenConfig interfaces. And generally means more work, of course.

macjohnny:
@FrankEssenberger thanks for your contribution!

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
FrankEssenberger: I added the doc to the new method `toModelImportMap`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
FrankEssenberger: removed the method.

FrankEssenberger:
I adjusted my approach and created a second method `Map<String,String> toModelImportMap(String input)` which will give the map between the fully quantified model for import and the initial name. For most cases this map will have only a single pair. However, for the typescript case if a type like "classA | classB" is found it will have two entries. 

This seemed much cleaner then to fiddle around with the imports in a post processing step.

FrankEssenberger:
I would like to run the formatter so that the diff looks better. Please let me know how to execute this and I am also happy for further feedback.

wing328:
cc @OpenAPITools/generator-core-team as the change impacts default codegen, generator

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -17,6 +17,8 @@
 
 package org.openapitools.codegen;
 
+import com.google.common.collect.Lists;
```
TiFu, line 20: Is `google.common.colllect` preferred over `java.util.*` in OAI generator?

FrankEssenberger: Just an old habit from other projects. I used the basic java methods to create new Lists and Maps.

macjohnny:
@FrankEssenberger could you please merge the current master and re-generate the samples?

FrankEssenberger:
I have done a `./bin/generate-samples.sh` and merged master. However, there were a lot of code changes in the samples which I commited. Not sure if this is correct?

macjohnny:
@FrankEssenberger did you
- merge master
- `mvn clean package`
- re-generate the samples
?

FrankEssenberger:
yes this is what I did. If I do this on the not forked repo I do not get the changes in the samples. Also the changes are reordering of dependencies for example:

```
#include "ApiResponse.h"	
#include "HttpContent.h"	
#include "Pet.h"	#include "Pet.h"
#include <cpprest/details/basic_types.h>
```
changes to
```
#include "Pet.h"	#include "Pet.h"
#include <cpprest/details/basic_types.h>
#include "ApiResponse.h"	
#include "HttpContent.h"	
```

For whatever reason. Is there a formatter task present?

macjohnny:
Could it be your IDE that performs code cleanup when committing?

FrankEssenberger:
ah, I just realized that with me map introduction I changed the order of imports. Most likely the hashmap has not a alphabetic ordering when I loop over the keys. I will fix this that the imports are order alphabetically.

FrankEssenberger:
ok, now only the typescript related samples have changed which is plausible.

FrankEssenberger:
looked at the failing tests and the missing imports are not correct. So I have broken something with my changes so that the for example the `Tag` and `Category` are not there anymore. But I have a local test running so I can fix this issue quickly I hope.

FrankEssenberger:
so I reverted a change to the angular typescript generator. It seems that the imports come from two places for api and model files. I did not expect that.

FrankEssenberger:
hm, I do not find good information on the failed test. All I see in the step is:
```
2020-07-12 11:31:33.713:INFO:oejs.AbstractNCSARequestLog:main: Opened /var/log/2020_07_12-requests.log
2020-07-12 11:31:33.743:INFO:oejs.ServerConnector:main: Started ServerConnector@1e8b7643{HTTP/1.1}{0.0.0.0:8080}
2020-07-12 11:31:33.745:INFO:oejs.Server:main: Started @5914ms

Exited with code exit status 1
CircleCI received exit code 1
```
which is rather unspecific. Are these tests known to be flaky? Should I trigger a rerun?

macjohnny:
@FrankEssenberger I re-triggered the CI build, let's see

macjohnny:
@FrankEssenberger can you merge the most recent master, again, please?

FrankEssenberger:
Sure. Actually I planned on the weekend to do something. I wanted to comment out the tests I added to see if they cause problems.

FrankEssenberger:
So it turned out that test should have the word `Test` somewhere in their name, I think this is some surefire default and that it is not a good idea to create files in a test and not clean them up. Since the tests are green now, it would be great to get finally the review done.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -221,6 +224,38 @@ public void processOpts() {
...
+    @Override
+    public Map<String,String> toModelImportMap( String name){
+        if(name.contains("|")){
+            List<String> names = Arrays.asList(name.replace(" ","").split("\\|"));
+            Map<String,String> result = Maps.newHashMap();
+            names.forEach(s->result.put(toModelImport(s),s));
+            return  result;
+        }
+        return Collections.singletonMap(toModelImport(name),name);
+    }
```
amakhrov, line 256: These two methods (toModelImportMap and toModelImport) look very similar (split by `|`, remove spaces, process each part). Should one method just delegate to the other? E.g. `toModelImport` could call `toModelImportMap` and then collect the resulting map back into a string.

FrankEssenberger: sure good point. I introduced small helper methods with a more descriptive name and used the map method inside the single entry one.

FrankEssenberger:
The pipeline seems a bit flaky. I see in the ci build the error: 
```
[error] sbt.librarymanagement.ResolveException: Error downloading com.softwaremill.sttp.client:json4s_2.13:2.2.0
[error]   Not found
[error]   Not found
[error]   not found: /home/circleci/.ivy2/local/com.softwaremill.sttp.client/json4s_2.13/2.2.0/ivys/ivy.xml
[error]   download error: Caught java.net.ConnectException: Connection timed out (Connection timed out) (Connection timed out (Connection timed out)) while downloading https://repo1.maven.org/maven2/com/softwaremill/sttp/client/json4s_2.13/2.2.0/json4s_2.13-2.2.0.pom
[error] 	at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:249)
```

FrankEssenberger:
And for the travis-ci I see the errorr:
```
 1) Isomorphic Fetch
       GET-Request:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/OpenAPITools/openapi-generator/samples/openapi3/client/petstore/typescript/tests/default/test/http/isomorphic-fetch.test.ts)
```
However, if I run the test locally it runs through so perhaps the timeout is reached for some strange reason in the pipeline.

macjohnny:
@FrankEssenberger I re-started the CI builds.

FrankEssenberger:
Cool, your restart helped. You have the magic green finger :-).

macjohnny:
cc @OpenAPITools/generator-core-team for final approval, as the change impacts default codegen, generator

FrankEssenberger:
merged master. Do you need something from my side?

macjohnny:
@FrankEssenberger we need to wait for a final review a a member of the @OpenAPITools/generator-core-team

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -1115,6 +1099,40 @@ private static String generateParameterId(Parameter parameter) {
         return operations;
     }
 
+    private Map<String,String> getAllImportsMapppings(Set<String> allImports){
```
wing328, line 1102: Can you please add docstrings to explain what these new (private) functions do (e.g. avoid duplicates)?

We want to document all methods in the default codegen/generator class to make code easier to understand for new contributors. (I do not expect you to document all methods missing the docstring. The community can help with separate PRs)

FrankEssenberger: I added some doc and used a Set to better represent that the entries are unique for the other method.

macjohnny:
@wing328 can we merge this?

FelschR:
`typescript-fetch` has the same issue. Would be nice to fix it there as well.

FrankEssenberger:
Yes, I put the change in the `AbstractTypeScriptClientCodegen.java` which should be the parent for all typescript generators I guess. In the test I piked three axios, node and angular and for all the union types were correctly handled.

macjohnny:
@FrankEssenberger thanks again for your contribution.
I hope we can merge it soon

wing328 (approved):
LGTM. Thanks for the PR.

PR #6822, opened 2020-06-30
[go-experimental] Fix oneOf generation of primitive types

zippolyte:
Given the following spec 
```yaml
components:
  schemas:
    Something:
      type: object
      oneOf:
        - $ref: '#/components/schemas/One'
        - $ref: '#/components/schemas/Two'
        - $ref: '#/components/schemas/Three'
    One:
      type: string
    Two:
      type: integer
    Three:
      type: object
      properties:
        foo:
          type: integer
```
The go-experimental generator would generate the following struct

```go
type Something struct {
        string *string
        int32  *int32
        Three  *Three
}
```

This fails because the field names `int32` and `string` are reserved keywords.

I fixed it by Capitalizing the field name, but let me know if you think there is a better way to do this.

### PR checklist


zippolyte:
cc @antihax (2017/11) @grokify (2018/07) @kemokemo (2018/09) @bkabrda (2019/07)

sebastien-rosset:
There is a way to specify that some string values are keywords:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java#L63

That list is used by [escapeReservedWord](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/GoClientExperimentalCodegen.java#L125) to escape words that are reserved keywords.

zippolyte:
Yeah in the openAPI model, the `oneOf` attribute is an array that contains the list of types composing the `oneOf`, so it's on purpose that it doesn't go through that escaping function I believe.

I didn't want to go into the generator code to change the openAPI model, so I thought of this solution instead.

It also has the benefit of making all fields of the `oneOf` struct exported by capitalizing, so it's more consistent too.

zippolyte:
apparently it's valid go now: https://play.golang.org/p/7BypWiBfakf

PR #6901, opened 2020-07-09
Correct allOf with only one child schema (no discriminator)

wing328:
Correct allOf with only one child schema (no discriminator) by removing the old (incorrect) behavior. 

Now it works correctly as model composition.

### PR checklist


m-mohr:
How to do enable inheritance without discriminator?

wing328:
Don't think you can do that according to the spec

> The OpenAPI Specification allows combining and extending model definitions using the allOf property of JSON Schema, in effect offering model composition. allOf takes an array of object definitions that are validated independently but together compose a single object.
>
> While composition offers model extensibility, it does not imply a hierarchy between the models. To support polymorphism, the OpenAPI Specification adds the discriminator field

Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#composition-and-inheritance-polymorphism

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1239,8 +1237,10 @@ public static String getParentName(ComposedSchema composedSchema, Map<String, Sc
...
                         // not a parent since discriminator.propertyName is not set
-                        hasAmbiguousParents = true;
-                        refedWithoutDiscriminator.add(parentName);
+                        // TOOD to be removed in 6.x release
+                        LOGGER.warn("[deprecated] inheritance without use of 'discriminator.propertyName' has been deprecated" +
+                                        " in the 5.x release. Composed schema name: {}. Title: {}",
+                                composedSchema.getName(), composedSchema.getTitle()); 
                     }
                 } else {
                     // not a ref, doing nothing, except counting the number of times the 'null' type
```
jeff9finger, line 1246: This block is no longer needed at all right? It only increments an unused variable `nullSchemaChildrenCount`

wing328:
@jeff9finger I've pushed some update to minimize the change and only fix the incorrect behavior. Please have another look when you've time.

typhoon2k:
Hi @wing328,

as far as I understand the only change applied here is change in log message content (and log level).
We have a very simple case:

base_schemas.yaml
```yaml
baseSchema:
  title: BaseSchema
  type: object
  description: Base schema description
  properties:
    property1:
      type: string
    property2:
      type: string
```

extension_schemas.yaml:
```yaml
extensionSchema
  title: ExtensionSchema
  type: object
  description: Extension schema description
  allOf:
    - $ref: 'base_schemas.yaml#/baseSchema'
  example:
    property1: 'value1'
    property2: 'value2'
```

With this definition (we have only one child schema here) we are still getting inheritance instead of composition and generation log contains a lot of WARN (previously)/INFO (with this update) messages.

So, behavior is not fixed for us...

best regards,
ty

typhoon2k:
I've investigated this a bit firther and it seems that some generators are using own logic for allOf cases. In our case we are using typescript-axios and plantuml generators and both are generating it as inheritance (for example, typescript-axios generator is using special allOf template).
But what worries me is that we are getting many many "inheritance without use of 'discriminator.propertyName' has been deprecated in the 5.x release" messages in generator log. Do we need to produce this spam now?

best regards,
ty

typhoon2k:
Hi @wing328 ,

what are the plans for this change? Can we get it in master sometime soon?

ty

P.S. It would be great to get rid of "inheritance without use of 'discriminator.propertyName' has been deprecated in the 5.x release" spamming message as well before merging this PR.

jimschubert:
@typhoon2k I don't think we can add a breaking change and deprecation without informing the user.

@wing328 can you wrap the above message in `once` as other spammy logs in `ModelUtils`? This way, users could tweak logging by changing the system properties and defaults:

* `-Dorg.openapitools.codegen.utils.oncelogger.cachesize=200`
* `-Dorg.openapitools.codegen.utils.oncelogger.expiry=2000`

typhoon2k:
@jimschubert , I just tried to play locally with once logger. I've made the following change to logging code:

```java
            once(LOGGER).info("[deprecated] inheritance without use of 'discriminator.propertyName' has been deprecated" +
                    " in the 5.x release. Composed schema name: " + composedSchema.getName() + ". Title: " + composedSchema.getTitle());
```

With this change I still get spam in logs. Do I need to configure once logger somehow additionally?

jimschubert:
Excellent suggestion. I'll go ahead and merge this, then change the info log to be generic and add a debug log with the name and title.

jimschubert:
@wing328 I'd merged this, but had to revert via #7323 as it caused all CI checks to fail in master. If you have a chance, could you merge master and take a look? It failed on `JavaClientCodegenTest#testAllowModelWithNoProperties` and I didn't readily see a way to allow for empty/marker classes on this change as the spec allows.

typhoon2k:
hi @wing328 , @jimschubert , is there any chance this update will get into master sometime soon?

PR #6924, opened 2020-07-13
[Dart] Treat float as double

sbu-WBT:
Double gets special special treatment in the mustache to prevent some errors like `type 'int' is not a subtype of type 'double'`.

https://github.com/OpenAPITools/openapi-generator/blob/dad931a25a54d6d1500654b96499d5239240c659/modules/openapi-generator/src/main/resources/dart2/class.mustache#L75-L82

Dart has no floats so they are replaced with doubles. Before this fix, it wasn't the case inside the mustache file so the floats had the type `double` in the class, but not the code from the `isDouble` block inside the mustache because only `isFloat` was true.

### PR checklist

@ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @athornz (2019/12) @amondnet (2019/12)


agilob:
Can we get a test for it pls?

wing328:
```
       {{#isDouble}} 
 {{name}} = (json['{{baseName}}'] == null) ? 
   null : 
   json['{{baseName}}'].toDouble(); 
       {{/isDouble}} 
       {{^isDouble}} 
 {{name}} = json['{{baseName}}']; 
       {{/isDouble}} 
```

What about using `{{#isNumber}} ... {{/isNumber}}` instead?

sbu-WBT:
Good idea, but would `{{#isNumber}} ... {{/isNumber}}` return true with an int? If not, I think it's the way to go @wing328

wing328:
`isNumber` will return false if it's integer or long.

FYI. `isNumeric` will return true if it's integer/long/double/float.

SimonIT:
@wing328 Replaced. I think the failing CI is not my fault because the master branch is also failing

sbu-WBT:
@wing328 Replaced. I think the failing CI is not my fault because the master branch is also failing

wing328:
Shippable CI not related to this PR.

wing328:
LGTM. We'll add tests after the dart2 client is ready to make the switch to use [fake petstore](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml)

PR #6938, opened 2020-07-14
[Java] Treat classes from java.lang as reserved words

joschi:
Classes in the package `java.lang` are available without explicitly importing them:
>  A compilation unit automatically has access to all types declared in its package and also automatically imports all of the public types declared in the predefined package `java.lang`.

Source: https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html

This leads to naming collisions in the generated code if there is a schema named like a class in `java.lang`.

For example a schema named `override` would clash with `java.lang.Override`, an annotation used in virtually every template for the Java generators.

As the least intrusive method (with regards to existing templates), we handle classes from `java.lang` as reserved words, so that they will be renamed accordingly.

### PR checklist

@bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger @karismann @Zomzog @lwlee2608 @bkabrda @wing328


joschi:
> As the least intrusive method (with regards to existing templates), we handle classes from `java.lang` as reserved words, so that they will be renamed accordingly.

As an alternative, we could also use the fully qualified class name (`java.lang.Override` instead of `Override`) in all templates. This seemed too intrusive for existing users, though.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -133,26 +134,35 @@ public AbstractJavaCodegen() {
...
+        ));
+        try {
+            ClassPath.from(ClassLoader.getSystemClassLoader())
+                    .getTopLevelClasses("java.lang").stream()
+                    .map(ClassPath.ClassInfo::getSimpleName)
+                    .filter(StringUtils::isNotBlank)
+                    .forEach(words::add);
+        } catch (IOException e) {
+            throw new UncheckedIOException("Couldn't load classes from java.lang", e);
+        }
```
joschi, line 164: The list of classes will depend on the version of the JVM being used in the generator.
As an alternative, we could list the classes explicitly (although this would be a very long list).

Zomzog: The list is empty with JDK11 and 14. I've tried to upgrade guava to `30.0-jre` without a better result.

I don't know if it is still possible to list java.lang classes. Maybe we must continue with the hand-made list.

Zomzog (changes requested):

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
joschi: @Zomzog Thanks for looking into this!

I could reproduce the issue with Java 11 and 15. 😢
Subsequently, I've opted to go with a static list of classes in `java.lang`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/JavaLangClassList.java
```diff
@@ -0,0 +1,261 @@
+package org.openapitools.codegen.languages;
```
wing328, line 1: Shall we put this file under `org.openapitools.codegen` instead as `org.openapitools.codegen.languages` contains a list of generators but this class is not a generator?

joschi: @wing328 Sorry for the late response.

I've moved the class to `org.openapitools.codegen.JavaLangClassList`.

wing328:
we've added name mapping support to java generators to provide another way to handle issues addressed by this PR: https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md#name-mapping

please give it a try with the latest master.

PR #6943, opened 2020-07-14
Fixes #6942: Added ability to prepend a basePath to typescript-redux-query generators

### LINKED ISSUE #6942: [REQ][typescript-redux-query] Allow client to set the base path to be prepended to endpoints in the generated query configs.
##### Description

If you generate a client using the `typescript-redux-query` generator, there is no way to set the base path for the url in the generated query configs.  This means you can't easily make your generated client communicate with your backend API if your client is hosted from a different domain.  One option is to override the network interface but that seems wrong considering other generators allow you to update the configuration and set the BASE_PATH.

Version: 5.0.0
My working version: commit cef1bec4666e7841c39a37eec97b2e81c96ad9f0

My suggested fix is to create a Configuration object in the `runtime.mustache` that defaults a variable called `basePath` to an empty string.  My decision for using an empty string is to keep this change backwards compatible.  We then update `apis.mustache` prepend the `runtime.Configuration.basePath` to the existing value being used for the `url` in the query config.  The below is a diff off of the `commit cef1bec4666e7841c39a37eec97b2e81c96ad9f0`.

stgraham2000:
Added a new Configuration object that has attributes that can be used to alter the behaviour of the query config generators.  In particular, added a basePath field which can be used to prepend to the urls in the query configs.  The old behaviour only used relative paths for the urls which means cross domain support was not possible.

Left the default basePath as an empty string to keep existing behaviour the same.

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-redux-query/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-redux-query/apis.mustache
@@ -157,7 +157,7 @@ function {{nickname}}Raw<T>({{#allParams.0}}requestParameters: {{operationIdCame
     {{/isListContainer}}
     {{/formParams}}
     const config: QueryConfig<T> = {
-        url: `{{{path}}}`{{#pathParams}}.replace(`{${"{{baseName}}"}}`, encodeURIComponent(String(requestParameters.{{paramName}}))){{/pathParams}},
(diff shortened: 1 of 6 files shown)


zefciu:
@stgraham2000 could you add some explanation on how to use it? The only way of modifying this value that I see is editing the auto-generated file.

Issue #4656, opened 2019-12-01
[BUG] typescript-node mixes require and import


DanielRuf:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When running `openapi-generator generate -i https://petstore3.swagger.io/api/v3/openapi.json -g typescript-node -o out -c config.json`

with the config.json file:
```json
{
    "npmName": "@swagger/typescript-fetch-petstore",
    "npmVersion": "1.0.0",
    "npmRepository": "https://skimdb.npmjs.com/registry",
    "snapshot": false,
    "supportsES6": true
}
```

it generates the following code which is not valid esm code:

```js
import localVarRequest = require('request');
import http = require('http');
```

##### openapi-generator version

`4.2.0`

macjohnny:
@DanielRuf would you like to implement a fix?

laubfall:
Just for your information: i ran into the same problem and fixed this issue in https://github.com/laubfall/openapi-generator/tree/typescript-node_mixes_require_and_import_4656 . PR is on its way.

laubfall:
https://github.com/OpenAPITools/openapi-generator/pull/6947

PR #6947, opened 2020-07-15
typescript-node: clean up require and import

laubfall:
… instead of import statement

- Resolves https://github.com/OpenAPITools/openapi-generator/issues/4656
- Fixed broken link in CONTRIBUTING.md

### PR checklist


modules/openapi-generator/src/main/resources/typescript-node/api-single.mustache
```diff
@@ -1,6 +1,13 @@
 {{>licenseInfo}}
+
+{{#supportsES6}}
+import localVarRequest from 'request';
+import http from 'http';
+{{/supportsES6}}
+{{^supportsES6}}
 import localVarRequest = require('request');
```
DanielRuf, line 8: The import var = require does not make much sense to me. This should be probably `const localVarRequest = require('request')`.

Or did I oversee something?

DanielRuf: Same for the other parts.

DanielRuf:
> * [ x]

You might want to remove the space inside the square brackets (it is either a space or x as single character) ;-)

macjohnny:
@laubfall could you please fix the CI issues?

modules/openapi-generator/src/main/resources/typescript-node/api-single.mustache
laubfall: Yeah thats right, i'll fix it :)

laubfall: I'm getting some typescript errors with const http = require. These are "TS2503: Cannot find namespace 'http'." and "error TS4055: Return type of public method from exported class has or is using private name 'http'.". As described here: https://stackoverflow.com/questions/31173738/typescript-getting-error-ts2304-cannot-find-name-require there are some ways to make this work, but to be honest i don't see a reason why to make a difference at this point. All other modules are imported with the "import .. from" stmt independent of the ES6 flag. So i removed the switch and changed the tsconfig to make the imports http/request work. What do you think about this solution?

DanielRuf: For TypeScript to use these you need the types for the Nodejs internal ones. Basically `@types/node`.

DanielRuf: The current solution should be fine.

wing328:

wing328:
I just pushed https://github.com/OpenAPITools/openapi-generator/pull/6947/commits/547e30521f96f8879fae76eca197ab7df20b99cc to fix the CI failure. Let's see how that goes.

samples/client/petstore/typescript-node/default/.openapi-generator/FILES
```diff
@@ -5,10 +5,10 @@ api/petApi.ts
 api/storeApi.ts
 api/userApi.ts
 git_push.sh
-model/apiResponse.ts
-model/category.ts
-model/order.ts
-model/pet.ts
-model/tag.ts
-model/user.ts
+model/./apiResponse.ts
```
DanielRuf: why this change again?

DanielRuf: Why this change? ./ makes not much sense in my opinion.

wing328: Agreed `./` is not needed. We'll tackle it separately.

laubfall: Ok, so there is nothing left here to todo for the PR?

laubfall:
> I just pushed [547e305](https://github.com/OpenAPITools/openapi-generator/commit/547e30521f96f8879fae76eca197ab7df20b99cc) to fix the CI failure. Let's see how that goes.

Thanks, i looked into the CI Failure but did not see the reason why the build failed

wing328:
Shippable CI failure not related to this PR.

PR #6960, opened 2020-07-16
[typescript-axios] switched to new URL API

### LINKED ISSUE #6859: [REQ] [typescript-axios] switching to new URL API

[Legacy `URL` API](https://nodejs.org/api/url.html#url_legacy_url_api) is now deprecated since Node.js 11 (Oct 2018).

Use [new `URL` API](https://nodejs.org/api/url.html#url_the_whatwg_url_api).
This will drop support Node.js below v6.0.0 (which was released Apr. 2016).

Not to use standard APIs and use packages.

sapphi-red:
Switched to new URL API since Node.js legacy URL API is deprecated.

This will drop support Node.js below v6.0.0 (which was released Apr. 2016).

close #6859

### PR checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -177,6 +177,7 @@ samples/client/petstore/python-tornado/.venv/
 samples/client/petstore/typescript-angular2/npm/npm-debug.log
 samples/client/petstore/typescript-node/npm/npm-debug.log
 samples/client/petstore/typescript-angular/tsd-debug.log
+samples/client/petstore/typescript-axios/tests/**/dist/
 samples/client/petstore/typescript-fetch/tests/**/dist/
 
 # aspnetcore
--- a/modules/openapi-generator/src/main/resources/typescript-axios/api.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/api.mustache
@@ -3,7 +3,6 @@
 {{>licenseInfo}}
 
 {{^withSeparateModelsAndApi}}
-import * as globalImportUrl from 'url';
 import { Configuration } from './configuration';
 import globalAxios, { AxiosPromise, AxiosInstance } from 'axios';
 // Some imports not used depending on template conditions
--- a/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
@@ -3,7 +3,6 @@
 /* eslint-disable */
 {{>licenseInfo}}
 
-import * as globalImportUrl from 'url';
 import globalAxios, { AxiosPromise, AxiosInstance } from 'axios';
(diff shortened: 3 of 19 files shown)


macjohnny:
thanks for your contribution

sapphi-red:
@macjohnny 
I found that this PR drops not only below v6 but below v10.
I overlooked the document that URL is not a global object until v10.
Since Node.js v7 to v9 is already EOL, I think this is not a serious problem.

macjohnny:
Thanks for the update

breeswish:
This PR introduces a bug that `const query = new URLSearchParams(localVarUrlObj.search);` may produce constant re-define when there is already a API parameter called `query`.

macjohnny:
@breeswish this could be fixed by renaming `query` to `queryParameters`, as `queryParameters` is in the list of the reserved words:
https://github.com/OpenAPITools/openapi-generator/blob/6909c888d2e3a2f0887e7b40f0811275474faf3d/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java#L111-L116

would you like to fix this?

sapphi-red:
I have made a PR #7475.

baurine:
Hi, it seems the above PR #7475 is not contained in the latest 5.0.0-beta2: https://github.com/OpenAPITools/openapi-generator/blob/v5.0.0-beta2/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache#L191

sapphi-red:
The latest version is 5.0.0 and that PR is included in 5.0.0 or 5.0.0-beta.3.

baurine:
> The latest version is 5.0.0 and that PR is included in 5.0.0 or 5.0.0-beta.3.

ok, thanks! let me try it later.

baurine:
hi @sapphi-red , sorry I didn't find the responding version in the https://www.npmjs.com/package/@openapitools/openapi-generator-cli , did version 5.0.0 and 5.0.0-beta.3 release to npm?

sapphi-red:
@baurine You need to install `2.x.x` for using `5.0.0-beta.2` and later. `yarn upgrade` won't bump versions which does not match with pattern written in `package.json`, so you need to run `yarn upgrade --latest @openapitools/openapi-generator-cli`. After installing it, set the version in config (https://www.npmjs.com/package/@openapitools/openapi-generator-cli#configuration).

baurine:
hi @sapphi-red , it works after upgrading to 2.1.16, thank you!

I just a little didn't understand, the old version just depends on ONE npm package while the new version depends on a ton of npm packages (likes nestjs, nuxtjs ...), isn't it just a wrapper of the jar? why does it need other npm packages?

image

image

sapphi-red:
I do not know about that, maybe creating an issue at the repository below would be better.
https://github.com/OpenAPITools/openapi-generator-cli

PR #6970, opened 2020-07-17
Fixes #6969: typescript-inversify client compile error, type casting …

### LINKED ISSUE #6969: Typescript-inversify generator is unusable
##### Description

Returns compile errors like TS2339: Property 'response' does not exist on type 'unknown'.

master

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/pull/6970

I've fixed this and also inconsistent apostrophes

michalzubkowicz:
This client was unusable for long time  because of typings, which fixes this PR, also I've changed apostrophes to be consistent in whole generator to more widely used (single).

fixes #6969


modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
```diff
@@ -73,21 +73,21 @@ export class {{classname}} {
...
-            queryParameters.push("{{paramName}}="+encodeURIComponent({{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}'])));
+            queryParameters.push('{{paramName}}='+encodeURIComponent({{paramName}}.join(COLLECTION_FORMATS['{{collectionFormat}}'])));
         {{/isCollectionFormatMulti}}
         }
         {{/isListContainer}}
         {{^isListContainer}}
         if ({{paramName}} !== undefined) {
         {{#isDateTime}}
-           queryParameters.push("{{paramName}}="+encodeURIComponent(<any>{{paramName}}.toISOString()));
+           queryParameters.push('{{paramName}}='+encodeURIComponent(({{paramName}} as Date).toISOString()));
```
macjohnny, line 87: what is the original type of `{{paramName}}`? can it be casted to Date without issues?

macjohnny:
cc @gualtierim

macjohnny:
@michalzubkowicz thanks for your PR!
can you please merge the most recent master into your branch?

modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
gualtierim: The type of paramName normally implements Date interface so this change should be correctly.

michalzubkowicz:
@macjohnny 

Only Date data type have toISOString method, so casting itself will not be a problem, also I'm using this generator for a while after these fixes. 

I'll merge last master

macjohnny:
> Only Date data type have toISOString method, so casting itself will not be a problem, also I'm using this generator for a while after these fixes

if it is `Date`, why is casting necessary at all? before, the casting was to `any`, so I guess it wasn't `any` or `Date`, so maybe it was `string`, but this cannot be directly cast to `Date`: https://www.typescriptlang.org/play/index.html#code/MYewdgzgLgBARgGwIYC4bQE4EswHMYC8MA5EhACYBmxA3AFB0ICmsAtk7qjACJJROF4yGGR58m9OkA

modules/openapi-generator/src/main/resources/typescript-inversify/api.service.mustache
macjohnny: I think the general type for `format: date-time` in the typescript generators is `string`, see https://github.com/OpenAPITools/openapi-generator/pull/5314

michalzubkowicz: @macjohnny 
> I think the general type for `format: date-time` in the typescript generators is `string`, see #5314

I'm pretty sure that is Date interface here, because I'm using it this generator.

michalzubkowicz:
> > Only Date data type have toISOString method, so casting itself will not be a problem, also I'm using this generator for a while after these fixes
> 
> if it is `Date`, why is casting necessary at all? before, the casting was to `any`, so I guess it wasn't `any` or `Date`, so maybe it was `string`, but this cannot be directly cast to `Date`: https://www.typescriptlang.org/play/index.html#code/MYewdgzgLgBARgGwIYC4bQE4EswHMYC8MA5EhACYBmxA3AFB0ICmsAtk7qjACJJROF4yGGR58m9OkA

You are right, this "cast" problably is not necessary, proper type should be introduced earlier, and for sure it's Date not string , because string doesn't have toISOString method. Just changed this type here to ease understanding why toISOString method is used.

Issue #6756, opened 2020-06-23
[REQ] Evolutive openapi to protobuf conversion


dobl1:
Hello,

### Is your feature request related to a problem? Please describe.

The converter from OAS to PB works great but only once !
Indeed if I update the original OAS file, let's say adding an attribute, then the .proto generated will have all the index shifted (and so not backward compatible).

## Describe the solution you'd like

One solution, certainly the simplest is to have attribute key hash as index, with some mechanics (increment ?) in case of conflict.
Also would be great to have the generated file named with the major version (ex : contact.v1.proto).

Regards.

dobl1:
Anyone ⁉️

wing328:
done by #7002.

@dobl1 let us know if that won't meet your requirement.

PR #7002, opened 2020-07-20
[Protobuf] Make it possible to override field numbers using x-protobuf-index

k0ral:
As described in issue https://github.com/OpenAPITools/openapi-generator/issues/6756 .
Cc: @dobl1 .

The protobuf generator currently does not guarantee that generated field numbers are preserved across multiple generations: it always generates them as a sequence from 0 to `n` (through the `x-index` vendor extension). If fields are re-ordered, added, or deleted, field numbers will almost always shift. This is a big deal, as field numbers are part of the binary interface of protobuf serialization format.

This change is making it possible to specify protobuf field numbers using the vendor extension `x-protobuf-index`. This is an opt-in feature at field level: if `x-protobuf-index` is not specified, then the current behavior based on `x-index` still applies. In other words: this is backward compatible.

For example, the following data definition:
```yaml
Pet:
  type: object
  properties:
    id:
      type: integer
      x-protobuf-index: 3
    name:
      type: string
      x-protobuf-index: 1
```
... will generate the below protobuf message:
```protobuf
message Pet {
  int64 id = 3;
  string name = 1;
}
```

I do not expect this pull-request to be merged as-is, I rather see it as a conversation-starter: do you agree with the general approach ? How would you suggest I improve the implementation ? I'd rather get feedback before investing time in writing automated tests, for example. For now, I've just manually validated the behavior on a toy example.

### PR checklist

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ProtobufSchemaCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ProtobufSchemaCodegen.java
@@ -28,7 +28,6 @@ import org.openapitools.codegen.meta.Stability;
 import org.openapitools.codegen.meta.features.DocumentationFeature;
 import org.openapitools.codegen.meta.features.SecurityFeature;
 import org.openapitools.codegen.meta.features.WireFormatFeature;
-import org.openapitools.codegen.utils.ProcessUtils;
 import org.openapitools.codegen.utils.ModelUtils;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


wing328:
@k0ral thanks for the PR. Please PM me via Slack (https://join.slack.com/t/openapi-generator/shared_invite/enQtNzAyNDMyOTU0OTE1LTY5ZDBiNDI5NzI5ZjQ1Y2E5OWVjMjZkYzY1ZGM2MWQ4YWFjMzcyNDY5MGI4NjQxNDBiMTlmZTc5NjY2ZTQ5MGM) when you've time.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ProtobufSchemaCodegen.java
```diff
@@ -216,13 +215,12 @@ public String toOperationId(String operationId) {
...
         objs = postProcessModelsEnum(objs);
         List<Object> models = (List<Object>) objs.get("models");
-        // add x-index to properties
-        ProcessUtils.addIndexToProperties(models, 1);
 
         for (Object _mo : models) {
             Map<String, Object> mo = (Map<String, Object>) _mo;
             CodegenModel cm = (CodegenModel) mo.get("model");
 
+            int index = 1
```
wing328, line 223: @k0ral overall it looks good 👍 but this line is missing the `;` at the end.

k0ral: Fixed, thanks !

wing328: 👍  Will merge after all the tests pass.

PR #7003, opened 2020-07-20
[Swift5] Fix Datetime default value

### LINKED ISSUE #5256: [BUG][JAVA][MAVEN] java.time.OffsetDateTime cannot be cast to java.lang.String
#### Bug Report Checklist

##### Description

Generation of client classes fails with class cast Exception while generation using maven Plugin.
Running `mvn clean compile` fails with 
```
[ERROR] 
java.lang.RuntimeException: Could not process operation:
  Tag: class Tag {
    name: Process
    description: REST-Services.
    externalDocs: null
}
  Operation: getProcessUsingGET
  Resource: get /api/prozesse
  Schemas: {}
  Exception: java.time.OffsetDateTime cannot be cast to java.lang.String
```

4.2.3
(Happens also in earlier Versions of 4.x.x)

The Spec used: (its valid according to https://apitools.dev/swagger-parser/online/ and https://editor.swagger.io/ )

```yaml
swagger: '2.0'
info:
  description: 'REST service'
  version: '2.1.9'
  title: REST
host: localhost
basePath: /
tags:
  - name: Process
    description: REST-Services.
paths:
  /api/prozesse:
    get:
      tags:
        - Process
      summary: Return a list of Process
      description: |
        Max response time MAX_RESPONSE_TIME_MS=20
      operationId: getProcessUsingGET
      produces:
        - application/json;charset=utf-8
      parameters:
        - name: dateFrom
          in: query
          description: Filters out Transactions before this date
          required: true
          type: string
          default: '2000-01-01T00:00:00.001Z'
          format: date-time
          allowEmptyValue: false
      responses:
        '200':
          description: OK
        '400':
          description: Bad Request
        '401':
          description: Unauthorized
        '403':
          description: Forbidden
        '404':
          description: Not Found
      security:
        - bearer: []
      deprecated: false
securityDefinitions:
  bearer:
    type: apiKey
    name: Authorization
    in: header
```

Used pom.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

	<modelVersion>4.0.0</modelVersion>
	<artifactId>failing-with-class-cast-exception</artifactId>
	<name>failing-with-class-cast-exception</name>
	<groupId>com.example</groupId>
	<version>0.0.1-SNAPSHOT</version>
	<build>
		<plugins>
			<plugin>
				<version>4.2.3</version>
				<executions>
					<execution>
						<id>kto-privatkunde</id>
						<goals>
							<goal>generate</goal>
						</goals>
						<configuration>
							<inputSpec>process.yml</inputSpec>
							<generateApiTests>false</generateApiTests>
							<configOptions>
								<library>jersey2</library>
								<hideGenerationTimestamp>true</hideGenerationTimestamp>
								<delegatePattern>true</delegatePattern>
								<java8>java8</java8>
								<dateLibrary>java8</dateLibrary>
							</configOptions>
							<additionalProperties>failOnEmptyBeans=false</additionalProperties>
						</configuration>
					</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
</project>
```

`mvn clean compile`

##### Steps to reproduce

Setup maven project and run `mvn clean compile`

Allezxandre:
This is a follow-up to #5256, which happens to affect the Swift5 code-gen.

If a default value is provided by the API spec for a date-time attribute, this PR fixes the generation code to build a correct Swift `Date` object.

To close #5256

### PR checklist

- [x] Copy the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language: @4brunu

### DIFF
--- a/samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/AdditionalPropertiesClass.swift
+++ b/samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/AdditionalPropertiesClass.swift
@@ -12,7 +12,7 @@ public struct AdditionalPropertiesClass: Codable {
     public var mapString: [String: String]?
     public var mapMapString: [String: [String: String]]?
 
-    public init(mapString: [String: String]?, mapMapString: [String: [String: String]]?) {
+    public init(mapString: [String: String]? = nil, mapMapString: [String: [String: String]]? = nil) {
         self.mapString = mapString
         self.mapMapString = mapMapString
     }
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
@@ -34,6 +34,10 @@ import java.io.File;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.time.OffsetDateTime;
+import java.time.Instant;
+import java.time.temporal.ChronoField;
+import java.util.concurrent.TimeUnit;
 
 import static org.openapitools.codegen.utils.StringUtils.camelize;
 
@@ -620,14 +624,19 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
                 }
             }
         }
-        if (ModelUtils.isIntegerSchema(p) || ModelUtils.isNumberSchema(p) || ModelUtils.isBooleanSchema(p)) {
-            if (p.getDefault() != null) {
+        if (p.getDefault() != null) {
+            if (ModelUtils.isIntegerSchema(p) || ModelUtils.isNumberSchema(p) || ModelUtils.isBooleanSchema(p)) {
                 return p.getDefault().toString();
-            }
-        } else if (ModelUtils.isStringSchema(p)) {
-            if (p.getDefault() != null) {
+            } else if (ModelUtils.isDateTimeSchema(p)) {
+                // Datetime time stamps in Swift are expressed as Seconds with Microsecond precision.
+                // In Java, we need to be creative to get the Timestamp in Microseconds as a long.
+                Instant instant = ((OffsetDateTime) p.getDefault()).toInstant();
+                long epochMicro = TimeUnit.SECONDS.toMicros(instant.getEpochSecond()) + ((long) instant.get(ChronoField.MICRO_OF_SECOND));
+                return "Date(timeIntervalSince1970: " + String.valueOf(epochMicro) + ".0 / 1_000_000)";
+            } else if (ModelUtils.isStringSchema(p)) {
                 return "\"" + escapeText((String) p.getDefault()) + "\"";
             }
+            // TODO: Handle more cases from `ModelUtils`, such as Date
         }
         return null;
     }
--- a/samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/Animal.swift
+++ b/samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/Animal.swift
@@ -12,7 +12,7 @@ public struct Animal: Codable {
     public var className: String
     public var color: String? = "red"
 
-    public init(className: String, color: String?) {
+    public init(className: String, color: String? = nil) {
         self.className = className
         self.color = color
     }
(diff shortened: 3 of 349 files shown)


4brunu:
Can you please run `./mvnw package && ./bin/generate-samples.sh bin/configs/other/swift5-*` and commit the changes?

wing328:
I've updated the samples via 80f9612c. Let's see how that goes.

4brunu:
It looks like the samples don't have a default date value, so this PR will be difficult to validate.
@Allezxandre could you please test this changes locally?

samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/AdditionalPropertiesClass.swift
```diff
@@ -12,7 +12,7 @@ public struct AdditionalPropertiesClass: Codable {
     public var mapString: [String: String]?
     public var mapMapString: [String: [String: String]]?
 
-    public init(mapString: [String: String]?, mapMapString: [String: [String: String]]?) {
+    public init(mapString: [String: String]? = nil, mapMapString: [String: [String: String]]? = nil) {
```
wing328, line 15: @Allezxandre looks like many parameters are default to `nil`. Is that expected?

4brunu: This is not related to this PR, it was introduced in another one, but apparently the samples were outdated.

Allezxandre:
> It looks like the samples don't have a default date value, so this PR will be difficult to validate.
> @Allezxandre could you please test this changes locally?

I tested this locally, and this yields the following default date when it is defined to March 17th 2020 10:00:00 UTC:

```
Date(timeIntervalSince1970: 1584439200000000.0 / 1_000_000)
```

samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/AdditionalPropertiesClass.swift
wing328: I actually tested with the master but didn't get these `nil` default value. Maybe my test wasn't done right.

4brunu: Let me test on master

4brunu: When I run `./mvnw package && ./bin/generate-samples.sh bin/configs/other/swift5-*` on the master, I get all of those `nil` changes

4brunu:
Is there any disadvantages in removing the division like this?
`Date(timeIntervalSince1970: 1584439200)`

samples/client/petstore/swift5/alamofireLibrary/PetstoreClient/Classes/OpenAPIs/Models/AdditionalPropertiesClass.swift
wing328: Thanks for confirming that. 

Is the `nil` default value expected? If not, we can remove it later in another PR. ( I think I know where to fix it )

4brunu: Yes it is expected. 
Basically this is just providing default `nil` values for for nullable types to shorten the constructor to the library user.

Allezxandre:
> Is there any disadvantages in removing the division like this?
> `Date(timeIntervalSince1970: 1584439200)`

This is because the date time can be expressed up to the microsecond precision. However, in Java they are represented as Microseconds stored in a long and in Swift they are represented as seconds represented in a double.

To avoid float precision issue, I let Java keep its storage as a long, and then add a `.0` and let the Swift compiler divide by a million to get the seconds stored in a double. 

In my example of a `1584439200.0` seconds timestamp, the date I used has 0 minutes, 0 seconds, 0 milliseconds, and 0 microseconds, so I guess it would be smarter in that case to remove the division. However, in other cases with dates that are not rounded like that, the division would prove useful.

4brunu:
Thanks for the explanation, looks good to me 👍

PR #7029, opened 2020-07-23
Fix string comparison in C++ UE4 client generator

wing328:
- Fix string comparison in C++ UE4 client generator

CC @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @martindelille (2018/03) @muttleyxd (2019/08)

### PR checklist

cc @Kahncode as well - the author of the generator


samples/client/petstore/cpp-ue4/Private/OpenAPIPetApiOperations.cpp
```diff
@@ -220,7 +220,6 @@ void OpenAPIPetApi::FindPetsByStatusResponse::SetHttpResponseCode(EHttpResponseC
 	switch ((int)InHttpResponseCode)
 	{
 	case 200:
-	default:
```
Kahncode, line 223: Has something else changed? These default clauses are needed if I remember correctly.

wing328: Yes, there was a fix. Previously the first response code is considered the `default`, which is incorrect as `default` should be defined separately.

Shall we revise the code to throw an exception instead as the response code is not defined in the spec?

Kahncode: No, this code does not support exceptions. Essentially if the response code was undefined in the spec I assumed it was a correct response. You may change that default behavior but I believe my tests ended up concluding that this was the best case scenario.

wing328: What about logging an error message instead?

wing328: I think I'll merge this PR which aims to fix the string comparison. 

The issue (default case has gone) is present in the generator regardless of this PR.

PR #7085, opened 2020-07-31
[ci] Verify supported JDK versions on master push

jimschubert:
This sets up a GitHub Workflow to verify JDK 8 (Linux, Windows), 11 (Linux), and 13 (Linux) on pushes to master. It also verifies that the ensure-up-to-date script works in both Linux and Windows.

This is applied to pushes to master only, because GitHub Workflows currently don't support run cancellation.

JDK 13 skips building the Maven and Gradle plugins for simplicity since the version of Gradle we use to build the Gradle plugin does not support JDK 13 and the Maven plugin is compiled and released using JDK 8. This supported version check is provided for those contributing generators and other functionality aside from plugins. we may add back in plugin checks to JDK 13 once our Gradle dependency is updated.

### PR checklist

### DIFF
--- a/.github/workflows/check-supported-versions.yaml
+++ b/.github/workflows/check-supported-versions.yaml
@@ -1,0 +1,80 @@
+name: Check Supported Java Versions
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


wing328:
LGTM 👍

PR #7149, opened 2020-08-06
Upgrade multiplatform libraries

### LINKED ISSUE #5346: [REQ] kotlin multiplatform - support newer kotlin version
Currently library versions are

- **Platform**: Kotlin multiplatform. 
- **Kotlin version**: 1.3.50
- **HTTP client**: Ktor 1.2.4. 
- **JSON processing**: Kotlinx Serialization: 0.12.0.

It should be nice to address those library versions:

- **Platform**: Kotlin multiplatform. 
- **Kotlin version**: 1.3.61
- **HTTP client**: Ktor 1.3.0. 
- **JSON processing**: Kotlinx Serialization: 0.14.0.

cromefire:
Fixes #5346

 - Upgraded the kotlin version
 - Changed the gradle files to kotlin
 - Upgraded ktor (and fixed the code, because there have been _many_ changes)
 - Removed explicit coroutines and kotlinx.serialization dependency
 - Fixed some codestyle

### PR checklist
 

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/api.mustache
```diff
@@ -66,19 +70,22 @@ import kotlinx.serialization.internal.StringDescriptor
...
             {{/hasBodyParam}}
 
-        val localVariableQuery = mutableMapOf<String, List<String>>()
-        {{#queryParams}}
-        {{{paramName}}}?.apply { localVariableQuery["{{baseName}}"] = {{#isContainer}}toMultiValue(this, "{{collectionFormat}}"){{/isContainer}}{{^isContainer}}listOf("${{{paramName}}}"){{/isContainer}} }
-        {{/queryParams}}
+        {{! TODO: Don't use all the listOf() }}
+        val localVariableQuery = mutableMapOf<String, List<String>?>(
+            {{#queryParams}}
+            "{{baseName}}" to {{#isContainer}}{{{paramName}}}?.let { toMultiValue(it, "{{collectionFormat}}") }{{/isContainer}}{{^isContainer}}listOf("${{{paramName}}}"){{/isContainer}}{{#hasMore}},{{/hasMore}}
```
cromefire, line 76: This seems to have to handle the special case of a map, which has to be split up into multiple query parameters, but I don't know how to do that

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/api.mustache
```diff
@@ -66,19 +70,22 @@ import kotlinx.serialization.internal.StringDescriptor
...
+        val localVariableHeaders = mutableMapOf<String, String?>(
+            {{#headerParams}}
+            "{{baseName}}" to {{#isContainer}}{{{paramName}}}?.joinToString(separator = collectionDelimiter("{{collectionFormat}}")){{/isContainer}}{{^isContainer}}this?.toString(){{/isContainer}}{{#hasMore}},{{/hasMore}}
+            {{/headerParams}}
+        )
 
         val localVariableConfig = RequestConfig(
             RequestMethod.{{httpMethod}},
-            "{{path}}"{{#pathParams}}.replace("{" + "{{baseName}}" + "}", "${{{paramName}}}"){{/pathParams}},
+            "{{path}}"{{#pathParams}}.replace("{{curly true}}{{baseName}}{{curly}}", "${{{paramName}}}"){{/pathParams}},
```
cromefire, line 88: This also seems like a waste, but I now sure, where and if I can transform that URL:
```kotlin
"/jellyfin/Audio/{itemId}/stream".replace("{itemId}", "$itemId").replace("{container}", "$container")
```

samples/client/petstore/kotlin-gson/src/main/kotlin/org/openapitools/client/models/Category.kt
```diff
@@ -1,29 +1,33 @@
...
  * A category for a pet
+ *
  * @param id 
  * @param name 
  */
 
-data class Category (
+data class Category(
     @SerializedName("id")
+
```
cromefire, line 27: This should go away

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/build.gradle.kts.mustache
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    }
+    ios {
+        binaries {
+            framework {
+                freeCompilerArgs = listOf("-Xobjc-generics")
+            }
+        }
+    }
```
cromefire, line 35: I can't test IOS

cromefire:
Technical committee:
@jimschubert, @dr4ke616 @karismann @Zomzog @andrewemery @4brunu @yutaka0m

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
```diff
@@ -521,7 +523,7 @@ private void processMultiplatformLibrary(final String infrastructureFolder) {
         typeMapping.put("object", "kotlin.String");  // kotlin.Any not serializable
 
         // multiplatform import mapping
-        importMapping.put("BigDecimal", "kotlin.Double");
+        importMapping.put("BigDecimal", "Double");
         importMapping.put("UUID", "kotlin.String");
```
cromefire, line 527: All of these types should also probably be mapped in a better way if possible somehow.
Maybe with custom classes that can be converted into local types then (if available)

cromefire: Loss of data should be avoided, but if it's possible to use some wrapper type, instead of the string, that can be converted to the equivalent types of that platform (like `LocalDatetime` or some of the other APIs for the jvm, `Date` for js and probably something for iOS).
Some of them, like URI, might already have primitives in kotlin std or ktor.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/api.mustache
wing328: Are you looking for `isMapContainer` for map?

cromefire: Yeah, thanks

cromefire: Is there documentation on which exist somewhere?

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/ApiClient.kt.mustache
```diff
@@ -0,0 +1,17 @@
...
+import {{packageName}}.apis.*
+
+class ApiClient(
+    baseUrl: String = "{{{basePath}}}",
+    httpClientEngine: HttpClientEngine? = null,
+    jsonConfiguration: JsonConfiguration = JsonConfiguration.Stable
+) {
+    {{#apiInfo}}
+    {{#apis}}
+    val {{{classVarName}}} = {{{classFilename}}}(baseUrl, httpClientEngine, jsonConfiguration)
```
cromefire, line 14: This may generate the following (invalid) code:
```kotlin
val package = PackageApi(baseUrl, httpClientEngine, jsonConfiguration)
```

wing328:
cc @jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10) @4brunu (2019/11) @yutaka0m (2020/03)

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/api.mustache
cromefire: (it also doesn't generate the curly braces for some reason)

wing328: I think I can fix this. Let's address this in a separate PR instead so please undo the change.

andrewemery:
LGTM

cromefire:
My last review comment (the `package` thing) should probably be fixed before merging though

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
jimschubert: What was the reason for changing `kotlin.Double` to `Double` here? I'd have to test, but iirc kotlin client would cause this to become a java.lang.Double (or something similar) and wouldn't honor kotlin's nullable functionality.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/ApiClient.kt.mustache
jimschubert: `package` is a reserved word and should be escaped. Did you perform a test that lead to the invalid example you provided? If so, could you share? I'd like to debug through to understand why package wasn't escaped.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
```diff
@@ -535,11 +538,16 @@ private void processMultiplatformLibrary(final String infrastructureFolder) {
         importMapping.put("Base64ByteArray", packageName + ".infrastructure.Base64ByteArray");
         importMapping.put("OctetByteArray", packageName + ".infrastructure.OctetByteArray");
 
+        // bundled client file
+        supportingFiles.add(new SupportingFile("ApiClient.kt.mustache", baseFolder, "ApiClient.kt"));
```
jimschubert, line 542: Note that renaming `ApiClient.kt.mustache` to `ApiClientBase.kt.mustache` is going to cause a lot of problems for anyone who has customized `ApiClient.kt.mustache`.

Can we find a naming strategy which retains `ApiClient.kt.mustache` without breaking existing extensions?

If not, we need to mark this PR as a breaking change and provide some guidelines in the PR for how one would migrate from using multiplatform before 5.0 to use it after this PR is merged.

jimschubert:
I left a few comments as inline comments rather than a "review", since they began as replies to existing comments.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
cromefire: Kotlin uses the `kotlin.*` types by default afaik also the scraems at me a lot otherwise for not leaving `kotlin.` away, so it seems fine.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
cromefire: It's breaking pretty much, because it used experimental / alpha state libraries (MPP itself is still experimental) and they had a lot of breaking changes themselves, so that not really something you can avoid. I tried to do my best and migrate it to stable versions so there's no deprecated, opt-in or otherwise experimental stuff anymore but it might be incompatible here and there.

cromefire: I'll see what I can do to add a little migration guide or so

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/ApiClient.kt.mustache
cromefire: I can get you an openapi def that reproduces it, but it's rather large

cromefire:
I hope I got everything

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/ApiClient.kt.mustache
jimschubert: I can put together a minimal doc. I just figured I'd test the same as you if it was handy, but not a big deal if it's cumbersome.

cromefire: This is the thing:
https://gitlab.com/-/snippets/2007909 (github sadly errors when I try to upload it)

But it's _huge_

cromefire: The thing that's causing it is probably the
```json
"tags": [
  "Package"
]
```

jimschubert:
> I hope I got everything

@cromefire Did you mean you'd hoped that you had touched each comment? I don't see any additional commits, so I'm confused about whether you were going to push additional work.

cromefire:
Yeah I meant comments because you said you put them there as inline comments and I only found 2

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/ApiClient.kt.mustache
cromefire: Where you able to find the issue that causes it's or should I try crafting a smaller file?

cromefire: Did a "hacky" solution, for now I just escape everthing. It works fine but it's not quite a best practice

cromefire:
Upgraded to Kotlin 1.4, Ktor 1.4 and kotlinx.serialization 1.4. That way even more experimental stuff has been replaced with more stable APIs. (So there won't be as many (hopefully none) breaking changes in the future)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinClientCodegen.java
cromefire: Just reverted it for now it shouldn't make a difference except generating some more warnings because of overqualified paths

cromefire:
Due to large code and structural differences between `kotlin-(okhttp|retrofit)` and `kotlin-multiplatform` it might be wise to split it to a separate generator (like done for many other languages)

modules/openapi-generator/src/main/resources/kotlin-client/libraries/multiplatform/build.gradle.kts.mustache
cromefire: So I basically need someone with a mac who can write a quick unittest for pet store or so and then run the test

cromefire:
CI seems to have some issues with keyserver.ubuntu.com timing out:
```
gpg: requesting key E084DAB9 from hkp server keyserver.ubuntu.com
```

cromefire:
Do you want me to do the split here or do you want me to create a 2nd PR for it after this one is merged? If it's fine doing it in this PR is easier and faster for me (for you it might increase the number of lines changed but decrease complexity due to less entangled code)

cromefire:
What I can also do if you want to keep compatibility is I can create a new generator (I've already seen that there are some `-next` ones) from this PR and keep the old so it can be gracefully deprecated

jimschubert:
I originally thought it might be better to have multiplatform separately for the same reasons… maybe now's the time to do that with the major version coming up.

Any thoughts? @dr4ke616 @karismann @Zomzog @andrewemery @4brunu @yutaka0m @OpenAPITools/generator-core-team

cromefire:
Also with the new version is there a timeframe when it has to land or if it's implemented as a "new" one can it land after the version is already released?

4brunu:
I also think it's a good idea to split this into a new generator, since it has more differences that similarities with the jvm kotlin clients.

cromefire:
Based one your feedback (it doesn't seem like anyone wants to oppose splitting this into a new one, although it's also only been 16h) I've decided that I'll provably close this PR on the weekend (if nobody has anything against it until then) and propose a new one, which adds a new client that uses the code from this PR. I'll still be watching feedback you give me here and incorporate it into the new PR.

Following that:
- What do I need to do to add a new client (other than creating the generator class for it)?
- Is there anything special I need to know about (command line) options? (e.g. special naming, common options vs. additional options, etc.)

jimschubert:
@cromefire I'm glad you asked. We have documentation about how to create a new generator here: https://openapi-generator.tech/docs/new-generator/.

While I was finding the link for you, I realized that it hasn't been updated with the new configuration-based approach that we've recently started following (rather than the sample script, it would be a sample config under `bin/config`).

jimschubert:
@cromefire I've updated the documentation via #7334. By the time you read this, it'll probably be live on the site and the `new.sh` script in master is now updated to output a yaml config under `./bin/configs` rather than a `.sh` and `.bat` script under bin.

cromefire:
Replaced by #7353

PR #7165, opened 2020-08-08
Add result in 500 status code

Gmtstephane:
Hi, in the Go server, we can only have Status code 200 or 500 wether our implementation function return an error or not.

In case or error (err!=nil), the server write the status code 500 without any messages .

```
result, err := c.service.AddPet(*pet)
	if err != nil {
		w.WriteHeader(500)
		return
	}	
	EncodeJSONResponse(result, nil, w)
```

In many cases, their can be several types of error in our implementation, and i wouldlike to return the  error message to the client.
I suggest to Encode the response with the result variable, with the statuscode 500, which will let us write responses in case of error.

```
        if err != nil {
		statusCode := 500
		EncodeJSONResponse(result, &statusCode, w)
		return
	}

```

@antihax (2017/11) @grokify (2018/07) @kemokemo (2018/09) @bkabrda (2019/07) @wing328 

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
+++ b/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
@@ -95,7 +95,7 @@ func (c *{{classname}}Controller) {{nickname}}(w http.ResponseWriter, r *http.Re
 	result, err := c.service.{{nickname}}({{#allParams}}{{#isBodyParam}}*{{/isBodyParam}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
 	if err != nil {
 		statusCode := 500
-		EncodeJSONResponse(result, &statusCode, w)
+		EncodeJSONResponse(err.Error(), &statusCode, w)
 		return
 	}
 	
--- a/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
+++ b/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
@@ -94,7 +94,8 @@ func (c *{{classname}}Controller) {{nickname}}(w http.ResponseWriter, r *http.Re
 	{{/isBodyParam}}{{/allParams}}
 	result, err := c.service.{{nickname}}({{#allParams}}{{#isBodyParam}}*{{/isBodyParam}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
 	if err != nil {
-		w.WriteHeader(500)
+		statusCode := 500
+		EncodeJSONResponse(result, &statusCode, w)
 		return
 	}
 	
--- a/samples/server/petstore/go-api-server/.openapi-generator/VERSION
+++ b/samples/server/petstore/go-api-server/.openapi-generator/VERSION
@@ -1,1 +1,1 @@
-5.0.0-SNAPSHOT
+5.0.0-beta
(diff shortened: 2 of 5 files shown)


Gmtstephane:
Still no review ? 
It's been a week and it's quiet a problem.

Thanks

kemokemo:
Hi @Gmtstephane . Thank you for your PR.

> I would like to return the error message to the client.

I agree with you. In order to achieve this, I felt it is better to include the content of `error` in the content to be returned. (I think a lot of implementation styles in the golang return an empty struct or nil if there is an error. So I don't think returning a `result` will often give us useful information.)

Maybe not so good, but how about doing the following, for example?

```go
EncodeJSONResponse(err.Error(), &statusCode, w)
```

or

```go
EncodeJSONResponse(fmt.Sprintf("result: %v, error: %v", result, err), &statusCode, w)
```

On the other hand, from a security point of view, I am concerned about the security risk of returning the contents of the error message to the client, which would normally be written to the log, because it makes the internal implementation predictable.
However, this point can be worked out on the endpoint implementation side, so it may not be as much of a problem as it should be.

@antihax @grokify @bkabrda @wing328
What do you all think?

Gmtstephane:
> I agree with you. In order to achieve this, I felt it is better to include the content of `error` in the content to be returned. (I think a lot of implementation styles in the golang return an empty struct or nil if there is an error. So I don't think returning a `result` will often give us useful information.)

Well i think you are right, it's better to return the error message. I just designed my api to return it as the result, by doing 
`return err.Error(),err` in the implementation, that is why.

Every solution is good to me, as long as we can read the error on the client side.

In my case i am cascading Api: 
 Client 1 make a request to server 1, wich make a request to a server 2 , and then the server 1 respond to the client. In case of error 500 and without any messages, their is no way from the client to know the type of error and it's location.

Thanks for you reply @kemokemo

kemokemo:
@Gmtstephane 
Thank you. It gave me more specifics and an idea of how to use it. That kind of usage would work well.

----
(This is not the main topic of this issue, and it is my personal opinion.)
If you want to find out where errors are occurring in multiple services, I think it would be better to use Prometheus or Elastic Stack to allow the server side administrator to quickly find out what is going on, rather than just from what the client receives.

Gmtstephane:
Still nothing new ?

I added a commit which change the result with the error content.

kemokemo:
I don't think there's a problem if the `service`'s endpoint function is implemented keeping in mind that the returned error content will be passed to the client side as well. I'm OK with this change.

Gmtstephane:
Ok,
I do not know who is able to accept and merge this PR so i allow myself to ping you all.
@antihax (2017/11) @grokify (2018/07) @kemokemo (2018/09) @bkabrda (2019/07) @wing328

Thanks

wing328:
@Gmtstephane can you please resolve the merge conflicts when you've time? I'll take a look this week when I can find the time.

wing328:
Looks like the master already contains a similar enhancement/fix:

```diff
<<<<<<< HEAD
        EncodeJSONResponse(err.Error(), &result.Code,{{#addResponseHeaders}} result.Headers,{{/addResponseHeaders}} w)
=======
        statusCode := 500
        EncodeJSONResponse(result, &statusCode, w)
>>>>>>> Add result in 500 status code
```

Closing this one for the time being. Can reopen if needed

PR #7186, opened 2020-08-11
[WIP][python-exp] Force camelization of imports

timgclark:
* Forces formatting of import classnames to handle edge case schemas

If merged this will resolve:
https://github.com/OpenAPITools/openapi-generator/issues/7187

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
@@ -330,7 +330,7 @@ public class PythonClientExperimentalCodegen extends PythonClientCodegen {
     @Override
     public String toModelImport(String name) {
         // name looks like Cat
-        return "from " + modelPackage() + "." + toModelFilename(name) + " import "+ name;
+        return "from " + modelPackage() + "." + toModelFilename(name) + " import "+ toModelName(name);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java
```diff
@@ -330,7 +330,7 @@ public String toDefaultValue(Schema p) {
     @Override
     public String toModelImport(String name) {
         // name looks like Cat
-        return "from " + modelPackage() + "." + toModelFilename(name) + " import "+ name;
+        return "from " + modelPackage() + "." + toModelFilename(name) + " import "+ toModelName(name);
```
spacether, line 333: Thank you for this fix!

This is definitely the right way to fix it.
Because our tests currently all pass, I am concerned that this corner case can break in the future and we would have no way to tell.

Can you add a test that verifies that the model class name is correct and the import is correct for a model named oddly? One could add this test in a couple locations:
- the [v3 sample tests](https://github.com/OpenAPITools/openapi-generator/tree/master/samples/openapi3/client/petstore/python-experimental/tests), how about adding a test of the model `_special_.model._name` which is named `SpecialModelName` in python
- the [java testing file](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/python/PythonClientExperimentalTest.java) for this generator

PR #7191, opened 2020-08-12
Properly decode $ref

### LINKED ISSUE #5720: [BUG] URI encoded $ref not properly decoded
#### Bug Report Checklist

##### Description

I wanted to generate a dart lib for flutter with the provided swagger yaml config. The problem is, that some of the generated class names contain spaces which are not valid. After writing a fix for it (PR #5693), I used swagger to edit the yaml. Swagger suggests to use `%20` instead of normal spaces inside `$ref` definition links because it's an URI (see [docs](https://swagger.io/docs/specification/using-ref/)). Now I got the problem, that some class names contain `%20`.

4.2.3

https://raw.githubusercontent.com/wbt-solutions/openproject_dart_sdk/master/openProject_swagger.yaml

I used the command `java -jar .\openapi-generator-cli-4.2.3.jar generate -g dart -i openProject_swagger.yaml  -o openproject_dart_sdk/` inside the Windows Powershell.

##### Steps to reproduce

Execute the command with the provided config and look for example in the *model\work_package.dart*.

##### Related issues/PRs

Haven't found any.

##### Suggest a fix

Properly decode the $ref attribute as URI.

sbu-WBT:
fixes #5720

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -41,6 +41,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.commons.io.FileUtils;
 
+import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
 import java.net.URI;
 import java.net.URLDecoder;
@@ -387,7 +388,11 @@ public class ModelUtils {
 
         }
 
-        return URLDecoder.decode(ref);
+        try {
+            return URLDecoder.decode(ref, "UTF8");
+        } catch (UnsupportedEncodingException e) {
+            return ref;
+        }
     }
 
     /**
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -389,7 +389,7 @@ public class ModelUtils {
         }
 
         try {
-            ref = URLDecoder.decode(ref, "UTF8");
+            ref = URLDecoder.decode(ref, "UTF-8");
         } catch (UnsupportedEncodingException ignored) {
         }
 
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -386,6 +388,11 @@ public static String getSimpleRef(String ref) {
 
         }
 
+        try {
+            ref = URLDecoder.decode(ref, "UTF-8");
+        } catch (UnsupportedEncodingException ignored) {
+        }
+
```
wing328, line 395: Shall we decode it at the start of the function `getSimpleRef` instead?

SimonIT: I had the same thought, but why decoding a String part which gets thrown away? I see no reason to do so

By the way, I think we should also unescape these special cases:
image

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -386,6 +388,13 @@ public static String getSimpleRef(String ref) {
...
+        ref = ref.replace("~1", "/").replace("~0", "~");
```
jimschubert, line 396:
```suggestion
        // see https://tools.ietf.org/html/rfc6901#section-3
        // Because the characters '~' (%x7E) and '/' (%x2F) have special meanings in
        // JSON Pointer, '~' needs to be encoded as '~0' and '/' needs to be encoded 
        // as '~1' when these characters appear in a reference token.
        // This reverses that encoding.
        ref = ref.replace("~1", "/").replace("~0", "~");
```

jimschubert (commented):
Suggested comment on the escapes, so they don't get removed accidentally in the future.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
sbu-WBT: @jimschubert Also added a unit test

jimschubert:
I've added OAS 3.0 spec support because the JSON pointer escapes make sense for refs pointing to operations/paths.

I think this approach is good to merge even before we cut 5.0.0-beta2. Any thoughts @wing328?

wing328 (approved):

Issue #7204, opened 2020-08-13
Support binary responses in Kotlin client


revmischa:
##### Description

Right now the Kotlin client doesn't support binary responses, only JSON. 

##### openapi-generator version

5.0.0-beta

##### OpenAPI declaration file content or url

It is impossible to consume this response with Kotlin right now:
```yaml
responses:
        '200':
          description: The generated PDF
          content:
            application/pdf:
              schema:
                type: string
                format: binary
```
[Throws error](https://github.com/OpenAPITools/openapi-generator/blob/f402126460e7d01251f9f13ff3213cb69bce7fae/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache#L173) `responseBody currently only supports JSON body`

##### Command line used for generation

Kotlin, gradle

##### Steps to reproduce

Make any request for a binary endpoint in Kotlin.

##### Related issues/PRs

Didn't see any.

##### Suggest a fix/enhancement

`AbstractKotlinCodegen` unfortunately maps the response type to `java.io.File`, which while convenient for sending a file request, is not well-suited to a response, because a `File` _must_ refer to a file on disk.
It would be much preferable to change the type mapping to `InputStream` or something similar for the response only, but I couldn't figure out how to do this without breaking the request type mapping as well. Maybe there is some way to leverage the `isResponseBinary` property but it wasn't clear to me how to change this.

So I changed it to generate a tempfile in my PR. If we could make it be an `InputStream` that would be better, but it's not clear to me how.

4brunu:
@wing328  This is implemented already, so maybe this issue can be closed

PR #7205, opened 2020-08-13
[Kotlin]] add support for binary response

revmischa:
Add missing support for handling binary responses in Kotlin clients.
It would be a lot better to use `InputStream` for the response type mapping, could use some advice on how to achieve that.

More discussion at issue: https://github.com/OpenAPITools/openapi-generator/issues/7204

@jimschubert , @dr4ke616  @karismann  @Zomzog  @andrewemery @4brunu  @yutaka0m 

### PR checklist
 

wing328:
```
e: /root/src/github.com/OpenAPITools/openapi-generator/samples/client/petstore/kotlin-jackson/src/main/kotlin/org/openapitools/c
> Task :compileKotlin FAILED
lient/infrastructure/ApiClient.kt: (121, 23): Unresolved reference: BufferedWriter
e: /root/src/github.com/OpenAPITools/openapi-generator/samples/client/petstore/kotlin-jackson/src/main/kotlin/org/openapitools/client/infrastructure/ApiClient.kt: (121, 38): Unresolved reference: FileWriter

* What went wrong:
Execution failed for task ':compileKotlin'.
> Compilation error. See log for more details

```

Shippable CI reports the above error. Please take a look when you've time.

revmischa:
Imports fixed. Not sure what's with the CircleCI error, looks like a transient issue to me

wing328:
All tests passed. If no further feedback on this PR, I'll merge it over the weekend.

revmischa:
I think my solution is not a very good solution. We shouldn't write out tempfiles IMO, we should pass the InputStream handle back to the caller. I was hoping to start some discussion on how to achieve that because I couldn't figure out how to change the type mapping only for the _response_.
I think the response type should be InputStream or something similar, but want to leave the request type as File because it already has support for transmitting files. It wasn't clear to me how to change the type mapping for the response while leaving the request type mapping as File.

wing328:
> I think the response type should be InputStream or something similar, but want to leave the request type as File because it already has support for transmitting files. It wasn't clear to me how to change the type mapping for the response while leaving the request type mapping as File.

What about using InputStream for both response and request ? In other words, we'll change the type mapping to map `binary` (format) to InputStream.

I'll put the merge on hold.

revmischa:
I think this would have to be changed and it would break existing users https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache#L100

wing328:
Upcoming version v5.0.0 is a major release and it's ok to include breaking changes without fallbacks.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
```diff
@@ -168,6 +170,15 @@ import org.threeten.bp.OffsetTime
         if (bodyContent.isEmpty()) {
             return null
         }
+        if (T::class.java == File::class.java) {
+            // return tempfile
+            val f = createTempFile()
```
wing328, line 175: Can we use `Files.createTempFile()` instead of `File.createTempFile()` has security issue?

ref: https://cwe.mitre.org/data/definitions/378.html

wing328: Added via https://github.com/OpenAPITools/openapi-generator/pull/9284

wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/9284

PR #7323, opened 2020-09-01
Revert "Correct allOf with only one child schema (no discriminator)"

jimschubert:
Reverts OpenAPITools/openapi-generator#6901


jimschubert:
@wing328 FYI

PR #7341, opened 2020-09-03
[kotlin] Fix optional header params in openhttp client

isycat:
This change adjusts the kotlin openhttp api template to ignore null headers when building the request rather than convert them to "null" string

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
@@ -40,7 +40,7 @@ import {{packageName}}.infrastructure.toMultiValue
     {{#isDeprecated}}
     @Deprecated(message = "This operation is deprecated.")
     {{/isDeprecated}}
-    {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}?{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
+    {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
         val localVariableBody: kotlin.Any? = {{#hasBodyParam}}{{#bodyParams}}{{{paramName}}}{{/bodyParams}}{{/hasBodyParam}}{{^hasBodyParam}}{{^hasFormParams}}null{{/hasFormParams}}{{#hasFormParams}}mapOf({{#formParams}}"{{{baseName}}}" to {{{paramName}}}{{#hasMore}}, {{/hasMore}}{{/formParams}}){{/hasFormParams}}{{/hasBodyParam}}
         val localVariableQuery: MultiValueMap = {{^hasQueryParams}}mutableMapOf()
 {{/hasQueryParams}}{{#hasQueryParams}}mutableMapOf<kotlin.String, List<kotlin.String>>()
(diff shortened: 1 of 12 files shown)


modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
```diff
@@ -40,7 +40,7 @@ import {{packageName}}.infrastructure.toMultiValue
     {{#isDeprecated}}
     @Deprecated(message = "This operation is deprecated.")
     {{/isDeprecated}}
-    {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}?{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
+    {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
```
jimschubert, line 43: note that this will be a breaking change if `sortParamsByRequiredFlag` is false. If you specify a default value before other parameters with no default value, you can only invoke the method with named parameters.

This will also potentially break Java interop as it would require `@JvmOverloads` on all methods, and this could have a non-trivial overhead on larger generated clients.

The only good way I see to address those concerns would be to put this functionality behind a new user option (something like `defaultOptionalArgsToNull`) and default it to true, which I think would work if we get this into the 5.0 release.

Then this change would become:

```
{{^required}}?{{#defaultOptionalArgsToNull}} = null{{/defaultOptionalArgsToNull}}{{/required}}
```

I don't know if `defaultOptionalArgsToNull` is the best name, though.

jimschubert:
cc kotlin technical committee @dr4ke616 @karismann @Zomzog @andrewemery @4brunu @yutaka0m

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
4brunu: I use the `sortParamsByRequiredFlag` to false, because of some bugs caused by the reorder of the parameters in the past, so I agree with the new flag, to ensure the correct behaviour of `sortParamsByRequiredFlag`.

isycat: Unless I have misunderstood something, there is no breakage within kotlim from adding = null. You can pass positioned params for a method that has defaulted params before others:


    private fun catOld(optionalParam: String?, requiredParam: Int, optionalParam2: String?): String {
        return "$optionalParam$requiredParam$optionalParam2"
    }

    private fun cat(optionalParam: String? = null, requiredParam: Int, optionalParam2: String? = null): String {
        return "$optionalParam$requiredParam$optionalParam2"
    }

    @Test
    fun testOptionalFirst() {
        // Without defaults, all 3 values must be specified
        assertEquals("123", catOld("1", 2, "3"))
        assertEquals("123", catOld("1", 2, optionalParam2 = "3"))
        assertEquals("123", catOld("1", requiredParam = 2, optionalParam2 = "3"))
        assertEquals("123", catOld(optionalParam = "1", requiredParam = 2, optionalParam2 = "3"))

        // With defaults, the same possible combinations are still valid
        assertEquals("123", cat("1", 2, "3"))
        assertEquals("123", cat("1", 2, optionalParam2 = "3"))
        assertEquals("123", cat("1", requiredParam = 2, optionalParam2 = "3"))
        assertEquals("123", cat(optionalParam = "1", requiredParam = 2, optionalParam2 = "3"))

        // plus:
        assertEquals("12null", cat("1", 2))
        assertEquals("12null", cat("1", requiredParam = 2))
        assertEquals("12null", cat(optionalParam = "1", requiredParam = 2))
        assertEquals("null23", cat(requiredParam = 2, optionalParam2 = "3"))
    }

(Was this a recent kotlin change?)

The Java overloads concern sounds legitimate (and therefore also the ability to turn the behavior off)

I am not sure when I will be able to squeeze in the new config param change, so I am going to separate the null defaulting from this PR, so as to not block the null string bug.

jimschubert: > (Was this a recent kotlin change?)

It's been around a while, but I recall being surprised by it when I started using Kotlin because it wasn't documented. The documentation was later [updated August 2017](https://github.com/JetBrains/kotlin-web-site/commit/c835ff2cd93b0b96f8fcb9b482a9cfe91e2c7c96#diff-fcd668d257886cdb850ee383b69170eb) to clarify the corner case. I think most people who started learning Kotlin before this doc update probably don't know it's even a thing, most likely because we're all using named args for clarity :)

jimschubert:
Looks good. Thanks for offering to address the default argument issue separately… it definitely makes the smaller reviews easier.

PR #7353, opened 2020-09-04
[kotlin-multiplatform-client] New, separate and refreshed

cromefire:
### PR checklist
 

cromefire:
Technical committee:

cromefire:
Some things to add (list may be expanded):
 - [ ] Fix up codegen
 - [ ] Client project mode (no gradle wrapper and settings)
 - [ ] Interop for the individual, non-kotlin languages (futures for java, promise for js)
 - [ ] Better ktor integration (remove unneeded request and response helpers and get more "native")
 - [ ] Lazy for individual clients in container client
 - [ ] More fitting initialization

cromefire:
I added the cli options, so if someone wants to have a look, if they're correctly implemented

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
```diff
@@ -0,0 +1,139 @@
...
+{{#operations}}
+@Suppress(
+    "unused",
+    "RemoveRedundantQualifierName",
+    "RemoveExplicitTypeArguments",
+    "RemoveRedundantCallsOfConversionMethods",
+    "DuplicatedCode",
+    "UnnecessaryVariable",
+)
+public class {{classname}} : ApiClientBase {
```
xsveda, line 36: It would be nice to have a conditional generation of an `open` modifier:
`public {{#openApiClass}}open {{/openApiClass}}class {{classname}} : ApiClientBase {`

**Motivation:** In our apps for every real `XyzApi` we create a `MockXyzApi` with mock responses and allow testers to switch the Api used at runtime. Its very useful either for development, testing or presentation purposes.

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
```diff
@@ -0,0 +1,139 @@
...
+     */
+    {{#returnType}}
+    @Suppress("UNCHECKED_CAST")
+    {{/returnType}}
+    public suspend fun {{operationId}}(
+        {{#allParams}}
+        {{{paramName}}}: {{{dataType}}}{{^required}}? = null{{/required}}{{#hasMore}},{{/hasMore}}
+        {{/allParams}}
+    ): HttpResponse<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}Unit{{/returnType}}> {
+        {{! TODO: Optimize all the ?.let to only the places where it's needed }}
```
xsveda, line 64: Until optimized, let's place one global `@file:Suppress("UNNECESSARY_SAFE_CALL")` to the top of the file.
We have many of them and Kotlin compiler output is is flooded by these warnings.

cromefire: Yeah but this will probably be gone until I declare it ready anyway, so I didn't add it just yet

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
cromefire: Yeah probably, I'll write that on the list of stuff that's still needed here

cromefire: This should only be needed for the mocking though and might be an option to implement platform specific async, I'd think, because most other use cases should be fine with extension functions

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
cromefire: Also I remember right now that this will be replaced with more "native" logic anyway if everything goes as planned, so that shouldn't be a problem. I just didn't have time yet to advance this

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
xsveda: Just to be clear, by mocking the apis I mean to create a subclass that overrides all the api functions. We are using these `MockXyzApi`s for testing of mobile apps running on real phone - mainly in app testing builds.
Although, sometimes we use these mocks also in production builds when users are allowed to try the app in some kind of a "demo" mode.

I've just realized that not only the api class but also the functions would need to be conditionally marked as `open`.

cromefire: Yeah but the models and helper classes should be fine. Also don't use this yet, it will probably have breaking changes, especially regarding the return types and maybe some other stuff, there's also no real android build currently, but JVM should work fine

cromefire:
I did not forget this, but I'll need ~5-6 Weeks until I have time to work on this again

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/data_class.mustache
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * {{{description}}}
+ *
+{{#allVars}}
+ * @param {{{name}}} {{{description}}}
+{{/allVars}}
+ */
+{{!TODO: Improve this?}}
```
kuFEAR, line 12: I have switch the array to list by default in KotlinClientCodegen.collectionType

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/interface_opt_var.mustache
```diff
@@ -0,0 +1,6 @@
+{{#description}}
+    /**
+     * {{{description}}}
+     */
+{{/description}}
+    @SerialName(value = "{{{vendorExtensions.x-base-name-literal}}}") {{>modelMutable}} {{{name}}}: {{#isEnum}}{{#isListContainer}}{{#isList}}kotlin.collections.List{{/isList}}{{^isList}}kotlin.Array{{/isList}}<{{classname}}.{{{nameInCamelCase}}}>{{/isListContainer}}{{^isListContainer}}{{classname}}.{{{nameInCamelCase}}}{{/isListContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}?
```
kuFEAR, line 6: https://github.com/OpenAPITools/openapi-generator/pull/6916/commits/c8d7395706893b5fefd1d258043b7a7ae8618555 a bit improve property templates for ability to add @Contextual or @Serializable(with = Adapter::class) annotations

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
tadfisher: Another option is to add an annotation like `@GeneratedApi` to the base class, then users needing open classes can apply the `kotlin-allopen` compiler plugin with something like:

```kotlin
allOpen {
    annotation("com.mypackage.infrastructure.GeneratedApi")
}
```

chris-hatton:
Hi @cromefire .  I came on a journey seeking OpenAPI / Multiplatform generation and, after looking at performing the changes myself, ended up here.  This is clearly already great progress; is there any way I can help to complete it?

cromefire:
I don't really think so I guess, there's not much left and I already have concrete ideas for it. I should be able to get it done in the next week's though.

fullkomnun:
Hey @cromefire I have been looking into your great work and tried to use it. 
I encountered some issues around Enum classes nested in data classes, 
will issue a PR on your fork so you can take a look at it. 

What are your ideas for platform specific interop?

cromefire:
> What are your ideas for platform specific interop?

Provide stuff like Futures for java or promises for JS, so you could also use the same client from a java/JS codebase instead of needing a kotlin shim or something like that. (For example if you publish a client for general use)

chris-hatton:
I have a burning use case for this. @fullkomnun did you find the current
fork state to be usable?

fullkomnun:
@chris-hatton Sorry for the late response. I also have a use-case that requires this support for the latest
Kotlin MP eco-system (with gradle Kotlin DSL, more recent releases of kotlinx.serialization, kotlinx.coroutines, ktor etc).

I have been testing this PR, mainly focusing on JVM/NodeJS as targets and it works great except for a few issues i have fixed and submitted for review [here](https://github.com/cromefire/openapi-generator/pulls).
I have not used the new 'kotlinx.datetime' yet though as my target API does not benefit from it...

I performed one more significant change which I will soon submit as a PR which allows clients to provide a custom HttpClient to ClientAPI. It is useful for using common Ktor features such as logging, timeouts and so on.
@cromefire would love to get your feedback =]

cromefire:
It should be already possible to use a custom `HttpClient` though.
Would also be great (if easily doable) could break it up into multiple PRs (I haven't looked at it yet, but it sounded like it's a lot of changes) so we can look at it on a feature by feature basis.

cromefire:
> What are your ideas for platform specific interop?

I've added interop for JS Promises now, [here](https://github.com/cromefire/openapi-generator/tree/multiplatform-client/samples/client/petstore/kotlin-multiplatform-new-async/src/js/main/org/openapitools/client) is the sample

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/ApiAbstractions.kt.mustache
```diff
@@ -46,6 +46,9 @@ public class Queries(public val queries: MutableMap<String, QueryParam> = mutabl
         addMulti(name, values.asIterable(), collectionFormat)
 
     // For Maps
+    public fun addMulti(name: String, values: Map<String, *>, collectionFormat: String): Unit =
+        queries.putAll(values.filterValues { it != null }.mapValues { QueryParam.Single(it.value.toString()) })
```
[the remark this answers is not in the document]
justasm: The `hasMore` tag was also not recognized previously, but I see that you got rid of it in the latest commit 95b1f4a 🙂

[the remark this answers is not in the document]
cromefire: Yeah me neither this comes from the code I "forked", I'll try that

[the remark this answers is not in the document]
cromefire: > The `hasMore` tag was also not recognized previously, but I see that you got rid of it in the latest commit 95b1f4a :slightly_smiling_face:

Yeah, I'm just using trailing comma, works better.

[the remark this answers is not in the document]
cromefire: Fixed

samples/client/petstore/kotlin-multiplatform-new/src/common/main/org/openapitools/client/models/Order.kt
```diff
@@ -0,0 +1,60 @@
...
+@Serializable
+public data class Order(
+    @SerialName(value = "id")
+    val id: kotlin.Long? = null,
+    @SerialName(value = "petId")
+    val petId: kotlin.Long? = null,
+    @SerialName(value = "quantity")
+    val quantity: kotlin.Int? = null,
+    @SerialName(value = "shipDate")
+    val shipDate: kotlinx.datetime.LocalDateTime? = null,
```
justasm, line 42: Using the default `dateLibrary`, or specifying `string`, still maps `date-time` to `kotlinx.datetime.LocalDateTime`. This results in an invalid model that does not compile because of a missing `UseSerializers` - but there should be no reference to `kotlinx.datetime` at all, the generated property should be `val shipDate: String? = null`.

I've run into this same issue running this generator locally, but have not managed to pin down the issue yet.

cromefire: Interesting...

cromefire: Yeah I was writing too much kotlin, forgot a `break`

cromefire: Is switched into the `STRING` case, but fell through and updated everything to `kotlinx`

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinMultiplatformClientCodegen.java
```diff
@@ -0,0 +1,618 @@
...
+                typeMapping.put("date", "String");
+                typeMapping.put("Date", "String");
+                typeMapping.put("DateTime", "String");
+                importMapping.put("Timestamp", "kotlin.String");
+                importMapping.put("LocalDateTime", "kotlin.String");
+                importMapping.put("LocalDate", "kotlin.String");
+                importMapping.put("LocalTime", "kotlin.String");
+            }
+            case KOTLINX: {
+                typeMapping.put("date-time", "LocalDateTime");
```
justasm, line 443: Mapping the `date-time` field to `kotlinx.datetime.LocalDateTime` may not be appropriate.

[`date-time` corresponds to RFC3339](http://spec.openapis.org/oas/v3.0.3#data-types) which includes the time zone offset, e.g. `2020-07-31T07:16:15Z`, `2020-07-31T09:16:15+02:00`.

`kotlinx.datetime.LocalDateTime` does not include any time zone offset, and in fact `LocalDateTime.parse(..)` fails to parse both of the above examples.

See a similar issue and discussion which changed [`java.time.LocalDateTime` to `java.time.OffsetDateTime`](https://github.com/swagger-api/swagger-codegen/pull/3015).

Unfortunately, `kotlinx.datetime` does not currently have an analogue for `java.time.OffsetDateTime`. `kotlinx.datetime.Instant` currently maps to / from only the first example time above. One potential approach that comes to mind is to map to a custom `OffsetDateTime` type until `kotlinx.serialization` adds support for common formats such as ISO 8601 / RFC 3339.

```kotlin
class OffsetDateTime(
  val localDateTime: kotlinx.serialization.LocalDateTime,
  val offset: kotlinx.serialization.ZoneOffset,
)
```

samples/client/petstore/kotlin-multiplatform-new/src/common/main/org/openapitools/client/models/Order.kt
justasm: Ah, nice find! Reading too much Kotlin myself too 😄

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/KotlinMultiplatformClientCodegen.java
cromefire: That is indeed [correct](https://github.com/Kotlin/kotlinx-datetime#type-use-cases)

cromefire: But maybe an instant would be more suitable here?

cromefire: Yeah something like a custom offset would probably be best, but I guess that means I have write some parsing...

cromefire: I hadn't checked this at all (just copied it over and changed it from Java to kotlinx), so feel free to check it.

justasm: I think both `Instant` and `Pair<LocalDateTime, ZoneOffset>` are suitable, but yeah, the first should be simpler to implement. The only potential downside with `Instant` I see is that in a round-trip server-client-server, the time zone information is lost.

`kotlinx.datetime.Instant.parse("2020-07-31T09:16:15+02:00")` currently fails (as of `kotlinx.datetime` version 0.1.1), however, so we'd need custom parsing logic regardless.

justasm: [The Java implementation ran into a similar conundrum](https://github.com/swagger-api/swagger-codegen/issues/2138#issuecomment-224254752)

cromefire: Yes the Zoning info probably requires us to use some offset stuff, I'm investigating writing a regex

cromefire: We can easily add a method to convert it to an instant, so the loss of information is optional

justasm: [This is something I threw together quickly when trying to investigate this.](https://gist.github.com/justasm/d34c06411940643874eda051e72ce2e4)

cromefire: I'll also raise an issue with them

cromefire: > [This is something I threw together quickly when trying to investigate this.](https://gist.github.com/justasm/d34c06411940643874eda051e72ce2e4)

Seems well enough, although the last case should probably raise an error

cromefire: Not complete yet and missing some documentation, but a start I guess: https://gitlab.com/-/snippets/2056386
(not tested yet, feel free to test it)

cromefire: Fixed, please test if that works for you

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
```diff
@@ -0,0 +1,89 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+/**
+ * Represents a [LocalDateTime] and the respective [TimeZone] of it.
+ */
+@Serializable(with = OffsetDateTime.Companion::class)
+public data class OffsetDateTime(val dateTime: LocalDateTime, val timeZone: TimeZone) {
```
justasm, line 16: Store a `LocalDateTime` and `ZoneOffset` instead - a `TimeZone` can represent more complex zone id representations such as "Europe/Paris" or "UTC+3". A `ZoneOffset`, on the other hand, is exactly what is represented in the RFC3339 / ISO8601 format. This is in line with [`java.time.OffsetDateTime`](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html), which is likely what `kotlinx.datetime` would use as a reference if implementing a similar class.

In `kotlinx.datetime` the `ZoneOffset` can be retrieved from the `TimeZone` at an `Instant`, so in the case where we have a `LocalDateTime` and `TimeZone`

```kotlin
val zoneOffset = with(timeZone) { offsetAt(localDateTime.toInstant()) }
```

justasm: `kotlinx.datetime` is just a bit too cutting-edge 🙂 [They have this open issue of parsing & formatting common interchange formats on the roadmap](https://github.com/Kotlin/kotlinx-datetime#knownopen-issues-work-tbd).

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Converts the [OffsetDateTime] to an [Instant]. This looses the [TimeZone] information, because the date and time
+     * is converted to UTC in the process.
+     */
+    public fun toInstant(): Instant = dateTime.toInstant(timeZone)
+
+    /**
+     * Creates an [OffsetDateTime] from an [Instant] in a given [TimeZone] ([TimeZone.UTC] by default).
+     */
+    public fun fromInstant(instant: Instant, timeZone: TimeZone = TimeZone.UTC): OffsetDateTime = OffsetDateTime(
```
justasm, line 34: This does not use the instance and would make more sense to be declared in the `companion object`.

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
```diff
@@ -0,0 +1,89 @@
...
+    public fun inTimeZone(newTimeZone: TimeZone): OffsetDateTime {
+        val instant = dateTime.toInstant(timeZone)
+        val newDateTime = instant.toLocalDateTime(newTimeZone)
+        return OffsetDateTime(newDateTime, newTimeZone)
+    }
+
+    /**
+     * Returns a new [OffsetDateTime] with the time zone [TimeZone.UTC].
+     */
+    public fun utc(): OffsetDateTime = inTimeZone(TimeZone.UTC)
```
justasm, line 51: Is there a use-case for this not covered by `toInstant(): Instant`? `Instant` is already intended to represent UTC timestamps.

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
cromefire: Yeah you're right

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
cromefire: Ehm, yeah this isn't supposed to be on the instance...

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
cromefire: Well this can be used to get a local utc time for example, you can get the same info from the instant too, but this would be a shortcut.

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/common/main/infrastructure/OffsetDateTime.kt.mustache
cromefire: So, the problem is that the offset isn't really something you can parse... (no `of()` or any kind of public constructor)

cromefire: Okay changed it, but for now it's a bit ugly until Kotlin/kotlinx-datetime#91 is addressed

cromefire:
So the last thing to get it actually compiling, at least for my schema, is the handling of `additionalProperties`. This model:
```json
"ProblemDetails": {
  "type": "object",
  "properties": {
    "type": {
      "type": "string",
      "nullable": true
    },
    "title": {
      "type": "string",
      "nullable": true
    },
    "status": {
      "type": "integer",
      "format": "int32",
      "nullable": true
    },
    "detail": {
      "type": "string",
      "nullable": true
    },
    "instance": {
      "type": "string",
      "nullable": true
    }
  },
  "additionalProperties": {
    "type": "object",
    "additionalProperties": false
  }
}
```
is being transformed into:
```kotlin
@Serializable
public data class ProblemDetails(
    @SerialName(value = "type")
    val type: kotlin.String? = null,
    @SerialName(value = "title")
    val title: kotlin.String? = null,
    @SerialName(value = "status")
    val status: kotlin.Int? = null,
    @SerialName(value = "detail")
    val detail: kotlin.String? = null,
    @SerialName(value = "instance")
    val instance: kotlin.String? = null
) : kotlin.collections.HashMap<String, kotlinx.serialization.json.JsonObject>
```
which obviously won't work.

So it would really help if anyone...
1) ...has ideas how this can be implemented.
2) ...more importantly (this is where I'm really stuck) knows how to detect this in the mustache template and which properties are available.

My ideas for 1) are currently:
- Add a property of a map type that includes the addition properties
- Create a parent class that behaves like a map

In both cases serialization will be rather... challenging.

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/data_class.mustache
cromefire: Yeah this also uses list now

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/api.mustache
cromefire: This might also be fixed now, at least I don't see warnings generated in my client

modules/openapi-generator/src/main/resources/kotlin-multiplatform-client/interface_opt_var.mustache
cromefire: I don't quite understand what you are getting at

fullkomnun:
For platform specific APIs, it could have been great to be able to just do:
common:
`expect class Later<T>`
JS:
`actual typealias Later<T> = Promise<T>`
JVM:
`actual typealias Later<T> = Deferred<T>`

But it is not [supported at the moment](https://youtrack.jetbrains.com/issue/KT-21846#focus=Comments-27-4480739.0-0) so i guess generating platform specific API sub-classes is fine.

I suppose that if the consumer is itself MP 'suspend' based API is preferred.
Otherwise, Promise<T> for JS and either Deferred<T> or CompletableFuture<T> for general JVM 
or reactive primitives for specific use-cases...

cromefire:
Because I've recently learned, that consuming kotlin libraries directly from js seems to be hard/impossible, I'm currently considering to just use generate async common code, with the use of `Deferred<T>`, so that users can use the [integrations](https://github.com/Kotlin/kotlinx.coroutines/blob/master/integration/README.md) themselves to convert it to their preferred primitive (I'll have to test whether you can use the integrations correctly from Java, otherwise I'll write a little wrapper or so).

cromefire:
So this actually works (not the nicest thing, but it's Java anyway):
```java
import com.example.ApiClient;
import com.example.infrastructure.OffsetDateTime;

import static kotlinx.coroutines.future.FutureKt.asCompletableFuture;

public class Main {
    public static void main(String[] args) {
        var client = new ApiClient();
        var def = client.getActivityLog().getLogEntriesAsync(0, 0, OffsetDateTime.parse(""), false);
        var fut = asCompletableFuture(def);
        fut.thenAccept((resp) -> System.out.println(resp.getStatus()));
    }
}
```

There's probably only some `@JvmStatic`s missing here and there.

cromefire:
`HttpResponse` does not work yet, but the plan is to get rid of it anyway

fullkomnun:
@cromefire Currently, nested enums in data classes are missing 'public' modifier for both enum class itself and it's 'value' property so explicitApi mode is not satisfied

cromefire:
> @cromefire Currently, nested enums in data classes are missing 'public' modifier for both enum class itself and it's 'value' property so explicitApi mode is not satisfied

Fixed

cromefire:
So I just did some major rewriting of the request logic (removed a lot of generic stuff because it's only ktor anyway). Authentication isn't there yet, but I'd be glad if some of you could test if it still builds for your schemas.

cromefire:
I'm also starting to reach the limits of mustache, because I'm not able to call custom functions or similar things.

fullkomnun:
@cromefire The latest changes, cause multiple build issues:

- Some generated api that uses form as body is missing an import for 'FormDataContent'
and uses an 'it' reference instead of 'this' when trying to build it with 'apply' kotlin extenstion:
```kotlin
            this.body = FormDataContent(
                ParametersBuilder(6).apply {
                    if (name != null) {
                        it.append("name", name)
                    }
                    if (loginName != null) {
                        it.append("loginName", loginName)
                    }
                    if (status != null) {
                        it.append("status", status)
                    }
                    if (role != null) {
                        it.append("role", role)
                    }
                    if (email != null) {
                        it.append("email", email)
                    }
                    if (locked != null) {
                        it.append("locked", locked.toString())
                    }
                }.build()
```

- all auth infrastructure classes fail to resolve `import com.something.infrastructure.Queries`

- ApiClientBase: 
```kotlin
    private fun RequestConfig.updateForAuth(authNames: List<String>) {
        for (authName in authNames) {
            val auth = authentications[authName] ?: throw Exception("Authentication undefined: $authName")
            auth.apply(queries, headers)
        }
    }
```
reports 'Unresolved reference: RequestConfig' and 
'Too many arguments for public inline fun <T> T.apply(block: T.() -> Unit): T defined in kotlin'

- 2 infrastructure files have non-standard naming (starts with lowercase letter):
'bytes.kt' and 'requestHelpers.kt'

cromefire:
> @cromefire The latest changes, cause multiple build issues:
> 
> * Some generated api that uses form as body is missing an import for 'FormDataContent'
>    and uses an 'it' reference instead of 'this' when trying to build it with 'apply' kotlin extenstion:
> 
> 
> ```kotlin
>             this.body = FormDataContent(
>                 ParametersBuilder(6).apply {
>                     if (name != null) {
>                         it.append("name", name)
>                     }
>                     if (loginName != null) {
>                         it.append("loginName", loginName)
>                     }
>                     if (status != null) {
>                         it.append("status", status)
>                     }
>                     if (role != null) {
>                         it.append("role", role)
>                     }
>                     if (email != null) {
>                         it.append("email", email)
>                     }
>                     if (locked != null) {
>                         it.append("locked", locked.toString())
>                     }
>                 }.build()
> ```
> 
> * all auth infrastructure classes fail to resolve `import com.something.infrastructure.Queries`
> 
> * ApiClientBase:
> 
> 
> ```kotlin
>     private fun RequestConfig.updateForAuth(authNames: List<String>) {
>         for (authName in authNames) {
>             val auth = authentications[authName] ?: throw Exception("Authentication undefined: $authName")
>             auth.apply(queries, headers)
>         }
>     }
> ```
> 
> reports 'Unresolved reference: RequestConfig' and
> 'Too many arguments for public inline fun T.apply(block: T.() -> Unit): T defined in kotlin'
> 
> * 2 infrastructure files have non-standard naming (starts with lowercase letter):
>   'bytes.kt' and 'requestHelpers.kt'

First 2 are fixed, 3 will be gone once I implement auth, 4th is correct like this because the files don't contain classes (official libraries follow the same schema)

cromefire:
3 is also fixed now, although the configuration is still incomplete

chris-hatton:
Hi @cromefire I built this PR branch and tried it out on the `swagger.json` for my use case, which is [here](https://barntools-d-api.azurewebsites.net/swagger/barn360/swagger.json).

The generation completes, but the resulting code does not compile.  Output attached in case this helps.
[output.txt](https://github.com/OpenAPITools/openapi-generator/files/5787243/output.txt)

cromefire:
Could you give me some screenshots of the errors? I see a lot of bad imports there, so that would make it a lot easier

chris-hatton:
Problems with this run break down into:

1. Bad imports/ordering around Kotlin date/time:
Screen Shot 2021-01-09 at 9 10 51 am

2. A parameter badly interpreted from the OpenAPI spec:
Screen Shot 2021-01-09 at 9 09 58 am

The Swagger JSON specifies a **header** value which is missing, and instead we see an `UNKNOWN_PARAMETER_NAME`.  This happens for several API calls, wherever the **header** variable is present:
Screen Shot 2021-01-09 at 9 21 34 am

3. Use of '`AnyType`?
Screen Shot 2021-01-09 at 9 14 25 am

chris-hatton:
I tested this using the `java` generator and the `UNKNOWN_PARAMETER_NAME` for my `swagger.json` is also a problem there, so that's not a bug in this PR; probably a bug in [`DefaultCodegen.java`](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java), or out-of-spec Swagger 🤔

cromefire:
Okay 1 just seems to be some bad conditions, I'll look at that, depending on how obvious the errors turn out to be, the options used (if any) might be useful.
Fot 2: `AnyType` should already be configured, I'll investigate why that doesn't work, could you get me the schema of that model, so I can look if there's something that hasn't been taken into account yet?

cromefire:
Also judging from the screenshots you're also using an outdated version of the branch, so some (all?) of those problems will probably go away if you generate using the latest version.

chris-hatton:
Yes; very silly mistake, I was using a code revision in my own fork of your PR which was out of date.
Anyway, I updated that but instead get the following error, on the same generation:
```
Exception in thread "main" java.lang.RuntimeException: Could not generate supporting file 'SupportingFile[templateFile='common/main/infrastructure/bytes.kt.mustache', folder='src/common/main/org/openapitools/client/infrastructure', destinationFilename='bytes.kt', canOverwrite=true]'
	at org.openapitools.codegen.DefaultGenerator.generateSupportingFiles(DefaultGenerator.java:726)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:882)
Caused by: org.openapitools.codegen.templating.TemplateNotFoundException: common/main/infrastructure/bytes.kt.mustache
	at org.openapitools.codegen.TemplateManager.getFullTemplateFile(TemplateManager.java:54)
	at org.openapitools.codegen.TemplateManager.getFullTemplateContents(TemplateManager.java:72)
	at org.openapitools.codegen.templating.MustacheEngineAdapter.compileTemplate(MustacheEngineAdapter.java:59)
	at org.openapitools.codegen.TemplateManager.write(TemplateManager.java:163)
	at org.openapitools.codegen.DefaultGenerator.processTemplateToFile(DefaultGenerator.java:1017)
	at org.openapitools.codegen.DefaultGenerator.generateSupportingFiles(DefaultGenerator.java:718)
	... 4 more
```

darizhap:
Hi @cromefire
Let me report a few issues here. I'm attaching the OpenApi file to this post. It fully complies with the specification. However, the http-client that is generated with this scheme cannot be compiled. Here is a list of compilation errors.

[openapi.yml.txt](https://github.com/OpenAPITools/openapi-generator/files/5835382/openapi.yml.txt)
[errors.txt](https://github.com/OpenAPITools/openapi-generator/files/5835393/errors.txt)

cromefire:
> Unfortunately, issue reporting is disabled in your repository. Therefore, I am writing here.

Yeah I believe this is the right place, my fork is only for the purpose of this PR and it also improves visibility

darizhap:
If you need help fixing these issues, please tell me how.

cromefire:
Ill probably just need a bit of time, need to setup the openapi spec, generate it, fix it and so ob

cromefire:
Can you tell me which flags you set?

cromefire:
So I fixed 2 of the issues, I have no idea for the last one...
image
It's missing the generic type, no idea what's happening there...

darizhap:
> Can you tell me which flags you set?

I just run this command:
java -jar openapi-generator-cli.jar generate -i openapi.yml -g kotlin-multiplatform -o client-new

darizhap:
> So I fixed 2 of the issues, I have no idea for the last one...
> It's missing the generic type, no idea what's happening there...

It works just fine, if you write the definition of that property like this:
```
items:
  $ref: '#/components/schemas/ListItem'
```

But the error appears, if you write:
```
items:
  allOf:
    - $ref: '#/components/schemas/ListItem'
    - description: "List of items in order"
```

cromefire:
Yeah the problem is, this is coming somewhere from the depth of openapi gen and I have absolutely no idea how to fix that, I only adapted some stuff and have no deeper insight how any of this works behind the scenes.

darizhap:
> Yeah the problem is, this is coming somewhere from the depth of openapi gen and I have absolutely no idea how to fix that, I only adapted some stuff and have no deeper insight how any of this works behind the scenes.

Great work, anyway!

cromefire:
I declare this "feature complete" from my point of view. So now only the 2 errors that are left need to get fixed (which I don't know how, so if someone has any ideas...) and if someone wants to add tests that'd probably a good thing too.

cromefire:
Travis tests fail because of rate limits: `toomanyrequests: You have reached your pull rate limit.`

wing328:
Restarted the failed jobs. Let's see how that goes.

cromefire:
Shippable now has some APT PPA related errors and warnings

Ingwersaft:
Awesome PR, really looking forward to the merge :)

I was trying it out and got the following error (in my project, but also in the sample: `openapi-generator/samples/client/petstore/kotlin-multiplatform-new`):

```
➜  kotlin-multiplatform-new git:(multiplatform-client) ✗ ./gradlew build
Downloading https://services.gradle.org/distributions/gradle-6.7.1-bin.zip
.........10%..........20%..........30%..........40%..........50%.........60%..........70%..........80%..........90%..........100%
Starting a Gradle Daemon (subsequent builds will be faster)

> Configure project :
Kotlin Multiplatform Projects are an Alpha feature. See: https://kotlinlang.org/docs/reference/evolution/components-stability.html. To hide this message, add 'kotlin.mpp.stability.nowarn=true' to the Gradle properties.

Warning: The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.
e: /Users/mk/workspace/openapi-generator/samples/client/petstore/kotlin-multiplatform-new/build.gradle.kts:43:5: Unresolved reference: native

* Where:
Build file '/Users/mk/workspace/openapi-generator/samples/client/petstore/kotlin-multiplatform-new/build.gradle.kts' line: 43

* What went wrong:
Script compilation error:

  Line 43:     native()
               ^ Unresolved reference: native

1 error

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.7.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 33s
➜  kotlin-multiplatform-new git:(multiplatform-client) ✗
```
Looks like there is [no native default target available](https://kotlinlang.org/docs/reference/mpp-supported-platforms.html).

Also you can [skip the kotlin-stdlib imports since 1.4](https://kotlinlang.org/docs/reference/migrating-multiplatform-project-to-14.html#simplify-your-build-configuration).

cromefire:
Oh yeah, right, native is still broken, I totally forgot about that

Ingwersaft:
Another problem I'm facing while using the subproject generation:
```
configOptions.set(
        mapOf(
            "dateLibrary" to "kotlinx",
            "async" to "false",
            "iosEnabled" to "true",
            "jsEnabled" to "true",
            "jsBrowser" to "true",
            "subproject" to "true",
            "packageName" to "com.liftric.portal.client"
        )
    )
```
Gradle errors with:
```
* What went wrong:
Could not create an instance of type org.jetbrains.kotlin.gradle.targets.js.subtargets.KotlinBrowserJs.
> Failed to apply plugin class 'org.gradle.language.base.plugins.LifecycleBasePlugin'.
   > Cannot add task 'build' as a task with that name already exists.
```

I found some slightly related youtrack issues with the workaround being to declare the plugins in the root project (with apply false), but I'm already doing it. Your generated subproject doesn't contain any version, so should be correct as far as I can tell, but maybe you have an Idea how to fix this?

With jsEnabled and jsBrowser both on false the error disappears.

Ingwersaft:
Another thing which would be awesome: Enable typescript definitions generation for the kotlin js IR compiler.

The js target would need to be executable
```
js(BOTH) {
        browser {
            binaries.executable()
        }
}
```
and all files containing toplevel stuff would need the JsExport treatment (otherwise no definitions will be generated):
```
@file:OptIn(ExperimentalJsExport::class)
@file:JsExport

import kotlin.js.ExperimentalJsExport
import kotlin.js.JsExport
[...]
```

cromefire:
> I found some slightly related youtrack issues with the workaround being to declare the plugins in the root project (with apply false), but I'm already doing it. Your generated subproject doesn't contain any version, so should be correct as far as I can tell, but maybe you have an Idea how to fix this?

Yeah you can only set a version once in the whole project so this why I did it that way, schools probably document that.

cromefire:
> Another thing which would be awesome: Enable typescript definitions generation for the kotlin js IR compiler.

I think we need some gradle extension script or so but I didn't get it to work yet.

Nutriz:
Hi, nice job. I have a problem with enums.

I tried to export the petstore sample and in models I have an unresolved reference for `CommonEnumSerializer`. I got hard time to find information about this class ! Seems to be renamed/removed since a [long time ago](https://github.com/Kotlin/kotlinx.serialization/commit/52c8290adea26b7c1aca4ad26666a19300564978#diff-181c69e85cc82a76fe041df047e2ff3980f71e0bef4e9eef9d1d869716f140dbL56) and you don't need it anymore. Enums are working for you ? I used your last commit `39cafb6db6 2021-01-28 Merge branch 'master' into multiplatform-client`

image

cromefire:
Are you sure it didn't cache anything or so while building? Because there isn't any reference to that in the code (anymore)

cromefire:
Maven seems really bad at caching so I think you do something like `clean build`

Nutriz:
Super thanks ! I will try again after a clean !

Ingwersaft:
> > Another thing which would be awesome: Enable typescript definitions generation for the kotlin js IR compiler.
> 
> I think we need some gradle extension script or so but I didn't get it to work yet.

I got it to work, but a lot of generated code issn't exportable to JS, so this would need some major changes to the generated client (with `@file:JsExport`):

```
Declaration of such kind (enum class) cant be exported to JS
Declaration of such kind (secondary constructor without @JsName) cant be exported to JS
Declaration of such kind (suspend function) cant be exported to JS
```

cromefire:
Well without enums and suspend functions you're probably not getting far, I've also tried some things there and settled that for a client to use from ts, you probably want to generate a "real" ja/ts client.

darizhap:
Hi,
Do you have any plans to make use of Kotlin multiplatform [date-time](https://github.com/Kotlin/kotlinx-datetime) library? It would be great if the generator could recognize definitions like this:
```
properties:
  someInstantProperty:
    format: date-time
    type: string
```
and generate properties or parameters of type [Instant](https://github.com/Kotlin/kotlinx-datetime/blob/master/core/common/src/Instant.kt).

cromefire:
Actually yes, the one I remember I couldn't figure out happens if you specify `additionalProperties` (or something similar, I can't look it up right now). The old code sends it to extend `HashMap` which doesn't work, because it's final, so there needs to be some other mechanism that deals with the unknown properties there and put them into a map or something similar (probably some custom serializer?)

xxfast:
Any update on this?

cromefire:
Nope, still those same problems and as I'm quite busy right now I haven't spent a lot of time on trying to find a solution for it.

silvaignacio:
Some update on this @cromefire , as the Android Client is quite outdated and it would be interesting to have Multiplatform support with Kotlin.

cromefire:
As always, there wasn't a lot of time to spare and too else much to do, there's not a lot of issues with the PR in it's current state except `additionalProperties` and it probably needs a Kotlin and Ktor update or two...

Also thought of maybe rewriting it without openapi-generator and Kotlin poet or so for more native code generation and less issues with templating... (as openapi-generator doesn't provide too much I think except the built-in templating) Maybe they'd even accept this generator not using tempating, but a code generator, but I don't think so and it would also add a dependency to the generator in the end (one that's not needed by most) and I don't think there's a plugin mechanism.

If you have a good solution to the issue with `additionalProperties` we might be able to get this over the finish line. For me it seems like I'd have to do the serialization manually in that case, for the whole object...

Overall I really could use this myself and I'd love to have this but I just didn't have the time and motivation yet to finish it.

wing328:
> If you have a good solution to the issue with additionalProperties we might be able to get this over the finish line

java, C# ,etc supports the additionalProperties. maybe we can these implementations as a starting point.

I think we can cherry-pick some of the enhancements in https://github.com/OpenAPITools/openapi-generator/pull/7353#issuecomment-687434057 to start with so that we can merge these enhancements separately.

If anyone wants to help on this, please reply to let us know.

most importantly thank you @cromefire for the awesome contributions 👍

cromefire:
> java, C# ,etc supports the additionalProperties. maybe we can these implementations as a starting point.

AFAIK they usually do that by inheriting from a list, which isn't doable in kotlinx.serialization last time I checked (3 years ago). In general it can of course be done, but it's a bit of work, especially since I have to do it in templating. I might have some time next week, I just have to remember to look at this again.

kroegerama:
Hi there, I have been following this branch for a long time, because I am also planning to write an OpenAPI generator for Kotlin multiplatform (or rather: rewrite my [existing one](https://github.com/kroegerama/openapi-kgen/) to KMM, instead of Retrofit+Moshi).

My plan is to just use `JsonElement` for unknown schemas.

The value `additionalProperties` can have two different values:
- `true` -> Use `JsonElement`. It is a "free-form" object
- schema XXX
  - if no own properties: use `Map<String, XXX>`
  - else: Use `JsonElement` to be sure, everything will be parsed

Issue #7361, opened 2020-09-06
[BUG] FreeForm object models with validations are not generated


spacether:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issuue still exists?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

FreeForm object models with validations are not generated
When I try to generate the model I see the message:
[main] INFO  o.o.codegen.DefaultGenerator - Model ObjectWithValidations not generated since it's a free-form object
But I should be able to generate and use ObjectWithValidations.

It requires that the payload be a pojo  {}
with a minimum of one property and that it accepts AnyType as the value of that property.
So this payload would pass validation:
{"hi": "there"}
and this one would fail validation:
{}
Because it was missing the minimum number of properties, 1.

##### openapi-generator version

5.0.0-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
  description: >-
    This spec is mainly for testing Petstore server and contains fake endpoints,
    models. Please do not use this for any other purpose. Special characters: "
    \
  version: 1.0.0
  title: OpenAPI Petstore
  license:
    name: Apache-2.0
    url: 'https://www.apache.org/licenses/LICENSE-2.0.html'
tags: []
paths: {}
components:
  schemas:
    ObjectWithValidations:
      type: object
      minProperties: 1
```

##### Generation Details

Generate the client in any language

##### Steps to reproduce

Generate the client in any language

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/issues/5387

##### Suggest a fix

Allow generation of free form objects with validations.
If a composed schema $ref uses a component that is a composed schema with validations, right now in master branch, that component model is not being generated.

PR #7373, opened 2020-09-08
Adds free form model generation in python-experimental

spacether:
If a free form object schema has validations I think that we should generate it as a model
- I think that we should do this to allow another composed schema to oneOf include an object schema with validations
- In the current master branch, those validations are omitted because there will not be a model to contain the validations

If a free form schema is used as an interface schema (another schema allOf includes it) I think that we should generate the interface as a model
- Note: this is another way of implementing interface models which is similar to https://github.com/OpenAPITools/openapi-generator/pull/6849 but doesn't require our users to know about a specific vendor extension.

If merged, this will resolve https://github.com/OpenAPITools/openapi-generator/issues/7361 for python-experimental only
- Tests added for these free form schema use cases

#### How this change was made:
- unaliasSchema is changed into a public method in each generator
- when generating models, if we are examining a FreeFormObject, we create a schema ref to it and see if unaliasSchema preserves the ref(meaning the model will be generated) or it unaliases the schema (meaning the model won't be generated)
- the method hasValidation is added to codegenModel to return a boolean saying if there is validation in the root level of the component schema that we are examining. This is needed to check if arrayModels have validations in all generators that make arraymodels. This is also needed in python-experimental to produce FreeFormObject models with validation.

### PR checklist
 
Python Technical Committee


wing328:
@spacether when you've time, can you please resolve the merge conflicts?

spacether:
> @spacether when you've time, can you please resolve the merge conflicts?

@wing328 merge conflicts have been resolved. Thank you for letting me know about them.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2491,6 +2491,9 @@ public CodegenModel fromModel(String name, Schema schema) {
                 } else { // type is number and without format
                     m.isNumber = Boolean.TRUE;
                 }
+            } else if (ModelUtils.isFreeFormObject(openAPI, schema)) {
+                addAdditionPropertiesToCodeGenModel(m, schema);
+                ModelUtils.syncValidationProperties(schema, m);
```
spacether, line 2496: @wing328 should we set `m.isMapModel = true;` here also?

wing328: Good question. free-form object is a [sub-type of dictionary](https://swagger.io/docs/specification/data-models/dictionaries/) so it may make sense to set `isMapModel` to true but my only concern is how it will impact other generators (e.g. aspnetcore, csharp-netcore, etc)

I would recommend setting it in another PR to see how it impacts other generators.

PR #7379, opened 2020-09-09
[cpp-ue4] add support for Content-Type containing charset

Falanwe:
Change to the api template for cpp-ue4 to handle a composite Content-Type (for instance application/json; charset=utf-8) by comparing the start of the string instead of looking for a perfect match.

Sadly, there are no test config for cpp-ue4 (yet?). I tested the change directly inside the UE4 project we have in which we are using openapi-generator (as it's a proprietary game I cannot share it obviously, but the resulting code did compile well).

@ravinikam @etherealjoy @MartinDelille @muttleyxd @Kahncode

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
@@ -56,11 +56,11 @@ void {{classname}}::HandleResponse(FHttpResponsePtr HttpResponse, bool bSucceede
 		FString ContentType = HttpResponse->GetContentType();
 		FString Content;
 
-		if(ContentType.IsEmpty())
+		if (ContentType.IsEmpty())
 		{
 			return; // Nothing to parse
 		}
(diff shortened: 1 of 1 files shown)


wing328:
@Falanwe thanks for the PR. 

Later we may want to use a regular expression to handle more edge cases, e.g. https://github.com/OpenAPITools/openapi-generator/blob/5693cb98635b04634ba875abfe95b0dbd7d69e42/samples/client/petstore/csharp/OpenAPIClient/src/Org.OpenAPITools/Client/ApiClient.cs#L364

We can go with your suggested fix for the time being.

Falanwe:
I'm not really familiar with using regexes in UE4 (I'm more of a C# programmer usually, but I think it would be a good idea to use them indeed). I'll probably create a new pull request with that when I have a little time, getting inspiration from C#'s regex (hopefully the regex syntax is the same between the two, at least for this specific regex).

Issue #6829, opened 2020-07-01
[BUG][GO] openapi-generator generates invalid Go code for response code ranges (e.g. 5XX)


edganiukov:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description
For response code ranges (e.g. 1XX, 2XX, 5XX) it generate invalid Go code, for example:
```go
        if localVarHTTPResponse.StatusCode == 404 {
            // ...
        }
        if localVarHTTPResponse.StatusCode == 5XX {  // <-- BUG is here
            // ...
        }
```

##### openapi-generator version
v4.3.1

##### OpenAPI declaration file content or url
```yaml
openapi: 3.0.3
info:
  version: '2.0.0'
  title: API
  description: Description
servers:
- {url: 'https://api-server/v2'}
security:
- bearerAuth: []
paths:
  /login:
    post:
      tags:
        - auth
      summary: Authenticate
      operationId: authenticate
      responses:
        200:
          description: OK
        400:
          description: Bad request. User ID must be an integer and larger than 0.
        401:
          description: Authorization information is missing or invalid.
        404:
          description: A user with the specified ID was not found.
        5XX:
          description: Unexpected error.
      security:
        - basicAuth: []
```

##### Command line used for generation
openapi-generator-cli  generate --generator-name go \
  --input-spec ${OAS_SPEC_FILE} \
  --api-package ${GO_API_PACKAGE} --package-name ${GO_PACKAGE} \
  --git-host github.com --git-user-id exampleuser --git-repo-id ${GO_ARTIFACT} \
  --output "${GO_ARTIFACT}"

wing328:
I don't think 5xx is a valid status code. You need to provide something like 500 instead.

edganiukov:
@wing328 
according to the doc it is a valid response status code for openapi3 - https://swagger.io/docs/specification/describing-responses
[...]

PR #7381, opened 2020-09-09
Add wildcardFirstChar for implicit response code

wing328:
Add wildcardFirstChar for implicit response code to support 4XX, 5XX, etc

wildcardFirstChar is set to 4 if it's 4XX for example and is set to null if the response code is not implicit (e.g. 404)

For https://github.com/OpenAPITools/openapi-generator/issues/6829

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
@@ -27,6 +27,7 @@ public class CodegenResponse implements IJsonSchemaValidationProperties {
     public boolean is3xx;
     public boolean is4xx;
     public boolean is5xx;
+    public String wildcardFirstChar; // store the first character of 2XX, 5XX, etc
     public String message;
     public boolean hasMore;
     public List<Map<String, Object>> examples;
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
```diff
@@ -27,6 +27,7 @@
     public boolean is3xx;
     public boolean is4xx;
     public boolean is5xx;
+    public String wildcardFirstChar; // store the first character of 2XX, 5XX, etc
```
jimschubert, line 30: @wing328 I don't think this naming is intuitive (Users could assume this is firstChar of something other than status).

Maybe `statusCodeClass` or `statusCodeCategory`? The term "class" is taken directly from [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6), but I'd be concerned this could become confused with the response class.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
```diff
@@ -272,9 +207,15 @@ public void setMultipleOf(Number multipleOf) {
 
     @Override
     public String toString() {
-        final StringBuilder sb = new StringBuilder("CodegenResponse{");
+        final StringBuffer sb = new StringBuffer("CodegenResponse{");
```
jimschubert, line 210: We probably don't need the threading synchronization overhead of `StringBuffer` in a toString method.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3908,6 +3908,12 @@ public CodegenResponse fromResponse(String responseCode, ApiResponse response) {
                     throw new RuntimeException("Invalid response code " + responseCode);
             }
         }
+
+        // for code ending with XX, e.g. 4XX, 5XX
+        if (r.code.length() == 3 && "XX".equals(r.code.substring(1))) {
+            r.wildcardFirstChar = r.code.substring(0, 1);
```
jimschubert, line 3914: It seems like rather than being a settable field, this property on response objects can be a calculated getter. See my comment in that class.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenResponse.java
jimschubert: I think rather than a mutable field, this can be a getter as it should always be the same based on the contents of this type.

Something like:

```
    public String getStatusCodeCategory() {
        if (code != null && code.length() > 0)
            return String.valueOf(code.charAt(0));
        return null;
    }
```

wing328:
@jimschubert thanks for the review. Will incorporate these this week or next.

wing328:
Actually we only need another mustache tag to indicate whether the response code is a range (e.g. 4XX) so I've fled https://github.com/OpenAPITools/openapi-generator/pull/7854 instead.

Issue #7372, opened 2020-09-08
[BUG] python-experimental broken on master (java.lang.NullPointerException)


keymone:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

python-experimental generator fails with `java.lang.NullPointerException`

##### openapi-generator version

master

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info: {title: foo, version: 1.0.0}
components:
  schemas:
    A: {type: object, additionalProperties: {$ref: '#/components/schemas/B'}}
    B: {type: array, items: {$ref: '#/components/schemas/C'}}
    C: {type: object}
paths: {}%
```

##### Generation Details

```
$ docker pull openapitools/openapi-generator-cli:latest
latest: Pulling from openapitools/openapi-generator-cli
Digest: sha256:ddbc3c6d4c9734f8bc80cab402a053cfbc856e036df24dfc6edb417376fc6534
Status: Image is up to date for openapitools/openapi-generator-cli:latest
$ docker run --rm -i --user `id -u`:`id -g` -v `pwd`:/src \
        openapitools/openapi-generator-cli:latest generate \
                -i /src/oapi.yaml \
                -g python-experimental \
                --package-name foo \
                -o /src/temp-openapi-client
[main] INFO  o.o.codegen.DefaultGenerator - Generating with dryRun=false
[main] WARN  o.o.c.ignore.CodegenIgnoreProcessor - Output directory does not exist, or is inaccessible. No file (.openapi-generator-ignore) will be evaluated.
[main] INFO  o.o.codegen.DefaultGenerator - OpenAPI Generator: python-experimental (client)
[main] INFO  o.o.codegen.DefaultGenerator - Generator 'python-experimental' is considered experimental.
[...]

PR #7383, opened 2020-09-09
Fixes isFreeFormObject detection

spacether:
We have a bug in master branch where this schema:
```
    A:
      type: object
      additionalProperties:
        $ref: '#/components/schemas/B'
```
Is incorrectly interpreted as FreeFormObject when it should be a MapSchema
Per our java documentation a schema is FreeForm if:
```
A free form object is an object (i.e. 'type: object' in a OAS document) that:
1) Does not define properties, and
2) Is not a composed schema (no anyOf, oneOf, allOf), and
3) additionalproperties is not defined, or additionalproperties: true, or additionalproperties: {}.
```
Schema A does not satisfy requirement 3. Requirement 3 only allows additionalproperties  values of null, true, and empty object {}. We have {$ref: '#/components/schemas/B'} which is different.

This PR fixes that bug by making sure that when we are in isFreeFormObject we check to see that a ref does not exist in additionalProperties.
- A test has been added demonstrating that the above model is generated with our updated code
- If merged, this PR will fix https://github.com/OpenAPITools/openapi-generator/issues/7372#issuecomment-689776396
 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -781,7 +781,7 @@ public class ModelUtils {
                         }
                     } else if (addlProps instanceof Schema) {
                         // additionalProperties defined as {}
-                        if (addlProps.getType() == null && (addlProps.getProperties() == null || addlProps.getProperties().isEmpty())) {
+                        if (addlProps.getType() == null && addlProps.get$ref() == null && (addlProps.getProperties() == null || addlProps.getProperties().isEmpty())) {
                             return true;
                         }
                     }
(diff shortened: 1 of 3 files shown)


keymone:
if i understand correctly, the error is in `modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java`, so it should be affecting all languages. why did i only get the exception in python-experimental?

keymone:
otherwise the fix looks good, thanks for addressing this so quickly!

spacether:
> if i understand correctly, if the error is in `modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java` then it should be affecting all languages. why did i only get the exception in python-experimental?

So there are two issues at play with why the error is seen in python-experimental and not other generators.
1. Is schema A seen as a FreeFormObject? In master branch yes it is incorrectly classified as FreeFormObject rather than MapSchema. That means that it will not be generated.
2. python-experimental post processes all models to make sure that their imports are formatted correctly for python. Because schema A was not generated, [this line ](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L432)throws a null pointer exception. This only happens in python-experimental because we are post processing all models (whether or not they are generated, and we hit some data where a value that we are trying to grab is null). If other generators were trying to get the "models" value in the schema A model they will also hit this exception. They don't do this same iteration, or they handle the null condition.

That's why it is only seen in python-experimental.

wing328 (approved):
LGTM

Issue #7385, opened 2020-09-10
[BUG] [Typescript-Node] Import mapping seems to be working incorrectly with typescript node


mahirk:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description
Importing mapping for generation on typescript-node does not properly render when generated. 

For mapping such as:
```
  "importMappings": {
    "Pet": "../dataModel/petModel"
  }
```
**Actual**
```
import { AModelPetModel } from '../../dataModel/petModel';
```
**Expected**
```
import { Pet } from '../dataModel/petModel';
```

For a mapping such as:
```
  "importMappings": {
    "Pet": "@company/prefix-zoo-store"
  }
```

**Actual**
```
import { NyPrefixZooStore } from '../@company/prefix-zoo-store';
```
**Expected**
```
import { Pet } from '@company/prefix-zoo-store';
```

##### openapi-generator version
Not a regression, using the latest tag on https://hub.docker.com/r/openapitools/openapi-generator-cli/tags.
This was also tested out with v4.1 which included https://github.com/OpenAPITools/openapi-generator/pull/3469 which allowed the usage of importMapping.

##### OpenAPI declaration file content or url

Using https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml

##### Generation Details

```
docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate \
        -i https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml \
        -c /local/openapi.config.json \
        -o /local/client
```

`openapi.config.json`
```
{
  "generatorName": "typescript-node",
  "additionalProperties": {
    "npmName": "pet-model-node-client",
    "npmRepository" : "http://pet",
    "npmVersion": "1.0.0",
    "supportsES6": true
[...]

PR #7410, opened 2020-09-14
[bug][typescript] Fix node client generator import file paths

mahirk:
### Bug and Recap
Issue: Resolves https://github.com/OpenAPITools/openapi-generator/issues/7385
> For mapping such as:
> ```
>   "importMappings": {
>     "Pet": "../dataModel/petModel"
>   }
> ```
> **Actual**
> ```
> import { AModelPetModel } from '../../dataModel/petModel';
> ```
> **Expected**
> ```
> import { Pet } from '../dataModel/petModel';
> ```
> 
> For a mapping such as:
> ```
>   "importMappings": {
>     "Pet": "@company/prefix-zoo-store"
>   }
> ```
> 
> **Actual**
> ```
> import { NyPrefixZooStore } from '../@company/prefix-zoo-store';
> ```
> **Expected**
> ```
> import { Pet } from '@company/prefix-zoo-store';
> ```

### Method
Below is the map in `operations.get("imports")` based on the import location:

_Model is created by generator_
```
[{import=model/pet, classname=Pet}]
```

_BYOModel_
```
[{import=@company/prefix-zoo-store, classname=Pet}]
```

Therefore the logic was updated to a simpler version, assuming the mapping is created by prior areas.

```
for (Map<String, Object> im : imports) {
     im.put("filename", im.get("import").toString());
}
```

This resulted in the following import for:

_Created Model_
```
import { Pet } from '../model/pet';
```

_BYOModel_
```
import { Pet } from '../@company/prefix-zoo-store';
```

Per the above, this fixed the issue of the improper render which led to the bad `classname` i.e. what led to the following `NyPrefixZooStore`

The next issue was the extra `../` which is embedded as part of the template. See: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-node/api-single.mustache#L8

For this purpose, the extra directory listing was remove from the template and added into the `Codegen` itself, where the `../` is appended by default _ONLY_ when the model is derived from an auto generated file. 

After this change here are the mappings in `operations.get("imports")`

_Model is created by generator_
```
[{import=../model/pet, classname=Pet}]
```

_BYOModel_
```
[{import=@company/prefix-zoo-store, classname=Pet}]
```

This resulted in the following import for:

_Created Model_
```
import { Pet } from '../model/pet';
```

_BYOModel_
```
import { Pet } from '@company/prefix-zoo-store';
```

#### Unit
1. Updated tests for `defaultModelImportTest` 
2. Added tests for `postProcessOperationsWithModels`

#### Manual
1. Ran the above expected v/s actual output comparisons to ensure the values provided are accurate

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -36,7 +36,8 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
     private static final Logger LOGGER = LoggerFactory.getLogger(TypeScriptNodeClientCodegen.class);
 
     public static final String NPM_REPOSITORY = "npmRepository";
-    private static final String DEFAULT_IMPORT_PREFIX = "./";
+    private static final String DEFAULT_MODEL_FILENAME_DIRECTORY_PREFIX = "./";
+    private static final String DEFAULT_MODEL_IMPORT_DIRECTORY_PREFIX = "../";
 
     protected String npmRepository = null;
     protected String apiSuffix = "Api";
@@ -153,7 +154,7 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
             return importMapping.get(name);
         }
 
-        return DEFAULT_IMPORT_PREFIX + camelize(toModelName(name), true);
+        return DEFAULT_MODEL_FILENAME_DIRECTORY_PREFIX + camelize(toModelName(name), true);
     }
 
     @Override
@@ -162,7 +163,7 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
             return importMapping.get(name);
         }
 
-        return modelPackage() + "/" + camelize(toModelName(name), true);
+        return DEFAULT_MODEL_IMPORT_DIRECTORY_PREFIX + modelPackage() + "/" + camelize(toModelName(name), true);
     }
 
     @Override
@@ -222,8 +223,7 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
         // Add additional filename information for model imports in the apis
         List<Map<String, Object>> imports = (List<Map<String, Object>>) operations.get("imports");
         for (Map<String, Object> im : imports) {
-            im.put("filename", im.get("import"));
-            im.put("classname", getModelnameFromModelFilename(im.get("filename").toString()));
+            im.put("filename", im.get("import").toString());
         }
 
         return operations;
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -309,10 +309,6 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
         return toApiFilename(name);
     }
 
-    private String getModelnameFromModelFilename(String filename) {
-        String name = filename.substring((modelPackage() + File.separator).length());
-        return camelize(name);
-    }
 @Override
     protected void addAdditionPropertiesToCodeGenModel(CodegenModel codegenModel, Schema schema) {
         super.addAdditionPropertiesToCodeGenModel(codegenModel, schema);
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptnode/TypeScriptNodeClientCodegenTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptnode/TypeScriptNodeClientCodegenTest.java
@@ -139,7 +139,7 @@ public class TypeScriptNodeClientCodegenTest {
     @Test(description = "correctly produces imports without import mapping")
     public void postProcessOperationsWithModelsTestWithoutImportMapping() {
         final String importName = "../model/pet";
-        Map<String, Object> operations = postProcessOperationsHelper(importName);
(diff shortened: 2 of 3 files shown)


mahirk:
Tagging the most recent contributors to the typescript generator per the contributor list:
@petejohansonxo 
@amakhrov
cc: @wing328

mahirk:
Tagging in a few more TS contributors from [the technical commitee](https://github.com/OpenAPITools/openapi-generator#62---openapi-generator-technical-committee) to help have a look at this:
@Vrolijkx @macjohnny @topce

mahirk:
+ @jimschubert as another active maintainer for a review 
+ @sebastianhaas as a TS Contributor

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -222,8 +223,7 @@ public String toModelImport(String name) {
         // Add additional filename information for model imports in the apis
         List<Map<String, Object>> imports = (List<Map<String, Object>>) operations.get("imports");
         for (Map<String, Object> im : imports) {
-            im.put("filename", im.get("import"));
-            im.put("classname", getModelnameFromModelFilename(im.get("filename").toString()));
```
amakhrov, line 226: looks like now the `getModelnameFromModelFilename` method is no longer used and can be removed?

modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptnode/TypeScriptNodeClientCodegenTest.java
```diff
@@ -134,4 +136,42 @@ public void mappedApiImportTest() {
...
+        final String importName = "@namespace/dir/category";
+        Map<String, Object> operations = postProcessOperationsHelper(importName);
+
+        codegen.postProcessOperationsWithModels(operations, Collections.emptyList());
+        List<Map<String, Object>> extractedImports = (List<Map<String, Object>>) operations.get("imports");
+
+        Assert.assertEquals(extractedImports.get(0).get("filename"), importName);
+    }
+
+    private Map<String, Object> postProcessOperationsHelper(String importName) {
```
amakhrov, line 160: Perhaps just nitpicking - but maybe there could be a better name for this method? Right now it's kind of unclear what its intention is (unless I read the actual method body).

amakhrov (commented):
Thanks for your contribution! The fix looks reasonable to me (just some minor suggestions).

I'm also wondering if this issue (and the fix) are typescript-node specific, or other typescript generators suffer from the same bug.

mahirk:
Thanks for the review @amakhrov! Made the suggestions as recommended above. 
The bug is likely other ts generators. Though it does mean more reviews, once this is approved, I was hoping to break down a pr for the current ts generators (and additionally try and add a sample as well so that it can be well tested)
Here is my testing so far:

| Generator | Issue                                                                                                                                                                                                                              | LoE                                                            |
|-----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
| node      | does not use import mappings correctly                                                                                                                                                                                             | low                                                            |
| axios     | does not support import mappings                                                                                                                                                                                                   | medium                                                         |
| fetch     | does not use import mappings correctly + expects "to" and "from" json methods                                                                                                                                                      | medium to high (will need to think of a better way to do this) |
| angular   | the template assumes everything to be in the mode directory, hence it fails https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L11 | low                                                            |


I'll potentially look at a quick fix for angular next and then attempt to evaluate axios and fetch since those are popular libraries. 

Another issue for typescript node is: https://github.com/OpenAPITools/openapi-generator/issues/7440

macjohnny (approved):
LGTM

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -162,7 +163,7 @@ public String toModelImport(String name) {
             return importMapping.get(name);
         }
 
-        return modelPackage() + "/" + camelize(toModelName(name), true);
+        return DEFAULT_MODEL_IMPORT_DIRECTORY_PREFIX + modelPackage() + "/" + camelize(toModelName(name), true);
```
jimschubert, line 166: @mahirk not a blocker on this pr, but wanted to comment since you said you were planning to look at ts implementation in other generators...

since modelPackage is passed by the user, this could result in `models//` when passed with trailing slash. It might be best to use StringUtils, something like:

```
return DEFAULT_MODEL_IMPORT_DIRECTORY_PREFIX + 
    StringUtils.appendIfMissing(modelPackage(), "/") +
    camelize(toModelName(name), true);
```

jimschubert (approved):
looks good. Just left a small comment if you're looking to do the same in other generators.

tomh4:
was this ever 

> Thanks for the review @amakhrov! Made the suggestions as recommended above.
> The bug is likely other ts generators. Though it does mean more reviews, once this is approved, I was hoping to break down a pr for the current ts generators (and additionally try and add a sample as well so that it can be well tested)
> Here is my testing so far:
> 
> Generator	Issue	LoE
> node	does not use import mappings correctly	low
> axios	does not support import mappings	medium
> fetch	does not use import mappings correctly + expects "to" and "from" json methods	medium to high (will need to think of a better way to do this)
> angular	the template assumes everything to be in the mode directory, hence it fails https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L11	low
> I'll potentially look at a quick fix for angular next and then attempt to evaluate axios and fetch since those are popular libraries.
> 
> Another issue for typescript node is: #7440

Did you ever figure out how to get this working with typescript-fetch ? The importMappings still seem to be ignored completely

PR #7416, opened 2020-09-14
Enable support for V3 annontations + use of springdoc

alygueye:
Update mustache template and codeGen Mojo to support spec v3 and springdoc


alygueye:
mistake

PR #7424, opened 2020-09-15
[Python-experimental] Don't mandate passing in required value as argument

bkabrda:
After a change done in https://github.com/OpenAPITools/openapi-generator/pull/7274, an enum model with a default value would not have the default value used in constructor. This PR fixes that.

### PR checklist
 

jirikuncar (approved):

spacether:
Thank you for the PR. Per the openapi spec, default is used to describe a value that the server will use if the client omits sending a value. Because of that I don't think that setting a default value with the default parameter is something that we should do. If there are two enum values and a default is set the spec is saying there are two possible options and if you don't send me this property or model I will use the default value. 

If there is only one enum value, then we are sure what value should be used. In that case we do use that value when making a new instance of that enum.

Does that make sense?

bkabrda:
Well to be fair I don't understand why the default couldn't be used even when instantiating the class in the client code. The openapi spec doesn't prohibit that, and it also doesn't say that the client should send no value when there is a default.

This is just a quality-of-life improvement that makes using the generated enum classes more convenient.

spacether:
> Well to be fair I don't understand why the default couldn't be used even when instantiating the class in the client code. The openapi spec doesn't prohibit that, and it also doesn't say that the client should send no value when there is a default.
> 
> This is just a quality-of-life improvement that makes using the generated enum classes more convenient.

The default could be used when instantiating the class in client code as you say. There are a couple of implications to doing that.
Right now python-experimental using the same logic for:
- endpoint params
- enum models
- object model properties
Enums of length one are set as the client side value when defined in the spec, otherwise the user must provide the value.
My preference is to keep our logic the same on all of these interfaces.

Do you suggest setting the default on all of these 3 use cases?
- on endpoint parameters and model properties, they are set in the endpoint function or the model init function.
The end result of applying the same logic to those is that those parameters with defaults will always be sent.
My preference is to not do that because it makes it so the client cannot choose not to send those parameters, they will always be included. What if we want to test the use case where we omit sending a model property or endpoint parameter?

bkabrda:
@spacether so I'll admit I'm not that deeply familiar with the python client, but here's my take on this:

First of all, I totally agree with keeping things consistent, so I do agree that all the three cases that you described should behave the same way.

What confuses me in your comment is this quote:

> The end result of applying the same logic to those is that those parameters with defaults will always be sent to the server if we depend only on the presence of defaultValue and set the value in a function signature.

Why is that? Can't you just pass `None` explicitly to these functions when you don't want this value sent? How do you *not* send them right now?

bkabrda:
I've been thinking some more and here's a slightly different take:

Maybe we can remove the `value` argument from the constructor altogether (which is the state that was before, I think) and just use the following already existing code snippet for this case, effectively getting:

```
def __init__(self, *args, **kwargs):
    if 'value' in kwargs:
        value = kwargs.pop('value')
    elif args:
        args = list(args)
        value = args.pop(0)
    else:
        value = {{{defaultValue}}}
    # raise if args is non-empty or kwargs is non-empty
```

Would that make more sense to you?


Taking a step back, what I want from an API library is making things as easy and comfortable as possible to use. While the behavior that we're discussing here is not defined by the OpenAPI spec, it is certainly more comfortable for use than the current state. WDYT?

spacether:
> I've been thinking some more and here's a slightly different take:
> 
> Maybe we can remove the `value` argument from the constructor altogether (which is the state that was before, I think) and just use the following already existing code snippet for this case, effectively getting:
> 
> ```
> def __init__(self, *args, **kwargs):
>     if 'value' in kwargs:
>         value = kwargs.pop('value')
>     elif args:
>         args = list(args)
>         value = args.pop(0)
>     else:
>         value = {{{defaultValue}}}
>     # raise if args is non-empty or kwargs is non-empty
> ```
> 
> Would that make more sense to you?
> 
> Taking a step back, what I want from an API library is making things as easy and comfortable as possible to use. While the behavior that we're discussing here is not defined by the OpenAPI spec, it is certainly more comfortable for use than the current state. WDYT?

Your suggestion looks good to me. You would also need mustache protection of whether or not the default block is generated like:
```
def __init__(self, *args, **kwargs):
    if 'value' in kwargs:
        value = kwargs.pop('value')
    elif args:
        args = list(args)
        value = args.pop(0)
{{#defaultValue}}
    else:
        value = {{{defaultValue}}}
{{/defaultValue}}
    # raise if args is non-empty or kwargs is non-empty
```

spacether:
>The end result of applying the same logic to those is that those parameters with defaults will always be sent to the server if we depend only on the presence of defaultValue and set the value in a function signature.
>
>Why is that? Can't you just pass None explicitly to these functions when you don't want this value sent? How do you not send them right now?

This was assuming that when defaultValue was set, we would assign it in the method or function signature.
Both required and optional parameters can have defaultValues set.
So if we are loading values into parameters based on whether there is a default value, regardless of whether the param is optional or required, a value would be assigned, and the payload would include it.

How it works today:
- required params are positional and must be included by the. dev
- optional params are kwarg params
- required with enums of length one are named arguments with a default [value see here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/python-experimental/api.mustache#L46)
The end result is that parameters are only included in payloads of they are passed by the user or are a named argument with a default. This means that unset variables do not exist in _data_store, and variables can have None explicitly assigned to them as a valid value.

bkabrda:
Ok, thanks for your reply. I'll work on my last suggestion that looked ok to you, hopefully will get it done during today or Monday.

bkabrda:
Ok, implemented. Hopefully this looks better to you @spacether. Thanks!

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
```diff
@@ -8,36 +8,47 @@
...
         elif args:
             args = list(args)
             value = args.pop(0)
+            have_value = True
+{{#defaultValue}}
         else:
             value = {{{defaultValue}}}
-{{/hasRequired}}
+            have_value = True
+{{/defaultValue}}
```
spacether, line 36: I think that you can do this without have_value. How about this?
do not define value, do not define have_value
```
        if 'value' in kwargs:
            value = kwargs.pop('value')
        elif args:
            args = list(args)
            value = args.pop(0)
{{#defaultValue}}
        else:
            value = {{{defaultValue}}}
{{/defaultValue}}
{{^defaultValue}}
        else:
            raise ApiTypeError(
                "value is required, but not passed in args or kwargs and doesn't have default",
                path_to_item=_path_to_item,
                valid_classes=(self.__class__,),
            )
{{/defaultValue}}
```
My understanding is that all ModelSimple instances need an assigned value.

spacether (changes requested):
Overall this looks good.
- Can you remove the unneeded have_value variable per my comment?
- Can you verify that required default values are used in endpoint parameters? If we don't have a test of this please add a test.
- Can you verify that required default values are used in model properties? If we don't have a test of this please add a test.
- Please update hasRequired to False here https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L741
- Please update the java doc here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L730
To:
```
        // 2. server has a default
        //      schema contains default
        //      cm.defaultValue set, cm.hasRequired = true
        //      different value here to differentiate between use case 3 below
        //      If the param with the defaultValue is required +
        //          1. an enum is instantiated without input, the defaultValue will be used as its value
        //          2 .an endpoint is called without the required defaultValue param input, the defaultValue will be used as the param value
        //          3. an model is instantiated without the required defaultValue param input, the defaultValue will be used as the param value
```
@bkabrda I am concerned that applying this logic in items 2.1 + 2.3 and will mutate payloads being received from the server. Maybe we should only do this logic if _spec_property_naming == False which means the data did not come from the server and we are making a model in client code. What do you think?

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
bkabrda: Yeah, that's a good point, I'll do it.

bkabrda:
So I removed usage of `have_value` as you requested. As for the other points:

* I unfortunately don't have enough time on my hands to work on the endpoint parameters and model properties. I don't believe that the level of inconsistency with approach in this PR would be so bad that we couldn't afford merging it at all if the other two don't get updated.

> Please update hasRequired to False here https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L741

That just points to a docstring of some method, so I'm not sure which one you want me to update. Can you use a specific commit in that link, so that the line numbers don't shift when someone merges PRs for this file?

> Please update the java doc here: ...

This keeps me wondering. I think you've referred several times to the fact that "server has a default". Where in the OpenAPI Spec is this coming from? The [definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#schema-object) of schema's `default` that I see in the OpenAPI Spec reads:

```
default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if type is string, then default can be "foo" but cannot be 1.
```

This doesn't at all say "server has a default", it talks in general about consumer, which can be server (if client is producer), but it can also be client (if server is producer). Is there something that I'm missing in the spec?

spacether:
> hasRequired

Doing only this change is fine here.

Sorry about pointing to the wrong place.
[This is the hasRequired](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L731) that I am asking that we update.

When reading that spec I assumed that the consumer was the server. Your interpretation of the client as another consumer is valid too. This question is a I have asked for clarification on [here](https://github.com/OAI/OpenAPI-Specification/issues/1789).

Based on that updated understanding can you update the java doc `server has default` to be more correct? Maybe something like:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientExperimentalCodegen.java#L716
```
        // 2. spec has a default
        //      schema contains default
        //      cm.defaultValue set, cm.hasRequired = false
        //      different value here to differentiate between use case 3 below
        //      This defaultValue is used when a consumer (client or server) lacks the input argument, defaultValue will be used
```

bkabrda:
@spacether I believe I've addressed all your comments now. There is a CI failure, but I don't believe it's related to anything that my PR did, probably an intermittent issue.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
```diff
@@ -8,30 +8,39 @@
...
-{{/hasRequired}}
         Keyword Args:
-{{^hasRequired}}
             value ({{{dataType}}}):{{#description}} {{description}}.{{/description}}{{#defaultValue}} defaults to {{{defaultValue}}}{{/defaultValue}}{{#allowableValues}}, must be one of [{{#enumVars}}{{{value}}}, {{/enumVars}}]{{/allowableValues}}  # noqa: E501
-{{/hasRequired}}
 {{> python-experimental/model_templates/docstring_init_required_kwargs }}
         """
 
-{{^hasRequired}}
+        value = None
```
spacether, line 23: Please delete this line setting value as None.
None may be a valid value for an instance, like nullable models.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
```diff
@@ -8,30 +8,39 @@
...
             value = kwargs.pop('value')
         elif args:
             args = list(args)
             value = args.pop(0)
+{{#defaultValue}}
         else:
             value = {{{defaultValue}}}
+{{/defaultValue}}
+{{^defaultValue}}
+{{#hasRequired}}
```
spacether, line 34: Please remove this hasRequired block.
In this case there is no default value. There must always be an input value for modelSimple models. We don't need to filter by hasRequired any more.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
```diff
@@ -8,30 +8,39 @@
...
     @convert_js_args_to_python_args
-    def __init__(self{{#hasRequired}}, value{{/hasRequired}}, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         """{{classname}} - a model defined in OpenAPI
 
-{{#hasRequired}}
+        Note that value can be passed either in args or in kwargs, but not in both.
+
         Args:
             value ({{{dataType}}}):{{#description}} {{description}}.{{/description}}{{#defaultValue}} if omitted the server will use the default value of {{{defaultValue}}}{{/defaultValue}}{{#allowableValues}}, must be one of [{{#enumVars}}{{{value}}}, {{/enumVars}}]{{/allowableValues}}  # noqa: E501
```
spacether, line 17: Can you remove value from args here?
We no longer have it in args, it is in kwargs in line 19.

spacether:
Thank you for making those changes. Just a few small tweaks are needed:
- https://github.com/OpenAPITools/openapi-generator/pull/7424#discussion_r494720863
- https://github.com/OpenAPITools/openapi-generator/pull/7424#discussion_r494721455
- https://github.com/OpenAPITools/openapi-generator/pull/7424#discussion_r494721993

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
spacether, line 17: How about changing FROM `if omitted the server will use the default value of {{{defaultValue}}}` TO `defaults to {{{defaultValue}}}`

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/method_init_simple.mustache
spacether: Thank you!

spacether (approved):
Looks good! Thank you for the PR and the updates.
I made some small tweaks to fix the none issue etc.

bkabrda:
Thanks! I got swamped up with different work and just wanted to address your last batch of comments. I appreciate that you took the time to fix and merge!

PR #7451, opened 2020-09-18
[php-slim4] Add package name and version to composer.json

### LINKED ISSUE #5054: [REQ][Slim4] Add package name and version to composer.json

I'm using php-slim4 to generate server code for our API. I want to integrate the generated package in our project with composer using a [custom repository of type "path"](https://getcomposer.org/doc/05-repositories.md#path).
Unfortunately the composer.json generated by php-slim4 is missing the required keys `name` and `version`.

`name` and `version` should be added to `modules/openapi-generator/src/main/resources/php-slim4-server/composer.mustache` like is is in `modules/openapi-generator/src/main/resources/php/composer.mustache`. That would solve the problem.
```
{
    "name": "{{gitUserId}}/{{gitRepoId}}",
    {{#artifactVersion}}
    "version": "{{artifactVersion}}",
    {{/artifactVersion}}

   ...
}
```

At the moment I use a custom `composer.mustache` when generating the server code. It is just a copy of the repository version with the above lines added. Unfortunately I'll will have to update the file manually every time php-slim4 changes dependencies.

## Additional context

The same problem applies when using php-silex and maybe other php-* generators. It would be nice to have more uniform `composer.json` files among the those generators.

TechPrivacyLab:
Resolves #5054

### PR checklist
 
cc @jebentier, @dkarlovi, @mandrean, @jfastnacht, @ackintosh, @ybelenko, @renepardon

### DIFF
--- a/samples/server/petstore/php-slim4/composer.json
+++ b/samples/server/petstore/php-slim4/composer.json
@@ -1,4 +1,6 @@
 {
+  "name": "GIT_USER_ID/GIT_REPO_ID",
+  "version": "1.0.0",
   "minimum-stability": "RC",
   "repositories": [
     {
--- a/modules/openapi-generator/src/main/resources/php-slim4-server/composer.mustache
+++ b/modules/openapi-generator/src/main/resources/php-slim4-server/composer.mustache
@@ -1,4 +1,8 @@
 {
+  "name": "{{gitUserId}}/{{gitRepoId}}",
+  {{#artifactVersion}}
+  "version": "{{artifactVersion}}",
+  {{/artifactVersion}}
   "minimum-stability": "RC",
   "repositories": [
     {


ybelenko:
I thought that we were going to remove these composer props from all php server generators at some point of time.

`Packagist` recommends not to use `version` tag in composer config:
> ### version
> The version of the package. In most cases this is not required and should be omitted (see below).
> Optional if the package repository can infer the version from somewhere, such as the VCS tag name in the VCS repository. In that case it is also recommended to omit it.
>
> **Note:** Packagist uses VCS repositories, so the statement above is very much true for Packagist as well. Specifying the version yourself will most likely end up creating problems at some point due to human error.

Ref: [The composer.json schema - Composer getcomposer.org](https://getcomposer.org/doc/04-schema.md#version)
I never used Slim generator with vanilla template, because it's always outdated to me :smile: I wanted to remove `name` and `version` tag because when `name` set to `GIT_USER_ID/GIT_REPO_ID` composer throws warning about invalid package name. Package name will be unlikely used because most of the users don't submit server implementation to packagist.

Conclusion, maybe we should add mustache flag `{{containsPackageName}}` in `AbstractPhpGenerator.java` and write `name` to `composer.json` only when it's real valid package name, not `GIT_USER_ID/GIT_REPO_ID`.
And `version` tag doesn't bother me, I can rewrite template like I usually do with `composer.json`.

ybelenko:
@wing328 you suggested to submit PR in #5054, so merge it if you still find it useful.
I described my point above. I'm not against neither for that update.

samples/server/petstore/php-slim4/composer.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "GIT_USER_ID/GIT_REPO_ID",
+  "version": "1.0.0",
```
dkarlovi, line 3: Generally, the version should NOT be added to the composer.json, it's deduced by Packagist from Git tags (namely, tags become the version).

PR #7522, opened 2020-09-28
Updated keywords to be aligned with website https://developer.salesfo…

henryh-force:
Reserved words were not aligned with the recent salesforce apex published list.

https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_reserved_words.htm


wing328:
@henryh-force thanks for the PR

cc @muenzpraeger

jimschubert:
I've updated the samples with the new changes.

PR #7528, opened 2020-09-28
[PHP] Small tweaks for php generator, PHPStan level 1

dkarlovi:
Continuation on #7376, fixes two small code issues and a single code style issue.

# PR checklist
 
@jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) heart, @ybelenko (2018/07), @renepardon (2018/12)

### DIFF
--- a/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
@@ -354,8 +354,8 @@ class ObjectSerializer
                     continue;
                 }
 
-                $propertyValue = $data->{$instance::attributeMap()[$property]};
-                if (isset($propertyValue)) {
+                if (isset($data->{$instance::attributeMap()[$property]})) {
+                    $propertyValue = $data->{$instance::attributeMap()[$property]};
                     $instance->$propertySetter(self::deserialize($propertyValue, $type, null));
                 }
             }
--- a/modules/openapi-generator/src/main/resources/php/api.mustache
+++ b/modules/openapi-generator/src/main/resources/php/api.mustache
@@ -63,18 +63,18 @@ use {{invokerPackage}}\ObjectSerializer;
      * @param ClientInterface $client
      * @param Configuration   $config
      * @param HeaderSelector  $selector
-     * @param int             $host_index (Optional) host index to select the list of hosts if defined in the OpenAPI spec
+     * @param int             $hostIndex (Optional) host index to select the list of hosts if defined in the OpenAPI spec
      */
     public function __construct(
         ClientInterface $client = null,
         Configuration $config = null,
         HeaderSelector $selector = null,
-        $host_index = 0
+        $hostIndex = 0
     ) {
         $this->client = $client ?: new Client();
         $this->config = $config ?: new Configuration();
         $this->headerSelector = $selector ?: new HeaderSelector();
-        $this->hostIndex = $host_index;
+        $this->hostIndex = $hostIndex;
     }
 
     /**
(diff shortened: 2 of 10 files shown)


modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -354,8 +354,8 @@ class ObjectSerializer
                     continue;
                 }
 
-                $propertyValue = $data->{$instance::attributeMap()[$property]};
-                if (isset($propertyValue)) {
+                if (isset($data->{$instance::attributeMap()[$property]})) {
+                    $propertyValue = $data->{$instance::attributeMap()[$property]};
```
dkarlovi, line 358: The previous statement basically made no sense, it's just a null check in disguise.

modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -63,18 +63,18 @@ use {{invokerPackage}}\ObjectSerializer;
      * @param ClientInterface $client
      * @param Configuration   $config
      * @param HeaderSelector  $selector
-     * @param int             $host_index (Optional) host index to select the list of hosts if defined in the OpenAPI spec
+     * @param int             $hostIndex (Optional) host index to select the list of hosts if defined in the OpenAPI spec
```
dkarlovi, line 66: Code style fix.

modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -567,14 +560,16 @@ use {{invokerPackage}}\ObjectSerializer;
         }
 
         // for model (json/xml)
-        if (isset($_tempBody)) {
-            // $_tempBody is the method argument, if present
+        {{#bodyParams}}
```
dkarlovi, line 563: This one produced nonsensical code unless you have body params:
1. sets `$_tempBody` to null
2. checks if `$_tempBody` is not null

It can never be not null because the code mutating it is not generated, why are the checks for it generated then?

dkarlovi:
@ybelenko @wing328 WDYT? I have another one in the pipeline once this one is accepted.

ybelenko:
I'm going to check this pr locally today.

modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -567,14 +560,16 @@ use {{invokerPackage}}\ObjectSerializer;
...
+        if (isset(${{paramName}})) {
             if ($headers['Content-Type'] === 'application/json') {
-                $httpBody = \GuzzleHttp\json_encode(ObjectSerializer::sanitizeForSerialization($_tempBody));
+                $httpBody = \GuzzleHttp\json_encode(ObjectSerializer::sanitizeForSerialization(${{paramName}}));
             } else {
-                $httpBody = $_tempBody;
+                $httpBody = ${{paramName}};
             }
-        } elseif (count($formParams) > 0) {
+        } else
```
ybelenko, line 570: This abandoned `else` really bothers me... Did you forgot to delete it? Or maybe you wanted it as `elseif` as it was before?

dkarlovi: The issue is the `if` is only needed when you have the body, otherwise I delete it and the `elseif` becomes `if`. I don't know how to do it better with Mustache. 🙈

ybelenko: Seems like I fixed it in my commit.

ybelenko (approved):
I don't quite understand the part with the body serialization, looks too complicated to understand it without tests, but I hope you know what you're doing. 👍

dkarlovi:
@ybelenko 

Basically, the code does this:
1. create a variable `$_tempBody = null`
2. if some condition in the generator, set the variable to some value
3. you have a if which checks `if ($_tempBody !== null)`

But, in case the second condition is not set, you generate code like this:

```php
$_tempBody = null;
// additional code here is not generated
if ($_tempBody !== null) {}
```

Which is nonsensical, the if will never match since the code mutating `$_tempBody` is not generated.

What I've done is changed the condition from 2. to generate the `if()` if the condition is set, but don't generate it if not. That way, the generated code makes sense in both cases and should work exactly as before, but doesn't have "generation sickness" (the code being clunky as a side-effect of it being generated).

PR #2266, opened 2019-02-28
[typescript-node] fixed enum generator for top level enums #665

christof-vollrath-spring-media:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh`, `./bin/openapi3/{LANG}-petstore.sh`, `./bin/security/{LANG}-petstore.sh` and `./bin/openapi3/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`~~, `3.4.x`, `4.0.x`~~. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

PR #7529, opened 2020-09-28
[typescript] Fix generation of enum models

### LINKED ISSUE #665: [typescript] support for enum
##### Description

Enum are not correctly generated in typescript client. Some investigation are requested to understand what is going wrong.

See also: https://www.typescriptlang.org/docs/handbook/enums.html

CC: @kayhuba

3.1.2

CLI to use the `typescript-node` generator.

bodograumann:
Due to the fact that the consolidated typescript template was forked from master so long ago, there are still some bugs left that where fixed for other typescript generators in the meantime. This is one of them.

This fixes #665 for the consolidated typescript generator. (With [sample definition](https://github.com/OpenAPITools/openapi-generator/issues/665#issuecomment-408769117))
The issue there is, that top-level enum schemas, which are not nested as attributes of other schemas are just generated as normal object schemas, which is not suited to represent the different allowed values.

Original fix for typescript-node was in PR #2266, merged as 8417c5bed0e23764dc841816c2322cf7dc4e9b0d in version 4.1.0.

CC: @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache
@@ -31,10 +31,13 @@ const supportedMediaTypes: { [mediaType: string]: number } = {
 let enumsMap: Set<string> = new Set<string>([
     {{#models}}
         {{#model}}
+            {{#isEnum}}
+    "{{classname}}{{enumName}}",
+            {{/isEnum}}
             {{#hasEnums}}
                 {{#vars}}
                     {{#isEnum}}
-					"{{classname}}{{enumName}}",
+    "{{classname}}{{enumName}}",
                     {{/isEnum}}
                 {{/vars}}
             {{/hasEnums}}
@@ -45,7 +48,9 @@ let enumsMap: Set<string> = new Set<string>([
 let typeMap: {[index: string]: any} = {
     {{#models}}
         {{#model}}
+            {{^isEnum}}
     "{{classname}}": {{classname}},
+            {{/isEnum}}
         {{/model}}
     {{/models}}
 }
--- a/modules/openapi-generator/src/main/resources/typescript/model/model.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript/model/model.mustache
@@ -11,6 +11,7 @@ import { HttpFile } from '../http/http{{extensionForDeno}}';
 * {{{description}}}
 */
 {{/description}}
+{{^isEnum}}
 export class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
 {{#vars}}
 {{#description}}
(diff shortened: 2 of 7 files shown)


macjohnny (approved):
LGTM

duhseekoh:
Is there anyway I can pick this change up prior to a proper 5.0 release? im using the docker cli approach with `typescript-axios`. I tried pulling in the `openapi-generator-cli:latest` instead of v4.3.1 and it still doesn't generate types for top level enums. Any ideas?

macjohnny:
@duhseekoh the `openapi-generator-cli:latest` should contain the changes. please check the output of the version information to make sure you are using the right version.

duhseekoh:
I confirmed in the output i'm using 5.0.0-SNAPSHOT.

An enum outlined in the swagger.json like this, isn't having a type generated for it when using the 5.0.0-SNAPSHOT + typescript-axios.

```
"components": {
  "schemas": {
    "Color": {
        "enum": [
          "BLUE",
          "YELLOW",
          "GREEN"
        ],
        "type": "string",
        "additionalProperties": false
  }
}
```

macjohnny:
The fix in this PR is for the new typescript generator.
Would you like to fix this for the (old) typescript-axios generator? Please open an issue with a minimal reproduction example as posted above

duhseekoh:
Oh, is the direction to use the new typescript generator? Is typescript-axios deprecated?

macjohnny:
@duhseekoh the `typescript-axios` generator is not deprecated, but the long-term goal would be to have all flavors of typescript generators united in the currently experimental `typescript` generator. This can happen if the community helps to port the current typescript generators.

duhseekoh:
It turns out typescript-axios does in fact generate TS enums for top level spec enums.

What I discovered, after pulling the source for openapi-generator down and through some trial and error, is that the generator fails when it hits an enum with `additionalProperties` attribute on it. In open api generator local dev mode, it exits the process when it encounters this. When it's ran via the packaged cli, it silently ignores those enums, and thus they are never included in the script. This is on 4.3.1 and the latest 5.x snapshot.

I can create a bug if needed. But first I was curious if it was expected behavior that it should fail when including that extra attribute?

Fails
```
"components": {
  "schemas": {
    "Color": {
        "enum": [
          "BLUE",
          "YELLOW",
          "GREEN"
        ],
        "type": "string",
        "additionalProperties": false
    }
  }
}
```

Works (generates Color enum)
```
"components": {
  "schemas": {
    "Color": {
        "enum": [
          "BLUE",
          "YELLOW",
          "GREEN"
        ],
        "type": "string"
    }
  }
}
```

PR #7546, opened 2020-09-29
Adds primitive base classes

spacether:
This is the first step to using dynamic base classes in the python-experimental client
In this PR I:
- store enum models in an Enum class
- other non object models inherit from the expected base class (str, int, etc) unless they are nullable
- nullable non-object models will return a dynamic base class of Enum type where the name is NONE and the value is None
  - the same tactic will be used for True and False booleans
  - this is done because None. True, and False cannot be inherited from BUT we can make enums that contain them
  - our enum class can will contain a base class like NullableString which allows us to check for this type when sending None/True/False values to the server

Long Term Benefits of Using Dynamic Base Classes
- This supports complicated use cases like allOf -OneOfComposedSchema. OneOfComposedSchema oneOf: -schemaA -schemaB
- We can reduce the number of times that validations and type conversion are done

### PR checklist
 

spacether:
I am still working on this but I have decided to make a general Schema class to do it rather than ModelSimple
That implementation is being worked in https://github.com/spacether/openapi-generator/tree/adds_schema_class

Issue #7557, opened 2020-09-30
[BUG] [Typescript] allow body on DELETE


dartzki:
##### Description

The typescript api currently does not support passing a body to a delete request.
DELETE Requests may have a body as per specification.
Generated services for a DELETE Request with a body simply drop the 'body' argument, leading to a wrong request.

##### openapi-generator version

5.0.0-beta2

##### Suggest a fix

the templates should be adapted to use `httpClient.request` instead of `httpClient.{{httpMethod}}` and pass the `{{httpMethod}}` as first, and the body as third argument.

Also, looking at the repo it seems that the openapi-generator suggests that DELETE request do not allow for a body by default:
https://github.com/OpenAPITools/openapi-generator/blob/ff68128c15363ece22d7b6376952eb4eabc1f0bd/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenOperation.java#L203

Imo this is very misleading.

FelixCoutinho:
OpenAPI Spec is going to introduce body as optional in DELETE operations in their version 3.1.0

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md

PR #7558, opened 2020-09-30
Fix issue7557 - allow body for DELETE requests

dartzki:
allow to have a body on a delete request.
DELETE requests support having a body per specification, therefore the openapi stubs should support this as well.

Instead of singling out the DELETE method, this PR uses the more generic `httpClient.request` that allows passing a 
body for any method.
Therefore it leaves the handling of a defined body on an unsupported method to the underlying implementation.
This is also the proposed solution by an angular developer, as the angular team is not willing to change the api of
`httpClient.delete` to support a body argument:
https://github.com/angular/angular/pull/37894

fixes issue: https://github.com/OpenAPITools/openapi-generator/issues/7557

### PR checklist
 
technical committee:

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -334,9 +334,10 @@ export class {{classname}} {
         }
 
     {{/isResponseFile}}
-        return this.httpClient.{{httpMethod}}{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}(`${this.configuration.basePath}{{{path}}}`,{{#isBodyAllowed}}
-            {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},{{/isBodyAllowed}}
+        return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}("{{httpMethod}}",
+            `${this.configuration.basePath}{{{path}}}`,
             {
+                body: {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},
     {{#hasQueryParams}}
                 params: queryParameters,
     {{/hasQueryParams}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -337,7 +337,14 @@ export class {{classname}} {
         return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}("{{httpMethod}}",
             `${this.configuration.basePath}{{{path}}}`,
             {
-                body: {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},
+    {{#hasBodyParam}}
+                body: {{#bodyParam}}{{paramName}}{{/bodyParam}},
+    {{/hasBodyParam}}
+    {{^hasBodyParam}}
+      {{#hasFormParams}}
+                body: convertFormParamsToString ? formParams.toString() : formParams,
+      {{/hasFormParams}}
+    {{/hasBodyParam}}
     {{#hasQueryParams}}
                 params: queryParameters,
     {{/hasQueryParams}}
--- a/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
@@ -334,7 +334,7 @@ export class {{classname}} {
         }
 
     {{/isResponseFile}}
-        return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}("{{httpMethod}}",
+        return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}('{{httpMethod}}',
             `${this.configuration.basePath}{{{path}}}`,
             {
     {{#hasBodyParam}}
--- a/samples/client/petstore/typescript-angular-v10-provided-in-root/builds/default/api/pet.service.ts
+++ b/samples/client/petstore/typescript-angular-v10-provided-in-root/builds/default/api/pet.service.ts
@@ -148,7 +148,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<any>("post",
+        return this.httpClient.request<any>('post',
             `${this.configuration.basePath}/pet`,
             {
                 body: body,
@@ -205,7 +205,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<any>("delete",
+        return this.httpClient.request<any>('delete',
             `${this.configuration.basePath}/pet/${encodeURIComponent(String(petId))}`,
             {
                 responseType: <any>responseType,
@@ -266,7 +266,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<Array<Pet>>("get",
+        return this.httpClient.request<Array<Pet>>('get',
             `${this.configuration.basePath}/pet/findByStatus`,
             {
                 params: queryParameters,
@@ -329,7 +329,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<Array<Pet>>("get",
+        return this.httpClient.request<Array<Pet>>('get',
             `${this.configuration.basePath}/pet/findByTags`,
             {
                 params: queryParameters,
@@ -385,7 +385,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<Pet>("get",
+        return this.httpClient.request<Pet>('get',
             `${this.configuration.basePath}/pet/${encodeURIComponent(String(petId))}`,
             {
                 responseType: <any>responseType,
@@ -447,7 +447,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<any>("put",
+        return this.httpClient.request<any>('put',
             `${this.configuration.basePath}/pet`,
             {
                 body: body,
@@ -524,7 +524,7 @@ export class PetService {
             responseType = 'text';
         }
 
-        return this.httpClient.request<any>("post",
+        return this.httpClient.request<any>('post',
             `${this.configuration.basePath}/pet/${encodeURIComponent(String(petId))}`,
             {
                 body: convertFormParamsToString ? formParams.toString() : formParams,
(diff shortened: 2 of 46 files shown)


modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -334,9 +334,10 @@ export class {{classname}} {
         }
 
     {{/isResponseFile}}
-        return this.httpClient.{{httpMethod}}{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}(`${this.configuration.basePath}{{{path}}}`,{{#isBodyAllowed}}
-            {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},{{/isBodyAllowed}}
+        return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}("{{httpMethod}}",
+            `${this.configuration.basePath}{{{path}}}`,
             {
+                body: {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},
```
amakhrov, line 340: maybe make this block conditional? only render it if bodyParam is present, otherwise just skip body (right now it emits `body: null` in this case). this would make generated code for GET requests a bit cleaner.

not a big deal - it's good enough as is

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -334,9 +334,10 @@ export class {{classname}} {
         }
 
     {{/isResponseFile}}
-        return this.httpClient.{{httpMethod}}{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}(`${this.configuration.basePath}{{{path}}}`,{{#isBodyAllowed}}
-            {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}{{#hasFormParams}}convertFormParamsToString ? formParams.toString() : formParams{{/hasFormParams}}{{^hasFormParams}}null{{/hasFormParams}}{{/bodyParam}},{{/isBodyAllowed}}
+        return this.httpClient.request{{^isResponseFile}}<{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}>{{/isResponseFile}}("{{httpMethod}}",
```
amakhrov, line 337: suggest using single quotes to wrap http method, to be consistent with the rest of the generated code

amakhrov:
It's funny that originally this `isBodyAllowed` method was introduced specifically to support typescript-angular generator.
However, since then it's been used by a few more generators, so we cannot safely get rid of it.

topce:
@dartzki  nice work however little bit dangerous  possible breaking change !
You are basically  changing code to call
lower level method request
instead of higher level  methods (get,put,post ..).

In my opinion it would be more robust to just to handle your problematic case delete method with body 
to use httpClient request.
That way you are sure that you did not break other people code .

You see how Angular team members are hesitant to do right stuff: to add body  to delete ,
and with your PR we would basically change  all typescript angular clients 
So my suggestion is to change code just to handle specific case that you point out.
Moreover you see in generated code lot of differences but in pet.yaml there is no example of 
DELETE call with body parameter.
Maybe to add this case delete with body to yaml file so when you generate code you would see code that 
is generated with your improvement.

@amakhrov  indeed isBodyAllowed is implemented like this 
```
/**
     * Check if body param is allowed for the request method
     *
     * @return true request method is PUT, PATCH or POST; false otherwise
     */
    public boolean isBodyAllowed() {
        return Arrays.asList("PUT", "PATCH", "POST").contains(httpMethod.toUpperCase(Locale.ROOT));
    }
```
so it is not just DELETE it is GET as well 
As @dartzki  point out by spec body is allowed in DELETE
so maybe method should be changed`
```
    public boolean isBodyAllowed() {
        return Arrays.asList("DELETE""PUT", "PATCH", "POST").contains(httpMethod.toUpperCase(Locale.ROOT));
        // or just use 
        return  !isRestfulShow(); 
   
    } 
```

also erlang and java could be impacted with that changes 
maybe @wing328  can also take a look on this global change

dartzki:
> @dartzki nice work however little bit dangerous possible breaking change !
> You are basically changing code to call
> lower level method request
> instead of higher level methods (get,put,post ..).
> 
> In my opinion it would be more robust to just handle your problematic case delete method with body
> to use httpClient request.
> That way you are sure that you did not break other people code .
> 
> You see how Angular team members are hesitant to do right stuff: to add body to delete ,
> and with your PR we would basically change all typescript angular clients
> So my suggestion is to change code just to handle specific case that you point out.

It's not lower level as in Using a Method of a different Service that is encapsulated - imo the Methods for the individual Methods are about convenience and not abstraction.
Changing the code to just handle the case of a DELETE request would result in more complexity in the open-api implementation and also make it harder to understand. That was my primary motivation for going this ways.

Also, I checked the angular implementation of the individual Methods and they directly forward to the `request` method.
Therefore, on a technical level, the only change is that body, if necessary for the call as defined by yaml, is always provided, which may not make sense for some request methods.
However, that would be a mistake in the yaml file for defining a body where it isn't possible.

All in all, taking into consideration the upcoming major version, I would tend to this solution as it results in a cleaner code-base.

> so it is not just DELETE it is GET as well
> As @dartzki point out by spec body is allowed in DELETE
> so maybe method should be changed`
> 
> ```
>     public boolean isBodyAllowed() {
>         return Arrays.asList("DELETE""PUT", "PATCH", "POST").contains(httpMethod.toUpperCase(Locale.ROOT));
>         // or just use 
>         return  !isRestfulShow(); 
>    
>     } 
> ```

Since I don't know the codebase, I wouldn't hazard to change such a central method. Shouldn't the logic for using the body parameter be the same as the logic for adding the body parameter to the service method?
Before a delete request would result in a service method that accepts a `body` param as input, but then doesn't use it in the method implementation.

topce:
@dartzki 
**Thank you for fast reply and contribution!**
I understand your point of view.
There are always tradeoffs : 
Either you make it more maintainable (easier to read code generation template)
or more safe for existing users and IMHO better generated code and more complicated template.

In my opinion it would be better that Angular changed there API 
so delete methods accept body parameters as French would say "comme il faut",
but they did not ?!
You could also vote (like) there if it is important for you
so you have moral satisfaction , probably would not change final outcome.

IMHO OpenAPITools do much better job keeping community happy  than Angular,
that why I contribute to this project not to other one ;-)

**But because Angular is ignorant should we change all our generated angular typescript code 
to fix problem that IMHO does not affect to many people and generate more code,
because one google employee decided  to refuse:
https://github.com/angular/angular/pull/37894**

**In my opinion not .**

However if you tested all supported version of Angular and you are sure 
that if existing users who switch to  5.0.0 would not have problem 
for me it is OK to accept PR otherwise  I would reject it.   

Generated code that I see does not prove that you fix initial issue
and possibly you introduce new issues  (because in client.ts HttpClient of angular there are so many overload 
also they use "any" on 103 places any that is not good sign if you want to right strong typed programs)
and there are hesitant to change any of code because they are afraid that they will break somebodies code.
**Why should we value our users less ?**

Regarding isBodyAllowed you rightly point out that body is allowed for delete .
so name of this method is confusing or implementation is wrong
that why I asked for @wing328  to take a look and contact maintainers   of erlang and java 
that use isBodyAllowed  and asked them if they  think they code will work if they add DELETE  to allowed body

amakhrov:
@topce You raised a good point about supporting angular 5. However, it looks like .request() method has the same signature even in angular 5: https://devdocs.io/angular~5/api/common/http/httpclient . So it doesn't seem to be an issue.

Also I don't believe it's fair to compare this PR and the angular team's decision on the issue. In Angular it would be public api change, affecting many users. But this PR in open-generator changes the internals of the generated code without changing surfaced api at all.
Could you please clarify how this PR could negatively impact users?

jimschubert:
> Also I don't believe it's fair to compare this PR and the angular team's decision on the issue. In Angular it would be public api change, affecting many users. But this PR in open-generator changes the internals of the generated code without changing surfaced api at all.
> Could you please clarify how this PR could negatively impact users?

It looks like [angular now allows DELETE body](https://github.com/angular/angular/blob/d1ea1f4c7f3358b730b0d94e65b00bc28cae279c/packages/common/http/src/request.ts#L28).

I agree that this shouldn't have any impact on users. We're wrapping angular's helper functions which all defer to request anyway. We're wrapping these calls in operations, and the users don't have access to the function call. We're not changing angular's http.request/delete/etc. @topce could you clarify that concern?

topce:
@amakhrov  my concern is that could introduce new errors
because it change generated code to call request instead of 
get delete put post patch methods
@jimschubert IMHO it is fair  to compare PRs because we change generated code
to call IMO low level function request instead of get delete post ...
and in this PR they reference Angular PR.

It is low level because get post etc internally call
Request

If 3 of you feel confident about change
and sure that previously generated code calls
correct  request method please proceed with a change.
Anyway if we broke something with this change 
Users can always switch back to version 4.3.1 until potential problem is fixed ;-)

topce:
@jimschubert 
> It looks like angular now allows DELETE body.
No it does not in delete method https://angular.io/api/common/http/HttpClient#delete
that why @dartzki  change it to use everywhere request there he can pass body with delete

You see also in excellent  documentation of Angular that delete methods have 15 overloads
get 15 
request 17 etc
before we use get/delete/put/post and my concern if we do not call
good  overloaded method maybe we break something that was working before.

topce:
@jimschubert 

Also another problem that @dartzki  and @amakhrov  
point out is method 
```
    /**
     * Check if body param is allowed for the request method
     *
     * @return true request method is PUT, PATCH or POST; false otherwise
     */
    public boolean isBodyAllowed() {
        return Arrays.asList("PUT", "PATCH", "POST").contains(httpMethod.toUpperCase(Locale.ROOT));
    }
```
Name of method isBodyAllowed but as @dartzki correctly  said body is allowed in delete
also this method is used in erlang and java template.
@amakhrov  argue that this method is maybe not needed anymore 
but that need to be check  with erlang and java maintainers.

Should we change implementation of method  to allow body for delete !?
That why I asked @wing328  to check this because it impacts other  code generators.

jimschubert:
I mean, we should match the http spec, not define our own rules. I can't speak for why angular has defined their own rules.

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
dartzki: fixed it, now body is only emitted if actually used

modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
dartzki: sure, I switched it to single quotes

dartzki:
@jimschubert @topce @wing328 Is there anything I can do to move this along?
Otherwise I'm forced to use my own fork, which I would prefer to avoid...

macjohnny:
@dartzki instead of forking you can also provide a custom version of specific templates, see https://openapi-generator.tech/docs/templating

but in order to move forward with this PR, maybe it would be less risky to introduce an option to allow the user to decide whether to use `httpClient.post()` vs `httpClient.request('post', ` ?

Mrsevic:
Any updates on this?

dartzki:
> @dartzki instead of forking you can also provide a custom version of specific templates, see https://openapi-generator.tech/docs/templating
> 
> but in order to move forward with this PR, maybe it would be less risky to introduce an option to allow the user to decide whether to use `httpClient.post()` vs `httpClient.request('post', ` ?

I can do that, If this means the PR will be merged.

However, from a personal perspective I see more downsides than upsides to hiding the change behind an option:
- the template will become even more complex (and imo there's already a lot of (necessary) complexity there)
- future changes are more likely to only take one of the implementations into account and breaking the other
- when/if the angular code will be adapted to actually support DELETE with body, the option will have to be removed by all which had it enabled. Imo it's an implementation detail and should not require users of openAPI generator do change their config either way.

That said, @macjohnny @topce If you say you'll approve the PR with the new option, I'll try to add it

topce:
Hi @dartzki 

I will accept this PR if you add option to generate request code.
By default it should be switched of so generated code should be same like before ,
because in  yaml file that we use for testing there is not delete with body.

You are right there is always pros and cons but I agree with @macjohnny  this is less risky.

So if somebody use delete with body he can switch on option that you add and generate code.
In my opinion not lot of people use this delete with body.

Also thanks to you, I found in our code problematic method 
isBodyAllowed with wrong implementation (see my previous comment)
maybe also it should be fixed to check with @jimschubert  and @wing328 
because it influence multiple clients.


You are right template is more complex , but that is price to pay.

Regarding other two downsides I am not afraid because you see Angular 
was hesitant to change their public API even it is oblivious there is error in API 
delete method without body parameter.
 
So if Angular does not change their HttpClient public API I do not expect that this part of application change a lot.
Even if they change method delete to add new body parameter ,we do not need to revert your changes.
Because maybe some users still want to use option that you will introduce.
Also your option should be tiny more efficient because there is on function call less

If you implement option it is not breaking change otherwise it is . 

Thank you  for your contribution!

dartzki:
superseded by https://github.com/OpenAPITools/openapi-generator/pull/10751

Issue #1309, opened 2018-10-24
Allow generation of reactive APIs w/o wrapping Mono<ResponseEntity>


jrobison153:
##### Description
We've been looking at the generated Spring Web Reactive interfaces and the response type wrappings seem superfluous, this is using version openapi-generator:3.3.1. For an API that returns a collection of type T, we get a Java interface with the following response signature

Mono<ResponseEntity<Flux<T>>>

From all of our testing this behaves identically to Flux<Foo> from a reactive client's perspective. The additional Mono<ResponseEntity> adds additional code to each controller to add the type wrappings.

Since the ServerWebExchange is already present and gives all the control over the HTTP session needed. It seems the generated code could be simplified

The suggestion is that an option could be added to generate the reactive interfaces w/o the wrapping Mono<ResponseEntity> 

##### openapi-generator version

openapi-generator:3.3.1

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/pull/571

cbornet:
In that case, that should be done for both reactive and non-reactive.
In non-reactive you also have the possibility to add headers by injecting the request-scoped `NativeWebRequest`.

P3tronius:
Any update on this? The signatures are barely usable as-is.

gmcouto:
I don`t like the ResponseEntity on the interface as well... there should be an option to remove it.

kdebski85:
I confirm the issue.
I am using openapi-generator-maven-plugin 4.3.0 with

```
<generatorName>spring</generatorName>
<library>spring-boot</library>
<configOptions>
  <reactive>true</reactive>
  <java8>true</java8>
  <interfaceOnly>true</interfaceOnly>
  <skipDefaultInterface>true</skipDefaultInterface>
</configOptions>
```

and the generator creates methods that return  `Mono<ResponseEntity<Flux<T>>>`

It should be just `Flux<T>` and `Mono<T>`.

erohana:
any workaround for spring reactive? I'm experiencing the same issues

gerardbosch:
I understood from https://github.com/OpenAPITools/openapi-generator/pull/571 that the "inconvenience" is not only the complex signature of return type, but the fact that `ResponseEntity.status()` need to be set upfront before fetching data which can be not always possible if the status depends on the data on the stream. Is that right?

[...]

PR #7566, opened 2020-10-01
generate simpler types for spring-boot reactive

maddingo:
resolves OpenAPITools/openapi-generator#1309
This PR generates Mono<T> for single-entity return types and Flux<T> for list-type return types.


maddingo:
This change generates code that looks nice but doesn't play along with Spring. investigating...

PR #7570, opened 2020-10-01
[core] Enable multiple oauth2 flows

### LINKED ISSUE #7193: [BUG] Oauth flows are not generated when there is more than 1
##### Description

When working with [multiple oauth flows](https://swagger.io/docs/specification/authentication/oauth2/) in definition and then using some kind of templates to use those flows, only first flow that is found is used. Problem is in [DefaultCodegen](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L4405) where it's not iterated through all flows, but just first flow is taken.

**4.2.3**

```yaml
openapi: 3.0.0
info:
  title: test api
  version: 0.0.1
paths:
  /imports:
    post:
      summary: Creates import
      operationId: createImport
      responses:
        201:
          description: created
      security:
        - oauth_auth:
            - import:create
  /imports/{importId}/state:
    put:
      summary: Changes import state
      operationId: changeImportState
      parameters:
        - name: importId
          in: path
          required: true
          schema:
            type: string
      responses:
        200:
          description: State changed
      security:
        - oauht_auth:
            - import:process
components:
  securitySchemes:
    oauth_auth:
      type: oauth2
      flows:
        password:
          tokenUrl: "../auth/realms/master/protocol/openid-connect/token"
          scopes:
            import:create: create import
        clientCredentials:
          tokenUrl: "../auth/realms/master/protocol/openid-connect/token"
          scopes:
            import:create: create import
            import:process: process import
```

##### Suggest a fix

Loop inside method `DefaultCodegen#fromSecurity` also through all flows and then they should be generated.

bilak:
This PR fixes #7193 generation of multiple oauth flows as it should be possible to do so by [specification](https://swagger.io/docs/specification/authentication/oauth2/)  

cc: @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04) @lwlee2608 (2019/10) @bkabrda (2020/01)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -1694,7 +1694,7 @@ public class DefaultCodegen implements CodegenConfig {
         } else if (Boolean.TRUE.equals(codegenParameter.isBoolean)) {
             codegenParameter.example = "true";
         } else if (Boolean.TRUE.equals(codegenParameter.isLong)) {
-            codegenParameter.example = "7•89";
+            codegenParameter.example = "789";
         } else if (Boolean.TRUE.equals(codegenParameter.isInteger)) {
             codegenParameter.example = "56";
         } else if (Boolean.TRUE.equals(codegenParameter.isFloat)) {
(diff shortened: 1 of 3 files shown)


jimschubert:
Looks good. Marked as a potential breaking change because people with custom templates may have to make changes.

Issue #7453, opened 2020-09-19
[BUG][CLI] crash when evaluating definitions without parameters


codymikol:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

When you provide openapi-generator an openapi definition that includes a url with path variables, but no defined parameters, it will throw a null pointer exception. 

```
Exception in thread "main" java.lang.NullPointerException
        at java.base/java.util.HashSet.<init>(HashSet.java:120)
        at org.openapitools.codegen.config.CodegenConfigurator.toContext(CodegenConfigurator.java:464)
        at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:507)
        at org.openapitools.codegen.cmd.Generate.execute(Generate.java:423)
        at org.openapitools.codegen.cmd.OpenApiGeneratorCommand.run(OpenApiGeneratorCommand.java:32)
        at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:61)
```

##### openapi-generator version

5.0.0-SNAPSHOT

##### OpenAPI declaration file content or url

https://github.com/codymikol/openapi-generator-kotlin-bug-demo/blob/master/generated-openapi-spec.json

##### Generation Details

This seems to happen regardless of generator-name, and the only other CLI options I am providing are the input-spec and output.

##### Steps to reproduce

I have created a minimal repro repository over [here](https://github.com/codymikol/openapi-generator-kotlin-bug-demo)
this includes a shell script with all of the exact steps that I took. The generated openapi spec is hardcoded in the repo, but I have also included a minimal Spring project that generates the spec via springfox 3 if you'd like to generate that yourself.

Note that the script uses the latest published openapi-generator and will need some tweaking to point at your own build. I have tested with both and they have the same result

##### Related issues/PRs

[...]

Issue #1376, opened 2018-11-05
[JAVA] Server and models with readOnly properties


tsiq-karold:
##### Description

Based on the description of `readOnly` from https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object which states that a response may contain a read-only property but a request must not contain it, I would expect generated client code to only contain getters for such objects and generated server code to contain both getters and setters.  The client side behaves as expected, the server generated code however does not allow setting such properties.

Is this a bug, a misunderstanding on my part or if neither then how are we supposed to set such properties in server code?

##### openapi-generator version

Master

##### OpenAPI declaration file content or url

```json
{
  "swagger": "2.0",
  "info": {
    "version": "0.0.1",
    "title": "Test"
  },
  "basePath": "/",
  "paths": {
    "/": {
      "get": {
        "operationId": "theGet",
        "responses": {
          "200": {
            "description": "response",
            "schema": {
              "$ref": "#/definitions/ResponseMessage"
            }
          }
        }
      }
    }
  },
  "definitions": {
    "ResponseMessage": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string",
          "readOnly": true
        },
        "value": {
          "type": "string",
          "readOnly": true
        }
      }
    }
  }
}
```

##### Command line used for generation

`java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i api.json -g jaxrs-jersey -o output`

##### Steps to reproduce

Generate code as mentioned above then attempt to create and return a `ResponseMessage` within `DefaultApiServiceImpl.theGet()`.

##### Related issues/PRs

Not that I could find.

##### Suggest a fix/enhancement

One possible solution is to remove the if statement around setter and builder method generation for read-only fields in the appropriate templates.  In my case this would be these two statements: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaJaxRS/pojo.mustache#L33 and https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/JavaJaxRS/pojo.mustache#L90.  The POJO templates for the client are separate so this would work.  However, I'm not sure it's ideal doing this in each set of templates especially if any generators share model templates between client and server.

[...]

PR #7597, opened 2020-10-04
CHORE[DEPENDENCIES]: bump swagger-parser => 2.0.22

codymikol:
this fixes a bug where the parser would return
null when missing parameters rather than reporting
some useful error.

Fixes N/A

This bumps the swagger-parser library form 2.0.20 to 2.0.22, this fixes a bug in the parser library that would cause openapi-generator to mysteriously fail when parameters are not defined in your api spec

My original issue was opened over here: https://github.com/OpenAPITools/openapi-generator/issues/7453 , but my fix just prevented openapi-generator from throwing a null pointer exception. This update fixes the underlying problem to that issue.

This is the specific PR that fixed that issue: https://github.com/swagger-api/swagger-parser/pull/1410 (thank you, I appreciate you)

2.0.22 changes: https://github.com/swagger-api/swagger-parser/releases/tag/v2.0.22

2.0.21 changes: https://github.com/swagger-api/swagger-parser/releases/tag/v2.0.21

## Changes Related to this version bump

The petstore examples have a new parameter relating to: [Add extension to keep track of original swagger version in PR #1374](https://github.com/swagger-api/swagger-parser/pull/1374)

The InlineModelResolver `fixStringModel` method needs to be adjusted per [Fix for Empty and Blank query parameters issue PR - #1376](https://github.com/swagger-api/swagger-parser/pull/1376) as it will no longer return null values for empty url parameters, but rather empty strings.

There is also a test that essentially just proves the same getExample method will return null, I have changed the test to instead just assert the returned value is an empty string. I do think that this is an unnecessary test for this project as it is essentially testing a 3rd party dependency and would be fine with removing it based on feedback.

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/InlineModelResolverTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/InlineModelResolverTest.java
@@ -28,6 +28,7 @@ import io.swagger.v3.oas.models.responses.ApiResponse;
 import io.swagger.v3.oas.models.responses.ApiResponses;
 import io.swagger.v3.parser.core.models.ParseOptions;
 import org.openapitools.codegen.utils.ModelUtils;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import java.util.HashMap;
@@ -927,7 +928,7 @@ public class InlineModelResolverTest {
         assertEquals("#/components/schemas/EmptyExampleOnStringTypeModels", schema.getItems().get$ref());
 
         assertTrue(ModelUtils.getReferencedSchema(openAPI, schema.getItems()) instanceof StringSchema);
-        assertNull(ModelUtils.getReferencedSchema(openAPI, schema.getItems()).getExample());
+        Assert.assertSame(ModelUtils.getReferencedSchema(openAPI, schema.getItems()).getExample(), "");
     }
 
     @Test
--- a/samples/client/petstore/go/go-petstore/api/openapi.yaml
+++ b/samples/client/petstore/go/go-petstore/api/openapi.yaml
@@ -2126,3 +2126,4 @@ components:
     http_basic_test:
       scheme: basic
       type: http
+x-original-swagger-version: "2.0"
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/InlineModelResolver.java
@@ -472,12 +472,20 @@ public class InlineModelResolver {
      * @param m Schema implementation
      */
     private void fixStringModel(Schema m) {
-        if (m.getType() != null && m.getType().equals("string") && m.getExample() != null) {
+        if (schemaIsOfType(m, "string") && schemaContainsExample(m)) {
             String example = m.getExample().toString();
-            if (example.substring(0, 1).equals("\"") && example.substring(example.length() - 1).equals("\"")) {
+            if (example.startsWith("\"") && example.endsWith("\"")) {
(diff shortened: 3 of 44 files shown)


codymikol:
From the looks of the CI, there will be quite a bit more work than just bumping the version number :/

codymikol:
It appears all of these failing tests come from changes in 2.0.21 : (

codymikol:
It seems to me that there was some workaround in openapigenerator for string values having escape characters baked into them that is no longer the case, I have an ugly fix that makes most of the tests pass, I'll try and clean it up farther.

codymikol:
There is a test `emptyExampleOnStringTypeModels` that seems to assert the swagger-models schema object has a null value for getExample(), which is no longer the case after this update. For the time being I have changed it to assert that the getExample() is an empty string, but seeing as this is testing outside the scope of openapi-generator, I think we can just go ahead and remove this test. Feedback on that would be great.

codymikol:
: /  It looks like all CIs are passing except circleci, but circleci won't let me look at the results of the build unless I give it read / write access to my repos and I'm not too keen on that.

modules/openapi-generator/src/test/java/org/openapitools/codegen/InlineModelResolverTest.java
```diff
@@ -927,7 +927,7 @@ public void emptyExampleOnStringTypeModels() {
         assertEquals("#/components/schemas/EmptyExampleOnStringTypeModels", schema.getItems().get$ref());
 
         assertTrue(ModelUtils.getReferencedSchema(openAPI, schema.getItems()) instanceof StringSchema);
-        assertNull(ModelUtils.getReferencedSchema(openAPI, schema.getItems()).getExample());
+        assertSame("", ModelUtils.getReferencedSchema(openAPI, schema.getItems()).getExample());
```
jimschubert, line 930: testng format is `assertSame(action, expected)`

samples/client/petstore/go/go-petstore/api/openapi.yaml
```diff
@@ -2126,3 +2126,4 @@ components:
     http_basic_test:
       scheme: basic
       type: http
+x-original-swagger-version: "2.0"
```
jimschubert, line 2129: @sebastien-rosset I see you added this to swagger-parser, but it doesn't seem to be consumed anywhere. 

Do you need to create a follow up to consume this as mentioned in #6290? 

> Limitation: due to issue swagger-parser #1369, achieving OAS compliance is not possible when the input document uses the OAS 2.0 schema. For OAS 2.0 documents, the legacy behavior should be used by setting the disallowAdditionalPropertiesIfNotPresent CLI option to true.

jimschubert (approved):
lgtm
the test assertion just needs to be flipped

pinging @OpenAPITools/generator-core-team for additional set of eyes, as I know there was some reason we weren't bumping to 2.0.21 but can't recall what it was.

samples/client/petstore/go/go-petstore/api/openapi.yaml
sebastien-rosset: yes, there needs to be a follow up. Actually even better would be to ensure the 2.0 parser is converting additionalProperties properly.
With `x-original-swagger-version`, at least we know the source is 2.0, but if the converter is not converting properly to 3.0 (and retaining the semantics), it's still going to be a problem.

modules/openapi-generator/src/test/java/org/openapitools/codegen/InlineModelResolverTest.java
codymikol: Switched to testng assert and flipped

Issue #7426, opened 2020-09-15
[Python] Incorrect return types if multiple responses are defined 


ksvirkou-hubspot:
### Description
If you define multiple responses, the codegen will only support the first one (typically 200).

## openapi-generator version
5.0.0-SNAPSHOT

## OpenAPI declaration file content or url

<img width="735" alt="Снимок экрана 2020-09-17 в 16 24 12" src="https://user-images.githubusercontent.com/57258237/93476330-56d0d980-f902-11ea-8c1a-32a56a4508c3.png">

Correctly generate the return-object based on the HTTP Status Code in the Response.

## Similar Issue for PHP

[link](https://github.com/OpenAPITools/openapi-generator/issues/125)

ksvirkou-hubspot:
## PR
https://github.com/OpenAPITools/openapi-generator/pull/7427

Issue #125, opened 2018-05-22
[PHP] Incorrect return types if multiple responses are defined


fMads:
##### Description

If you define multiple responses, the codegen will only support the first one (typically 200).

##### openapi-generator version

3.0.0-SNAPSHOT

##### OpenAPI declaration file content or url

```yaml
  /path
    post:
      responses:
        200:
          schema:
            $ref: '#/definitions/EmptyResponse'
        202:
          schema:
            $ref: '#/definitions/InBackgroundResponse'
```

##### Suggest a fix/enhancement

Correctly generate the return-object based on the HTTP Status Code in the Response.
Also fix ReturnType info in DocBlock

ackintosh:
@fMads #128 is merged! Thanks for your contribution.

slind85:
This is still happening for Java

PR #7634, opened 2020-10-09
[Ruby] Incorrect return types if multiple responses are defined

uyermolkin-hubspot:
### Description
If you define multiple responses, the codegen will only support the first one (typically 200).

### openapi-generator version
4.3.1 and higher

### OpenAPI declaration file content or url
93476330-56d0d980-f902-11ea-8c1a-32a56a4508c3

### Similar Issue for PHP

[https://github.com/OpenAPITools/openapi-generator/issues/125](url)

### Similar Issue for Python
[https://github.com/OpenAPITools/openapi-generator/issues/7426](url)


uyermolkin-hubspot:
@spacether could you please check this PR? Thanks!

spacether (approved):
looks good to me

spacether (changes requested):
Can you fix the travis ci errors?
There are failing ruby tests [there](https://travis-ci.org/github/OpenAPITools/openapi-generator/builds/736342547).

uyermolkin-hubspot:
> Can you fix the travis ci errors?
> There are failing ruby tests [there](https://travis-ci.org/github/OpenAPITools/openapi-generator/builds/736342547).

I'm confused, because these tests are failed not because of my changes. It fails even on master branch. Can you take a look? @spacether

spacether:
Ruby Technical Committee:
@cliffano (2017/07) @zlx (2017/09) @autopp (2019/02)
Any insights here?

atanasiuk-hubspot:
Hi, any updates on this?

samples/client/petstore/ruby-faraday/lib/petstore/api/pet_api.rb
```diff
@@ -183,20 +189,26 @@ def find_pets_by_status_with_http_info(status, opts = {})
...
-      return_type = opts[:debug_return_type] || 'Array<Pet>'
-
       # auth_names
       auth_names = opts[:debug_auth_names] || ['petstore_auth']
 
+      # return_type
+      return_type = opts[:debug_return_type]
+
+      return_types_map = {
+        200 => 'Array&lt;Pet&gt;',
```
spacether, line 199: This map value is incorrect. Can you update it to 'Array<Pet>'?

samples/client/petstore/ruby-faraday/lib/petstore/api/pet_api.rb
```diff
@@ -183,20 +189,26 @@ def find_pets_by_status_with_http_info(status, opts = {})
...
+        400 => '',
```
spacether, line 200: Should this be empty string or nil?

modules/openapi-generator/src/main/resources/ruby-client/api.mustache
```diff
@@ -175,20 +175,38 @@ module {{moduleName}}
...
       new_options = opts.merge(
         :operation => :"{{classname}}.{{operationId}}",
         :header_params => header_params,
         :query_params => query_params,
         :form_params => form_params,
         :body => post_body,
         :auth_names => auth_names,
-        :return_type => return_type
+        :return_type => return_type,
+        :return_types_map => return_types_map
```
spacether, line 209: You are using two parameters to pass in the same data here:
- return_type
- return_types_map
How about only passing in return_types_map?
If the user sets a :debug_return_type should we set all values of the map to  :debug_return_type as it looks the old code replace the return type with the debug_return_type if it is passed in.

spacether (changes requested):
Thanks for this PR overall it looks good.
- One mapped value is incorrect and needs a fix.

I have a questions about:
- using empty string vs nil
- asking if we can use one parameter to store the data types rather than two for simplicity

PR #7636, opened 2020-10-09
Fixes #7635: typescript-inversify generator wrongly handles array typ…

### LINKED ISSUE #7635: typescript-inversify generator wrongly handles array type parameters
##### Description
api.service.mustache generates such code 

```
if (param !== undefined) {
    param.forEach((element) => {
    queryParameters.push('param='+encodeURIComponent(String(param)));
})
}
```
Instead of 
```
if ($param !== undefined) {
    $param.forEach((element) => {
    queryParameters.push('param='+encodeURIComponent(String(element)));
})
}
```
Wrong variable is used in loop. 

##### Suggest a fix
I'll provide PR.

michalzubkowicz:
Wrong variable is used in loop generating array type url parameters


macjohnny (approved):
LGTM

PR #7637, opened 2020-10-09
[python-experimental] implement `in` operator for model classes

keymone:
### PR checklist
 
Allow to check for attribute presence in model using `in` operator.

Unless I'm missing something, right now the only way to correctly get an attribute which may not be present is this:

    try:
        attr_value = instance.attr_name
    except ApiAttributeError:
        attr_value = None

which is quite cumbersome.

@spacether

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
@@ -69,3 +69,19 @@
                 "the same".format(name, type(self).__name__),
                 path_to_item
             )
+
+    def __contains__(self, name):
+        """this allows us to use `in` operator: `'attr' in instance`"""
+
+        if name in self.required_properties:
+            return name in self.__dict__
+
+        model_instances = self._var_name_to_model_instances.get(
+            name, self._additional_properties_model_instances)
+
+        if model_instances:
+            for model_instance in model_instances:
+                if name in model_instance._data_store:
+                    return True
+
+        return False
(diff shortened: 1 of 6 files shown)


spacether (approved):
Thank you for the PR. This looks great!

keymone:
@spacether relevant question - is there value in raising an exception instead of returning `None` for access of legit attribute that doesn't have value? now that it's possible to test for presence of an attribute, it's easy to differentiate between that and legit `None` value.

code i want to simplify is this:

    value = instance.attr if 'attr' in instance else None
    if value is not None:
        # do something

with default `None` it's just be:

    if instance.attr is not None:
        # do something

thoughts?

spacether:
In my opinion there is value. Attributes can be nullable which is different than being unset.
Let's say
attribute_a has value None if we access it we should get None
Now let's say attribute a is unset if we try to access it directly with model.attribut_a we should get some kind of exception. This lets us differentiate between these use cases.

spacether:
If you want this functionality why not ad a default missing input argument to getattt in ModelNormal and ModelComposd? That would make it work like python dict.get.
Then one could do model.get(param_name, default_value_if_missing)

keymone:
> Now let's say attribute a is unset if we try to access it directly with model.attribute_a we should get some kind of exception. This lets us differentiate between these use cases.

we definitely can differentiate in both cases via exception or by using `'attr' in instance`, i just think that using exceptions for that leads to more complex code than necessary. it is much more often that one needs to react to `None` value regardless if it was set or the attribute is missing, than to differentiate between set and unset `None`.

> Then one could do model.get(param_name, default_value_if_missing)

yes, that's another method i want to add, but it would be nice if using attributes directly was more convenient. right now i have to wrap in try/except way too much :(

> Why not use model._data_store.get(name)?

i don't think all attributes are in there? required attributes seem to be in model.__dict__ directly.

spacether:
> i don't think all attributes are in there? required attributes seem to be in model.__dict__ directly.

The attribute that are not in _data_store is data about the class or the composed instance.
I think that all payload key value pairs are in _data_store.

PR #7653, opened 2020-10-11
[Default] Generate referenced superclass even without properties

### LINKED ISSUE #7638: [BUG] If superclass has no properties, it will not be generated, but derived class does extend not existing/generated class
##### Description

I use the openapi gradle-plugin to generate java classes  - but if the superclass has no properties, the class will not be generated but the derived class extends the non existing class.
So the code does not compile.

Using the command-line I see the following message: 

> Model Pet not generated since it's a free-form object

As a workaround we modify the openapi json to add some dummy attribute in the superclass, so that it will be generated.

org.openapi.generator 4.3.1

```json
{
    "openapi": "3.0.1",
    "info": {
        "title": "Abstract Generation problem",
        "description": "Abstract Generation problem",
        "version": "1.0.0-SNAPSHOT"
    },
    "servers": [],
    "paths": {},
    "components": {
        "schemas": {
            "Pet": {
                "title": "AbstractPet",
                "type": "object",
                "description": "model containing all the details of a pet",
                "discriminator": {
                    "propertyName": "petTypeName"
                }
            },
            "Dog": {
                "title": "DogDetails",
                "description": "The model containing the details of the dog",
                "allOf": [
                    {
                        "$ref": "#/components/schemas/Pet"
                    },
                    {
                        "title": "DogDetails",
                        "type": "object",
                        "properties": {
                            "legs": {
                                "type": "string",
                                "description": "the number of legs"
                            }
                        },
                        "description": "The model containing the details of the dog"
                    }
                ]
            }
        }
    }
}
```

##### Generation Details

java -jar openapi-generator-cli-4.3.1.jar generate -i openapi.json -g spring -o ./temp

##### Generated Models

There were two java classes generated:

- Dog.java
- DogDetails.java

Dog.java does extend Pet, that does not exist!

```java
public class Dog extends Pet  {
  @JsonProperty("legs")
  private String legs;

  public Dog legs(String legs) {
    this.legs = legs;
    return this;
  }
...
```

##### Workaround

change definition of Pet to:

```json
            "Pet": {
                "title": "AbstractPet",
                "type": "object",
                "properties": {
                    "dummyProperty": {
                        "type": "string",
                        "description": "Workaround - OpenAPI generator does not consider definitions without properties"
                    }
                },
                "description": "model containing all the details of a pet",
                "discriminator": {
                    "propertyName": "petTypeName"
                }
            },
```

Zomzog:
I've done the change on the default branch because I can reproduce with Java and Kotlin and I think it may be an issue on other languages. So I add the core team and not just the Java technical committee.

May fix #7638

@wing328
@jimschubert
@cbornet
@ackintosh
@jmini
@etherealjoy
@spacether

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -1072,7 +1072,7 @@ public class ModelUtils {
         }
 
         if (schema != null && StringUtils.isNotEmpty(schema.get$ref())) {
-            String simpleRef = ModelUtils.getSimpleRef(schema.get$ref());
+            String simpleRef = getSimpleRef(schema.get$ref());
             if (importMappings.containsKey(simpleRef)) {
                 LOGGER.debug("Schema unaliasing of {} omitted because aliased class is to be mapped to {}", simpleRef, importMappings.get(simpleRef));
                 return schema;
@@ -1088,7 +1088,7 @@ public class ModelUtils {
                 if (isGenerateAliasAsModel(ref)) {
                     return schema; // generate a model extending array
                 } else {
-                    return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())),
+                    return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())),
                             importMappings);
                 }
             } else if (isComposedSchema(ref)) {
@@ -1101,22 +1101,51 @@ public class ModelUtils {
                         return schema; // generate a model extending map
                     } else {
                         // treat it as a typical map
-                        return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())),
+                        return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())),
                                 importMappings);
                     }
                 }
             } else if (isObjectSchema(ref)) { // model
                 if (ref.getProperties() != null && !ref.getProperties().isEmpty()) { // has at least one property
                     return schema;
-                } else { // free form object (type: object)
-                    return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())),
+                } else {
+                    // free form object (type: object)
+                    if (hasValidation(ref)) {
+                        return schema;
+                    } else if (hasDescendants(simpleRef, openAPI)) {
+                        return schema;
+                    }
+                    return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())),
                             importMappings);
                 }
             } else {
-                return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())), importMappings);
+                return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())), importMappings);
             }
         }
         return schema;
+    }
+
+    private static boolean hasDescendants(String simpleRef, OpenAPI openAPI) {
+        for (Map.Entry<String, Schema> entry : getSchemas(openAPI).entrySet()) {
+            if (!isSameSchema(simpleRef, entry.getKey())) {
+                Schema schema = entry.getValue();
+                if (isComposedSchema(schema)) {
+                    ComposedSchema composedChild = (ComposedSchema) schema;
+                    boolean b = composedChild.getAllOf()
+                            .stream().map(Schema::get$ref)
+                            .map(ModelUtils::getSimpleRef)
+                            .anyMatch(simpleRef::equals);
+                    if (b) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    private static boolean isSameSchema(String simpleRef, String schema) {
+        return simpleRef.equals(schema);
     }
 
     /**
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -1127,7 +1127,7 @@ public class ModelUtils {
 
     private static boolean hasDescendants(String simpleRef, OpenAPI openAPI) {
         for (Map.Entry<String, Schema> entry : getSchemas(openAPI).entrySet()) {
-            if (!isSameSchema(simpleRef, entry.getKey())) {
+            if (!simpleRef.equals(entry.getKey())) {
                 Schema schema = entry.getValue();
                 if (isComposedSchema(schema)) {
                     ComposedSchema composedChild = (ComposedSchema) schema;
@@ -1144,10 +1144,6 @@ public class ModelUtils {
             }
         }
         return false;
-    }
-
-    private static boolean isSameSchema(String simpleRef, String schema) {
-        return simpleRef.equals(schema);
     }
 
     /**
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
@@ -1132,7 +1132,9 @@ public class ModelUtils {
                 if (isComposedSchema(schema)) {
                     ComposedSchema composedChild = (ComposedSchema) schema;
                     boolean b = composedChild.getAllOf()
-                            .stream().map(Schema::get$ref)
+                            .stream()
+                            .map(Schema::get$ref)
+                            .filter(Objects::nonNull)
                             .map(ModelUtils::getSimpleRef)
                             .anyMatch(simpleRef::equals);
                     if (b) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2857,8 +2857,8 @@ public class DefaultCodegen implements CodegenConfig {
     }
 
     protected List<MappedModel> getAllOfDescendants(String thisSchemaName, OpenAPI openAPI) {
-        ArrayList<String> queue = new ArrayList();
-        List<MappedModel> descendentSchemas = new ArrayList();
+        ArrayList<String> queue = new ArrayList<>();
+        List<MappedModel> descendentSchemas = new ArrayList<>();
         Map<String, Schema> schemas = ModelUtils.getSchemas(openAPI);
         String currentSchemaName = thisSchemaName;
         Set<String> keys = schemas.keySet();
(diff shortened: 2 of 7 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1101,24 +1101,55 @@ public static Schema unaliasSchema(OpenAPI openAPI,
...
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    private static boolean isSameSchema(String simpleRef, String schema) {
+        return simpleRef.equals(schema);
+    }
```
spacether, line 1151: nit:
How about removing this function? It is one line and you can use `simpleRef.equals(schema)` in the call site instead. Also there is only one call site so it may not make sense to make this a function.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1101,24 +1101,55 @@ public static Schema unaliasSchema(OpenAPI openAPI,
...
                 if (ref.getProperties() != null && !ref.getProperties().isEmpty()) { // has at least one property
                     return schema;
-                } else { // free form object (type: object)
-                    return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())),
+                } else {
+                    // free form object (type: object)
+                    if (hasValidation(ref)) {
+                        return schema;
+                    } else if (hasDescendants(simpleRef, openAPI)) {
+                        return schema;
```
spacether, line 1116: @jimschubert this will turn on interface schemas on everywhere
Is that what we want?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
```diff
@@ -1101,24 +1101,55 @@ public static Schema unaliasSchema(OpenAPI openAPI,
...
             } else if (isObjectSchema(ref)) { // model
                 if (ref.getProperties() != null && !ref.getProperties().isEmpty()) { // has at least one property
                     return schema;
-                } else { // free form object (type: object)
-                    return unaliasSchema(openAPI, allSchemas.get(ModelUtils.getSimpleRef(schema.get$ref())),
+                } else {
+                    // free form object (type: object)
+                    if (hasValidation(ref)) {
+                        return schema;
+                    } else if (hasDescendants(simpleRef, openAPI)) {
```
spacether, line 1115: How about replacing this with:
`} else if (getAllOfDescendants(simpleRef, openAPI).size() > 0) {`
like we do in python-experimental?
That way you can delete the new hasDescendants function and use our existing getAllOfDescendants.

jimschubert: I prefer the easier to read `hasDescendants()`.

`getAllOfDescendants(…).size() > 0` could easily be read as getting all descendants rather than getting the descendants for allOf composition.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
jimschubert: I agree here. `isSameSchema(string, string)` doesn't give us any way to control drift of which strings are passed in here. This would better be suited as a comment above the conditional:

```
// check if model names are the same
if (!simpleRef.equals(entry.getKey())) {
```

Also, I wonder if checking `simpleRef` like this will work in all cases? For example, if I have `#/components/schemas/Cat` and `#/components/responses/Cat`, is there ever a situation where I'll see one `Cat` where I expect to see the other (see `ModelUtils.getSimpleRef`)?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
jimschubert: @spacether I don't think this will "turn on interface models"… it's just returning a model that has no properties, which is a perfectly valid use case. We should have some way to indicate that the schema is free form, but we currently only store that boolean on response, parameter, and property codegen types. 

One question I would have is whether we have sufficient test coverage on this code path, because only a single typescript file has changed. For example, does `ModelUtils.isFreeFormObject` also have to be updated with logic around validation and descendants?

spacether: The reason that I think that this will turn on interface models everywhere is that this is [how python-experimental turned it on](https://github.com/OpenAPITools/openapi-generator/commit/35d42728d9df60da65300fe7cd900e4056719969) and this change applies to all generators that use ModelUtils.unaliasSchema.
If you look [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java#L463)
you see that when generating models, for a free form model, we attempt to unalias a ref to it and:
- if we follow the reference and return the schema that we ref point to, then we skip generating the model
- if we return the ref schema, we DO generate the model
This code makes it so we return the ref schema, so the interface model will be generated.
One way to confirm this would be to update the v2 and v3 petstore sample to include an interface model and see if it is generated.

In my opinion these interface models are free form models. That's why I included the comments:
```
                    // there are 3 free form use cases
                    // 1. free form with no validation that is not allOf included in any composed schemas
                    // 2. free form with validation
                    // 3. free form that is allOf included in any composed schemas
                    //      this use case arises when using interface schemas
                    // generators may choose to make models for use case 2 + 3
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: Then how about having hasDescendants call getAllOfDescendants under the hood? The code is very similar to what we already have.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: What are your suggested updates to ModelUtils.isFreeFormObject @jimschubert?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
Zomzog: I've created `hasDescendants` because `getAllOfDescendants` is a method of DefaultCodegen that call other methods of the DefaultCodegen like `toModelName` and I only need to know if it's used once.

But maybe I can move everything in DefaultCodegen.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
Zomzog: I've inlined the method. I don't manage to create such a case where I have a conflict where I have the wrong model because all entries came from `getSchemas`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
Zomzog: I'm not sure to understand but the case is `AbstractPet` must be both (in Java) a Class and a Map<String,String> (FreeForm)?
For the test coverage, I can add `composed-schemas` on some other generators if it's required.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
spacether: This is fine as is. As you mention we only need to test presence, not qty.

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/ModelUtils.java
jimschubert: @spacether I think you're mixing the concepts of "free form object" and interfaces. These aren't exactly the same things. An interface is something that's used as a base for other instances, while a "free form object" is one which is defined as having an unbounded set of allowed properties. The current implementation before this pull request doesn't allow marker interfaces or free form objects in composed models, although these are supported by the specification. If python has issues with supporting the spec, I think the changes will be needed in python so as not to block spec support in everything else.

> What are your suggested updates to ModelUtils.isFreeFormObject

If we're going to return the schema for the two new cases in this method (for block labeled as "free form objects"), then I think we should have the same two conditionals in the `ModelUtils.isFreeFormObject` method. To put it differently, this method is now:

```
if (ref.getProperties() != null && !ref.getProperties().isEmpty()) { // has at least one property
    return schema;
} else {
    // free form object (type: object)
    if (hasValidation(ref)) {
        return schema;
    } else if (hasDescendants(simpleRef, openAPI)) {
        return schema;
    }
    return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())),
            importMappings);
}
```

if the else block is still correct that we have free form objects in three different states now, then I think the helper method needs to be updated.

If, instead, the logic is meant to be this:

```
// return schema when it has at least one property, validations, or descendants
if (ref.getProperties() != null && !ref.getProperties().isEmpty()) {
    return schema;
}
if (hasValidation(ref)) {
    return schema;
} 
if (hasDescendants(simpleRef, openAPI)) {
    return schema;
}

// free form object (type: object)
return unaliasSchema(openAPI, allSchemas.get(getSimpleRef(schema.get$ref())),
        importMappings);
```

… then I think comments and structure might have to be updated in this block for clarity and `ModelUtils.isFreeFormObject` still makes sense as-is. Again, I question the amount of coverage we have on these scenarios because this PR hasn't changed any tests or samples which would indicate it's "turned on interface models".

spacether: > If python has issues with supporting the spec, I think the changes will be needed in python so as not to block spec support in everything else.
Python-experimental supports all of these use cases and is not blocked.

I have lots of questions here. Are you free to discuss it in Slack?

1. How you would classify these use cases, InterfaceModel, ModelWithValidations
InterfaceModel:
  type: object
ModelWithValidations:
  type: object
  maxProperties: 3
ComposedSchema:
  allOf
"$ref": "#/components/schemas/ErrorModel"
2. Should ModelUtils.isMapSchema be true for these?
3. Can a free form object be used as an interface?
4. Doesn't a schema type:object that is allOf included in other schemas constrain the types of properties to be AnyType because additionalProperties={} by default? That is more constraining than an empty class definition (interface). These languages have many types not covered by the openapi spec.
5. Must those other instances that inherit from the interface be type object or can they be primitives like string with validations? If so then  aren't we talking about AnyType (type=null)?

Zomzog: Does the discussion has occurred? :)

pchesneau:
Hello ! 
Any update on this PR ? Or is there anything we (non maintainers) can do to help? I just faced the same issue and wonder if I should wait for an update of openapi generator, or fork and build patched version?

BennyKitchell:
Following up, I am facing the same issue for my project.

spacether:
Sorry that this feel through the cracks. @Zomzog can you rebase this on master and resolve the merge conflict?

harmvanderwal:
Are there any updates on this ticket? Would be great if this'd be implemented in a future release.

BennyKitchell:
Bump on this @Zomzog, is there anything we can do in the community to help get this merged? Thanks!

TarasLuchak (approved):

PuskyWasTaken:
I am facing the same issue for my project. Is this ever going to get merged?

adamperea:
Will this change (or something equivalent) ever be merged @wing328? 🥺

gianlucagallo:
@wing328 plan about merging?

ldeluigi:
I need this one merged. Maybe someone can adopt the task? @Zomzog, do you have any insight on the situation?

PR #7709, opened 2020-10-13
python-experimental adds dynamic base classes [DRAFT]

spacether:
## python-experimental adds dynamic base classes

In this PR:
- when deserializing payloads or creating Schema/Model instances the new instance will dynamically inherit from the correct Schema classes (models) and python primitive classes (dict, str, list, int, float, etc)

So an object model Box will inherit from Box, Schema, and dict.
Caviats:
- None, True, and False are singletons in python so we can't inherit from them
  - For those cases, we return an enum option with the below NAME and value
    -  NONE = None
    - TRUE = True
    - FALSE = False
This allows our None/True/False use cases to still identify themselves as a Model/Schema and contain the expected values

For a composed schema Triangle that oneOf includes ScaleneTriangle, the resulting instance will inherit from Triangle + ScaleneTriangle

## Long Term Benefits of Using Dynamic Base Classes
- any depth of inline models can be supported
- model definition is shorter
- This supports complicated use cases like allOf -OneOfComposedSchema. OneOfComposedSchema oneOf: -schemaA -schemaB
- We can reduce the number of times that validations and type conversion are done
- when one has an instance of a composed schema, one will know what schemas passed validation for that payload

### PR checklist
 

spacether:
I am closing this PR because a rebase was too painful on master

PR #7746, opened 2020-10-18
Sanitize additionalProperties with CliOptions

### LINKED ISSUE #6591: [BUG] spring generator using maven plugin generates correct code, while cli/gradle generates wrong code
#### Bug Report Checklist

##### Description

I have a simple sample spec. When invoking the generator from maven plugin, code is generated as expected. But both gradle plugin and cli generate a different code.

on the generated interface `com.example.api.DefaultApi`, when running from Gradle/CLI, this code is generated:

```java
    ResponseEntity<InlineResponse200> getGreetingsMessage(ServerWebExchange exchange);
```
Which is wrong as it is using reactive classes, which are disabled on the config file.

When using maven plugin, the following code is generated instead:

```java
    ResponseEntity<InlineResponse200> getGreetingsMessage();
```

which seems to be the correct code.

4.3.1

```yaml
openapi: '3.0.3'
info:
  title: 'Greetings API'
  version: '1.0.0'
paths:
  /api/v1/greetings:
    get:
      operationId: getGreetingsMessage
      responses:
        200:
          description: 'A greetings message'
          content:
            application/json:
              schema:
                type: object
                properties:
                  message:
                    type: string
```

```sh
openapi-generator generate \
    --input-spec spec.yaml \
    --config config.json \
    --generator-name spring \
    --output . \
    --invoker-package com.example \
    --api-package com.example.api \
    --model-package com.example.model
```

##### Config file used for generation

```json
{
  "useTags": "true",
  "interfaceOnly": "true",
  "hideGenerationTimestamp": "true",
  "sourceFolder": "src/gen/java",
  "java8": "true",
  "library": "spring-boot",
  "dateLibrary": "java8",
  "reactive": "false",
  "useOptional": "true",
  "useBeanValidation": "true",
  "skipDefaultInterface": "true"
}
```

##### Steps to reproduce

copy both `spec.yaml` and `config.json` to a folder and run `open-api-generator-cli` as described above.

##### Related issues/PRs

No related issues I could find.

##### Suggest a fix

-

Zomzog:
fix #6591

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
@@ -42,13 +42,14 @@ import java.io.File;
 import java.io.IOException;
 import java.util.*;
 
+import static io.swagger.v3.parser.util.SchemaTypeUtil.BOOLEAN_TYPE;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
 /**
  * A class which manages the contextual configuration for code generation.
  * This includes configuring the generator, templating, and the workflow which orchestrates these.
- *
+ * <p>
  * This helper also enables the deserialization of {@link GeneratorSettings} via application-specific Jackson JSON usage
  * (see {@link DynamicSettings}.
  */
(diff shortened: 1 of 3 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java
```diff
@@ -615,6 +620,26 @@ public ClientOptInput toClientOptInput() {
                 .config(config)
                 .userDefinedTemplates(userDefinedTemplates);
 
-        return input.openAPI((OpenAPI)context.getSpecDocument());
+        return input.openAPI((OpenAPI) context.getSpecDocument());
+    }
+
+    private Object sanitizeAdditionalPropertyValue(CodegenConfig config, String opt, Object value) {
```
jimschubert, line 626: This isn't sanitization, but coercion from string properties to those declared in CLI options.

I love the idea behind this, but I have quite a few reservations.

I think this is potentially going to become very confusing for folks because `additionalProperties` is intended for users to be able to pass extra properties to templates. We just happen to also use it for generator's options to be passed to templates as well. What happens for a user who wants to pass a custom property like `isSupported: "false"`? We now have some properties as boolean and some as string.

It's generally up to generators themselves to determine whether or not a boolean value should be type bool in the template or type string, and this change could potentially break templates which expect string true/false in templates rather than bools. There are about 65 places where this might be affected in code, not counting built-in templates. We may also potentially be breaking some languages where booleans are cased differently (like Python), or even support them being worded differently (`yes` and `no` in YAML). In those cases, any sort of check like `if ("false".equals(property))` would now fail, and we'd have to be certain our testing covers all of those permutations of options.

I'd much rather fix this specific instance of `reactive: false` at the source of the problem in SpringCodegen with this code:

```
        if (additionalProperties.containsKey(REACTIVE)) {
            if (!SPRING_BOOT.equals(library)) {
                throw new IllegalArgumentException("Currently, reactive option is only supported with Spring-boot");
            }
            convertPropertyToBooleanAndWriteBack(REACTIVE);
            this.setReactive((boolean)additionalProperties.get(REACTIVE));
        }
```

and figure out a way separately for generators to opt-in to automatically coercing CliOpts to their defined types in a way which is appropriate for the target languages.

I love the idea you've used, and maybe we could move it to DefaultCodegen with a `coerceAdditionalProperties` flag which defaults to false, but when set to true invokes this logic? I guess that would be another potential change I'd feel comfortable with if we set the spring generator to true here. But, I think that would also mean cleaning up all the `convertPropertyToBooleanAndWriteBack` usages in the Spring generator as well.

Zomzog: Ok, I think I understand the idea, so I move the check to a method in SpringCodegen and AbstractJavaCodegen. The impact on samples is due to the fact I set default values on additionalProperties so now everything is java8 by default.

The idea is that each generator provides his list of properties to check/convert to boolean so custom properties will stay has to String if needed. 

Let me know if it's an acceptable solution for you.

Zomzog: @jimschubert I've reworked the solution when I understand that default values can be modified. I know there is a test fail on CI with java8 and maybe CliOption must be reworked to use the "defaultValue" method. But I hope you can confirm I'm on the right way before pushing further.

Issue #7530, opened 2020-09-28
[BUG][PHP] Too verbose PHPDoc blocks in each file


dkarlovi:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

## Description

Current,y the `php` generator produces files with [very large headers, close to or exceeding 40 lines](https://github.com/OpenAPITools/openapi-generator/blob/8b393358591a025f9789401bfdc6d8fc8e1fb69e/samples/client/petstore/php/OpenAPIClient-php/lib/ApiException.php#L2-L40). Not only is that way too verbose, i[t actually lowers the quality of the code](https://localheinz.com/blog/2018/05/06/cost-and-value-of-docblocks/) since the signal/noise ratio is way off.

This can also be seen [in other places](https://github.com/OpenAPITools/openapi-generator/blob/8b393358591a025f9789401bfdc6d8fc8e1fb69e/samples/client/petstore/php/OpenAPIClient-php/test/Model/DogTest.php#L82) where the information is repeated in the PHPDoc block and immediately following it in code, bringing no value.

## Solution proposals

- merge information from multiple file PHPDoc blocks into a single shorter one, not repeating itself, take into account [code style fixers can change/remove them](https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:header_comment) and design accordingly (for example, one file level one with the the data currently in `partial_header.mustache`, one class level one, example below)
- remove PHPDoc comments which just repeat the informaton already obvious from the method APIs: if you have a PHPDoc, it should create value, if it's just repeating the same information, it's not doint that

## Example

```php
<?php

/**
 * Reddit.com.
 *
 * Reddit.com API
 *
 * The version of the OpenAPI document: 1.0.0
 *
 * Generated by: https://openapi-generator.tech
 * OpenAPI Generator version: 5.0.0-SNAPSHOT
 */

[...]

PR #7779, opened 2020-10-21
Tweaks for PHP client Markdown files suggested by markdownlint

dkarlovi:
# Summary

This is a set of tweaks to the generated documentation of the `php` generator, based on `markdownlint` and other similar tools packaged in my https://github.com/dkarlovi/docker-docqa Docker image.

The main README was slightly rearranged to make it more readable, moving the parts a package user would actually be interested in to the top.

See an example here https://github.com/flexolabs/reddit-client

### PR checklist
 
@jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) heart, @ybelenko (2018/07), @renepardon (2018/12)


dkarlovi:
Failures unrelated, the build process seems flake-prone.

thomaspicquet (approved):

modules/openapi-generator/src/main/resources/php/README.mustache
```diff
@@ -4,29 +4,19 @@
 {{{appDescriptionWithNewLines}}}
 {{/appDescriptionWithNewLines}}
 
-This PHP package is automatically generated by the [OpenAPI Generator](https://openapi-generator.tech) project:
```
wing328, line 7: Personally, I prefer putting this block on top for growth hacking but I'm ok to give your suggestion a try to move it to the bottom of the page.

dkarlovi: @wing328 as discussed in #7530, the "growth hacking" effort is possible without it being overwhelming.

 From the POV of the:
1. **generator developer**, the generator is the most important thing, the code generated is a side-effect, just test files.
2. **generator user**, the point of the generator is to generate the best possible package, not promote the generator itself, perpetuum mobile style. :)
3. **package user** (the developer using the generated code), the generator is totally irrelevant, it might as well not exist, what's important is the code offered in the package, how to use it and get stuff done.

POV in 1. and 3. are in total opposition, this change is trying to shift the focus a bit more toward 3. to better realize 2. The goals of 1. should still be realized, but in a more subtle way.

dkarlovi: BTW this same issue is visible [here](https://github.com/flexolabs/reddit-client/blob/25feb73a9df32154d14598c298726c0cb3d017bb/composer.json#L13): this URL is **not** the homepage of the package built, it's either the [`info.contact.url` provided in the spec](https://github.com/flexolabs/openapi-specs/blob/4718374fd1df2cbcef94529e6e5c01a02bb31989/reddit/v1/index.yaml#L9) or something the user provides.

wing328: > the generator is totally irrelevant, it might as well not exist, what's important is the code offered in the package, how to use it and get stuff done.

Many users still not aware that the API clients can be auto-generated instead of manually written. It doesn't hurt to make them aware there's such thing called OpenAPI Generator that they can use to generate code.

dkarlovi: @wing328 I understand that as do I understand your desire to promote the generator, it's perfectly understandable and not unreasonable.

My intention here is to point out **generator's users have no inherent interest to promote the generator**, they're not its core developer team. What they want from the generator is generating high-quality code which they can offer as "SDKs" for their API's users (or use themselves). The generator for group 2. is a tool, a means to an end, it's not the destination.

The group 3. doesn't care about the generator at all, they're primarily interested to use the code generated, they're twice removed from it, for them, the code generated is the product. The code and the docs generated should reflect that better, aligning them a bit more toward the interest of users of the code generated goes a long way.

It's not about "informing the users of the package about the generator which made the package", it's about the package itself. The meta-documentation obviously has its place, but it's shouldn't be the primary focus.

dkarlovi: Think about it like this: if this was a printing press, it would be trying to add a chapter about itself to each book it's printing because it's trying to inform the readers about itself. :) Yes, it's in the printing press's best interest to put the word out, but there's more than one way to do it, being too direct can seem obnoxious and have the opposite effect from the desired one.

wing328: I never said the promotional message should be the focus. I'm not against generating a better package. Putting the message at the top of the package doesn't mean we (openapi-generator contributors) are not focusing on the quality of the package.

dkarlovi:
@wing328 what about these changes?

Issue #7781, opened 2020-10-21
[REQ] Change Python models to not raise exception when unset attribute is accessed


keymone:
Python models raise exception when accessing an attribute that wasn't set. This leads to very defensive coding style that relies heavily on exceptions for control-flow, which is generally accepted to be bad practice. Until recently it wasn't possible to even check if an attribute is set without accessing internal fields of the model, leading to code like this:

    # intended code:
    if model.attr:
        do_things()

    # actual code:
    try:
        attr = model.attr
    except ApiAttributeError:
        attr = None
    if attr:
        do_things()

After #7637 was merged things improved:

    # better code:
    attr = if 'attr' in model model.attr else None
    if attr:
        do_things() 

However this is still cumbersome compared to the intended code.

I'd like to request that accessing unset attribute no longer raises an exception. There are two potential problems we need to consider:

1. What if one needs to differentiate between attribute being unset and attribute being set to actual value `None`?

This is possible thanks to #7637: `'attr' in model and model.attr is None`.

2. Backward compatibility: if some projects rely on exception for control-flow we will break their functionality.

We should hide this behavior behind a flag to code generator.

One argument I've heard is that functionality I need is already available when accessing attributes via internal `_data_store` field, but to me this is leaking implementation details. In future we may want to change underlying implementation of model's data storage (convert models to Python 3's dataclasses, etc) which will break existing code.

Would such feature be merged if I prepared a PR for it?

cc/ @spacether

spacether:
Such a feature would be merged. I have a very strong preference that it be implemented with a get method like dictionaries do

spacether:
Feature has been merged

PR #7784, opened 2020-10-21
[python] Add option to return None instead of raising exception when accessing unset attribute

keymone:
See #7781

Add option to return None instead of raising exception when accessing unset attribute.

### PR checklist
 
cc @spacether

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
@@ -85,3 +85,6 @@
                     return True
 
         return False
+
+    __setattr__ = __setitem__
+    __getattr__ = {{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}
--- a/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
+++ b/modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
@@ -1,5 +1,5 @@
     def __setitem__(self, name, value):
-        """this allows us to set a value with instance.field_name = val"""
+        """set the value of an attribute using square-bracket notation: `instance[attr] = val`"""
         if name in self.required_properties:
             self.__dict__[name] = value
             return
@@ -29,7 +29,7 @@
     __unset_attribute_value__ = object()
 
     def get(self, name, default=None):
-        """this allows us to get a value with val = instance.field_name"""
+        """returns the value of an attribute or some default value if the attribute was not set"""
         if name in self.required_properties:
             return self.__dict__[name]
 
@@ -61,8 +61,9 @@
             )
 
     def __getitem__(self, name):
-        value = self.get(name, self.__unset_attribute_value__)
-        if value is self.__unset_attribute_value__:
+        """get the value of an attribute using square-bracket notation: `instance[attr]`"""
+        value = self.get(name, self.__dict__['__unset_attribute_value__'])
+        if value is self.__dict__['__unset_attribute_value__']:
             raise ApiAttributeError(
                 "{0} has no attribute '{1}'".format(
                     type(self).__name__, name),
@@ -71,7 +72,7 @@
         return value
 
     def __contains__(self, name):
-        """this allows us to use `in` operator: `'attr' in instance`"""
+        """used by `in` operator to check if an attrbute value was set in an instance: `'attr' in instance`"""
 
         if name in self.required_properties:
             return name in self.__dict__
@@ -85,6 +86,3 @@
                     return True
 
         return False
-
-    __setattr__ = __setitem__
-    __getattr__ = {{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}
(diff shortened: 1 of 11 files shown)


spacether:
Can this be implemented with a get method? That way we do not need to add a new generator additionalProperty or configuration property.
This would behave like dict.get
So one can have:
```
pet_kwargs = {"name": "Hannah", "color": "black"}
pet = Pet(**pet_kwargs)
animal_type = pet.get("animal_type") # None is the default_value automatically
# animal_type == None
animal_type = pet.get("animal_type", "dog")
# animal_type == "dog"
```

keymone:
@spacether I understand the motivation behind `.get()` approach and we can have that as well without any conflict with this PR.

But why not have both? Compared to current implementation both `.get` and not throwing exceptions are preferable and lead to better code quality.

Yes, model instances can be treated as dictionaries and dictionaries throw a `KeyError` when non-existant key is being retrieved, however I don't think retrieving non-existent key is the same as retrieving existing but unset attribute. So maybe I should add throwing exception when non-existant attribute is being retrieved?

Also, dictionaries don't pretend to be objects and don't expose direct `model.attr` interface. Why have this interface, which is appealing for it's conciseness and simplicity but burden users with having to deal with exceptions or explicitly verify presence of an attribute all the time? Let's have an option for a more relaxed interface.

spacether:
> Yes, model instances can be treated as dictionaries and dictionaries throw a KeyError when non-existant key is being retrieved, however I don't think retrieving non-existent key is the same as retrieving existing but unset attribute. So maybe I should add throwing exception when non-existant attribute is being retrieved?

This gets into very murky territory
We either should raise key errors or we shouldn't. We shouldn't have case where we don't raise errors all the time then have a corner case where we always do raise them. That's why I think that providing an interface that raises an exception (dot access) clearly describes what is happening and adding get, that allows you to control the default provides clear predicatable behavior here.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
```diff
@@ -1,4 +1,4 @@
-    def __setattr__(self, name, value):
+    def __setitem__(self, name, value):
         """this allows us to set a value with instance.field_name = val"""
```
spacether, line 2: How about updating the docstring here? 
```
we use this to set values for instance.field_name = val and instance["field_name"] = val
```

samples/openapi3/client/extensions/x-auth-id-alias/python-experimental/x_auth_id_alias/model_utils.py
```diff
@@ -346,30 +341,30 @@ class ModelNormal(OpenApiModel):
...
 
-        path_to_item = []
-        if self._path_to_item:
-            path_to_item.extend(self._path_to_item)
-        path_to_item.append(name)
         raise ApiAttributeError(
             "{0} has no attribute '{1}'".format(
                 type(self).__name__, name),
-            [name]
+            [e for e in [self._path_to_item, name] if e]
```
spacether, line 367: How about just using name here? We know that name was not found.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
```diff
@@ -20,28 +20,22 @@
...
 
-        path_to_item = []
-        if self._path_to_item:
-            path_to_item.extend(self._path_to_item)
-        path_to_item.append(name)
         raise ApiAttributeError(
             "{0} has no attribute '{1}'".format(
                 type(self).__name__, name),
-            path_to_item
+            [e for e in [self._path_to_item, name] if e]
```
spacether, line 26: How about just using name here? We know that name was not found.
- this same comment applies on line 70 too

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
```diff
@@ -20,28 +20,22 @@
...
         )
 
-    def __getattr__(self, name):
+    __unset_attribute_value__ = object()
```
spacether, line 29: How about __unset_attribute_value__ = object
Then we always point to a singleton and don't need to instantiate additional empty objects

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
keymone: i just compacted the previous functionality to find `path_to_item`, if you think just `name` is enough instead of the old path code, i'm fine with that

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/methods_setattr_getattr_composed.mustache
keymone: well, somebody *could* concievably set an attribute to `object` value, `__unset_attribute_value__` will generate unique object only once per class, so i'm not worried about performance here

spacether:
This looks great, thanks you!

Can you add a python test in one of our [manual tests](https://github.com/OpenAPITools/openapi-generator/tree/master/samples/openapi3/client/petstore/python-experimental/tests_manual) that accesses a type:object's model properties with:
- when a property is stored
  - access instance.prop and assert that the result is what you expect
  - access instance["prop"] and assert that the result is what you expect
  - access instance.get("prop") and assert that the result is what you expect
- when a property is not stored
  - access instance.prop and assert that the result is what you expect or the expected exception is raised
  - access instance["prop"] and assert that the result is what you expect or the expected exception is raised
  - access instance.get("prop") and assert that the result is what you expect or the expected exception is raised
  - access instance.get("prop", None) and assert that the result is None

That will verify that this is working and throw a test error if this feature is accidentally changed/broken in the future.

keymone:
@spacether i think all those cases are covered except `.get` with existing and absent attribute value. i've added those but test broke for unrelated reason.

spacether (approved):
Thanks for adding those tests. Your CI error is unrelated so I will merge this.
Thank you for the your PR!

PR #7813, opened 2020-10-26
[Protobuf] Fix support for AllOf in protobuf-schema

guidoussin:
### PR checklist
 
### Description

This fix is in 2 steps:

#### Fixed index for vars in generated protobuf files
To ensure the backward compatibility between two schema generations, I'm fixing the index generated for each attribute base on its name.

#### Correct support of AllOf
As there is no inheritance in protobuf, the use case _3_0/allOf_composition_discriminator.yaml_ doesn't work.
If the parent is referenced in the response, we can't use one of its children as instantiation (Dog for Pet for example).
The only way to make it work is to replicate all attributes from the children in the parents, including imports.

Before the fix, the generated Pet was:
```
syntax = "proto3";

package somepackage;

message Pet {

  string petType = 0;

}
```

After the fix:
```
syntax = "proto3";

package somepackage;

import public "models/lizard_all_of.proto";
import public "models/snake_all_of.proto";
import public "models/cat_all_of.proto";
import public "models/dog_all_of.proto";

message Pet {

  string petType = 140636936;

  bool hasLegs = 159828448;

  bool lovesRocks = 499337491;

  string name = 3373707;

  string bark = 3016376;

}
```


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ProtobufSchemaCodegen.java
```diff
@@ -37,12 +35,17 @@
...
 import java.util.regex.Pattern;
 
 import static org.openapitools.codegen.utils.StringUtils.camelize;
 import static org.openapitools.codegen.utils.StringUtils.underscore;
 
 public class ProtobufSchemaCodegen extends DefaultCodegen implements CodegenConfig {
+  
+    private static final String IMPORT = "import";
+
+  private static final String IMPORTS = "imports";
```
wing328, line 48: We'll change the indention to 4-spaces in another PR.

wing328: Done via https://github.com/OpenAPITools/openapi-generator/pull/7857

PR #7840, opened 2020-10-29
[rust-server] Add support for bearer token authentication

mattdowdell:
As far as I can tell, the bulk of the work to support bearer tokens was already implemented to support OAuth, but never explicitly turned on. This was unexpectedly simple, so I susect I may have missed something.

Changes:
- Added bearer token to the security features for rust server.
- Supplemented the basic auth condition in the context template to handled basic auth and bearer token separately.
- Repurpose an exising sample to confirm the code generation works as expected.

The new sample builds without problems, so the only possibly issue is that the underlying `swagger` crate didn't implement this. However, there's already bearer support in the client and it's also used in the OAuth support in the server, so I don't see why it wouldn't work.

### PR checklist
 
- [X] Run the shell script `./bin/generate-samples.sh`to update all Petstore samples related to your fix. This is important, as CI jobs will verify _all_ generator outputs of your HEAD commit as it would merge with master. These must match the expectations made by your contribution. You may regenerate an individual generator by passing the relevant config(s) as an argument to the script, for example `./bin/generate-samples.sh bin/configs/java*`. For Windows users, please run the script in [Git BASH](https://gitforwindows.org/).

CC @frol @farcaller @richardwhiuk @paladinzh


richardwhiuk:
Nope looks good to me. I think we've just not had a requirement to support Bearers outside of OAuth before, so it's just not come up.

Thanks!

richardwhiuk:
(Client support is already present - https://github.com/OpenAPITools/openapi-generator/blob/4ea4ceb0ff13077100448ec93d694d9d43166a9b/modules/openapi-generator/src/main/resources/rust-server/client-operation.mustache#L324-L335 )

PR #7858, opened 2020-11-02
[python-experimental] Add mypy for the client

meadsteve:
Partially implements: https://github.com/OpenAPITools/openapi-generator/issues/7836

Adds mypy static type checker with a minimal setup running on the experimental python client code.

I've made a small amount of changes to get the build passing but there's plenty of follow up that could be added to make the typing richer.

### PR checklist
 
/cc @wing328 @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @arun-nalla (2019/11) @spacether (2019/11)


modules/openapi-generator/src/main/resources/python/python-experimental/tox.mustache
```diff
@@ -7,3 +7,4 @@ deps=-r{toxinidir}/requirements.txt
 
 commands=
    {{^useNose}}pytest --cov={{{packageName}}}{{/useNose}}{{#useNose}}nosetests{{/useNose}}
+   mypy petstore_api
```
meadsteve, line 10: This is the core part of the change

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
```diff
@@ -62,7 +62,7 @@
 
 {{#additionalPropertiesType}}
     @cached_property
-    def additional_properties_type():
+    def additional_properties_type(cls):
```
spacether, line 65: Why did you add cls as an input here? The cls input is not used.
- this same question applies to openapi_types + discriminator

meadsteve: @spacether all methods on a class technically require one argument. Which is self for instances of the object or the cls for a class method. The only exception is if it's decorated with `@staticmethod`. Although it "works" without this first argument mypy detects it (correctly) as an error.

Now that I look at this again for these it really should be called self not cls. This won't affect the functionality but I'll make the change anyway as it's confusing for it to have the wrong name.

meadsteve: Actually ignore me. I misunderstood your question. I think this is a little more complicated to do correctly because of the way the @cached_property decorator works.

spacether:
Can you rebase on master branch? Not sure why those drone errors are happening.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
meadsteve: @spacether So I've updated this to self so that `@ cached_property ` can be swapped for `@property` without any code changes.

meadsteve:
@spacether just rebased and pushed. Let's see what happens.

spacether:
Hey there. Sorry to see you close this. I wwould likeit merged if wecan get tests passing. How do you feel? I apologize that this has taken so long.

meadsteve:
@spacether I was clearing out old PRs. Let me rebase and see if it passes now as I don't think the problems related to the PR. I've lost a lot of mental context of this over christmas though.

meadsteve:
I'll need to find the time to fix the conflicts though as I think things have moved on a bit since I opened this.

spacether:
Yup python-experimental became python while this PR was open

meadsteve:
@spacether I think this might be a lot of work to fix. I checked out the latest master and it looks like the the sample client in `samples/openapi3/client/petstore/python` hasn't been updated. So when I update that (without applying any changes) then I get a bunch of errors. Possibly this means master is broken which needs fixing first.

Apologies if this is incorrect. I'm just getting back up to speed with this.

spacether:
What do you mean by: `the the sample client in samples/openapi3/client/petstore/python hasn't been updated`?
Every time a PR is merged those files must be up to date. Otherwise CI tests won't pass.
Your PR may be easier to re-write from scratch because the python generators changed while yours was open.

modules/openapi-generator/src/main/resources/python/python-experimental/model_templates/classvars.mustache
spacether: If we could avoid this error from being triggered in mypy that would be ideal. Without a cls or self input arg this code runs fine. Doesn't that make it valid python?

meadsteve:
@spacether yeah I think you might be right. I'll close this in favour of another PR. There could well be some state I need to clear out.

PR #7862, opened 2020-11-02
Allow metadata file generation to be disabled using additionalProperties.

jrouly:
When `generateMetadataFiles = false` is passed in `additionalParameters`, do not generate `.openapi-generator-ignore` or `.openapi-generator/{FILE,VERSION}`.

* Moves the boolean for this into `CodegenConfig` so that custom implementations can set this now.
* Removes the mutable variable from `DefaultGenerator` and instead relies on the functions in the `CodegenConfig` interface. Looking for input on this point.
* Adds a `CodegenConstant` for `generateMetadataFiles` and hooks it into `DefaultCodegen#processOpts()`
* Adds a unit test for suppressing metadata file generation.

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -84,7 +84,6 @@ public class DefaultGenerator implements Generator {
     private Boolean generateApiDocumentation = null;
     private Boolean generateModelTests = null;
     private Boolean generateModelDocumentation = null;
-    private Boolean generateMetadata = true;
     private String basePath;
     private String basePathWithoutHost;
     private String contextPath;
@@ -167,10 +166,13 @@ public class DefaultGenerator implements Generator {
      * or other metadata files used by OpenAPI Generator.
      *
      * @param generateMetadata true: enable outputs, false: disable outputs
+     * @deprecated use {@link CodegenConfig#setGenerateMetadataFiles(boolean)} instead
+     * @throws NullPointerException if the config is not already instantiated
      */
     @SuppressWarnings("WeakerAccess")
+    @Deprecated
     public void setGenerateMetadata(Boolean generateMetadata) {
-        this.generateMetadata = generateMetadata;
+        this.config.setGenerateMetadataFiles(generateMetadata);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -224,6 +224,9 @@ void configureGeneratorProperties() {
         generateApiTests = GlobalSettings.getProperty(CodegenConstants.API_TESTS) != null ? Boolean.valueOf(GlobalSettings.getProperty(CodegenConstants.API_TESTS)) : getGeneratorPropertyDefaultSwitch(CodegenConstants.API_TESTS, true);
         generateApiDocumentation = GlobalSettings.getProperty(CodegenConstants.API_DOCS) != null ? Boolean.valueOf(GlobalSettings.getProperty(CodegenConstants.API_DOCS)) : getGeneratorPropertyDefaultSwitch(CodegenConstants.API_DOCS, true);
 
+        // allows disabling the default generation of OpenAPI Generator metadata files.
+        generateMetadata = config.additionalProperties().containsKey(CodegenConstants.GENERATE_METADATA_FILES) ? Boolean.valueOf(config.additionalProperties().get(CodegenConstants.GENERATE_METADATA_FILES).toString()) : getGeneratorPropertyDefaultSwitch(CodegenConstants.GENERATE_METADATA_FILES, true);
```
jrouly, line 228: It seems like this `generateMetadata` flag was only ever being used during testing. Not sure why - it's a useful flag to expose to consumers.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -167,10 +166,13 @@ public TemplateProcessor getTemplateProcessor() {
...
      *
      * @param generateMetadata true: enable outputs, false: disable outputs
+     * @deprecated use {@link CodegenConfig#setGenerateMetadataFiles(boolean)} instead
+     * @throws NullPointerException if the config is not already instantiated
      */
     @SuppressWarnings("WeakerAccess")
+    @Deprecated
     public void setGenerateMetadata(Boolean generateMetadata) {
-        this.generateMetadata = generateMetadata;
+        this.config.setGenerateMetadataFiles(generateMetadata);
```
jrouly, line 175: Depending on reviewer preference, I could alternatively:
* default to `true` if `config == null`
* continue to maintain `generateMetadata` in `DefaultGenerator` and back out the commit that moves it into `CodegenConfig`, instead doing the `additionalProperties` check in `DefaultGenerator`
* delete this method instead of deprecating it
* something else!

jrouly:
Tagging the scala technical committee @clasnake @jimschubert @shijinkui @ramzimaalej @chameleon82 @Bouillie

Y'all may be interested in this PR because it allows for the use of `sbt-openapi-generator` in sbt `sourceGenerators`. Here's an example from some production code I'm working on:
```scala
lazy val models = project
  .settings(commonSettings)
  .settings(
    openApiGeneratorName := "scala-basic-models",
    openApiInputSpec := ((resourceDirectory in Compile).value / "openapi.yaml").getPath,
    openApiOutputDir := ((sourceManaged in Compile).value).getPath,
    openApiPackageName := "com.rallyhealth",
    openApiValidateSpec := Some(true),
    openApiAdditionalProperties ++= Map("generateMetadataFiles" -> "false"),
    sourceGenerators in Compile += openApiGenerate
  )
```

Previously this wasn't possible because the metadata files would fail to compile.

chameleon82:
`sourceManaged` is preferable way to manage generated files in `Scala`. 
I would like to have this change in favor to `sbt-openapi-generator`.

wing328:
cc @OpenAPITools/generator-core-team as well

jimschubert:
I don't think additional properties is appropriate here. This is tool-specific, not template-specific. Exposing this as a property now means it's another thing we'd have to ensure exists for templates.

This should really be done in the plugin, via option (copying cli), and using the existing setter. I'll look at doing that this evening to show what I mean.

jrouly:
I was also on the fence about Additional Properties. What do you think
about a Global property?

jrouly:
@jimschubert I was also on the fence about additional properties for the same reason.

> via option (copying cli)

So, are you thinking this should be set in `GlobalSettings` or `CodegenConfigurator` then? Or _entirely_ within the plugin?

jrouly:
@jimschubert is this what you were thinking? https://github.com/OpenAPITools/sbt-openapi-generator/pull/23

jimschubert:
> @jimschubert is this what you were thinking? https://github.com/OpenAPITools/sbt-openapi-generator/pull/23

Yeah, that's what I was talking about.

jrouly:
If you're cool with that approach, I'm cool with it too. I'll close this PR.

PR #7916, opened 2020-11-10
[Java/okhttp] Add dynamic operations option

cbornet:
### PR checklist
 
This PR adds support for dynamic operations.
When the option is selected, the generated code will get the details for the operations from an OAS that is provided at runtime. The generated code will look for `operationId` and the parameter names in this OAS to build the requests.
This way the client is more decoupled from the server. You can make changes in the server API and the client will still work without needing to be regenerated. This is a major gain when ensuring the client is up-to-date is hard (eg. mobile apps).

Examples of API changes that keep compatibility with dynamic clients (and that break non-dynamic clients):
* Changing the operation URL (eg changing /api/user/{id} to /api/users/{id})
* Changing the operation method (eg. a POST becomes a PUT)
* Moving an operation parameter between path, header and cookie

@bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04) @lwlee2608 (2019/10) @bkabrda (2020/01)


wing328 (approved):
Given that the feature is off by default, I'm happy to include it in the upcoming 5.0.0 release.

Issue #3894, opened 2019-09-15
Springboot codegen : Compilation error in generated controllers if the skipDefaultInterface tag is set to true


devdream02:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] What's the version of OpenAPI Generator used?
- [ ] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

Springboot codegen : Compilation error in generated controllers if the skipDefaultInterface tag is set to true

##### openapi-generator version

3.0.2
##### OpenAPI declaration file content or url

petstore.yaml
##### Command line used for generation

spring boot
maven plugin config
`                             <configOptions>
                                <java8>true</java8>
                                <library>spring-boot</library>
                                <dateLibrary>java8</dateLibrary>
                                <sourceFolder>src/main/java</sourceFolder>
                                <useBeanValidation>true</useBeanValidation>
                                <configPackage>com.solvezy.sample.config</configPackage>
                                <swaggerDocketConfig>true</swaggerDocketConfig>
                                <skipDefaultInterface>true</skipDefaultInterface>
                            </configOptions>`

##### Steps to reproduce

Set skipdefaultInterface flag as true

###### Error message
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project openapi-spring: Compilation failure: Compilation failure:
[ERROR] /C:/Users/1984/Downloads/Codebase/openapipoc/src/main/java/com/solvezy/sample/api/UserApiController.java:[11,8] com.solvezy.sample.api.UserApiController is not abstract and does not override abstract me
thod updateUser(java.lang.String,@javax.validation.Valid com.solvezy.sample.model.User) in com.solvezy.sample.api.UserApi
[ERROR] /C:/Users/1984/Downloads/Codebase/openapipoc/src/main/java/com/solvezy/sample/api/UserApiController.java:[20,5] method does not override or implement a method from a supertype
[...]

Issue #2984, opened 2019-05-23
[Java][Spring] Disable default methods for interfaces


eriktim:
## The problem

For our microservices we use the Java Spring OpenAPI generator. When creating a new service:
* we write an OpenAPI yaml file;
* we generate interfaces and DTOs using `interfaceOnly` and `java8`;
* we create controllers that implement the interfaces.

This works really well. However, all interfaces have a default implementation (resulting in a `HttpStatus.NOT_IMPLEMENTED`). These will only give errors on runtime and does not tell me what controller methods I need to implement.

## The suggested solution

It would be great to be able to disable the default implementation so that we get errors at compile time showing what methods we need to implement.

## Alternatives

The alternative would be to also generate an initial controller which we modify according our needs. Unfortunately, this will not always work when you update an existing interface and we would like to keep generated/manual code separated.

Are there any plan on such a feature?

cc @diyfr

macjohnny:
@eriktim it would be great if you could implement this feature

PR #7959, opened 2020-11-17
Make the generated API controller code compile.

developer79433:
This change causes the generated API controller class to be declared 'abstract' if the (also generated) interface that it implements doesn't have any default methods, because the programmer set the Spring generator's "skipDefaultInterface" option.

Without this change, the generated code doesn't compile. With this change, it does.

Apart from the existing project tests, I tested that my POC Spring project that uses OpenAPI generator compiles when I use this branch of OpenAPI generator, which it previously doesn't.

I could have also added 'Base' to the end of the class name when it's abstract, as I discussed in the associated GitLab issue about this (#3894 ).
I didn't do this, that's a larger change, as the generated filename then needs to change too. In my opinion this would add unjustified additional complexity to OpenAPI Generator for something that's mostly just of cosmetic value. But if other people deem this additional change worthwhile, I'm happy to implement it.

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/JavaSpring/apiController.mustache
+++ b/modules/openapi-generator/src/main/resources/JavaSpring/apiController.mustache
@@ -49,7 +49,7 @@ import java.util.concurrent.Callable;
 @RequestMapping("${openapi.<%title%>.base-path:<%>defaultBasePath%>}")
 <%={{ }}=%>
 {{#operations}}
-public class {{classname}}Controller implements {{classname}} {
+public {{^jdk8-default-interface}}abstract {{/jdk8-default-interface}}class {{classname}}Controller implements {{classname}} {
 {{#isDelegate}}
 
     private final {{classname}}Delegate delegate;


macjohnny:
cc @bbdouglas (2017/07) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09) @jeff9finger (2018/01) @karismann (2019/03) @Zomzog (2019/04) @lwlee2608 (2019/10) @bkabrda (2020/01)

developer79433:
That CircleCI failure appears to be due to a TCP connection reset encountered while downloading dependencies. This is robably a transient error? I'm not sure how to trigger a retry.

In any case, I don't think that failure was caused by this change.

@macjohnny , what does 'breaking change without fallback' mean? I don't want to break anything. I don't understand what a fallback is. Is there something I can do to fix whatever is wrong here?

Thanks for looking at this, I appreciate the attention. This might seem like a minor issue, but it improves my developer experience significantly.

macjohnny:
@developer79433 I re-started the CI, but I think you will have to re-generate the samples.
it seems like a breaking change, since it changes the output in an incompatible way: the class now needs to be extended. E.g. if someone has a script using the generated class, not expecting the `abstract` keyword, this would break. But I prefer to delegate the discussion to the java technical committee.

developer79433:
Thanks for restarting the CI job.
Yep, the class does now need to be extended, which it didn't before. But previously, there was no compiled class at all, because the generated code didn't compile :-) So I think (correct me if I'm wrong) that this change won't break any Java compilations that weren't already breaking. This change might break scripts (not javac) that parsed the Java code. I wasn't aware such post-processors existed. But I'm hoping it would be easy for such scripts to begin permitting the presence of the 'abstract' keyword.

macjohnny:
@developer79433 yes I agree that such a post-processing script would be easy to adapt, nevertheless, from the current point of view, it is a breaking change, even if it is easy to fix. but since version the upcoming 5 *can* contain such breaking changes, it should be fine to include this change.

developer79433:
@macjohnny Can you give me an example of something that this change would break? Maybe it's something I could fix.
In any case, I'm glad to hear this change can be included in an upcoming version.

developer79433:
@macjohnny A little housekeeping - can you add the 'bug' label, please? And link issue #3894 , which this MR fixes?

I don't seem to be able to make these changes myself.

developer79433:
@bbdouglas Can we get your review of this change, please?

This is a trivial change that fixes an actual, real bug in OpenAPI Generator - it generates Java code code that doesn't compile, which this change fixes.

I want to demo OpenAPI Generator to a large, risk-averse organisation, but it's not currently usable to them, due solely to this bug.
I don't wish to include in my demo to this organisation what is the truth, that I had to fork OpenAPI Generator to get it to work, then maintain my fork.
I'm concerned that could create an impression that OpenAPI Generator was not mature enough for their use, or that upstream were not responsive to bugfixes, so the project must be forked to ensure timely fixes.
That could cause this organisation to conclude that use of OpenAPI Generator was not sustainable, which would be most unfortunate.

As far I understand it, this change is blocked awaiting a major release, because there are concerns that it could break unspecified Java code post-processor(s).
Such post-processors must therefore be depending on OpenAPI Generator's output Java code continuing not to compile.
I've been unable to find any such post-processors. I do not currently possess any evidence that they exist.

Although I agree that breaking third-party tools would be unfortunate, I believe that any tools that post-process Java code, but which depend on that code not compiling, are already fundamentally broken, and should be fixed forthwith.
I believe such a bug should not be worked around in OpenAPi Generator, still less at cost of delaying fixes to OpenAPI Generator.

developer79433:
@macjohnny I think you may have misunderstood the nature of this change. When you said:

> E.g. if someone has a script using the generated class, not expecting the abstract keyword, this would break.

There is nobody using the generated class, because there is no class, because the code doesn't compile.

There are two cases:
1. The generated code is usable. OpenAPi Generator is configured in such a way that this change does nothing whatsoever.
2. The generated code is unusable. OpenAPI is configured in such a way that with this change, the code would become usable.

There is no case in which any Java code compiles without this change, but does not compile with this change.

macjohnny:
@developer79433 thanks for the detailed analysis and the effort to improve the generator!
I am wondering whether it would be a suitable solution to change
https://github.com/OpenAPITools/openapi-generator/blob/ebac0a8ed42815f18bb42a493eafa430cbff0a2d/modules/openapi-generator/src/main/resources/JavaSpring/apiController.mustache#L93

to
```
{{^jdk8-default-interface}} 
```
so that implementation methods are generated whenever it is not jdk8 or the `skipDefaultInterface` option is true.


in any case, the potentially breaking change behavior is just some concern that it *might* affect users relying on the generated code not to be abstract, but I think it is fine to include your change.

@cbornet @bbdouglas could you please review this change?

@developer79433 please re-generate the samples to we can merge this change if there are no objections.

macjohnny:
maybe @wing328 also has an opinion on this

cbornet:
The `jdk8-default-interface` option was precisely introduced to make the compile fail. See https://github.com/OpenAPITools/openapi-generator/issues/2984
Personnally I don't understand why you would set this flag to false. Just set it to true and everything will work fine.

macjohnny:
@cbornet I think using `skipDefaultInterface` makes most sense in combination with `interfaceOnly`, so we get an interface without default implementations, and we manually create a class that needs to implement these methods.
However, then we might miss which controlles we need to implement.

macjohnny:
@eriktim would you like to join the discussion here?

eriktim:
Thanks @macjohnny.

`skipDefaultInterface` was indeed intended to be used with `interfaceOnly` as described in #2984. The idea is that you get compile time errors of unimplemented methods instead of HTTP errors. I do agree the generated code should be free of errors though, so making it `abstract` might fix that. I cannot judge on the impact of this as I haven't seen any use case for generating controller classes anyway so far. I always enable `interfaceOnly`. An alternative solution could be to only allow the flag to be set in conjunction with `interfaceOnly`.

macjohnny:
@developer79433 does the `interfaceOnly` option resolve the issue in your project?

Issue #8047, opened 2020-11-30
[BUG] [ruby] port `:-1` generated in configuration.rb


jfeltesse-mdsol:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [x] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

In the case _no port_ is defined in the server's URL, in the generated `configuration.rb` file the `@host` variable would feature `:-1` as the port, which is of course invalid in a URL.

##### openapi-generator version

master @ cb1a620628

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.3
info:
  title: test api
  version: v1
servers:
  - url: https://hello-{stage}.world.{tld}{basePath}
    variables:
      stage:
        default: sandbox
      tld:
        default: net
      basePath:
        default: /v1
paths:
  /test:
    get:
      description: this is a test
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                type: string
```

##### Generation Details

```sh
openapi-generator generate -i port-openapi.yml -g ruby --library faraday -o ./port-client
```

##### Steps to reproduce

1. run the above command with the above openapi.yaml file
2. `cat port-client/lib/openapi_client/configuration.rb | grep -n hello-sandbox`
3. prints out: `137:      @host = 'hello-sandbox.world.net:-1'`

##### Related issues/PRs

- https://github.com/OpenAPITools/openapi-generator/pull/7464
- https://github.com/OpenAPITools/openapi-generator/pull/7531

@cliffano (2017/07) @zlx (2017/09) @autopp (2019/02) @wing328

wing328:
@jfeltesse-mdsol thanks for the details. I'll take a look later today.

wing328:
UPDATE: filed #8049 to fix it. Local tests using the spec you provided look good.

PR #8049, opened 2020-11-30
Fix port when the URL doesn't have port specified

wing328:
To fix https://github.com/OpenAPITools/openapi-generator/issues/8047

TODO: will add a test case shortly

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -769,7 +769,7 @@ public class DefaultGenerator implements Generator {
         bundle.put("basePathWithoutHost", basePathWithoutHost);
         bundle.put("scheme", URLPathUtils.getScheme(url, config));
         bundle.put("host", url.getHost());
-        if (url.getPort() != 80 && url.getPort() != 443 ) {
+        if (url.getPort() != 80 && url.getPort() != 443 && url.getPort() != -1) {
             bundle.put("port", url.getPort());
         }
         bundle.put("contextPath", contextPath);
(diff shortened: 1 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -769,7 +769,7 @@ private void generateSupportingFiles(List<File> files, Map<String, Object> bundl
         bundle.put("basePathWithoutHost", basePathWithoutHost);
         bundle.put("scheme", URLPathUtils.getScheme(url, config));
         bundle.put("host", url.getHost());
-        if (url.getPort() != 80 && url.getPort() != 443 ) {
+        if (url.getPort() != 80 && url.getPort() != 443 && url.getPort() != -1) {
```
jfeltesse-mdsol, line 772: I guess that does the trick but why was the port `-1` to begin with? Some default value? Unfortunate conversion or something?
If there's no way to make it so it doesn't get the `-1` value I suppose a comment explaining why `-1` is a possible value and needs to be handled would help in the future. 
Port 80 and 443 people can figure it out, -1, not so much I guess.

wing328: My guess is that it's a "special logic/convention" in swagger parser but don't have time to dig into that code base to confirm.

wing328: > The getPort() function is a part of URL class. The function getPort() returns the port of a specified URL. The function returns the port number or -1 if the port is not set

Ref: https://www.geeksforgeeks.org/url-getport-method-in-java-with-examples/

jfeltesse-mdsol (approved):

PR #8054, opened 2020-11-30
Fixes issue 8052: Stackoverflow in toExampleValue() for python client

fbl100:
Added a Set<String> to the toExampleValueRecursive() method to detect circular references.
Added a watered down GeoJson API that, prior to this commit, would have thrown a StackOverflowException
Added a unit test (testRecursiveToExample()) and expected output confirming that the fix works.

@spacether this solution may be a little heavy handed.  I tried not to introduce another parameter to the toExampleValueRecursive(), but this seemed like the cleanest way to do it.  

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
@@ -16,6 +16,7 @@
 
 package org.openapitools.codegen.languages;
 
+import com.google.common.collect.Sets;
 import io.swagger.v3.core.util.Json;
 import io.swagger.v3.oas.models.media.*;
 import io.swagger.v3.oas.models.media.ArraySchema;
@@ -879,7 +880,7 @@ public class PythonClientCodegen extends PythonLegacyClientCodegen {
 
     public String toExampleValue(Schema schema, Object objExample) {
         String modelName = getModelName(schema);
-        return toExampleValueRecursive(modelName, schema, objExample, 1, "", 0);
+        return toExampleValueRecursive(modelName, schema, objExample, 1, "", 0, Sets.newHashSet());
     }
 
     private Boolean simpleStringSchema(Schema schema) {
@@ -925,9 +926,12 @@ public class PythonClientCodegen extends PythonLegacyClientCodegen {
      *                    ModelName( line 0
      *                        some_property='some_property_example' line 1
      *                    ) line 2
+     * @param seenSchemas This set contains all the schemas passed into the recursive function. It is used to check
+     *                    if a schema was already passed into the function and breaks the infinite recursive loop. The
+     *                    only schemas that are not added are ones that contain $ref != null
      * @return the string example
      */
-    private String toExampleValueRecursive(String modelName, Schema schema, Object objExample, int indentationLevel, String prefix, Integer exampleLine) {
+    private String toExampleValueRecursive(String modelName, Schema schema, Object objExample, int indentationLevel, String prefix, Integer exampleLine, Set<Schema> seenSchemas) {
         final String indentionConst = "    ";
         String currentIndentation = "";
         String closingIndentation = "";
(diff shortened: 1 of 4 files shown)


spacether:
Hey there. Thanks for your PR. Yup adding a new parameter is the cleanest way to handle it.
Can you please run 'bin/utils/ensure-up-to-date' locally and commit changes? That will fix the CI error.

spacether (changes requested):
please regenerate the samples to eliminate the CI error

samples/client/petstore/python/docs/FakeApi.md
```diff
@@ -175,55 +173,7 @@ configuration = petstore_api.Configuration(
...
-        prefix_ns_integer=-2,
-        prefix_ns_boolean=True,
-        prefix_ns_array=[
-            1,
-        ],
-        prefix_ns_wrapped_array=[
-            1,
-        ],
-    ) # XmlItem | XmlItem Body
+    xml_item = XmlItem(XmlItem) # XmlItem | XmlItem Body
```
spacether, line 176: Your update is breaking these samples.
- Client
- AnimalFarm
- XmlItem

Can you update your code so these work correctly?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
```diff
@@ -948,6 +949,18 @@ private String toExampleValueRecursive(String modelName, Schema schema, Object o
...
         }
+
+        // check for recursive model
+        if (processedModels.contains(modelName)) {
+            // have already processed this model
+            // this happens when you have recursive or circular models
+            return fullPrefix + modelName + closeChars;
+        } else if(modelName != null && !modelName.isEmpty()){
+            // if the model is non-null and non-empty, add it to the set
+            processedModels.add(modelName);
```
spacether, line 960: Shouldn't we only add a model when we have finished generating its sample?
What are our recursive base cases here for cycles?

fbl100: I'll go back and take a look at this again, thanks.  I didn't realize that the 'ran /bin/utils/ensure-up-to-date' was required and didn't pick up on the fact that my examples had changed that.

fbl100:
@spacether We took another crack at the toExampleValueRecursive() issue.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
fbl100: @spacether Got a CircleCI build error that I'm having  a hard time interpreting.  Do you mind taking a look and providing pointers for how to reproduce locally?

modules/openapi-generator/src/test/resources/3_0/issue_8052_recursive_model_expected_value.txt
```diff
@@ -0,0 +1,9 @@
+GeoJsonGeometry(
+        type="GeometryCollection",
+        geometries=[
+            GeoJsonGeometry(
+                type="GeometryCollection",
+                geometries=null,
```
spacether, line 6: null is invalid in Python. Can you update this to be None?

spacether (changes requested):
Thank you for making these changes. It is almost there!
Your tests sample includes null which is not valid python code.
Can you update the string null to be the string None?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonClientCodegen.java
spacether: That error is unrelated. We. can ignore it.

fbl100:
Hi @spacether, We have updated the example value to be None in the recursive case.

spacether (changes requested):
Thank you for making that update. So looking at your spec None is not a valid value for geometries because geometries is type array and that array is not Nullable. Please change the empty value to be empty list for your sample.

spacether:
When /i is passed in can we turn on case insensitive in RgxGen? If you don't want to do this now we can leave it to the future.
The author of that library explained that we can use the CASE_INSENSITIVE option like [this code does](https://github.com/curious-odd-man/RgxGen/blob/9697254f1fd56082930a049560adc407df6c2488/src/test/java/com/github/curiousoddman/rgxgen/CombinedCaseInsensitiveTests.java#L40).

fbl100:
We changed the default value for the recursive item to:
None - if it's nullable
[] - if it's an array type
{} - if it's an object type

fbl100:
Hello @spacether, I noticed that this PR had conflicts with the latest master.  I went ahead and fixed the conflicts and pushed the latest.  Let me know if you need any other changes.

spacether (approved):
Not: the Shippable CI error is unrelated
@fbl100 thank you for your PR and for making those updates. This looks great!

spacether:
This PR broke master so I am reverting it. There is a java test failure that we missed

fbl100:
@spacether everything passed on my side.  Can you point me to the build log that failed?  I'll figure it out.

spacether:
Thank you so much! Here is a log
https://github.com/OpenAPITools/openapi-generator/runs/1622405497?check_suite_focus=true
By eye it looks like maybe an os newline issue?

PR #8085, opened 2020-12-03
[dart] Cleanup and OAS3 support

kuhnroyal:
* cleanup old docs across all dart generators
* regenerate all tests
* cleanup .gitignore
* minor formatting improvements in dart-dio
* unified integration-test layout for dart/dart-dio generators
* add generation and integration-tests for OAS3 petstore to dart/dart-dio
* add generation and integration-tests for OAS3 petstore-with-fake-api to dart/dart-dio

**Explicitly not adding the new integration tests for petstore-with-fake-api to the master POM. They do not work but having them allows for quickly iterating on open issues without breaking existing basic OAS2/OAS3 petstore examples. Instead they should be run manually until everything is fixed.**

### PR checklist
 
**TODO:**
- [x] Verify new inline objects are correct, they seem to be unused
- [x] Fix compile errors in dart regarding File/MultipartFile
- [x] Fix tests for POST/PUT operations that now return `200` with content


kuhnroyal:
The new inline objects in OAS3 are caused by the PetAPI update with FormData operations.

kuhnroyal:
Don't think we can generate a useful `fromJson` function for these inline objects that contain a `MultipartFile` property.
Best would be to somehow skip these inline request body models from being generated in the first place.

kuhnroyal:
CC @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @athornz (2019/12) @amondnet (2019/12)

kuhnroyal:
@wing328 This should be ready, CircleCI seems to have failed with some sort of timeout, maybe you can restart the step.

kuhnroyal:
Not clear whats happening in CircleCI, seems to be related to some scala artifacts.

wing328:
> The new inline objects in OAS3 are caused by the PetAPI update with FormData operations. The same models are generated for java/kotlin etc. and are also unused. Why?

These models are generated due to how form parameters are defined in OAS3. Usually these are not used.

We'll revise accordingly given that it causes confusion to users.

```
[main] INFO  o.o.codegen.DefaultGenerator - Model inline_object (marked as unused due to form parameters) is generated due to the system property skipFormModel=false (default)
```

One way to prevent these from being generate is to set the `skipFormModel` option.

PR #8108, opened 2020-12-06
[Python][Client] Default to system CA instead of certifi

### LINKED ISSUE #6506: [REQ] [Python] Use urllib3 default (system) CA bundle instead of certifi

Since version [1.25.3](https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1253-2019-05-23) urllib3 defaults to using the system Certificate Authority (CA) bundle:

> 1.25.3 (2019-05-23)
> - Change HTTPSConnection to load system CA certificates when ca_certs, ca_cert_dir, and ssl_context are unspecified. (Pull [#1608](https://github.com/urllib3/urllib3/pull/1608), Issue [#1603](https://github.com/urllib3/urllib3/issues/1603))

However, the Python client currently overrides this default, using a CA bundle from certifi, instead of the system bundle, if the user does not specify `configuration.ssl_ca_cert`.

https://github.com/OpenAPITools/openapi-generator/blob/db5941379f1b6d55075a7a8d85b63cde3e149eb8/modules/openapi-generator/src/main/resources/python/rest.mustache#L62

I propose removing certifi as a dependency, requiring urllib3 >= 1.25.3 and relying on urllib3 to handle the case when no CA bundle is specified by the user.

My main reasoning is that it’s up to urllib3 to handle the ssl verification and it would be a better separation of concerns to simply pass `configuration.ssl_ca_cert` through to `urllib3.PoolManager` or `urllib3.ProxyManager`’s `ca_certs` argument without additional logic.  The generated client should still work "out of the box" for most people now that urllib3 has a default that allows for ssl verification using the system CA store.  

My secondary concern is having certifi as a default and required dependency in a project like this that's intended to run in a lot of different environments:
- The Windows system CA store is a registry entry, not a file that can be passed to `ca_certs`, so it seems quite difficult to override the certifi default with the Windows CA bundle.  However, if the system CA bundle was the default then it would be easy for the user to override it with `configuration.ssl_ca_cert=certifi.where()` if they choose.
- In a corporate setting the system CA store may be centrally managed to ensure frequent updates and/or company signed certificates for connecting to company services or proxies.  An external bundle like certifi will not include company root certificates and its installation may itself present a security policy violation under these circumstances.

If there’s some general agreement on this I’d be happy to submit a pull request.

rparini:
Resolves #6506 

Gets the Python client to use the system Certificate Authority bundle by default for verifying ssl connections (like the Python standard library does, as described in [PEP 476](https://www.python.org/dev/peps/pep-0476/#trust-database)) instead of relying on the external certifi python package for the default.  See https://github.com/OpenAPITools/openapi-generator/issues/6506 for the reasons for this change.

For the urllib3 python client this is achieved by passing `configuration.ssl_ca_cert` directly to urllib3.PoolManager or urllib3.ProxyManager’s `ca_certs` argument.  Then `ca_certs` be `None` by default and in this case urllib3 (from version 1.25.3) will load the system CA certificates, as described [here](https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1253-2019-05-23).

I have done the same with the asyncio python client since the `ssl.create_default_context` will also use the system’s default CA certificates if `cafile`, `capath` and `cadata` are all `None`, as documented [here](https://docs.python.org/3.8/library/ssl.html#ssl.create_default_context).

I have additionally:
- Removed `certifi` as a dependency
- Required `urllib3 >= 1.25.3`

These changes would be breaking for anyone relying on the certifi certificates, although I imagine most people’s system default CA certificates are adequate and in this case the change would not be noticed.  Any user still wanting to use the certifi certificates could continue to do so by setting `configuration.ssl_ca_cert=certifi.where()`.

### PR checklist
 
@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess @arun-nalla @spacether

### DIFF
--- a/modules/openapi-generator/src/main/resources/python-legacy/README_onlypackage.mustache
+++ b/modules/openapi-generator/src/main/resources/python-legacy/README_onlypackage.mustache
@@ -25,9 +25,8 @@ This python library package is generated without supporting files like setup.py 
 
 To be able to use it, you will need these dependencies in your own package that uses this library:
 
-* urllib3 >= 1.15
+* urllib3 >= 1.25.3
 * six >= 1.10
-* certifi
 * python-dateutil
 {{#asyncio}}
 * aiohttp
--- a/modules/openapi-generator/src/main/resources/python-legacy/asyncio/rest.mustache
+++ b/modules/openapi-generator/src/main/resources/python-legacy/asyncio/rest.mustache
@@ -9,7 +9,6 @@ import re
 import ssl
 
 import aiohttp
-import certifi
 # python 2 and python 3 compatibility library
 from six.moves.urllib.parse import urlencode
 
@@ -43,14 +42,7 @@ class RESTClientObject(object):
         if maxsize is None:
             maxsize = configuration.connection_pool_maxsize
 
-        # ca_certs
-        if configuration.ssl_ca_cert:
-            ca_certs = configuration.ssl_ca_cert
-        else:
-            # if not set certificate file, use Mozilla's root certificates.
-            ca_certs = certifi.where()
-
-        ssl_context = ssl.create_default_context(cafile=ca_certs)
+        ssl_context = ssl.create_default_context(cafile=configuration.ssl_ca_cert)
         if configuration.cert_file:
             ssl_context.load_cert_chain(
                 configuration.cert_file, keyfile=configuration.key_file
--- a/modules/openapi-generator/src/main/resources/python-legacy/configuration.mustache
+++ b/modules/openapi-generator/src/main/resources/python-legacy/configuration.mustache
@@ -76,6 +76,8 @@ class Configuration(object):
     :param server_operation_variables: Mapping from operation ID to a mapping with
       string values to replace variables in templated server configuration.
       The validation of enums is performed for variables with defined enum values before.
+    :param ssl_ca_cert: str - the path to a file of concatenated CA certificates 
+      in PEM format 
 
 {{#hasAuthMethods}}
     :Example:
@@ -174,6 +176,7 @@ conf = {{{packageName}}}.Configuration(
 {{/hasHttpSignatureMethods}}
                  server_index=None, server_variables=None,
                  server_operation_index=None, server_operation_variables=None,
+                 ssl_ca_cert=None,
                  ):
         """Constructor
         """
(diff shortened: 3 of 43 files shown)


spacether:
@rparini this looks good. Fundamentally this code is swapping the ssl cert source from certifi to to urllib3.
I am a little concerned about the breaking use case. How about adding a ssl_ca_cert input argument to Configuration [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/configuration.mustache#L173) to support their use case?

rparini:
Yes I think that makes sense to better document the `ssl_ca_cert` config option.  I have done this in the latest commit.

I have not touched the `python-legacy` client.  Would it be worth also making these changes there?

I also notice that running `./bin/generate-samples.sh` has not updated the `python-asyncio` petstore sample (https://github.com/rparini/openapi-generator/tree/python-default-ca/samples/client/petstore/python-asyncio) even though I have changed `src/main/resources/python/asyncio/rest.mustache`.  Do I need to do something else to get it to apply these changes?

spacether:
asyncio is still built from python-legacy. If you want to build this into python-legacy that's fine too. I have no preference either way. What do you prefer?

rparini:
Sure, I will add to python-legacy.  It looks like the changes should be the same.

rparini:
Not sure if the CircleCI error is related to my changes.  The error seems to be related to an unavailable dependency:
```
Dec 10, 2020 11:05:45 PM org.apache.maven.wagon.providers.http.httpclient.impl.execchain.RetryExec execute
INFO: Retrying request to {s}->https://repo.maven.apache.org:443
[ERROR] Plugin org.apache.maven.plugins:maven-javadoc-plugin:3.1.1 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-javadoc-plugin:jar:3.1.1: Could not transfer artifact org.apache.maven.plugins:maven-javadoc-plugin:pom:3.1.1 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]
```

spacether (approved):
Thank you for you PR and adding the new `__init__` argument! This looks great!

PR #8144, opened 2020-12-09
[Go][Server] Use the correct parameter name

randomswdev:
Implementing two minor changes:

1. When extracting the parameter value use `baseName` instead of `paramName`. The two variables usually are not different, but there is a difference in case the path parameter name includes a special character like a `-`. For example in:
`/api/resource/{resource-id}`
`baseName` is equal to `resource-id`, while `paramName` is eqaul to `resourceId`. Using the latter is incorrect.
2. When failing to parse a parameter, like for example when the code expects an integer and parsing the integer fails, I would think it is more appropriate to return StatusBadParam isntead of StatusServerInternalError, because the failure was caused by the client sending invalid data and not because of an issue on the server.

@antihax @grokify @kemokemo @bkabrda

### DIFF
--- a/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
+++ b/modules/openapi-generator/src/main/resources/go-server/controller-api.mustache
@@ -35,60 +35,60 @@ func (c *{{classname}}Controller) Routes() Routes {
 func (c *{{classname}}Controller) {{nickname}}(w http.ResponseWriter, r *http.Request) { {{#hasFormParams}}
 	err := r.ParseForm()
 	if err != nil {
-		w.WriteHeader(500)
+		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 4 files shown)


wing328:
@randomswdev thanks for the PR. Can you please resolve the merge conflicts when you've time?

randomswdev:
> 
> 
> @randomswdev thanks for the PR. Can you please resolve the merge conflicts when you've time?

I updated the pull request.

randomswdev:
Not sure why the build failed here. Any clue?

wing328:
Let me look into it over the weekend. A typical restart of the job may fix it.

wing328:
CircleCI tests passed via https://app.circleci.com/pipelines/github/OpenAPITools/openapi-generator?branch=randomswdev-fix-goserver-parameter-names

PR #8149, opened 2020-12-09
[dart][dart-dio] Enum improvements

kuhnroyal:
* use raw strings for enum value where possible
* special handling for dart-dio as raw string don't work as expected with the built_value generator
* use the correct datatype in dart enum templates

### PR checklist
 
CC @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @josh-burton (2019/12) @amondnet (2019/12)
FYI @agilob 

@noordawod I reverted some of your changes to the enum templates. As far as I can tell `isEnum` is always true in enum templates.

### DIFF
--- a/samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/model/inline_object2.dart
+++ b/samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/model/inline_object2.dart
@@ -27,10 +27,10 @@ abstract class InlineObject2 implements Built<InlineObject2, InlineObject2Builde
 class InlineObject2EnumFormStringArray extends EnumClass {
 
   /// Form parameter enum test (string array)
-  @BuiltValueEnumConst(wireName: '>')
+  @BuiltValueEnumConst(wireName: r'>')
   static const InlineObject2EnumFormStringArray greaterThan = _$inlineObject2EnumFormStringArray_greaterThan;
   /// Form parameter enum test (string array)
-  @BuiltValueEnumConst(wireName: '$')
+  @BuiltValueEnumConst(wireName: r'\$')
   static const InlineObject2EnumFormStringArray dollar = _$inlineObject2EnumFormStringArray_dollar;
 
   static Serializer<InlineObject2EnumFormStringArray> get serializer => _$inlineObject2EnumFormStringArraySerializer;
@@ -45,13 +45,13 @@ class InlineObject2EnumFormStringArray extends EnumClass {
 class InlineObject2EnumFormString extends EnumClass {
 
   /// Form parameter enum test (string)
-  @BuiltValueEnumConst(wireName: '_abc')
+  @BuiltValueEnumConst(wireName: r'_abc')
   static const InlineObject2EnumFormString abc = _$inlineObject2EnumFormString_abc;
   /// Form parameter enum test (string)
-  @BuiltValueEnumConst(wireName: '-efg')
+  @BuiltValueEnumConst(wireName: r'-efg')
   static const InlineObject2EnumFormString efg = _$inlineObject2EnumFormString_efg;
   /// Form parameter enum test (string)
-  @BuiltValueEnumConst(wireName: '(xyz)')
+  @BuiltValueEnumConst(wireName: r'(xyz)')
   static const InlineObject2EnumFormString leftParenthesisXyzRightParenthesis = _$inlineObject2EnumFormString_leftParenthesisXyzRightParenthesis;
 
   static Serializer<InlineObject2EnumFormString> get serializer => _$inlineObject2EnumFormStringSerializer;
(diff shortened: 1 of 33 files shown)


wing328:
> As far as I can tell isEnum is always true in enum templates.

Can you please elaborate on that? Sounds like a bug if it's always true independent of whether it's an enum or not.

noordawod:
Could you also elaborate more on `use the correct datatype in dart enum templates`?

kuhnroyal:
I have not found a model that get's passed into any of the `enum(_inline)` templates in which `isEnum` is false.
Which seems logical when you look at the `model` and `class` templates. They only render the `enum` templates when `isEnum` is true.

```mustache
{{#models}}
{{#model}}
{{#isEnum}}
{{>enum}}
{{/isEnum}}
{{^isEnum}}
{{>class}}
{{/isEnum}}
{{/model}}
{{/models}}
```

Inside the enum template the value will always be a `String` which prevents `int/double` based enums from being correctly generated:
```mustache
   /// The underlying value of this enum member.
  {{#isEnum}}	
  final String value;	
  {{/isEnum}}	
  {{^isEnum}}	
  final {{{dataType}}} value;
  {{/isEnum}}
```
After removing this check, the correct enum value type is generated: https://github.com/OpenAPITools/openapi-generator/pull/8149/files#diff-4671c37577153f83b40b0ad23a216cc1c6e1bbd2bbe7c074733af003c266f8b3L18

samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/model/inline_object2.dart
```diff
@@ -27,10 +27,10 @@ abstract class InlineObject2 implements Built<InlineObject2, InlineObject2Builde
 class InlineObject2EnumFormStringArray extends EnumClass {
 
   /// Form parameter enum test (string array)
-  @BuiltValueEnumConst(wireName: '>')
+  @BuiltValueEnumConst(wireName: r'>')
   static const InlineObject2EnumFormStringArray greaterThan = _$inlineObject2EnumFormStringArray_greaterThan;
   /// Form parameter enum test (string array)
-  @BuiltValueEnumConst(wireName: '$')
+  @BuiltValueEnumConst(wireName: r'\$')
```
agilob, line 33: Found a mismatch between dart-dio and dart. a few files below you will find:

```
static const dollar = EnumArraysJustSymbolEnum._(r'$');
```

You can see the `$` isn't escaped with `\` but it is here, both are raw strings. I suspect this is breaking compatibility between them?

samples/openapi3/client/petstore/dart2/petstore_client_lib_fake/lib/model/inline_object2.dart
```diff
@@ -86,23 +86,22 @@ class InlineObject2EnumFormStringArrayEnum {
...
 
   /// The underlying value of this enum member.
-  final String value;
+  final List<String> value;
 
   @override
   bool operator ==(Object other) => identical(this, other) ||
-      other is InlineObject2EnumFormStringArrayEnum && other.value == value ||
-      other is String && other == value;
+      other is InlineObject2EnumFormStringArrayEnum && other.value == value;
```
agilob, line 93: `other.value == value`

this is comparing references not contents of lists, so it's not what I would expect from `==` on enum. What are your expectations?

agilob (changes requested):

agilob:
Please change title to include dart-dio

samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/model/inline_object2.dart
kuhnroyal: See the comment here, this is specific to `built_value`s Dart generator: https://github.com/OpenAPITools/openapi-generator/pull/8149/files#diff-f325b1ca8b03ba6f1ee25144eec1079de2cc2d853d6237eac40f157663bda2acR120

agilob: ah thanks for that, missed comment. Well, it's fine then I expressed my opinion on use of built_value in another PR.

agilob (approved):

samples/openapi3/client/petstore/dart2/petstore_client_lib_fake/lib/model/inline_object2.dart
agilob: Not a change for request, just asking.

kuhnroyal: Yes this is a list but is actually completely wrong. This needs to be fixed in some other PR. An enum with a list as value is not valid. And the inline objects are gonna be removed soon.
I know what causes this but it will get too big.

Issue #4865, opened 2019-12-24
[BUG] [dart][dart-dio] generator doesn't supported nested schemas whose name contains the word "List" or "Map"


coreysprague:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used? master
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description
In my project's yaml we rely heavily on component schemas for re-using data structures. In our domain, we also use the term "Listing" as one of our primary entities.

When generating with the dart-dio generator, there is a string replace (introduced here: https://github.com/OpenAPITools/openapi-generator/pull/4018/files#diff-76962401fd8881ac1ddf6cb5e64909c5R206) that over-eagerly replaces all occurrences of "List" in the type name with "BuiltList". It unfortunately will do something like the following:

Instead of:
```BuiltList<Listing>```
we get
```BuiltList<BuiltListing>```

Thankfully, this is a compilation error and it is a trivial search/replace to post-process, but it is an unexpected trap.

##### openapi-generator version

dart-dio (tried 4.2.2 and master)

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
  title: Sample API
  version: 0.1.9
servers:
  - url: http://api.example.com/v1
    description: Optional server description, e.g. Main (production) server
paths:
  /listings:
    get:
      summary: Returns a collection of listings.
      responses:
        "200": # status code
          description: An array of listings
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ListingCollection"
components:
  schemas:
    ListingCollection:
      type: object
      properties:
        totalResults:
          type: number
        offset:
          type: number
        pageNumber:
          type: number
        listings:
          type: array
          items:
            $ref: "#/components/schemas/Listing"
    Listing:
      type: object
      properties:
        name:
          type: string
```

##### Command line used for generation

vanilla generation using the dart-dio template

##### Steps to reproduce

Your yaml must have a model class that contains a list of types who name contains the word List.

[...]

PR #8153, opened 2020-12-10
[dart][dart-dio] Use built_value collection types without string replacement

kuhnroyal:
Do not use String replacement for built_value collection types. 
This fixes https://github.com/OpenAPITools/openapi-generator/issues/4865 but also changes return/parameter types and would be a breaking change.
I think this is the correct thing to do but the changed parameter types may reduce usability in cases of list body/parameters.

@josh-burton Did you intentionally prevent the return/parameter types from being built_value types? What is your opinion on this?

### PR checklist
 
CC @ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @josh-burton (2019/12) @amondnet (2019/12) @sbu-WBT (2020/12) @kuhnroyal (2020/12) @agilob (2020/12)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
@@ -21,23 +21,17 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
+import io.swagger.v3.oas.models.Operation;
+import io.swagger.v3.oas.models.media.StringSchema;
+import io.swagger.v3.oas.models.servers.Server;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.openapitools.codegen.CliOption;
-import org.openapitools.codegen.CodegenConstants;
-import org.openapitools.codegen.CodegenType;
-import org.openapitools.codegen.DefaultCodegen;
-import org.openapitools.codegen.SupportingFile;
+import org.openapitools.codegen.*;
 import org.openapitools.codegen.meta.features.ClientModificationFeature;
 import org.openapitools.codegen.meta.features.DocumentationFeature;
 import org.openapitools.codegen.meta.features.GlobalFeature;
@@ -106,9 +100,6 @@ public class DartClientCodegen extends DefaultCodegen {
                 )
         );
 
-        // clear import mapping (from default generator) as dart does not use it at the moment
-        importMapping.clear();
-
         outputFolder = "generated-code/dart";
         modelTemplateFiles.put("model.mustache", ".dart");
         apiTemplateFiles.put("api.mustache", ".dart");
@@ -174,8 +165,6 @@ public class DartClientCodegen extends DefaultCodegen {
 
         // These are needed as they prevent models from being generated
         // which would clash with existing types, e.g. List
-        // Importing dart:core doesn't hurt but a subclass may choose to skip
-        // dart:core imports.
         importMapping.put("dynamic", "dart:core");
         importMapping.put("Object", "dart:core");
         importMapping.put("String", "dart:core");
@@ -187,6 +176,8 @@ public class DartClientCodegen extends DefaultCodegen {
         importMapping.put("Map", "dart:core");
         importMapping.put("Set", "dart:core");
         importMapping.put("DateTime", "dart:core");
+
+        defaultIncludes = new HashSet<>(Collections.singletonList("dart:core"));
 
         cliOptions.add(new CliOption(PUB_LIBRARY, "Library name in generated code"));
         cliOptions.add(new CliOption(PUB_NAME, "Name in generated pubspec"));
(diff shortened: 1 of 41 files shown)


josh-burton (approved):

josh-burton:
@kuhnroyal no not intentional, more a case of not knowing how to correctly substitute the built collection types. Even though a breaking change I think it's the right change. Nice work :)

kuhnroyal:
@wing328 Is master the correct branch for a breaking change? How should this be documented?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
```diff
@@ -188,6 +177,8 @@ public DartClientCodegen() {
         importMapping.put("Set", "dart:core");
         importMapping.put("DateTime", "dart:core");
 
+        defaultIncludes = new HashSet<>(Collections.singletonList("dart:core"));
```
agilob, line 180: `Sets.newHashSet` from guava

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
```diff
@@ -474,34 +465,62 @@ public String toDefaultValue(Schema schema) {
 
     @Override
     public String getTypeDeclaration(Schema p) {
```
agilob, line 467: can you add unit tests to this method?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
```diff
@@ -474,34 +465,62 @@ public String toDefaultValue(Schema schema) {
...
     }
 
     @Override
     public Map<String, Object> postProcessModels(Map<String, Object> objs) {
         return postProcessModelsEnum(objs);
     }
 
+    @Override
+    public CodegenOperation fromOperation(String path, String httpMethod, Operation operation, List<Server> servers) {
+        final CodegenOperation op = super.fromOperation(path, httpMethod, operation, servers);
```
agilob, line 509: can you add unit tests for this method? is it worth to make it more dart-dio specific as the rest of the changes are dart-dio?

agilob:
I'm not a fan of these changes TBH, it has the following disadvantages I see straight away:

1. Breaking change in terms of input/output to methods
2. 3rd party library to support such basic concepts like lists and maps
3. I heave to know how to use that 3rd party library to use lists/maps (I was not aware this library even exists until now)
4. When I generate an Owner with Pets I want to add and remove pets from existing array as I go and not be limited by  immutable arrays that it holds by default. It restricts how classes can be used.
5. TypeScript and Java generators that I use produce mutable lists so it's also introducing incompatibility between languages and how generated code can be used

If anything, this should be a runtime flag in custom properties.

That said, I much prefer shelf and not planing to use dart-dio, so whatever floats your boat works for me, but please don't put it in main dart. If these changes landed in any generator I use I would be really pissed as it completely breaks 100% of cases how I use generated classes in Java and Dart. Autogenerated code are my DTOs and small cached objects, so shared schema and how classes are used between languages absolutely has to be the same, eg. lists and maps have to be mutable.

#4865 should be solved in a way that's compatible with other generators, so contributors from scala or python can come to the dart generators and be familiars with code generation process.

agilob:
Any opinion on this @sbu-WBT ?

kuhnroyal:
> When I generate an Owner with Pets I want to add and remove pets from existing array as I go and not be limited by immutable arrays that it holds by default. It restricts how classes can be used.

IMO this depends on the language/framework features. I usually prefer immutability. Many other generators for languages that have immutable collections by default do the same. Java and Typescript are not some of those as we all know :)

If you choose to use this generator, you also choose to use `built_value`, all the models you get from the API are immutable built_value classes, everything you put into it as well. It only makes sense to use the immutable data structures provided by built_value which use the same builder pattern as the model classes. It is much easier for responses which are deserialized as `BuiltList/Map/Set` anyways.

> but please don't put it in main dart

I would never dream of it. This is here for dart-dio because the used collection classes are part of the same serialization package that is used.

> It takes dart-dio generated code further from dart generator, making them less and less compatible. TypeScript is suffering from similar problems, each generator is completely different, has different codes, settings and produces different outputs. Java and Python generators are all similar, you can swap gson with jackson generator and almost no difference in use. We should be aiming for less split between generators, not more.

Sounds great and would very much be possible if we could use dart mirrors in Flutter, like we use reflection in Java :) Languages that don't have a form reflection will have the problem of generators being a lot more different to each other because everything that is different has to be expressed in code and can not be weaved in at runtime.


It is a great idea to provide a flag to choose a specific serialization framework, as other generators do. But dart-dio doesn't do this yet and it is way out of scope for this issue.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
kuhnroyal: There are tests for dart and dart-dio that test the type returned for `array` and `map` OAS types.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
kuhnroyal: Could possibly be moved to the dart-dio generator.

agilob:
>Sounds great and would very much be possible if we could use dart mirrors in Flutter, like we use reflection in Java :) Languages that don't have a form reflection will have the problem of generators being a lot more different to each other because everything that is different has to be expressed in code and can not be weaved in at runtime.

>You can swap gson with jackson generator and almost no difference in use

I didn't mean reflection or mirrors, just content of X1Generator and X2Generator and .mustache files should be as similar as possible to reduce duplication, produce outputs that are more similar. You can swap gson for jackson and as long as there are correct dependencies, it will all work without any fuss.

But as above, I'm not familiar with dart-dio, so whatever works for you here.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
agilob: makes sense to keep it here for all darts.

agilob (approved):
Minor change requests not worth effort of a commit, good to go.

PR #8174, opened 2020-12-13
[dart][dart-dio] More enum fixes for inner types

kuhnroyal:
### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/dart2/class.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/class.mustache
@@ -15,23 +15,8 @@ class {{{classname}}} {
     {{#maximum}}
   // maximum: {{{maximum}}}
     {{/maximum}}
-  {{{dataType}}} {{{name}}};
   {{/isEnum}}
-  {{#isEnum}}
-    {{#isContainer}}
-      {{#isArray}}
-  List<{{{classname}}}{{{enumName}}}> {{{name}}};
-      {{/isArray}}
-      {{#isMap}}
-  Map<String, dynamic> {{{name}}};
-      {{/isMap}}
-    {{/isContainer}}
-    {{^isContainer}}
-      {{#allowableValues}}
-  {{{classname}}}{{{enumName}}} {{{name}}}{{#required}}{{#defaultValue}} = {{{classname}}}{{{enumName}}}._({{{defaultValue}}}){{/defaultValue}}{{/required}};
-      {{/allowableValues}}
-    {{/isContainer}}
-  {{/isEnum}}
+  {{{datatypeWithEnum}}} {{{name}}};
 
   {{/vars}}
   @override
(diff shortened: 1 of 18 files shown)


samples/openapi3/client/petstore/dart2/petstore_client_lib_fake/lib/model/enum_arrays.dart
```diff
@@ -159,7 +159,7 @@ class EnumArraysArrayEnumEnum {
   const EnumArraysArrayEnumEnum._(this.value);
 
   /// The underlying value of this enum member.
-  final List<String> value;
```
kuhnroyal, line 162: @agilob This commit also fixes the wrong value type that was appearrent in the other PR.

modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -15,23 +15,8 @@ class {{{classname}}} {
...
-  Map<String, dynamic> {{{name}}};
-      {{/isMap}}
-    {{/isContainer}}
-    {{^isContainer}}
-      {{#allowableValues}}
-  {{{classname}}}{{{enumName}}} {{{name}}}{{#required}}{{#defaultValue}} = {{{classname}}}{{{enumName}}}._({{{defaultValue}}}){{/defaultValue}}{{/required}};
-      {{/allowableValues}}
-    {{/isContainer}}
   {{/isEnum}}
+  {{{datatypeWithEnum}}} {{{name}}};
```
kuhnroyal, line 19: I am not sure about all of this change, especially the enum default value. Currently all default values are set in the constructor. As far as I can tell in that case a wrong constructor default value would be generated.

agilob:
Took me long time to understand from class name whether this is correct or not... Do we **need** `...EnumEnum`? Do we need to append every enum class with `Enum`? Kind of excessive.

dart.dev doesn't recommend it https://dart.dev/guides/language/language-tour#enumerated-types 

No other generator does it. It was even difficult to reverse engineer to find original spec from the names.

modules/openapi-generator/src/main/resources/dart2/class.mustache
agilob: I think the default value should stay as null, unless field is marked as non-nullable, required or has default value provided, this looks good to me.

kuhnroyal:
The `EnumEnum` is just happening because the model/property name already ends with `Enum`.
Appending `Enum` is done in `DefaultCodegen` and some generators do their own modifications but it is that way in several languages: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/java/native/src/main/java/org/openapitools/client/model/EnumArrays.java

This is already this way in the dart generator and has always been. (hence no changes to dart output, only dart-dio)

I am not particularly fond of the naming, I was actually thinking about removing it. Removing it will be a breaking change in the dart generator instead of dart-dio. Would be good to get some some more opinions on this.

agilob (approved):

PR #8177, opened 2020-12-14
[typescript-angular] Dates Path Parameters are not encoded properly

ratoaq2:
Dates path parameters should also use .toISOString()

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -421,7 +421,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
                         // Add the more complicated component instead of just the brace.
                         CodegenParameter parameter = findPathParameterByName(op, parameterName.toString());
                         pathBuffer.append(toParamName(parameterName.toString()));
-                        if (parameter != null && parameter.isDateTime) {
+                        if (parameter != null && (parameter.isDateTime || parameter.isDate)) {
                             pathBuffer.append(".toISOString()");
                         }
                         pathBuffer.append("))}");


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -421,7 +421,7 @@ public void postProcessParameter(CodegenParameter parameter) {
                         // Add the more complicated component instead of just the brace.
                         CodegenParameter parameter = findPathParameterByName(op, parameterName.toString());
                         pathBuffer.append(toParamName(parameterName.toString()));
-                        if (parameter != null && parameter.isDateTime) {
+                        if (parameter != null && (parameter.isDateTime || parameter.isDate)) {
```
macjohnny, line 424: given that the actual `TypeScript` type of a `date` or `date-time` field is `string`, this if statement should be removed

https://github.com/openapitools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularModelTest.java#L51
https://github.com/openapitools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/typescript/typescriptangular/TypeScriptAngularModelTest.java#L96

see also https://github.com/OpenAPITools/openapi-generator/pull/5314

@amakhrov what is your opinion?

amakhrov: Agree.

@ratoaq2 - could you post an example spec and codegen options you use that result in an issue you're facing?

ratoaq2: Yes, sure.

My scenario is: All date and date-time schemas from openapi spec are mapped to Moment (from Moment.js) type. We have an Angular interceptor that makes sure that all json from http response's models have the dates and date-times converted to Moment

Also all input parameters (query params, path params, etc) that are date or date-time are Moment.

This is our maven plugin configuration:

```
				<configuration>
					<generatorName>typescript-angular</generatorName>
					<inputSpec>${apiSpecLocation}</inputSpec>
					<output>${apiGeneratedTypescriptLocation}</output>
					<templateDirectory>${openApiTemplateDirectory}/typescript-angular</templateDirectory>
					<generateApis>true</generateApis>
					<addCompileSourceRoot>false</addCompileSourceRoot>
					<generateSupportingFiles>true</generateSupportingFiles>
					<configOptions>
						<providedInRoot>false</providedInRoot>
						<modelPropertyNaming>original</modelPropertyNaming>
						<sortParamsByRequiredFlag>true</sortParamsByRequiredFlag>
						<ensureUniqueParams>true</ensureUniqueParams>
						<supportsES6>false</supportsES6>
						<taggedUnions>false</taggedUnions>
						<enumPropertyNaming>original</enumPropertyNaming>
						<useSingleRequestParameter>true</useSingleRequestParameter>
						<ngVersion>10.0.5</ngVersion>
					</configOptions>
					<importMappings>
						<importMapping>Moment=moment</importMapping>
					</importMappings>
					<typeMappings>
						<typeMapping>object=any</typeMapping>
						<typeMapping>set=Array</typeMapping>
						<typeMapping>date=Moment</typeMapping>
						<typeMapping>DateTime=Moment</typeMapping>
					</typeMappings>
				</configuration>
```

And we had to customize two templates:
Added a workaround in `models.mustache` since the apis generates a wrong import for the moment import mapping: `import { Moment } from '../model/models';`
```
/*
 * - Based on https://github.com/OpenAPITools/openapi-generator/blob/cf29908/modules/openapi-generator/src/main/resources/typescript-angular/models.mustache
 * - Added workaround to fix service imports for Moment.js
 */
export { Moment } from 'moment';
```


And `api.service.mustache`:
```
/*
 * - Based on https://github.com/OpenAPITools/openapi-generator/blob/08fb590/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache
 * - Use isMoment when handling Moment dataType
 * - Change bodyParam to always be named 'body' (since x-codegen-request-body-name doesn't seem to work with inheritance/oneOf)
 * - Workaround for Date path param  ( https://github.com/OpenAPITools/openapi-generator/pull/8177 )
 */

// isMoment fix
import { isMoment } from 'moment';


export interface {{#prefixParameterInterfaces}}{{classname}}{{/prefixParameterInterfaces}}{{operationIdCamelCase}}RequestParams {
{{#allParams}}
    {{#description}}/** {{description}} */
    // Body param name fix
    {{/description}}{{^isBodyParam}}{{paramName}}{{/isBodyParam}}{{#isBodyParam}}body{{/isBodyParam}}{{^required}}?{{/required}}: {{{dataType}}}{{#isNullable}} | null{{/isNullable}};
{{/allParams}}

...

    private addToHttpParams(httpParams: HttpParams, value: any, key?: string): HttpParams {
        // isMoment fix
        if (typeof value === "object" && value instanceof Date === false && !isMoment(value)) {
            httpParams = this.addToHttpParamsRecursive(httpParams, value);
        } else {

...

// Workaround for Date path param
{{#useSingleRequestParameter}}
        const {{paramName}} = requestParameters.{{^isBodyParam}}{{paramName}}{{/isBodyParam}}{{#isBodyParam}}body{{/isBodyParam}}{{#isDate}}?.clone(){{/isDate}};
{{/useSingleRequestParameter}}
{{#required}}
        if ({{paramName}} === null || {{paramName}} === undefined) {
            throw new Error('Required parameter {{paramName}} was null or undefined when calling {{nickname}}.');
        }
{{/required}}
{{#useSingleRequestParameter}}
{{#isPathParam}}
{{#isDate}}
        // workaround for https://github.com/OpenAPITools/openapi-generator/pull/8177
        {{paramName}}.toString = () => {{paramName}}.toISOString();
{{/isDate}}
{{/isPathParam}}
{{/useSingleRequestParameter}}
```

I understand that OpenAPI spec 3 defines date and date-time as string. But they have a format (date or date-time)
IMO, openapi-generator should allow us to use isDate and isDateTime functions to know if that's the case and allow us to customize date/date-time parameters.

amakhrov: Thanks for the details, it makes sense to me.

So you have custom type mapping and custom template changes that support that mapping. This is totally legit.

On the other hand, changing the default behavior of angular codegen to do `.toISOString()` without changing the default mapping (which has good reasons to be mapped to `string` currently) would result in broken generated code. Or am I missing the point?

ratoaq2: Yes, that `toISOString()` is just a workaround because of the issue I tried to fix in this PR. That's not a suggested fix. I just shared the whole scenario to have a better picture.

macjohnny: if I am not mistaken @amakhrov refers to https://github.com/OpenAPITools/openapi-generator/pull/8177/files#diff-4602e23d3a025b64a09d0826cd43f90aaf25035660e9db37442cdc93c5e2f4e3R425

@ratoaq2 I think we need to find an alternative solution in order to avoid a breaking change, but still allow the customization you are looking for. 
how about a callback function e.g. in the configuration that could transform the value as needed? it would be great if you could extend the generated readme how to use the generated angular code with moment dates

ratoaq2: could you explain better this "callback function in the configuration"?

I'm very interested in helping to have a way to use moment.js out-of-the-box without the need to override and customize templates for that

macjohnny: @ratoaq2 we could introduce a "processing function" that can be added to the configuration similar to the `accessToken` callback
https://github.com/OpenAPITools/openapi-generator/blob/9b6fd5a27df34e6b7da1e20a791015aa912cf53d/samples/client/petstore/typescript-angular-v10-provided-in-root/builds/default/configuration.ts#L13

and pipe the input to the api call through that function, which could process / transform e.g. dates.
we could limit this to the case when `useSingleRequestParameter` is enabled, so all input parameters can be passed to the function as one object and will be returned as one object.

https://github.com/OpenAPITools/openapi-generator/blob/9b6fd5a27df34e6b7da1e20a791015aa912cf53d/modules/openapi-generator/src/main/resources/typescript-angular/api.service.mustache#L160-L166

ratoaq2:
It's possible to use Param encoders now

PR #8201, opened 2020-12-15
[dart][dart-dio] Add support for default values

kuhnroyal:
This PR adds support for default values to the dart-dio generator.

* return `null` when there is no default value instead of `"null"` (this also removes the "defaults to null" from model docs)
* don't need any collection literal default values for built_value
* use built_value's `_initializeBuilder` to set default values
*  fix formatting of default values in model doc

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartDioClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartDioClientCodegen.java
@@ -125,20 +125,13 @@ public class DartDioClientCodegen extends DartClientCodegen {
 
     @Override
     public String toDefaultValue(Schema schema) {
-        if (ModelUtils.isMapSchema(schema)) {
-            return "const {}";
-        } else if (ModelUtils.isArraySchema(schema)) {
-            return "const []";
-        }
-
         if (schema.getDefault() != null) {
             if (ModelUtils.isStringSchema(schema)) {
                 return "'" + schema.getDefault().toString().replaceAll("'", "\\'") + "'";
             }
             return schema.getDefault().toString();
-        } else {
-            return "null";
         }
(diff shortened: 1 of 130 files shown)


agilob:
>return null when there is no default value instead of "null" (this also removes the "defaults to null" from model docs)

:+1:

agilob (approved):

agilob (approved):

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartDioClientCodegen.java
```diff
@@ -125,20 +125,13 @@ public String getHelp() {
 
     @Override
     public String toDefaultValue(Schema schema) {
-        if (ModelUtils.isMapSchema(schema)) {
```
agilob, line 128: Does it make sense to remove these 4 lines from `DartClient` too?

kuhnroyal: No, in the dart generator you need the collection default values. But in dart-dio there are none, we use a builder pattern there.

kuhnroyal: The models are mutable and you said you would expect to have an empty collection that you can just add to after instantiating the object. This is what the collection default values are used for.

agilob: >This is what the collection default values are used for.

Yes, when I worked on that I kept it for backwards compatibility, I would do it differently now: keep compatibility with most popular generators where default value is null, unless field is required or defined as non-null. At some point we'll have to break it, but not rush.

PR #8219, opened 2020-12-17
Fix logic for removal of forward slash in RegGen pattern

code-lucidal58:
Fix logic for removal of forward slash in RegGen pattern
@spacether


wing328:
I wonder if a test can be added to cover this issue moving forward.

code-lucidal58:
I shall add tests for the same.

wing328:
Looks like this is causing issues to the python-legacy generator:
```

Exception in thread "main" java.lang.RuntimeException: Could not process model 'format_test'.Please make sure that your schema is correct!
--
3624 | at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:499)
3625 | at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:875)
3626 | at org.openapitools.codegen.cmd.Generate.execute(Generate.java:432)
3627 | at org.openapitools.codegen.cmd.OpenApiGeneratorCommand.run(OpenApiGeneratorCommand.java:32)
3628 | at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:66)
3629 | Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 0
3630 | at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
3631 | at java.base/java.lang.String.substring(String.java:1874)
3632 | at org.openapitools.codegen.languages.PythonLegacyClientCodegen.toExampleValueRecursive(PythonLegacyClientCodegen.java:800)
3633 | at org.openapitools.codegen.languages.PythonLegacyClientCodegen.toExampleValue(PythonLegacyClientCodegen.java:718)
3634 | at org.openapitools.codegen.DefaultCodegen.fromProperty(DefaultCodegen.java:3095)
3635 | at org.openapitools.codegen.DefaultCodegen.addVars(DefaultCodegen.java:4881)
3636 | at org.openapitools.codegen.DefaultCodegen.addVars(DefaultCodegen.java:4823)
3637 | at org.openapitools.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:2547)
3638 | at org.openapitools.codegen.DefaultGenerator.processModels(DefaultGenerator.java:1250)
3639 | at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:494)
3640 | ... 4 more
```

Can you please take a look?

code-lucidal58:
For this error, all patterns in the YAML files should start with ^ and end with $. Is it correct to make this assumption? If not I can add a line to `toExampleValueRecursive` after the two `while` statements that if the pattern string is empty (meaning it never encountered ^ and $, then use the original pattern string.

wing328:
> For this error, all patterns in the YAML files should start with ^ and end with $. Is it correct to make this assumption?

I don't think that's a correct assumption

spacether:
[From the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#properties)
```
pattern (This string SHOULD be a valid regular expression, according to the [ECMA 262 regular expression dialect](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5))
```
Per there we can have valid regexes like:
`b/hi /g`
or `/hi/g`

code-lucidal58:
Updated logic:
if regex starts with `/` trim it. If it ends with `/i` `/g` `/m` or `/` trim it

modules/openapi-generator/src/test/resources/3_0/examples.yaml
```diff
@@ -118,3 +118,16 @@ components:
...
+      properties:
+        name:
+          type: string
+          pattern: '^[a-z\s]+$'
+        number:
+          type: string
+          pattern: '^[0-9]{10}$'
+        address:
+          type: string
+          pattern: '^[a-z0-9\s]+$'
```
spacether, line 132: How about updating these examples to include: leading / trailing / and /g  /i /m ?
That way we show that this feature is working and if it is changed in the future the test will break.
- slash use case `/^[a-z\s]+$/`
- /g use case `/^[a-z\s]+$/g`
- /i use case `/^[a-z\s]+$/i`
- /m use case `/^[a-z\s]+$/m`

spacether:
Thank you for updating your sample spec, that looks good. Now the tests are failing. Can you take a look here and update them?
```
[ERROR] testGeneratedExampleValues(org.openapitools.codegen.python.PythonLegacyClientCodegenTest)  Time elapsed: 0.06 s  <<< FAILURE!
java.lang.AssertionError: expected [true] but found [false]
	at org.openapitools.codegen.python.PythonLegacyClientCodegenTest.testGeneratedExampleValues(PythonLegacyClientCodegenTest.java:108)
```
Closing and re-opening the PR to kick off CI again. Some of these failures were unrelated.

code-lucidal58:
The testcases are failing in PythonLegacyTest. Need some to fix it. Next 24hours would be enough

wing328:
CircleCI tests passed via https://app.circleci.com/pipelines/github/OpenAPITools/openapi-generator/7685/workflows/780c57e5-ce78-4609-965b-5217897de966/jobs/21214

PR #8226, opened 2020-12-17
Go onoef discriminator err

code-lucidal58:
Here is the logic I have tried to implement
```txt
if useOneOfDiscriminatorLookup and discriminator:
  use discriminator to marshal
else:
  use object marshalling and check if it fits into any struct
```
Since mustache does not have `and` operator, there is a block of code that is used twice. 
```txt
if useOneOfDiscriminatorLookup:
  if discriminator:
    use discriminator to marshal
  else:
    use object marshaling and check if fits into any struct
else:
  use object marshaling and check if fits into any struct
  ```

@antihax (2017/11) @grokify (2018/07) @kemokemo (2018/09) @bkabrda (2019/07)

### DIFF
--- a/bin/configs/go-petstore.yaml
+++ b/bin/configs/go-petstore.yaml
@@ -7,3 +7,4 @@ additionalProperties:
   packageName: petstore
   disallowAdditionalPropertiesIfNotPresent: false
   generateInterfaces: true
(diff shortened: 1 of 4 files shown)


wing328:
Closed via https://github.com/OpenAPITools/openapi-generator/pull/11178

PR #8235, opened 2020-12-18
[dart][dart-dio] Nullable support/improvements

kuhnroyal:
### dart-dio/built_value

+ **Disable all fields nullable by default:**
This is not in line with OAS and will prevent future Dart nullabilty features (NNBD) from being useful as all types would be optional.

+ **Properties are `@nullable` when not required AND not nullable**

+ **Support nullable/required fields in JSON**
  * properties in built_value need to be `@nullable` when they are nullable in OAS or when they are not required in OAS
  * built_value does not support serializing `null` values by default as it is based on a serialization format based on iterables/lists and not maps
  * dart-dio uses the built_value json plugin to convert the built_value format to regular json
  * by generating a custom serializer for each class we can add support for serializing `null` values if the property is required AND nullable in OAS
  * this is a breaking change as not all properties in the models are nullable by default anymore

### Dart
* basically just serialize `null` when the field is required and nullable

This is based on https://github.com/OpenAPITools/openapi-generator/pull/8220 which needs to be merged first.

### PR checklist
 

kuhnroyal:
@josh-burton Would really like your input on this.

kuhnroyal:
@agilob Please check it out, serialization for the normal Dart generator is a mess so I am not so sure

modules/openapi-generator-gradle-plugin/gradle.properties
```diff
@@ -1,5 +1,6 @@
 # RELEASE_VERSION
-openApiGeneratorVersion=5.0.1-SNAPSHOT
+openApiGeneratorVersion=5.1.0-SNAPSHOT
+>>>>>>> origin/master
```
agilob, line 3: merge conflict message here

modules/openapi-generator-cli/pom.xml
```diff
@@ -4,7 +4,7 @@
         <groupId>org.openapitools</groupId>
         <artifactId>openapi-generator-project</artifactId>
         <!-- RELEASE_VERSION -->
-        <version>5.0.1-SNAPSHOT</version>
+        <version>5.1.0-SNAPSHOT</version>
```
agilob, line 7: I think version bump should go in a separate PR, but leaving this to @wing328

modules/openapi-generator/src/main/resources/dart2/enum_inline.mustache
```diff
@@ -1,14 +1,14 @@
 {{#description}}/// {{{description}}}{{/description}}
-class {{{classname}}}{{{enumName}}} {
+class {{{enumName}}} {
```
agilob, line 2: I think that's a big breaking change for no big benefit, enum serialisation still won't be as good as in other languages. I think it's time to give up on doing this in openapi and use a more popular 3rd party library do it for us. I'm only familiar with json_serializable but it doesn't handle standalone enums so can't use it here. I'll look around soon.

kuhnroyal:
Yea, wrong target and the enum branch this is based up on has already been discussed.

modules/openapi-generator/src/main/resources/dart2/enum_inline.mustache
kuhnroyal: This is actually part of  #8220 but required for this to work.

modules/openapi-generator-cli/pom.xml
kuhnroyal: Wrong target branch

Issue #8179, opened 2020-12-14
[REQ][dart] Dart full OAS3 support


kuhnroyal:
This is an epic to track remaining work for full dart/dart-dio OAS3 support.

**General:**
- [ ] OAS3 fake petstore contains a model called `File`. This is generated as a separate model in the Go client but as `java.io.File` in Java/Kotlin. It should probably be a separate model. Currently this leads to compile errors.

**Dart:**
- [ ] There are problems regarding deserialization of maps, de-serialization code is trying to call `Map.mapFromJson`. There are open issues for this and the problem is now evident in the generated fake petstore.
Related issues: https://github.com/OpenAPITools/openapi-generator/issues/8029
- [ ] Mapping error in `array_test.dart` de-serialization
- [ ] Support enums of type `double`,  de-serialization via switch statement doesn't work here
- [ ] De-serialization of `List<Object>` tries to call `Object.mapFromJson`
- [ ] OAS3 fake petstore contains a model called `Client`, this conflict with `http.Client`

**Dart-Dio:**
- [ ] Compile errors regarding header parameters of type `bool` and `List`, basically everything that is not a `String`

kuhnroyal:
@agilob FYI if you have time to tackle some of the Dart issues

agilob:
Might have more time in 2 weeks during Xmas break, but can't promise it.

kuhnroyal:
Dart-dio now compiles with full OAS3 fake petstore.

Grohden:
I'm curious, is `dart` generator working for you all using the current templates?
I had to fix a lot of cast issues from json conversions for the `dart` generator, here's my files if it help...
https://gist.github.com/Grohden/1fd8d110c806d8b239e88d07847c1e5a

But still I'm not able to even compile the project, there's still some issues with enum names :/

wing328:
@Grohden are you using the latest master to generate the Dart API client? There's been many enhancements, bug fixes to the Dart generators

Grohden:
@wing328 I'm manually copying the current master dart templates.
But I'm still on the latest released beta (must be on the beta3). Is there any way to use the master 'release' without cloning and building it myself? (a wget url)

wing328:
You can find the snapshot version of the latest master in the project's README

kuhnroyal:
@Grohden you can not only copy the templates, that is not enough, there are many changes in the generators as well.

The problems regarding json serialization are listed as todos.

[...]

PR #8244, opened 2020-12-20
[dart] fix some compilation issues and added casts for analyzer implicit-casts flag

Grohden:
### Details

I've made some fixes for the **dart** templates, @kuhnroyal and I had some discussion on #8179.

My changes were necessary to be able to compile **my personal project** and also the fake petstore generated clients.

**Not all changes were made to fix compile errors**: I've mistakenly enabled the `implicit-casts:false` on the generated client and then fixed the resulting analyzer errors.. 
 * if fixing casts 'issues' and supporting that flag is not desired I can revert my cast changes.
 * if fixing casts 'issues' and supporting that flag is desired we could enable the strong mode like @kuhnroyal did in #8231

From the changelist listed in #8179 I've fixed:
  - [x] De-serialization of maps of maps e.g. Map<String, Map<String, String>>

(I've only fixed the compilation error, not sure if it will work properly on runtime)

Maybe I've fixed other issues.. but I'm not sure, would be glad if someone test it and confirm exactly what is still broken

### PR checklist
 
@kuhnroyal could you comment on the casts you said that were uneeded?

@swipesight @jaumard @josh-burton @amondnet @sbu-WBT @kuhnroyal @agilob


modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -190,22 +188,22 @@ class {{{classname}}} {
...
     return map;
   }
 
   // maps a json object with a list of {{{classname}}}-objects as value to a dart map
   static Map<String, List<{{{classname}}}>> mapListFromJson(Map<String, dynamic> json, {bool emptyIsNull, bool growable,}) {
     final map = <String, List<{{{classname}}}>>{};
-    if (json != null && json.isNotEmpty) {
-      json.forEach((String key, dynamic v) {
-        map[key] = {{{classname}}}.listFromJson(v, emptyIsNull: emptyIsNull, growable: growable);
+    if (json?.isNotEmpty == true) {
```
agilob, line 204: do we need `== true`?

modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -190,22 +188,22 @@ class {{{classname}}} {
...
+      json.forEach((key, value) {
+        map[key] = {{{classname}}}.listFromJson(value as List<dynamic>, emptyIsNull: emptyIsNull, growable: growable,);
```
agilob, line 206: can you assess if this has any performance hit? Can casting things inside a loop slower be slower than casting upfront in loop definition?

modules/openapi-generator/src/main/resources/dart2/class.mustache
Grohden: yes

```dart
void main() {
  if(null) {
    print("x");
  }
}

// Uncaught Error: Assertion failed: "boolean expression must not be null"
```

modules/openapi-generator/src/main/resources/dart2/api.mustache
```diff
@@ -192,7 +192,7 @@ class {{{classname}}} {
       {{/isArray}}
       {{^isArray}}
         {{#isMap}}
-      return {{{returnType}}}.from(apiClient.deserialize(_decodeBodyBytes(response), '{{{returnType}}}'));
+      return {{{returnType}}}.from(apiClient.deserialize(_decodeBodyBytes(response), '{{{returnType}}}') as Map<dynamic, dynamic>);
```
agilob, line 195: we already know it's a map of things, what does knowing it is `<dynamic, dynamic>` bring?

modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -109,15 +109,15 @@ class {{{classname}}} {
           ? null
           : (json[r'{{{baseName}}}'] as List).map(
       {{#items.complexType}}
-              {{items.complexType}}.listFromJson(json[r'{{{baseName}}}'])
+              {{items.complexType}}.listFromJson(json[r'{{{baseName}}}'] as List<dynamic>)
```
agilob, line 112: same here, we know that the method accepts `List<dynamic> json` as first param, so why the cast here?

modules/openapi-generator/src/main/resources/dart2/api.mustache
Grohden: With the implicit casts false we get this:
image

which means that we can't implicit cast a type to a more specialized type (`dynamic` to `Map<dynamic, dynamic>`)

modules/openapi-generator/src/main/resources/dart2/class.mustache
Grohden: same as above, with the `implicit-casts:false` we cant implicit cast `dynamic` to `List<dynamic>`, but as I said in the description, if that flag is not desirable I can remove almost all casts :thinking: and I'm pretty sure it will still work

modules/openapi-generator/src/main/resources/dart2/class.mustache
Grohden: > can you assess if this has any performance hit?

I lack the knowledge for that...

> Can casting things inside a loop slower be slower than casting upfront in loop definition?

U mean to move this to the method signature? declaring `Map<String, dynamic> json` as `Map<String, List<dynamic>> json`?

On that note.. isn't this a microoptimization? I know that dart runtime will validate all assignment types whether you cast or not something, maybe the compiler make some optimizations when you cast before to not validate it again :thinking:... but again, I lack the knowledge for that.

oh, and [json_serializable](https://pub.dev/packages/json_serializable) also generates theses casts that I've added

modules/openapi-generator/src/main/resources/dart2/api.mustache
Grohden: maybe we could just cast it instead of applying the map.from? :thinking: gonna try that in the https://dartpad.dev/

agilob (approved):

modules/openapi-generator/src/main/resources/dart2/api.mustache
Grohden: just casting it would work:
```dart
void main() {
  print(foo({"a": 1}));
}

Map<String, int> foo(dynamic response) {
  return response as Map<String, int>;
}
```

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
```diff
@@ -70,7 +70,7 @@ class ApiClient {
   Future<Response> invokeAPI(
     String path,
     String method,
-    Iterable<QueryParam> queryParams,
+    List<QueryParam> queryParams,
```
agilob, line 73: User of this API should be able to pass `Set` as well, eg. let dart remove duplicated elements and not relay on order of elements in Iterable.

modules/openapi-generator/src/main/resources/dart2/api.mustache
agilob: that's fine I'm cool with it, just was asking

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
Grohden: I've changed this one based on the call to `_updateParamsForAuth`, this function expect the param to be a list, passing a set would be accepted but would result in runtime error when you call that function

image

```
void main() {
  foo({"a", "b"});
}

void foo(Iterable<String> authNames) {
  bar(authNames);
}

void bar(List<String> authNames) {
}

// Uncaught Error: TypeError: Instance of '_LinkedHashSet<String>': type '_LinkedHashSet<String>' is not a subtype of type 'List<String>'
```

agilob: There doesn't seem to be any reason for `_updateParamsForAuth` being limited to `List` it's not order dependent either

```dart
  /// Update query and header parameters based on authentication settings.
  /// @param authNames The authentications to apply
  void _updateParamsForAuth(List<String> authNames, List<QueryParam> queryParams, Map<String, String> headerParams) {
    authNames.forEach((authName) {
      Authentication auth = _authentications[authName];
      if (auth == null) throw new ArgumentError("Authentication undefined: " + authName);
      auth.applyToParams(queryParams, headerParams);
    });
  }
```

Set seems to be even more correct options as some frameworks (I remember one PHP based doing it) will throw error if you pass the same url param twice.

Grohden: Well, what you said make sense, I'm gonna try to change the signature of `_updateParamsForAuth` and `applyToParams` to receive a `Iterable` instead of a `List` then

Grohden: I forgot to thank you for the review, thanks for that :heart: 

It turns out that I can't change the signature of `applyToParams` to a `Iterable`: in `ApiKeyAuth` there's a call to `.add`, and Iterables don't have that method :/

We could change the method and add a `.toList()` call:
```dart
  void _updateParamsForAuth(
    List<String> authNames,
    Iterable<QueryParam> queryParams,
    Map<String, String> headerParams,
  ) {
    for(final authName in authNames) {
      final auth = _authentications[authName];
      if (auth == null) {
        throw ArgumentError('Authentication undefined: $authName');
      }
      auth.applyToParams(queryParams.toList(), headerParams);
    }
  }
```

But I'm pretty sure that this would make the `applyToParams` implementers `.add` calls pointless.. they rely on mutability, and `.toList` would create a copy of the original list

modules/openapi-generator/src/main/resources/dart2/class.mustache
kuhnroyal: Casts usually don't have a runtime impact, they are mostly there to make the compiler happy.

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
agilob: Does that mean it currently throws at runtime if we pass Set in place of the existing iterable, which at the end is expected to be a List?

kuhnroyal:
You should add a matching `analysis_options.yaml` as a new support template file, like it is done in the dio-generator.

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
agilob: so basically for now it has to be specific implementation Set or List but not parent class `Iterable` because it doesn't have `.add` method :clap:. 

sure, leave it as list then

agilob:
@kuhnroyal  @Grohden 
 what do you two think about adding something more popular and opinionated that a custom analysis? Would you add effective-dart here?

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
agilob: >We could change the method and add a .toList() call:

My personal choice would be to pass iterable to allows sets and any other lists and sets, but it's also outside of scope of your PR, so I'm leaving the decision with you.

Grohden:
@agilob would totally use use effective dart, I can make the needed changes if @kuhnroyal also agrees

kuhnroyal:
I would love to see a linter in dart/dart-io at some point.
Need to decide which makes the most sense here.
https://pub.dev/packages/effective_dart
https://pub.dev/packages/lint
https://pub.dev/packages/pedantic

I usually use lint.
And yes, should be done in another PR and added to both generators as well as to the integration tests.

agilob:
Pedantic always makes my eyes bleed, it makes Dart have more strict syntax than Java. I personally stick with effective_dart, but I've never used `lint`... and I think I know why ` An opinionated, community-driven set of lint rules for Dart and Flutter projects. Like pedantic but stricter `

agilob:
>edit: effective could be done in another PR to not bloat this one

I disagree here, since you're adding lint fixes it makes sense to add the linter too!

Grohden:
> I disagree here, since you're adding lint fixes it makes sense to add the linter too!

Somehow this PR ended up as lint fixes, but my primary problem was the `Map.mapfromJson` not letting me compile my project.. I've mistakenly enabled the analyzer and fixed these cast issues :laughing: I can remove the `implicit-casts:false` and its fixes and enable it in a separated PR for the linter if that's better

kuhnroyal:
Yes, we should have a separate PR for the strong-mode/implicit changes. Can you use the same options I used here https://github.com/OpenAPITools/openapi-generator/pull/8231
I just checked out this branch and tried with the `analysis_options` from my branch and all the `json as Foo` casts are somehow unnecessary. So I would really love to see your analyzer options.

agilob:
No probs, so just add the analysis.yaml so we know what the changes are related to.

Grohden:
@kuhnroyal  my analysis options for the petstore is just:
```yaml
analyzer:
  strong-mode:
    implicit-casts: false
```

kuhnroyal:
Ok so I see the problem now. I never disable implicit casts because that doesn't do much good.
Assigning `dynamic` to a `String` is just as safe as casting it, if it is not a `String` it will fail either way.

On the other hand having `implicit-dynamic: false` actually prevents type errors.
After realizing we are talking about different settings I have to say I am against `implicit-casts: false`.

modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -133,17 +133,15 @@ class {{{classname}}} {
   	          {{/items.complexType}}
           {{/items.isArray}}
           {{^items.isArray}}
-        {{{name}}}: json[r'{{{baseName}}}'] == null
-          ? null
-          : {{{complexType}}}.mapFromJson(json[r'{{{baseName}}}']),
+        {{{name}}}: json[r'{{{baseName}}}'] as {{{datatype}}},
```
kuhnroyal, line 136: Are we sure that this can only be a primitive type here?

Grohden: This is the map of maps case...

I don't see this case in the petstore, but my project had this case:
image

I'm gonna change this line to the same as the list treatment (above)
```mustache
          {{#items.isMap}}
            {{{name}}}: json[r'{{{baseName}}}'] == null
              ? null
                {{#items.complexType}}
                    : {{items.complexType}}.mapFromJson(json[r'{{{baseName}}}']),
                {{/items.complexType}}
                {{^items.complexType}}
                    : (json[r'{{{baseName}}}'] as Map).cast<String, Map>(),
                {{/items.complexType}}
          {{/items.isMap}}
          {{^items.isMap}}
              {{{name}}}: json[r'{{{baseName}}}'] // primitive? Probably Map<String, primitive>
          {{/items.isMap}}
```

This is what it will generate:
image

Its that ok?

Grohden:
@kuhnroyal @agilob 
I've removed the fixes related to the `implicit-casts: false`, I've also changed the way I deal with the `Maps of maps` case

> After realizing we are talking about different settings I have to say I am against implicit-casts: false.

I've only learned from your comments about the `implicit-dynamic: false`, so agree with you and my recent changes are respecting this flag, but I'm gonna open another PR for the analysis file with it.

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
Grohden: I'm gonna keep it as a List, ideally we would use iterable and not rely on mutability (as it may affect code outside of this generated lib).

I'm gonna try to open another PR to make it accept a iterable and also rely on pure functions instead of mutating the list

agilob (changes requested):
Can you add the analysis yaml file pls?

Grohden:
@agilob 
U mean the one from #8231 or
one with
```
analyzer:
  strong-mode:
    implicit-casts: true
```
(which is to **explicitly accept** `implicit-casts`)
?

agilob:
The one that triggers warnings you fixed here

Grohden:
@agilob file added

Grohden:
@kuhnroyal I was bugged about why I had analyzer errors reported on the `_updateParamsForAuth` from the `invokeAPI`, it turns out that it was the `implicit-casts: false` that was reporting that error to me.

So just so you know, this code exemplifies a case where this flag can be helpful:
```dart
void main() {
  foo({"a", "b"});
}

void foo(Iterable<String> authNames) {
  // implicit-casts: false -> The argument type 'Iterable<String>' can't be assigned to the parameter type 'List<String>'.
  // implicit-casts: true -> ok, just runtime error :D
  bar(authNames); 
}

void bar(List<String> authNames) {
  authNames.add("bar");
}
```

agilob:
>>  // implicit-casts: false -> The argument type 'Iterable<String>' can't be assigned to the parameter type 'List<String>'.

Yea, that's a good fix. I'll address the bigger issue some time later, it shouldn't be List there :(

kuhnroyal:
I agree that this is one of the few cases were `implicit-cast: false` would highlight the potential bug. The question is, is the overhead in all the other places worth it? I am not sure. What do you guys think?

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
```diff
@@ -209,12 +209,12 @@ class ApiClient {
     List<QueryParam> queryParams,
     Map<String, String> headerParams,
   ) {
-    authNames.forEach((authName) {
+    for(final authName in authNames) {
```
kuhnroyal, line 212: Missing space

modules/openapi-generator/src/main/resources/dart2/enum.mustache
```diff
@@ -78,4 +78,4 @@ class {{{classname}}}TypeTransformer {
 
   /// Singleton [{{{classname}}}TypeTransformer] instance.
   static {{{classname}}}TypeTransformer _instance;
-}
\ No newline at end of file
+}
```
kuhnroyal, line 81: Remove the newline

modules/openapi-generator/src/main/resources/dart2/enum_inline.mustache
```diff
@@ -78,4 +78,4 @@ class {{{classname}}}{{{enumName}}}TypeTransformer {
 
   /// Singleton [{{{classname}}}{{{enumName}}}TypeTransformer] instance.
   static {{{classname}}}{{{enumName}}}TypeTransformer _instance;
-}
\ No newline at end of file
+}
```
kuhnroyal, line 81: Remove the newline

modules/openapi-generator/src/main/resources/dart2/class.mustache
kuhnroyal: I understand this change but in any case we don't care about wether it is empty. `foreach` doesn't run if it is empty.
Maybe something like this but thats just my preference, up  to you.
```dart
return json?.map((key, dynamic value) => MapEntry(key, {{{classname}}}.listFromJson(value, emptyIsNull: emptyIsNull, growable: growable,),) ?? <String, {{{classname}}}>{};
```

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
kuhnroyal: Is this external API? If so, this is a breaking change, 5.0.0 was just released. But it also fixes a bug so I am not sure what the process is here. @agilob You gonna fix this before the next release? :)

kuhnroyal (changes requested):

modules/openapi-generator/src/main/resources/dart2/class.mustache
kuhnroyal: Ideally you could change this in another PR as it is probably related to some analyzer rule.

Grohden:
@kuhnroyal I am the "if the compiler say its ok, then its ok" guy, so after seeing the flag capturing an actual bug I'm not against it - I'm gonna keep using it in personal projects. 

But I agree that having to cast every single dynamic is annoying and unnecessary, and the dart API that we have is not huge and is not expected to change very often.. so I would keep the flag (explicitly) disabled

agilob:
>would highlight the potential bug. The question is, is the overhead in all the other places worth it? I am not sure. What do you guys think?

Let's keep what we have here and address linter in another PR. I think it's fair to add the strictest linter in the generated code. I can do it after this gets in.

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
agilob: where is the space missing? It's all good.

modules/openapi-generator/src/main/resources/dart2/enum.mustache
agilob: It's all good, this is autogenerated code, shouldn't even be committed to version control, there are plenty of generators that leave +5 empty whitespaces at end of files.

agilob:
>But I agree that having to cast every single dynamic is annoying and unnecessary, and the dart API that we have is not huge and is not expected to change very often.. so I would keep the flag (explicitly) disabled

In case of autogenerated code I think correctness and type-safety is more important than "I dont like it has 2 casts every 5 lines in files I never modify manually"

agilob (approved):

kuhnroyal:
I agree, since it is generated we can add the casts. But they need to be the correct ones and that is not always easy to see.
For the bug that was found, most people would have just blindly added the cast `as List` to make it compile. It would still fail with a `Set` but the compiler is happy.

modules/openapi-generator/src/main/resources/dart2/enum.mustache
Grohden: These new lines were added by AS, but I'm with @agilob on this one.. I always apply dartfmt on the generated client so things like empty lines are removed.. but when we add the effective dart guide the generated code will need to be properly formatted when generated isn't?

kuhnroyal: The problem is that it will change again for the next PR and the next and the next. That is lots of noise that doesn't actually affect the commit. And both enum templates are actually inlined so you suddenly end up with duplicate/triple empty lines between classes. But whatever, up to you.

Grohden: would be nice to properly format the **mustaches** templates and leave the dart fonts reformat for the user.. I find very confusing to read if else conditions on the current mustache templates

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
kuhnroyal: `for (`. Whatever, we need to add post-processing to the generated examples so we have diffs that don't have this problem.

kuhnroyal (approved):

wing328:
@Grohden can you please resolve the merge conflicts when you've time? Thank you.

Grohden:
@wing328 I've merged the latest master today and regenerated the needed files..

I probably need to test if the new added anaysis_options.yaml file will affect the new (?) json serializable option... but that's gonna take time if it depends on me

agilob:
>I probably need to test if the new added anaysis_options.yaml file will affect the new (?) json serializable option...

I can look into it after this one is meged

wing328:
@Grohden sorry do you mind resolving the merge conflicts again? We'll put other Dart PRs on hold for the time being.

Grohden:
@wing328 no problem

I've merged today's master and re-generated all dart codes (`./bin/generate-samples.sh ./bin/configs/dart*`)

wing328:
Travis CI failure not related to this PR.

Issue #8254, opened 2020-12-22
[BUG][DART-DIO] break in for loop in 'ApiKeyAuthInterceptor'


vancegillies:
Hi,

Sorry for not following the bug report template but the issue is so simple and can probably be answered without having to have a repro.

There is a break statement in '/modules/openapi-generator/src/main/resources/dart-dio/auth/api_key_auth.mustache'

```dart
for (var info in authInfo) {
    final authName = info["name"];
    final authKeyName = info["keyName"];
    final authWhere = info["where"];
    final apiKey = apiKeys[authName];
    if (apiKey != null) {
        if (authWhere == 'query') {
            options.queryParameters[authKeyName] = apiKey;
        } else {
            options.headers[authKeyName] = apiKey;
        }
        break; // <- THIS
    }
}
```

A break in a for loop stops lopping, so if there is more then one apiKey only the first is added to the headers

I think 'continue' was intended? but also unneeded as its at the end of the loop anyway.

These are the headers i'm using 
```yaml
securitySchemes:
  device_id:
    type: apiKey
    in: header
    name: X-DEVICE-ID
  app_id:
    type: apiKey
    in: header
    name: X-APP-ID
  session_token:
    type: apiKey
    in: header
    name: Authorization
```
Only the app-id was being added to the headers, but removing the break statement fixed my issue.

Unless I'm missing something this seems like a bug

Again sorry not using a the template.

Thanks

kuhnroyal:
I think noone thought of having multiple API-keys and the break is just case of wrong optimization.
Can you add a PR for this?
As a workaround you need to manually add a custom `AuthInterceptor`.

vancegillies:
Happy to do a PR later today, if not, after Christmas. Yes my workaround was just removing that break and it then works as expected

PR #8274, opened 2020-12-26
[DART-DIO] Remove break statement in api auth key for loop to allow more than one api key

vancegillies:
Fixes bug from #8254


agilob:
You removed PR template which had list of steps to do to fix testing pipeline.

kuhnroyal:
Can you please close this one and create a new PR with the template and follow the steps.
Especially the sample regeneration and CCing the technical committee, otherwise this will just get lost and the build will fail.

Issue #8300, opened 2020-12-30
[BUG][Python] Generated python client fails to correctly process optional response bodies


telday:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

Attempting to use optional request bodies in the Python generated client, then passing nothing in for the body kwarg, causes the client to crash and output:

```
Reason: Cannot prepare a request message for provided
                             arguments. Please check that your arguments match
                             declared content type.
```

This is because in the rest client the if statement choosing how to format the request body fails to account for a body = `None`

##### openapi-generator version

4.3.1 However the offending portion of the rest client generates identically in the latest version.

##### OpenAPI declaration file content or url

```yaml
paths:
  '/test/':
    post:
      operationId: 'testing'
      requestBody:
        required: false
        content:
          text/plain:
            schema:
              type: string
      responses:
        "200":
          description: "The secret value was added successfully"
```

##### Steps to reproduce

Create a spec file with the above path in it, generate a python client, then attempt to make a request without passing the body kwarg

spacether:
Please upgrade to v5.0.0 and use the python genertor and set nullable to True for your response body schema.
Nullable schemas are supported in v5.0.0 python.

spacether:
If you need this behavior, please use the `python-experimental` generator
This feature exists in that generator
One can see that option request bodies are only included if they are set per this code: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python-experimental/petstore_api/paths/fake_refs_number/post.py#L71

PR #8301, opened 2020-12-30
Changed generated python client to accept None as request body

telday:
This allows optional request bodies to work properly, previously they would always fail when body was None.

Resolves: #8300

Signed-off-by: Ellis Wright <ellis.wright@cyberark.com>

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/python/api.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api.mustache
@@ -123,6 +123,12 @@ class {{classname}}(object):
             kwargs['{{paramName}}'] = \
                 {{paramName}}
 {{/requiredParams}}
+
+            if 'body' not in kwargs:
+                body = 'body' in self.params_map['all']
+                body_nullable = 'body' in self.params_map['nullable']
+                if body and not body_nullable:
+                    kwargs['body'] = ''
             return self.call_with_http_info(**kwargs)
 
         self.{{operationId}} = Endpoint(
--- a/modules/openapi-generator/src/main/resources/python/rest.mustache
+++ b/modules/openapi-generator/src/main/resources/python/rest.mustache
@@ -170,7 +170,7 @@ class RESTClientObject(object):
                 # Pass a `string` parameter directly in the body to support
                 # other content types than Json when `body` argument is
                 # provided in serialized form
-                elif isinstance(body, str) or isinstance(body, bytes):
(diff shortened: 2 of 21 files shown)


spacether:
Hey there. So I am concerned that we are not coving all of our use cases here.
- Required can be True or False
- Body schema can be nullable or not nullable
So sending None/null as a body is valid and should be different than omitting the body parameter.
How do you feel about reworking our code to cover these two use cases?
1) send + receive None nullable body (required= True or False, nullable=True)
2) send and receive request with body omitted (required=False)

Some ways to do this would be to:
- extract body from kwargs in the method `request`
- set the body default value to a unset_value class parameter:
```
class RESTClientObject:
    unset_value = object()
    def request(self, method, url, query_params=None, headers=None,
        body=unset_value, post_params=None, _preload_content=True,
        _request_timeout=None):
```

telday:
@spacether Just added another chunk of code which checks if the body is nullable then converts the value to an empty string if it is not. I struggled with where to put this logic, let me know what you think.

modules/openapi-generator/src/main/resources/python/api.mustache
```diff
@@ -123,6 +123,12 @@ class {{classname}}(object):
             kwargs['{{paramName}}'] = \
                 {{paramName}}
 {{/requiredParams}}
+
+            if 'body' not in kwargs:
+                body = 'body' in self.params_map['all']
+                body_nullable = 'body' in self.params_map['nullable']
+                if body and not body_nullable:
+                    kwargs['body'] = ''
```
spacether, line 131: Why are we setting body to empty string here?

spacether:
@telday thank you for making those changes. It's not clear to me if this code will code the use cases that I asked for.
Can you add tests showing it working?
The cases that will prove that this is working are:
- required body parameter, body nullable, show transmit and receiving none working
- optional body parameter, body nullable, show transmit and receiving none working

One can edit the sample spec file to add those endpoints here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/python/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml
and the apple schema is nullable and could be used for these endpoints.

For the tests, one could add them to the tag fake_api and add a test like we have here:
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/tests_manual/test_fake_api.py#L212

PR #8308, opened 2021-01-02
Dart - (de)serialize Enum from map

agilob:
Testing something, might be a bug in generator

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart2/class.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/class.mustache
@@ -143,7 +143,7 @@ class {{{classname}}} {
         {{{name}}}: null, // No support for decoding binary content from JSON
             {{/isBinary}}
             {{^isBinary}}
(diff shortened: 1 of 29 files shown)


modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -143,15 +143,22 @@ class {{{classname}}} {
...
+            {{#isEnum}}
+        enum
+        {{{name}}}: _$enumDecode(_${{{complexType}}}EnumMap, json[r'{{{baseName}}}']),
+            {{/isEnum}}
+            {{^isEnum}}
+        not enum
         {{{name}}}: {{{complexType}}}.fromJson(json[r'{{{baseName}}}']),
+            {{/isEnum}}
             {{/isBinary}}
         {{/isMap}}
```
kuhnroyal, line 155: You are in a `isMap` block, I don't think `Map<String, SomeEnum>` is possible.

PR #8310, opened 2021-01-03
Dart: use JSON-serializable

agilob:
@noordawod @kuhnroyal I attempted to use json_serializable to replace all our manual code. Here's almost finished work, but unfortunately I found a bug in the library

https://github.com/google/json_serializable.dart/issues/784 - google doesn't seem to care too much about handling enums in this library so looks like it won't go any further, unless you have any idea how to fix it manually? Generate _$decode and _$encode on our own?

Here's another attempt to generate enum handling code to produce the same autogenerated code as json_serializer https://github.com/OpenAPITools/openapi-generator/pull/8308 here I hit another bug where complexItem / enum isn't recognised correctly what you can see in [second commit](https://github.com/OpenAPITools/openapi-generator/pull/8308/commits/f8a53f839be451f8a35d1f0d6500513c30c2cb72)

The general direction is to remove our manual work and relay on some library, but I can't find any where enums aren't 3rd class citizens.

Any advice is welcome.

// this is PoC


noordawod:
Just wondering, why rely on another library to do this? At least in my case, I further adjust the templates for my own projects depending on the needs. Having another layer to worry about (and have no direct control over) is problematic, for me at least.

agilob:
To reduce complexity of code we maintain here. It's still far from good, for me it's still unusable, especially bad when it comes to enums, default values and error handling on invalid input (number as bool, string as number etc).

noordawod:
Well, I don't know... I'm fine with the existing template (which I don't find extremely complicated) so I'm neutral here.

agilob:
>I'm fine with the existing template

how many models do you have and how complex are they? I'm working with +30 models,  3 top level models, most children are interchangeable so nullability and configuration of fields is important. I use it in TS and java without problems, but in dart it's simply easier to reproduce models manually, generated code is still unusable for me. I can use only 4 classes from all specified.

noordawod:
Let's not compare the sizes of our d*cks here :) I don't have anything concrete against using it, so if you can get it to work then I'm fine with it...

agilob:
was just asking out of curiosity, how you use the generator

noordawod:
Currently using it in 3 projects (1 at work, 2 at home) and in one of the latter I'm using 4 templates (TS, Dart, Java and Kotlin). I'm using another generator for the db schemas as well, so you can say that I'm well accustomed to writing and wiring up templates.

As projects evolve few changes are required, so I prefer to patch upstream when possible (hence why I changed so much in the Dart template), but when not possible I do it locally. I just don't have an experience with the 3rd party library you're trying to integrate, and cannot anticipate at this time what kind of impact it would have on me -- for both professional and personal projects.

With that being said, I'm all up for changes so that's why I'm neutral.

So now, why did you decide to go on this endeavor in the first place?

kuhnroyal:
I like the idea of removing the manual `toJson`/`fromJson` as the most of the remaining serialization problems are located there. Other than that, I don't really have an opinion as I don't use it.

kuhnroyal:
You could try keeping the enums as they are, json_serializable should work with manual `to/fromJson` for parts of the structure.

kuhnroyal:
@agilob When you have time, can you try to just keep the manual `toJson/fromJson` for the enums and have the rest generated with `json_serializable` this should work I think. We probably need some test for this but I think we should head this way.

k-paxian:
> The general direction is to remove our manual work and relay on some library, but I can't find any where enums aren't 3rd class citizens.

What do you say about this https://github.com/k-paxian/dart-json-mapper ?

agilob:
I see it supports default values on fields, but no mention of required and nullable or not-null?

k-paxian:
Just let me know the requirements and we will make it happen.

agilob:
Thanks mate, I opened 3 new issues to keep conversation in their own topics https://github.com/k-paxian/dart-json-mapper/issues/108 https://github.com/k-paxian/dart-json-mapper/issues/110 https://github.com/k-paxian/dart-json-mapper/issues/109

agilob:
@k-paxian would you be available for a chat on matrix/gitter or slack? Where can I find you?

I'm trying to fit dart-json-mapper into openapi but either I'm getting infinite loops from build_runner or crashes "before?" start

```
Launching lib/main.dart on SM T713 in debug mode...
                                                                        
* What went wrong:                                                      
Could not initialize class org.codehaus.groovy.runtime.InvokerHelper    
                                                                        
* Try:                                                                  
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
                                                                        
* Get more help at https://help.gradle.org
````

[Documentation](https://pub.dev/packages/dart_json_mapper#basic-setup) also says to run method `initializeJsonMapper()` on start in main() method, which might be quite difficult to force openapi users to remember to add this to all projects, it would create bi-directional relationship between my_package and openapi_generated_code package.

If possible I would like to move this conversation to a chat or email, what's best way to reach you?

k-paxian:
I'm in openapi-generator slack, put a private message to you. @agilob

PR #8331, opened 2021-01-05
[WIP][csharp-netcore] Add "httpclient" library support

wing328:
Add support for HttpClient library

### PR checklist
 

devhl-labs:
So I should carry on with what I was doing and later do pr's to your new branch of csharp-netcore-httpclient? The files at `/csharp-netcore-httpclient/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient` are those the only files that I can be library specific, or can I just add the templates that are not present yet?

wing328:
@devhl-labs yes please work on the `csharp-netcore-httpclient` branch.

> are those the only files that I can be library specific, or can I just add the templates that are not present yet?

Feel free to add more. Those are the one I found in your PRs. If the generator cannot find the mustache templates in `modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient`, it will use the default in `modules/openapi-generator/src/main/resources/csharp-netcore/`

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/api.mustache
```diff
@@ -0,0 +1,704 @@
...
+        /// <summary>
+        /// {{summary}}
+        /// </summary>
+        /// <remarks>
+        /// {{notes}}
+        /// </remarks>
+        /// <exception cref="{{packageName}}.Client.ApiException">Thrown when fails to make API call</exception>
+        {{#allParams}}/// <param name="{{paramName}}">{{description}}{{^required}} (optional{{#defaultValue}}, default to {{.}}{{/defaultValue}}){{/required}}</param>
+        {{/allParams}}/// <returns>ApiResponse of {{#returnType}}{{returnType}}{{/returnType}}{{^returnType}}Object(void){{/returnType}}</returns>
+        ApiResponse<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}Object{{/returnType}}> {{operationId}}Response({{#allParams}}{{{dataType}}} {{paramName}}{{^required}}{{#optionalMethodArgument}} = default({{{dataType}}}){{/optionalMethodArgument}}{{/required}}{{^-last}}, {{/-last}}{{/allParams}});
```
wing328, line 46: @devhl-labs I noticed that you've renamed `WithHttpInfo` with `Response`. I'm not against a different naming for methods that return HTTP headers, status code, etc but if the method name already ends with `response`, the method name may end up like `getResponseResponse`.

We may need to add an option to customize this instead.

wing328:
@devhl-labs for other generators (e..g Java, Ruby), we only updated the ApiClient.mustache file with a different Http library most of the time (In other words, the ApiClient is swappable with a different implementation of HTTP library). For other templates (e.g. csproj files that list out the dependencies), we can use `{{#isHttpClient}} ... {{/isHttpClient}}` mustache tag instead to avoid maintaining too many customized files under `modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient`

wing328:
@devhl-labs when you've time, can you please PM me via Slack to have quick chat on this?

devhl-labs:
@wing328 can you merge this and also 8465? There is still work to be done but im not doing it all, so might as well release it with partial support.

Blackclaws:
Just tested it out. 

Found a weird reference to CocApi in the generated Client which seems superfluous.

Also it uses a couple of C#9 language features which makes the generated output unnecessarily incompatible with netstandard 2.1. It should at least allow a fallback to that, as one of the requested and one of our usecases as well is usage of generated code within Unity which is incompatible with Net 5.0.

If you want I can open a pullrequest with changes on this branch. Let me know!

devhl-labs:
I'll remove the unrelated api reference. I think the only C# 8 feature you're referring to is the nullable reference types, which we can fix with a small change to the post processing powershell.

How can I create only samples for my library so I don't have conflicts with unrelated libraries? I currently see a conflict with typescript which I did not touch.

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/api.mustache
```diff
@@ -0,0 +1,325 @@
...
+
+            {{#produces}}
+            request.Headers.Accept.Add(new System.Net.Http.Headers.MediaTypeWithQualityHeaderValue("{{{mediaType}}}"));
+            {{/produces}}
+
+            using System.Net.Http.HttpResponseMessage responseMessage = await _httpClient.SendAsync(request, cancellationToken.GetValueOrDefault()).ConfigureAwait(false);
+
+            string responseContent = await responseMessage.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+            ApiResponse<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}Object{{/returnType}}> apiResponse = new(responseMessage, responseContent);
```
Blackclaws, line 315: This is C# 9 syntax.

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/netcore_project.mustache
```diff
@@ -0,0 +1,39 @@
...
+    {{#useCompareNetObjects}}
+    <PackageReference Include="CompareNETObjects" Version="4.61.0" />
+    {{/useCompareNetObjects}}
+    <PackageReference Include="JsonSubTypes" Version="1.8.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <!--<PackageReference Include="RestSharp" Version="106.11.4" />
+    <PackageReference Include="Polly" Version="7.2.0" />-->
+    <PackageReference Include="Microsoft.Extensions.Http" Version="5.0.0" />    
+    {{#validatable}}
+    <!--<PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />-->
```
Blackclaws, line 35: This reference is needed for netstandard

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/netcore_project.mustache
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <!--<TargetFramework>{{targetFramework}}</TargetFramework>--><TargetFramework>net5.0</TargetFramework>
```
Blackclaws, line 5: Target Framework should not be hardcoded

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/api.mustache
```diff
@@ -0,0 +1,325 @@
...
+        /// <remarks>
+        /// {{notes}}
+        /// </remarks>
+        /// <exception cref="{{packageName}}.Client.ApiException">Thrown when fails to make API call</exception>
+        {{#allParams}}
+        /// <param name="{{paramName}}">{{description}}{{^required}} (optional{{#defaultValue}}, default to {{.}}{{/defaultValue}}){{/required}}</param>
+        {{/allParams}}
+        /// <param name="cancellationToken">Cancellation Token to cancel the request.</param>
+        /// <returns>Task of ApiResponse{{#returnType}} ({{returnType}}){{/returnType}}</returns>
+        System.Threading.Tasks.Task<ApiResponse<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}Object{{/returnType}}>> {{operationId}}WithHttpInfoAsync({{#allParams}}{{#required}}{{{dataType}}} {{paramName}}{{^-last}}, {{/-last}}{{/required}}{{^required}}{{{dataType}}}??? {{paramName}} = null{{^-last}}, {{/-last}}{{/required}}{{/allParams}}{{#allParams.0}}, {{/allParams.0}}System.Threading.CancellationToken? cancellationToken = null);
```
Blackclaws, line 34: What are the ??? supposed to do? I think there are two superfluous ?. This is throwing errors for me at least

Blackclaws (changes requested):
I noted a couple of the issues that directly pop up when you change to netstandard and c# 8.

If you also go below C#8 you of course run into the nullable reference type issue.

The ??? mark issue is also present without changing to netstandard.

In general I think this might be better suited as a separate generator instead of a library for the normal csharp-netcore generator as it is quite a change from what the restsharp version is producing.

devhl-labs:
The ??? is a placeholder. It gets fixed in the post processing. Check the readme file to see a powershell script that both runs the jar and then executes the fix. The issue is that a reference type is declared by Type not Type? so i need some placeholder to mark the object as nullable.

Blackclaws:
> The ??? is a placeholder. It gets fixed in the post processing. Check the readme file to see a powershell script that both runs the jar and then executes the fix. The issue is that a reference type is declared by Type not Type? so i need some placeholder to mark the object as nullable.

Which Readme? Its not in the generated Readme.md. I think if there is post processing required especially via a tool like powershell I'm leaning more and more towards this being a separate generator. Powershell Core is included in current dotnet/sdk docker images so it might not be an issue for people using that in their pipeline, that might not be everyone though. I'd specifically mention that as a requirement and maybe even check whether the postprocessing couldn't be done within the generator itself if it is needed.

devhl-labs:
It's here at the top:

<https://github.com/devhl-labs/openapi-generator/tree/csharp-netcore-httpclient/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient>

What does a generator mean to you? We seem to have different definitions. I've already suggested in the slack channel that we restructure the csharp folders like below.  We have a bunch of csharp folders and it's very confusing which to use. Even one aspnet generator which you may not even see if you go looking for csharp options.

```
csharp
  client
    libraries(http, restsharp, etc)
      version
  server
    libraries(http, restsharp, etc)
      version
```

Blackclaws:
We might indeed have different definitions. Let me try to give mine:

A generator has a specific feature set that allows me to fine tune the output and supports a specific language. I expect a generator to be as compatible as possible between versions as I have code depending on the output.

It might also have to do with how we use generators in our workflow. We don't create a server implementation, export the openapi spec, then generate a client and just keep modifying that. We build the openapi spec and then create abstract server base classes and automatic client libraries via generation, those also aren't committed but are regenerated on CI runs, hence why we kind of rely on the fact that they don't break the exposed API surface on upgrading the generator (we could just stick to one version of the generator, but that also keeps us from getting bug fixes). 

I think your restructuring is indeed a good idea when we talk about the template folders. I don't think its possible at least not easily to merge the generator classes into one. That is due to the fact that the tasks they perform might be very different and they expose different command line switches. Having a switch that pertains to the server implementation exposed on the client generator is confusing.

lucamazzanti:
Is this still useful? Unfortunately it seems to have gone in a different direction since the latest changes.
Probably if there are contributions it could be useful to redo the PR from the current master and apply just the mustache parts.
Otherwise it can be closed.

devhl-labs:
Ill be working on this this weekend.

PR #8354, opened 2021-01-06
[Play Framework] Update the bean validation to use version 2.0.

JFCote:
For a reason I don't know, it was not working anymore with version 1 of the bean validation. I think it was related that the POJO was modified globally to use jakarta annotation and these doesn't work anymore with version 1. 

So instead of trying to fix it, I updated to version 2 which is the latest and used by many other generators. I tested it with my own very huge and complicated project and everything seems to work **but please review this carefully because I'm not a specialist of bean validation**.

### PR checklist
 
- [X] Run the shell script `./bin/generate-samples.sh`to update all Petstore samples related to your fix. This is important, as CI jobs will verify _all_ generator outputs of your HEAD commit as it would merge with master. These must match the expectations made by your contribution. You may regenerate an individual generator by passing the relevant config(s) as an argument to the script, for example `./bin/generate-samples.sh bin/configs/java*`. For Windows users, please run the script in [Git BASH](https://gitforwindows.org/).


samples/server/petstore/java-play-framework-api-package-override/app/apimodels/Pet.java
```diff
@@ -19,16 +19,20 @@
...
 
   @JsonProperty("photoUrls")
-  private List<String> photoUrls = new ArrayList<>();
+  @NotNull
+private List<String> photoUrls = new ArrayList<>();
 
   @JsonProperty("tags")
-  private List<Tag> tags = null;
+  @Valid
+private List<Tag> tags = null;
```
wing328, line 35: What about keeping the 2-space indentation?

JFCote: I tried but either I was adding an unnecessary empty line or removing the 2 space indentation.
I will give it another try!

wing328: I can give it a try tomorrow as I'm pretty "experience" in mustache spacing.

JFCote: I just tried like 10 different ways I could think of and there is always a problem... If you find anything, please let me know!

JFCote: The problem is that the `beanValidation` partial file only contains "ifs". Most of the time, none of the "ifs" is true so when mustache renders the partials, it just see the file `\n` which then creates an empty line for no reason. I searched for an option in the mustache library that could disable this but haven't found one so far.

wing328: Pushed https://github.com/OpenAPITools/openapi-generator/pull/8354/commits/05bbacd646e0b1a6b308b6e9200219c4d34e538c. Still have a line break after annotation.

To fix that, need to remove all new line in the bean validation mustache template.

What about showing a message at the end of code generation to advise users using a code formatter?

JFCote: @wing328 Not sure I understand your last sentence about the code formater. 

Meanwhile, I had an idea during the weekend. I need to test it. The problem we have with the "partials" `beanValidation` is because sometime it doesn't have anything and it then just adds the line break.

I was thinking of always having something at the end of the `beanValidation`, something like that: `**javaplayframeworkdummynode**`

Then, in the post processing, there must be a way to search for this string + `\n` and to delete it. I think it can work. That is a strange workaround but Mustache templating is strange too, so we must adapt ;)

I'll give it a try today or tomorrow.

wing328: > @wing328 Not sure I understand your last sentence about the code formater.

I mean recommending users to use something like https://github.com/jhipster/prettier-java to auto-format the code, which fixes all those indentation issues, extra spaces, etc in the output.

JFCote: @wing328 I checked other generator like Spring (on which Play Framework generator was based) and they have the same problem, even worse. See this for example: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/springboot-beanvalidation/src/main/java/org/openapitools/model/Animal.java#L42

So I think that with you fix, having only one `\n` is not that bad after all.

So if you agree, you can merge this PR.

wing328:
CI failure already fixed in the master.

PR #8369, opened 2021-01-07
[dart][dart-dio] Improve form param handling, respect required flag

kuhnroyal:
* simplify template
* respect required flag (only null check when not required)
* minor formatting

### PR checklist
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
@@ -52,26 +52,15 @@ class {{classname}} {
         ];{{/hasConsumes}}
         {{#hasFormParams}}
 
-        final formData = <String, dynamic>{};
+        final formData = <String, dynamic>{
+            {{#formParams}}
+            {{^required}}if ({{paramName}} != null) {{/required}}r'{{baseName}}': {{#isFile}}MultipartFile.fromBytes({{paramName}}, filename: r'{{baseName}}'){{/isFile}}{{^isFile}}parameterToString(_serializers, {{paramName}}){{/isFile}},
+            {{/formParams}}
+        };
         {{#isMultipart}}
-        {{#formParams}}
-        {{^isFile}}
-        if ({{paramName}} != null) {
-            formData[r'{{baseName}}'] = parameterToString(_serializers, {{paramName}});
-        }
-        {{/isFile}}
-        {{#isFile}}
-        if ({{paramName}} != null) {
-            formData[r'{{baseName}}'] = MultipartFile.fromBytes({{paramName}}, filename: r'{{baseName}}');
-        }
-        {{/isFile}}
-        {{/formParams}}
         bodyData = FormData.fromMap(formData);
         {{/isMultipart}}
         {{^isMultipart}}
-        {{#formParams}}
-        formData['{{baseName}}'] = parameterToString(_serializers, {{paramName}});
-        {{/formParams}}
         bodyData = formData;
         {{/isMultipart}}
         {{/hasFormParams}}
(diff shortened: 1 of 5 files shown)


josh-burton (approved):

samples/client/petstore/dart-dio/petstore_client_lib/lib/api/pet_api.dart
```diff
@@ -427,9 +427,10 @@ class PetApi {
             'application/x-www-form-urlencoded',
         ];
 
-        final formData = <String, dynamic>{};
-        formData['name'] = parameterToString(_serializers, name);
-        formData['status'] = parameterToString(_serializers, status);
+        final formData = <String, dynamic>{
+            if (name != null) r'name': parameterToString(_serializers, name),
```
agilob, line 431: I would put null-check inside `parameterToString` to remove repeating code, what's the effect of `formData['name'] = null` ?

samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/api/fake_api.dart
```diff
@@ -369,7 +369,7 @@ class FakeApi {
             'application/json',
         ];
 
-        var serializedBody = body;
+        final serializedBody = body;
```
agilob, line 372: is this copy even needed?

agilob (approved):

samples/openapi3/client/petstore/dart-dio/petstore_client_lib_fake/lib/api/fake_api.dart
kuhnroyal: We expect the serializedBody below.

samples/client/petstore/dart-dio/petstore_client_lib/lib/api/pet_api.dart
kuhnroyal: There is one but it defaults to `''`. I am unsure what the consequences are if I remove this.

kuhnroyal: Also the null-check is only done when the param is not `required`.

modules/openapi-generator/src/main/resources/dart-dio/api.mustache
```diff
@@ -52,26 +52,15 @@ class {{classname}} {
         ];{{/hasConsumes}}
         {{#hasFormParams}}
 
-        final formData = <String, dynamic>{};
+        final formData = <String, dynamic>{
+            {{#formParams}}
+            {{^required}}if ({{paramName}} != null) {{/required}}r'{{baseName}}': {{#isFile}}MultipartFile.fromBytes({{paramName}}, filename: r'{{baseName}}'){{/isFile}}{{^isFile}}parameterToString(_serializers, {{paramName}}){{/isFile}},
```
wing328, line 57: What about required fields that are nullable (`nullable: true`)? Do we need to handle it above?

kuhnroyal: Well these are formParams and not fields, I have never seen them `nullable` but since only required fields can be `nullable: true` they would currently be empty strings `''` if they are `null`.

wing328: optional fields can also have `nullable: true`, which means the field will explicitly emit `null`. For example in JSON:

```json
{
  "name": "something",
  "type": null
}
```
whereas `type` is defined as both optional and nullable.

kuhnroyal: Hmm ok, but if `type` is optional then how will it ever be not present in the output if it is `nullable`? If `type` is set to `null` it needs to be serialized in JSON, if it is anything else it also needs to be serialized in JSON - that basically means it is not optional?

In any case I didn't see anything about parameter objects being `nullable` in the spec, although `CodegenParameter` has a `nullable` flag, so I am not sure if this needs to be handled here.

kuhnroyal: Updated with an additional `{{^nullable}}` check.

Issue #7256, opened 2020-08-20
[BUG] Java oneof Classes are not generated


neopaf:
##### Description

(Thanks for fantastic tool!)

I expect OneOfGeographicLocationGeographicLocationRef.java be created.

Actual: no such file is generated

##### openapi-generator version

Tried with
* 4.3.1
* openapi-generator-cli-5.0.0-20200813.162255-527.jar

##### OpenAPI declaration file content or url

[Swagger.json.txt](https://github.com/OpenAPITools/openapi-generator/files/5102449/Swagger.json.txt)

##### Generation Details

```bash
openapi-generator  generate -I Swagger.json.txt -g jaxrs-cxf-client 
```

##### Steps to reproduce

Run above command, you'll get
src/gen/java/org/openapitools/model/GeographicLocation.java 
which contains
```java
import org.openapitools.model.OneOfGeographicLocationGeographicLocationRef;
```
But no such class was generated.

kito99:
Just ran into this as well with the same version. Can we get this prioritized for the next release?

neopaf:
@kito99 Kito, my sympathy... 
For now I've just handmade missing files... just so I could still use generated code.
But I'm not happy :(

wing328:
Please try the `java` (library: jersey2) client generator in the latest master, which offers better support for oneOf, anyOf.

seybi87:
> @kito99 Kito, my sympathy...
> For now I've just handmade missing files... just so I could still use generated code.
> But I'm not happy :(

@neopaf I am also facing this issue :-(

As you have resolved these problems by manually creating the required classes, do you have a MWE example with the created classes online? 
This would be much appreciated 👍

neopaf:
I did just a trivial hack, not something that works as client/server.

I need code for investigation purposes only...

sarumont:
@seybi87 I just got back to this today and need a compiling JAR. I *think* this will work (about to test it, and I'll report back here):

```java
package com.devsupply.dana.model;

import java.util.Objects;
import com.fasterxml.jackson.annotation.JsonProperty;

/**
 * OrgAllOf
 */
public class OneOfUserApiKey {
  private User user;
  private ApiKey apiKey;

  public void setUser(User user) {
    user(user);
  }

  public void setApiKey(ApiKey apiKey) {
    apiKey(apiKey);
  }

  public OneOfUserApiKey user(User user) {
    this.user = user;
    this.apiKey = null;
    return this;
  }

[...]

Issue #5903, opened 2020-04-13
[BUG][java] oneOf in `responses`: interface not generated.


jfeltesse-mdsol:
##### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used?
- [x] Have you search for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

As mentioned in #5706, if the openapi spec file defines a oneOf schema in responses "as is", the generator references a model in the DefaultApi but the model doesn't get generated. See examples below.

##### openapi-generator version

latest master: cc09118ffc2f13ca5d251049dc3d7552f08b670a

##### OpenAPI declaration file content or url

OpenApi spec file:
```yaml
openapi: 3.0.3
info:
  title: oneOf test
  version: '1.0'
paths:
  /state:
    get:
      operationId: getState
      responses:
        '200':
          description: get states
          content:
            application/json:
              schema: # doesn't work
                oneOf:
                  - $ref: '#/components/schemas/ObjA'
                  - $ref: '#/components/schemas/ObjB'
                discriminator:
                  propertyName: realtype
  /state_ref:
    get:
      operationId: getStateRef
      responses:
        '200':
          description: 'get states using $ref'
          content:
            application/json:
              schema: # logically same as above but works
                $ref: '#/components/schemas/ObjAOrObjB'
components:
  schemas:
    ObjA:
      type: object
      properties:
        realtype:
          type: string
        message:
          type: string
    ObjB:
      type: object
      properties:
        realtype:
          type: string
        code:
          type: integer
    ObjAOrObjB:
      oneOf:
        - $ref: '#/components/schemas/ObjA'
        - $ref: '#/components/schemas/ObjB'
      discriminator:
        propertyName: realtype
```

##### Command line used for generation

`openapi-generator generate -i spec.yaml -g java --library native -o ./out/java_oneof`

##### Steps to reproduce

- use spec file above
- run command above
- notice how `src/main/java/org/openapitools/client/api/DefaultApi.java` features
  ```
  import org.openapitools.client.model.GetState200;
[...]

Issue #2892, opened 2019-05-14
[BUG] [JAVA] UNKNOWN BASE TYPE when using allOf in request body schema declaration


Emdee89:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] What's the version of OpenAPI Generator used?
- [X] Have you search for related issues/PRs?
- [X] What's the actual output vs expected output?

##### Description

Code generation spills out:
```
[WARNING] allOf with multiple schemas defined. Using only the first one: SomewhatIn1. To fully utilize allOf, please use $ref instead of inline schema definition
[WARNING] allOf with multiple schemas defined. Using only the first one: SomewhatIn1. To fully utilize allOf, please use $ref instead of inline schema definition
[WARNING] allOf with multiple schemas defined. Using only the first one: SomewhatIn1. To fully utilize allOf, please use $ref instead of inline schema definition
[WARNING] allOf with multiple schemas defined. Using only the first one: SomewhatIn1. To fully utilize allOf, please use $ref instead of inline schema definition
[WARNING] The following schema has undefined (null) baseType. It could be due to form parameter defined in OpenAPI v2 spec with incorrect consumes. A correct 'consumes' for form parameters should be 'application/x-www-form-urlencoded' or 'multipart/form-data'
[WARNING] schema: class ComposedSchema {
    class Schema {
        type: null
		...
		
[WARNING] codegenModel is null. Default to UNKNOWN_BASE_TYPE

Thereafter, we encounter compilation failures:
api/SomewhatApi.java:[7,71] cannot find symbol
  symbol:   class UNKNOWN_BASE_TYPE
```

##### openapi-generator version

- 3.3.4
- 4.0.0

Both versions are affected.

##### OpenAPI declaration file content or url

Gist URL:
https://gist.github.com/Emdee89/b3bcab67c46fd6c49d9282a528313990

##### Command line used for generation

Maven Plugin execution:

```
<plugin>
        <groupId>org.openapitools</groupId>
        <artifactId>openapi-generator-maven-plugin</artifactId>
        <version>4.0.0</version>
        <executions>
          <execution>
            <goals>
              <goal>generate</goal>
            </goals>
            <configuration>
              <inputSpec>${project.basedir}/src/main/resources/spec.yaml</inputSpec>
              <output>${project.basedir}</output>
[...]

PR #8423, opened 2021-01-12
[Generator][WIP]Extract schema name for requestBody when using oneOf

hugo-ma-alves:
Hello

I was having some difficulties generating the java (feign) client when the requestBody included the oneOf directive.  Let me try to illustrate with an example using the following spec.

```yaml
openapi: 3.0.1
info:
  title: Sample OneOf
  description: Sample OneOf
  version: v1
servers:
  - url: 'http://localhost:8080'
    description: Generated server url
paths:
  /pet:
     post:
      tags:
        - creation
      summary: Create a new Pet
      operationId: create
      parameters:
        - name: Accept-Language
          in: header
          description: Language
          schema:
            type: string
            enum:
              - FR
              - DE
              - EN
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
              - $ref: '#/components/schemas/Lizard'
              x-one-of-name: Pet
      responses:
        '201':
          description: The was created
components:
  schemas:
    Pet:
      type: object
      required:
      - petType
      properties:
        petType:
          type: string
      discriminator:
        propertyName: petType
        mapping:
          dog: Dog
          cat: Cat
          lizzard: Lizzard
    Cat:
      allOf:
      - type: object
        properties:
          name:
            type: string
    Dog:
      allOf:
      - type: object
        properties:
          bark:
            type: string
    Lizard:
      allOf:
      - type: object
        properties:
          lovesRocks:
            type: boolean
```

And to generate the client I use the following command:

```generate -i "sample.yml" -g java -o pet-client --additional-properties=library=feign```

So for the first try I used the code on the master branch and I got the following API class:

```java
public interface CreationApi extends ApiClient.Api {

  /**
   * Create a new Pet
   * 
   * @param acceptLanguage Language (optional)
   * @param UNKNOWN_BASE_TYPE  (optional)
   */
  @RequestLine("POST /pet")
  @Headers({
    "Content-Type: application/json",
    "Accept: application/json",
    "Accept-Language: {acceptLanguage}"
  })
  void create(@Param("acceptLanguage") String acceptLanguage, UNKNOWN_BASE_TYPE UNKNOWN_BASE_TYPE);
}
```

The generator couldn't extract the correct type that encapsulates the subtypes of the Pet, so it uses the default value "UNKNOWN_BASE_TYPE.
I tracked this value to the line https://github.com/OpenAPITools/openapi-generator/blob/49f55b0d8e3a9531e3962bd831b74181ac71abd8/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L5971. By default if it doesn't have another value it uses this one.
Not sure if it is related with this ticket https://github.com/OpenAPITools/openapi-generator/issues/7256

Then I tried to use the extension x-one-of-name to manually define the model name to "Pet". This type is correctly created as a Java class from where the subtypes extend. And I got the same result.

I searched the code and found the function that parses the requestBody: https://github.com/OpenAPITools/openapi-generator/blob/49f55b0d8e3a9531e3962bd831b74181ac71abd8/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L6022-L6025

It looks to me that it expects a $ref for the requestBody schema and it doesn't takes in consideration that it can contain a oneOf ( I believe it is called inline schema? not sure about the naming).  I then ran the generator with the change included in this PR and I got the following API.

```java
public interface CreationApi extends ApiClient.Api {
  /**
   * Create a new Pet
   * 
   * @param acceptLanguage Language (optional)
   * @param pet  (optional)
   */
  @RequestLine("POST /pet")
  @Headers({
    "Content-Type: application/json",
    "Accept: application/json",
    "Accept-Language: {acceptLanguage}"
  })
  void create(@Param("acceptLanguage") String acceptLanguage, Pet pet);
}
```

Now it uses the value of the extension as the Parameter name, and the client compiles correctly. 

I saw some other issues that may be related with this topic, https://github.com/OpenAPITools/openapi-generator/issues/2892 https://github.com/OpenAPITools/openapi-generator/issues/5903 https://github.com/OpenAPITools/openapi-generator/issues/7256
And I believe there is also other option where the OneOf classes should be generated right? In this case the class OneOfDogCatLizard. But I couldn't generate these classes, do I need to pass diferente parameters to the generator cli?

So to sum up, if this change is technically correct I can complete the PR by fixing the unit tests and updating the documentation

Hugo

### PR checklist
 
- [ ] Run 
  ```
  mvn clean package
  ./bin/generate-samples.sh
  ``` 
  to update all Petstore samples related to your fix. 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -6021,6 +6021,8 @@ public class DefaultCodegen implements CodegenConfig {
 
         if (StringUtils.isNotBlank(schema.get$ref())) {
             name = ModelUtils.getSimpleRef(schema.get$ref());
+        }else {
+            name = getSchemaType(schema);
         }
         schema = ModelUtils.getReferencedSchema(this.openAPI, schema);
 


kuhnroyal:
This does not work as soon as the type is not imported already in the API class or there is a model suffix/prefix configured.
After some investigation into the same problem, I now think that there is currently zero support for generating inline composed result/response/parameter in `InlineModelResolver`

hugo-ma-alves:
Thanks for the feedback @kuhnroyal
I'm interested in investigating this problem a bit more.
Do you have any example, open api spec or a sample, that I can use to replicate the situation you describe? it would be helpful  for starting.

hugo-ma-alves:
@wing328  Do you have any example to reproduce this problem?

nermin-imamovic:
[openapi.zip](https://github.com/OpenAPITools/openapi-generator/files/6296339/openapi.zip)

Can you test with this file. 

When we can expect to resolve this :)

wing328:
About the UNKNOWN_BASE_TYPE issue, it should be fixed in v6.0.0. Please pull the latest to give it another try.

PR #8431, opened 2021-01-13
[Ruby] Incorrect return types if multiple responses are defined (reopened)

plaurynovich-hubspot:
### Description

If you define multiple responses, the codegen will only support the first one (typically 200).

### Previous PR

https://github.com/OpenAPITools/openapi-generator/pull/7634


samples/client/petstore/ruby-faraday/lib/petstore/api/store_api.rb
```diff
@@ -114,20 +117,25 @@ def get_inventory_with_http_info(opts = {})
...
-      return_type = opts[:debug_return_type] || 'Hash<String, Integer>'
-
       # auth_names
       auth_names = opts[:debug_auth_names] || ['api_key']
 
+      # return_type
+      return_type = opts[:debug_return_type]
+
+      return_types_map = {
+        200 => “Hash&lt;String, Integer&gt;“,
```
wing328, line 127: Please use `{{{ ... }}}` instead of `{{ ... }}` to use the original value in the mustache tag.

PR #8472, opened 2021-01-19
[feature][python-flask] Add CORS support to python-flask server

ceresek:
Adds basic CORS support to the server generated with the `python-flask` generator (and the derived `python-aiohttp` and `python-blueplanet`). Enabled with `featureCORS` option (same as similar option in other generators). Disabled by default.

Mentioning TC members for Python as instructed in the PR checklist: @taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess @arun-nalla @spacether.


spacether:
@ceresek thank you for your PR. The feature generally looks good but it broke tests in circle and travis. Can you get them passing

ceresek:
@spacether Thank you for your comments. I fixed the Travis issue (forgot to update the samples and that was the check that failed), but I might need some advice with Circle. The parallel logs are a bit challenging to read, but the only serious failure I see is Maven failing to fetch a dependency from the Maven Central, with lots of "Connection reset" errors, and the same test builds and runs fine locally. What am I missing ?

spacether:
Thank you for your sample fix. It looks like circle balked when it tried to download a maven package. I just restarted it which may resolve it. Let's wait and see.

wing328:
All tests passed 👍

PR #8476, opened 2021-01-19
[swift5] Fix target SDKs for Combine option

### LINKED ISSUE #7728: [BUG][SWIFT5] Swift5 generator sets wrong platforms requirement in Package.swift
##### Description

Swift5 generates `Package.swift` a manifest file for Swift Package Manger. The manifest file contains a `platforms` directive that specifies the minimum iOS, macOS, tvOS, watchOS SDK version required to compile the generaged sources. Swift5 generator config options include a `responseAs` configuration that specifies what libraries the generated code should use. When `responseAs` is set to `Combine`, the generated code will use the Combine Framework, which is available on iOS 13, tvOS 13, macOS 10.15, watchOS 6 and above. The `platforms` directive in the `Package.swift` file does not reflect this requirement.

###### Expected `Package.swift`:

```
platforms: [
        .iOS(.v13),
        .macOS(.v10_11),
        .tvOS(.v9),
        .watchOS(.v3)
    ],
```

###### Actual `Package.swift`:
```
platforms: [
        .iOS(.v9),
        .macOS(.v10_11),
        .tvOS(.v9),
        .watchOS(.v3)
    ],
```

4.3.1

Gradle.build.kts:
```kotlin
openApiGenerate {
    validateSpec.set(true)
    generatorName.set("swift5")
    outputDir.set(openApiOutputDir.toString())
    inputSpec.set(file("openapi.yaml").toString())
    configOptions.set(mapOf(
            "nonPublicApi" to "false",
            "responseAs" to "Combine",
            "projectName" to "SpeedtestPlus",
            "swiftUseApiNamespace" to "false"
    ))
}
```

##### Steps to reproduce

Create a Gradle build file with the above configuration, use any openapi.yaml, and generate.

##### Suggest a fix

When `responseAs` is set to `Combine`, the `platforms` directive in `Package.swift` should be as follows:

```
platforms: [
        .iOS(.v13),
        .macOS(.v10_15),
        .tvOS(.v13),
        .watchOS(.v6)
    ],
```

fl034:
Fixes #7728 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


fl034:

4brunu:
Hi,

First of all, thanks for taking the time to create this PR. 🙂

Could you please describe your use case on why to make this change?

I would like to discuss this PR.

The methods that use `Combine` are already marked as `@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)`.
So this should already protect agains users of iOS 12 and bellow.

Is this not enough?

It should be possible to choose more than one `responseAs` at the same time, for example `Result` and `Combine`.
In this case, the minimum supported version of iOS should be 9, although it still uses `Combine`.

The current setup of mark the `Combine` methods with `@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)` should allow the users of the library to choose `Combine` in iOS 13 and above, and `Result` on iOS 12 and bellow.

One thing that maybe we could do is to change the minimum iOS version to 13 when `Combine` is the only selected `responseAs` option?

@fl034 What do you think about this?

4brunu:
Hi @fl034,

Looks like a good solution to me.

Are you thinking in adding `#canImport Combine` in the `import Combine` and also the methods that use `Combine?

4brunu:
Hi @fl034,

Thanks for making the changes that we talked about 👍 

I think that we should also wrap the methods that use `Combine` with  `#canImport Combine`.

What do you think about that?

fl034:
Man you're fast! I was just writing about it. 
I was thinking to replace the `@available` if `#canImport`, but this leads to errors.
And leaving it as it is and only wrapping the import statement also leads to errors.
So you're right, we need both.
I will update it soon, again.

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -8,7 +8,10 @@
 import Foundation{{#usePromiseKit}}
 import PromiseKit{{/usePromiseKit}}{{#useRxSwift}}
 import RxSwift{{/useRxSwift}}{{#useCombine}}
-import Combine{{/useCombine}}{{#swiftUseApiNamespace}}
+#if canImport(Combine)
+import Combine
+#endif
+{{/useCombine}}{{#swiftUseApiNamespace}}
```
4brunu, line 14: Could you please put `#endif{{/useCombine}}{{#swiftUseApiNamespace}}` in the same line to avoid generating an extra line?

4brunu (approved):
Looks good, I just left a small comment. 👍 
Thanks for your help!

modules/openapi-generator/src/main/resources/swift5/api.mustache
fl034: Jep :-)

fl034:
@4brunu I noticed that in the latest snapshot version, using URLSession as library, I don't get callbacks from my requests. It works with Alamofire as library. Also it works in 5.0.0 stable. 
Do you know if there's already an issue for that?

4brunu:
No, I didn't knew about that.
Do you have a sample project to show the issue?
Because I'm using the snapshot version with URLSession and it's working.

fl034:
@4brunu I opened an issue #8511 regarding my observation.
Since it's unrelated to this PR, you can merge this PR if you want :-).

4brunu:
I don't have permissions to merge the PRs.
I only help reviewing them. 🙂 
Let's just wait for the CI to finish.

PR #8498, opened 2021-01-21
Raise an error when passing None authentication values in Python

therve:
When using a client that gets value from the environment, it's very easy
to end up with None in api_key, which gives a pretty error about
bytes-like object down in the HTTP code. This raises an error earlier.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api_client.mustache
@@ -618,6 +618,8 @@ class ApiClient(object):
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
+                if auth_setting['value'] is None:
+                    raise ApiValueError('Invalid authentication parameter value')
                 if auth_setting['in'] == 'cookie':
                     headers['Cookie'] = auth_setting['value']
                 elif auth_setting['in'] == 'header':
--- a/modules/openapi-generator/src/main/resources/python/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/python/api_client.mustache
@@ -618,12 +618,12 @@ class ApiClient(object):
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
-                if auth_setting['value'] is None:
-                    raise ApiValueError('Invalid authentication parameter value')
                 if auth_setting['in'] == 'cookie':
                     headers['Cookie'] = auth_setting['value']
                 elif auth_setting['in'] == 'header':
                     if auth_setting['type'] != 'http-signature':
+                        if auth_setting['value'] is None:
+                            raise ApiValueError('Invalid authentication parameter value')
                         headers[auth_setting['key']] = auth_setting['value']
 {{#hasHttpSignatureMethods}}
                     else:
--- a/samples/client/petstore/python/petstore_api/api_client.py
+++ b/samples/client/petstore/python/petstore_api/api_client.py
@@ -606,12 +606,12 @@ class ApiClient(object):
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
-                if auth_setting['value'] is None:
-                    raise ApiValueError('Invalid authentication parameter value')
                 if auth_setting['in'] == 'cookie':
                     headers['Cookie'] = auth_setting['value']
                 elif auth_setting['in'] == 'header':
                     if auth_setting['type'] != 'http-signature':
+                        if auth_setting['value'] is None:
+                            raise ApiValueError('Invalid authentication parameter value')
                         headers[auth_setting['key']] = auth_setting['value']
                 elif auth_setting['in'] == 'query':
                     querys.append((auth_setting['key'], auth_setting['value']))
--- a/samples/client/petstore/python/petstore_api/api_client.py
+++ b/samples/client/petstore/python/petstore_api/api_client.py
@@ -606,6 +606,8 @@ class ApiClient(object):
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
+                if auth_setting['value'] is None:
+                    raise ApiValueError('Invalid authentication parameter value')
                 if auth_setting['in'] == 'cookie':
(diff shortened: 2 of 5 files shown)


therve:
@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess @arun-nalla @spacether 🙇

modules/openapi-generator/src/main/resources/python/api_client.mustache
```diff
@@ -622,6 +622,8 @@ class ApiClient(object):
                     headers['Cookie'] = auth_setting['value']
                 elif auth_setting['in'] == 'header':
                     if auth_setting['type'] != 'http-signature':
+                        if auth_setting['value'] is None:
+                            raise ApiValueError('Invalid authentication parameter value')
```
spacether, line 626: Multiple auth_settings can be passed in in auth_settings.
our use cases are:
1. qty 1 auth_settings and it is 'http-signature'
2. qty >= 2 auth_settings, one is 'http-signature', and both are required
3. qty >= 2 auth_settings, one is 'http-signature', and only one is required
As far as I know we don't have a way to differentiate between use case 2 + 3.
Are you assuming that we are in use case 1 or 2?
In mind use case 1 and 3 are much more likely and we should support them. What do you think?
How about only raising an exception if we are in use case 1?
What do you think about adding tests to demonstrate this behavior?
If we don't have tests then this feature could break in the future and we would have no visibility into it.

therve: I'm not assuming anything, but I'm not sure it matters? In this particular case I just prevent `None` being present in the headers, which will fail serialization later on. I think it's an issue regardless of the number of auth_settings, but maybe I'm missing something.

spacether: So your solution adds all present auth_settings onto the request. Which implies that all auth settings are required. This fix is only needed on the auth settings that we choose to send. If we are in use case 3 we have what we need and can choose to not send http-signatureaauth amd we willnnot hit the exception.

Hereare are some examples

1. qty 1 auth_settings and it is 'http-signature'
a. We have not set the auth info in  configuration, errors should be raised
b. We have set it, everything works

2. qty >= 2 auth_settings, one is 'http-signature', and both are required
a. We have set none or only one. We hot an exception.
b. All are included and everything works

3. qty >= 2 auth_settings, one is 'http-signature', and only one is required
a. Only one auth is included in the request and we skip adding http-signature because its value is not set. Everything works when the request is sent. 
b) all auth info is sent even though only one is needed. Everything works. 

Use case 3a is why I think the qty of authorizations and whether or not they are required matters. Do you get it? Which use cases use cases do you think are the most important and how do you want to proceed?

therve: Sorry I must miss something. I don't add anything anywhere? I just raise an error when a value is passed (and that value is passed today?). Is there case where we set the header to None, but we don't end up sending it?

spacether: Yes. Use case 3a is a case where we could have 2 or more auth settings on an endpoint, and only one is required. 
```
auth_settings = ['api_key', 'cookie']
config.Configuration(cookie_auth='blah, api_key=None)
# when we send the request only include cookie auth because api_key auth is unset. The server only needs one or the other
```
That is a use case where we could skip adding an authorization scheme if the values are unset (None).

therve: Ah ok, but if you do that, you don't get a value in `self.configuration.auth_settings().get(auth)` thus you don't go in the section below, no?

spacether: Yes that's right for 2 auths where one is present we would include the one auth and for the other one we would skip it (not go into the section below).

How about about making this change?
```
        used_auth = 0
        for auth in auth_settings:
            if auth_setting:
                uses_auth_value = (auth_setting['in'] in {'cookie', 'query'} or (auth_setting['in'] == 'header' and auth_setting['type'] != 'http-signature'))
                if uses_auth_value and auth_setting['value'] is None:
                    # Use cases:
                    # 1. qty 1 auth is required, if not included we raise an exception below
                    # 2. qty >= 2 auth are included, all are required, config only includes one, server throws an error
                    # 3. qty >= 2 auth are included, only one is required, everything works if only one included. If all are omitted we raise an exception below before making request to the server.
                    continue
                if auth_setting['in'] not in {'cookie', 'header', 'query'}:
                    raise ApiValueError(
                        'Authentication token must be in `query` or `header`, or `cookie`'
                    )
                used_auth += 1
                if auth_setting['in'] == 'cookie':
                    headers['Cookie'] = auth_setting['value']
                elif auth_setting['in'] == 'header':
                    if auth_setting['type'] != 'http-signature':
                        headers[auth_setting['key']] = auth_setting['value']
{{#hasHttpSignatureMethods}}
                    else:
                        # The HTTP signature scheme requires multiple HTTP headers
                        # that are calculated dynamically.
                        signing_info = self.configuration.signing_info
                        auth_headers = signing_info.get_http_signature_headers(
                                            resource_path, method, headers, body, querys)
                        headers.update(auth_headers)
{{/hasHttpSignatureMethods}}
                elif auth_setting['in'] == 'query':
                    querys.append((auth_setting['key'], auth_setting['value']))
                else:
                    raise ApiValueError(
                        'Authentication token must be in `query` or `header`, or `cookie`'
                    )
        if used_auth == 0:
            raise ApiValueError('This route defines auth but your configuration is missing the required auth. Set auth info in your configuration')
```

therve: I guess that would work? Seems pretty complicated.

spacether: I am open to other solutions here. Do you have a simpler way of supporting the use case where multiple auths are defined and only one is used?

therve: Sorry that didn't come up as I wanted. I'm not super familiar with that code, it would need some digging to give a real interesting feedback. I was trying to solve an easier problem initially :).

PR #8509, opened 2021-01-22
Add getters for private default Headers and authentications

NANASHI0X74:
Rationale for this is: At the moment I'm in a situation where code generation (or rather the scheme generation for our backend code) doesn't generate code that works with our backend, so I need to do one specific request by hand. I'd like to still use the same default headers and authentications that the Apiclient uses though.

This is a really minor change, so I thought I'd show you the code before talking. Submitting this as draft so I can do the build and sample generation later

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/bin/utils/export_docs_generators.sh
+++ b/bin/utils/export_docs_generators.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 SCRIPT="$0"
 echo "# START SCRIPT: ${SCRIPT}"
--- a/bin/utils/export_docs_generators.sh
+++ b/bin/utils/export_docs_generators.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 SCRIPT="$0"
 echo "# START SCRIPT: ${SCRIPT}"
--- a/bin/utils/export_generator.sh
+++ b/bin/utils/export_generator.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 SCRIPT="$0"
 
--- a/bin/utils/export_generator.sh
+++ b/bin/utils/export_generator.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 SCRIPT="$0"
 
--- a/modules/openapi-generator/src/main/resources/dart2/api_client.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/api_client.mustache
@@ -49,6 +49,12 @@ class ApiClient {
      _defaultHeaderMap[key] = value;
   }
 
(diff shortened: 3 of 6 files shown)


kuhnroyal:
Makes sense, follow the PR check list and regenerate the samples. Otherwise it won't build.

I would recommend returning immutable copies of the collections in the getters to avoid the problems that you added as documentation.

NANASHI0X74:
returning an immutable list for the authentications sounds like a good idea.
If I were to add entries to the default header map, those would get picked up though, right? I think it would have benefits to leave that mutable, for example I could use addAll() to add multiple headers at once and the like

kuhnroyal:
Hmm are you sure this is the correct template? The `dart` template folder is unmaintained.

NANASHI0X74:
so, I finally got round to doing this stuff. Thanks btw, I did mistakenly use the wrong template folder.
I took the liberty btw to also make the shebangs in two of the scripts portable (i.e. use /usr/bin/env bash) because the hardcoded /bin/bash actually caused problems on my nixOS machine.
Now, to mention the rest of the technical committee...
@swipesight @jaumard @josh-burton @amondnet @sbu-WBT @agilob

bin/utils/export_docs_generators.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
agilob, line 1: Out of scope for this PR

bin/utils/export_generator.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
agilob, line 1: Out of scope for this PR

agilob:
I don't understand why should none be added or deleted. If I want to add another default header conditionally from my code, I should be able to. Should your code copy the headers before any other action rather than making it immutable for everyone by force?

bin/utils/export_docs_generators.sh
agilob: If you want to add this in this PR, for git-log clarity it should come in own separate commit.

NANASHI0X74:
I would actually advocate for leaving the returned headermap mutable. For the Authentications it definitely makes sense to return them immutably since the authentications are defined in the spec and the generated code sort of enforces those by only respecting authentications on certain keys in the map from what I understand, but leaving the headermap mutable has the benefit that you can use the full dart collection api to mutate it if you want and not be constrained to only adding one header at a time via a function call. I don't really see any downside to that either...

kuhnroyal:
Fine by me

kuhnroyal (approved):

wing328:
CircleCI failure not related to this change.

kuhnroyal:
Broken master build fixed in https://github.com/OpenAPITools/openapi-generator/pull/8581

Issue #5014, opened 2020-01-16
[REQ] Swift Vapor server generator


wing328:
### Is your feature request related to a problem? Please describe.

We would like to create a Swift Vapor server generator similar to what we've done for NodeJS Express and other server generators.

Ref: https://github.com/vapor/vapor

I'm working with @vkill on this new generator.

MysterionRise:
Do we need to have server side Swift generator as well? For now, I mostly see only client ones

Dracks:
apple has already a generator: https://github.com/apple/swift-openapi-generator (I use this)

Issue #117, opened 2018-05-21
Travis CI issue


wing328:
##### Description

Travis CI timed out due to the following:
```
openapi-petstore-0.1.0.0: Test suite tests passed
Completed 14 action(s).
Compiling src/Byte.elm
Some new packages are needed. Here is the upgrade plan.

  Install:
    NoRedInk/elm-decode-pipeline 3.0.0
    elm-lang/core 5.1.1
    elm-lang/html 2.0.0
    elm-lang/http 1.0.0
    elm-lang/virtual-dom 2.0.4
    justinmimbs/elm-date-extra 2.1.1
```
##### openapi-generator version

Current master

##### Related issues/PRs

https://travis-ci.org/OpenAPITools/openapi-generator/builds/380911655

##### Suggest a fix/enhancement

We may want do disable the elm client test for the time being to see if the issue goes away

FantasyTeddy:
This issue should be resolved with PR #118 (f222245).

wing328:
Yup, thanks for the PR to fix that.

PR #8515, opened 2021-01-23
[swift] Add Swift Vapor client library

aymanbagabas:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
@@ -292,6 +292,7 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
 
         supportedLibraries.put(LIBRARY_URLSESSION, "[DEFAULT] HTTP client: URLSession");
         supportedLibraries.put(LIBRARY_ALAMOFIRE, "HTTP client: Alamofire");
+        supportedLibraries.put(LIBRARY_VAPOR, "HTTP client: Vapor");
 
         CliOption libraryOption = new CliOption(CodegenConstants.LIBRARY, "Library template (sub-template) to use");
         libraryOption.setEnum(supportedLibraries);
@@ -441,7 +442,7 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
         additionalProperties.put(READONLY_PROPERTIES, readonlyProperties);
 
         // Setup swiftUseApiNamespace option, which makes all the API
-        // classes inner-class of {{projectName}}API
+        // classes inner-class of {{projectName}}
         if (additionalProperties.containsKey(SWIFT_USE_API_NAMESPACE)) {
             setSwiftUseApiNamespace(convertPropertyToBooleanAndWriteBack(SWIFT_USE_API_NAMESPACE));
         }
@@ -484,57 +485,59 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
         additionalProperties.put("apiDocPath", apiDocPath);
         additionalProperties.put("modelDocPath", modelDocPath);
 
-        supportingFiles.add(new SupportingFile("Podspec.mustache",
-                "",
-                projectName + ".podspec"));
-        supportingFiles.add(new SupportingFile("Cartfile.mustache",
-                "",
-                "Cartfile"));
+        if (!getLibrary().equals(LIBRARY_VAPOR)) {
+            supportingFiles.add(new SupportingFile("Podspec.mustache",
+                    "",
+                    projectName + ".podspec"));
+            supportingFiles.add(new SupportingFile("Cartfile.mustache",
+                    "",
+                    "Cartfile"));
+            supportingFiles.add(new SupportingFile("CodableHelper.mustache",
+                    sourceFolder,
+                    "CodableHelper.swift"));
+            supportingFiles.add(new SupportingFile("OpenISO8601DateFormatter.mustache",
+                    sourceFolder,
+                    "OpenISO8601DateFormatter.swift"));
+            supportingFiles.add(new SupportingFile("JSONDataEncoding.mustache",
+                    sourceFolder,
+                    "JSONDataEncoding.swift"));
+            supportingFiles.add(new SupportingFile("JSONEncodingHelper.mustache",
+                    sourceFolder,
+                    "JSONEncodingHelper.swift"));
+            supportingFiles.add(new SupportingFile("git_push.sh.mustache",
+                    "",
+                    "git_push.sh"));
+            supportingFiles.add(new SupportingFile("SynchronizedDictionary.mustache",
+                    sourceFolder,
+                    "SynchronizedDictionary.swift"));
+            supportingFiles.add(new SupportingFile("XcodeGen.mustache",
+                    "",
+                    "project.yml"));
+            supportingFiles.add(new SupportingFile("APIHelper.mustache",
+                    sourceFolder,
+                    "APIHelper.swift"));
+            supportingFiles.add(new SupportingFile("Models.mustache",
+                    sourceFolder,
+                    "Models.swift"));
+        }
         supportingFiles.add(new SupportingFile("Package.swift.mustache",
                 "",
                 "Package.swift"));
-        supportingFiles.add(new SupportingFile("APIHelper.mustache",
-                sourceFolder,
-                "APIHelper.swift"));
         supportingFiles.add(new SupportingFile("Configuration.mustache",
                 sourceFolder,
                 "Configuration.swift"));
         supportingFiles.add(new SupportingFile("Extensions.mustache",
                 sourceFolder,
                 "Extensions.swift"));
-        supportingFiles.add(new SupportingFile("Models.mustache",
-                sourceFolder,
-                "Models.swift"));
         supportingFiles.add(new SupportingFile("APIs.mustache",
                 sourceFolder,
                 "APIs.swift"));
-        supportingFiles.add(new SupportingFile("CodableHelper.mustache",
-                sourceFolder,
-                "CodableHelper.swift"));
-        supportingFiles.add(new SupportingFile("OpenISO8601DateFormatter.mustache",
-                sourceFolder,
-                "OpenISO8601DateFormatter.swift"));
-        supportingFiles.add(new SupportingFile("JSONDataEncoding.mustache",
-                sourceFolder,
-                "JSONDataEncoding.swift"));
-        supportingFiles.add(new SupportingFile("JSONEncodingHelper.mustache",
-                sourceFolder,
-                "JSONEncodingHelper.swift"));
-        supportingFiles.add(new SupportingFile("git_push.sh.mustache",
-                "",
-                "git_push.sh"));
-        supportingFiles.add(new SupportingFile("SynchronizedDictionary.mustache",
-                sourceFolder,
-                "SynchronizedDictionary.swift"));
         supportingFiles.add(new SupportingFile("gitignore.mustache",
                 "",
                 ".gitignore"));
         supportingFiles.add(new SupportingFile("README.mustache",
                 "",
                 "README.md"));
-        supportingFiles.add(new SupportingFile("XcodeGen.mustache",
-                "",
-                "project.yml"));
 
         switch (getLibrary()) {
             case LIBRARY_ALAMOFIRE:
@@ -548,6 +551,9 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
                 supportingFiles.add(new SupportingFile("URLSessionImplementations.mustache",
                         sourceFolder,
                         "URLSessionImplementations.swift"));
+                break;
+            case LIBRARY_VAPOR:
+                additionalProperties.put("useVapor", true);
                 break;
             default:
                 break;
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
@@ -63,11 +63,13 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
     public static final String LENIENT_TYPE_CAST = "lenientTypeCast";
     public static final String USE_SPM_FILE_STRUCTURE = "useSPMFileStructure";
     public static final String SWIFT_PACKAGE_PATH = "swiftPackagePath";
+    public static final String USE_CLASSES = "useClasses";
     public static final String USE_BACKTICK_ESCAPES = "useBacktickEscapes";
     public static final String GENERATE_MODEL_ADDITIONAL_PROPERTIES = "generateModelAdditionalProperties";
     public static final String HASHABLE_MODELS = "hashableModels";
     protected static final String LIBRARY_ALAMOFIRE = "alamofire";
     protected static final String LIBRARY_URLSESSION = "urlsession";
+    protected static final String LIBRARY_VAPOR = "vapor";
     protected static final String RESPONSE_LIBRARY_PROMISE_KIT = "PromiseKit";
     protected static final String RESPONSE_LIBRARY_RX_SWIFT = "RxSwift";
     protected static final String RESPONSE_LIBRARY_RESULT = "Result";
@@ -81,6 +83,7 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
     protected boolean swiftUseApiNamespace = false;
     protected boolean useSPMFileStructure = false;
     protected String swiftPackagePath = "Classes" + File.separator + "OpenAPIs";
+    protected boolean useClasses = false;
     protected boolean useBacktickEscapes = false;
     protected boolean generateModelAdditionalProperties = true;
     protected boolean hashableModels = true;
@@ -280,6 +283,8 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
                 + " and set the source path to Sources" + File.separator + "{{projectName}} (default: false)."));
         cliOptions.add(new CliOption(SWIFT_PACKAGE_PATH, "Set a custom source path instead of "
                 + projectName + File.separator + "Classes" + File.separator + "OpenAPIs" + "."));
+        cliOptions.add(new CliOption(USE_CLASSES, "Use final classes for models instead of structs (default: false)")
+                .defaultValue(Boolean.FALSE.toString()));
 
         cliOptions.add(new CliOption(HASHABLE_MODELS,
             "Make hashable models (default: true)")
@@ -468,6 +473,10 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
             setHashableModels(convertPropertyToBooleanAndWriteBack(HASHABLE_MODELS));
         }
         additionalProperties.put(HASHABLE_MODELS, hashableModels);
+
+        if (additionalProperties.containsKey(USE_CLASSES)) {
+            setUseClasses(convertPropertyToBooleanAndWriteBack(USE_CLASSES));
+        }
 
         setLenientTypeCast(convertPropertyToBooleanAndWriteBack(LENIENT_TYPE_CAST));
 
@@ -879,6 +888,10 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
 
     public void setSwiftPackagePath(String swiftPackagePath) {
         this.swiftPackagePath = swiftPackagePath;
+    }
+
+    public void setUseClasses(boolean useClasses) {
+        this.useClasses = useClasses;
     }
 
     public void setUseBacktickEscapes(boolean useBacktickEscapes) {
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
@@ -120,7 +120,6 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
                         "Bool",
                         "Void",
                         "String",
-                        "URL",
                         "Data",
                         "Date",
                         "Character",
@@ -224,8 +223,8 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
         typeMapping.put("float", "Float");
         typeMapping.put("number", "Double");
         typeMapping.put("double", "Double");
-        typeMapping.put("file", "URL");
-        typeMapping.put("binary", "URL");
+        typeMapping.put("file", "Data");
+        typeMapping.put("binary", "Data");
         typeMapping.put("ByteArray", "Data");
         typeMapping.put("UUID", "UUID");
         typeMapping.put("URI", "String");
@@ -602,7 +601,7 @@ public class Swift5ClientCodegen extends DefaultCodegen implements CodegenConfig
 
     @Override
     public boolean isDataTypeFile(String dataType) {
-        return "URL".equals(dataType);
+        return "Data".equals(dataType);
     }
 
     @Override
--- a/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
@@ -5,16 +5,25 @@
 //
 
 import Foundation
+{{#useVapor}}
+import Vapor
+{{/useVapor}}
 
-{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{projectName}}API {
+{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{projectName}} {
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var basePath = "{{{basePath}}}"
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var credential: URLCredential?
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var customHeaders: [String: String] = [:]{{#useAlamofire}}
+    {{#useVapor}}
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var customHeaders: HTTPHeaders = [:]
+    {{/useVapor}}
+    {{^useVapor}}
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var customHeaders: [String: String] = [:]
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var credential: URLCredential?{{#useAlamofire}}
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var requestBuilderFactory: RequestBuilderFactory = AlamofireRequestBuilderFactory(){{/useAlamofire}}{{#useURLSession}}
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var requestBuilderFactory: RequestBuilderFactory = URLSessionRequestBuilderFactory(){{/useURLSession}}
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} static var apiResponseQueue: DispatchQueue = .main
+    {{/useVapor}}
 }
 
+{{#useVapor}}{{/useVapor}}{{^useVapor}}
 {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class RequestBuilder<T> {
     var credential: URLCredential?
     var headers: [String: String]
@@ -33,7 +42,7 @@ import Foundation
         self.parameters = parameters
         self.headers = headers
 
-        addHeaders({{projectName}}API.customHeaders)
+        addHeaders({{projectName}}.customHeaders)
     }
 
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func addHeaders(_ aHeaders: [String: String]) {
@@ -42,7 +51,7 @@ import Foundation
         }
     }
 
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, Error>) -> Void) { }
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, Error>) -> Void) { }
 
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} func addHeader(name: String, value: String) -> Self {
         if !value.isEmpty {
@@ -52,7 +61,7 @@ import Foundation
     }
 
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func addCredential() -> Self {
-        credential = {{projectName}}API.credential
+        credential = {{projectName}}.credential
         return self
     }
 }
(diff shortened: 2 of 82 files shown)


aymanbagabas:
@4brunu @tanner0101

Hi all,

This PR is not done yet but I would like to have your feedback on this. I still need to do the API docs and testing, the main components are pretty much done. This would resolve https://github.com/OpenAPITools/openapi-generator/issues/5014 and https://github.com/OpenAPITools/openapi-generator/issues/4957

4brunu:
Hi @aymanbagabas, 
First of all, great work, and thanks for taking the time and effort to do this contribution 👍

Since Vapor is a server side framework, and this introduces a lot of changes, I think it makes more sence to separate this into a new generator.
Looking at other generators, for example `kotlin`, it has `AbstractKotlinCodegen.java`, which has the shared base code for all kotlin generator, and then `KtormSchemaCodegen.java`, `KotlinClientCodegen.java`, `KotlinServerCodegen.java`, `KotlinSpringServerCodegen.java`, `KotlinVertxServerCodegen.java`.
I think we could do the same, since we already have `Swift5ClientCodegen.java`.

What do you think about this?
Related to Vapor, I never used it, so it's difficult to me to review this implementation.

I will comment on some changed on the code, to try to better understand some of the changes.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenOperation.java
```diff
@@ -29,7 +29,7 @@
             isArray, isMultipart,
             isResponseBinary = false, isResponseFile = false, hasReference = false,
             isRestfulIndex, isRestfulShow, isRestfulCreate, isRestfulUpdate, isRestfulDestroy,
-            isRestful, isDeprecated, isCallbackRequest, uniqueItems;
```
4brunu, line 32: Hi, could you elabore on what's the use case for this default response?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -75,7 +76,7 @@
     protected boolean readonlyProperties = false;
     protected boolean swiftUseApiNamespace;
     protected String[] responseAs = new String[0];
-    protected String sourceFolder = "Classes" + File.separator + "OpenAPIs";
```
4brunu, line 78: Is this change related to some requirement of Vapor or SPM?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -109,7 +110,6 @@ public Swift5ClientCodegen() {
                         "Bool",
                         "Void",
                         "String",
-                        "URL",
```
4brunu, line 112: Why did you dele URL here?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -195,7 +195,6 @@ public Swift5ClientCodegen() {
 
         typeMapping = new HashMap<>();
         typeMapping.put("array", "Array");
-        typeMapping.put("List", "Array");
```
4brunu, line 198: Why did you delete this?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -499,7 +504,7 @@ public String escapeReservedWord(String name) {
         if (this.reservedWordsMappings().containsKey(name)) {
             return this.reservedWordsMappings().get(name);
         }
-        return "_" + name;  // add an underscore to the name
```
4brunu, line 502: Could you explain this change please?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -807,7 +812,7 @@ public void setSwiftUseApiNamespace(boolean swiftUseApiNamespace) {
     @Override
     public String toEnumValue(String value, String datatype) {
         // for string, array of string
-        if ("String".equals(datatype) || "[String]".equals(datatype) || "[String: String]".equals(datatype)) {
```
4brunu, line 810: Could you explain this change please?

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
```diff
@@ -5,16 +5,25 @@
 //
 
 import Foundation
+{{#useVapor}}
+import Vapor
+{{/useVapor}}
 
-{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{projectName}}API {
```
4brunu, line 9: Hi, could you explain the why of this change?

modules/openapi-generator/src/main/resources/swift5/Package.swift.mustache
```diff
@@ -1,12 +1,12 @@
...
+// swift-tools-version:5.1
 
 import PackageDescription
 
 let package = Package(
     name: "{{projectName}}",
     platforms: [
         .iOS(.v9),
-        .macOS(.v10_11),
+        .macOS(.v10_15),
```
4brunu, line 9: I think here we should put an if uses Vapor

4brunu (commented):
Hey, I just left a few comments, to try to understand the changes 🙂

4brunu:
Is this already working already? 🙂 
Nice work, thanks once again for your contribution, and please don't feel discouraged by all the questions I left, I'm just trying to understand the changes.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenOperation.java
aymanbagabas: I needed a way to determine if operation responses contain a default response. The way I'm handling each operation return value is to return an enum based on the response status code.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: Here I'm just following swift libraries' common conventions. https://swift.org/package-manager/#creating-a-library-package

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: duped in line [#117](https://github.com/OpenAPITools/openapi-generator/pull/8515/files#diff-4a35291cd15aa856298c1f88a25af4ed1043f7776bff67e3803f20008a7763a6R117)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: List is not a Swift type. I had a problem while testing [DocuSign specs](https://github.com/docusign/OpenAPI-Specifications) where they define a type `List` and that got mapped to the wrong value `Array` which is a primitive type in swift.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: Swift uses backticks to escape reserved words. Refer to the note section [here](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID313)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: [Character](https://developer.apple.com/documentation/swift/character) uses quotes when initialized same as strings it needs to be surrounded by quotes `"` when defined as an enum.

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
aymanbagabas: I figured it's extraneous to have `API` added to the project name since in most cases the project name _is_ in fact the library/class name.

modules/openapi-generator/src/main/resources/swift5/Package.swift.mustache
aymanbagabas: Agree.

aymanbagabas (commented):
Thank you @4brunu for taking the time to review this pr.

While in fact Vapor is a server-side framework, here I'm using Vapor's [client](https://docs.vapor.codes/4.0/client/)/[content](https://docs.vapor.codes/4.0/content/) APIs to generate Vapor compatible APIs.

aymanbagabas:
> Looking at other generators, for example `kotlin`, it has `AbstractKotlinCodegen.java`, which has the shared base code for all kotlin generator, and then `KtormSchemaCodegen.java`, `KotlinClientCodegen.java`, `KotlinServerCodegen.java`, `KotlinSpringServerCodegen.java`, `KotlinVertxServerCodegen.java`.
> I think we could do the same, since we already have `Swift5ClientCodegen.java`.
> 
> What do you think about this?
> Related to Vapor, I never used it, so it's difficult to me to review this implementation.

While it will be cleaner, I don't see that as necessary since all the libraries (including this one) are client-side generators. And the language mapping is the same, the only difference I can see is that Vapor already implements all the boiler plate code needed for generating a client API, i.e. encoding, decoding, and other helpers are all built-in Vapor.

What I would do is separate Vapor templates from other libraries' templates, I think that would be a good middle ground. What do you think?

4brunu:
> While it will be cleaner, I don't see that as necessary since all the libraries (including this one) are client-side generators. And the language mapping is the same, the only difference I can see is that Vapor already implements all the boiler plate code needed for generating a client API, i.e. encoding, decoding, and other helpers are all built-in Vapor.
> 
> What I would do is separate Vapor templates from other libraries' templates, I think that would be a good middle ground. What do you think?

From what I understand, this will be used to generate a client project, that will be used on a Vapor server, so that the Vapor server can invoke other API servers, is this correct or did I understand it wrong?

The main blocker that I see in this PR, is that it has a lot of breaking changes.
So I suggest to separate the breaking changes from the new library option `Vapor` to make it easier to merge.
I will comment all the breaking changes so that we can discuss them.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: I like this change, and I think in the future we should adopt it, but this is a breaking change, so we can't merge it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -214,8 +213,8 @@ public Swift5ClientCodegen() {
         typeMapping.put("float", "Float");
         typeMapping.put("number", "Double");
         typeMapping.put("double", "Double");
-        typeMapping.put("file", "URL");
```
4brunu, line 217: Why did you change the `file` and `binary` formats from `URL` to `Data`?
This will break the existing generator.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: Can you elaborate more on this issue please? Perhaps show an example of the issue?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: I added a comment on line 217

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -353,7 +353,7 @@ public void processOpts() {
         } else {
             additionalProperties.put(PROJECT_NAME, projectName);
         }
-        sourceFolder = projectName + File.separator + sourceFolder;
```
4brunu, line 356: This I also a breaking change, that I think is related to the comments in line 78

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: This is a breaking change, that I suggested to open a new PR with, against the `6.0.x` branch

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -546,7 +551,7 @@ public String getSchemaType(Schema p) {
 
     @Override
     public boolean isDataTypeFile(String dataType) {
-        return dataType != null && dataType.equals("URL");
```
4brunu, line 549: This change is related to the one in line 217

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -720,11 +725,16 @@ public String toVarName(String name) {
...
         name = camelize(name, true);
 
-        // for reserved word or word starting with number, append _
-        if (isReservedWord(name) || name.matches("^\\d.*")) {
+        // for reserved words surround with ``
+        if (isReservedWord(name)) {
             name = escapeReservedWord(name);
         }
 
+        // for words starting with number, append _
```
4brunu, line 733: This is also a breaking change, can you explain the why of this change please?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -745,11 +755,16 @@ public String toParamName(String name) {
         // pet_id => petId
         name = camelize(name, true);
 
-        // for reserved word or word starting with number, append _
```
4brunu, line 748: Same here as in line 733

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: So this is a bug fix? It's not working in the current release?

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
4brunu: This is a breaking change the affects a lot of files, so I suggest to revert it.

modules/openapi-generator/src/main/resources/swift5/Package.swift.mustache
```diff
@@ -34,8 +42,8 @@ let package = Package(
         // Targets can depend on other targets in this package, and on products in packages which this package depends on.
         .target(
             name: "{{projectName}}",
-            dependencies: [{{#useAlamofire}}"Alamofire", {{/useAlamofire}}{{#usePromiseKit}}"PromiseKit", {{/usePromiseKit}}{{#useRxSwift}}"RxSwift"{{/useRxSwift}}],
-            path: "{{projectName}}/Classes"
+            dependencies: [{{#useVapor}}"Vapor", {{/useVapor}}{{#useAlamofire}}"Alamofire", {{/useAlamofire}}{{#usePromiseKit}}"PromiseKit", {{/usePromiseKit}}{{#useRxSwift}}"RxSwift"{{/useRxSwift}}],
+            path: "Sources"
```
4brunu, line 46: This change is related to `Swift5ClientCodegen.java` line 356

modules/openapi-generator/src/main/resources/swift5/Configuration.mustache
```diff
@@ -5,10 +5,19 @@
 //
 
 import Foundation
+{{#useVapor}}import Vapor{{/useVapor}}
+{{#swiftUseApiNamespace}} 
 
+extension {{projectName}} {
```
4brunu, line 11: Make this class an extension of `{{projectName}}` is also a breaking change.

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -10,15 +10,19 @@ import PromiseKit{{/usePromiseKit}}{{#useRxSwift}}
...
+import Vapor{{/useVapor}}{{#swiftUseApiNamespace}}
 
-extension {{projectName}}API {
+extension {{projectName}} {
 {{/swiftUseApiNamespace}}
 
 {{#description}}
 /** {{description}} */{{/description}}
 {{#objcCompatible}}@objc {{/objcCompatible}}{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{classname}}{{#objcCompatible}} : NSObject{{/objcCompatible}} {
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} init() { }
```
4brunu, line 22: Is there any advantage in using the instance method instead of calling a static method?
If this change is made, this should be surrounded with
```
{{#useVapor}}
...
{{/useVapor}}
```

4brunu (commented):
I think I commented all the breaking changes

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: I have been thinking about this and what we can do is to introduce a flag like `useSPMFileStructure` that is false by default to keep backwards compatibility.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: I noticed that `List` is not a Swift nor OAS data type. I was working with DS specs and one of their models is named `List`, defined as below.  I was facing a problem where the generator was naming this model `Array` instead of `List`.

```
...
    "list": {
      "type": "object",
      "properties": {
        "anchorAllowWhiteSpaceInCharacters": {
          "description": "",
          "type": "string"
        },
        "anchorAllowWhiteSpaceInCharactersMetadata": {
          "$ref": "#/definitions/propertyMetadata",
          "description": ""
        },
        "anchorCaseSensitive": {
...
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: I noticed that this breaks `objCompatible` files, so I changed it to escape `objCompatible` reserved words with a `_` and back-ticks otherwise.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: `file` and `binary` both refer to binary format and `URL` != binary.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: All Swift reserved words can be escaped using backticks as stated [here](https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID412).

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: Commit dropped. I was trying something out and forgot to remove this line.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: You are right, thanks for this fix and for the extra info 👍

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -485,6 +509,13 @@ public void processOpts() {
                         sourceFolder,
                         "URLSessionImplementations.swift"));
                 break;
+            case LIBRARY_VAPOR:
+                additionalProperties.put("useVapor", true);
+                typeMapping.put("object", "JSON");
```
4brunu, line 514: @aymanbagabas I think the change of `AnyType` and `object` to `JSON`, could be applied to all libraries.
What do you think?

modules/openapi-generator/src/main/resources/swift5/Extensions.mustache
```diff
@@ -192,4 +194,206 @@ extension RequestBuilder {
         }
         return deferred.promise
     }
-}{{/usePromiseKit}}
+}{{/usePromiseKit}}{{#useVapor}}
+
+public enum JSON: Content {
```
4brunu, line 199: @aymanbagabas What was the reference implementation of this?

aymanbagabas: The JSON type is based on https://github.com/vapor/vapor/commit/b7786e8a60c71c2e86765eacd9f403f076a081b8 by @tanner0101. I'll add that to the commit description.

4brunu: We just need to be sure that there is no licensing issue here.

4brunu:
Could you please run the following commands to update the other samples?
Because the changes to the other samples are making a lot of noise in this PR. 
```
./mvnw clean package 
./bin/generate-samples.sh
```

4brunu:
Thanks for updating the samples 👍 
When you have some time could you please investigate the changes on the other generators?
I see a lot of changes in the other generators that are not suppose to be there.
Ideally the other generators should only have intencional changes like the JSON object.

aymanbagabas:
```
Fetching https://github.com/vapor/vapor.git
error: the package dependency graph could not be resolved; unable to find any available tag for the following requirements:
    https://github.com/vapor/vapor.git @ 4.0.0..<5.0.0
```

hmm, `xcodebuild-test` doesn't like fetching Vapor from Github.

4brunu:
Sure, that can be done in `bitrise.yml` on the root directory of the project.

aymanbagabas:
> Sure, that can be done in `bitrise.yml` on the root directory of the project.

I don't think you can change the stack being used from `bitrise.yml` it has to be done from Bitrise UI.

4brunu:
It is, try it 🙂

aymanbagabas:
> It is, try it 🙂

https://discuss.bitrise.io/t/defining-which-stack-to-use-in-the-yml-file/2882

4brunu:
@wing328 could you please help changing the Xcode version in bitrise?

modules/openapi-generator/src/main/resources/swift5/modelObject.mustache
```diff
@@ -1,4 +1,4 @@
-{{^objcCompatible}}{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} struct {{classname}}: Codable{{#vendorExtensions.x-swift-hashable}}, Hashable{{/vendorExtensions.x-swift-hashable}} {
+{{^objcCompatible}}{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} {{#useClassModels}}final class{{/useClassModels}}{{^useClassModels}}struct{{/useClassModels}} {{{classname}}}: {{#useVapor}}Content{{/useVapor}}{{^useVapor}}Codable{{/useVapor}} {
```
4brunu, line 1: @aymanbagabas just curious, why do you want to use a class here, instead of a struct?

4brunu:
@aymanbagabas since this PR it quite big now, would you mind isolate some changes here, and send them in a different PR?
Two that come to mind are `useSPMFileStructure` and the `JSON` object.
What do you think of this?

aymanbagabas:
The build is failing because of `xcodebuild`, possibly because of Xcode 11 which comes with swift 5.1. As of today, the latest stable swift version is 5.3.3 which ships with Xcode 12. Some people reported the same problem with swift 5.1 [here](https://forums.swift.org/t/spm-dependency-resolution-version-vs-tag-branch/30089), [here](https://github.com/dropbox/SwiftyDropbox/issues/252), and [here](https://github.com/saoudrizwan/Disk/issues/80).

```
Fetching https://github.com/vapor/vapor.git
error: the package dependency graph could not be resolved; unable to find any available tag for the following requirements:
    https://github.com/vapor/vapor.git @ 4.0.0..<5.0.0
[INFO] BUILD FAILURE
[INFO] Total time:  7.332 s
[INFO] Finished at: 2021-03-30T12:06:22-07:00
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (xcodebuild-test) on project Swift5PetstoreClientTests: Command execution failed.: Process exited with an error: 1 (Exit value: 1) -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
|                                                                              |
+---+---------------------------------------------------------------+----------+
| x | Run Swift5 tests (exit code: 1)                               | 4.9 min  |
+---+---------------------------------------------------------------+----------+
| Issue tracker: https://github.com/bitrise-io/steps-script/issues             |
| Source: https://github.com/bitrise-io/steps-script                           |
+---+---------------------------------------------------------------+----------+
```

4brunu:
> > @aymanbagabas since this PR it quite big now, would you mind isolate some changes here, and send them in a different PR?
> > Two that come to mind are `useSPMFileStructure` and the `JSON` object.
> > What do you think of this?
> 
> I did make `useSPMFileStructure` into its own [PR](https://github.com/OpenAPITools/openapi-generator/pull/9074) but not the other one. Man splitting and restructuring git commits is _not_ fun at all.

Thanks, I already reviewed it.

samples/client/petstore/swift5/urlsessionLibrary/PetstoreClient/Classes/OpenAPIs/Models/Order.swift
```diff
@@ -30,5 +31,61 @@ public struct Order: Codable {
...
+        case petId
+        case quantity
+        case shipDate
+        case status
+        case complete
+    }
+
+    // Encodable protocol methods
+
+    public func encode(to encoder: Encoder) throws {
```
4brunu, line 45: @aymanbagabas this should only be done when using classes, not structs.

aymanbagabas: [Codable](https://developer.apple.com/documentation/swift/codable) can be applied to both AFAICT. I added `encode(encoder:)` and `init(decoder:)` explicitly to decode null values since you _need_ to use `encoderIfPresent()` and `decoderIfPresent()` to capture nulls instead of ignoring the optionals all together.

4brunu: I know we can use explicitly `encode(encoder:)` to `structs`, but I think we should rely on the implicit implementation, because it's less code to maintain and one less source of bugs.

4brunu: As far as I now, the implicit implementation of `encode(encoder:)` already deals with the nulls, or not?

aymanbagabas: I believe you're right. I meant to encode nullable variables. To do so, you gotta use `encode(encoder:)` instead of `encodeIfPresent(encoder:)` for optional values so that it encodes nulls. We really don't need `init(decoder:)` so I'm just gonna drop that one. Here's a good example of what I'm talking about https://stackoverflow.com/a/47268112/10913628.

These are the possible four cases of encoding. I'm using String as an example.

| isNullable | isRequired | type | func
|-----------|------------|------|-------|
| true | true | String? | encode() |
| true | false | String? | encodeIfPresent() |
| false | true | String | encode() |
| false | false | String? | encodeIfPresent() |

My question is, does the generator always has `isNullable` and `isRequired` available? Because I know that `nullable` is an OAI 3.0 feature and it doesn't exist in either 2.0 nor 3.1.

4brunu: Since that's a question related to OAI, I don't know the answer.
But I think everything comes down to `String?` or `String`.
Looking at the first case, `isNullable` `true`, and `isRequired` `true`, I'm not sure if the `func` should be `encode()` or  `encodeIfPresent()`.

aymanbagabas: If we want to encode `null` in the generated JSON, we have to use `encode()` instead of `encodeIfPresent()`. Note that `encodeIfPresent()` is the one used for optionals and that's the default implementation of `func encode(to:)`.

aymanbagabas: @4brunu I will make this change to encode `null`s in the encoded JSONs. I'm also gonna start splitting this PR into smaller ones.

4brunu: Thanks for your effort in splitting this big PR into small ones, that will make it a lot easier to review each PR.

Some good candidates to a small PR are this one, Any to JSON, reserved word _ or `.

Ideally this PR end's up just having the Vapor changes.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: @wing328  @4brunu I noticed that there is inconsistency between `typeMapping` and `isDataTypeFile/isDataTypeBinary`. I believe that the typeMapping for both `file` and `binary` **should** be mapped to `Data` in Swift. [OpenAPI 3.0 file upload specs](https://swagger.io/docs/specification/describing-request-body/file-upload/)

Should I fix this and put it in a separate PR?

```
    typeMapping.put("file", "URL");
    typeMapping.put("binary", "URL");

    ....

    @Override
    public boolean isDataTypeFile(String dataType) {
        return "URL".equals(dataType);
    }

    @Override
    public boolean isDataTypeBinary(final String dataType) {
        return "Data".equals(dataType);
    }
```

aymanbagabas:
@4brunu @wing328 I've split most of the changes in the following PRs
https://github.com/OpenAPITools/openapi-generator/pull/9074 https://github.com/OpenAPITools/openapi-generator/pull/9203 https://github.com/OpenAPITools/openapi-generator/pull/9204  https://github.com/OpenAPITools/openapi-generator/pull/9205 https://github.com/OpenAPITools/openapi-generator/pull/9206
Adding Vapor client is the last one left which depends heavily on these PRs. I'll keep this PR open for discussion.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
4brunu: The `file` being associated with `URL` is right, because the `file` upload works based on the `file` `URL`.
The `binary` maybe it should be associated to `Data`, but I'm not 100% sure.

aymanbagabas: From what I understand from the [2.0 specs](https://swagger.io/specification/v2/), `file` should *also* be associated with `Data`. You could try it yourself. Go to [Petstore](https://petstore.swagger.io/#/pet/uploadFile), open your browser's devtools and go to the network tab. Try the uploadFile endpoint and choose an arbitrary id and file then see what your request looks like.

The request has a `content-type` of `multipart/form-data` and the body contains the raw binary octets representing the file being uploaded. This is the same as the `binary` type except that `file` uses `multipart/form-data` and `binary` uses other MIME types like `application/octet-stream`.

4brunu: The Java and Kotlin generators also map `file` to `java.io.File`, and the `URL` is the closest we have to file in Swift.
Then the URLSession and Alamofire will configure the `multipart/form-data` with the file `URL`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -580,7 +594,7 @@ public String getSchemaType(Schema p) {
 
     @Override
     public boolean isDataTypeFile(String dataType) {
-        return "URL".equals(dataType);
+        return isDataTypeBinary(dataType);
```
4brunu, line 597: @aymanbagabas this will result in a StackOverflow exception.

PR #8551, opened 2021-01-27
Fix handling of 1xx and 3xx in Rust Reqwest

gferon:
An `Err(...)` was returned for 1xx and 3xx status codes, which are totally valid response code and should be returned with `Ok(...)` instead.

Also removed a redundant `use` (found using `clippy`).

I filed this on the `5.1.x` branch since I think it qualifies as a hotfix, but feel free to change it to something more appropriate.

Tagging @farcaller for the Rust review.


wing328:
@gferon thanks for the PR, which looks like a bug fix to me so I've retargeted this PR to the current master instead.

cc https://github.com/OpenAPITools/openapi-generator#62---openapi-generator-technical-committee

gferon:
Thanks! I'll rebase the branch on master then so the diff looks better.

wing328:
Closing this one as I've cherry-pick your commits and filed https://github.com/OpenAPITools/openapi-generator/pull/8574 instead. If no further feedback from anyone, I'll merge that PR over the weekend.

PR #8556, opened 2021-01-27
make shebangs portable

NANASHI0X74:
change the shebangs for the scripts that generate samples to locate bash via `/usr/bin/env`
Some systems, like nixOS, don't use the linux filesystem hierarchy standard and/or have a bash binary in that path. `/usr/bin/env` is available on most systems even that don't go with the regular FHS and will locate the interpreter using the user's $PATH instead. for more info: https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

### DIFF
--- a/bin/utils/export_docs_generators.sh
+++ b/bin/utils/export_docs_generators.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 SCRIPT="$0"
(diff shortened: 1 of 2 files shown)


wing328:

NANASHI0X74:
oh yeah thanks for pointing that out... I guess I'll have to fix the email address. But for this commit I don't really care :p you can still merge it regardless

PR #8558, opened 2021-01-27
[swift5] URLSession: Fix memory leak of SessionDelegate

fl034:
`cleanupRequest()` is only called on error for now. So `sessionDelegate` doesn't get deallocated after request.

I think it's okay to call `cleanupRequest()` directly after the processing of the response, instead of putting it in the completion closure, since it will only deallocate when not in use anymore. What do you guys think?

Bildschirmfoto 2021-01-27 um 19 37 57

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
@@ -24,9 +24,6 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
 
 {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class URLSessionRequestBuilder<T>: RequestBuilder<T> {
 
-    // swiftlint:disable:next weak_delegate
-    fileprivate let sessionDelegate = SessionDelegate()
-
     /**
      May be assigned if you want to control the authentication challenges.
      */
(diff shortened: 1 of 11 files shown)


fl034:
@4brunu

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -126,6 +124,7 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
         }
 
         let cleanupRequest = {
+            urlSessionStore[urlSessionId]?.finishTasksAndInvalidate()
```
4brunu, line 127: Interesting, I didn't know this API.
Can you please comment on the need of this?
Just curious.
Thanks 😊

4brunu (approved):
Looks good to me.
Thanks for your contribution.

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
fl034: This invalidates the session after the all tasks of this session have completely finished. New tasks cannot be started on this session.
This call is not necessary, but good practice, since we're creating a session for each request, so after a task as finished we don't need the session anymore.

4brunu: Nice, thanks 👍

4brunu:
For me it's ready to merge 👍

PR #8567, opened 2021-01-28
upgrade scala-sttp generator to use sttp3

micossow:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


michalo2882:
cc: @clasnake (2017/07), @jimschubert (2017/09) ❤️, @shijinkui (2018/01), @ramzimaalej (2018/03), @chameleon82 (2020/03), @Bouillie (2020/04)
cc2: @adamw

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/api/PetApi.scala
```diff
@@ -34,15 +33,14 @@ class PetApi(baseUrl: String) {
...
-  def addPet(pet: Pet): RequestT[Identity, Either[ResponseException[ErrorModel, Exception], Pet], Any] =
+  def addPet(pet: Pet): Request[Either[ResponseException[ApiModel, Exception], Pet], Any] =
     basicRequest
       .method(Method.POST, uri"$baseUrl/pet")
       .contentType("application/json")
       .body(pet)
       .response(fromMetadata(
-        asJsonEither[ErrorModel, Pet],
-        ConditionalResponseAs(_.code == StatusCode(200), asJsonEither[ErrorModel, Pet]),
-        ConditionalResponseAs(_.code == StatusCode(405), asJsonEither[MethodNotAllowed, Pet])
```
chameleon82: It will be nice to have all statuses mapped to Error Models rather than have one generic ApiModel. Otherwise we loose specification provided by OpenApi and have to implement it outside of autogenerated client.

Ideally it will be nice to convert model into exception with ErrorModel supported, unfortunately OpenApi doesn't support it directly.

here the example what i mean. Something like this:
```scala
case class PetModelNotAllowed[ErrorModel](result: ErrorModel) extends Exception

client.addPet(myPetRequest) match {
 case Left(successfulPetModel) => doSomething(successfulPetModel)
 case Right(ex: PetModelNotAllowed, _) =>
   logger.error(s"Oops, something went wrong. Code for reference ${ex.errorCode}. ${ex.errorMessage}
   case Right(_, ex: Throwable) =>
     logger.error("Request failed", ex)
 }
```

chameleon82: Even generic `ApiException` can be fine to handle ErrorModel 
```scala
case class ApiException[T: ErrorModel](statusCode: HttpStatus, errorModel: T) extends Exception

// optional:
case class PetModelNotAllowed extends ApiException[ErrorModel]
```
but the main point is to have httpCode to Model mapping specified in api specification as it provided by OpenApiSpec

michalo2882: Thanks for your input, however you seem to be reviewing the original commit. In 77025d8 I got rid of `ErrorModel`, because after all it doesn't make any sense - there is no official model for error, so now it will only handle the error content if it's explicitly specified in the spec, for example:
```yml
responses:
  '200':
    content:
      application/xml:
        schema:
          $ref: '#/components/schemas/Pet'
      application/json:
        schema:
          $ref: '#/components/schemas/Pet'
  '404':
    description: Pet not found
    content:
      application/xml:
        schema:
          $ref: '#/components/schemas/Error'
      application/json:
        schema:
          $ref: '#/components/schemas/Error'
```

modules/openapi-generator/src/main/resources/scala-sttp/project/build.properties.mustache
```diff
@@ -1 +1 @@
-sbt.version=1.3.12
+sbt.version=1.4.3
```
eugeniyk, line 1:
```suggestion
sbt.version=1.4.5
```

docs/generators/scala-sttp.md
```diff
@@ -14,7 +14,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |disallowAdditionalPropertiesIfNotPresent|If false, the 'additionalProperties' implementation (set to true by default) is compliant with the OAS and JSON schema specifications. If true (default), keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.|<dl><dt>**false**</dt><dd>The 'additionalProperties' implementation is compliant with the OAS and JSON schema specifications.</dd><dt>**true**</dt><dd>Keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.</dd></dl>|true|
 |ensureUniqueParams|Whether to ensure parameter names are unique in an operation (rename parameters that are not).| |true|
 |jodaTimeVersion|The version of joda-time library| |2.10.6|
```
eugeniyk, line 16: Let's also migrate to java8 DateTime instead?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
```diff
@@ -212,21 +213,25 @@ public String escapeReservedWord(String name) {
...
+                        if (response.isDefault) {
                             continue;
                         }
                         try {
                             int code = Integer.parseInt(response.code);
+                            hasOnlyDefaultResponse = false;
                             if (code >= 600) {
                                 unknownCodes.add(code);
                             }
+                            response.vendorExtensions.put("x-error-code", response.is4xx || response.is5xx);
```
eugeniyk, line 229: what does `vendorExtensions` provide?

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/api/PetApi.scala
```diff
@@ -32,13 +33,15 @@ class PetApi(baseUrl: String) {
    * 
    * @param pet Pet object that needs to be added to the store
    */
-  def addPet(pet: Pet
-): Request[Either[ResponseError[Exception], Pet], Nothing] =
+  def addPet(pet: Pet): Request[Either[ResponseException[ApiModel, Exception], Pet], Any] =
```
eugeniyk, line 36: I think for `Left` case we should have more concrete generated type such as `ExceptionApiModel`
```scala
sealed trait ApiModel
case class Pet extends ApiModel
case class PetNotFound extends ExceptionApiModel
```

```suggestion
  def addPet(pet: Pet): Request[Either[ResponseException[ExceptionApiModel, Exception], Pet], Any] =
```

michalo2882: I'm not sure if the change would be valid. The same model can be used for handling both 200 and 400 responses in theory.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
michalo2882: It's used to determine whether response should be `Left` or `Right` in template generator. See `ignoredResponseReturnType.mustache`. Is there any better way to achieve this?

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/api/PetApi.scala
adamw: if `ApiModel` and `ExceptionApiModel` are traits (by default corresponding to 2xx and 4xx/5xx responses), then if a concrete model class is re-used, it could implemet both of these traits?

docs/generators/scala-sttp.md
michalo2882: 👍

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaSttpClientCodegen.java
eugeniyk: I see; no idea tbh never used mustache

modules/openapi-generator/src/main/resources/scala-sttp/apiModel.mustache
```diff
@@ -2,4 +2,6 @@
 package {{invokerPackage}}
 
 trait ApiModel
+trait ExceptionApiModel
 case object EmptyApiModel extends ApiModel {}
+case object EmptyExceptionApiModel extends ExceptionApiModel {}
```
eugeniyk: I think would be nice to add comments here since it will be part of public API

modules/openapi-generator/src/main/resources/scala-sttp/model.mustache
```diff
@@ -24,7 +24,7 @@ case class {{classname}}(
   {{/description}}
   {{{name}}}: {{^required}}Option[{{/required}}{{^isEnum}}{{dataType}}{{/isEnum}}{{#isEnum}}{{classname}}Enums.{{datatypeWithEnum}}{{/isEnum}}{{^required}}] = None{{/required}}{{^-last}},{{/-last}}
   {{/vars}}
-) extends ApiModel
+) extends ApiModel with ExceptionApiModel
```
eugeniyk, line 27: imo `ExceptionApiModel` should just extend `ApiModel`

modules/openapi-generator/src/main/resources/scala-sttp/apiModel.mustache
```diff
@@ -2,4 +2,6 @@
 package {{invokerPackage}}
 
 trait ApiModel
+trait ExceptionApiModel
```
eugeniyk, line 5:
```suggestion
trait ExceptionApiModel extends ApiModel
```

samples/client/petstore/scala-sttp/src/main/scala/org/openapitools/client/model/User.scala
```diff
@@ -27,6 +27,6 @@ case class User(
   phone: Option[String] = None,
   /* User Status */
   userStatus: Option[Int] = None
-) extends ApiModel
+) extends ApiModel with ExceptionApiModel
```
eugeniyk, line 30: and howcome user is part of `ExceptionApiModel`? Like when you create new user and post is unsuccessful you will return existing user in response body?

michalo2882: I didn't find any example on how to enrich the model metadata with the info on whether or not it is used for processing 4xx codes. That's why all models are also `ExceptionApiModel`.

chameleon82: @michalo2882 could you please provide an example how to map response into results?
Let me give an example with models i expect:
```scala
case class Pet(...) extends ApiModel
case class PetBadRequest(code: Int, message: String) extends ExceptionApiModel
case class PetValidationFailed(code: Int, message: String, errors: Seq[String]) extends ExceptionApiModel
```
and spec mapping:
```
200 -> Pet
400 -> PetNotFound
422 -> PetValidationFailed
```

I expect something like this
```scala
val api = new PetApi(...)
val pet = Pet(...)

val response = api.addPet(pet)

response.mapResponse { 
  case Right(createdPet) => createdPet
  case Left(responseException: ResponseException[PetBadRequest, Exception]) => (response.model.code, response.model.message)
  case Left(responseException: ResponseException[PetValidationFailed, Exception]) => response.model.errors.foreach(println)
  case Left(responseException: DeserializationException[Exception]) => responseException.message
  case Left(responseException: HttpError[Exception]) => responseException.statusCode
}
```

PR #8594, opened 2021-02-01
[kotlin] fix Date types usages

4brunu:
`AbstractKotlinCodegen.java` has some issues related to date types.

The first one is that in the `typeMapping` it defines `Date` as `java.time.LocalDate` which is correct.
But then in `importMapping` it defines `Date` as `java.time.LocalDate`  which is incorrect and doesn't match with the `typeMapping`
It should define `Date` as `java.time.LocalDate` in the `typeMapping`.

The second one is that it defines `DateTime` as `java.time.LocalDateTime` in the `typeMapping` and in the `importMapping`, which is wrong, it should be `java.time.OffsetDateTime`.
This can cause bugs because the servers and the clients can (and most probably are) in different timezones.
This is already fixed in `KotlinSpringServerCodegen` and `KotlinClientCodegen` but not in `AbstractKotlinCodegen`.

Finally since `java.util.Date` is never used, it's not necessary in the  Kotlin Client `Serializer`, so it's removed.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


4brunu:
This PR has lots of changes, because I generate the samples `server` and `openapi3`, which were really outdated, because they are "hidden" in the directories `bin/configs/other/openapi3/kotlin*` and `bin/configs/other/kotlin*`.

jimschubert:
I was going to ask about the  `DateTime` default from `java.time.LocalDateTime` to `java.time.OffsetDateTime`, but then I thought about it more and it's really the only type that fully supports the spec's date-time requirement to match [rfc3339](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14).

We'll need to consider this a bug fix in that case. Folks may need to pass type-mappings to override for backward compatibility with previously generated code.

jimschubert (approved):
LGTM

PR #8597, opened 2021-02-02
[dart][dart-dio] Improve API code & wrap serialization errors

kuhnroyal:
### Background
Currently during (de)serialization errors, four types of exceptions can be thrown:
* `StateError` - during (de)serialization when built_value can not find a matching serializer (very generic dart:core error, can be thrown from almost anything)
* `FormatException` - `json.decode` failed (very generic dart:core error, can be thrown from almost anything)
* `JsonUnsupportedObjectError` - `json.encode` failed
* `DeserializationError` - built_value deserialization failed

When implementing a general error handling in a consuming Dart application it is not clear that these errors are thrown by the API client as most of these can be thrown from other sources, especially when using built_value for other things as well.

By default when using dio with plain `json.encode/decode` all (de)serialization exceptions would be wrapped in `DioError` instances but since our built_value (de)serialization occurs before/after the dio call this is not the case.

The errors need to be manually wrapped in a `DioError` instance. To achieve this the API method call code has to be restructured:
* first construct a `RequestOptions` instance with all parameters etc. inline (this can be used to fill the `DioError` in case of serialization failure)
* then construct and serialize the body data
* then send the request and await the response
* then deserialize the response
* always use the `specifiedType` for built_value (de)serialization instead of a mix of ways (this does the same as finding the serializer directly)
* wrap (de)serialization in try/catch and throw a `DioError` (this uses the new `RequestOptions` and possible the response instances) - generic errors are now wrapped in a `DioError` which makes it clear that they originated from serialization code inside an API call

### Breaking change
The API call signature doesn't change but the type of thrown errors changes. For users who are checking for those errors it could be a breaking change.
 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
+++ b/modules/openapi-generator/src/main/resources/dart-dio/api.mustache
@@ -32,132 +32,133 @@ class {{classname}} {
         ProgressCallback onSendProgress,
         ProgressCallback onReceiveProgress,
     }) async {
-        final String _path = '{{{path}}}'{{#pathParams}}.replaceAll('{' r'{{baseName}}' '}', {{{paramName}}}.toString()){{/pathParams}};
+        final _request = RequestOptions(
+          path: r'{{{path}}}'{{#pathParams}}.replaceAll('{' r'{{{baseName}}}' '}', {{{paramName}}}.toString()){{/pathParams}},
+          method: '{{#lambda.uppercase}}{{httpMethod}}{{/lambda.uppercase}}',
+          {{#isResponseFile}}
+          responseType: ResponseType.bytes,
+          {{/isResponseFile}}
+          headers: <String, dynamic>{
+            {{#httpUserAgent}}
+            r'User-Agent': r'{{{.}}}',
+            {{/httpUserAgent}}
+            {{#headerParams}}
+            r'{{baseName}}': {{paramName}},
+            {{/headerParams}}
+            ...?headers,
+          }..removeWhere((_, dynamic value) => value == null),
+          queryParameters: <String, dynamic>{
+            {{#queryParams}}
+            r'{{baseName}}': {{paramName}},
+            {{/queryParams}}
+          }..removeWhere((_, dynamic value) => value == null),
+          extra: <String, dynamic>{
+            'secure': <Map<String, String>>[{{^hasAuthMethods}}],{{/hasAuthMethods}}{{#hasAuthMethods}}
+              {{#authMethods}} {
+                'type': '{{type}}',
+                'name': '{{name}}',{{#isApiKey}}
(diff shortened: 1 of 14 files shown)


kuhnroyal:
@josh-burton I would like your input on this and it would be great if you could test this on some project if possible at all.

CC @agilob 

I am unsure about the breaking change that I have documented and how much that would actually affect consumers. Maybe we should target this into 5.1.0 but there is no fallback, so in theory this is for 6.0.0 which is too far off.
@wing328 What do you think?

wing328:
> The API call signature doesn't change but the type of thrown errors changes. For users who are checking for those errors it could be a breaking change.

@kuhnroyal can it be an option to choose which behaviour (new or existing) to use? 

For R, we've 2 options related to error handling:

`exceptionPackage`: Specify the exception handling package	

`returnExceptionOnFailure`: Throw an exception on non success response codes	


Ref: https://openapi-generator.tech/docs/generators/r/

kuhnroyal:
> > The API call signature doesn't change but the type of thrown errors changes. For users who are checking for those errors it could be a breaking change.
> 
> @kuhnroyal can it be an option to choose which behaviour (new or existing) to use?
> 
> For R, we've 2 options related to error handling:
> 
> `exceptionPackage`: Specify the exception handling package
> 
> `returnExceptionOnFailure`: Throw an exception on non success response codes
> 
> Ref: https://openapi-generator.tech/docs/generators/r/

I could add a generator flag (something like `wrapSerializationErrors`) but I would also argue that the changes are the only logical option in the future. So that switch would just exist to fallback to the existing behavior for 5.0.1 and be removed in 5.1.0?

This is also not about errors for HTTP response codes, the dio library already throws errors by default for non-success codes. This change only addresses runtime (de)serialization errors to match that same behavior which was intended by dio for their default serialization mechanism.

wing328:
The reason why I suggest an option is that the change can get into 5.1.0 release (5.0.1 will be released this week). Otherwise, you will need to wait for 6.0.0 release (which is months away from now).

In 5.1.0 release, you can set it as the default as users can easily fallback to existing behaviour if they want to.

kuhnroyal:
> The reason why I suggest an option is that the change can get into 5.1.0 release (5.0.1 will be released this week). Otherwise, you will need to wait for 6.0.0 release (which is months away from now).
> 
> In 5.1.0 release, you can set it as the default as users can easily fallback to existing behaviour if they want to.

Sounds good, will do that.

wing328:
FYI. I'm syncing master to 5.1. and 6.0.x

kuhnroyal:
Gonna split this up in 2 new PRs

Issue #8552, opened 2021-01-27
[BUG][SCALA] Akka-client cannot be build agains scala 2.13.x version


jaroslawZawila:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [ ] Have you searched for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

I am trying to generate akka-client for scala 2.13 and it fails as it cannot find the underlying dependencies.

```
com.typesafe.akka/akka-http_2.13
org.json4s/json4s-jackson_2.13
org.json4s/json4s-ext_2.13
org.scalatest/scalatest_2.13
com.typesafe.akka/akka-stream_2.13
com.typesafe.akka/akka-actor_2.13
de.heikoseeberger/akka-http-json4s_2.13
```

##### OpenAPI declaration file content or url

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/scala-akka-client/build.sbt.mustache
```yaml
(here your code)
```
(for YAML code) or
```json
(here your code)
```
(for JSON code), so it becomes more readable. If it is longer than about ten lines,
please create a Gist (https://gist.github.com) or upload it somewhere else and
link it here.
  -->

##### Suggest a fix

Update above dependencies to below;
```
  "com.typesafe" % "config" % "1.4.1",
  "com.typesafe.akka" %% "akka-actor" % "2.5.24",
  "com.typesafe.akka" %% "akka-stream" % "2.5.24",
  "com.typesafe.akka" %% "akka-http" % "10.1.9",
{{#joda}}
  "joda-time" % "joda-time" % "2.10.1",
{{/joda}}
  "org.json4s" %% "json4s-jackson" % "3.6.7",
  "org.json4s" %% "json4s-ext" % "3.6.7",
  "de.heikoseeberger" %% "akka-http-json4s" % "1.27.0",
  // test dependencies
[...]

PR #8624, opened 2021-02-05
[scala][akka] Update dependencies for scala 2.13 in scala-akka

ex0ns:
The generated files for the scala-akka binding are not compatible with scala 2.13
This issue was raised in https://github.com/OpenAPITools/openapi-generator/issues/8552

I used the versions lised on this issues, but I think that we could event use more recent version of akka/akka_http, what do you think ?

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [x] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master` (I'm not sure that this is the correct branch, might be 5.1 instead)

Could you please review this

@clasnake (2017/07), @jimschubert (2017/09) ❤️, @shijinkui (2018/01), @ramzimaalej (2018/03), @chameleon82 (2020/03), @Bouillie (2020/04)


ex0ns:
We need to implement the support for multiple scala version inside `maven` or switch to sbt as `scala-sttp` does.
I tried to implement the first version by having multiple passes in the `pom.xml` but there is a strange behavior when generating the template, changes to `modules/openapi-generator/src/main/resources/scala-akka-client/pom.mustache` are not visible in the generated ` samples/client/petstore/scala-akka/pom.xml` after running `./bin/generate-samples.sh bin/configs/scala*` 

Is there a special processing for the `pom.mustache` file ?

wing328:
> could you please clarify why this was added ? Thanks

Could be due to some manual changes in pom.xml to run the integration tests (used in the CI)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
```diff
@@ -166,7 +166,6 @@ public void processOpts() {
         additionalProperties.put(CodegenConstants.INVOKER_PACKAGE, invokerPackage);
 
         supportingFiles.add(new SupportingFile("README.mustache", "", "README.md"));
-        supportingFiles.add(new SupportingFile("pom.mustache", "", "pom.xml"));
```
wing328, line 169: @ex0ns should this change be reverted?

ex0ns:
I'll come back to this pull request soon, but I think that in order to support compilation for  multiple scala's versions, we will have to use sbt here and replace the `pom.xml` content by some generic content (such as the one in the sttp generator).

I have successfully used the new template in a 2.13 project, but I'm still facing some issues regarding the compatibility with 2.12

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
ex0ns: I'm not sure, the goal is to compile for both 2.13 and 2.12 which is not possible/easy with maven.
I migrated this codegen to sbt.

ex0ns:
I am finally back with an updated version that uses akka 2.6 and that fixes a lot of warning with 2.13.
The codegen seems to be working fine:
```sh
./bin/generate-samples.sh ./bin/configs/scala-akka.yaml
```
And then run sbt in the generated `samples/client/petstore/scala-akka`

```sh
sbt '+compile'
```

I'm however not able to successfully run `sbt test` in this folder, but I might be missing something (because even on the source branch those tests are not running for me)

samples/client/petstore/scala-akka/src/test/scala/PetApiTest.scala
```diff
@@ -118,6 +119,5 @@ class PetApiTest extends AsyncFlatSpec with Matchers {
         }
       }
   }
-  */
+   */
```
ramzimaalej, line 122: Can you remove this commented out block as well please?

ramzimaalej (changes requested):
One minor comment in addition to fixing the broken test.

samples/client/petstore/scala-akka/src/test/scala/PetApiTest.scala
ex0ns: That is a good catch, thanks. 
On the other hand I was not able to reproduce the CircleCI bug and I would really like some help on that, I'm not sure to understand what the `circe_parallel` script is supposed to do, but I can't run it locally.

ex0ns:
This is ready to be review, could someone take a look ?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/ScalaAkkaClientCodegen.java
wing328: the next version is a patch release so removing this file seems to be a breaking change.

can we simply keep it and update the dependency versions?

ex0ns: I've added the pom.xml back (as well as the pom.mustache) :ok_hand:

wing328:
Tested locally and the output compiles without issues.

PR #8660, opened 2021-02-09
fix: remove double URL encoding

zapek:
DefaultWebClient() already encodes URL query parameters.

Without this PR, I have a request with title=foo bar which gets converted to title=foo%25%20bar.
With this PR, the request becomes title=foo%20bar

This fixes #8171

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/webclient/ApiClient.mustache
@@ -573,7 +573,7 @@ public class ApiClient{{#jsr310}} extends JavaTimeFormatter{{/jsr310}} {
             builder.queryParams(queryParams);
         }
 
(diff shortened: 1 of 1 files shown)


wing328:
Tested locally and the result is good:
```
HTTP: GET /v2/pet/12345?query_test=foo%20bar HTTP/1.1
``

wing328:
Filed https://github.com/OpenAPITools/openapi-generator/pull/8972 to update `webclient` dependencies. Please kindly review when you've time.

wing328:
@zapek thanks for the PR, which has been included in the v5.1.0 release: https://twitter.com/oas_generator/status/1373636654024380423

fbecaria:
@wing328 A question regarding this issue, as fixed the example you mention before, I see that fails if the query param include an "&", is this correct? My solution so far is to remove encoding at all and encode query params manually.

Issue #8548, opened 2021-01-27
[BUG][JAVA] import mapping of alias types in API interface


PhilippParis:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [ ] Have you searched for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

when using importMapping for alias types, the generated code does not use the importmapping in the api interface

##### Expected

```
    @POST
    @Path("/{id}")
    Response typeAliasIdPost(@PathParam("id") foo.bar.TypeAlias id);
```

##### Actual
```
    @POST
    @Path("/{id}")
    Response typeAliasIdPost(@PathParam("id") String id);
```

##### openapi-generator version

5.0.0

##### OpenAPI declaration file content or url

https://gist.github.com/PhilippParis/28c40d590b4a9019cb8040b143e09884

##### Generation Details

```
<generatorName>jaxrs-spec</generatorName>
<interfaceOnly>true</interfaceOnly>
 <java8>true</java8>
<importMappings>
    <importMapping>TypeAlias=foo.bar.TypeAlias</importMapping>
</importMappings>
```

##### Steps to reproduce

run openapi-generator 5.0.0 with specified generation details

##### Related issues/PRs

https://github.com/OpenAPITools/openapi-generator/issues/3589

borsch:
The problem here is that you describe this object by your own here https://gist.github.com/PhilippParis/28c40d590b4a9019cb8040b143e09884#file-openapi-yaml-L36

to work with alias do the following
1. don't create you own definition for alias in yaml file
2. in place where you need to use that alias don't `$ref` them, but just use `type: TypeAlias`

At least this is how I configured alias for my project. Version of openapi-maven-plugin-generator is 4.3.1
Example:


```yaml
/type-alias/{id}:
    post:
      parameters:
        - name: id
          in: path
          required: true
          schema:
            type: TypeAlias     <--
      responses:
        200:
          description: OK
```

[...]

PR #8661, opened 2021-02-09
Allow custom import to override models from openapi specification

borsch:
Related issue: https://github.com/OpenAPITools/openapi-generator/issues/8548

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -432,17 +432,17 @@ void generateModels(List<File> files, List<Object> allModels, List<String> unuse
...
+//                if (config.importMapping().containsKey(name)) {
+//                    LOGGER.debug("Model {} not imported due to import mapping", name);
+//
+//                    for (String templateName : config.modelTemplateFiles().keySet()) {
+//                        // HACK: Because this returns early, could lead to some invalid model reporting.
+//                        String filename = config.modelFilename(templateName, name);
+//                        Path path = java.nio.file.Paths.get(filename);
+//                        this.templateProcessor.skip(path,"Skipped prior to model processing due to import mapping conflict (either by user or by generator)." );
+//                    }
+//                    continue;
```
borsch, line 444: Question to maintainers: does it make sense to do this stuff configurable? Like introduce property this will turn this feature on/off

borsch:

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
Zomzog: I'm not sure to understand why you want to remove this part.

borsch: Technically there can be a case, when client of yaml/json specification would like you you his own DTO instead of generated by openapi specification

Regarding solution: I would prefer to introduce configuration parameter so that client can configure this behaviour

Zomzog: It's the case of this block, skip models override in the importMapping.
It's also possible to just skip all model creation and just use the right `modelPackage` if his own DTO has the same name like this :
```
<modelPackage>my.package.of.dto</modelPackage>
<generateModels>false</generateModels>
```

borsch: @Zomzog idea in original issue was to allow custom override for some of models and not all

Zomzog: Yep and model per model it's the `importMapping` option. But I think I missing something. Could you give me an example where this part creates issues? The related example doesn't work but I think for another reason.

borsch: @Zomzog https://github.com/OpenAPITools/openapi-generator/blob/326d890c4a1882953f88dabe828244100b1bb9ed/modules/openapi-generator/src/test/resources/3_0/issue_8548.yaml

example from `samples/server/petstore/spring-mvc-custom-import-with-override` contains custom import mapping for object `CustomObject` and therefore this object won't be generated(https://github.com/OpenAPITools/openapi-generator/pull/8661/files#diff-21792377bac6a73df584a020e525e172900373ab1ff04f3112fe08df88183878R24)

example from `samples/server/petstore/spring-mvc-custom-import-without-override` doesn't have override and therefore `CustomObject` will be generated by generator

Not sure why but this commented peace of code doesn't allow partial model override

wing328:
cc @OpenAPITools/generator-core-team since the change impacts the default generator

PR #8693, opened 2021-02-12
Rust: Add support for lists (and sets) of enum values

gferon:
For a field described like:

```yaml
matrixAttributes:
  type: array
  uniqueItems: true
  items:
    type: string
    enum: ["travelTimes", "distances"]
  default: ["travelTimes"]
```

the generated field in Rust would not be correct and was:

```rust
#[serde(rename = "matrixAttributes", skip_serializing_if = "Option::is_none")]
pub matrix_attributes: Option<MatrixAttributes>,
```

with this patch, it is now correctly wrapped in a collection, depending on whether `uniqueItems` is set or not:

For a list:

```rust
#[serde(rename = "matrixAttributes", skip_serializing_if = "Option::is_none")]
pub matrix_attributes: Option<Vec<MatrixAttributes>>,
```

For a set:

```rust
#[serde(rename = "matrixAttributes", skip_serializing_if = "Option::is_none")]
pub matrix_attributes: Option<std::collections::HashSet<MatrixAttributes>>,
```

cc @frol (2017/07) @farcaller (2017/08) @richardwhiuk (2019/07) @paladinzh (2020/05) for review

### DIFF
--- a/modules/openapi-generator/src/main/resources/rust/model.mustache
+++ b/modules/openapi-generator/src/main/resources/rust/model.mustache
@@ -51,7 +51,7 @@ pub struct {{{classname}}} {
     /// {{{description}}}
     {{/description}}
     #[serde(rename = "{{{baseName}}}"{{^required}}, skip_serializing_if = "Option::is_none"{{/required}})]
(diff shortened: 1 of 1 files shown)


wing328:
```
  type: array
  uniqueItems: true
```
@gferon thanks for the PR. Are we on the same page that this is a set instead of array?

I've not checked if we have correctly mapped this to a proper set type in Rust in the generator.

gferon:
@wing328 great catch! you're absolutely right, I've added a test for `{{#uniqueItems}}` now to support both sets and lists of enum values.

wing328 (approved):

wing328:
@gferon thanks for the PR, which has been included in the v5.1.0 release: https://twitter.com/oas_generator/status/1373636654024380423

Issue #8258, opened 2020-12-22
[Feature] [TYPESCRIPT] Generator parses to Array instead of Set


SchillingSebastian:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

If I generate a **typescript-angular** client with the latest 5.0 stable release, query the mentioned endpoint and try to use `user.authority.has(UserAuthorityEnum.Admin)` within a component I get the following error. (In Both Template/Component the Error occurs)
```
ERROR TypeError: _angular_core__WEBPACK_IMPORTED_MODULE_0__.ɵɵpipeBind1(...).authority.has is not a function
    at DashboardComponent_Template (main.js:2043)
    at executeTemplate (vendor.js:65785)
    at refreshView (vendor.js:65654)
    at refreshComponent (vendor.js:66801)
    at refreshChildComponents (vendor.js:65460)
    at refreshView (vendor.js:65704)
    at refreshEmbeddedViews (vendor.js:66755)
    at refreshView (vendor.js:65678)
    at refreshComponent (vendor.js:66801)
    at refreshChildComponents (vendor.js:65460)
    at refreshView (vendor.js:65704)
    at refreshEmbeddedViews (vendor.js:66755)
    at refreshView (vendor.js:65678)
    at refreshComponent (vendor.js:66801)
    at refreshChildComponents (vendor.js:65460)
```
With further investigation it seems like the field authority is an Array instead of a Set. I am not sure if this is a bug or if this is a result of a faulty spec. I validated with the cli validator, as well as the swagger editor and the referenced validator. 
This behaviour was also present in the 5.0 beta 3 and I hoped it would be solved with the stable release.

Can anyone point me in the right direction?
If there are any more specs/configs I can help you with, don't hesitate to ask for them.

##### openapi-generator version

Latest stable release (5.0.0)

##### OpenAPI declaration file content or url
```yaml
openapi: 3.0.1
info:
  title: Some Api
  description: Some Desc
  version: 0.2.0
servers:
[...]

PR #8695, opened 2021-02-13
[typescript-fetch] handle uniqueItems correctly in model and api

t-h-e:
uniqueItems produces Set instead of Array as type, but the corresponding code did not treat Set correctly

use `Array.from` and `new Set` to convert from Set to Array and vice versa, if uniqueItems is true.

related to https://github.com/OpenAPITools/openapi-generator/issues/8258, but only fixes typescript-fetch

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache
@@ -109,7 +109,7 @@ export class {{classname}} extends runtime.BaseAPI {
             queryParameters['{{baseName}}'] = requestParameters.{{paramName}};
             {{/isCollectionFormatMulti}}
             {{^isCollectionFormatMulti}}
-            queryParameters['{{baseName}}'] = requestParameters.{{paramName}}.join(runtime.COLLECTION_FORMATS["{{collectionFormat}}"]);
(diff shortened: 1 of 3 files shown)


macjohnny:
@t-h-e thanks for your contribution. please follow the steps described in https://github.com/OpenAPITools/openapi-generator/blob/master/.github/PULL_REQUEST_TEMPLATE.md

macjohnny (commented):
the code looks good, but I think this might be a breaking change if currently someone uses a type-mapping. I am not sure how we should proceed. @wing328 any thoughts?

t-h-e:
@macjohnny Thanks for the hint. I updated the samples.

As @macjohnny mentioned correctly, this change will break setups, where type-mapping has been used.

I think I found another issue, unless I'm mistaken. Pet.ts in samples/client/petstore/typescript-fetch/builds/default-3.0/models may not work correctly in line 94 and 112, as photoUrls of type Set<string> is treated as primitive. This worked when photoUrls was of type Array<string>, but not with Set.

wing328:
For other generators, we do not consider it as a breaking change as we're making a correction to the output. Let's see if any TS users have a different opinion on this.

macjohnny:
Cc @TiFu, @taxpon, @sebastianhaas, @kenisteward, @Vrolijkx, @macjohnny, @topce, @akehir, @petejohansonxo, @amakhrov

wing328:
@t-h-e thanks for the PR, which has been included in the v5.1.0 release: https://twitter.com/oas_generator/status/1373636654024380423

Issue #6192, opened 2020-05-06
[BUG] Type for map>array>item not generated


bard:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] What's the version of OpenAPI Generator used? 4.3.1
- [x] Have you search for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Bounty to sponsor the fix ([example](https://www.bountysource.com/issues/66123212-javascript-client-produces-a-wrong-object-for-a-string-enum-type-that-is-used-with-ref))

##### Description

When an item of an array of a map is specified through a referenced schema, the type for the item is not generated. 

For example, for an endpoint specification of:

```yaml
  "/characters":
    get:
      responses:
        "200":
          content:
            application/json:
              schema:
                additionalProperties:
                  type: array
                  items:
                    $ref: "#/components/schemas/Character"
```

And a schema specification of:

```yaml
    Character:
      required: ["first", "name"]
      properties:
        id:
          type: integer
        name:
          type: string
```

The `Character` type is not generated:

```
api/apis/DefaultApi.ts:25:82 - error TS2304: Cannot find name 'Character'.

25     async charactersGetRaw(): Promise<runtime.ApiResponse<{ [key: string]: Array<Character>; }>> {
                                                                                    ~~~~~~~~~

api/apis/DefaultApi.ts:42:59 - error TS2304: Cannot find name 'Character'.

42     async charactersGet(): Promise<{ [key: string]: Array<Character>; }> {
                                                             ~~~~~~~~~


Found 2 errors.
```

##### openapi-generator version

4.3.1

##### OpenAPI declaration file content or url

https://github.com/bard/openapi-generator-missing-type/blob/master/example.yaml

Excerpt:

```yaml
  "/characters":
    get:
      responses:
        "200":
          content:
            application/json:
              schema:
                additionalProperties:
                  type: array
                  items:
                    $ref: "#/components/schemas/Character"
```

##### Command line used for generation

```sh
openapi-generator generate -g typescript-fetch --additional-properties=typescriptThreePlus=true -i example.yaml -o api
```

##### Steps to reproduce

1. Clone https://github.com/bard/openapi-generator-missing-type
2. Run `yarn install && yarn test`

##### Related issues/PRs

[...]

PR #8699, opened 2021-02-14
Add object type to imports when value of map is array of objects

rgroothuijsen:
This PR fixes a bug that occurs when a method returns a map containing arrays of custom objects, for example `Map<int, TestObject[]`>. While `TestObject` in this case is recognized mostly correctly, it is not added to the imports because for the imports it only looks at the array itself, rather than the type of the values held in the array. The result is that the user needs to manually add the import to the generated code.

The parameterized type, in this case, appears to be instead located in `additionalProperties.complexType` and should be added to the imports whenever the container type is a map and the map's value type is not a built-in type.

This is a fix for issue #6192.


hg:
I was looking into this myself. This solution is probably not generic enough — the current generator misses imports for:

- key to model: `{ [key: string]: TestModel; }`
- key to array of models: `{ [key: string]: Array<TestModel>; }`
- nested maps: `{ [key: string]: { [key: string]: Array<TestModel>; }; }`

and so on.

I this the more universal solution would look something like this:

```java
for (CodegenProperty addProp = r.additionalProperties;
     addProp != null;
     addProp = addProp.additionalProperties
) {
    if (addProp.complexType != null) {
        imports.add(addProp.complexType);
    } else if (addProp.baseType != null) {
        if (!defaultIncludes.contains(addProp.baseType) &&
                !languageSpecificPrimitives.contains(addProp.baseType)) {
            imports.add(addProp.baseType);
        }
    }
}
```

although I'm sure this does not cover everything (at least the tests seem to be passing).

@rgroothuijsen would you like to update this PR, or should I open a separate one? Please see attached the schemas I tested with.

[specs.zip](https://github.com/OpenAPITools/openapi-generator/files/6056292/specs.zip)

rgroothuijsen:
@hg Thanks for the test cases, I'll have a look at them. I figured I would keep the fix specific to the problem I was having, because I didn't know if making more general changes would produce unexpected side effects. Personally I was able to reproduce the second and third case in my scenario, though the first case (key to model) was in fact working for me.

rgroothuijsen:
@hg Turns out you were right about it not being general enough. Your test cases failed on my code, so I've changed my code. It now recursively checks nested `additionalProperties` for data types that can be added to the imports.

SKF-SIXT:
Are there any updates on this PR? We are facing the same problem. Can we help to solve it?

PR #8772, opened 2021-02-19
[typescript-axios] Support array query arguments

### LINKED ISSUE #7966: [BUG][typescript-axios] Multi-value query params are incorrectly encoded as a comma separated string
#### Bug Report Checklist

- [X] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [X] Have you searched for related issues/PRs?

##### Description

The generated code improperly converts query list params to comma-separated strings.

If passing an array to a query param - eg `include=["relation_1", "relation_2"]` I expect it would output `include=relation_1&include=relation_2` in the query string. Instead, it returns `include=relation_1,relation_2`.

* `5.0.0-beta2`

If you post the code inline, please wrap it with
```yaml
info:
  title: API
  version: 1.0.0
openapi: 3.0.0
paths:
  /some-resource:
    get:
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                properties:
                  id:
                    type: string
                type: object
      parameters:
        - in: query
          name: include
          required: false
          schema:
            default: []
            type: array
            items:
              enum:
                - relation_1
                - relation_2
              type: string
```
##### Generation Details

Use [typescript-axios] with default options.

```yaml
generatorName: typescript-axios
```

##### Steps to reproduce

1. Run generator
2. Call `await DefaultApiAxiosParamCreator().someResourceGet(["relation_1", "relation_2"])`.
3. Observe that called url has `include=relation_1,relation_2` in the querystring.

##### Suggest a fix
The problematic code is here:

https://github.com/OpenAPITools/openapi-generator/blob/v5.0.0-beta2/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache#L191-L194

```typescript
const localVarQueryParameter = {} as any;

if (include) {
    localVarQueryParameter['include'] = include;
}

const query = new URLSearchParams(localVarUrlObj.search);
for (const key in localVarQueryParameter) {
    query.set(key, localVarQueryParameter[key]);
}
```

This line `query.set(key, localVarQueryParameter[key]);`  converts the second argument to a string, so the array gets converted to `relation_1,relation_2`.

The proper approach would be to check if the value of the given `localVarQueryParameter` is an array, and if so, to loop over it and append, eg:

```
for (const key in localVarQueryParameter) {
    if(Array.isArray(localVarQueryParameter[key])) {
        localVarQueryParameter[key].forEach(v => {
            query.append(key, v)
        })
    } else {
        query.set(key, localVarQueryParameter[key]);
    }
}
```

### LINKED ISSUE #7973: [BUG] [typescript-axios] Query parameters do not support `explode: true, style: form`
#### Bug Report Checklist

- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?

##### Description

OpenAPI supports parameters defined with the following.

```yaml
name: color
in: query
style: form
explode: true
```
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#style-values

Parameters defined like this and given an array output query strings like `color=blue&color=black&color=brown`.

However currently a typescript-axios client made by openapi-generator would instead output `color=blue%2Cblack%2Cbrown`.

```
$ openapi-generator version
5.0.0-beta2
```

```yaml
parameters:
- name: color
  in: query
  style: form
  explode: true
  schema:
    type: array
    item:
      type: array
        - in: query
          name: Balance.Currency
          style: form
          explode: true
          schema:
            type: string
```
##### Generation Details

```
openapi-generator generate -i openapi.yaml -g typescript-axios -c config.yaml
```

##### Suggest a fix

I think that apiInner.mustache needs to be updated to include handling for this type of collection format.

https://github.com/OpenAPITools/openapi-generator/blob/78611b3e387128b5f768865e4dfa8b4736ba571c/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache#L107-L138

dantman:
typescript-axios does not correctly output query arguments when you use `explode: true` for a parameter.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
  - CC: @TiFu @taxpon  @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov 

fixes #7966
fixes #7973

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-axios/common.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/common.mustache
@@ -83,7 +83,7 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
             if (Array.isArray(object[key])) {
                 searchParams.delete(key);
                 for (const item of object[key]) {
-                    searchParams.set(key, item);
+                    searchParams.append(key, item);
                 }
             } else {
                 searchParams.set(key, object[key]);
--- a/modules/openapi-generator/src/main/resources/typescript-axios/common.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-axios/common.mustache
@@ -80,7 +80,14 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
     const searchParams = new URLSearchParams(url.search);
     for (const object of objects) {
         for (const key in object) {
-            searchParams.set(key, object[key]);
+            if (Array.isArray(object[key])) {
+                searchParams.delete(key);
+                for (const item of object[key]) {
+                    searchParams.set(key, item);
+                }
+            } else {
+                searchParams.set(key, object[key]);
+            }
         }
     }
     url.search = searchParams.toString();
--- a/samples/client/petstore/typescript-axios/builds/composed-schemas/common.ts
+++ b/samples/client/petstore/typescript-axios/builds/composed-schemas/common.ts
@@ -94,7 +94,7 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
             if (Array.isArray(object[key])) {
                 searchParams.delete(key);
                 for (const item of object[key]) {
-                    searchParams.set(key, item);
+                    searchParams.append(key, item);
                 }
             } else {
                 searchParams.set(key, object[key]);
--- a/samples/client/petstore/typescript-axios/builds/composed-schemas/common.ts
+++ b/samples/client/petstore/typescript-axios/builds/composed-schemas/common.ts
@@ -91,7 +91,14 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
     const searchParams = new URLSearchParams(url.search);
     for (const object of objects) {
         for (const key in object) {
-            searchParams.set(key, object[key]);
+            if (Array.isArray(object[key])) {
+                searchParams.delete(key);
+                for (const item of object[key]) {
+                    searchParams.set(key, item);
+                }
+            } else {
+                searchParams.set(key, object[key]);
+            }
         }
     }
     url.search = searchParams.toString();
--- a/samples/client/petstore/typescript-axios/builds/default/common.ts
+++ b/samples/client/petstore/typescript-axios/builds/default/common.ts
@@ -94,7 +94,7 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
             if (Array.isArray(object[key])) {
                 searchParams.delete(key);
                 for (const item of object[key]) {
-                    searchParams.set(key, item);
+                    searchParams.append(key, item);
                 }
             } else {
                 searchParams.set(key, object[key]);
--- a/samples/client/petstore/typescript-axios/builds/default/common.ts
+++ b/samples/client/petstore/typescript-axios/builds/default/common.ts
@@ -91,7 +91,14 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
     const searchParams = new URLSearchParams(url.search);
     for (const object of objects) {
         for (const key in object) {
-            searchParams.set(key, object[key]);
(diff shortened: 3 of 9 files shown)


modules/openapi-generator/src/main/resources/typescript-axios/common.mustache
```diff
@@ -80,7 +80,14 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
     const searchParams = new URLSearchParams(url.search);
     for (const object of objects) {
         for (const key in object) {
-            searchParams.set(key, object[key]);
+            if (Array.isArray(object[key])) {
+                searchParams.delete(key);
```
TiFu, line 84: What's the reason this was added? 

If this is to avoid duplicate keys for different objects, we should IMO instead raise an error instead of silently removing the previously set object.

dantman: It's to make the array behaviour behave the same as the non-array behaviour.

`set` replaces any current query arguments. So the current non-array behaviour of setSearchParams is to delete any existing query argument when it replaces it. `append` doesn't clear previous values so I included a `delete` to ensure that `{foo: 'bar'}` and `{foo: ['bar']}` do not exhibit different behaviours, as that would be unexpected.

TiFu: Agreed on the behavior of `set` vs `append`. Why do we need `searchParams.delete(key)` before the `for` loop that uses `append`?

The case I have in mind is the following:
If `objects = [{foo: 'bar'}, {foo: ['baz', 'bazzzz']}]`, then the code currently would 
* `set("foo", "bar")`
* `delete("foo")`
* append("foo", "baz")
* append("foo", "bazzz")

Shouldn't we retain the `foo` from `objects[0]` or preferably throw an error because this is entirely undefined behavior?

Obviously, overlapping keys shouldn't happen but I'd prefer throwing an error in that case instead of silently dropping the first key.

dantman: In the current implementation that object will do the following:

- `set("foo", "bar")`
- `set("foo", ['baz', 'bazzzz'])`

Which will result in a query string of `foo=baz,bazzzz`. The current behaviour is that any key overwrites any previous value.

The comma-ness is a critical bug, it results in `style: form, explode: true` in the OpenAPI schema result in `foo=baz,bazzzz` when according to the OpenAPI spec it's supposed to result in ``foo=baz&foo=bazzzz`.

I am trying to fix this bug, ensuring that an array correctly results in setting multiple `key=` query parameters instead of incorrectly stringifying the array.

I'm not completely opposed to making `setSearchParams` always append (however in that case the `set` for non-arrays should also be `append`). However, that would technically be a breaking change (AFAICT making query options append instead of overwrite OpenAPI parameters).

I am trying to fix a critically broken bug in the typescript-axios generator that makes it unusable if you have a `style: form, explode: true` OpenAPI spec. So I do not want to introduce any breaking changes that would delay the release of the bugfix.

TiFu: :man_facepalming: Thanks for clearing this up! :) Let's retain the old behavior, for now, to not break any current setups.

TiFu (approved):

TiFu:
Merged - Thank you for your pull request (and patience ;))!

wing328:
@dantman thanks for the PR, which has been included in the v5.1.0 release: https://twitter.com/oas_generator/status/1373636654024380423

Issue #8771, opened 2021-02-19
[BUG] [python] additional_properties_type not set correctly in models


spacether:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [X] Have you searched for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

Using a v3 spec for an object schema, when additionalProperties is True or unset or {} the type should be set for it
Looking at our generated models, the type is not set at the model level though it IS being set in CodegenProperty
The expected output is to have `additional_properties_type = {str, int, float...}` when it is incorrectly being set to `additional_properties_type = None`

This issue: https://github.com/OpenAPITools/openapi-generator/issues/8813
Is also a blocker on getting this working because I need to define the vars that only the composed schema includes at the composed schema level.

##### openapi-generator version

5.0.0

##### OpenAPI declaration file content or url

```
# openapi-generator generate -i issue_8743v3_wrong_response_type.yaml -o issue_8743v3 -g python
openapi: 3.0.1
info:
  title: Quick Base API
  version: 1.0.0
servers:
- url: //api.quickbase.com/
paths:
  /fields/{fieldId}:
    get:
      description: Gets the properties of an individual field, based on field id.
      parameters:
      - name: fieldId
        in: path
        required: true
        schema:
          type: integer
      responses:
        200:
          description: Success
          content:
            '*/*':
              schema:
                type: object
                properties:
                  id:
                    type: integer
                  fieldType:
                    type: string
                  label:
                    type: string
                  properties:
                    $ref: '#/components/schemas/FieldProperties'
components:
  schemas:
    FieldProperties:
      type: object
      properties:
        comments:
          type: string
        format:
          type: string
        maxLength:
          type: integer
        allowNewChoices:
          type: boolean
        displayAsLink:
          type: boolean
        allowHTML:
          type: boolean
      description: 'Additional properties for the field. '
```

##### Generation Details

[...]

PR #8802, opened 2021-02-23
[python] Fixes additional_properties_type for models

spacether:
Before this PR, master branch python client code incorrectly set the additional_properties_type of models to None when it was unset, when it should be set to allow any type. Allowing any type is the default behavior when additionalProperties is unset per openapi versions >= 2.
This PR fixes additionalProperties values for models #8771, correctly setting  additional_properties_type.
- updates sample model docs
- adds test of additionalProperties in Tag
- fixes Fruit and GmFruit tests
- a [new v2 python sample has been added for disallowAdditionalPropertiesIfNotPresent=true](https://github.com/spacether/openapi-generator/tree/fixes_python_model_addprops_type/samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent). So now for v2 specs we have one sample for each boolean value of disallowAdditionalPropertiesIfNotPresent. The standard python v2 sample now uses disallowAdditionalPropertiesIfNotPresent=false (the option is omitted which default to false in the python generator)
- I have updated the discard_unknown_keys test @sebastien-rosset
- For v3 specs, where additionalProperties: False works, we have a test of it [here](https://github.com/spacether/openapi-generator/blob/fixes_python_model_addprops_type/samples/openapi3/client/petstore/python/tests_manual/test_composed_schema_with_props_and_no_add_props.py#L41)
- For v3 specs where a composed schema contains schemas where additionalProperties is False we have tests [here](https://github.com/spacether/openapi-generator/blob/fixes_python_model_addprops_type/samples/openapi3/client/petstore/python/tests_manual/test_fruit_req.py#L56)

## requiredVars have been moved out of the composed schema init signature
requiredVars have been moved out of the composed schema init signature because the required vars were probably defined in the composed oneOf/anyOf/allOf schemas and this schema does not know about them. Every schema is validated separately. So that means that for a composed schema with no type required or properties or additionalProperties definition inside it:
- all types are valid
- there are no required properties
- for object types, all properties are allowed because additionalProperties omitted defaults to additionalProperties: True which allows string property names and any type of value
- THEN the payload is checked against any oneOf/anyOf/allOf schemas

This change is needed because in our java code, if a oneOf class include required variables, those variable are set in composed schema that includes them. This is incorrect because those variable are not required in the composed schema, they are optional there.
One can see this by looking at the Fruit is a oneOf apple, banana example.
This PR updates it so apple and banana include required parameters. Before making the composed schema init change, those parameters were incorrectly present as positional arguments in Fruit. That's because the java code added those apple and banana requiredVars into Fruit's requiredVars which is incorrect.

## v2 Spec Implications
[There is a bug in swagger-parser for v2 specs](https://github.com/swagger-api/swagger-parser/issues/1369). With that bug, values of True, False, and unset are all seen as null additionalProperties which now becomes AnyType schema.
Previously, the default behavior was to treat all 3 conditions as setting additional_properties_type to None, which did not allow any additional properties to be sent for a given object payload.
In order to have additionalProperties True/False, and unset work correctly, users need to upgrade their specifications to v3.
If users still need the old behavior where additionalProperties was None for additionalProperties=True/False/unset, then they should set disallowAdditionalPropertiesIfNotPresent to True. Doing that will break composition usages oneOf/anyOf/allOf because for that to work, in the composed schema additionalProperties must be any type to allow in properties not defined in the composed schema.properties.
There's [another bug](https://github.com/swagger-api/swagger-parser/issues/1378) where schemas with unset type (AnyType) have their type set to object. This PR works around that bug. If you need to differentiate between AnyType schemas and free form objects (type: object with no properties) then update your spec to v3 or higher.

## Breaking Change with Fallback
For v2 specs that use compositiona and were setting disallowAdditionalPropertiesIfNotPresent=True this is a breaking change because composition (allOf schemas) will stop working. If you need composition to work, you must omit setting disallowAdditionalPropertiesIfNotPresent which will then default it to False. Composed schemas will then work. This conforms with the openapi tech spec.

Note: this different PR https://github.com/OpenAPITools/openapi-generator/pull/8816 will fix it so the variables defined in any schema/model in python only include variables for schema.properties and do not include composed schema properties.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator/src/main/resources/python/model_templates/method_init_composed.mustache
```diff
@@ -10,7 +10,52 @@
         '_additional_properties_model_instances',
     ])
 
-{{> model_templates/method_init_shared }}
+    @convert_js_args_to_python_args
+    def __init__(self, *args, **kwargs):  # noqa: E501
```
spacether, line 14: requiredVars have been moved out of the init signature because the required vars were defined in composed oneOf/anyOf/allOf schemas and this schema does not know about them. Every schema is validated separately. So that means that for a composed schema with no `type` `required` or `properties` or `additionalProperties` definition inside it.
- all types are valid
- there are no required properties
- for object types, all properties are allowed because additionalProperties omitted defaults to additionalProperties: True which allows string property names and any type of value
- THEN the payload is checked against any oneOf/anyOf/allOf schemas

spacether: This change will need to be made but it should be made in the future when I fix the value of vars and requiredVars to only include the info in each schema. Reverting this back to what it was before. That w ay these updates are incremental.

spacether:
This PR is blocked by the fact that v2 specs do not correctly set additionalProperties.
I will write a separate PR to add a fix for v2 specs which allows us to default additionalProperties to {} when the input additionalProperties is False or unset.

spacether:
Re-opening to run CI against the 5.2.X branch

Issue #8813, opened 2021-02-23
[BUG] ComposedSchema vars incorrectly includes extra properties


spacether:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [ ] Have you searched for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

composed schemas incorrectly include extra properties when the should only include self schema properties
It looks like they include composed schema properties
Keeping these separate is necessary for correct json schema validation. One must check that:
- types conform to any constraints at the composed schema level
- validations conform to any constraints at the composed schema level
- payload properties match properties defined at the composed schema level
- additional properties match additionalProperties constraints at the composed schema level (omitted additionalProperties allows any type to be ingested)
- defined composed schemas must validate to the payload

##### openapi-generator version

master 3.1.0

PR #8816, opened 2021-02-24
Adds fix for composed schema model vars

spacether:
Adds fix for composed schema model vars
Per [guidance from json schema and openapi people](https://github.com/OAI/OpenAPI-Specification/issues/2478), when a composed schema has properties or additionalProperties
object payloads are validated to the properties or additionalProerties definition at the composed schema level, independent of properties in composed oneOf/anyOf/allOf definition

So that means that CodegenModel.vars must store only the properties present in the composed schema, if they are present.
This has implications to how we validate composed schemas across generators, and that all generators should eventually support supportsAdditionalPropertiesWithComposedSchema = True.
For now only python and java support it.

This is a fix for: https://github.com/OpenAPITools/openapi-generator/issues/8813

### Blocker
Landing this PR is blocked by: https://github.com/OpenAPITools/openapi-generator/pull/8802
because we need additionalProperties to be set at the model level before we can use it to allow any value in in all key value pairs.

@sebastien-rosset should the go generator set supportsAdditionalPropertiesWithComposedSchema to True?

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -2619,7 +2619,7 @@ public class DefaultCodegen implements CodegenConfig {
         return m;
     }
 
-    private void setAddProps(Schema schema, IJsonSchemaValidationProperties property){
+    protected void setAddProps(Schema schema, IJsonSchemaValidationProperties property){
         if (schema.equals(new Schema())) {
             // if we are trying to set additionalProperties on an empty schema stop recursing
             return;
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -6174,7 +6174,7 @@ public class DefaultCodegen implements CodegenConfig {
         return codegenParameter;
     }
 
-    protected void addVarsRequiredVarsAdditionaProps(Schema schema, IJsonSchemaValidationProperties property){
+    protected void addVarsRequiredVarsAdditionalProps(Schema schema, IJsonSchemaValidationProperties property){
         setAddProps(schema, property);
         if (!"object".equals(schema.getType())) {
             return;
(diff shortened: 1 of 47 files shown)


wing328:
> Per guidance from json schema and openapi people, when a composed schema has properties or additionalProperties

Can you please share the links to the discussions?

spacether:
> > Per guidance from json schema and openapi people, when a composed schema has properties or additionalProperties
> 
> Can you please share the links to the discussions?

Sure, here it is: https://github.com/OAI/OpenAPI-Specification/issues/2478

samples/client/petstore/java/jersey2-java8-localdatetime/src/main/java/org/openapitools/client/model/Cat.java
```diff
@@ -46,33 +38,6 @@
...
-
-  public Boolean getDeclawed() {
-    return declawed;
-  }
-
-
-  public void setDeclawed(Boolean declawed) {
-    this.declawed = declawed;
-  }
-
```
wing328, line 75: @spacether the `declawed` property has been removed as part of the change, which doesn't seem to be correct according to the definition of `Cat`:
```
    Cat:
      allOf:
        - $ref: '#/components/schemas/Address'
        - type: object
          properties:
            declawed:
              type: boolean
```
ref: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/java/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml#L1411-L1418

spacether: That is correct as declawed only exists in that allOf third schema. It does not have an explicit definition inside Cat because cat has by default:
```
Cat:
  properties: {} # no declawed property here
  additionalProperties: true
```

wing328: > The CodegenModel Cat not containing declawed when vars only contains properties in the current schema is correct

Can you share the link(s) to discussion supporting this claim so that everyone is on the same page?

spacether: Yup it is in the link I pasted above https://github.com/OAI/OpenAPI-Specification/issues/2478

spacether:
So this change works to remove the vars that were hoisted into composed schemas from those composed schemas. They still exist in the source schemas and can be assigned as additional properties in the composed schemas.

I am concerned about merging this because it will remove the variables from the properties and signatures of the composed schemas. While not a breaking change, I suspect that many users will not want that.

Issue #853, opened 2018-08-20
[C#] Codegen for Unity


iBicha:
##### Description
This is an issue to discuss how a C# generator can be effectively useful in a Unity project.
I've been told the C# codegen is under a [refactor](https://github.com/OpenAPITools/openapi-generator/pull/737) phase, so my suggestion is to consider the next few points, to maybe include in this refactor.

I don't know if Unity will require a separate template, but it's a possibility.

##### Related issues/PRs
This issue will extend the points discussed here, kindly consider taking a look at it before continuing:
https://github.com/swagger-api/swagger-codegen/issues/8569

##### Suggest a fix/enhancement

Here's a list of things one might want to consider when supporting codegen for Unity:

- Using UnityWebRequest (see issue above): the usual way to call a UnityWebRequest and wait for a response is through Coroutines ([example](https://docs.unity3d.com/ScriptReference/Networking.UnityWebRequest.Get.html)) or through the [completed](https://docs.unity3d.com/ScriptReference/AsyncOperation-completed.html) event.

Note that the response is returned on the main thread (and of course, the deserialization of the response should happen on a background thread).

Also, this can easily be made into an `async/await` style by creating an Awaiter, maybe something like this:
```
    //UnityWebRequestAsyncOperation is an AsyncOperation
    public static TaskAwaiter GetAwaiter(this AsyncOperation asyncOperation)
    {
        var taskCompletionSource = new TaskCompletionSource<bool>();

        asyncOperation.completed += operation => { taskCompletionSource.TrySetResult(true); };

        return ((Task) taskCompletionSource.Task).GetAwaiter();
    }
```
Which will bring me to the next point:

- Depending on the configuration of a Unity project, the .Net version running can be 3.5 or 4.x equivalent. For that reason, some C# features are not always accessible. Fortunately, Unity has [define symbols](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html) allowing to identify the configuration, and have code to support it accordingly. For example:

```
public class API {
//
//APIs using Coroutines here, which is supported on both runtime versions
//

#if NET_4_6

//
//APIs using async/await here, only available in 4.6
//

#endif
}
```

- As for Json serialization, the obvious choice would be Newtonsoft's, although it is known that there are few modifications needed for it to work correctly on different platforms. This [asset](https://assetstore.unity.com/packages/tools/input-management/json-net-for-unity-11347) could be of help. 

[...]

PR #8821, opened 2021-02-24
C# Netcore WebRequest and HttpClient functionality

### LINKED ISSUE #6800: [REQ] New C# Generator that uses HttpClient instead of RestSharp
The existing C# generators produce clients that rely on the RestSharp library instead of the built-in System.Net.Http.HttpClient. This is particularly annoying if you're building apps using the ASP.NET *Typed HttpClient* pattern. In addition because the generated clients don't use System.Net.Http.HttpClient, it's impossible (or not very obvious) where you would add delegating handlers to handle cross-cutting concerns that span across all requests. Examples of cross-cutting concerns might be request logging, trace header propagation and retry logic.

What I would like to see is an alternative C# generator that would generate code to support developers in using the *Typed HttpClient* pattern.

For information about the System.Net.Http.HttpClient API.
https://docs.microsoft.com/dotnet/api/system.net.http.httpclient

For information about making requests with System.Net.Http.HttpClient.
https://docs.microsoft.com/aspnet/core/fundamentals/http-requests

For a quick guide to System.Net.Http.HttpClient and Delegating Handlers, I highly recommend the first half of this talk.
https://youtu.be/ojDxK_-I-To

Blackclaws:
This is WIP work tracking the addition of WebRequest and HttpClient as replacements for RestSharp, as some environments might find it undesirable to include an additional library (Unity, Blazor, etc.).

This is a bit orthogonal to the already open pull request and could be considered a different take on the subject:

https://github.com/OpenAPITools/openapi-generator/pull/8331

It tries to keep as much of the original interface intact as possible. WebRequest more so than HttpClient which simply does not offer many of the customizations that WebRequest and RestSharp do.

You can try it out by setting the library to httpclient by cliswitch or adding:

library: httpclient

to the global config in the config file you're using. Should hopefully alleviate/fix #6800 and #853

If you also want to get rid of the dependency on Polly add

additionalProperties:
  supportsRetry: false

aswell

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

 @mandrean @frankyjuang @shibayan


Blackclaws:
I think this is at a point now where it is ready to be reviewed. It still doesn't work with Blazor but it should work with Unity, I'll test that later today.

In general I think a separate generator such as @shibayan or @devhl-labs has been working on might be needed to fully support Blazor because quite a few options simply aren't possible there.

wing328:
Thanks for the PR. I played with it locally and it looks very good to start with 👍 

I'll merge it so that users can try it out in the upcoming 5.1.0 release (this month or next), and will submit PRs to further improve it (e.g. README, petstore integration tests in CI, etc)

modules/openapi-generator/src/main/resources/csharp-netcore/ApiClient.mustache
```diff
@@ -19,18 +19,28 @@ using System.Web;
 {{/netStandard}}
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
+using ErrorEventArgs = Newtonsoft.Json.Serialization.ErrorEventArgs;
+{{#useRestSharp}}
```
wing328, line 23: I'll create a separate ApiClient.mustache just for `httpclient` library similar to how we support multiple HTTP libraries in other client generators as too many `{{#useRestSharp}} ... {{/useRestSharp}}` etc make the template hard to maintain.

gbt518:
Thank you for this PR!

I found one small issue as I was testing this out in 5.1.0. I noticed that a new HttpClient is created for each call. 

From https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-5.0:

> HttpClient is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors.

VincentNe:
I would like to try out the new HttpClient generator but I don't really have a clue on how to enable the new generator with the cli. 

So far I've tried the following command:
`openapi-generator-cli generate -i api.json -g csharp-netcore -o Api.Client.NetCore --additional-properties library=httpclient
`
And I see that it's still using the RestSharp library, can anyone give me an example cli command on how to make use of the new library.

devhl-labs:
There are two versions of HttpClient, mine and Blackclaws. Mine only uses one instance of HttpClient that you pass in, however, it only supports basic requests and JWT tokens at the moment. Looks like Blackclaw's version may have made the 5.1 release, but mine didn't. :(

@VincentNe try just `--library httpclient` without the additional properties.

Blackclaws:
> Thank you for this PR!
> 
> I found one small issue as I was testing this out in 5.1.0. I noticed that a new HttpClient is created for each call.
> 
> From https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-5.0:
> 
> > HttpClient is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors.

Yes, this was indeed the case in order to be mostly a drop in replacement for restsharp. As creating new HttpClients for each request is indeed problematic in some cases there is now a new pull request to change this behaviour to be by default a static HttpClient for each ApiClient which however is also customizable from outside.

sleipper:
Some great work here @Blackclaws. I have some feedback if you're interested:

1) ApiClient attempts to deserialise to a model regardless of status code. For example a 404, 401, 500 would likely have a different model say with validation or problem details (https://docs.microsoft.com/en-us/aspnet/core/web-api/handle-errors?view=aspnetcore-5.0). Could also not return JSON or no content in the case of a 204 response code. Typically in the case at present the error returned is:
Most of ApiException : Unexpected character encountered while parsing value: P. Path '', line 0, position 0.
  Stack Trace: 
  
2) Generated API doesn't work with just HttpClient only, this is called out in the Readme but would be better to not permit invalid use. E.g new PetApi(client); the use of this will error as you need to set disableHandlerFeatures to true.

3) ApiClient NewRequest method uses Properties e.g. request.Properties["CookieContainer"], but this is obsolete, should used options instead. There is a TODO but due to your PR to avoid HttpClient being used for each call looks like this is per request but doesn't appear to be a way to set Cookies as localVarRequestOptions is always created. 

4) Minor issue
	- RetryConfiguration class not static

PR #8831, opened 2021-02-25
[C++][Qt5] added Oauth Access Token Support from Implicit Flow

basyskom-dege:
This PR will add support for Oauth tokens from the implicit flow. It provides setter functions and will add the token in the header for authorization.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

PING @wing328 @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @MartinDelille (2018/03) @muttleyxd (2019/08)


etherealjoy:
@basyskom-dege 
When you have time could you please resolve the conflicts, it is trivial .

basyskom-dege:
resolved conflicts

etherealjoy:
Thanks for taking up this topic, another important feature!.


Seems like there are some steps missing.
1. Call addPet() APIand receives error
2. Server redirects to OAuth2 token server
3. Present authorization to token Server (multiple possibilities here, name and password or)
4. Receive Token response
5. Use the token in the next request

Seems we are only doing step 5 in this PR. 
In normal cases the token received from step 4 is changing and would be unknown to client side. Only the data in step 3 is known to client side.

basyskom-dege:
yes im aware that im just doing step 5 there. I did some research on this topic and it seems like the impiclit flow is not supported by the Qt Oauth2 module. I created a class that does the job but then i stumbled across [this](https://github.com/pipacs/o2) repository that does this already. I wrote a small wrapper using this library and it worked as intendet. Unfortunalty I need to work on other things for the moment so I just put step 5 on this PR for now. I will come back to this in the future. But I might need to know if the whole implicit flow should be in the generated code or if the user should do that by himself and just set the token with the function i provieded here.

etherealjoy:
@basyskom-dege 
Ok thanks for the clarification, it is just that we can't advertise supporting OAuth2 implicit flow when we really are not doing it.

I did try to do something with this ealier and added a gist here
https://gist.github.com/etherealjoy/439679c7eca4ed7f2d58af3ade52430a
Idea is 
1. Store the `HttpRequestInput` when making API call until the callback comes
2. If Authentication error is received dont emit error signal, emit authencation needed signal and handle in callback
3. In authentication needed Callback send the credentials/digest or something 
4. Once authenticated callback comes extract token and inject the additional header. If authentication fail here fire the actual error signal. 

I never proposed it here because I was concerned about too much extra code.

basyskom-dege:
> Ok thanks for the clarification, it is just that we can't advertise supporting OAuth2 implicit flow when we really are not doing it.

Yes that should not be. But I saw that other generators that support it, dont do all of that either. I.E `go` client. But the way you propose the authentication looks good to me. I will pick it up later on and will inform you once I got something.

basyskom-dege:
@etherealjoy Hey. I was gone for a while and now im back to work on the Oauth2 support. Since the Implicit flow is deprecated i thought it would be the best idea to use Oauth2 Authorization Code Flow. It would be the easiest way to use Qt´s  [Oauth Module](https://doc.qt.io/archives/qt-5.9/qoauth2authorizationcodeflow.html) but this module has a commercial and gpl license. So is that suitable for the generator?

etherealjoy:
> EDIT: i will just try implement this without the qnetworkauth module so this wont be an issue

Thats would be great. The GPL could be a problem for some users indeed.

basyskom-dege:
Closed because of #10183

PR #8840, opened 2021-02-25
[dart] First stab add correct Dart lib layout

kuhnroyal:
Initial draft for new Dart library layout that closely follows Darts best practices.

* move everything under a `src` folder
* add new top level file called `$pubName.dart` that exports all other files
* users now only require one `import 'package:$pubName/$pubName.dart';` 

Most of the changes are just moved files and changed imports.

Looking for feedback @agilob @josh-burton 

I assume changing the file layout and requiring users to change their imports is considered a breaking change? @wing328


agilob:
Technically that's a big breaking change, but also super easy to fix

kuhnroyal:
Closing this and applying it to the next-gen dart-dio generator.

PR #8842, opened 2021-02-25
Correct use of loggers with placeholders and remove some redundant operations

agilob:
1. Removed some redundant operations from `static{}` blocks, unnecessary casts, object creation and a few more.
2. Fixed use of loggers in DefaultCodegen to use object placeholders which are faster, decrease object allocation in String pool and adds nullsafety in loggers


agilob:
@wing328  @jimschubert

agilob:
Another commit replaces very frequently used `string.replace()` with pattern / matcher / replace where pattern is compiled ahead of time to reduce number of times of compiling pattern in `java.lang.String.replaceAll`. Method doing String.replace is called in loops.

wing328:

agilob:
No one from this list ever responds

wing328 (approved):
LGTM

agilob:
Thanks

PR #8854, opened 2021-02-27
[dart] Default value is only available when nullable is not set

agilob:
Dart json_serializable does not allow setting `nullable: false` AND `defaultValue` at the same time.

As `nullable: true` with default value doesn't make sense, neither `nullable: false` it requires removal of `nullable`.

It throws exception:

```
Error with `@JsonKey` on `mapProperty`. Cannot use `defaultValue` on a field with `nullable` false.
package:openapi/model/additional_properties_class.dart:31:23
   ╷
31 │   Map<String, String> mapProperty;
   │                       ^^^^^^^^^^^

```

@kuhnroyal


modules/openapi-generator/src/main/resources/dart2/class.mustache
```diff
@@ -50,7 +50,7 @@ class {{{classname}}} {
   {{/isEnum}}
   {{^isBinary}}
   @JsonKey(
-    nullable: {{#nullable}}true{{/nullable}}{{^nullable}}false{{/nullable}},
+    {{#defaultValue}}defaultValue: {{{defaultValue}}},{{/defaultValue}}{{^defaultValue}}nullable: {{#nullable}}true,{{/nullable}}{{^nullable}}false,{{/nullable}}{{/defaultValue}}
```
kuhnroyal, line 53: I noticed a problem earlier today. It's not `{{nullable}}` - it is `{{isNullable}}`.
And then you can shorten this to `nullable: {{isNullable}}`

agilob: aa that's why it was behaving weird sometimes!

kuhnroyal: Ya found some places in other templates that had wrong output when I tried some NNBD stuff.

kuhnroyal (approved):

agilob:
@kuhnroyal bump

PR #8868, opened 2021-03-02
Proposed solution for issue 8833

nappa85:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
+++ b/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
@@ -80,7 +80,7 @@ pub enum {{{operationIdCamelCase}}}Error {
 /// {{{.}}}
 {{/notes}}
 {{#vendorExtensions.x-group-parameters}}
-pub {{#supportAsync}}async {{/supportAsync}}fn {{{operationId}}}(configuration: &configuration::Configuration{{#allParams}}{{#-first}}, params: {{{operationIdCamelCase}}}Params{{/-first}}{{/allParams}}) -> Result<{{#supportMultipleResponses}}ResponseContent<{{{operationIdCamelCase}}}Success>{{/supportMultipleResponses}}{{^supportMultipleResponses}}{{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}{{/supportMultipleResponses}}, Error<{{{operationIdCamelCase}}}Error>> {
+pub {{#supportAsync}}async {{/supportAsync}}fn {{{operationId}}}<A: configuration::ApiKey>(configuration: &configuration::Configuration<A>{{#allParams}}{{#-first}}, params: {{{operationIdCamelCase}}}Params{{/-first}}{{/allParams}}) -> Result<{{#supportMultipleResponses}}ResponseContent<{{{operationIdCamelCase}}}Success>{{/supportMultipleResponses}}{{^supportMultipleResponses}}{{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}{{/supportMultipleResponses}}, Error<{{{operationIdCamelCase}}}Error>> {
     // unbox the parameters
     {{#allParams}}
     let {{paramName}} = params.{{paramName}};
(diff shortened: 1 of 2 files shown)


wing328:
Please update the samples by running:
```
./mvnw clean package 
./bin/generate-samples.sh
```

wing328:
cc @frol (2017/07) @farcaller (2017/08) @richardwhiuk (2019/07) @paladinzh (2020/05)

wing328:
Is it correct to say that this is a breaking change?

What if the API doesn't use API key for authentication? For example, it only uses HTTP basic for authentication.

nappa85:
> Is it correct to say that this is a breaking change?
> 
> What if the API doesn't use API key for authentication? For example, it only uses HTTP basic for authentication.

You need to specify a type that impls ApiKey, even if not using it.
E.g:
```rust
let conf = Configuration::<String>::default();
```

But this approach allows dynamic ApiKeys, for example in my use case there is a nonce header that must be different and incremental for every call, and a signature header that's an HMAC based on a secret, the nonce and the URL.
With a basic HashMap approach (the minimum for multiple keys support) I would need to keep a mutable instance of Configuration and keep editing the keys, having to foresee the URL that will be composed with GET parameters.
With my approach I simply need a struct that impls ApiKey, it's everything simpler and more dynamic.

I think it's a standard Rust approach

wing328:
@nappa85 what about following other clients and use a hash to store the API key instead?

e.g. Configuration.cs in the C# client: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/csharp/OpenAPIClient/src/Org.OpenAPITools/Client/Configuration.cs#L88-L98

There's also something call API key prefix but it's not something that needs to be included in this PR.

nappa85:
> @nappa85 what about following other clients and use a hash to store the API key instead?
> 
> e.g. Configuration.cs in the C# client: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/csharp/OpenAPIClient/src/Org.OpenAPITools/Client/Configuration.cs#L88-L98
> 
> There's also something call API key prefix but it's not something that needs to be included in this PR.

It's a less powerful approach, and, as you already told, it doesn't support the prefixes.

In my use case I had to expand my approach passing also a reference to the RequestBuilder to be able to retrieve already setted headers and the URL:
```rust
use openapi::apis::configuration::ApiKey;

use reqwest::RequestBuilder;

use chrono::Utc;

use hmac::{Hmac, Mac, NewMac};

use log::error;

type HmacSha512 = Hmac<sha2::Sha512>;

/// TheRockTrading ApiKeys manager
pub struct TrtKeys {
    api_key: String,
    secret: String,
}

impl TrtKeys {
    /// costructor
    pub fn new(api_key: String, secret: String) -> Self {
        TrtKeys {
            api_key,
            secret
        }
    }
}

impl ApiKey for TrtKeys {
    fn get_prefix(&self, _: &str) -> Option<String> {
        None
    }
    fn get_key(&self, name: &str, req_builder: &RequestBuilder) -> Option<String> {
        match name {
            "X-TRT-KEY" => Some(self.api_key.clone()),
            "X-TRT-NONCE" => {
                let timestamp = Utc::now().timestamp_nanos();
                Some(timestamp.to_string())
            },
            "X-TRT-SIGN" => {
                let rb = req_builder.try_clone()?;
                let req = rb.build()
                    .map_err(|e| error!("error building request: {}", e))
                    .ok()?;
                let nonce = req.headers().get("X-TRT-NONCE")?;
                let mut mac = HmacSha512::new_varkey(self.secret.as_bytes())
                    .map_err(|e| error!("error initializing HMAC-SHA521: {}", e))
                    .ok()?;

                // this way we skip the utf8 check
                let mut bytes = nonce.as_bytes().to_vec();
                let mut url = req.url().to_string().into_bytes();
                bytes.append(&mut url);
                mac.update(&bytes);

                Some(hex::encode(mac.finalize().into_bytes()))
            },
            _ => None,
        }
    }
}
```

I think it's a poweful example

7Hazard:
It's evident this has been forgotten about, but is still an issue in 7.7.0.

PR #8876, opened 2021-03-03
[csharp-netcore] Make Blazor Webassembly compatible

Blackclaws:
This is a first step that in general makes the API Client compatible with Blazor Webassembly applications. I'll add some more detailed instructions later but it does work already out of the box. Performance improvements as well as size improvements are something that is definitely desirable in that scenario but that can come later.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/samples/client/petstore/csharp-netcore/OpenAPIClient-httpclient/src/Org.OpenAPITools/Client/ApiClient.cs
+++ b/samples/client/petstore/csharp-netcore/OpenAPIClient-httpclient/src/Org.OpenAPITools/Client/ApiClient.cs
@@ -337,10 +337,13 @@ namespace Org.OpenAPITools.Client
 
             if (response != null)
             {
-                foreach (Cookie cookie in handler.CookieContainer.GetCookies(uri))
-                {
-                   transformed.Cookies.Add(cookie);
+                try {
+                    foreach (Cookie cookie in handler.CookieContainer.GetCookies(uri))
+                    {
+                       transformed.Cookies.Add(cookie);
+                    }
                 }
+                catch (PlatformNotSupportedException) {}
             }
 
             return transformed;
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
@@ -349,10 +349,13 @@ namespace {{packageName}}.Client
 
             if (response != null)
             {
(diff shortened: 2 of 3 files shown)


samples/client/petstore/csharp-netcore/OpenAPIClient-httpclient/src/Org.OpenAPITools/Client/ApiClient.cs
```diff
@@ -337,10 +337,13 @@ private ApiResponse<T> ToApiResponse<T>(HttpResponseMessage response, object res
...
-                foreach (Cookie cookie in handler.CookieContainer.GetCookies(uri))
-                {
-                   transformed.Cookies.Add(cookie);
+                try {
+                    foreach (Cookie cookie in handler.CookieContainer.GetCookies(uri))
+                    {
+                       transformed.Cookies.Add(cookie);
+                    }
                 }
+                catch (PlatformNotSupportedException) {}
```
wing328, line 346: Shall we log a warning message or use other channels to notify the users/developers? It seems like the exception is now discarded silently.

Blackclaws: A good question and one I've also asked myself. The issue is that PlatformNotSupported is a very specific exception that I think is only thrown on Blazor WASM right now (as its running in Browser).

If we start logging this now this will clutter up the console log of anyone using the generated code for any api call on that platform. We could introduce a logging level cli switch for generating code that would introduce log messages here and at other places throughout the code. That might be an interesting addition anyway.

wing328:
cc @mandrean (2017/08) @frankyjuang (2019/09) @shibayan (2020/02) @Blackclaws (2021/03)

samples/client/petstore/csharp-netcore/OpenAPIClient-httpclient/src/Org.OpenAPITools/Client/ApiClient.cs
wing328: > If we start logging this now this will clutter up the console log of anyone using the generated code for any api call on that platform

What about we start with this and revise the logging strategy later if users complain about it?

I still prefer over-communication to the users/developers to start with because I'd bad experiences dealing with errors being silent before.

wing328: Let's go with what you've so far to start with. We can always improve later with another PR.

PR #8883, opened 2021-03-03
[dart] Add option to set linting library, initial support effective_dart

agilob:
@kuhnroyal 

this is wide for all generators, but I tested only dart and dart with json_serializable


kuhnroyal:
I am not convinced we should go this way. We need to decide for one library that we use while generating so that we can validate it in CI. Allowing to configure it will mean that it will always fail with some of them because they use different rule-sets. If the user really wants something other than what is preconfigured, then they need to use a custom template imo.

agilob:
In that case we should remove all linting rules and the other PR which adds  analysis file as it might not lint in a way a user might expect.

kuhnroyal:
No, we need to lint to make sure we have no bugs in the examples. The generator creates a library project, if you consume it in your project, you don't really care how it was linted as you don't consume the analysis_options.

agilob:
well, ok then. we can reconsider in the future maybe :P

Issue #8904, opened 2021-03-05
[BUG][JavaScript] client generator: Reserved words detector is not case-sensitive


delenius:
#### Bug Report Checklist

- [X] Have you provided a full/minimal spec to reproduce the issue?
- [X] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [X] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [X] Have you searched for related issues/PRs?
- [X] What's the actual output vs expected output?

##### Description

The JS client generator is case-insensitive when it comes to detecting reserved words, while JavaScript itself is not.
For example, it complains about 'InstanceOf' as a model name, and renames it to 'ModelInstanceOf'.

##### openapi-generator version

5.1.0-SNAPSHOT, commit `976b4aa757e989fbaa28f92a721ade39735332e6`.

##### OpenAPI declaration file content or url

See full details in [this test repo](https://github.com/delenius/openapi-issues/tree/main/reserved-words).

##### Suggest a fix

The culprit is this line in `DefaultCodegen.java`:

```java
    protected boolean isReservedWord(String word) {
        return word != null && reservedWords.contains(word.toLowerCase(Locale.ROOT));
    }
```

Removing `.toLowerCase` here should fix the problem. I will submit a PR.

PR #8944, opened 2021-03-10
[Java] Make reserved word detection case-insensitive

delenius:
The Java generators are not case-sensitive w.r.t reserved word detection. This is the same for Java as #8904 was for JavaScript. There are likely other languages affected, but I am still fixing it on a case-by-case basis because some tests broke when I fixed it in `DefaultCodegen.java`.

Pinging Java tech committee:
@bbdouglas @sreeshas @jfiala  @lukoyanov @cbornet @jeff9finger @karismann @Zomzog @lwlee2608 @nmuesch

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -142,7 +142,7 @@ public AbstractJavaCodegen() {
 
         hideGenerationTimestamp = false;
 
-        setReservedWordsLowerCase(
+        reservedWords = new HashSet<>(
```
agilob, line 145: `reservedWords = Sets.newHashSet(....)`

samples/client/petstore/java/rest-assured-jackson/src/main/java/org/openapitools/client/model/ModelApiResponse.java
```diff
@@ -35,7 +35,7 @@
   ModelApiResponse.JSON_PROPERTY_TYPE,
   ModelApiResponse.JSON_PROPERTY_MESSAGE
 })
-@JsonTypeName("ApiResponse")
+@JsonTypeName("_ApiResponse")
```
agilob, line 38: Is this change intentional? Change in the JsonTypeName will be a breaking change without fallback. This PR won't get far as is.

delenius: Hmm, no, certainly not. I'll look into it.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java
delenius: Fixed.

samples/client/petstore/java/rest-assured-jackson/src/main/java/org/openapitools/client/model/ModelApiResponse.java
delenius: I fixed this by modifying `AbstractJavaCodegen.fromName` so that it keeps the original name here, rather than calling `escapeReservedWord` for these. I tried changing it directly in `DefaultCodegen`, but that caused other problems as usual (in the Kotlin generator, this time). Please take another look.

PR #8965, opened 2021-03-12
[typescript-axios] handle uniqueItems in query and header parameters

t-h-e:
Fixes https://github.com/OpenAPITools/openapi-generator/issues/6904
`uniqueItems` for query and header parameter have not been handled correctly

add endpoint /fake/test-unique-paramters to petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml and typescript-axios config for demonstration

@TiFu  @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov


modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
```diff
@@ -117,7 +127,12 @@ export const {{classname}}AxiosParamCreator = function (configuration?: Configur
     {{#headerParams}}
             {{#isArray}}
             if ({{paramName}}) {
+                {{#uniqueItems}}
+                let mapped = Array.from({{paramName}}).map(value => (<any>"{{{dataType}}}" !== "Set<string>") ? JSON.stringify(value) : (value || ""));
```
t-h-e, line 131: I'm not 100% sure about the comparison here `<any>"{{{dataType}}}" !== "Set<string>"` as I'm not sure what the comparison in the `Array` case is supposed to do. If someone can verify that this is correct or let me know what the correct version is supposed to be, I would be grateful 😃

t-h-e:
The commit appears bigger than it actually is due to the generated code from the additional typescript-axios config

modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache
macjohnny: it looks correct, I think the idea is to serialize the items of a non-string array/set

macjohnny (approved):
LGTM

macjohnny:
@t-h-e thanks for your contribution. please re-generate the samples

macjohnny:
note: this PR is similar to https://github.com/OpenAPITools/openapi-generator/pull/8695

PR #8975, opened 2021-03-15
[DART] Add standard generated classes as reserved

sbu-WBT:
A friend create a specification with a model named `ApiClient` which resulted for me in two classes with the same name which doesn't compile. So I thought it might be handy to mark the default included classes also as reserved.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@swipesight (2018/09) @jaumard (2018/09) @josh-burton (2019/12) @amondnet (2019/12) @sbu-WBT (2020/12) @kuhnroyal (2020/12) @agilob (2020/12)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
@@ -164,8 +164,7 @@ public abstract class AbstractDartCodegen extends DefaultCodegen {
                 "File",
                 "Client",
                 "Future",
-                "Response",
-                "StreamedRequest"
+                "Response"
         );
 
         cliOptions.add(new CliOption(PUB_LIBRARY, "Library name in generated code"));
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
@@ -164,7 +164,8 @@ public abstract class AbstractDartCodegen extends DefaultCodegen {
                 "File",
                 "Client",
                 "Future",
-                "Response"
+                "Response",
+                "StreamedRequest"
         );
 
         cliOptions.add(new CliOption(PUB_LIBRARY, "Library name in generated code"));
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
@@ -17,6 +17,7 @@
 
 package org.openapitools.codegen.languages;
 
+import com.google.common.collect.Sets;
 import org.openapitools.codegen.CliOption;
 import org.openapitools.codegen.CodegenConstants;
 import org.openapitools.codegen.SupportingFile;
@@ -36,6 +37,18 @@ public class DartClientCodegen extends AbstractDartCodegen {
 
     public DartClientCodegen() {
         super();
+
+        additionalReservedWords.addAll(
+                Sets.newHashSet(
+                        "ApiClient",
+                        "QueryParam",
+                        "Authentication",
+                        "HttpBasicAuth",
+                        "HttpBearerAuth",
+                        "ApiKeyAuth",
+                        "OAuth"
+                )
+        );
 
         final CliOption serializationLibrary = new CliOption(CodegenConstants.SERIALIZATION_LIBRARY,
                 "Specify serialization library");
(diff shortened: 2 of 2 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
```diff
@@ -164,7 +164,8 @@ public AbstractDartCodegen() {
                 "File",
                 "Client",
                 "Future",
-                "Response"
+                "Response",
+                "StreamedRequest"
```
kuhnroyal, line 168: This should also go to the normal dart client as it is part of the `http` package and not used in the dio generator?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
```diff
@@ -37,6 +38,18 @@
     public DartClientCodegen() {
         super();
 
+        additionalReservedWords.addAll(
+                Sets.newHashSet(
```
agilob, line 42: Thinking about the opposite problem. I don't generate ApiClient class, I would generate only models. This makes ApiClient a non-reserved word, thus produces unexpected results.

The problem exists, but I'm not convinced this is the right solution here.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
agilob: This change is in AbstractDartCodegen.java so it affects all dart clients, including dartdio.

kuhnroyal: Yea so it should also go to `DartClientCodegen`.

sbu-WBT: I thought of this,  too. But then I saw [`Response`](https://pub.dev/documentation/http/latest/http/Response-class.html) which is also a `http` class (If I'm not wrong) so I thought I put it there. I have no problem to put it in the other set

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
kuhnroyal: I think this is the correct solution, doing this as well in the dio generator. You may have a couple false positives but that is ok I'd say.

sbu-WBT: There is the option to only generate the models? Then just add the classes to the set only if the parameter is not set. Do you mind telling me the option name?

agilob: >Or if you have another idea just tell us, I'm open to better solutions

Well, no. It would have to be conditionally added increasing complexity of such solution. I would prefer to see global PR for this problem rather than each generator doing it on their own.

@wing328 is there any DefaultCodeGen solution for this? A set of _potentially_ used keywords like `ApiClient` to prevent generated files collide with project files. It would have to be managed independently by each code generator and their cmd options.

Looks like big effort to handle some small cases, but correctness of the output should be priority(?).

agilob:
Im happy to approve it after @kuhnroyal approves it. I think the problem in other comments needs a wider planning for all generators, making it out of scope for your problem @sbu-WBT

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractDartCodegen.java
kuhnroyal: Yea, I added `Response` here because it also exists in Dio (a different class but same name). Same goes for `Client`.
I think it doesn't hurt here as the `http` package can basically be considered part of Dart.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartClientCodegen.java
kuhnroyal: > is there any DefaultCodeGen solution for this? A set of _potentially_ used keywords like `ApiClient` to prevent generated files collide with project files. It would have to be managed independently by each code generator and their cmd options.

The problem is that most other generators/languages have a concept of absolute/fq imports but Dart does not.
So in other generators you can use the import mapping which can be configured as cmd parameters or in the generator class. The dart generators basically miss uses the import mapping, meaning we HAVE TO use it differently due to the absence of the absolute/fq imports concept.

The `additionalReservedWords ` is just something I came up with for the dio generator to solve this in a simple manner.

kuhnroyal: Well that is also not fully correct. Dart has named imports `import "package:foo:foo.dart" as bar;` but I tried this and it fails with a lot of builders/build_runner (`json_serializable`, `built_value`).

kuhnroyal (approved):

agilob (approved):

agilob:
@sbu-WBT re-run the tests

wing328:
@sbu-WBT thanks for the PR, which has been included in the v5.1.0 release: https://twitter.com/oas_generator/status/1373636654024380423

PR #8988, opened 2021-03-17
[Java][Jaxrs-Resteasy] Fixes generator devaultValues for int64/float/…

### LINKED ISSUE #8986: Query Parameters of type integer in format int64 seem to create a broken rest easy API
##### Description
Very similar to https://github.com/OpenAPITools/openapi-generator/issues/1226, but for rest easy instead.
Created a API Spec with a query parameter of type integer, with format int64 which results to Long in Java. Leads to a broken resteasy Api with the following signature:
@DefaultValue("0l") @QueryParam("offset") Long offset,

5.1.0

  causing the problem (line of code or commit), or simply make a suggestion -->
remove the suffix for long, float, double etc, as in 
https://github.com/OpenAPITools/openapi-generator/pull/4969
https://github.com/OpenAPITools/openapi-generator/pull/4969/files#diff-9de34b4aad5de707e8aa0c031f20a6573f8f50826a437f0652c54b3788b37d41R869

lucafuji:
This PR fixes #8986 
That issue describes that when we have an in64, float, or double CodegenParameter, the string defaultValue includes unneeded suffix characters l,f, or d. Those suffixes were added by AbstractJavaCodegen because that generator assumed that those default values would be used as java values of the correct type. Instead, the spring generator uses String type default values for CodegenParameter only. Typically those values are handled with parseLong/parseFloat/parseDouble etc which fail if the value includes a character suffix.

 we now use values which lack the suffix so parseLong/parseDouble/parseFloat will work
CodegenParameter defaultValues:

"21474836487" (int64)
"3.14159" (float)
"3.14159" (double)

Verification Test
A test has been added at https://github.com/OpenAPITools/openapi-generator/pull/8988/files#diff-ecbab4cf643943739b9ce7e56d73b1b391b4ce6d09310326e8ea9d023da164b8R41


lucafuji:
@wing328 Not sure how to assign this PR to you but please take a look, thanks

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/PostProcessUtils.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2018 OpenAPI-Generator Contributors (https://openapi-generator.tech)
+ * Copyright 2018 SmartBear Software
```
wing328, line 3: @lucafuji new files added to openapi-generator project doesn't require this line: https://opensource.stackexchange.com/questions/7300/copyright-notice-in-the-file-header-apache-v2-license

lucafuji: removed this line

lucafuji:
@wing328 It seems the build keeps failing due to the docker pull limit
`
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "docker pull swaggerapi/petstore" failed and exited with 1 during .`

Any thing we can do to let it pass?

modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/PostProcessUtils.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openapitools.codegen.utils;
+
+import org.openapitools.codegen.CodegenParameter;
+
+public class PostProcessUtils {
+  public static void postProcessParameter(CodegenParameter p) {
```
wing328, line 22: What about putting this function in `AbstractJavaCodegen` class instead as this seems to be relevant to Java only?

wing328: Or simply put the function in the `postProcessParameter` of `AbstractJavaCodegen` to avoid calling it in the Java client/server generator java classes ?

wing328 (approved):
LGTM. Thanks for the fix 👍

PR #8995, opened 2021-03-17
[csharp][aspnetcore] Update aspnetcore nuspec and csproj with authors and description

Blackclaws:
Update aspnetcore nuspec and csproj with authors and description

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -24,6 +24,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |operationResultTask|Set methods result to Task&lt;&gt;.| |false|
 |packageAuthors|Specifies Authors property in the .NET Core project file.| |OpenAPI|
 |packageCopyright|Specifies an AssemblyCopyright for the .NET Framework global assembly attributes stored in the AssemblyInfo file.| |No Copyright|
+|packageDescription|Specifies a AssemblyDescription for the .NET Framework global assembly attributes stored in the AssemblyInfo file.| |A library generated from a OpenAPI doc|
 |packageGuid|The GUID that will be associated with the C# project| |null|
 |packageName|C# package name (convention: Title.Case).| |Org.OpenAPITools|
 |packageTitle|Specifies an AssemblyTitle for the .NET Framework global assembly attributes stored in the AssemblyInfo file.| |OpenAPI Library|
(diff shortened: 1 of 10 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
```diff
@@ -21,6 +21,7 @@
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.parser.util.SchemaTypeUtil;
+import org.mozilla.javascript.optimizer.Codegen;
```
wing328, line 24: What is this for?

docs/generators/aspnetcore.md
```diff
@@ -24,6 +24,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |operationResultTask|Set methods result to Task&lt;&gt;.| |false|
 |packageAuthors|Specifies Authors property in the .NET Core project file.| |OpenAPI|
 |packageCopyright|Specifies an AssemblyCopyright for the .NET Framework global assembly attributes stored in the AssemblyInfo file.| |No Copyright|
+|packageDescription|Specifies a AssemblyDescription for the .NET Framework global assembly attributes stored in the AssemblyInfo file.| |A library generated from a OpenAPI doc|
```
wing328, line 27: Instead of adding another CLI option, what about using `{{appDescription}}` instead ?

Ref: https://github.com/OpenAPITools/openapi-generator/blob/ba4aae5074c6d786e375887b24a38703cf7a6410/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java#L296

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AspNetCoreServerCodegen.java
Blackclaws: Nothing. Removed it. Was added by accident and I didn't catch it

docs/generators/aspnetcore.md
Blackclaws: We could do that. The way I did it makes it consistent with the csharp-netcore generator. We might in general think about refactoring that one as well then and getting rid of duplicate options, though it might be an incompatible change so that might have to wait until 6.0.0

I could just add it in additional properties instead. Would that be better? I think having it at least somewhat consistent between generators of the same language family makes sense.

wing328: 👌 let's go with what you've so far.

PR #9006, opened 2021-03-18
Bug - python client deserialization when attribute named `self`

gbmarc1:
@taxpon  @frol @mbohlool  @cbornet @kenjones-cisco @tomplus @Jyhess @arun-nalla @spacether

When an attribute is named `self`  in model  `__init__` fails because `self` is in kwargs. Generation maps `self` to `_self` properly. However, when calling the endpoint the deserialization fails.

```yaml
openapi: 3.0.2
info:
  title: bug report
  description:  Bug report
  version: 0.60.1
servers:
  - url: /inference/v1
paths:
  /datasets:
    get:
      responses:
        '200':
          description: description
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/datasets'

components:
  schemas:
    datasets:
        type: object
        properties:
          self:
            type: string
```

```python

    import open_api
    from open_api.api import datasets_api

    # Defining the host is optional and defaults to http://localhost/inference/v1
    # See configuration.py for a list of all supported configuration parameters.
    configuration = open_api.Configuration(
        host="/inference/v1",
    )

    # Enter a context with an instance of the API client
    with open_api.ApiClient(configuration) as api_client:
        # Create an instance of the API class
        api_instance = datasets_api.DatasetsApi(api_client)

        api_instance.v1_inference_datasets_get()  # FAILS HERE
```

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


gbmarc1:
#9006

wing328:

gbmarc1:
@wing328 , it should be ok now. Thank you

modules/openapi-generator/src/main/resources/python/model_utils.mustache
```diff
@@ -910,6 +910,13 @@ def deserialize_model(model_data, model_class, path_to_item, check_type,
                    _configuration=configuration,
                    _spec_property_naming=spec_property_naming)
 
+    # An attribute named `self` received from the api will conflicts with the reserved `self`
+    # parameter of a class method. During generation, `self` attributes are mapped
+    # to `_self` in models. Here, we map it directly.
+    if "self" in input_value:
+        input_value["_self"] = input_value["self"]
+        del input_value["self"]
```
spacether, line 918: So I am concerned that this code is manually handling one case only.
Why does the convert_js_args_to_python_args fn not correctly handle this parameter?
That fn is used to decorate the init method: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/petstore_api/model/pet.py#L120

Fn definition is here:
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/petstore_api/model_utils.py#L1559

Which calls change_keys_js_to_python
https://github.com/OpenAPITools/openapi-generator/blob/46326249d2c9df8ec530199ea6dbb8821acf1dd1/samples/openapi3/client/petstore/python/petstore_api/model_utils.py#L1041
Which uses model_class.attribute_map to convert to the spec property name.
If the property name is self and the property is optional, it is passed in kwargs.

spacether: Hmm this is happening because the kwargs self parameter is assigned to the self argument of the model:
```
>>> class A:
...   def __init__(self, **kwargs):
...     self.kwargs = kwargs
... 
>>> 
>>> A()
<__main__.A object at 0x10673fdc0>
>>> kwargs = dict(self=5)
>>> kwargs
{'self': 5}
>>> A(**kwargs)
  File "<stdin>", line 1, in <module>
TypeError: __init__() got multiple values for argument 'self'
```

spacether (changes requested):
@gbmarc1 Thank you for your PR. One goal with the python generator is to not treat any variable as special. If a name collision can happen, I have been adding an underscore in on the python side to prevent property name collisions.
So we don't use `nullable` as a class property, we use `_nullable` to allow a model to have a property named nullable.

Can you rename the self parameter in the model init signature to _self [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/model_templates/method_init_shared.mustache#L2)?
That way our existing code will work and this variable is not handled in a special way outside of our model mustache templates.

Also, can you add a new schema that demonstrates this problem to our sample spec here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/python/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml#L2320
And can you add a model test like [we have here](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/tests_manual/test_number_with_validations.py#L33)?
If we don't have a test of this and someone changes the code in the future we will not be able to detect when this feature breaks.

modules/openapi-generator/src/main/resources/python/model_utils.mustache
gbmarc1: In fact this is my error:
```
  File ".../model_utils.py", line 1287, in attempt_convert_item
    return deserialize_model(input_value, valid_class,
  File ".../model_utils.py", line 1201, in deserialize_model
    return model_class(**kw_args)
TypeError: wrapped_init() got multiple values for argument 'self'
```
So `self` is assigned as a decorator parameter and not as the kwargs to process.

gbmarc1: Other solution would be to deviate from python standard and set the self parameter to something highly improbable to set as an attribute in an api

```
def convert_js_args_to_python_args(fn):
    from functools import wraps

    @wraps(fn)
    def wrapped_init(__wrapped_self__no_collision, *args, **kwargs):
        spec_property_naming = kwargs.get("_spec_property_naming", False)
        if spec_property_naming:
            kwargs = change_keys_js_to_python(
                kwargs, __wrapped_self__no_collision.__class__
            )
        return fn(__wrapped_self__no_collision, *args, **kwargs)

    return wrapped_init
```

spacether: > set the self parameter to something highly improbable
Yes, that's why my asked solution is to rename it to `_self`. One would need to change wrapped_init self to `_self` and the model `__init__`  self to `_self` This prevents the collision

modules/openapi-generator/src/main/resources/python/model_utils.mustache
```diff
@@ -1277,11 +1277,16 @@ def get_valid_classes_phrase(input_classes):
 def convert_js_args_to_python_args(fn):
     from functools import wraps
     @wraps(fn)
-    def wrapped_init(self, *args, **kwargs):
+    def wrapped_init(_self, *args, **kwargs):
+    """
+    An attribute named `self` received from the api will conflicts with the reserved `self`
+    parameter of a class method. During generation, `self` attributes are mapped
+    to `_self` in models. Here, we name `_self` instead of `self` to avoid conflicts.
```
spacether, line 1284: How about we stop mapping `self` to `_self` in generation?
That's what I was asking for when I said:
> And I think that we should remove the self word as a [reserved python word.](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PythonLegacyClientCodegen.java#L122) Then we can pass payload property self through to the init with no mutations and the java code will not rename self to _self.

gbmarc1: I don't agree it should still be reserved.
The problem would be moved elsewhere like in the `__init__` methods of the models.

https://github.com/OpenAPITools/openapi-generator/blob/ee2f585f37c16834f312e0e8440da777350f6560/modules/openapi-generator/src/main/resources/python/model_templates/method_init_simple.mustache#L11

And what if we have an attribute named `_self`. We are back to the same problem. Now, if the user has both `_self` and `self`, one will overwrite the other. I think it is reasonable to assume that it is less likely to happen.

 I think only modifying the argument in the wrapper is appropriate.

spacether: Right, that's why I asked that we rename all selfs to _selfs, so we can treat the spec as the source of truth and not have a collision. But it sounds like we disagree about that. If you want to only change the wrapper and can get tests passing that is fine.

gbmarc1: Any idea why circleci doesn't pass?

spacether: Yup
```

	modified:   samples/openapi3/client/petstore/python/.openapi-generator/FILES
	modified:   samples/server/petstore/kotlin-server/ktor/src/main/kotlin/org/openapitools/server/apis/PetApi.kt
	modified:   samples/server/petstore/kotlin-server/ktor/src/main/kotlin/org/openapitools/server/apis/StoreApi.kt
	modified:   samples/server/petstore/kotlin-server/ktor/src/main/kotlin/org/openapitools/server/apis/UserApi.kt

This script runs in pull requests against the anticipated merge commit (as if the PR was merged now).
When you see unexpected files here, it likely means that there are newer commits in master that you need to 
rebase or merge into your branch.

Please run 'bin/utils/ensure-up-to-date' locally and commit changes (UNCOMMITTED CHANGES ERROR)
2
```
Can you remove the kotlin changes from this pr?

samples/server/petstore/kotlin-server/ktor/src/main/kotlin/org/openapitools/server/apis/PetApi.kt
```diff
@@ -1,80 +1,80 @@
...
+    }
+
+
+    get<Paths.findPetsByStatus> {  _: Paths.findPetsByStatus ->
+        val principal = call.authentication.principal<OAuthAccessTokenResponse>()
+        
+        if (principal == null) {
+            call.respond(HttpStatusCode.Unauthorized)
+        } else {
+            val exampleContentType = "application/json"
```
spacether, line 77: Why was this file changed?

gbmarc1: required in the PR checklist runnning:
./mvnw clean package 
./bin/generate-samples.sh

Changed the files

samples/server/petstore/kotlin-server/ktor/src/main/kotlin/org/openapitools/server/apis/PetApi.kt
```diff
@@ -1,80 +1,80 @@
...
+
+
+    delete<Paths.deletePet> {  _: Paths.deletePet ->
+        val principal = call.authentication.principal<OAuthAccessTokenResponse>()
+        
+        if (principal == null) {
+            call.respond(HttpStatusCode.Unauthorized)
+        } else {
+            call.respond(HttpStatusCode.NotImplemented)
+        }
```
spacether, line 67: Adding this `^M` looks incorrect

gbmarc1: ok I will revert that changes for tis file

gbmarc1:
@spacether 
Could you just regenerate the file for me.
Generation fails on my side. I don't know why

`[INFO] Reactor Summary for openapi-generator-project 5.1.0-SNAPSHOT:                                                                                                                                             [INFO]                                                                                                                                                                                                           [INFO] openapi-generator-project .......................... SUCCESS [  5.004 s]                                                                                                                                  [INFO] openapi-generator-core ............................. SUCCESS [01:25 min]                                                                                                                                  [INFO] openapi-generator (core library) ................... FAILURE [31:35 min]                                                                                                                                  [INFO] openapi-generator (executable) ..................... SKIPPED                                                                                                                                              [INFO] openapi-generator (maven-plugin) ................... SKIPPED                                                                                                                                              [INFO] openapi-generator-gradle-plugin (maven wrapper) .... SKIPPED                                                                                                                                              [INFO] openapi-generator-online ........................... SKIPPED `

gbmarc1:
Any idea why CI\Circleci build fails? Error seems unrelated to this PR

spacether:
> Any idea why CI\Circleci build fails? Error seems unrelated to this PR

That error is unrelated to this PR.

spacether (approved):
Thank you for your PR and adding the test case. This looks good.

PR #9012, opened 2021-03-19
[scala-akka-http-server] allows generation of managed sources only

cchafer:
When the official OpenAPITools / sbt-openapi-generator sbt plugin is used to generate a scala-akka-http-server stub, the resulting files are a project in their own right, including:
- a build file
- a README.md file
- a default controller (controller.scala)

The problem is that it can be interesting not to generate a project, but just the source files needed to compile a project.
By generating just scala files, which are placed in the managedSources of sbt, one can automate the generation of models and APIs within a project, without having to filter unnecessary files that prevent compilation.

This is what this modification allows.
By adding the **asManagedSources** parameter in the additional properties of the generator, the build.sbt, README.md and controller.scala files will be ignored.

It is then possible, in the build file of the project, to simply define sbt-openapi-generator as a source generator:

```
openApiAdditionalProperties := Map("asManagedSources" -> "true")
Compile / sourceGenerators += openApiGenerate
```
By doing that, not a single manual operation is needed when openapi specification changes.
Just by recompiling the project, the developer will see the templates and APIs updated and compiled at the same time.

Btw, syntax is far from perfect, due to the usage of a string to express the Boolean, but this is all due to sbt-openapi-generator...

### DIFF
--- a/docs/generators/scala-akka-http-server.md
+++ b/docs/generators/scala-akka-http-server.md
@@ -12,6 +12,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |apiPackage|package for generated api classes| |null|
 |artifactId|artifactId| |openapi-scala-akka-http-server|
 |artifactVersion|artifact version in generated pom.xml. This also becomes part of the generated library's filename| |1.0.0|
+|asManagedSources|Wether we want to generate an autonomous akka project, or sources to be included in sbt Managed Sources | If false, an autonomous project will be generated. If true, only source files to be added in ManagedSources will be generated|false]
 |dateLibrary|Option. Date library to use|<dl><dt>**joda**</dt><dd>Joda (for legacy app)</dd><dt>**java8**</dt><dd>Java 8 native JSR310 (prefered for JDK 1.8+)</dd></dl>|java8|
 |disallowAdditionalPropertiesIfNotPresent|If false, the 'additionalProperties' implementation (set to true by default) is compliant with the OAS and JSON schema specifications. If true (default), keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.|<dl><dt>**false**</dt><dd>The 'additionalProperties' implementation is compliant with the OAS and JSON schema specifications.</dd><dt>**true**</dt><dd>Keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.</dd></dl>|true|
 |ensureUniqueParams|Whether to ensure parameter names are unique in an operation (rename parameters that are not).| |true|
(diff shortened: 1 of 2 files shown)


wing328:
What about using .openapi-generator-ignore to have complete control on what files to skip during code generation?

cchafer:
> What about using .openapi-generator-ignore to have complete control on what files to skip during code generation?

I think it would not work.
Documentation says : 
> create a file named .openapi-generator-ignore in the root of the output directory with the contents

When doing a sbt clean, the output directory of openapi is deleted, meaning my .openapi-generator-ignore would also be deleted.
In addition of that, adding a target / generated / build directory in VCS would be a bad practice, and I don't want my whole team to have to recreate this file when cloning the project.

If I can add this file somewhere else, or reference it from another directory, si yes, it could work.

BUT... in the same time I still think i shouldn't have to specify by myself the name of the files.
You have an option to prevent models creation, even if people could add every model file in the openapi-generator-ignore file.
But for convenience, it's much easier.


PS: sorry for my crap english...

wing328:
```
@@ -12,7 +12,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |apiPackage|package for generated api classes| |null|
 |artifactId|artifactId| |openapi-scala-akka-http-server|
 |artifactVersion|artifact version in generated pom.xml. This also becomes part of the generated library's filename| |1.0.0|
-|asManagedSources|Wether we want to generate an autonomous akka project, or sources to be included in sbt Managed Sources | If false, an autonomous project will be generated. If true, only source files to be added in ManagedSources will be generated|false]
+|asManagedSources|Resulting files cab be used as managed resources. No build files or default controllers will be generated| |false|
 |dateLibrary|Option. Date library to use|<dl><dt>**joda**</dt><dd>Joda (for legacy app)</dd><dt>**java8**</dt><dd>Java 8 native JSR310 (prefered for JDK 1.8+)</dd></dl>|java8|
 |disallowAdditionalPropertiesIfNotPresent|If false, the 'additionalProperties' implementation (set to true by default) is compliant with the OAS and JSON schema specifications. If true (default), keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.|<dl><dt>**false**</dt><dd>The 'additionalProperties' implementation is compliant with the OAS and JSON schema specifications.</dd><dt>**true**</dt><dd>Keep the old (incorrect) behaviour that 'additionalProperties' is set to false by default.</dd></dl>|true|
Perform git status
On branch pull/9012

	modified:   docs/generators/scala-akka-http-server.md

```

I'll update the doc after merging this PR.

Let's go with what you've so far unless someone later comes up with a better approach.

Issue #5218, opened 2020-02-05
[REQ] Feature Request Description


dmcguire81:
### Is your feature request related to a problem? Please describe.

Several of the generators we've used (rust, typescript-axios, swift5) forcefully serialize and deserialize to JSON irrespective of the setting of the `Content-Type` header. For example, see this [needsSerialization declaration](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-axios/apiInner.mustache#L197), which serializes to JSON any body argument that is not a `string`, when the underlying library, `axios`, will happily handle native types that correspond to useful MIME types like `image/*`, etc.

This leaves me in a bind as an API designer: either use correct MIME types that tools like cURL and Postman handle well and forego meaningful code generation, or forego external, code-agnostic testing and diagnostic tools, and get code that works on shallow `application/json`-based objects that have MIME-types embedded in them.

## Describe the solution you'd like

I'd like to be able to use native MIME types where appropriate and useful.

## Describe alternatives you've considered

Modeling media as shallow objects that will encode to `application/json`, just for the purposes of code-generation.

## Additional context

Really, this is a philosophical question about how the owners see this tool being used. Should I file defects when I encode MIME types and they are not respected, as detailed above. Should I just make peace with the trade-offs, as described, and choose as best I can?

richardwhiuk:
In https://github.com/OpenAPITools/openapi-generator/pull/5339 the Rust Server generator was tweaked so that mime types for which the generator doesn't explicitly have support for (i.e. JSON and XML), are treated as unstructured data, and passed to the user as a string or a byte array.

I'm intrigued whether this would satisfy your requirements?

ahilke:
For `typescript-axios`, this has been fixed in https://github.com/OpenAPITools/openapi-generator/pull/7580, available in v5.0.0-beta3.

jclulow:
FYI: If I understand correctly, this issue may be at least related to #9021, where I'm trying to pass different `Into<Body>` things as the request body parameter for an endpoint that accepts `application/octet-stream`.

PR #9021, opened 2021-03-19
rust: better support for binary body parameters

jclulow:
The Rust-based reqwest client does not correctly handle a binary body
parameter today.  It creates a `PathBuf` parameter and attempts to pass
that to the `RequestBuilder` `json()` method, which sends the path itself
as the body of the request.

Instead, it would help to parameterise operations that accept a binary
body so that they may accept any object that implements the `Into<Body>`
trait.  This would allow the caller to pass a binary body in several
forms, such as a slice of bytes or a `Stream` that reads from some other
source.

Note possible related issue: #5218

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [x] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request. (cc @ahl @frol @farcaller @richardwhiuk @paladinzh)

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
@@ -18,9 +18,11 @@
 package org.openapitools.codegen.languages;
 
 import com.google.common.base.Strings;
+import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.media.ArraySchema;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.media.StringSchema;
+import io.swagger.v3.oas.models.servers.Server;
 import io.swagger.v3.parser.util.SchemaTypeUtil;
 import org.openapitools.codegen.*;
 import org.openapitools.codegen.meta.features.*;
@@ -501,6 +503,38 @@ public class RustClientCodegen extends DefaultCodegen implements CodegenConfig {
         }
 
         return StringUtils.underscore(sanitizedOperationId);
+    }
+
+    @Override
+    public CodegenOperation fromOperation(String path, String httpMethod, Operation operation, List<Server> servers) {
+        CodegenOperation op = super.fromOperation(path, httpMethod, operation, servers);
+
+        // If this endpoint has a single binary body parameter, we want
+        // to accept a body argument with the trait bound Into<Body>.
+        // This requires a type parameter in the function definition,
+        // in advance of actual parameter, and is thus far easier to decide
+        // here than in the template.
+        if (REQWEST_LIBRARY.equals(getLibrary()) && op.bodyParams != null && op.bodyParams.size() == 1) {
+            CodegenParameter cp = op.bodyParams.get(0);
+
+            if (cp.isBinary) {
+                cp.baseType = cp.dataType = "B";
+                op.vendorExtensions.put("x-rust-type-parameter", "B: Into<reqwest::Body>");
+
+                // A copy of the body parameter will appear somewhere in the
+                // allParams list.  We need to doctor that one as well.
+                int seen = 0;
+                for (CodegenParameter ap : op.allParams) {
+                    if (ap.isBinary && ap.isBodyParam) {
+                        ap.baseType = ap.dataType = "B";
+                        seen++;
+                    }
+                }
+                assert seen == 1;
+            }
+        }
+
+        return op;
     }
 
     @Override
--- a/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
+++ b/modules/openapi-generator/src/main/resources/rust/reqwest/api.mustache
@@ -88,7 +88,65 @@ pub {{#supportAsync}}async {{/supportAsync}}fn {{{operationId}}}(configuration: 
 
 {{/vendorExtensions.x-group-parameters}}
 {{^vendorExtensions.x-group-parameters}}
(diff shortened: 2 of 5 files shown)


jclulow:
While the sample application does not currently have any binary body parameters, I have one in another application I am working on.  This portion of the API definition document:

```
    "/v1/worker/task/{task}/chunk": {
      "post": {
        "operationId": "worker_task_upload_chunk",
        "parameters": [
          {
            "in": "path",
            "name": "task",
            "required": true,
            "schema": {
              "type": "string"
            },
            "style": "simple"
          }
        ],
        "requestBody": {
          "content": {
            "application/octet-stream": {
              "schema": {
                "type": "string",
                "format": "binary"
              }
            }
          },
          "required": true
        },
        "responses": {
          "201": {
            "description": "successful creation",
            "content": {
              "application/json": {
                "schema": {
                  "title": "UploadedChunk",
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "string"
                    }
                  },
                  "required": [
                    "id"
                  ]
                }
              }
            }
          }
        }
      }
    },
```

... becomes the following generated routine ...

```rust
pub async fn worker_task_upload_chunk<B: Into<reqwest::Body>>(
    configuration: &configuration::Configuration,
    task: &str,
    body: B,
) -> Result<crate::models::UploadedChunk, Error<WorkerTaskUploadChunkError>> {

    let local_var_client = &configuration.client;

    let local_var_uri_str = format!("{}/v1/worker/task/{task}/chunk", configuration.base_path, task=crate::apis::urlencode(task));
    let mut local_var_req_builder = local_var_client.post(local_var_uri_str.as_str());

    if let Some(ref local_var_user_agent) = configuration.user_agent {
        local_var_req_builder = local_var_req_builder.header(reqwest::header::USER_AGENT, local_var_user_agent.clone());
    }
    local_var_req_builder = local_var_req_builder.body(body);

    let local_var_req = local_var_req_builder.build()?;
    let local_var_resp = local_var_client.execute(local_var_req).await?;

    let local_var_status = local_var_resp.status();
    let local_var_content = local_var_resp.text().await?;

    if !local_var_status.is_client_error() && !local_var_status.is_server_error() {
        serde_json::from_str(&local_var_content).map_err(Error::from)
    } else {
        let local_var_entity: Option<WorkerTaskUploadChunkError> = serde_json::from_str(&local_var_content).ok();
        let local_var_error = ResponseContent { status: local_var_status, content: local_var_content, entity: local_var_entity };
        Err(Error::ResponseError(local_var_error))
    }
}
```

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
```diff
@@ -503,6 +505,38 @@ public String toOperationId(String operationId) {
...
+                // A copy of the body parameter will appear somewhere in the
+                // allParams list.  We need to doctor that one as well.
+                int seen = 0;
+                for (CodegenParameter ap : op.allParams) {
+                    if (ap.isBinary && ap.isBodyParam) {
+                        ap.baseType = ap.dataType = "B";
+                        seen++;
+                    }
+                }
+                assert seen == 1;
```
wing328, line 533: Should we use `assert` in test code only?

wing328: What about simply using `isBinary` in the mustache template instead of creating `x-rust-type-parameter` ?

jclulow: > Should we use `assert` in test code only?

I put the assert here because it should be an invariant.  The generated code won't be correct (something is seriously wrong) if there is more than one binary body parameter in the `allParams` list, when there was only one in the `bodyParams` list.  It seems like it would be better to crash rather than to generate incorrect code if that is ever true, as it presumably represents a bug in the code.

> What about simply using `isBinary` in the mustache template instead of creating `x-rust-type-parameter` ?

I'm not sure how to do that with the limited functionality available in the templates.  We need three things to be true:

- that it's binary
- that it's a body parameter
- that it's the only body parameter

In addition to these three conditions, we need to make that determination _outside_ of the `{{#allParams}}` loop in the template where we emit function arguments.  I could not see how to use the `isBinary` property of a specific index in the `bodyParams` list as a predicate in the appropriate part of the template, or how to use the length of the list as a predicate.

I also expect that as the Rust code generator becomes more complete we will need to do more of this.  That is, add more trait bounds and generic parameters as certain other types of arguments are added; e.g., today for strings we generate code like:

```rust
async fn endpoint(
    arg1: &str,
    arg2: &str,
)
```

But it would be more common to accept something like an `AsRef<str>`; e.g.,

```rust
async fn endpoint<S1: AsRef<str>, S2: AsRef<str>>(
    arg1: S1,
    arg2: S2,
)
```

This means you can pass either a `&str` or a `String` or, I believe, something that implements `Display`.

What are your thoughts?

wing328:
>I put the assert here because it should be an invariant. The generated code won't be correct (something is seriously wrong)

Don't think we should use assert in non-test code. We usually use condition to perform a check and throw an exception instead.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
wing328: >I put the assert here because it should be an invariant. The generated code won't be correct (something is seriously wrong)

Don't think we should use assert in non-test code. We usually use condition to perform a check and throw an exception instead.

dkatz23238:
Thanks Joshua for your work on implementing a more generic approach to uploading bytes, this specific feature is quite useful for a project we are building at www.smartlayers.io

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
jayvdb: @wing328 which assertion should be raised here?

jayvdb:
@bcourtine this seems relevant to the `TODO` you put in https://github.com/OpenAPITools/openapi-generator/pull/1890

jayvdb:
I believe that this PR should enhance https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/rust/petstore.yaml to include an endpoint which triggers this new code, and then the samples will generate the new code.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
wing328: maybe just a RuntimeException

aDogCalledSpot:
What's the status here? I need a Vec<u8> in the body and PathBuf isn't very helpful.

PR #9025, opened 2021-03-20
[typescript][axios] Handle boolean enum correctly

### LINKED ISSUE #9024: [BUG] typescript-axios - boolean enum is generated as string enum
##### Description
following schema
```json
    "schemas": {
      "BooleanEnum": {
        "enum": ["true", "false"],
        "type": "boolean"
      },
```
currently generates
```ts
export enum BooleanEnum {
    True = 'true',
    False = 'false'
}

```
This bug is affecting every typescript codegen, though a fix does not fix all of those

latest master

https://gist.github.com/haedaal/a771d6f3df337b624ae8a2c2d176c0db

##### Suggest a fix

It should generate this
```ts
export enum BooleanEnum {
    True = true,
    False = false
}
```
or this because of TS18033
```ts
export type BooleanEnum = true | false
```

haedaal:
closes #9024 

with following schema
```json
    "schemas": {
      "BooleanEnum": {
        "enum": ["true", "false"],
        "type": "boolean"
      },
```
https://gist.github.com/haedaal/a771d6f3df337b624ae8a2c2d176c0db

typescript-fetch currently generates
```ts
export enum BooleanEnum {
    True = 'true',
    False = 'false'
}
```

with this pr
```ts
export type BooleanEnum = true | false
```

by using 'type' rather than 'enum' we breaks consistency, but TS18033 does not allow this
```ts
export enum BooleanEnum {
    True = true, // Only numeric enums can have computed members, but this expression has type 'true' ...
    False = false
}
```

this pr has 2 changes.
1. by adding isBoolean, generate model correctly
2. fix typescript-axios modelEnum.mustache

accidentally typescript-angular is fixed without template changes, but many others need their own fixes on template.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


tusharchutani (approved):
This looks good

macjohnny:
What is the use case to have an enum with two boolean values?

haedaal:
I think it's main use case is boolean literal type, not an enum with two boolean value (though I took the example above)

```ts
type FooResponse = FooSucessResponse | FooFailureResponse
type FooSuccessResponse = {
  success: true,
  body: {
    foo: string
  }
}
type FooFailureResponse = {
  success: false,
  body: {
    reason: string
  }
}
```

by handling boolean enum correctly, typescript can infer body type if value of success is given

haedaal:
@tusharchutani any plan to merge it..?

amakhrov:
Is it even a valid or reasonable schema in the first place?
I know it satisfies the validator - but the following one also does.
```
type: boolean
enum: [true, false, fuzzy]
```
So it tells us more about strictness of the validator, rather than correctness of the spec.

`type: boolean` already unambiguously indicates that the only two possible values are `true` and `false` (but not `"true"` or `"false"`!)

In this case we just need to decide whether we give priority to `type` attribute or `enum` attribute. In the former case, the generated code should not emit an additional type at all - just use `boolean` for the model property. In the latter case a string enum, as it's generated today, sounds correct.

kenisteward:
This is true the original schema is not valid. If you want to use the enum
properly it should not be 'true' but just true

kenisteward:
Technically the base validation is invalid due to the fact that the items
'true' and 'false' are not actually Boolean anyway.

haedaal:
as a result of assigning string value to the value of boolean enum,
following pattern is required for now
```ts
if (response.success === 'true') { }
```

haedaal:
@amakhrov does it make sense? I'm wondering what's blocking point.

amakhrov:
@haedaal I still struggle to understand the use case, honestly. Why define a boolean value in the spec as an enum? Why are the enum values are defined as strings in the spec - even though you clearly want them to be boolean in the generated code.

haedaal:
> @haedaal I still struggle to understand the use case, honestly. Why define a boolean value in the spec as an enum? Why are the enum values are defined as strings in the spec - even though you clearly want them to be boolean in the generated code.


@amakhrov 

1. Why define a boolean value in the spec as an enum?
```ts
type GetNextOrderResponse = {success: true, data: Order} | {success: false, data?: unknown} // or any other FailReason type for data in fail case
```
2. Why are the enum values are defined as strings in the spec - even though you clearly want them to be boolean in the generated code.
> that's the whole point of this pr. now codegen is generating string enum value rather than boolean enum value even though I clearly specify them as a boolean! this pr is a bugfix (if not a hotfix)

haedaal:
> This is true the original schema is not valid. If you want to use the enum properly it should not be 'true' but just true
> […](#)

that's exactly what I thought is a problem, but if you try it, making schema like this
```json
    "schemas": {
      "BooleanEnum": {
        "enum": [true, false],
        "type": "boolean"
      },
    }
```
does not fix the wrong result.

it's because we check only for number type and unconditionally wrap it's value with double quote as you can see here.
And this pr add "boolean" type to the cases where you have to use value as it is.

https://github.com/OpenAPITools/openapi-generator/blob/11deb438290483410b34ddf8dc10a858dc215244/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L249-L255

https://github.com/OpenAPITools/openapi-generator/blob/5d58afe5e79c425452716d48dc7f6c2bf39147d7/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L719-L723

amakhrov:
Ok, I think now I see what you're trying to do (sorry, it took me a while :) ).

You don't actually need `BooleanEnum = true | false` (since it's redundant). You want smth like `TrueBooleanEnum = true` and `FalseBooleanEnum = false` - is that correct?

If so, I would rather try to avoid generating an `enum` in the first place. The ideal generated model would be exactly as in your example
```
{success: true, data: Order}
```
(no enum whatsoever)

However, I admit this would probably be a bigger change in the templates. So as a quick solution your PR makes total sense to me

amakhrov (approved):

haedaal:
> You don't actually need `BooleanEnum = true | false` (since it's redundant). You want smth like `TrueBooleanEnum = true` and `FalseBooleanEnum = false` - is that correct?

Yes! Exactly! :)

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
```diff
@@ -64,7 +64,7 @@
     public String defaultValue;
     public String arrayModelType;
     public boolean isAlias; // Is this effectively an alias of another simple type
-    public boolean isString, isInteger, isLong, isNumber, isNumeric, isFloat, isDouble, isDate, isDateTime, isShort, isUnboundedInteger;
+    public boolean isString, isInteger, isLong, isNumber, isNumeric, isFloat, isDouble, isDate, isDateTime, isShort, isUnboundedInteger, isBoolean;
```
wing328, line 67: The following function also needs to be updated as well:

https://github.com/OpenAPITools/openapi-generator/blob/5d58afe5e79c425452716d48dc7f6c2bf39147d7/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L764

https://github.com/OpenAPITools/openapi-generator/blob/5d58afe5e79c425452716d48dc7f6c2bf39147d7/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L854

https://github.com/OpenAPITools/openapi-generator/blob/5d58afe5e79c425452716d48dc7f6c2bf39147d7/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java#L872

I'll update these in another PR after merging this one.

wing328:
cc @OpenAPITools/generator-core-team as the change impacts the codegen model class.

modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenModel.java
wing328: Filed https://github.com/OpenAPITools/openapi-generator/pull/9802

PR #9073, opened 2021-03-25
csharp-netcore Adding template files for HttpClient library

devhl-labs:
Added template files for HttpClient library. This should be identical to the HttpClient branch except for https://github.com/OpenAPITools/openapi-generator/commit/a88e01f80f1a4922f7db407b9ef68f91fb73bc15 and https://github.com/OpenAPITools/openapi-generator/pull/9073/commits/c4c19900449467d701575e0c07b6c0c1c44ec777

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


devhl-labs:
Actually I think this will override the HttpClient from Blackclaw wont it? There are significant differences, so that's not great. For instance I require an HttpClient to be injected and I support null reference types.

Blackclaws:
So I think we should talk about where this implementation of httpclient support should land if it lands. The name httpclient is now already taken by the implementation I supplied and changing that to be a totally different implementation might be confusing.

I'm still rather unhappy about the fact that there is the need to postprocess the output from this generator, which I think is not actually needed if you play with the generator / the templates a bit more and is confusing for people using this in a CI/CD setup.

I haven't made a point for point comparison between the code on both sides, it appears there are some features not supported by your version however such as file upload, custom user agents, no setting of basePath per Api instance, etc.

As it is right now I would not recommend merging it into master.

devhl-labs:
I started my work on this merely writing new templates because it appeared the demand for HttpClient was being neglected. Using the library feature to host my templates was Cheng's idea, not mine. So seeing your iteration hosted in this library folder was confusing, even for me. 

Obviously the post processing should be moved. Writing the templates is a big effort, and I did not intend to do it all myself in one pr. 

The HttpClient is injected into this library, so the things like basePath, retry logic, short circuiting are handled by the end users before the templates are even instantiated. You should not be creating a new HttpClient inside the template files. You should be injecting either an HttpClient, or perhaps an IHttpClientFactory.

Blackclaws:
> I started my work on this merely writing new templates because it appeared the demand for HttpClient was being neglected. Using the library feature to host my templates was Cheng's idea, not mine. So seeing your iteration hosted in this library folder was confusing, even for me.
> 

I think the idea was that my version since it provided all the existing features that people were expecting from the csharp-netcore generator fits better as the httpclient iteration of that generator. I think your version would fit better as a new generator alltogether csharp-experimental for example.


> Obviously the post processing should be moved. Writing the templates is a big effort, and I did not intend to do it all myself in one pr.

Gotcha. The reason I didn't expand upon your pull request and instead opted to do the work from scratch was that it seemed like your direction was going to be breaking with existing API surfaces for the client which is something I wanted to avoid. I think you have some good ideas in your code and I also think some refactoring might be needed to streamline the current state a bit more.

Introducing support for nullable reference types will also require checking the targetFramework but might be a welcome addition for some.

> 
> The HttpClient is injected into this library, so the things like basePath, retry logic, short circuiting are handled by the end users before the templates are even instantiated. You should not be creating a new HttpClient inside the template files. You should be injecting either an HttpClient, or perhaps an IHttpClientFactory.

I think that itself means people need more knowledge about how to use this version, if you expect them to pass in the HttpClient by default, they might also be falling into the trap that they instantiate a new HttpClient anytime they instantiate the API class itself.

Creating a new HttpClient inside the generated code is fine as long as you do it sparingly. My initial iteration was too close to the RestSharp way of things so it didn't catch the real big issue popping up even if reusing the client with one HttpClient per ApiClient. 

If you take a look at my latest pull request that fixes this issue its now possible to set the HttpClient manually before first usage of the APIClient, if its not set it will be created by the APIClient.

Setting such things as basePath in HttpClient doesn't work in case you have an API that accesses different hosts but maybe wants to share the HttpClient anyway. IHttpClientFactory would be great but it again introduces a dependency on a NuGet package so that was a no-go for me (it works for .Net 5 but we are targeting .Net Standard 2.0 compatibility).

devhl-labs:
> Creating a new HttpClient inside the generated code is fine as long as you do it sparingly. My initial iteration was too close to the RestSharp way of things so it didn't catch the real big issue popping up even if reusing the client with one HttpClient per ApiClient.

No it's not fine, because then the end users can't control retry logic, short circuiting, message handlers, and the myriad of other things a user may want to do in their client.

> Setting such things as basePath in HttpClient doesn't work in case you have an API that accesses different hosts but maybe wants to share the HttpClient anyway. IHttpClientFactory would be great but it again introduces a dependency on a NuGet package so that was a no-go for me (it works for .Net 5 but we are targeting .Net Standard 2.0 compatibility).

In that case and users would use a named HttpClient configured the way they want and instantiate a new api.mustache object.

Blackclaws:
> > Creating a new HttpClient inside the generated code is fine as long as you do it sparingly. My initial iteration was too close to the RestSharp way of things so it didn't catch the real big issue popping up even if reusing the client with one HttpClient per ApiClient.
> 
> No it's not fine, because then the end users can't control retry logic, short circuiting, message handlers, and the myriad of other things a user may want to do in their client. Here is a good example, if I were to use your templates in one of my solutions right now, it wouldn't work at all, because you're not setting the MaxConnectionsPerServer property (I assume). It would work for a few dozen requests but the time to reply would slowly increase, and then I would get a dozen timeouts.
> 

Right now retry logic is configurable via the retry policies using Polly. If you want to adjust the HttpClient or its Handler manually you can do that now after the last pull request by simply accessing

```csharp
{{packageName}}.Api.ApiClient.HttpClient
{{packageName}}.Api.ApiClient.HttpClientHandler
```
as they are now static properties of the ApiClient class. So you can do whatever you want in order to change the base HttpClient being used for a specific ApiClient. It currently doesn't allow you to specify an HttpClient per API which is something we might want to evaluate as well.

> > Setting such things as basePath in HttpClient doesn't work in case you have an API that accesses different hosts but maybe wants to share the HttpClient anyway. IHttpClientFactory would be great but it again introduces a dependency on a NuGet package so that was a no-go for me (it works for .Net 5 but we are targeting .Net Standard 2.0 compatibility).
> 
> In that case and users would use a named HttpClient configured the way they want and instantiate a new api.mustache object.
> 

You could do that as well, true. Its just a break from the current approach using the Configuration class, hence why I think it might be an issue for people currently using csharp-netcore that just want to migrate away from RestSharp.

> I can explore starting a new generator. It's just unfortunate that my csharp-netcore templates wont be in the csharp-netcore folder.

We could also talk about yours being httpclient-experimental as it breaks quite a few existing conventions.

PR #9074, opened 2021-03-25
[swift5] Add useSPMFileStructure

aymanbagabas:
Signed-off-by: Ayman Bagabas <ayman.bagabas@gmail.com>

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator/src/main/resources/swift5/Package.swift.mustache
```diff
@@ -35,7 +35,7 @@ let package = Package(
         .target(
             name: "{{projectName}}",
             dependencies: [{{#useAlamofire}}"Alamofire", {{/useAlamofire}}{{#usePromiseKit}}"PromiseKit", {{/usePromiseKit}}{{#useRxSwift}}"RxSwift"{{/useRxSwift}}],
-            path: "{{projectName}}/Classes"
+            path: "{{#useSPMFileStructure}}Sources{{/useSPMFileStructure}}{{^useSPMFileStructure}}{{projectName}}/Classes{{/useSPMFileStructure}}"
```
wing328, line 38: What about adding an option `swiftPackagePath` to customize the whole value instead and default it to `{{projectName}}/Classes` so that users have full control of it?

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -414,6 +418,16 @@ public void processOpts() {
             additionalProperties.put(POD_AUTHORS, DEFAULT_POD_AUTHORS);
         }
 
+        if (additionalProperties.containsKey(USE_SPM_FILE_STRUCTURE) && (boolean)additionalProperties.get(USE_SPM_FILE_STRUCTURE)) {
+            setSwiftUseApiNamespace(convertPropertyToBooleanAndWriteBack(SWIFT_USE_API_NAMESPACE));
```
4brunu, line 422: @aymanbagabas I think maybe what you wanted to do was `setUseSPMFileStructure(convertPropertyToBooleanAndWriteBack(USE_SPM_FILE_STRUCTURE));`?

4brunu:
Otherwise looks good to me 👍 
Could you please create a new sample project to ensure this works and keep working as expected? Or change a current one to use this flag?
Thanks

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
```diff
@@ -414,6 +418,16 @@ public void processOpts() {
             additionalProperties.put(POD_AUTHORS, DEFAULT_POD_AUTHORS);
         }
 
+        if (additionalProperties.containsKey(USE_SPM_FILE_STRUCTURE) && (boolean)additionalProperties.get(USE_SPM_FILE_STRUCTURE)) {
+            setUseSPMFileStructure(convertPropertyToBooleanAndWriteBack(USE_SPM_FILE_STRUCTURE));
+            sourceFolder = "Sources";
```
4brunu, line 423: @aymanbagabas according to https://swift.org/package-manager/
`By convention, a target includes any source files located in the Sources/<target-name> directory.`
So I would suggest to change the structure to `Sources/ProjectName`.
What do you think?

samples/client/petstore/swift5/urlsessionLibrary/.openapi-generator/FILES
```diff
@@ -2,64 +2,64 @@
 Cartfile
 Package.swift
 PetstoreClient.podspec
```
4brunu, line 4: @aymanbagabas I almost forgot that we need to update the `podspec` to match the new file structure.
Could you please update the property `source_files` in the `podspec` file?
Thanks

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/Swift5ClientCodegen.java
aymanbagabas: I think you're right, at some point, I dropped the projectName from the source folder somehow.

4brunu:
One last thing that I noticed is that the options `USE_SPM_FILE_STRUCTURE` and `SWIFT_PACKAGE_PATH` don't apear in the docs.

To make them apear, you make something like 
```
 cliOptions.add(new CliOption(PROJECT_NAME, "Project name in Xcode"));
 cliOptions.add(CliOption.newBoolean(USE_COROUTINES, "Whether to use the Coroutines adapter with the retrofit2 library.")); 
```

aymanbagabas:
@4brunu do you have any idea why this is failing? I couldn't tell what I'm missing in the last commit.

```
[ERROR] Tests run: 1437, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 66.084 s <<< FAILURE! - in TestSuite
[ERROR] checkOptionsProcessing(org.openapitools.codegen.swift5.Swift5OptionsTest)  Time elapsed: 0.019 s  <<< FAILURE!
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Boolean
[ERROR] Failures: 
[ERROR]   Swift5OptionsTest>AbstractOptionsTest.checkOptionsProcessing:51 » ClassCast ja...
[ERROR] Tests run: 1437, Failures: 1, Errors: 0, Skipped: 0
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M3:test (default-test) on project openapi-generator: There are test failures.
```

4brunu:
Looks like you already fix it.
Can you run the following commands to update the sample projects and docs?
That's why circle ci is failing.
```
./mvnw clean package 
./bin/generate-samples.sh
```

4brunu:
CI is now passing.
@wing328 for me you can merge this PR

docs/generators/swift5.md
```diff
@@ -33,7 +33,9 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |responseAs|Optionally use libraries to manage response.  Currently PromiseKit, RxSwift, Result, Combine are available.| |null|
 |sortModelPropertiesByRequiredFlag|Sort model properties to place required parameters before optional parameters.| |true|
 |sortParamsByRequiredFlag|Sort method arguments to place required parameters before optional parameters.| |true|
+|swiftPackagePath|Set a custom source path instead of OpenAPIClient/Classes/OpenAPIs.| |null|
 |swiftUseApiNamespace|Flag to make all the API classes inner-class of {{projectName}}API| |null|
+|useSPMFileStructure|Use SPM file structure and set the source path to Sources/{{projectName}} (default: false).| |null|
```
wing328, line 38: @aymanbagabas thanks for taking my feedback.

Shouldn't just `swiftPackagePath` would be enough? In other words, we want to avoid duplicated options do very similar things.

4brunu:
I would like to give you some context regarding this PR.
The current file structure is `Classes/OpenAPIs/*` because it was like this since the beginning and it’s fine, there is nothing wrong with this.
But then Apple introduced Swift Package Manager(SPM), that is an official package manager for Swift.
With that a standard for the file structure was introduced, `Sources/projectName/*`.
Since that was introduced, the libraries in the community are being migrated to this new structure.
I think that we should also migrate for this file structure, for good practice reasons.
The only problem is that that change would introduce a breaking change, for people that integrate the generated code manually. For those that use the SPM, Cocoapods or Carthage integration it will work automatically.
I think that a custom path isn’t needed here, because we are implementing a community standard.

My suggestion is that we go with Three of the two options:
1 - Make the breaking change, because this is a community standard, it's very easy to fix, and Swift 5 generator is still in beta.
2 - Make the breaking change, but also introduce a boolean that allows the user go back to the old behaviour.
3 - Don't make the breaking change, by introducing a boolean, and keep the legacy behaviour on by default.

@aymanbagabas @wing328 what do you think of this?

aymanbagabas:
TBH, I'm all for making the breaking change since the generator is still in beta and users don't expect it to be stable. But instead of introducing a boolean to allow the old behavior, users could use `swiftPackagePath` to set the old behavior back.

wing328:
Only `experimental` generators allows breaking changes in the upcoming patch release (5.1.1)

I prefer 3

> 3 - Don't make the breaking change, by introducing a boolean, and keep the legacy behaviour on by default.

In 5.2.0, we can change the option to use the new behaviour by default.

4brunu:
Looks option 3 also looks good to me.
And what about the option `swiftPackagePath` should we keep it along side with `useSPMFileStructure`?
Or should we just keep `useSPMFileStructure`?
For me I think we can just keep `useSPMFileStructure.`
@aymanbagabas @wing328 what do you think?

modules/openapi-generator/src/main/resources/swift5/Podspec.mustache
```diff
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   {{#podDocumentationURL}}
   s.documentation_url = '{{podDocumentationURL}}'
   {{/podDocumentationURL}}
-  s.source_files = '{{projectName}}/Classes/**/*.swift'
```
4brunu, line 28: @aymanbagabas another place that I remembered that I think we need to update the path is in the `XcodeGen.mustache` file, the `sources` property

wing328:
> what do you think?

Prefer one option instead of two for simplicity.

aymanbagabas:
I think this discussion should've taken place earlier. Anyway, since both options have been implemented already and work as expressed, I don't see the point of dropping one of them.

As @wing328 mentioned, we can change this behavior and make the breaking changes in the 5.2.0 release. Drop `useSPMFileStructure`, and set the default source path to `Sources/{{projectname}}` and use `swiftPackagePath` to have the old behavior back.

4brunu:
Looks good to me.
@aymanbagabas could you please fix the merge conflicts and the `XcodeGen.mustache` `sources` property?

modules/openapi-generator/src/main/resources/swift5/XcodeGen.mustache
```diff
@@ -4,7 +4,7 @@ targets:
     type: framework
     platform: iOS
     deploymentTarget: "9.0"
-    sources: [{{projectName}}]
+    sources: [{{#swiftPackagePath}}{{swiftPackagePath}}{{/swiftPackagePath}}{{^swiftPackagePath}}{{#useSPMFileStructure}}Sources/{{projectName}}{{/useSPMFileStructure}}{{^useSPMFileStructure}}{{projectName}}/Classes{{/useSPMFileStructure}}{{/swiftPackagePath}}]
```
4brunu, line 7: @aymanbagabas Here the logic is a bit different, XcodeGen only need's the root directory.
If should be `swiftPackagePath` or `Sources` or `projectName`

```
sources: [{{#swiftPackagePath}}{{swiftPackagePath}}{{/swiftPackagePath}}{{^swiftPackagePath}}{{#useSPMFileStructure}}Sources{{/useSPMFileStructure}}{{^useSPMFileStructure}}{{projectName}}{{/useSPMFileStructure}}{{/swiftPackagePath}}]
```

aymanbagabas: Note that `swiftPackagePath` won't be the root directory, it would be the full path to the source files.

4brunu: Yes, I know.
It will work as well.
This change is more to keep the current behaviour for the existing `XcodeGen.`

aymanbagabas: Changes pushed

4brunu:
@aymanbagabas thanks for the changes.
Looks good to me.
@wing328 for me this PR is ready to merge 👍

wing328:
Sorry for the delay. PR merged.

Issue #9060, opened 2021-03-24
[BUG][csharp-netcore] C# Client for Net Core generate port exaustion


lucamazzanti:
##### Description
I moved from Net Framework 4.7 to Net Core 3.1 and 5 and I have a critical issue.

The csharp-netcore client uses RestSharp and is not stable with Net Core actually. See [this](https://github.com/restsharp/RestSharp/issues/1406) issue. 
It seems to be a net core breaking change they won't patch in time and actually this is impacting on RestSharp.

It silently open at every call a new socket, creating a new HttpWebRequest each time. 
This is a behavior totally different in NetFramework where it keep the same.
You can reproduce it easily generating a simple client and using it in both verisons.
Test multiple calls and watch the connection in Wireshark, I can further add images or a report if you prefer.

While NetFramework honors the keep-alive header (supported also by the server) in Net Core it doesnt.
- In NetFramework  it do the handshake and TLS connection just 1 time, and it consumes just one socket keeping working on it with the next calls.
- In Net Core not, it open a new socket each times.

This is critical because you cannot understand it when moving your client to net core until you have a feedback from the server closing the connections to all.

I see there is a version with HttpClient, i tried it but it create every time:
` var handler = new HttpClientHandler(); var client = new HttpClient();`

##### openapi-generator version
openapi-generator-cli-5.1.0.jar with csharp-netcore client without settings.

wing328:
The latest version of OpenAPI Generator v5.1.0 supports HttpClient (--library httpclient). Please give it a try.

lucamazzanti:
I see there is a version with HttpClient, i tried it but it create every time:

```
var handler = new HttpClientHandler(); 
var client = new HttpClient();
```

So the issue remains

devhl-labs:
You'll want this pr once it's merged. You pass in the HttpClient, so you control it entirely.

https://github.com/OpenAPITools/openapi-generator/pull/9073

Note that it requires a bit of post processing, but here is a script for it, just edit the parameters to suite your needs.

```ps1
java -jar "../openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar" generate `
    -g csharp-netcore `
    -i "your-swagger-file.yml" `
    -c generator-config.json `
    -o output `
    -t templates `
    --library httpclient

$files = Get-ChildItem output -Recurse | Where-Object {-Not($_.PSIsContainer)}
foreach ($file in $files)
{
[...]

PR #9085, opened 2021-03-25
[csharp][netcore-httpclient] Reuse HttpClient, Allow use of external HttpClient, Fix Socket Exhaustion, Alternative With Constructor Injection

Blackclaws:
Fixed an existing issue that made the HttpClient open new sockets for each connection even if being reused as the UserAgent was being set on the Client and not the request. Also makes HttpClient settable from external.
Fixes: https://github.com/OpenAPITools/openapi-generator/issues/9060

Alternative to https://github.com/OpenAPITools/openapi-generator/pull/9080

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
```diff
@@ -189,30 +187,28 @@ namespace {{packageName}}.Client
...
-            _baseUrl = {{packageName}}.Client.GlobalConfiguration.Instance.BasePath;
-    {{#reUseHttpClient}}
-            _httpClientHandler = new HttpClientHandler();
-            _httpClient = new HttpClient(_httpClientHandler);
-    {{/reUseHttpClient}}
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ApiClient" />
+            /// Initializes a new instance of the <see cref="ApiClient" />
```
wing328, line 195: Minor suggestion: remove these spaces

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
```diff
@@ -189,30 +187,28 @@ namespace {{packageName}}.Client
...
 
             _baseUrl = basePath;
-    {{#reUseHttpClient}}
-            _httpClientHandler = new HttpClientHandler();
-            _httpClient = new HttpClient(_httpClientHandler);
-    {{/reUseHttpClient}}
+            /* TODO: Decide how to handle this case
+            if(client != null && handler == null) {
+                throw new ArgumentException("if providing HttpClient, you also need to provide its handler");
+            }*/
```
wing328, line 208: From the code sample in the readme:
```
HttpClientHandler yourHandler = new HttpClientHandler();
HttpClient yourHttpClient = new HttpClient(yourHandler);
```
Looks like if the HttpClient is provided, one can obtain its HttpClientHandler, right?

If that's the case, I would suggest simply ask for HttpClient in the constructor to make it easier for the user.

Blackclaws: That is unfortunately not the case. You can not acquire the handler from the HttpClient once it is constructed. The handler is  used for setting/retrieving cookies as well as configuring the proxy and client certificates. We could say that in case a user passed in their own HttpClient that they would be handling all of that themselves or simply have to live with the fact that it doesn't work. Hence why I was wondering whether we should at least give some kind of warning.

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
Blackclaws: Will do. I'll also maybe make the HttpClient private again on the class as it is now injected and doesn't need to be modified after creation anymore.

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
wing328: Just saw this: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-5.0#httpclient-and-net-core

Shall we simply use `HttpClient = new HttpClient();` as it should pick the best available transport based on the platform? In other words, ApiClient will not store the HttpClientHandler.

Blackclaws: Again, we are no longer able to set proxy options and client certificates anymore if we do that. Instantiating HttpClientHandler creates an underlying actual handler which is SocketsHttpHandler.

Blackclaws: Just to be clear, what we are doing by default is exactly what HttpClient and its default constructor are doing:
        public HttpClient()
            : this(new HttpClientHandler())
        {
        }

lucamazzanti: Having a unique public static instance of HttpClient and HttpClientHandler is not the right way.
You need to inject them from the ctor. You cannot have the lifecycle of each ApiClient bound to this singletons, is not the way is intended to use in net core.

Blackclaws: We do not have a static instance in this pull request

lucamazzanti: Sorry i reviewed the other one.
It is correct if you inject it, but if it is created inside then is no more disposed.

Blackclaws: If possible you keep the HttpClient alive for as long as possible, if you create a new one per request even if you dispose of it you will face SocketExhaustion pretty soon. HttpClientFactory would be a good solution its not available for .netstandard 2.0 however only as a nuget package.

You are correct that we should probably dispose the HttpClient when the ApiClient is disposed, if it was created initially. I'll add that.

lucamazzanti: Reading the [guide](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net)
Here the scenarios:
- Disposing each time the handler is not enough because windows has a tcp timeout that keep connected the socket after the dispose (I can also say that there is also the keep-alive default true to think about), so socket exaustion. So the handler must be pooled, the client must be recreated each times fith the disposeHandler parameter set to off.
- keeping static shared is viable only for some scenarios like small lifed programs (for the DNS issue, if the handler became unhealthy).
So the best way is to:
1. create each times a client, with disposeHandler false, passing the right handler pooled
2. keep an in memory concurrent pool with an evict time to avoid the long timed issue

And this is what the [DefaultHttpClientFactory](https://github.com/aspnet/HttpClientFactory/blob/master/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs) does (see also [here](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#benefits-of-using-ihttpclientfactory)).
Unfortunately the factory has high coupling with the DI packages (logging, options, ...)

So possibilities:
- we code, test, something similar
- we avoid that at all, keeping only the constructor with the injection
- document well the usage of a new HttpClient\Handler not disposed each call, so people is aware of that 

I hope i can help you all to build the netcore client, let me know your thoughts.

Blackclaws: Actually found another bug while implementing this.

lucamazzanti: We can actually dispose the client if created inside, for sure is better than dont do that at all.
And document the issue in the ctor or in a good place, to let uderstand clearly to the devs that if they don't manage theirself the lifecycle outside, the behavior is to create\dispose.

Blackclaws: Check the latest commit. It now disposes when the ApiClient or the ApiAccessClass is disposed. If not passed in one HttpClient per ApiAccessClass is created.

lucamazzanti: The commit is ok, it is all right, to bound the httpClient to the entire apiClient lifecycle can emprove multiple calls, and I like it. One thing: if you pass the handler to the client ctor and you dispose both at the end, you need to pass also the second parameter disposeHandler to false. In that way you dispose only the one created inside in all the scenarios.

On the counterpart to moving the ApiClient to an IDisposable in't too much strong? 
The disposable interface will lead to the using scenario and scopes, and this is not useful when injected. 
Another possibility is to keep the lifecycle of the client\hander (if created) inside the ExecAsync call:
`
private async Task<ApiResponse<T>> ExecAsync<T>(HttpRequestMessage req .... )
{
       var httpClientHandler = _httpClientHandler ?? new HttpClientHandler();
       var disposeHandler = _httpClientHandler == null;
       var httpClient = _httpClient ?? new HttpClient(httpClientHandler, false);
       var disposeClient = _httpClient == null;

       ....

       if(disposeClient ) httpClient.Dispose();
       if(disposeHandler) httpClientHandler.Dispose();

       return result;
}
`

Blackclaws: No, we will not scope the lifecycle to the ExecAsync. We have that in the current version, it leads to socket exhaustion even with disposing, so that is not an option. If you dispose the handler manually if its already disposed nothing will happen, so that is fine to do.

Moving ApiClient to IDisposable is no issue at all.

lucamazzanti: Ok fine, watch out the disposeHandler parameter and all is fine.

Blackclaws: Yep added that

lucamazzanti: I just realized that I cannot get the handler from the client, and the DI give us just the client yet configured with inside the handler. I will check the NSwag implementation that pass just the client to understand better.

lucamazzanti: Ok NSwag is more simple than that, there is no support for that options related to the handler.
So, if I inject just the client (i think the normal behavior), i loose some settings, this can lead to some misunderstand.
If I inject both, i have the all configuration woriking, but is not so clean in the base DI of a net core project.
I think we can document the settings and the ctor to warn about that.
I also prefer to document some remarks in the ctor to inform in the intellisense about the optional parameters, the behaviors, and the dispose meaning.
Do you think is too much?

Blackclaws: > I just realized that I cannot get the handler from the client, and the DI give us just the client yet configured with inside the handler. I will check the NSwag implementation that pass just the client to understand better.

Yeah, that limitation is the reason why I have to ask for the handler in the first place. If you inject just the client you will get an exception right now telling you you need to also inject the handler.

Blackclaws: > Ok NSwag is more simple than that, there is no support for that options related to the handler.
> So, if I inject just the client (i think the normal behavior), i loose some settings, this can lead to some misunderstand.
> If I inject both, i have the all configuration woriking, but is not so clean in the base DI of a net core project.
> I think we can document the settings and the ctor to warn about that.
> I also prefer to document some remarks in the ctor to inform in the intellisense about the optional parameters, the behaviors, and the dispose meaning.
> Do you think is too much?

If you'd like to add further details to the parameters feel free to fork my version of the repo and open a pull request to my branch with the changes.

lucamazzanti: I also thinked about the validation of both parameters as you.
But actually i think that I won't to get a complex DI for these few not so useful parameters.

Blackclaws: Yeah you need customize the DI injection by running a lambda function that then sets the correct configuration at that point.

Right now adding just the HttpClient is not a supported UseCase. If you say this is relevant for you as well, and you want to use the client from IHttpClientFactory then we'll figure out a way to do that. 

I talked about this also with @wing328 and we actually wanted to add a way to do this later if people request it. It appears you may have a usecase already?

Blackclaws: I've added a new version that includes a switch to disable the exception and the features that don't work if handler isn't present.

lucamazzanti: I don't understand why in those 3+ years of net core, no one had this issue using a swagger codegen.
Maybe because in the official startup documentation they promote NSwag, effectively the only one actually support the injection of an HttpClient. They also underline very well the issues\tradeoffs in the HttpClient and how they manage it with the factory in the DI.

We moved to net core with a few projects in this years, those services consume others using rest api, with a Swagger codgen. In this days I monitored a strange pressure on the production environment, and I found the socket exsaustion, done by ourself with a chatty client. I compared the different behavior of the vanilla codegen in both framework, monitoring with wireshark:
- one uses 1 socket, 1 TCP handshake, 1 TLS handshale, N calls
- the other one open each times the pipeline consuming sockets

The breaking change is inside RestSharp, it has an issue open but the project seems aware to this difficult change to support net core and I think it has no future plans to fix it. The real change is inside the HttpWebRequest that actually in Net Core no more has a static lifecycle and create a new HttpClient.

So RestSharp is not a valid production library for net core, at all.
This means also Swagger is not a valid choice for net core.
Swagger seems no more active, the main author, William, is here now on OpenApi, reviewing you now (William thank you for the projects, they are great). 

So the choice for us are:
- create new contracts with NSwag for the net core projects, leaving Swagger.
- cooperate on OpenApi to have a net-core client fully integrated in the core DI. OpenApi c# codegen is 99% the same as swagger so moving to OpenApi it seems the right way to continue using Swagger.

Your fix is just enough to avoid the exsaustion now.
But having a client that not fit easily in the DI can move us to choose another more solid way for the future.
But I also see a lot of interest in this PR and on this project to move to a working client for net core so I will await and cooperate with you if I can give some support.

wing328: I've different suggestions but let's go with what you've so far. We can resume the discussion after the merge.

wing328: I've different suggestions but let's go with what you've so far. We can resume the discussion after the merge.

wing328:
> I've added a new version that includes a switch to disable the exception and the features that don't work if handler isn't present.

Instead of adding a switch, what about adding a constructor that takes just one argument (HttpClient) as follows instead?
```
        public ApiClient(HttpClient client = null) :

        public ApiClient(HttpClient client = null, HttpClientHandler handler) :
```
(the handler cannot be null)

lucamazzanti:
If I could I also want to expose with a constructor in the api part the inject of an httpclient to complete the support, I will open a PR.

lucamazzanti:
> > I've added a new version that includes a switch to disable the exception and the features that don't work if handler isn't present.
> 
> Instead of adding a switch, what about adding a constructor that takes just one argument (HttpClient) as follows instead?
> 
> ```
>         public ApiClient(HttpClient client = null) :
> 
>         public ApiClient(HttpClient client = null, HttpClientHandler handler) :
> ```
> 
> (the handler cannot be null)

I created a PR #9109 with that change, if you and @Blackclaws find that changes useful.

wing328:
@lucamazzanti thanks for the PR 👍 . Will try to review later today.

PR #9100, opened 2021-03-26
[Dart2] Add initial support to use an isolate to offload JSON serialization/deserialization

noordawod:
I have noticed that as my API grew and its associated data requirements grew with it, a Flutter app I work on started showing hickups (jank) when a lengthy JSON hit the wire. So, I wanted to offload to a `compute()` function but was unable due to the fact that `ApiClient._deserialize` is private. Hence this pr.

In addition, I had to adjust the signatures of some functions and methods to be futures and that means that this pr is not backward-compatible. So if users have overridden `serialize()` or `deserialize()` then they'll need to update their code to match the new signature. A matter of 5 minutes but the change allows a better concurrency regardless if one uses an isolate or not.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@ircecho (2017/07) @swipesight (2018/09) @jaumard (2018/09) @athornz (2019/12) @amondnet (2019/12) @wing328 @sbu-WBT @agilob


agilob (approved):

agilob:
While technically this is backwards incompatible change, it doesn't feel like it. Good practice should be accepted.

noordawod:
> While technically this is backwards incompatible change, it doesn't feel like it. Good practice should be accepted.

That's great to hear! Thanks for the approval. I'll get the pr to pass all tests asap.

agilob:
It's still up to @wing328  to decide whether this gets in 5.2 or 6.0, but for me it's a candidate for 5.2. Breaking change is minimal, but benefits are high.

noordawod:
> It's still up to @wing328 to decide whether this gets in 5.2 or 6.0, but for me it's a candidate for 5.2. Breaking change is minimal, but benefits are big.

I agree. Let's wait and see what @wing328 thinks. Hoping for 5.2!

noordawod:
@agilob @wing328 tests are green! :tada:

noordawod:
ps: after implementing the offloading to via `compute()`, the app is significantly more responsive and the jank (even the slightest) is gone. I can truly say now that the app achieves 60 frames/sec with ease even with very high network load.

agilob:
Nice, what do you think about creating a service with isolates for api client? apiClient would be wrapped in an isolate rather than `Future`s? Would it be better practice?

noordawod:
I think this is beyond the scope of what the API client should be. One can always use it inside an isolate, but from my experience it's not needed at all. The heaviest operations in the client are JSON serialization/deserialization, and I think this pr would solve that bottleneck with minimal changes.

wing328:
> It's still up to @wing328 to decide whether this gets in 5.2 or 6.0, but for me it's a candidate for 5.2. Breaking change is minimal, but benefits are big.

Is there a way to fallback? via an option?

noordawod:
> > It's still up to @wing328 to decide whether this gets in 5.2 or 6.0, but for me it's a candidate for 5.2. Breaking change is minimal, but benefits are big.
> 
> Is there a way to fallback? via an option?

I could introduce 2 new methods that return `Future`s and mark the previous one deprectaed?

wing328:
> I could introduce 2 new methods that return Futures and mark the previous ones deprecated?

Sounds good. We can then include this change in the v5.2.0 release (breaking changes with fallbacks)

noordawod:
> > I could introduce 2 new methods that return Futures and mark the previous ones deprecated?
> 
> Sounds good. We can then include this change in the v5.2.0 release (breaking changes with fallbacks)

I've made the changes, what do you think?

wing328:
I saw some changes to the test. Are those changes a must in order for the test code to work? https://github.com/OpenAPITools/openapi-generator/pull/9100/files#diff-66038a916518dd193fcc27d1a54f5c8331e871c276ce8ec9443843fe261c3f17R42

In other words, does the user need to make any code change to their Dart app if upgrading to newer auto-generated Dart clients with this change?

noordawod:
> I saw some changes to the test. Are those changes a must in order for the test code to work? https://github.com/OpenAPITools/openapi-generator/pull/9100/files#diff-66038a916518dd193fcc27d1a54f5c8331e871c276ce8ec9443843fe261c3f17R42
> 
> In other words, does the user need to make any code change to their Dart app if upgrading to newer auto-generated Dart clients with this change?

Nope, no changes needed for 3rd party apps.

Although both the tests and the internal workings of the generated API client code call the new `async` methods (`serializeAsync()`, `deserializeAsync()`), these just forward to the non-async versions `serialize()` and `deserialize()`. So in effect, no changes are needed for client apps even if the non-async methods were overloaded.

However, clients that wish to use an isolate to offload serialization/deserialization, can now overload the async methods (`serializeAsync()`, `deserializeAsync()`) and accomplish that.

wing328:
Understood. Then I think this change can be included in the current master.

Let's see if the Dart technical committee has further opinion on this change.

Thanks again for the PR.

agilob:
I'm good with all these changes.

noordawod:
For the life of me, I cannot fathom why the tests are failing... I believe I touched everywhere that needs to be modified. Any help guys? @wing328 @agilob

noordawod:
> For the life of me, I cannot fathom why the tests are failing... I believe I touched everywhere that needs to be modified. Any help guys? @wing328 @agilob

Forget it, found the reason!

kuhnroyal:
I don't recommend this change as the default. There can be serious performance penalties due to the isolate message passing especially with small payloads. The general deserialization can be much slower with this change when you have multiple small requests/responses.
I would recommend adding a switch for this.

From the https://pub.dev/packages/dio_flutter_transformer docu:
> Through practical experience, we find that although using compute can make tasks go on in the background, it may lead to slow task execution. So please think carefully before using it.

kuhnroyal:
The switch could be implemented by leaving the `async` serialize/deserialize API and just switch the implementation of the `ApiClient`. Maybe even without a switch by using some sort of `SerializationHandler` interface and 2 different implementations that can be passed to the `ApiClient`. One with the current behavior (default) and one with isolate.

Then one can construct it like `ApiClient(handler: IsolateSerializationHandler())` and there would also not be a breaking change.

noordawod:
> I don't recommend this change as the default. There can be serious performance penalties due to the isolate message passing especially with small payloads. The general deserialization can be much slower with this change when you have multiple small requests/responses.
> I would recommend adding a switch for this.
> 
> From the https://pub.dev/packages/dio_flutter_transformer docu:
> 
> > Through practical experience, we find that although using compute can make tasks go on in the background, it may lead to slow task execution. So please think carefully before using it.

The pr doesn't incorporate isolates. It's just a refactor so whoever wants to use isolates, can.

noordawod:
> I don't recommend this change as the default. There can be serious performance penalties due to the isolate message passing especially with small payloads. The general deserialization can be much slower with this change when you have multiple small requests/responses.
> I would recommend adding a switch for this.
> 
> From the https://pub.dev/packages/dio_flutter_transformer docu:
> 
> > Through practical experience, we find that although using compute can make tasks go on in the background, it may lead to slow task execution. So please think carefully before using it.

From my practical experience, using `compute()` under Flutter drastically improved the UX and I see no drawbacks whatsoever. The app literally achieves native speeds with no hickups.

kuhnroyal:
I see, should have checked the changes more carefully.
How would you use isolates with the `DeserializationMessage`? I don't see it being constructed anywhere.

Instead of overloading I would probably add a handler interface with a default implementation.
Other than that I think it is a good change. Sorry for the miss understanding.

noordawod:
> I see, should have checked the changes more carefully.
> How would you use isolates with the `DeserializationMessage`? I don't see it being constructed anywhere.
> 
> Instead of overloading I would probably add a handler interface with a default implementation.
> Other than that I think it is a good change. Sorry for the miss understanding.

`DeserializationMessage` is to be used with isolates. Flutter's `compute()` accepts a single argument, but it can be anything. Hence this data class.

A snippet from my own app after overloading the new async methods:

```dart
  import 'package:generated_openapi_code' as openapi;

  @override
  Future<dynamic> deserializeAsync(String json, String targetType, {bool? growable}) =>
      compute<openapi.DeserializationMessage, dynamic>(
        openapi.deserializeAsync,
        openapi.DeserializationMessage(
          json: json,
          targetType: targetType,
          growable: growable,
        ),
      );

  @override
  Future<String> serializeAsync(Object value) => compute<Object, String>(openapi.serializeAsync, value);
```

kuhnroyal:
I see, not a fan of the overriding for a generated API, would like to see a `handler` for this but it the end it's fine with me either way. I don't use the generator.

noordawod:
@wing328 @agilob 
I have a feeling the failed test isn't related to the changes. Seems to be in the server.
Anyway, merged master now so let's see.

kuhnroyal:
The failing build should be solved by https://github.com/OpenAPITools/openapi-generator/pull/9122

noordawod:
Thanks @kuhnroyal for the note!

noordawod:
@wing328 @agilob
what to do while the CI is broken? wait until it's sane as @kuhnroyal noted?

kuhnroyal:
These are failing tests in this PR.

noordawod:
But the pr changes nothing in the tests. This cannot be right only because of the changes in the pr.
For the record, tests used to pass in the past before making the pr bacnward-compatible.

agilob:
Can you add a comment that the synchronous method is deprecated and marked for removal in 6.0?

noordawod:
> Can you add a comment that the synchronous method is deprecated and marked for removal in 6.0?

There is a deprecation already, but I'll add the note about removal in 6.0.

noordawod:
@agilob @wing328 @kuhnroyal 
Can you guys take a look at the failing test and understand why it's happening? As far as I can see, I didn't change the tests at all -- just called the async versions serialize/deserialize.

kuhnroyal:
You previous changes to the tests should have solved this and the test succeeds for me. The issue is causes by adding multiple pets to the `FakeClient` which internally now awaits the serialization but by that time the expected response has already been overwritten.  Remember the test exists twice.

Although we can probably remove one of the tests (openapi-2).
Additionally the `FakeClient` tests also don't really test anything. They will succeed even if the serialization is wrong, since the client and the expectations both use the clients serialization fucntions.

noordawod:
@wing328 @agilob what do you think?

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
```diff
@@ -147,9 +146,48 @@ class ApiClient {
 
     throw ApiException(HttpStatus.badRequest, 'Invalid HTTP operation: $method $path',);
   }
+{{#native_serialization}}
+
+  Future<dynamic> deserializeAsync(String json, String targetType, {bool growable}) async =>
+    // ignore: deprecated_member_use_from_same_package
+    deserialize(json, targetType, growable: growable);
+
+  @Deprecated('Scheduled for removal in 6.0. Use deserializeAsync() instead.')
```
wing328, line 155: minor suggestion: replace `6.0` with `OpenAPI Generator 6.x`

as users may have no idea what `6.0` refers to

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
```diff
@@ -147,9 +146,48 @@ class ApiClient {
...
+    return targetType == 'String'
+      ? json
+      : _deserialize(jsonDecode(json), targetType, growable: growable == true);
+  }
+{{/native_serialization}}
+
+  // ignore: deprecated_member_use_from_same_package
+  Future<String> serializeAsync(Object value) async => serialize(value);
+
+  @Deprecated('Scheduled for removal in 6.0. Use serializeAsync() instead.')
```
wing328, line 170: minor suggestion: replace `6.0` with `OpenAPI Generator 6.x`

as users may have no idea what `6.0` refers to

noordawod: Most definitely, coming up.

kuhnroyal:
> Although we can probably remove one of the tests (openapi-2).
> Additionally the `FakeClient` tests also don't really test anything. They will succeed even if the serialization is wrong, since the client and the expectations both use the clients serialization fucntions.

This should be done separately, for now you should just fix both tests in the same way you did earlier.
`samples/openapi3/client/petstore/dart2/petstore/test/pet_faked_client_test.dart`
`samples/client/petstore/dart2/petstore/test/pet_faked_client_test.dart`

noordawod:
> > Although we can probably remove one of the tests (openapi-2).
> > Additionally the `FakeClient` tests also don't really test anything. They will succeed even if the serialization is wrong, since the client and the expectations both use the clients serialization fucntions.
> 
> This should be done separately, for now you should just fix both tests in the same way you did earlier.
> `samples/openapi3/client/petstore/dart2/petstore/test/pet_faked_client_test.dart`
> `samples/client/petstore/dart2/petstore/test/pet_faked_client_test.dart`

I rolled back my "fix", but can certainly bring it back. Do you think it'll solve the tests?

kuhnroyal (approved):

noordawod:
woo hoo! that was an awesome input, thank @kuhnroyal.

noordawod:
What's left for merging? :)

wing328:
@noordawod just merged. Thanks for the PR.

@agilob @kuhnroyal thanks for the review.

vladaman:
@noordawod any chance to add support for null-safety and dart >= 2.12 ? Thanks!

noordawod:
@vladaman I talked with @agilob a while ago about adding null-safety but I just don't have time right now. But definitely in the planning.

PR #9109, opened 2021-03-27
[csharp][netcore-httpclient] Reuse HttpClient, Allow use of external HttpClient. Patch to previous PR.

lucamazzanti:
Fixed an existing issue that made the HttpClient open new sockets for each connection. 
Also makes HttpClient settable from external.
Fixes: #9060
Patches: #9085

Updated constructors to be more explicit in the usage.
Updated documentation to underline the use cases and the issue.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@mandrean (2017/08) @frankyjuang (2019/09) @shibayan (2020/02) @Blackclaws (2021/03)

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/README.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/README.mustache
@@ -55,7 +55,8 @@ Install-Package CompareNETObjects
 ```
 {{#useRestSharp}}
 
-NOTE: RestSharp versions greater than 105.1.0 have a bug which causes file uploads to fail. See [RestSharp#742](https://github.com/restsharp/RestSharp/issues/742)
+NOTE: RestSharp versions greater than 105.1.0 have a bug which causes file uploads to fail. See [RestSharp#742](https://github.com/restsharp/RestSharp/issues/742).
+NOTE: RestSharp for .Net Core creates a new socket for each api call, which can lead to a socket exhaustion problem. See [RestSharp#1406](https://github.com/restsharp/RestSharp/issues/1406).
 
 {{/useRestSharp}}
 <a name="installation"></a>
@@ -102,7 +103,10 @@ c.Proxy = webProxy;
 ```
 {{#useHttpClient}}
 
-To use your own HttpClient instances just pass them to the ApiClass constructor.
+### Connections
+Each ApiClass (properly the ApiClient inside it) will create an istance of HttpClient. It will use that for the entire lifecycle and dispose it when called the Dispose method.
+
+To better manager the connections it's a common practice to reuse the HttpClient and HttpClientHander (see [here](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net) for details). To use your own HttpClient instance just pass it to the ApiClass constructor.
 
 ```csharp
 HttpClientHandler yourHandler = new HttpClientHandler();
@@ -110,17 +114,20 @@ HttpClient yourHttpClient = new HttpClient(yourHandler);
 var api = new YourApiClass(yourHttpClient, yourHandler);
 ```
 
-If you want to use an HttpClient and don't have access to the handler, for example in a DI context in aspnetcore when
-using IHttpClientFactory. You need to disable the features that require handler access:
+If you want to use an HttpClient and don't have access to the handler, for example in a DI context in Asp.net Core when using IHttpClientFactory. 
 
 ```csharp
 HttpClient yourHttpClient = new HttpClient();
-var api = new YourApiClass(yourHttpClient, null, true);
+var api = new YourApiClass(yourHttpClient);
 ```
+You'll loose some configuration settings, the features affected are: Setting and Retrieving Cookies, Client Certificates, Proxy settings. You need to either manually handle those in your setup of the HttpClient or they won't be available.
 
-The features affected are: Setting and Retrieving Cookies, Client Certificates, Proxy settings.
-You need to either manually handle those in your setup of the HttpClient or they won't be available.
+Here an example of DI setup in a sample web project:
 
+```csharp
+services.AddHttpClient<YourApiClass>(httpClient => 
+   new PetApi(httpClient));
+```
 
 {{/useHttpClient}}
 
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
@@ -421,10 +421,10 @@ namespace {{packageName}}.Client
             return request;
         }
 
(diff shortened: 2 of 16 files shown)


Blackclaws:
Looked over it real quick, looks fine from a code standpoint on first glance. We should discuss whether we really want to just assume that having these features missing is going to be noticed by users and having the info in the readme itself is going to be fine. 

I'm not even sure how many people know about the handler in the first place. Though I do get it makes for much easier usage and maybe proxy, cookies and client certificates are special use cases anyway.

lucamazzanti:
> Looked over it real quick, looks fine from a code standpoint on first glance. We should discuss whether we really want to just assume that having these features missing is going to be noticed by users and having the info in the readme itself is going to be fine.

You are right, I removed the disable feature parameter because it was 1:1 related with the handler, but I understood your idea to ask the permission to disable some features.

> I'm not even sure how many people know about the handler in the first place. Though I do get it makes for much easier usage and maybe proxy, cookies and client certificates are special use cases anyway.

Seeing that nobody has view the issue I could say not a lot of people. An issue that you can see in a real production scenario, nothing you can easily test here. Thinking about my usage in the Net Core, the base documentation is clear and cover a few but detailed points, one is the DI and its features. One of that is the HttpClient and its handlers lifcecyle, plus a resilient integration with Polly for retries, caching. The handlers are pooled, and you recieve a configured, working HttpClient, not the handler.
So I think that someone that embrace the current features will not use them, just the client.
If you don't relies on the IHttpClientFactory you need to do some not so simple work: a thread safe pool with time evictions.
So probably the handler will be used only for this remaining features, not a basic scenario.

> It could throw exceptions when those features are used and no handler is set. That would be another way to do things.

Its a good idea, they depends a lot on the configuration so probably they will not comes out easily at coding time, but in some customer settings during the production setup. I still prefer to have a clear error probaly than a silent behavior.

Blackclaws:
> Think about to remove the handler\clients from local variables and from method signature and use them as class fields as they are now.

Yeah, do that change, its a leftover from when they were created locally.

Blackclaws:
> Its a good idea, they depends a lot on the configuration so probably they will not comes out easily at coding time, but in some customer settings during the production setup. I still prefer to have a clear error probaly than a silent behavior.

I'd say do that then. Introduce an exception with "Operation not supported" or something like that, including a useful error message. That of course doesn't work for the extraction of cookies in the ToApiResponse part, but I think we can leave that one.

lucamazzanti:
> > Its a good idea, they depends a lot on the configuration so probably they will not comes out easily at coding time, but in some customer settings during the production setup. I still prefer to have a clear error probaly than a silent behavior.
> 
> I'd say do that then. Introduce an exception with "Operation not supported" or something like that, including a useful error message. That of course doesn't work for the extraction of cookies in the ToApiResponse part, but I think we can leave that one.

lucamazzanti:
S

lucamazzanti:
I don't know maybe the check can be made at constructor time when you have both handler valued and the configuration with something not supported. I see the configuration is passed within with a readonly interface, so probably it can change between calls from a change outside the client, in the global configuration.

Blackclaws:
I think InvalidOperationException or NotSupportedException both kind of fit. In general yeah I think the csharp surface needs a bit of a rework, as its used by the generated code I think the generation never is used after the initial client setup. So maybe I'll rework that afterwards as well.

Blackclaws:
Looks good to me. I like having exceptions in just these specific cases where it actually matters. So good to be merged from my points of view.
Thanks for the pull request!

wing328:
Circle CI failure (not related to this change) already fixed in the master

wing328:
@lucamazzanti when you've time, can you please PM me via https://join.slack.com/t/openapi-generator/shared_invite/enQtNzAyNDMyOTU0OTE1LTY5ZDBiNDI5NzI5ZjQ1Y2E5OWVjMjZkYzY1ZGM2MWQ4YWFjMzcyNDY5MGI4NjQxNDBiMTlmZTc5NjY2ZTQ5MGM to discuss a little bit more about your use case? Thank you.

lucamazzanti:
Thanks you all for the project and the support

lucamazzanti:
> @lucamazzanti when you've time, can you please PM me via https://join.slack.com/t/openapi-generator/shared_invite/enQtNzAyNDMyOTU0OTE1LTY5ZDBiNDI5NzI5ZjQ1Y2E5OWVjMjZkYzY1ZGM2MWQ4YWFjMzcyNDY5MGI4NjQxNDBiMTlmZTc5NjY2ZTQ5MGM to discuss a little bit more about your use case? Thank you.

Done

lucamazzanti:
So this merge will be released in milestone 5.1.1, I'll test it and move in production after the release.

Blackclaws:
> So this merge will be released in milestone 5.1.1, I'll test it and move in production after the release.

I might be doing another pull request to clean up the huge amount of constructors we have right now. I think we can boil those down a bit.

lucamazzanti:
I understand there are so much constructors, it was just to manage better the parameters and intellisense in code completion hints.

lucamazzanti:
I have a question for the multiple ctor, if I can make good my work.
They are too much, so I want to open a PR fix to try to solve it.
We have in ApiClient those constructors: 
- empty
- basePath

I changed the previous code trying to manage better the new parameters, but this put a multiplier on ctors:
- empty
- basePath
- httpClient
- httpClient, basePath
- httpClient, httpClientHandler
- httpClient, httpClientHandler, basePath

This move to more ctors in the ApiClass generated where we moved from:
- empty
- basePath
- configuration

to:
- empty
- basePath
- configuration
- httpClient
- httpClient, basePath
- httpClient, configuration
- httpClient, httpClientHandler
- httpClient, httpClientHandler, basePath
- httpClient,httpClientHandler, configuration

The position of the services before the configurations is preferred, also in other point of the project, but here it blocks from insert a default null value on the httpClientHandler and compress some ctors.
Another point for that choice was the documentation in the intellisense and the lack of optional services in the sign.

But yes for sure they are too much.
So probabily it was better before, I can do a PR to fix that and return to @Blackclaws version.

So the sign for ApiClient will be:
- httpClient = null, httpClientHandler = null
- basePath, httpClient = null, httpClientHandler = null
with the precondition to avoid an httpClient null when passed an handler

and the ApiClass will return to:
- httpClient = null, httpClientHandler = null
- basePath, httpClient = null, httpClientHandler = null
- configuration, httpClient = null, httpClientHandler = null

**Or** we can keep the client as first parameter and move the optional handler as third:
- httpClient, httpClientHandler = null
- httpClient, basePath, httpClientHandler = null
And in ApiClass:
- empty
- basePath
- configuration
- httpClient, httpClientHandler = null
- httpClient, basePath, httpClientHandler = null
- httpClient, configuration, httpClientHandler = null
so here we mantain the httpClient as first parameter when available and mandatory, then the mandatory configurations, and least the optional handler.

lucamazzanti:
Created PR #9145 to reduce the number of constructors.

PR #9166, opened 2021-04-04
Add Hashable for Struct

### LINKED ISSUE #9030: [REQ] [Swift5] Hashable struct models

I'm always frustrated because the generated struct models aren't Hashable.
That happens often because DiffableDatasource and SwiftUI.List require a Hashable model.
Originally, we just had to define Hashable in the original type declaration of the struct models, but now we have to make the following choices:

- Customize the Template
- Manually add a lot of x-swift-hashable to swagger.yml
- Write a lot of boilerplate code with Extension (because it will be a separate file from the definition)

Always make all struct models Hashable like Codable. This doesn't cause any problems and it's a very simple solution.
In addition, x-swift-hashable is no longer needed, reducing code.
The work of the Swift compiler and the code generated by the Swift compiler will be a lot, but it will be negligible.

Add a flag to additional-properties. If we need a Hashable compliant model, specify it.

maoyama:
Resolves #9030

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/swift5/modelObject.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/modelObject.mustache
@@ -1,4 +1,4 @@
-{{^objcCompatible}}{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} struct {{classname}}: Codable{{#vendorExtensions.x-swift-hashable}}, Hashable{{/vendorExtensions.x-swift-hashable}} {
+{{^objcCompatible}}{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} struct {{classname}}: Codable, Hashable {
 {{/objcCompatible}}{{#objcCompatible}}@objc {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} class {{classname}}: NSObject, Codable {
 {{/objcCompatible}}
 
--- a/modules/openapi-generator/src/test/resources/2_0/swift/petstore-with-fake-endpoints-models-for-testing.yaml
+++ b/modules/openapi-generator/src/test/resources/2_0/swift/petstore-with-fake-endpoints-models-for-testing.yaml
@@ -1086,7 +1086,6 @@ definitions:
     xml:
       name: Order
   Category:
-    x-swift-hashable: true
     type: object
     required:
       - name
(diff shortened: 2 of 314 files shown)


wing328:
@maoyama thanks for the PR. What about adding an option (e.g. `hashableStruct`) to add `x-swift-hashable` to all models in postProcessModels if x-swift-hashable is not yet defined in the model?

Ref: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/CodegenConstants.java#L133

maoyama:
@wing328 I didn't want to add more options and make it more complicated for the user. I like the simple specs by this PR.

wing328:
Understood. Let's see if the Swift technical committee has a different option on this as I do not know if everyone likes hashable struct.

wing328:
Another use case worth considering is that a user wants all models except a few models to extends hashable. An option I described above can achieve that.

4brunu:
I don't see any downside in making all the models conform to `Hashable`.
For me we can go with all models conform to `Hashable` and if an issue appears to someone we can introduce a flag to disable some cases.

wing328:
OK. Let's go with this and add an option if needed later.

fl034:
Hey guys @4brunu . I've experienced a downside of this feature. I have many custom implementations of Hashable that seem to be ignored now, since the model has a synthesized implementation.
In my custom implementations I exclude for example timestamps like "updatedAt".

So for me it will be crucial to be able to control this feature. For me it would be best to be able to control this at model-level. Like with `x-swift-hashable` or `x-swift-disable-hashable`.

We could also have a configuration flag to set the default behavior.

4brunu:
A configuration flag to control the behaviour seems a good option to me.
@fl034 could you please open a PR with this?

fl034:
Sorry @4brunu I don't have resources for it this time. My workaround was to downgrade to 5.1.0.
I added an issue to track it.

maoyama:
@fl034 @4brunu 
> I have many custom implementations of Hashable that seem to be ignored now, since the model has a synthesized implementation.

I want to know more about this custom implementation of Hashable.
I think that we can exclude the specified property by custom implementation of Hashable like below:
```swift
// Custom implementations of Hashable
extension Animal {
    public func hash(into hasher: inout Hasher) {
        hasher.combine(className)
    }
}

// Generated Model
public struct Animal: Codable, Hashable {
    public var className: String
    public var color: String? = "red"

    public init(className: String, color: String? = "red") {
        self.className = className
        self.color = color
    }
}
```
 If this feature has a downside, we need an option to control its behavior.

fl034:
@maoyama I suppose when the extension is in the same file, the custom implementation is used. But this is not applicable for generated code files, of course.

In my case I even get a compiler warning, that my implementations will be ignored.

Bildschirmfoto 2021-05-12 um 11 54 11

maoyama:
@fl034 Thank you for your reply.
I understand that we need an option to control its behavior.

Here are the results of the custom implementation I've tried:
- Hashable can be custom implementation in the same module, not limited to the defined file.
- Hashable cannot be custom implementation with different modules. (Equatable is possible)

I can open a PR with adding options.

4brunu:
@maoyama a PR is welcome 👍

Issue #8255, opened 2020-12-22
[BUG] Gradle plugin doesn't support remote locations for inputSpec


joschi:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?

##### Description

The OpenAPI Generator Gradle Plugin doesn't support remote locations (i. e. HTTPS URLs) for the input specification (`inputSpec`) anymore.

This is a breaking change between OpenAPI Generator Gradle Plugin 4.3.1 and 5.0.0 and thus should either be listed in the release notes or fixed in a patch release.

Repository to illustrate and reproduce the issue:
https://github.com/joschi/openapi-generator-gradle-remote-input

The following setup was working with the OpenAPI Generator Gradle Plugin 4.3.1:

<details>
<summary><a href="https://github.com/joschi/openapi-generator-gradle-remote-input/blob/fe51395acffdcee1a71c7357673b86aaddb7e69f/build.gradle
">build.gradle with OpenAPI Generator Gradle Plugin 4.3.1</a></summary>

```groovy
plugins {
  id "org.openapi.generator" version "4.3.1"
}

repositories {
  mavenCentral()
}

openApiGenerate {
  inputSpec = "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml"
  outputDir = "$projectDir/client"
  configOptions = [
      java8                  : "true",
      dateLibrary            : "java8",
      serializationLibrary   : "jackson",
      useRuntimeException    : "true"
  ]
  generatorName = "java"
  library = "rest-assured"
  groupId = "com.example"
  id = "client"
  version = "1.0.0-SNAPSHOT"
}
```
</details>
<details>
<summary>Build output</summary>

```
./gradlew --stacktrace openApiGenerate

> Task :openApiGenerate
Output directory does not exist, or is inaccessible. No file (.openapi-generator-ignore) will be evaluated.
4.3.0 has deprecated the use of vendor extensions which don't follow lower-kebab casing standards with x- prefix.
Successfully generated code to /Users/joschi/Projects/openapi-generator-gradle-remote-input/client

[...]

PR #9192, opened 2021-04-06
[Kotlin] Issue #8255 fix (inputSpec param not accepts URL)

Orachigami:
Fixes: #8255
- Replaced `@InputFile` and `@PathSensitive` annotations with `@Optional` and `@Input` annotations for the `inputSpec` parameter
- Added test that checks project generation with a remote spec

### PR checklist
 
- [x] Pull Request title clearly describes the work in the pull request and Pull Request description provides details about how to validate the work. Missing information here may result in a delayed response from the community.
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt
```diff
@@ -92,8 +89,8 @@ open class GenerateTask : DefaultTask() {
     /**
      * The Open API 2.0/3.x specification location.
      */
-    @get:InputFile
-    @PathSensitive(PathSensitivity.RELATIVE)
+    @Optional
+    @Input
```
joschi, line 93: This would unfortunately mean that even if the input specification in a local file changed, the task would be up-to-date because it's only checking the filename but not its contents.

Maybe splitting the input specification location into local specs and remote specs would make sense?

Orachigami: @joschi, if I extend this PR with the solution to check the modification status of the remote source will it do the trick? I mean let's imagine that the Up-to-date check works with the Url/Path then will you agree to use specificationUrl?

Issue #9093, opened 2021-03-26
[Announcement] New generator for Go Echo


ph4r5h4d:
I'm working on a new code generator for the GO Echo microframework.
It's part of what we are using in our teams, and I was thinking let's contribute it to the main project.
It's almost finished, and I'll create a PR in a couple of days.

I've tried to see if I'd be able to get a good result by just modifying the GO Gin generator's templates, but it was not the case, and I wanted it a little bit opinionated. I'll share the codes in a bit, and maybe if you have any idea on how I can improve the codes and templates, you can share so I can apply them to the project.

P.S: I was not sure if it should be an announcement or anything else. Maybe a moderator can help on this part.

PR #9224, opened 2021-04-09
Add Go echo server codegen

ph4r5h4d:
This is the first draft for go-echo-server codegen for OpenAPI-codegen. the purpose is that we are internally using Go Echo, and we needed to create this codegen. Still, I thought that it might be worth sharing it with everyone.  
Related issue: https://github.com/OpenAPITools/openapi-generator/issues/9093

I'm not a Java guru, and I mostly used go-gin-codegen codes to build this new codegen, and it's a bit opinionated.

Kindly let me know what needs to be done so we can merge this PR.

@antihax  @grokify  @kemokemo  @jirikuncar 
Kindly let me know what needs to be improved.


kemokemo:
@ph4r5h4d
Thank you so much for your PR! 😄  It looks nice.
I'm very happy about this PR because I also used [echo](https://github.com/labstack/echo) for my company's internal web service and like it.  
I like [echo](https://github.com/labstack/echo) as well as [gin](https://github.com/gin-gonic/gin).

I have confirmed that the generated `go-echo-server`'s code builds, runs, and works as a container.
It looks good, but there are a few things I would like to discuss. Can you please see the following?

## `golint ./...` reports some issues

I feel that it would be better to add comments for methods exported in the package, so that issues reported by the golint tool can be resolved. What do you think?

## docker run

This may be a problem only in my macOS environment. I had to add the `-p` flag when running the docker container.  
How about in your environment?

- (I can not access to this app) `docker run --rm -it openapi`
- (I can access to this app) `docker run --rm -p 8080:8080 -it openapi`

ph4r5h4d:
I'll take a look into the items you've mentioned and will issue an update by the end of the week.
Thank you for putting time into testing this PR.

ph4r5h4d:
Can you kindly check again and see if everything is fine?

wing328:
Overall it looks good to start with 👍 

I'll merge this and file another PR with minor enhancements.

Thanks for the new generator!

Issue #9148, opened 2021-04-01
[BUG] [cpp-pistache-server] Generated cmake project does not build


JulianGmp:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

A recent [pull request](https://github.com/pistacheio/pistache/pull/870) in the pistache repository made changes to their build system, in which the `install` target was removed from the `CMakeLists.txt` (see [diff](https://github.com/pistacheio/pistache/commit/9dc080b9ebbe6fc1726b45e9db1550305938313e#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL107) ).

That target is used by the pistache templates, and thus building the server will currently fail.
Tested on current master (7bf792a3485b967a2ae74f6b3bb3a337ba47e1bb).

Build commands used
```sh
cmake -S . -B build/
cmake --build build/ -j 1 --target all
```

Last bits of the output:
```
[ 16%] Performing build step for 'PISTACHE'
[ 91%] Built target pistache
[ 95%] Built target pistache_static
[100%] Built target pistache_shared
[ 20%] Performing install step for 'PISTACHE'
gmake[3]: *** No rule to make target 'install'.  Stop.
gmake[2]: *** [CMakeFiles/PISTACHE.dir/build.make:93: PISTACHE-prefix/src/PISTACHE-stamp/PISTACHE-install] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:154: CMakeFiles/PISTACHE.dir/all] Error 2
gmake: *** [Makefile:103: all] Error 2
```

##### openapi-generator version

openapi-generator-cli 5.1.1-SNAPSHOT
commit : 7bf792a
built : 2021-04-01T11:56:22+02:00

##### OpenAPI declaration file content or url

[...]

PR #9251, opened 2021-04-13
[C++] [Pistache] Model validation, general overhaul

JulianGmp:
This PR includes a variety of smaller changes, as well as implementing model validation for the Pistache C++ server.

List of changes:
- Implemented validation of model data through mustache templates, this validation includes...
  - enum values (both “inline” enums and enums as their own schema)
  - `date` and `date-time` formatted strings as per `full-date`/`date-time` in RFC3339
  - string length (min and max) using `std::string::length`
  - numeric validations for
    - min value (exclusive and non-exclusive)
    - max value (exclusive and non-exclusive)
    - multipleOf (Note this may cause issues for float/double, see https://swagger.io/docs/specification/data-models/data-types/ section “Multiples”)
  - Array length (min and max)
  - Unique array elements
  - **Recursive validation for array elements**, this includes all checks mentioned above

- Generate arrays with `uniqueItems: true` as `std::vector`s. The current implementation used a type called `Set` which did not exist and caused build issues. The standard vector was chosen above a set to ensure the order of items remains intact (see https://github.com/OpenAPITools/openapi-generator/pull/5466#issuecomment-593305650);

- Implement the `==` operator (and `!=`) for model classes. This is needed for checking if an array only has unique elements.

- Change get functions for array type. They now just return a vector, instead of a const reference to the vector. This copies the underlying vector, which is expensive. My justification is: returning a const reference may still allow the caller to modify the individual elements of the vector, thus the get function may not be declared as `const`. This currently causes the issue that you have to use a const cast in your API impl, since the model is always passed by const reference. By copying the vector the get function can safely be declared as const, and the original data cannot be modified.

- Create virtual methods `handleParsingException` and `handleOperationException` for API classes. These will map a thrown exception to an HTTP status code and a response. The intended use is to overload these if needed, to ensure a certain return structure for error cases that are handled in the API class (such as a json parsing error).

- Made the API classes' `base` property `static const`. That way you can read the base path without having to create an instance of the API.

- Smaller visual changes, for example one line namespace declarations.

- Minor cleanups, for example passing an `int32_t` by value, not by const reference.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

*C++ technical committee*: @ravinikam @stkrwork @etherealjoy @martindelille @muttleyxd


modules/openapi-generator/src/main/resources/cpp-pistache-server/api-header.mustache
```diff
@@ -14,24 +14,21 @@
...
 {{#imports}}{{{import}}}
 {{/imports}}
 
-{{#apiNamespaceDeclarations}}
-namespace {{this}} {
-{{/apiNamespaceDeclarations}}
-
-{{#hasModelImport}}
-using namespace {{modelNamespace}};{{/hasModelImport}}
+namespace {{apiNamespace}}
```
muttleyxd, line 22: This will only work with C++17 or newer, while Pistache is guaranteed to work with C++14, so I'd return to older syntax. This comment applies to all namespace definition changes.

modules/openapi-generator/src/main/resources/cpp-pistache-server/api-impl-header.mustache
```diff
@@ -22,17 +22,16 @@
...
 
 {{#hasModelImport}}
 using namespace {{modelNamespace}};{{/hasModelImport}}
 
 class {{classname}}Impl : public {{apiNamespace}}::{{classname}} {
 public:
-    {{classname}}Impl(std::shared_ptr<Pistache::Rest::Router>);
-    ~{{classname}}Impl() {}
+    explicit {{classname}}Impl(const std::shared_ptr<Pistache::Rest::Router>& rtr);
+    ~{{classname}}Impl() override;
```
muttleyxd, line 34: We can remove destructor body from .cpp file, if we define destructor as:
```
~{{classname}}Impl() override = default;
```

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-validation-body.mustache
```diff
@@ -0,0 +1,125 @@
+{{!
```
muttleyxd, line 1: I didn't review this yet, I'll do it once samples have been generated.

muttleyxd (changes requested):
Good changes in general, some minor things to fix.
One missing thing are updated samples - you should run
`./bin/generate-samples.sh bin/configs/other/cpp-pistache-server-cpp-pistache.yaml` in root directory of openapi-generator - it will update cpp-pistache-server samples. When it's done, I'll investigate validation a bit more.

modules/openapi-generator/src/main/resources/cpp-pistache-server/api-header.mustache
muttleyxd: Pistache merged switch C++17 2 hours after my comment, so you can ignore this comment.

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-validation-body.mustache
JulianGmp: Samples have been generated.
On that note, running only `./bin/generate-samples.sh` as the PR template states didn't generate the pistache samples, only running
`./bin/generate-samples.sh bin/configs/other/cpp-pistache-server-cpp-pistache.yaml`
seems to have worked.

I'm not sure if this is intentional for the script?

JulianGmp: Also, for checking the validation code, I recommend running the code through a formatter first, as the indentation is an absolute mess

wing328: Please move `bin/configs/other/cpp-pistache-server-cpp-pistache.yaml` to `bin/configs/cpp-pistache-server-cpp-pistache.yaml` as part of this PR.

wing328: > I recommend running the code through a formatter first, as the indentation is an absolute mess

Yup, that's generally what we recommend the users to do after code generation.

JulianGmp: File moved in ec0bf98

wing328:
If no further feedback on this, I'll merge it later this week:

cc @ravinikam @stkrwork @etherealjoy @MartinDelille @muttleyxd

etherealjoy:
> If no further feedback on this, I'll merge it later this week:
> 
> cc @ravinikam @stkrwork @etherealjoy @MartinDelille @muttleyxd

@wing328 I will check it within next days, sorry I was a bit busy

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
```diff
@@ -18,13 +18,60 @@ namespace {{this}} {
...
+        msg << _pathPrefix << ": has no value;";
+    }
+    {{/isEnum}}
+    {{^isEnum}}
+    {{#vars}}
+    {{^required}}if ({{nameInCamelCase}}IsSet()){{/required}}
+    {{#required}}/* {{name}} */ {{/required}}{
+        const {{{dataType}}}& value = m_{{name}};
+        const std::string currentValuePath = _pathPrefix + ".{{nameInCamelCase}}";
+        {{> model-validation-body }}
```
muttleyxd, line 53: I wonder if it's possible to detect here if any content has been generated by model-validation-body. Only thing I don't like here is that it generates a lot of if branches for things, that can't be validated. 

@wing328 maybe you know a way to do it? I tried experimenting with hasValidation property, but it's always false for me.

etherealjoy:
LGTM
@JulianGmp 
Seems like a major improvement. Thanks for the effort.
Two points I want to ask
- For the struct type there is no validation? 
- For the equality operator, what about containers, like arrays and maps?

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
wing328: One way is to do it in the generator's (java) model post-processing function to add something like `x-cpp-has-validation` to indicate whether there's any validation

JulianGmp: There is a field called `has-validation`, I'm not sure if we can use it in this particular context, I will experiment with it.

JulianGmp:
> LGTM
> @JulianGmp
> Seems like a major improvement. Thanks for the effort.
> Two points I want to ask
> 
>     * For the struct type there is no validation?
> 
>     * For the equality operator, what about containers, like arrays and maps?

Sorry for the delayed reply. To address your points:

As of now, there isn't a validation for the struct types. Frankly I forgot they are an option for the generator, but it should not be much effort to add the validation code to the structs too.

The C++ standard library defines the `==` operator for std::vector and std::map as:
> Checks if the contents of lhs and rhs are equal, that is, they have the same number of elements and each element in lhs compares equal with the element in rhs at the same position.

So this, as to my understanding, this compares individual elements of the container, in regards to order. I tested this for the `std::map` and the "position" of elements is based on the key, not the order in which they were added in (`g++ (GCC) 10.2.0`).

Sources:
 - [operator==,!=,<,<=,>,>=,<=>(std::map)](https://en.cppreference.com/w/cpp/container/map/operator_cmp)
 - [operator==,!=,<,<=,>,>=,<=>(std::vector)](https://en.cppreference.com/w/cpp/container/vector/operator_cmp)

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
JulianGmp: Alright I tested the `hasValidation` a little.
In principle it works, I can check the value and only generate the `const {{{dataType}}}& value....` if it is true.
However, this will break for arrays. Given the following example:
```yaml
myOtherArray:
  type: array
  items:
    type: integer
    format: int64
    minimum: 10
    maximum: 100
```
`hasValidation` remains false for `myOtherArray`, since the array itself does not have validation, but the items inside of it do.
I don't think there's a way to recursively check if the items of an array (which of course can be arrays too) have some sort of validation, without modifying the java code for the openapi-generator. I would try to do that, but it's a fairly large codebase and I don't know my way around it overall.

But I can use this for all other checks, so the unnecessary checks are only generated for array types.

JulianGmp:
In regards to implementing this for struct models too.

I have noticed that struct models use `Pistache::optional` for values that aren't required.
This means that the validation body has to be slightly different from the one for classes, but I had two other questions:

1. why do the struct models use an optional type, and the classes use their own “is set” flags. I get that changing this now would be an insane breaking change, but I'm kinda curious.

2. Pistache have recently removed their `Pistache::optional` and opted for `std::optional` during their switch to C++17 (see https://github.com/pistacheio/pistache/commit/86ef2a43e1b8b05ecafe45634390963989a09b57). So this has to be changed as well.

I'm fairly busy right now, but I'll try to find some time for this, unless you consider opening a seperate issue for the structs.

On another note, #9148 still prevents building the cpp server, since the cmake target always pulls the latest master from pistache.

etherealjoy:
> I'm fairly busy right now, but I'll try to find some time for this, unless you consider opening a seperate issue for the structs.

We can handle it with a different issue for structs.

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
etherealjoy: We can improve this in the future, You can list the limitations in the Description

muttleyxd (approved):

wing328:
Shippable failure not related to this change.

JulianGmp:
> > I'm fairly busy right now, but I'll try to find some time for this, unless you consider opening a seperate issue for the structs.
> 
> We can handle it with a different issue for structs.
> 
> We will have to adapt for the optional eventually, but as a separate issue

Thank you for merging my PR.
I created according issues for the things mentioned: #9538 and #9540.

I will also update this PRs description to document the limitations to this feature as described in https://github.com/OpenAPITools/openapi-generator/pull/9251#discussion_r624503266.

PR #9283, opened 2021-04-16
Qt5 support date format enum

### LINKED ISSUE #9279: [REQ] [Qt5] [C++] Allow for dateTime format to be ISODateWithMs (not just ISODate)

I would like to use ISODateWithMs and right now the only way to do that is to call: 
```
SerializerSettings
bool setDateTimeFormat(const QString& dateTimeFormat);
```

and hope that you can supply a QString that meets your needs.

Here is the current code for serializing datetimes
```
QString toStringValue(const QDateTime &value) {
    // ISO 8601
    return SerializerSettings::getInstance()->getDateTimeFormat().isEmpty()? value.toString(Qt::ISODate):value.toString(SerializerSettings::getInstance()->getDateTimeFormat());
}
```

I would like to also have the option of setting Qt::ISODateWithMs to the serializer. trying to do it manually with the string is difficult due to timezones.

I can do a few things:

1. add a boolean to SerializerSettings called useMsInDateTime, which then switches between Qt::ISODate and Qt::ISODateWithMs
2. add a setter for setDateTimeFormat overloaded to accept a Qt::DateFormat and then adjust the logic in toStringValue to also look for if this was set and use it if so.

"yyyy-MM-ddTHH:mm:ss.zzzZ" "yyyy-MM-ddTHH:mm:ss.zzz[Z|[+|-]HH:mm]" etc all have issues with timezones when passed in as just a string format. So the ability to pass the enum as well would fix this

xconverge:
fixes #9279

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

cc @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @MartinDelille (2018/03) @muttleyxd (2019/08)

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
@@ -8,9 +8,9 @@ namespace {{this}} {
 {{/cppNamespaceDeclarations}}
 
 struct {{prefix}}CustomDateTimeFormat{
-    bool setString = false;
+    bool isStringSet = false;
     QString formatString;
-    bool setEnum = false;
+    bool isEnumSet = false;
     Qt::DateFormat formatEnum;
 };
 
@@ -21,12 +21,14 @@ public:
     }
 
     static void setDateTimeFormatString(const QString &dtFormat){
-        getInstance()->customDateTimeFormat.setString = true;
+        getInstance()->customDateTimeFormat.isStringSet = true;
+        getInstance()->customDateTimeFormat.isEnumSet = false;
         getInstance()->customDateTimeFormat.formatString = dtFormat;
     }
 
     static void setDateTimeFormatEnum(const Qt::DateFormat &dtFormat){
-        getInstance()->customDateTimeFormat.setEnum = true;
+        getInstance()->customDateTimeFormat.isEnumSet = true;
+        getInstance()->customDateTimeFormat.isStringSet = false;
         getInstance()->customDateTimeFormat.formatEnum = dtFormat;
     }
 
@@ -39,8 +41,8 @@ public:
 private:
     explicit {{prefix}}SerializerSettings(){
         instance = this;
-        customDateTimeFormat.setString = false;
-        customDateTimeFormat.setEnum = false;
+        customDateTimeFormat.isStringSet = false;
+        customDateTimeFormat.isEnumSet = false;
     }
     static {{prefix}}SerializerSettings *instance;
     {{prefix}}CustomDateTimeFormat customDateTimeFormat;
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
```diff
@@ -7,14 +7,29 @@
...
-    static void setDateTimeFormat(const QString & dtFormat){
-        getInstance()->dateTimeFormat = dtFormat;
+    static {{prefix}}CustomDateTimeFormat getCustomDateTimeFormat() {
+        return getInstance()->customDateTimeFormat;
     }
-    static QString getDateTimeFormat() {
-        return getInstance()->dateTimeFormat;
+
+    static void setDateTimeFormatString(const QString &dtFormat){
+        getInstance()->customDateTimeFormat.setString = true;
```
etherealjoy, line 24: Here maybe you can set the other flag as false, otherwise the check in the other API will have no effect.

modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
```diff
@@ -7,14 +7,29 @@
 namespace {{this}} {
 {{/cppNamespaceDeclarations}}
 
+struct {{prefix}}CustomDateTimeFormat{
+    bool setString = false;
```
etherealjoy, line 11: Since it is a boolean, you could use `isStringSet` and same for other flag below.

etherealjoy (changes requested):

modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
xconverge: done

modules/openapi-generator/src/main/resources/cpp-qt5-client/helpers-body.mustache
xconverge: yep sounds good, done

etherealjoy (approved):

xconverge:
thanks for the prompt review!

samples/client/petstore/cpp-qt5/client/PFXHelpers.h
```diff
@@ -50,7 +50,8 @@ class OptionalParam {
     }
 };
 
-bool setDateTimeFormat(const QString&);
+bool setDateTimeFormat(const QString &format);
```
wing328, line 53: Is this a breaking change as there are changes in the method signature?

etherealjoy: same signature, just the parameter is not specified earlier and it is perfectly fine c++ thing

xconverge: yep, was just done for consistency within the file

xconverge: also can see that it was not a breaking change because this file did not need to be changed to compile https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/cpp-qt5/PetStore/main.cpp interesting enough, you can see that is trying to do a "similar" thing to this in a more roundabout way than just assigning IsoDataMs

wing328: 👌

PR #9392, opened 2021-05-03
upgrade dart generator dependencies

Adi1231234:
upgrade dart generator dependencies

### DIFF
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -15,9 +15,9 @@ dependencies:
   intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
-  json_annotation: '^4.0.1'{{/json_serializable}}
+  json_annotation: '^3.1.1'{{/json_serializable}}
 dev_dependencies:
   test: '>=1.3.0 <1.18.0'
 {{#json_serializable}}
-  build_runner: '^2.0.2'
-  json_serializable: '^4.1.1'{{/json_serializable}}
+  build_runner: '^1.10.9'
+  json_serializable: '^3.5.1'{{/json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -17,7 +17,7 @@ dependencies:
 {{#json_serializable}}
   json_annotation: '^3.1.1'{{/json_serializable}}
 dev_dependencies:
-  test: '>=1.3.0 <1.18.0'
+  test: '>=1.16.0 <1.18.0'
 {{#json_serializable}}
   build_runner: '^1.10.9'
   json_serializable: '^3.5.1'{{/json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -12,7 +12,7 @@ environment:
   sdk: '>=2.0.0 <3.0.0'
 dependencies:
   http: '>=0.12.0 <0.13.0'
-  intl: '^0.16.1'
+  intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
   json_annotation: '^3.1.1'{{/json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -11,7 +11,7 @@ homepage: '{{{pubHomepage}}}'
 environment:
   sdk: '>=2.0.0 <3.0.0'
 dependencies:
-  http: '>=0.12.0 <0.13.0'
+  http: '>=0.12.0 <0.13.3'
   intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -15,9 +15,9 @@ dependencies:
   intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
-  json_annotation: '^3.1.1'{{/json_serializable}}
+  json_annotation: '^4.0.1'{{/json_serializable}}
 dev_dependencies:
-  test: '>=1.3.0 <=1.17.3'
+  test: '>=1.3.0 <1.18.0'
 {{#json_serializable}}
   build_runner: '^2.0.2'
   json_serializable: '^4.1.1'{{/json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -19,5 +19,5 @@ dependencies:
 dev_dependencies:
   test: '>=1.3.0 <=1.17.3'
 {{#json_serializable}}
-  build_runner: '^1.0.0'
-  json_serializable: '^3.5.1'{{/json_serializable}}
+  build_runner: '^2.0.2'
+  json_serializable: '^4.1.1'{{/json_serializable}}
--- a/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
+++ b/modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
@@ -11,7 +11,7 @@ homepage: '{{{pubHomepage}}}'
 environment:
   sdk: '>=2.12.0 <3.0.0'
 dependencies:
-  http: '>=0.12.0 <=0.13.3'
+  http: '>=0.13.0 <0.14.0'
   intl: '^0.17.0'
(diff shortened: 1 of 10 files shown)


wing328:
cc @swipesight (2018/09) @jaumard (2018/09) @josh-burton (2019/12) @amondnet (2019/12) @sbu-WBT (2020/12) @kuhnroyal (2020/12) @agilob (2020/12)

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -11,8 +11,8 @@ homepage: '{{{pubHomepage}}}'
 environment:
   sdk: '>=2.0.0 <3.0.0'
```
jaumard, line 12: intl 0.17.0 minsdk is 2.12, you might want to enforce that here too

agilob: would this generate a project that doesn't compile?

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -11,8 +11,8 @@ homepage: '{{{pubHomepage}}}'
...
 dependencies:
-  http: '>=0.12.0 <0.13.0'
-  intl: '^0.16.1'
+  http: '>=0.12.0 <0.13.3'
```
kuhnroyal, line 14:
```suggestion
  http: '>=0.13.0 <0.14.0'
```
The changes no longer compile with `0.12.x`.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
jaumard: If you use a dart < 2.12 yes, pub get will tell you that intl can be use starting 2.12

kuhnroyal: Yes this would have to be `>=2.12.0 <3.0.0`

kuhnroyal (changes requested):
I want to leave https://github.com/OpenAPITools/openapi-generator/pull/8978#issuecomment-799767908 for consideration.
@agilob I am not sure if `json_serializable` is still involved but as you noted previously, this will most likely be a breaking change.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
kuhnroyal: The problem is, that if we update the lower bound to `>=2.12` then other dependencies will also require updates (`json_serializable`, maybe others`).

kuhnroyal:
We basically had the same discussion in https://github.com/OpenAPITools/openapi-generator/pull/8978 - the question that remains is: Can we live with a breaking change (due to Dart 2.12.0 requirement) or do we create a new generator like I did with the dart-dio-next generator? 
The `json_serializable` option is basically still experimental, if our only breaking changes are there, then we can go for it imo.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -9,15 +9,15 @@ authors:
...
-  http: '>=0.12.0 <0.13.0'
-  intl: '^0.16.1'
+  http: '>=0.12.0 <=0.13.3'
+  intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
   json_annotation: '^3.1.1'{{/json_serializable}}
 dev_dependencies:
-  test: '>=1.3.0 <1.16.0'
+  test: '>=1.3.0 <=1.17.3'
```
kuhnroyal, line 20:
```suggestion
  test: '>=1.16.0 <1.18.0'
```

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -9,15 +9,15 @@ authors:
...
-  sdk: '>=2.0.0 <3.0.0'
+  sdk: '>=2.12.0 <3.0.0'
 dependencies:
-  http: '>=0.12.0 <0.13.0'
-  intl: '^0.16.1'
+  http: '>=0.12.0 <=0.13.3'
+  intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
   json_annotation: '^3.1.1'{{/json_serializable}}
```
kuhnroyal, line 18: This needs an update and probably code changes.

agilob:
> The json_serializable option is basically still experimental, if our only breaking changes are there, then we can go for it imo.

agreed. braking change in this generator will be new field in an annotation, one line change to fix, but it will require D2.12

agilob:
Is this going to break a lot of other things? the whole nnbd, is it going to even compile with 2.12?

kuhnroyal:
Well, you can use Dart 2.12 without using NNBD. These changes are for now just Dart 2.12.
But `json_serializable` seems to be using the Dart type to determine nullability, this would require NNBD to be working.
Maybe null handling is now broken for the `json_serializable` option, not completely sure.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -9,15 +9,15 @@ authors:
...
-  intl: '^0.16.1'
+  http: '>=0.13.0 <0.14.0'
+  intl: '^0.17.0'
   meta: '^1.1.8'
 {{#json_serializable}}
-  json_annotation: '^3.1.1'{{/json_serializable}}
+  json_annotation: '^4.0.1'{{/json_serializable}}
 dev_dependencies:
-  test: '>=1.3.0 <1.16.0'
+  test: '>=1.3.0 <1.18.0'
```
kuhnroyal, line 20: This still needs a lower bound of `>=1.16.0`

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
```diff
@@ -9,15 +9,15 @@ authors:
...
 {{#json_serializable}}
-  build_runner: '^1.0.0'
-  json_serializable: '^3.5.1'{{/json_serializable}}
+  build_runner: '^2.0.2'
```
kuhnroyal, line 22: I would go with `^2.0.0` here, no need to force `2.0.2`.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
Adi1231234: why?

kuhnroyal:
This is starting to look like it may work, nice work @Adi1231234!
 
@agilob The tests for `petstore_json_serializable_client_lib_fake` are currently disabled but that is cause they don't compile fully anyways right?
```xml
<!--<module>samples/openapi3/client/petstore/dart2/petstore_json_serializable_client_lib_fake</module>-->
```

All the other samples are compiling. That sadly doesn't mean everything will work at runtime :) We need some people to test this.

modules/openapi-generator/src/main/resources/dart2/pubspec.mustache
kuhnroyal: Because https://pub.dev/packages/test/changelog#1160-nullsafety8 is the first version that fully support running on Dart 2.12 which is now required. Anything below 1.16.0 will not compile.

agilob:
>@agilob The tests for petstore_json_serializable_client_lib_fake are currently disabled but that is cause they don't compile fully anyways right?

yea, never managed to enable them as that generator was never finalized.

kuhnroyal (commented):
Ok, in the current state, I believe this may work but I have not tested it. I suggest to not merge this before the 5.1.1 release. I think this should go into 5.2.0

Adi1231234:
> Ok, in the current state, I believe this may work but I have not tested it. I suggest to not merge this before the 5.1.1 release. I think this should go into 5.2.0

Do I have to do anything else to get it into 5.2.0?

kuhnroyal:
> > Ok, in the current state, I believe this may work but I have not tested it. I suggest to not merge this before the 5.1.1 release. I think this should go into 5.2.0
> 
> Do I have to do anything else to get it into 5.2.0?

I think you can just switch the target branch to 5.2.x and then it can be merged, might have to rebase.

Adi1231234:
> > > Ok, in the current state, I believe this may work but I have not tested it. I suggest to not merge this before the 5.1.1 release. I think this should go into 5.2.0
> > 
> > 
> > Do I have to do anything else to get it into 5.2.0?
> 
> I think you can just switch the target branch to 5.2.x and then it can be merged, might have to rebase.

Thanks, I just did it, I would be happy if you could approve it

kuhnroyal (approved):
LGTM, was waiting with approve only so that @wing328 doesn't merge it into master.
Thanks for your work on this!

agilob (approved):

agilob:
I'm guessing it's time to upgrade json_serializable after this gets in

kuhnroyal:
Yea, and NNBD :)

Adi1231234:
Hi, I do not know so much how merges works here. but I noticed that the branch has not been marged.
I wanted to ask, should I do something else or talk to someone in order the branch to be merge?

agilob:
this PR will be closed after it's merged and you will receive a notification

kuhnroyal:
I am sure @wing328 was busy with the 5.1.1 release. This will be merged in the next few days, no worries.

PR #9426, opened 2021-05-07
[csharp][httpclient] Introduce FileParameter as an abstraction

Blackclaws:
This introduces a new class FileParameter as an abstraction to use in our ApiClients. It contains an automatic implicit conversion for existing stream parameters so its backwards compatible to existing code. 

The reason for this is, that there are platforms that have file like parameters and we currently have no way to pass a filename into our api clients aside from FileStreams. However for Browser platforms there are no FileStreams there are IBrowserFiles that do not map. So we need another abstraction.

This could be further expanded to add content types to the files.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


wing328:
cc @lucamazzanti as well (new C# technical committee member)

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
```diff
@@ -214,6 +214,7 @@ namespace {{packageName}}.Client
             _httpClient = new HttpClient(_httpClientHandler, true);
             _disposeClient = true;
             _baseUrl = basePath;
+            #warning
```
lucamazzanti, line 217: probably to remove, this generates a c# warning directive without a message

lucamazzanti (approved):
I had the same needs, and I must override the client locally to avoid to clone and customize the apiclient.mustache file. After that PR, i will revert my customization and keep using it vanilla 100%.

A really useful feature to support also uploads from FileStreams but with the need to override the filename during upload.

It seems good, just watch out the warning i think you left.

modules/openapi-generator/src/main/resources/csharp-netcore/libraries/httpclient/ApiClient.mustache
Blackclaws: Ah yes, that was an accident I'll remove it

PR #9429, opened 2021-05-07
[csharp][aspnetcore] Allow the use of endpoints (single class per operation)

Blackclaws:
This adds support for endpoints, i.e. a single class per operation to the generator. Unfortunately the way the generator currently works they are still all part of the same file. We can discuss whether we want to put them under an even further nested namespace. 
As a side effect enabled actual usage of annotations. Might want to expand the annotations of the normal controllers later on as well.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/docs/generators/aspnetcore.md
+++ b/docs/generators/aspnetcore.md
@@ -36,6 +36,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |useCollection|Deserialize array types to Collection&lt;T&gt; instead of List&lt;T&gt;.| |false|
 |useDateTimeOffset|Use DateTimeOffset to model date-time properties| |false|
 |useDefaultRouting|Use default routing for the ASP.NET Core version.| |true|
+|useEndpoints|Use one class per operation| |false|
 |useFrameworkReference|Use frameworkReference for ASP.NET Core 3.0+ and PackageReference ASP.NET Core 2.2 or earlier.| |false|
 |useNewtonsoft|Uses the Newtonsoft JSON library.| |true|
 |useSwashbuckle|Uses the Swashbuckle.AspNetCore NuGet package for documentation.| |true|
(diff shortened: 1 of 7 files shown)


wing328:
cc @lucamazzanti as well (new C# technical committee member)

lucamazzanti (commented):
The namespace doesn't respect the folders, are you sure to keep the .Endpoints part?
Classes doesn't respect the Controller suffix convention, do you prefer to remove it because it can became a long name?

Blackclaws:
> The namespace doesn't respect the folders, are you sure to keep the .Endpoints part?
> Classes doesn't respect the Controller suffix convention, do you prefer to remove it because it can became a long name?

Good point, maybe we should actually change the folder instead? I think I want to keep the Endpoints part because I want to clearly separate it from the existing functionality regarding controllers. 

In general I would also like to have one file per operation however the way the generator works right now this simply isn't possible and would require a new option for the template system that allows creating one file per operation instead of per tag.

The Controller suffix convention is just that a convention and we are doing route generation via annotations anyway. I was thinking about actually adding an "Endpoint" Suffix but that might be redundant for people that already call it Endpoint in their apispec. Now that I\m thinking about it that might be something we can parse out however .... 

When you go by how other people are implementing a similar feature you can end up with even more different classes/names.

https://github.com/ardalis/CleanArchitecture/tree/main/src/Clean.Architecture.Web/Endpoints/ProjectEndpoints

Basically the structure would be: Endpoints.TAGEndpoints followed by classes that are named by the operationid

wing328:
i've filed https://github.com/OpenAPITools/openapi-generator/pull/17161 to add a new rule in openapi-normalizer to set `tags` to the operationId so as to support single "class" per operation (or per operationId) to be more precise.

Please take a look when you've time.

wing328:
#17161 merged. please pull the latest master to give it a try. closing this one.

Issue #9200, opened 2021-04-07
PHP 7.2 end of life


ybelenko:
PHP 7.2 reached end of life at 30 Nov 2020

> End of life - A release that is no longer supported. Users of this release should upgrade as soon as possible, as they may be exposed to unpatched security vulnerabilities.

### [PHP: Supported Versions](https://www.php.net/supported-versions.php):
- ~7.2 until 30 Nov 2020~
- 7.3 until 6 Dec 2021
- 7.4 until 28 Nov 2022
- 8.0 until 26 Nov 2023

It seems that we need to change minimum PHP version to 7.3 in all php generators.

### Current Requirements:
| Generator | PHP version |
|:---------:|:------------:|
| ~php client~ | >= 7.3 \|\| >= 8.0         |
| Laravel server | >= 7.2.5      |
| Lumen server | >= 7.2.5        |
| ~Slim4 server~ | >= 7.3             |
| **Symfony server** | >= **7.1.3**     |
| ~Ze-ph server~ | >= 7.3            |

cc @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh @renepardon

ybelenko:
Before I start to update PHP version in generators I would like to know what community thinks about 8.0 support.
1. Can we say that code which works under 7.3 could also works under 8.0?
2. Is there any appropriate/recommended TravisCI config to test against 8.0? I just checked Travis docs, seems like there is no Ubuntu environment with preinstalled php 8.0. Neither mentions about 8.0 in general.
3. Is composer constraint `"php": "^7.3 || ^8.0"` correct?

cc @Articus as active PHP contributor

Articus:
>     1. Can we say that code which works under 7.3 could also works under 8.0?

I do not think that it is safe to assume that - PHP 8.0 is a major release after all with [few backward incompatible changes](https://www.php.net/manual/en/migration80.incompatible.php) and even if generated code runs fine on 8.0 its dependencies may not support it yet.

>     2. Is there any appropriate/recommended TravisCI config to test against 8.0? I just checked Travis docs, seems like there is no Ubuntu environment with preinstalled php 8.0. Neither mentions about 8.0 in general.

[...]

PR #9445, opened 2021-05-10
[php-mezzio-ph] PHP 8 support

Articus:
Huge update for php-mezzio-ph generator:
- minimal PHP version bump to 7,3 - #9200 
- proper support for container data type - multidimensional arrays and maps are handled correctly now
- custom CLI option "modern" to generate code using modern language syntax for PHP 8 projects
- support for referenced query parameters and referenced query parameter schemas 
- removal of obsolete samples folder `samples/openapi3/server/petstore/php-mezzio-ph`

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


wing328:
Tested locally and can start a server without issues
```
$ php -S 0.0.0.0:8080 -t ./public
[Fri May 21 14:38:31 2021] PHP 7.4.3 Development Server (http://0.0.0.0:8080) started
```

wing328:
Some users may consider this as a breaking change and we may add a `php-mezzio-ph-deprecated` generator to keep them happy.

Articus:
Thought about that :) But I believe it should be done next year after PHP 8.1 release. Currently there are [three supported version of PHP](https://www.php.net/supported-versions.php) - 7.3 , 7.4 , 8.0 . By default (without `modern` option) `php-mezzio-ph` generates [code](https://github.com/Articus/openapi-generator/tree/php-mezzio-ph-update/samples/server/petstore/php-mezzio-ph) that is compatible with them all (and even with 7.1 and 7.2 that are no longer supported). With `modern` option generated [code](https://github.com/Articus/openapi-generator/tree/php-mezzio-ph-update/samples/server/petstore/php-mezzio-ph-modern) heavily uses new features introduced in 8.0 and can no longer run on 7.* . So right now full PHP 8 support is an optional feature, not a breaking change.

ybelenko:
> Some users may consider this as a breaking change and we may add a `php-mezzio-ph-deprecated` generator to keep them happy.

It's hard to say what we can consider breaking changes in server codegens. If I remember correct @jimschubert thinks that server generators cannot cause breaking changes at all because users generate fresh stubs during migration anyway.

.gitignore
```diff
@@ -177,7 +177,6 @@ samples/client/petstore/python-tornado/.venv/
 # PHP
 samples/client/petstore/php/OpenAPIClient-php/composer.lock
 samples/openapi3/server/petstore/php-symfony/SymfonyBundle-php/composer.lock
-samples/openapi3/server/petstore/php-mezzio-ph/composer.lock
```
ybelenko, line 180: @Articus Could you explain why do you consider to commit `composer.lock`? Since we build library not a project it's not really important to install exactly the same build on each contributor's machine. 

I don't get it, in 99% cases lock file is just an additional problem to me, especially with autodeploy. When I install new package and commit both `composer.json` and `composer.lock` then server runs `composer install` it always says like `Your lock file is different from deps, please run composer update bla-bla` So I need to manually SSH to remote server and run `composer update` over there which updates all deps and now `composer.lock` contains uncommitted changes, so next GitHub hook will overwrite lock file again and build will be broken, crazy. Maybe I miss something about composer CLI, cannot find a decent way to build remote server on a webhook.

@wing328 What do you think about committing `composer.lock` in our case?

samples/server/petstore/php-mezzio-ph-modern/composer.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "git_user_id/git_repo_id",
+    "description": "",
+    "license": "unlicense",
+    "version": "1.0.0",
+    "type": "project",
+    "require": {
+        "php": "^8.0",
```
ybelenko, line 8: Does it install properly with PHP 7.3? I thought we need `^7.3 || ^8.0` version constraint.

modules/openapi-generator/src/main/resources/php-mezzio-ph/README.md.mustache
```diff
@@ -5,9 +5,9 @@ Generated by the [OpenAPI Generator](https://openapi-generator.tech) project.
 ## Overview
 This server stub aims to provide light, yet comprehensive structure for your API project using:
 
-- PHP: >=7.2
+- PHP: >=7.3
```
ybelenko, line 8: Maybe it makes sense to leave a note about PHP 8 support in README? If you didn't test it carefully under 8.0 then I would add a notice about it.

.gitignore
Articus: Fully agree about copmposer.lock :) But this line is removed because folder `samples/openapi3/server/petstore/php-mezzio-ph` does not exist any longer.

For historical reasons there were two samples for `php-mezzio-ph` - for OAS 2 spec file `samples/server/petstore/php-mezzio-ph` and for OAS 3 spec file `samples/openapi3/server/petstore/php-mezzio-ph`. During [my previous PR](https://github.com/OpenAPITools/openapi-generator/pull/8145) these folders became equal because according [this comment](https://github.com/OpenAPITools/openapi-generator/pull/8145#issuecomment-744177698) samples for OAS 2 spec are no longer needed. So I removed `samples/openapi3/server/petstore/php-mezzio-ph` entirely.

samples/server/petstore/php-mezzio-ph-modern/composer.json
Articus: Nope, samples in `samples/server/petstore/php-mezzio-ph-modern` are generated with `modern=true` (it is a new CLI option for `php-mezzio-ph`) and this code can run only on PHP 8.0 . Samples in `samples/server/petstore/php-mezzio-ph` are generated with `modern=false` and [its `composer.json`](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/php-mezzio-ph/composer.json) contains valid `"php": "^7.3 || ^8.0"`.

modules/openapi-generator/src/main/resources/php-mezzio-ph/README.md.mustache
Articus: As I wrote [in previous comment](https://github.com/OpenAPITools/openapi-generator/pull/9445#issuecomment-845716159) it is tested and works against 8.0 :) I can add extra text to emphasize that but aren't  requirements in [composer.json](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/php-mezzio-ph/composer.json) enough?

ybelenko:
@Articus thanks for clarifications I wasn't focused enough, didn't notice you've added two generators. That's not common codegen flow.

Issue #9448, opened 2021-05-11
[Kotlin][Retrofit] Change OkHttpClient parameter to Call.Factory


shanselm-ergon:
With OpenApi version 5 a new constructor parameter got added to pass an `OkHttpClient` to the `ApiClient` ([#6855]). We thought this is good enough for us, but now we run into some problems with this approach. The parameter should be changed to a `Call.Factory` instead.

- Smaller interface
- Only the Call.Factory is used (see `retrofitBuilder.client(usedClient)`, this just sets the `Call.Factory`)
- It is more like it was before version 5 of OpenApi (the exposed `adapterBuilder` had a `callFactory` method)
- In our use case, we need to replace the underlying http-client during runtime. If we only have to pass the `Call.Factory` we can easily do this, but if we have to pass the client itself we can't achieve this properly. 

## Describe the solution you'd like

Change the parameter from `OkHttpClient` to `Call.Factory`. Each `OkHttpClient` is also a `Call.Factory`, so minimal effort to adopt this change. 

## Describe alternatives you've considered

With the knowledge that only the `Call.Factory` interface is used by the `ApiClient` we could also create our own version of a `OkHttpClient` and only overwrite the `newCall(request)` method. But this is based on internal knowledge, which should not be part of the `ApiClient` API.

I will create a pull request suggestion for this improvement.

shanselm-ergon:
See pull request [#9451]

shanselm-ergon:
Resolved with [#9451]

PR #9451, opened 2021-05-11
[Kotlin][retrofit2] replace okhttpclient with callfactory

shanselm-ergon:
[Kotlin][retrofit2] replace okhttpclient with callfactory

fixes [#9448] 

### PR checklist
 
- [/] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [/] Pull Request title clearly describes the work in the pull request and Pull Request description provides details about how to validate the work. Missing information here may result in delayed response from the community.
- [/] Run the following to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) and update samples:
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [/] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `5.1.x`, `6.0.x`
- [/] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request.

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-retrofit2/infrastructure/ApiClient.kt.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-retrofit2/infrastructure/ApiClient.kt.mustache
@@ -21,6 +21,7 @@ import {{packageName}}.auth.HttpBearerAuth
 {{/authMethods}}
 {{/hasAuthMethods}}
 
+import okhttp3.Call
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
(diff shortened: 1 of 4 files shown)


4brunu:
@shanselm-ergon first of all, thanks for contributing to the kotlin client generator.
The problem I see with this, is that this will create a breaking change without fallback, this implies to targeting the branch `6.0.x`.
Do you have any idea on how to make this changes backwards compatible?

shanselm-ergon:
The only way to keep this backwards compatible would be, to use the same parameter name `okHttpClient` for the `Call.Factory`. The rest would still be compatible, because every `OkHttpClient` is also a `Call.Factory`. Under the hood we anyway used the client only as factory, so there would not be a change for the users. The naming would just not be as accurate as it could be.

4brunu:
@shanselm-ergon Thanks a lot for your explanation.
Since `OkHttpClient` is also a `Call.Factory`, I think maybe we can keep this PR as is and consider it a breaking change with fallback because if the users of this library don't use named parameters, they don't need to change anything, and if they use, the only thing that they have to do is to change the name from `okHttpClient` to `callFactory` which is a compile time error, and the IDE helps with the migration, so it seems reasonable to me.
Let's see what @wing328 thinks of this.

wing328:
Let's go with this. If users prefer a way to make it backward compatible, we can introduce an option (but not something I prefer to avoid having too many options for a generator)

4brunu:
Looks good to me.
As @shanselm-ergon mentioned, the worth thing would be if someone uses named parameters, that would need to change this
```
ApiClient(..., okHttpClient = OkHttpClient(),  ...)
```
into this
```
ApiClient(..., callFactory = OkHttpClient(),  ...)
```

And if the users don't use named parameters, they don't need to change nothing.
@wing328 for me you can merge this PR. 👍

PR #9457, opened 2021-05-11
Fix loss of multiple header values for python api client

jkupferer:
Fix Python library headers loss of multiple values by switching from `dict` to use of urllib3 `HTTPHeaderDict`. In particular this fix is needed for passing multiple values to `Impersonate-Group` header.

https://github.com/kubernetes-client/python/issues/1462
https://github.com/kubernetes-client/python/pull/1463

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@taxpon @frol @mbohlool @cbornet @kenjones-cisco @tomplus @Jyhess @arun-nalla @spacether


spacether:
Thank you for your PR. Your code looks like it fixes the issue but there are no tests. So if someone changes this code in the future there would be no CI test telling them that they broke a feature.
How about adding a test of your needed use case here?
https://github.com/OpenAPITools/openapi-generator/tree/master/samples/openapi3/client/petstore/python/tests_manual
That will future-proof this feature.

samples/openapi3/client/petstore/python/tests_manual/test_api_call_headers.py
```diff
@@ -0,0 +1,90 @@
...
+            mock_method.return_value = self.mock_response([])
+
+            api_client.call_api(
+                '/fake',
+                'GET',
+                collection_formats = {'Impersonate-Group': 'multi'},
+                header_params = {
+                    'Impersonate-Group': ['dev', 'admin'],
+                },
+            )
```
spacether, line 72: Hey there, this looks great. This is almost what I want. You correctly used response mocking and asserted that the mocked method was called with your expected inputs. But call_api should be a private method that is not used by a developer.
Can you call a real method that exists on the api_client and keep your assert called. with in line 74?
You can see examples of us doing that here: https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/tests_manual/test_fake_api.py#L103

If you need to add a new endpoint to our python spec that you use for this test, you can add it to this file: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/python/petstore-with-fake-endpoints-models-for-testing-with-http-signature.yaml

jkupferer: Ah, well the reason I'm using `call_api` opens up a number of more questions about headers in the library. While the ApiClient has a `set_default_header` method, it has no mechanism to specify the collection format to use with the header.

The use of `collection_formats` in the code seems a bit odd in the way that this setting is used in so many different places in calls to `parameters_to_tuples` for headers as well as path, query, and post parameters... As best I can tell there is no reasonable way to configure the client with default `collection_formats` nor a way to pass a setting from a higher level call to the library... looking into another approach.

jkupferer: Okay, I have pushed a new implementation and I think a better one. This moves the usage of HTTPHeaderDict to the default headers and adds a method to add header values, ex:

```
        api_client = ApiClient()
        api_client.add_default_header('Impersonate-Group', 'dev')
        api_client.add_default_header('Impersonate-Group', 'admin')
```

I still need to refine the test as you asked...

jkupferer:
I believe I just spotted an issue with `test_fake_api.py`:

https://github.com/OpenAPITools/openapi-generator/blob/master/samples/openapi3/client/petstore/python/tests_manual/test_fake_api.py#L78-L85

```
        if 'body' in kwargs:
            used_kwargs['body'] = kwargs['body']
        else:
            mock_method.assert_called_with(
                http_method,
                url,
                **used_kwargs
            )
```

When `body` is passed as a keword argument the `assert_called_with` is skipped and no test condition is evaluated.

jkupferer:
Fixing the bug in the test suite where it wasn't testing requests with `body` unmasked what look like errors in the library.

For example:

```
AssertionError: expected call not found.
Expected: request('POST', 'http://petstore.swagger.io:80/v2/fake/refs/arraymodel', _preload_content=True, _request_timeout=None, headers=HTTPHeaderDict({'Accept': 'application/json', 'User-Agent': 'OpenAPI-Generator/1.0.0/python', 'Content-Type': 'application/json'}), query_params=[], post_params=[], body=[{'className': 'Cat', 'color': 'black'}])
Actual: request('POST', 'http://petstore.swagger.io:80/v2/fake/refs/arraymodel', headers=HTTPHeaderDict({'User-Agent': 'OpenAPI-Generator/1.0.0/python', 'Accept': 'application/json', 'Content-Type': 'application/json'}), query_params=[], post_params=[], _preload_content=True, _request_timeout=None, body=[{'className': 'Cat', 'color': 'black', 'class_name': 'Cat'}])
```

And the tests for `getInlineAdditionalPropertiesPayload` were entirely broken... I have included fixes for the test suite... but it is not really testing for correct behavior like this. I am thinking fixing the `test_fake_api.py` tests needs to just be separated from this issue and fixed by whoever goes after the issues it has been hiding.

The tests for the headers can be kept separate?

spacether:
Thank you for uncovering that issue. Can you file a separate issue on it?
Yes we should probably keep the tests in a separate file for that reason.
So you have uncovered an issue with how our python code sets headers. Basically the current way that it chooses a header is probably not great. For required headers where there is only one value we should probably load in the known value as the default value. For all other headers we should probably require the client to choose and explicitly pass in headers.

So far, our api_client has no state and configuration data inside itself except for a reference to the Configuration class as far as I know. So I am concerned about adding this new info there.
Can you further describe your use case?
1. are these spec headers that you are setting with data that the spec describes types for?
2. are these spec headers that you are setting with data that the spec describes enum values for?
3. are these additional headers that the spec does not know about at all?

jkupferer:
> Can you further describe your use case?
> 
> 1. are these spec headers that you are setting with data that the spec describes types for?
> 2. are these spec headers that you are setting with data that the spec describes enum values for?
> 3. are these additional headers that the spec does not know about at all?

I am afraid I don't know how to answer these questions. I started working with the Python kubernetes-client (https://github.com/kubernetes-client/python), which is generated using the openapi-generator. I needed a way to pass multiple `Impersonate-Group` headers according to the Kubernetes API documentation (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation).

It looks like the Kubernetes API spec (https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json) does not say anything about this header.

The values for this header are group name strings such as would be provided by an identity and access management system. These are definitely not something that can be enumerated.

spacether:
Thanks for providing that extra info. The spec that I was referring to in questions 1-3 is your Kubernetes API spec.
So because that spec does not include definition of the `Impersonate-Group` header, our autogen code does not know anything about it.

How about you add an _additional_headers dict to all endpoints that allows a user to provide additional headers that are passed through to endpoint calls. This should allow you to use any normal endpoint call and pass in your required additional headers. The reason that I prefer the leading underscore on the variable name is that I want a user to be able to define a different additional_headers parameter that will not collide with our parameter.

One could add that parameter here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api.mustache#L47
As a distinct parameter, or one can add it to the kwargs docstring there, and check for its presence in our call_with_http_info
here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api_client.mustache#L858
And one would probably want to start with the _additional_header and then add the params['header'] on top of them.
Something like:
```
sent_headers = {}
if kwargs['_additional_headers']:
    sent_headers.update(kwargs['_additional_headers'])
sent_headers.update(params['header'])

return self.api_client.call_api(
    self.settings['endpoint_path'], self.settings['http_method'],
    params['path'],
    params['query'],
    sent_headers,
    ...
````
    
```

jkupferer:
I have reverted changes to `test_fake_api.py` and opened an issue documenting the logic error in the test code:

https://github.com/OpenAPITools/openapi-generator/issues/9482

jkupferer:
> How about you add an _additional_headers dict to all endpoints that allows a user to provide additional headers that are passed through to endpoint calls. This should allow you to use any normal endpoint call and pass in your required additional headers...

I see how that could be helpful, though if `sent_headers` is a plain python dict it wouldn't support multiple instances of the same header. If it were an HTTPHeaderDict then it would provided the needed functionality.

Perhaps everywhere the library is dealing with headers it should be using HTTPHeaderDict rather than plain python dictionaries?

spacether:
Having _additional_headers be of HTTPHeaderDict is fine too. Does dict work also if we include your HTTPHeaderDict processing deeper in the code?
I remember one of your earlier inputs being:
{'Impersonate-Group': ['dev', 'admin']} or am I misremembering?

jkupferer:
> I remember one of your earlier inputs being:
> {'Impersonate-Group': ['dev', 'admin']} or am I misremembering?

Right, though that causes some trouble with `params_to_tuples`:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api_client.mustache#L502

The `collection_formats` here would somehow need the key of the header name to configure it to unpack the list value as `multi`. The way that `collection_formats` configures header, path, query, and post parameters seems a bit odd. It is unlikely that a header and another parameter would have the same name... but it seems like a fragile design that this one setting influences the processing of so many logically unrelated behaviors. I would rather not depend on it with any changes. Putting HttpHeaderDict at a higher level feels cleaner.

spacether:
Aren't those behaviors related?
They are how the collectionFormation directs the tooling to send the parameters to the server per:
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#parameter-object
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#style-values

When you pass in an instance of HTTPHeaderDict for _additional_headers sounds good.

jkupferer:
> Aren't those behaviors related?

Definitely related, but... well the specification says, "A unique parameter is defined by a combination of a name and location." If I understand the code correctly, the implementation of `collection_formats` is only configuring by name, not name + location. If an API used the same name for a path parameter and a post parameter the collection format could not treat these independently.

spacether:
> > Aren't those behaviors related?
> 
> Definitely related, but... well the specification says, "A unique parameter is defined by a combination of a name and location." If I understand the code correctly, the implementation of `collection_formats` is only configuring by name, not name + location. If an API used the same name for a path parameter and a post parameter the collection format could not treat these independently.

If you want to refactor this to better conform to the guidance on serializing parameter data go right ahead. That would be a great addition.

jkupferer:
Update... I found that repeated headers with different values is not entirely supported in the python urllib3 library. My early tests did not have request bodies. When I tested with a POST request I found that the multiple headers get translated into a single header with a comma separated value.

https://github.com/urllib3/urllib3/issues/2242

spacether:
> Update... I found that repeated headers with different values is not entirely supported in the python urllib3 library. My early tests did not have request bodies. When I tested with a POST request I found that the multiple headers get translated into a single header with a comma separated value.
> 
> [urllib3/urllib3#2242](https://github.com/urllib3/urllib3/issues/2242)

Couldn't you just set the value of the header to be '1, 2' rather than trying to assign 1 + 2 as values to the same header?

spacether:
> Thank you for uncovering that issue. Can you file a separate issue on it?
> Yes we should probably keep the tests in a separate file for that reason.
> So you have uncovered an issue with how our python code sets headers. Basically the current way that it chooses a header is probably not great. For required headers where there is only one value we should probably load in the known value as the default value. For all other headers we should probably require the client to choose and explicitly pass in headers.
> 
> So far, our api_client has no state and configuration data inside itself except for a reference to the Configuration class as far as I know. So I am concerned about adding this new info there.
> Can you further describe your use case?
> 
> 1. are these spec headers that you are setting with data that the spec describes types for?
> 2. are these spec headers that you are setting with data that the spec describes enum values for?
> 3. are these additional headers that the spec does not know about at all?

FYI If you rebase on master I fixed the test_fake_api problems. Thanks for filing that issue on it :)

jkupferer:
> Couldn't you just set the value of the header to be '1, 2' rather than trying to assign '1' + '2' as values to the same header?

I wish I could, but the Kubernetes API doesn't adhere to RFC 7230 (https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2). Comma separated values for `Impersonate-Group` are not treated the same as repeating the header with different values. From what I have seen from searching around it seems a number of APIs have similar issues with headers. It seems developers have often ignored this rule regarding header behavior.

jkupferer:
It looks like this needs to be shelved for now, at least until the 2.x release of urllib3. For now, here's a link to the rather ugly little hotfix patch I'm doing to urllib3 and the kubernetes API.

https://github.com/redhat-cop/babylon/blob/main/catalog/api/hotfix.py

PR #9489, opened 2021-05-15
New generator | Arduino cpp

AndersSpringborg:
This is a new generator for the Arduino platform.

We have targeted it for a esp32, and esp 8266 chip, and used a [library for HTTP request](https://github.com/espressif/arduino-esp32/tree/master/libraries/HTTPClient) for HTTP protocol. It is implemented with HTTPS, which is a challenge for microcontrollers, as there's a limited amount of flash.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


wing328:
cc @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @martindelille (2018/03) @muttleyxd (2019/08)

modules/openapi-generator/src/main/resources/tiny-cpp-client/helpers-header.mustache
```diff
@@ -0,0 +1,24 @@
+#ifndef _HELPERS_H_
+#define _HELPERS_H_
+
+#include <string>
+#include "bourne/json.hpp"
+
+using namespace std;
```
muttleyxd, line 7: This should be removed. Stuff from std namespace will leak everywhere where helpers are used. If you really need it, then please move it to headers-body, but I'd prefer prefixing with `std::` everywhere it's required.

https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

modules/openapi-generator/src/main/resources/tiny-cpp-client/model-header.mustache
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	void {{setter}}({{dataType}} {{#isContainer}}{{#isMap}}<std::string, std::string>{{/isMap}}{{^isMap}}<{{#items}}{{dataType}}{{/items}}>{{/isMap}}{{/isContainer}} {{name}});
+	{{/vars}}
+
+
+    private:
+    {{#vars}}
+	{{dataType}} {{#isContainer}}{{#isMap}}<std::string, std::string>{{/isMap}}{{^isMap}}<{{#items}}{{dataType}}{{/items}}>{{/isMap}}{{/isContainer}}{{name}};
+	{{/vars}}
+    void __init();
```
muttleyxd, line 66: Using double underscore is undefined behavior, as specified by C++ standard. Problem is that standard library implementation can use something like `#define __init whatever` - it's what C++ standard allows them to do.

I'd change to `init()` for simplicity. If you really want an underscore for private members, then I'd put it at the end `init_()`

https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

modules/openapi-generator/src/main/resources/tiny-cpp-client/model-body.mustache
```diff
@@ -0,0 +1,177 @@
...
+	{{^isContainer}}
+	{{name}} = {{defaultValue}};
+	{{/isContainer}}
+	{{/vars}}
+}
+
+void
+{{classname}}::__cleanup()
+{
+    //{{#vars}}if({{name}} != NULL) {
```
muttleyxd, line 38: Since this is unused, I guess it can be removed (unless it's WIP)

modules/openapi-generator/src/main/resources/tiny-cpp-client/model-body.mustache
```diff
@@ -0,0 +1,177 @@
+
+{{#models}}{{#model}}
+#include "{{classname}}.h"
+
+using namespace Tiny;
+
+{{classname}}::{{classname}}()
+{
+    __init();
```
muttleyxd, line 9: Is there a reason for keeping constructor's body (and destructor's too) in a separate method?

modules/openapi-generator/src/main/resources/tiny-cpp-client/helpers-body.mustache
```diff
@@ -0,0 +1,102 @@
+#include "Helpers.h"
+#include <string>
+#include <sstream>
+
+bool isprimitive(string type){
+	if(type.compare("std::string") == 0||
```
muttleyxd, line 6: I guess it'd be a little easier to read if it was like
```
if (type == "std::string" ||
    type == "int" ||
    type == "long" || ...```

File: modules/openapi-generator/src/main/resources/tiny-cpp-client/helpers-body.mustache [hunk 6]
```diff
@@ -0,0 +1,102 @@
...
+    else if (type.compare("int") == 0) 
+    {
+		int* val = static_cast<int*> (target);
+		*val = value.to_int();
+	} 
+    
+    else if (type.compare("float") == 0) 
+    {
+		float* val = static_cast<float*> (target);
+		*val = (float)(value.to_float());
```
muttleyxd, line 43: I'd replace this C-style cast with C++ static_cast

modules/openapi-generator/src/main/resources/tiny-cpp-client/helpers-header.mustache
```diff
@@ -0,0 +1,24 @@
...
+
+bool isprimitive(string type);
+
+void jsonToValue(void* target, bourne::json value, string type);
+
+std::string stringify(long input);
```
muttleyxd, line 13: I think all of these methods could be replaced with std::to_string()

modules/openapi-generator/src/main/resources/tiny-cpp-client/service/api-body.mustache
```diff
@@ -0,0 +1,236 @@
...
+            }
+            payload = tmp_arr.dump();
+            {{/isPrimitiveType}}
+            {{/items}}
+            {{/isArray}}
+            {{/isContainer}}
+
+            {{^isContainer}}
+            {{#isPrimitiveType}}
+            payload = stringify({{paramName}});
```
muttleyxd, line 112: If you remove stringify method then change this to

```
payload = {{^isString}}std::to_string({{/isString}}{{paramName}}{{^isString}}){{/isString}};
```

modules/openapi-generator/src/main/resources/tiny-cpp-client/helpers-header.mustache
```diff
@@ -0,0 +1,24 @@
+#ifndef _HELPERS_H_
```
muttleyxd, line 1: Same as with double underscore - underscore and upper character combination is reserved by C++ standard.

Also I think you could switch to `#pragma once`

samples/client/petstore/tiny/cpp/lib/Models/ApiResponse.h
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	std::string getMessage();
+
+	/*! \brief Set 
+	 */
+	void setMessage(std::string  message);
+
+
+    private:
+	int code;
```
muttleyxd, line 72: We could use

```
int code{};
...
std::string type{};
```

This will zero-initialize primitive variables - with it we could skip having to initialize it explicitly in the constructor.

modules/openapi-generator/src/main/resources/tiny-cpp-client/service/api-header.mustache
```diff
@@ -0,0 +1,74 @@
...
+{{#operations}}
+/**
+ *  Class {{basename}}
+ * Generated with openapi::tiny-cpp-client
+ */
+
+class {{classname}} : public AbstractService {
+public:
+    {{! Constructor }}
+    {{classname}}();
```
muttleyxd, line 26: Constructor and destructor are always empty - they could be set to default and removed from api-body.mustache.

```
{{classname}}() = default;
virtual ~{{classname}}() = default;
```

modules/openapi-generator/src/main/resources/tiny-cpp-client/service/api-body.mustache
```diff
@@ -0,0 +1,236 @@
...
+            {{/isPrimitiveType}}
+
+            {{^isPrimitiveType}}
+            payload = {{paramName}}.toJson().dump();
+            {{/isPrimitiveType}}
+            {{/isContainer}}
+
+            {{/bodyParam}}
+            int httpCode = http.sendRequest("{{httpMethod}}", (uint8_t *) payload.c_str(), payload.length());
```
muttleyxd, line 121: Another C-style cast. I guess it should become reinterpret_cast.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TinyCppClientCodegen.java
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+
+    // FilePaths
+    private static final String sourceFolder = "src";
+    private static final String libFolder = "lib";
+    private static final String serviceFolder = libFolder + File.separator + "service";
+
+    // Types
+    private static final String cpp_array_type = "std::list";
```
muttleyxd, line 181: This is only used in `makeTypeMappings()`, so I guess it could be moved there.

muttleyxd (changes requested):
Good addition in general, but has some minor issues - nothing hard to fix.

AndersSpringborg:
Great feedback, we'll look into it soon

modules/openapi-generator/src/main/resources/tiny-cpp-client/model-body.mustache
AndersSpringborg: No, I think it was copied from another cpp class, where they used it outside the class

samples/client/petstore/tiny/cpp/lib/Models/ApiResponse.h
AndersSpringborg: If this is primitive types, does it call the defualt constructor for classes like std::string of std::list?

muttleyxd: Yeah, it will work for stuff that has constructors too. 
I think there is only one case where `{}` isn't enough - if you want to initialize a field using a parameter passed to constructor, eg.

```
class Something
{
public:
   Something(int some_important_parameter) : important_parameter(some_important_parameter) {}

private:
   int important_parameter; // you can't do something like '{some_important_parameter}'
};
```

modules/openapi-generator/src/main/resources/tiny-cpp-client/service/api-body.mustache
AndersSpringborg: We could't get reinterpret_cast<uint8_t*>() to work. We've copied what the underlying library does like this:
https://github.com/espressif/arduino-esp32/blob/a1d8b959b05e3fb1391883ce6f9f2ed87af7da0f/libraries/HTTPClient/src/HTTPClient.cpp#L513-L516

samples/client/petstore/tiny/cpp/lib/Models/ApiResponse.h
AndersSpringborg: OKay cool, thanks 🍰

modules/openapi-generator/src/main/resources/tiny-cpp-client/model-header.mustache
AndersSpringborg: We've removed init completely :)

modules/openapi-generator/src/main/resources/tiny-cpp-client/service/api-body.mustache
muttleyxd: Probably because you're casting away const from `const char* c_str()`. You can use `payload.data()`, take a look at std::string specification and see what's the difference between c_str and data.

https://en.cppreference.com/w/cpp/string/basic_string/c_str
https://en.cppreference.com/w/cpp/string/basic_string/data

AndersSpringborg:
Thanks for using your time on feedback, tips and tricks. This is our first time writing cpp and Java code, so every small detail is a huge help 😁

muttleyxd:
You're welcome :grin:
C++ isn't easy to learn for beginners, mostly because many libraries and learning materials are basing on old habits before C++11 even existed, so it's easy to pick them up and in result whole learning and programming process is much harder than it could be.

wing328:
@AndersSpringborg thanks for the PR. I'll merge this into master so that we can collect feedbacks from the users.

I'll submit a PR to make some minor enhancements to the new generator.

wing328:
@AndersSpringborg I've filed with the enhancements (renaming generator to conform to the convention, minor code format enhancement, etc): https://github.com/OpenAPITools/openapi-generator/pull/9560

Can you please PM me via Slack for a quick chat (IM) when you've time?

PR #9515, opened 2021-05-18
[dart][ci] Move Dart samples from CircleCI to Github actions

kuhnroyal:
**Suggestions/Assumptions**
* future release branches (`5.2.x`) should be named `release/5.2.x` or some use different pattern that can be globbed in the workflow files
* there should be a separate workflow for Samples of each platform/language

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/.github/workflows/samples-dart.yaml
+++ b/.github/workflows/samples-dart.yaml
@@ -4,11 +4,11 @@ on:
   push:
     branches:
       - master
-      - release/**
+      - '[5-9]+.[0-9]+.x'
   pull_request:
     branches:
       - master
-      - release/**
+      - '[5-9]+.[0-9]+.x'
 
 jobs:
   tests-dart-2-10:
--- a/.github/workflows/samples-dart.yaml
+++ b/.github/workflows/samples-dart.yaml
@@ -11,8 +11,8 @@ on:
       - release/**
 
 jobs:
-  tests:
-    name: Tests
+  tests-dart-2-10:
+    name: Tests Dart 2.10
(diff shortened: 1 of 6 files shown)


kuhnroyal:
@wing328 This also fixes a new formatting CI failure due to yesterdays Flutter/Dart release which now gets installed in CircleCI.

.github/workflows/samples-dart.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - release/**
+  pull_request:
+    branches:
+      - master
+      - release/**
```
wing328, line 11: What about using something like [5-9].[0-9].[0-9]?

ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

kuhnroyal: This will work for `pull_request` but for `push` it will also match tags. I guess that could be ok.

kuhnroyal: Nvm. tags are prefixed so this should work.

kuhnroyal: But you always use `x` for the patch version?

wing328: yup, please use `[5-9].[0-9].x`

wing328:
Both drone.io, travis CI failures have been fixed in the master.

Issue #9570, opened 2021-05-24
[BUG] [Python] Adding supporting files and custom package name with Python generator throws exception


shylacs:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

Python generator throws an exception when adding supporting files and custom package name

##### openapi-generator version

5.1.1

##### OpenAPI declaration file content or url
```yaml
openapi: 3.0.0
info:
  title: Some title
  version: 1.0
paths: {}
components:
  schemas: {}
  responses: {}
  parameters: {}
  examples: {}
  requestBodies: {}
  headers: {}
  securitySchemes:
    basic_auth:
      type: http
      scheme: basic
    api_key:
      type: apiKey
      name: id
      in: header
    Bearer:
      type: apiKey
      name: Authorization
      in: header
  links: {}
  callbacks: {}
tags: []
servers: []
```
runtime-config.yaml
```json
packageName: "infra_sdk.runtime"
projectName: "infra-sdk-runtime"
packageVersion: "0.1.0"
appDescription: "Some description"
generatorName: "python"
templateDir: "mustache-templates/python/runtime"
files:
  deprecated-api/translation-scheme.json:
    folder: deprecatedapitranslator
    destinationFilename: translation-scheme.json
    templateType: SupportingFiles
```

##### Generation Details

java -jar openapi-generator-cli.jar generate -g python -i openapi-runtime.yaml -c config/python/runtime-config.yaml -o ../out

##### Steps to reproduce

Using openapi-runtime.yaml as input spec and config as runtime-config.yaml and empty templates dir, trying to generate python bindings throws below exception
[...]

PR #9572, opened 2021-05-24
Fix python generation when custom files and templates are specified

shylasrinivas:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
@@ -938,7 +938,7 @@ public class DefaultGenerator implements Generator {
         // TODO: initial behavior is "merge" user defined with built-in templates. consider offering user a "replace" option.
         if (userDefinedTemplates != null && !userDefinedTemplates.isEmpty()) {
             Map<String, SupportingFile> supportingFilesMap = config.supportingFiles().stream()
-                    .collect(Collectors.toMap(TemplateDefinition::getTemplateFile, Function.identity()));
+                    .collect(Collectors.toMap(TemplateDefinition::getTemplateFile, Function.identity(), (oldValue, newValue) -> oldValue));
 
             // TemplateFileType.SupportingFiles
             userDefinedTemplates.stream()
--- a/docs/generators/tiny-cpp.md
+++ b/docs/generators/tiny-cpp.md
@@ -165,7 +165,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |Maps|✗|ToolingExtension
 |CollectionFormat|✓|OAS2
 |CollectionFormatMulti|✓|OAS2
-|Enum|✓|OAS2,OAS3
+|Enum|✗|OAS2,OAS3
 |ArrayOfEnum|✓|ToolingExtension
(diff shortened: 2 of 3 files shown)


shylasrinivas:
@taxpon  @frol  @mbohlool  @cbornet  @kenjones-cisco  @tomplus  @Jyhess  @arun-nalla  @spacether

shylasrinivas:
Fix for https://github.com/OpenAPITools/openapi-generator/issues/9570

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultGenerator.java
```diff
@@ -938,7 +938,7 @@ private void processUserDefinedTemplates() {
         // TODO: initial behavior is "merge" user defined with built-in templates. consider offering user a "replace" option.
         if (userDefinedTemplates != null && !userDefinedTemplates.isEmpty()) {
             Map<String, SupportingFile> supportingFilesMap = config.supportingFiles().stream()
-                    .collect(Collectors.toMap(TemplateDefinition::getTemplateFile, Function.identity()));
+                    .collect(Collectors.toMap(TemplateDefinition::getTemplateFile, Function.identity(), (oldValue, newValue) -> oldValue));
```
spacether, line 941: What is this lambda doing? Why is it returning the oldValue?
I have no way to tell from this PR if this code works or not.
If this feature is changed in the future, we will not have any test that fails and the feature will break.
How about adding a test of processUserDefinedTemplates to ensure that this feature will keep working?

shylasrinivas: @spacether Absolutely valid points. 
Collectors.toMap throws IllegalStateException when dup keys are found, hence the logic is to return a single value when dups are encountered. oldValue is an existing key in the map, whereas newValue is a key found during later iterations of the list (in this case config.supportingFiles()). The code simply retains the existing key (oldValue) in the map when a dup (newValue) is encountered. Based on my testing, I have found oldValue to be user defined templates, which is desired to be retained in the map.
I will look into adding a test. Agreed there is no way to tell if the code works, how do you typically ensure PR quality?

spacether: Thanks for explaining what this code is doing.
Most of our most basic tests are in: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java

shylasrinivas: @spacether Thank you for the pointers. I have added a test, but it looks like the config (line 684 of DefaultCodegenTest.java) I want to specify is not supported in CodegenConfigurator unless, I am missing something. Lmk if the test looks ok. I can remove the commented check based on your input.

spacether: Thank you for explaining that and adding the test. Your test looks good.

spacether (approved):
Thank you for your PR. This looks good

PR #9577, opened 2021-05-25
[cpp-restsdk] Update json double/float parse.

adameste:
Fix cpp-restsdk double and float parse.

double.NaN was parsed when server doesn't ensure the floating point value has decimal point in in.

Fixed incorrect server response json parsing. For example the following json:

```json
{
    "double_value": 42
}
```

As of now, the value 42 is parsed as NaN, because json library returns false on is_double().
Additional condition ensures NaN is not returned when not needed.
Implemented for both float & double data types.

### DIFF
--- a/docs/generators/tiny-cpp.md
+++ b/docs/generators/tiny-cpp.md
@@ -165,7 +165,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |Maps|✗|ToolingExtension
 |CollectionFormat|✓|OAS2
 |CollectionFormatMulti|✓|OAS2
-|Enum|✓|OAS2,OAS3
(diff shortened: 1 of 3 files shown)


wing328:
Thanks for the PR. Please follow the instruction in the PR template below to update the samples so that CI can test the change:

  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

cc @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @martindelille (2018/03) @muttleyxd (2019/08)

adameste:
Ran all three above mentioned commands & commited changes.

muttleyxd (approved):

PR #9583, opened 2021-05-25
Swift5 Vapor 4 client library

aymanbagabas:
Opened in favor of https://github.com/OpenAPITools/openapi-generator/pull/8515

TODOs:
- [x] Rebase onto https://github.com/OpenAPITools/openapi-generator/pull/9419

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


aymanbagabas:
@4brunu CI is passing, Travis is failing cause of Docker.

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
```diff
@@ -5,16 +5,25 @@
 //
 
 import Foundation
+{{#useVapor}}
+import Vapor
+{{/useVapor}}
 
-{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{projectName}}API {
+{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class {{projectName}} {
```
4brunu, line 12: @aymanbagabas this is a breaking change that affects a lot of file, and what I remember from the other PR is that this change was only related to naming, so I suggest to revert it please.

4brunu:
@aymanbagabas I left a comment on an issue that is a breaking change and creates a lot of noise.
Could you please address it?
This PR contains a lot a changed files.
Ideally to make this PR easier to review and merge, this PR should only contain changes related to vapor that only affects the vapor clients.
Once this is addressed I will continue with the code review.

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
aymanbagabas: Yes, it's only related to naming. However, it doesn't make sense to forcefully append "API" to the project name. The user should have a choice of having that in the name. Simply when desired, the user can append "API" to the project name and it would give the old behavior back.

I'll make a separate PR for that one. Should I base it on top of the 5.2.x branch?

4brunu: Honestly I don't think this is an option worth to create.
It will add extra complexity and we won't add nothing new or extra functionality to the project.

4brunu: But that would break every project that uses OpenAPI Swift Generator with compilation errors.

aymanbagabas: Why? The only change would be to append "API" to the project name of each project that uses the OpenAPI Swift Generator

4brunu: That has other side effects like changing the output directory name

aymanbagabas: Right, but for a generated library, I would rather have the project name matches the library class name since there is only one class. This also matches the [SPM file structure](https://swift.org/package-manager/) for example the [`PlayingCard`](https://github.com/apple/example-package-playingcard) library
```
example-package-playingcard
├── Sources
│   └── PlayingCard
│       ├── PlayingCard.swift
│       ├── Rank.swift
│       └── Suit.swift
└── Package.swift
```

4brunu: I see your point.
My main issue here is creating an extra burden on the user side.
To try to make this transition as smooth as possible, we could do something like this.

```
@available(*, deprecated, renamed: "PlayingCard")
public typealias PlayingCardAPI = PlayingCard

open class PlayingCard {
    ...
}
```

This way the old name PlayingCardAPI would still work, and the user could migrate when he wanted it.
Plus the IDE would presente a warning in a quick fix.

aymanbagabas: Is that a yes? Should I create a PR for this?

4brunu: You last comment made sense to me, so for me it's a yes.
Could you please open a PR?

aymanbagabas: @4brunu I'm also thinking about doing a similar thing for `Configuration.mustache` and `model.mustache` when using the `swiftUseApiNamespace` option. I believe when users enable `swiftUseApiNamespace`, the Configuration class and models _should_ also be under the API namespace like what `api.mustache` has. Because it doesn't make sense to do so for the API classes and not for the other classes. For example:
```
@available(*, deprecated, renamed: "PlayingCard.Configuration")
public typealias Configuration = PlayingCard.Configuration

extension PlayingCard {
    open class Configuration {
        ...
    }
}
```
and 
```
@available(*, deprecated, renamed: "PlayingCard.Rank")
public typealias Rank = PlayingCard.Rank

extension PlayingCard {
    enum Rank: Int {}
}
```

PR #9594, opened 2021-05-26
[csharp][netcore] Do not validate enum parameters

### LINKED ISSUE #9586: [BUG][C#][csharp-netcore] maxLength + enum = generated code doesn't build
#### Bug Report Checklist

- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?

##### Description

When an object property features `maxLength` (and possibly other similar attributes?) alongside an `enum`, the generated code does not compile.
Now, I know this is kind of weird to have both enum values and maxLength defined but the spec doesn't prevent those for the reasons stated [here](https://github.com/OAI/OpenAPI-Specification/issues/2153#issuecomment-590190022).
Which means that the validators won't catch these, the generator will run fine, only at the final step of building a package the error would show up.

master @ e09409fbce

```yaml
openapi: 3.0.3
info:
  title: C# enum length repro
  version: v1
paths:
  '/hello':
    get:
      responses:
        '200':
          description: returns a greeting
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Greeting'

components:
  schemas:
    Greeting:
      type: object
      properties:
        reply:
          type: string
          maxLength: 20
          enum:
            - hi
            - howdy
```

##### Generation Details

See steps below.

##### Steps to reproduce

```
openapi-generator generate -i openapi.yaml -g csharp-netcore --library httpclient -o out/csharp_enum
cd out/csharp_enum 
dotnet build --configuration Release
```

produces:

```
(...)
out/csharp_enum/src/Org.OpenAPITools/Model/Greeting.cs(141,49): error CS1061: 'Greeting.ReplyEnum?' does not contain a definition for 'Length' and no accessible extension method 'Length' accepting a first argument of type 'Greeting.ReplyEnum?' could be found (are you missing a using directive or an assembly reference?)
(...)

Build FAILED.
```

##### Related issues/PRs

None that I could find.

##### Suggest a fix

No idea... 🙇 

@mandrean @frankyjuang @shibayan @Blackclaws @lucamazzanti @wing328 @prajon84

Blackclaws:
This fixes #9586 . We do not wish to validate enum type parameters further as they are already constrained by their possible values so any validation is superfluous.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@mandrean (2017/08) @frankyjuang (2019/09) @shibayan (2020/02) @Blackclaws (2021/03) @lucamazzanti (2021/05)

### DIFF
--- a/modules/openapi-generator/src/main/resources/csharp-netcore/modelGeneric.mustache
+++ b/modules/openapi-generator/src/main/resources/csharp-netcore/modelGeneric.mustache
@@ -300,6 +300,7 @@
             {{/parent}}
             {{#vars}}
             {{#hasValidation}}
+            {{^isEnum}}
             {{#maxLength}}
             // {{{name}}} ({{{dataType}}}) maxLength
             if(this.{{{name}}} != null && this.{{{name}}}.Length > {{maxLength}})
@@ -343,6 +344,7 @@
 
             {{/isByteArray}}
             {{/pattern}}
+            {{/isEnum}}
             {{/hasValidation}}
             {{/vars}}
             yield break;


modules/openapi-generator/src/main/resources/csharp-netcore/modelGeneric.mustache
```diff
@@ -300,6 +300,7 @@
             {{/parent}}
             {{#vars}}
             {{#hasValidation}}
+            {{^isEnum}}
             {{#maxLength}}
             // {{{name}}} ({{{dataType}}}) maxLength
             if(this.{{{name}}} != null && this.{{{name}}}.Length > {{maxLength}})
```
prajon84, line 306: Instead of not validating for enum params. What about if:
`this.{{{name}}}.ToString().Length > {{maxLength}}` so that it will work for other case like:

```
	uuid_property:
      description: This is description
      type: string
      format: uuid
      maxLength: 36
```

Else, it will also break at: `UuidProperty.Length > 36`

jfeltesse-mdsol: I'm not sure I understand what you mean, are you saying strings with `format: uuid` are not String objects and thus need the `toString()` call? Looks like a separate issue if that's what you meant.

Here the fix simply skips adding validation checks if the property specifies an enum, which is a very reliable and future-proof way, me thinks.

Blackclaws: Strings with format: uuid are of Guid type in c#. The comment is valid however and worth some thought. If there is a simple way to keep validation and still have the enum behavior that might be worth exploring. However having validation for enum parameters would mean that they might break at some point in the future when one decides to use the forbidden values for the first time instead of breaking immediately.

prajon84: Yes. I meant to say is `Strings with format: uuid are of Guid type in c#`.
And, as the above validations like `MaxLength`/ `MinLength` are specific to `string` type. Its better to convert to string type using `ToString()` before validating the property.
Similar to that I found in `ruby-client` -> https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/ruby-client/partial_model_generic.mustache#L302 ( `to_s` converts to `string` IIUC)

Besides, it might be separate issue but the gist of the bug is same.

Blackclaws: Guids really don't have the same issue however. Reason being that it performs its own validation. Of course the same applies you could technically have the validation parameters in the spec I guess. It just makes no sense whatsoever.

PR #9625, opened 2021-05-30
Swift5 Vapor 4 client library

aymanbagabas:
This PR includes all commits from https://github.com/OpenAPITools/openapi-generator/pull/9624 and https://github.com/OpenAPITools/openapi-generator/pull/9608

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


4brunu:
Before review this PR, let's wait for https://github.com/OpenAPITools/openapi-generator/pull/9624 to get merged.

4brunu:
Hey, I opened a PR with some spacing and CI improvements here https://github.com/aymanbagabas/openapi-generator/pull/17.

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/Extensions.swift
```diff
@@ -0,0 +1,134 @@
...
+        } catch {
+            return request.eventLoop.makeFailedFuture(error)
+        }
+    }
+}
+
+extension Set: Content where Element: Content { }
+
+extension AnyCodable: Content {}
+extension AnyEncodable: Content {}
```
4brunu, line 133: This line is creating the following build error
`Extension outside of file declaring struct 'AnyEncodable' prevents automatic synthesis of 'init(from:)' for protocol 'Decodable'`

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/Extensions.swift
```diff
@@ -0,0 +1,134 @@
...
+extension AnyDecodable: Content {}
```
4brunu, line 134: This line is creating the following build error
`Extension outside of file declaring struct 'AnyDecodable' prevents automatic synthesis of 'encode(to:)' for protocol 'Encodable'`

4brunu (changes requested):
@aymanbagabas could you please take a look at those two build errors? Thanks

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/Extensions.swift
aymanbagabas: I think they should be removed

4brunu:
@aymanbagabas thanks for all the work until now, and for extracting a lot of PRs from the original one on vapor.
I don't see nothing wrong in this PR.
The api file with vapor is a lot different from the others, but I think that is fine, since it's a totally different implementation.

One question that I would like to have your opinion on is, you choose an enum to represent the http status code, and I was thinking in suggesting to use a just one response object with an int for http status code.
But I also understand that with an enum to represent the http status code, you have a finite possibilities on the response.
So, unless you think that having only one response object with an int for http status code it's better than the enum, I think this PR is ready to merge 👍

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/APIs/StoreAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+            try request.content.encode(body, using: Configuration.contentConfiguration.requireEncoder(for: Order.defaultContentType))
+            
+            try beforeSend(&request)
+        }
+    }
+
+    public enum PlaceOrder {
+        case http200(value: Order, raw: ClientResponse)
+        case http400(value: Void, raw: ClientResponse)
+        case http0(value: Order, raw: ClientResponse)
```
4brunu, line 206: @aymanbagabas I think at least the default case, should have an int with the status code, because it can be something that may be important for debug proposes like 401 or 403 for example.
And Maybe instead of `http0` it could be `httpUnknown` to be more clearer?

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/APIs/StoreAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+            try Configuration.apiWrapper(&request)
+            
+            
+            try request.content.encode(body, using: Configuration.contentConfiguration.requireEncoder(for: Order.defaultContentType))
+            
+            try beforeSend(&request)
+        }
+    }
+
+    public enum PlaceOrder {
```
4brunu, line 203: @aymanbagabas looking at this enum I'm not sure if I feel confident about it.
I see the reason on why you did it this way, but I'm afraid if this will be a restrictive API to use in the future, and maybe a class response would be better.
Looking at this it is not obvious to someone that doesn't know the http status code if it succeeded or not.
With the other libraries, it is obvious if the request was successful or not.

4brunu (commented):
I left some questions that I would like to discuss with you

samples/client/petstore/swift5/vaporLibrary/Sources/PetstoreClient/APIs/StoreAPI.swift
aymanbagabas: [`ClientResponse`](https://api.vapor.codes/vapor/main/Vapor/ClientResponse/#clientresponse.status) type has a `.status` property that has the response code. I agree, `httpUnknown` is more clearer. Also, I think that case should optionally decode the response since we don't really know what the response would look like.

4brunu: You are right, I haven't noticed that ClientResponse has http status code, headers, etc, that's nice 👍

aymanbagabas: yeah, Vapor is a really cool framework, it forces you to write beautiful code. You should try it sometime. There's [vapor-server-codegen](https://github.com/thecheatah/vapor-server-codegen), I think bringing that to the OpenAPI generator would be a great idea.

4brunu: When you have time, please try to change `http0` to `httpUnknown` and check the `value` Type in each enum case, to see what makes more sense, for us to merge this 🙂

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -275,7 +275,7 @@ extension {{projectName}} {
         case http{{code}}(value: {{#dataType}}{{{dataType}}}{{/dataType}}{{^dataType}}Void{{/dataType}}, raw: ClientResponse)
```
4brunu, line 275: @aymanbagabas I think only the status code should have 200 should have something like this
If it's 400 we don't know the response type.
```
case http200(value: Double, raw: ClientResponse)
case http400(value: ????, raw: ClientResponse)
```

aymanbagabas: Per the [specs](https://swagger.io/docs/specification/describing-responses/#default), you can absolutely define responses for any status code. Maybe I should remove the value and only return `ClientResponse` when the response is empty instead of returning `Void` for the value.

aymanbagabas: That's how it is right now, when the response in the specs is empty, it returns `Void` for the value. As long as the specs contain a response schema, it would have that in the value. I will change it to `ClientResponse` only when the response is empty aka no `dataType` for the response.

4brunu: I think now it put's the success response in all status code's, or am I wrong?

aymanbagabas: No, the Petstore specs are different, for examples `2_0/petstore.yaml` doesn't have a 200 response schema on POST `/pet` while the `3_0` version does. Thus, the generated code for `2_0` doesn't have a 200 response enum case for POST `/pet` aka `addPet` operation.

4brunu:
CI is failing because there are some stuff that are outdated.
Can you please run the following commands and commit the changes?
```
./mvnw clean package 
./bin/generate-samples.sh
```
Thanks

aymanbagabas:
looks like something is causing the `csharp` generator to fail

4brunu:
@aymanbagabas Thanks for your effort in contributing with the vapor integration, and also in splitting the initial big PR into smaller ones that made multiple improvements to the Swift generator 👍

Issue #9305, opened 2021-04-21
[BUG] Description


dinobal:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

- Is there a way to not have a wrapper object as an input parameter and instead just take an union type?
- The return types are correct but it seems not to be handled correctly, a bug perhaps?
- If everything works as intended how can I change my schema so the generated client has a method that can parse the response and type it correctly?

##### openapi-generator version

5.1.0

##### OpenAPI declaration file content or url

**_specs.yaml_**
```yaml
openapi: 3.0.1
info:
  title: Sample API
  description: Optional multiline or single-line description in [CommonMark](http://commonmark.org/help/) or HTML.
  version: 0.1.9
servers:
  - url: http://api.example.com/v1
    description: Optional server description, e.g. Main (production) server
  - url: http://staging-api.example.com
    description: Optional server description, e.g. Internal staging server for testing
paths:
  /pets:
    post:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
              discriminator:
                propertyName: pet_type
                mapping:
                  obj1: '#/components/schemas/Cat'
                  obj2: '#/components/schemas/Dog'
      responses:
        '200':
          description: Created
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Cat'
                  - $ref: '#/components/schemas/Dog'
                discriminator:
                  propertyName: pet_type
                  mapping:
                    obj1: '#/components/schemas/Cat'
                    obj2: '#/components/schemas/Dog'
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
      discriminator:
        propertyName: pet_type
    Dog:
      type: object
      allOf:
[...]

Issue #5335, opened 2020-02-16
[REQ][typescript] Use more sophisticated yaml spec for typescript generators samples


amakhrov:
Currently samples for all typescript generators are built based on petstore.yaml (2_0 and 3_0).
This spec doesn't include features like "date" format, allOf / oneOf, discriminators, enum aliases (for OAS3).

As a result, issues related to those features are not obvious, and also it's not obvious how fixes addressing those issues impact the end result.

A number of other generators already use a more advanced test spec for that: petstore-with-fake-endpoints-models-for-testing

Suggestion is to change all existing typescript samples from petstore to petstore-with-fake-endpoints-models-for-testing

PR #9628, opened 2021-05-31
[Typescript] Fix  oneOf

ksvirkou-hubspot:
https://github.com/OpenAPITools/openapi-generator/issues/9305

### Doesn’t work "oneOf" 

```
"items":{
   "oneOf":[
      {
         "$ref":"#/components/schemas/SingleFieldDependency"
      },
      {
         "$ref":"#/components/schemas/ConditionalSingleFieldDependency"
      }
   ]
}
```

```
import { OneOfSingleFieldDependencyConditionalSingleFieldDependency } from './OneOfSingleFieldDependencyConditionalSingleFieldDependency';

...

'inputFieldDependencies'?: Array<OneOfSingleFieldDependencyConditionalSingleFieldDependency>;
```

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
@@ -147,6 +147,7 @@ public class TypeScriptClientCodegen extends DefaultCodegen implements CodegenCo
         typeMapping.put("object", "any");
         typeMapping.put("integer", "number");
         typeMapping.put("Map", "any");
+        typeMapping.put("map", "any");
         typeMapping.put("date", "string");
         typeMapping.put("DateTime", "Date");
         typeMapping.put("binary", "any");
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
@@ -19,6 +19,7 @@ package org.openapitools.codegen.languages;
 
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.parameters.Parameter;
 import org.apache.commons.lang3.StringUtils;
@@ -32,6 +33,8 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.text.SimpleDateFormat;
 import java.util.*;
+import java.util.stream.Collectors;
+
 import static org.openapitools.codegen.utils.StringUtils.camelize;
 import static org.openapitools.codegen.utils.StringUtils.underscore;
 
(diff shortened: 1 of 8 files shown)


ksvirkou-hubspot:
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov

ksvirkou-hubspot:
Any updates on this?
@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
```diff
@@ -144,10 +147,12 @@ public TypeScriptClientCodegen() {
         typeMapping.put("object", "any");
         typeMapping.put("integer", "number");
         typeMapping.put("Map", "any");
+        typeMapping.put("map", "any");
```
amakhrov, line 150: why do we need this?

amakhrov (commented):
Thanks for your contribution! The fix looks good to me - as far as I can see what it does.
Could you please add unit tests, both to demonstrate the fix working and to prevent regressions in future?

And if you even feel like going an extra mile here - it would be super beneficial (and highly appreciated! :) ) to add a sample with oneOf/allOf (looks like no existing typescript samples use a spec with this feature - https://github.com/OpenAPITools/openapi-generator/issues/5335)

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java
ksvirkou-hubspot: > Thanks for your contribution! The fix looks good to me - as far as I can see what it does.
> Could you please add unit tests, both to demonstrate the fix working and to prevent regressions in future?
> 
> And if you even feel like going an extra mile here - it would be super beneficial (and highly appreciated! :) ) to add a sample with oneOf/allOf (looks like no existing typescript samples use a spec with this feature - #5335)

Hi 
there is only oneOf but after this fix It would be much easy to add allOf 
I don’t know about unit tests but I'll try to look

ksvirkou-hubspot: I added
```typeMapping.put("map", "any"); ```
because I saw it in codegen
``` import { ModelMap } from './ModelMap';```
it was adding to every model 
I researched how it was fixed at typescript-node
and done the same

Comment (ksvirkou-hubspot) [reply modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java:150]: https://github.com/OpenAPITools/openapi-generator/blob/46f8a6733a652544a2914b6bd6dcabf09e49b78c/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java#L163-L164

Comment (ksvirkou-hubspot) [reply modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptClientCodegen.java:150]: @amakhrov ^^

**Comment by ksvirkou-hubspot** [comment]
Hi 
I 've just committed  unit tests
@amakhrov  @macjohnny

**Comment by TiFu** [comment]
I have spent some time thinking about how to "properly" implement oneOf/anyOf/allOf in the context of the generator. 

I believe the current solution is not sufficient, due to 2 reasons
- Implementing `anyOf` with a union type is incorrect, see the below example which to my understanding would emit `AnyOfPet = PetByAge | PetByType` instead of allowing `AnyOfPet = PetByAge | PetByType | (PetByAge & PetByType)`
- ObjectSerializer is not able to correctly serialize/deserialize is not able to properly deserialize/serialize union types (as far as I am aware) - i.e., something like "Array<A | B | C>" will not be correctly parsed (currently: No further serialization/deserialization is done)

Fixing this will be quite a bit of effort, as we need to
- [Easy] Make ObjectSerializer "understand" union types (both `|`, `&`)
- Add an algorithm to properly check `anyOf` (i.e., it can be A & B & C, or A&B or A&C but not A & (half of) B. One approach could be a) For an object `o`, check for all types `t` in `anyOfTypes`,  whether `o` has all (required)  properties of  `t`. Then, check whether all properties in `o` are covered by the `t`'s that matched `o`, and that there are no conflicts 
 
```
        AnyOfPet:
          anyOf:
            - $ref: '#/components/schemas/PetByAge'
            - $ref: '#/components/schemas/PetByType'

        PetByAge:
          type: object
          properties: 
            age: 
              type: integer
            nickname: 
              type: string
          required:
            - age              

        PetByType:
          type: object
          properties:
            pet_type:
              type: string
              enum: [Cat, Dog]
            hunts:
              type: boolean
          required: 
            - pet_type
```
as

amakhrov:
Frankly, accurately representing `anyOf` or non-discriminated `oneOf` in TS (or in general, I guess) is so cumbersome that I feel like it's not worth efforts. For starters, you're doomed to use type assertions every time you work with such a combined type, since TS has no way to know which combination of A/B/C is matched in a particular instance of our (anyOf A/B/C). Or you have to generate a bunch of type guards (one for each of A, B, C)

Also for the purpose of (de)serialization, we don't have to be type strict. Whatever properties are present in the input, we deserialize then to a corresponding type, assuming there is no conflicting props between A/B/C. And it's a job for the validator to ensure no conflicts, not a job for the serializer. Imo :)

ksvirkou-hubspot:
Hi @TiFu 
I agree with @amakhrov 
Can we merge it ?
it doesn’t work now
it would be great  to merge it 
Because for us it is blocker

TiFu:
I agree with @amakhrov :-) 

Both CI failures are unrelated (1 due to pull rate limit, 1 due to CSharp Generator Tests).

TiFu:
Alright. Merging this :-). Thank you for the PR!

- Appveyor fails due to an issue in the csharp samples.
- Travis fails due to request limit

PR #9653, opened 2021-06-02
Avoid using Properties.clone() which is broken in Gradle

### LINKED ISSUE #7601: [BUG] Use of GlobalSettings in gradle-plugin leads to race conditions in Gradle 6.6+
##### Description

On the latest version of Gradle 6.6 and latest generator 4.3.1 I am seeing inconsistent generation of outputs, some projects where model generation is disabled, intermittently models appear and the next run they do not appear. I believe this is due to running my multi-module gradle project with a number of openapi generator tasks running at once, the current problem is with ~5 concurrent generations running.

It appears that there is a race condition in the GeneratorTask and the use of a global singleton GlobalSettings:

https://github.com/OpenAPITools/openapi-generator/blob/61b543f03dc546386467c9a7afe8864f6f609ed1/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt#L513

GlobalSettings appears to defend against this via a ThreadLocal map which seems to work in Gradle 6.5, but fail in Gradle 6.6+. I'm not sure whats changed in Gradle land but it seems like any settings configured via GlobalSettings conflict/collide with each other when running concurrently, if I set `--max-workers 1` everything works as expected, if workers is > 1 then the problems start.

4.3.1

n/a

##### Generation Details

n/a

##### Steps to reproduce

n/a

##### Related issues/PRs

None that I can see 

##### Suggest a fix

Not sure, it seems like something in upstream Gradle has changed in the 6.5 -> 6.6 upgrade, I was able to work around it by moving from configuration that ended up in GlobalSettings and re-implemented it in ignore files.

chschu:
This PR fixes #7601 by working around https://github.com/gradle/gradle/issues/17344. It does not change generator output.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/GlobalSettings.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/GlobalSettings.java
@@ -34,7 +34,10 @@ public class GlobalSettings {
     private static ThreadLocal<Properties> properties = new InheritableThreadLocal<Properties>() {
         @Override
         protected Properties initialValue() {
-            return (Properties) System.getProperties().clone();
+            // avoid using System.getProperties().clone() which is broken in Gradle - see https://github.com/gradle/gradle/issues/17344
+            Properties copy = new Properties();
+            copy.putAll(System.getProperties());
+            return copy;
         };
     };
 


wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/config/GlobalSettings.java
```diff
@@ -34,7 +34,9 @@
     private static ThreadLocal<Properties> properties = new InheritableThreadLocal<Properties>() {
         @Override
         protected Properties initialValue() {
-            return (Properties) System.getProperties().clone();
+            Properties copy = new Properties();
```
wing328, line 37: Thanks for the PR. Can you please add a 1-liner comment explaining why and mentioning https://github.com/gradle/gradle/issues/17344 ?

chschu: Done. Thanks for reviewing!

PR #9656, opened 2021-06-02
[kotlin][client] silence deprecation warning

4brunu:
Silence deprecated warning when used internally

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
@@ -41,6 +41,9 @@ import {{packageName}}.infrastructure.toMultiValue
     @Deprecated(message = "This operation is deprecated.")
     {{/isDeprecated}}
     {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}?{{/required}}{{^-last}}, {{/-last}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
+        {{#isDeprecated}}
+        @Suppress("DEPRECATION")
+        {{/isDeprecated}}
         val localVariableConfig = {{operationId}}RequestConfig({{#allParams}}{{{paramName}}} = {{{paramName}}}{{^-last}}, {{/-last}}{{/allParams}})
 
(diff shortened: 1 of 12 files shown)


modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/api.mustache
```diff
@@ -41,6 +41,9 @@ import {{packageName}}.infrastructure.toMultiValue
     @Deprecated(message = "This operation is deprecated.")
     {{/isDeprecated}}
     {{^doNotUseRxAndCoroutines}}{{#useCoroutines}}suspend {{/useCoroutines}}{{/doNotUseRxAndCoroutines}}fun {{operationId}}({{#allParams}}{{{paramName}}}: {{{dataType}}}{{^required}}?{{/required}}{{^-last}}, {{/-last}}{{/allParams}}) : {{#returnType}}{{{returnType}}}{{#nullableReturnType}}?{{/nullableReturnType}}{{/returnType}}{{^returnType}}Unit{{/returnType}} {
+        {{#isDeprecated}}
+        @Suppress("DEPRECATION")
+        {{/isDeprecated}}
```
wing328, line 46: For the warning, do you mean `This operation is deprecated` (line 41) ?

4brunu: Yes.
Basically in the kotlin generator we have two methods per endpoint.
One that returns the endpoint configuration.
And another that call's the first method to get the endpoint configuration, does the server call and returns the server request.

If this operation is deprecated, both methods will be deprecated.
And when the second method gets the endpoint configuration will show a deprecation warning, but this is just noise for the users of this library.

I'm not sure if I explained it well or not. :)

4brunu: Here are some screenshots to show the issue.

First of all, this change will not impact the users in any way, it's only an internal change.
When the users call a deprecated api, they will still see the deprecated warning, both in the method call and in the request config method.

Screenshot 2021-06-08 at 09 02 30

The issue that I found and that this PR fixes, is that there are deprecated warnings on the generated code, that don't add any value, and only create noise.


Basically for each api method's, there are two generated methods on the kotlin client.

One that holds the request configurations and only returns it, but it doesn't perform the server request.
Screenshot 2021-06-08 at 09 01 24

And another one that actually makes the server request, and get's the configuration of this api method in the previous method.
Screenshot 2021-06-08 at 09 01 18

When an api method is deprecated, both of this methods will be marked as deprecated.

So when the second method that actually makes the server request get's the configuration request, it will show a warning.
But this will only create noise to the user, because this method call is right, and there is nothing to fix here.

That's why I introduced this change.

Screenshot 2021-06-08 at 09 02 53

wing328: Thanks for the detailed explanation

PR #9661, opened 2021-06-02
[csharp] Move nullable reference types switch to abstract

devhl-labs:
Moving the nullable reference types switch from aspnet generator to abstract so all csharp can use it.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@mandrean @frankyjuang @shibayan @Blackclaws @wing328


wing328:
@devhl-labs thanks for the PR as always.

Do we also need to update the mustache templates to actually use the option (the mustache tag)?

In other words, if we move the option to the C# abstract class, we need to make sure all C# generators that inherit the abstract generator properly support the option. Otherwise, we can only add the option to individual generators to start with.

devhl-labs:
Yes, it would have to be handled by each set of template files. If we don't put it in the abstract, then we are removing the option for users to use their own custom templates to handle this option for generators that don't support it yet.

Blackclaws:
I think that, technically, there is no need to actually change any part of the templates aside from the csproj files to use this. The reason is that with nullable reference types you simply change the post process function to treat every type as nullable and voila you get a nullable reference annotation for each type that is optional

devhl-labs:
I can add it to the csproj files later today.

wing328:
> I think that, technically, there is no need to actually change any part of the templates aside from the csproj files to use this.

Yes, that what I meant, e.g. https://github.com/OpenAPITools/openapi-generator/pull/9620/files#diff-b62dd64c4e02abc058c477127e8c9c85250c66b616d6683c9ee143470bcf2cd9R10

devhl-labs:
Im tracking i just didnt get to it today.

devhl-labs:
Aspnetcore was already handled.  Skipped dotnet-core2 because it's deprecated. Skipped nancyfx because the framework itself is deprecated as far as I can tell.  Added to csharp-netcore.

wing328:
Can you please resolve the merge conflicts when you've time?

devhl-labs:
Done

devhl-labs:
I see drone had an unrelated error. Is that why app veyor failed?

wing328:
AppVeyor tests passed via https://ci.appveyor.com/project/WilliamCheng/openapi-generator/builds/39524940

Issue #8579, opened 2021-01-29
[BUG][Python] Constructing an object with a polymorphic base class parameter causes RecursionError


bherw:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [ ] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [ ] Have you searched for related issues/PRs?
- [ ] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

Generated client code crashes with a RecursionError when trying to construct a TestRequest object with a base type parameter. The request specifies a parent type, I'm trying to set the child type as the parameter value (standard abstract base class/child class modelling).

##### openapi-generator version

5.0.0

##### OpenAPI declaration file content or url

```yaml
openapi: 3.0.0
info:
 description: The Bug schema
 title: recursion bug
 version: 0.1.11

paths: {}

components:
 schemas:
   TestRequest:
     type: object
     required:
       - data
       - parentType
     properties:
       data:
         type: string
       parentType:
         $ref: '#/components/schemas/ParentType'
   ParentType:
     type: object
     required:
       - objectType
     properties:
       objectType:
         type: string
     discriminator:
       propertyName: objectType
       mapping:
         child1: '#/components/schemas/ChildType1'
         child2: '#/components/schemas/ChildType2'
   ChildType1:
     allOf:
       - $ref: '#/components/schemas/ParentType'
     properties:
       objectType:
         type: string
         default: child1
       attribute1:
         type: string
         format: byte
   ChildType2:
     allOf:
       - $ref: '#/components/schemas/ParentType'
     type: object
     properties:
       objectType:
         type: string
         default: child2
```

##### Generation Details

```sh
java -jar /home/bherweyer/dev/spa/spa-openapi/openapi-generator-cli-5.0.0.jar  generate -i /home/bherweyer/dev/spa/spa-openapi/specification/bug.yaml -g python -o /home/bherweyer/dev/spa/spa-openapi/target/client/python/bug --package-name bug

```

##### Steps to reproduce

```python
[...]

Issue #9684, opened 2021-06-05
[BUG][python] allOf with subobjects generates "Invalid inputs"


mfmarche:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [x] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

The current testcases for allOf in petstore seem to work on the assumption that passing model arguments in __init__() can be passed to all objects in the allOf.

In particular, the method "validate_get_composed_info" description says:

    Exceptions are raised if:
    - 0 or > 1 oneOf schema matches the model_args input data
    - no anyOf schema matches the model_args input data
    - any of the allOf schemas do not match the model_args input data

This seems to suggest that I cannot instantiate a composedObject (via its constructor), with arguments that are associated with a particular object in the oneOf list.

I made a very simple change to the petstore api model spec to highlight this:

```
+    SpecialAttributes:
+      type: string
+      minLength: 1
+      maxLength: 100
     ShapeInterface:
       properties:
         shapeType:
           type: string
+        name:
+          $ref: '#/components/schemas/SpecialAttributes'
       required:
         - shapeType
```

Adding a name to ShapeInterface, with min/max arguments now makes it ModelNormal, (rather than a simple string).

instantiating this:
```
equilateral_triangle.EquilateralTriangle(triangle_type="EquilateralTriangle", shape_type="Triangle", name=SpecialAttributes("aname"))
```

gives the following error:
```
petstore_api.exceptions.ApiValueError: Invalid inputs given to generate an instance of 'TriangleInterface'. The input data was invalid for the allOf schema 'TriangleInterface' in the composed schema 'EquilateralTriangle'. Error=Invalid type for variable 'name'. Required value type is one of [NoneType, bool, date, datetime, dict, float, int, list, str] and passed type was SpecialAttributes at ['name']
```

[...]

PR #9715, opened 2021-06-07
[python] fix: 8579

mfmarche:
Updates the python generator to fix a recursion issue when checking discriminators
Issue 8579 is where recursive models would cause an endless loop. Resolve this by keeping tracking of discriminator
models that have been seen.

Also updated petstore openapi with a new property that would previously fail without this change.

fixes:
- https://github.com/OpenAPITools/openapi-generator/issues/8579

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


samples/client/petstore/python/petstore_api/model_utils.py
```diff
@@ -596,8 +596,17 @@ def __setitem__(self, name, value):
                 )
         # attribute must be set on self and composed instances
         self.set_attribute(name, value)
+        found_intersection = False
         for model_instance in self._composed_instances:
-            setattr(model_instance, name, value)
+            intersection = find_args_intersection(model_instance.__class__, {name: value})
```
spacether, line 601: Intersections are allowed, why is this code choosing not to assign the value to the composed model instance?

samples/client/petstore/python/petstore_api/model_utils.py
```diff
@@ -596,8 +596,17 @@ def __setitem__(self, name, value):
...
+            if len(intersection):
+                setattr(model_instance, name, value)
+                found_intersection = True
+        if not found_intersection:
+            # we didn't find an intersection for the property, it could be an
+            # additional type, so pass it through
+            for model_instance in self._composed_instances:
+                setattr(model_instance, name, value)
```
spacether, line 609: Our current code has the same properties applied to all model instances which is consistent with openapi schemas. Why are you only assigning properties here if there are no intersections?

samples/client/petstore/python/petstore_api/model_utils.py
mfmarche: that was the orignal issue. Intersections only work when they are a type that is allowed. What happens when you have a SimpleModel defined in on of the allOf models, it throws since its not an allowed type.

samples/client/petstore/python/petstore_api/model_utils.py
```diff
@@ -1747,7 +1771,8 @@ def get_allof_instances(self, model_args, constant_args):
     for allof_class in self._composed_schemas['allOf']:
 
         try:
-            allof_instance = allof_class(**model_args, **constant_args)
+            intersected_args = find_args_intersection(allof_class, model_args)
+            allof_instance = allof_class(**intersected_args, **constant_args)
```
spacether, line 1775: Every schema/class instance should ingest all arguments. Not the subset of intersected ones.
This is required by openapi. Why are you passing in the intersection only?

samples/client/petstore/python/petstore_api/model_utils.py
spacether: Can you give an example of a schema that has allof and that SimpleModel in it? It's hard to understand what yuo mean without real examples.

mfmarche: Hi, thanks for your comments.

I have updated the petshop openapi spec with an example exactly for testing this use case.

see the updated openapi spec:
https://github.com/OpenAPITools/openapi-generator/pull/9715/files#diff-a402132df00046e31cfa3d1521283e26d2f3df03ea998809d4d65fa047ee7cedR2309

I specifically added ShapeName, with min/max lengths. With the min and max lengths, this ends up creating a SimpleModel, and generated here: https://github.com/OpenAPITools/openapi-generator/pull/9715/files#diff-be491650dcc0d4230a1f7b60e5eacf92b36a14630614b718a2f4f93ec0e5e664

I added a test here:
https://github.com/OpenAPITools/openapi-generator/pull/9715/files#diff-0fdf80691bfcf48b97e56ab4bbd24139a7ab2571f2419cba5de94524756161b8R54

When the EquilateralTriangle is instantiatied, it will walk through the allOf instances. What would have previously occured, is the the model ShapeName would be passed to each of the allOf instances (ShapeInterface and TriangleInterface). The ShapeName would not be accepted in TriangleInterface (since its not part of that model spec), and it would raise an exception.

WIthout the changes for the intersection, this test would fail, complaining that the type is not supported. This is example is also outlined in the issue I raised here: https://github.com/OpenAPITools/openapi-generator/issues/9684

samples/client/petstore/python/petstore_api/model_utils.py
mfmarche: This could be a misunderstanding of how a constructor for a composed object is supposed to work. If the composed schema can handle all the arguments (that could be handled by separate allOf instances), its possible that some of the allOf structures cannot handle them all. As per my pervious comment above, how can the TriangleInterface handle a ModelSimple class it knows nothing about (eg. the ShapeName example I gave)?

spacether: Because ModelSimple classes types must be seen as the underlying base type str/int/float etc.
TriangleInterface should accept name/ShapeName as a str even though it is a ModelSimple. 
I left a comment about that on https://github.com/OpenAPITools/openapi-generator/issues/9684

mfmarche: After reviewing the discussion in #9684 and the requirement that all properties must be applied to all (allOf) instances, it occured to me that the configuration property "discard_unknown_keys" will satisfy the requirement where certain types are not accepted by the allOf instances. I removed the find_args_intersection code, but kept the original test that now applies this configuration.

spacether: Discard unknown keys may work for you. One other use case to be aware of is that and of these composed schemas A through E may have minproperties or maxProperties validations. Those validations require that a dict have a certain number of keys. Which again points us to each schema ingesting all payload params. In my opinionthe best way toffix this is to fix the get_simple class logic.

mfmarche: would it suffice to leave this PR as is, and introduce that as another fix to resolve the general use case?

spacether: This PR has done a great job identifying problems and working to solve them. For the reasons that I laid out above I don't want to merge the intersection code. 
How about separate prs to solve the discriminator issue and the ingesting ModelSimple issue with additionalProperties?
this could becomeoone ffo thosePRs.

mfmarche: Thanks for the feedback. I removed the intersection code with the recent force push. The only thing this PR does is now solve the recursion issue (8579). It doesn't truly solve #9684, as per your comments. However, I have made no changes with regard to that, other than add a testcase that shows the behavior of using the discard to achieve this one use case.

modules/openapi-generator/src/main/resources/python/model_utils.mustache
```diff
@@ -799,18 +799,19 @@ def remove_uncoercible(required_types_classes, current_item, spec_property_namin
             results_classes.append(required_type_class)
     return results_classes
 
-def get_discriminated_classes(cls):
+def get_discriminated_classes(cls, seen_discriminator, possible_classes):
```
spacether, line 802: Can we remove possible classes as an input? Why are we routing it in here? Isn't possible_classes only in the context of the current schema that we are looking at?

spacether:
Just updated the description to remove:
> When instantiating allOf objects, the constructor arguments need to be
filtered according to the openapi_types that are supported.

Because it's not what openpi directs us to doper our discussion. All matching schemas must ingest the send payload assuming that we are talking about the root schema and not properties/items/additionalProperties.

spacether:
@gavinmh sure thing. Can you update the PR to eliminate the possible_classes input argument?
The fn already returns possible_classes so we should be able to recursively call it passing in only cls and seen_discriminator which should maybe be named first_seen_discriminator_cls.
If you can't edit the PR, can you clone it and continue work there?

PR #9724, opened 2021-06-08
operationId: Allow getting only the method name (#9719)

### LINKED ISSUE #9719: [REQ] operationId: Allow getting only the method name
I have an OpenAPI server based on Python Connexion (https://github.com/zalando/connexion). Since this is a Python module the operationId fields take the form of "_package_name.api.module_name.method_". For example, the operationId for the "/users" GET operation will be "my_server.api.users.get". 
As a result, the Go code generator will create an service called UsersApiService that will have a method called MyServerApiUsersGet. To me, this name has a lot of unneeded sections. The Python package name is obviously not needed, the word Api is not needed and also the Service name is not needed since the method is already inside a service with that name. So, instead, I would like to have the method be just "get".
Please do not focus on this specific "GET" operation, the naming could be anything.

Today, the code generator allows removing the prefix, but it assumes the delimiter is an underscore and takes everything after the first one. In my case, the delimiter is a period and I would like to get everything after the last one.

So, I would like to suggest broadening this option by adding two configuration fields:
1. Delimiter - allowing the user to choose which delimiter to use. The default should be underscore
2. Count - The X occurrence of the delimiter that constitutes a prefix. The default should be 1 and -1 should be used for last

I saw your comment about using templates. But, since this change relates to all languages, I think my suggestion is the way to go

ygalblum:
This PR adds the functionality I requested in #9719 

Add new configuration flags: REMOVE_OPERATION_ID_PREFIX_DELIMITER and REMOVE_OPERATION_ID_PREFIX_COUNT
Add the corresponding fields in the DefaultCodegen
Sanitize the operationId according to the new fields
Add delimiter to list of nonNameElementPattern
Add unit test for different cases (different delimiters and counts)

### PR checklist

  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

fixes #9719

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -177,6 +177,8 @@ public class DefaultCodegen implements CodegenConfig {
     protected List<CliOption> cliOptions = new ArrayList<CliOption>();
     protected boolean skipOverwrite;
     protected boolean removeOperationIdPrefix;
+    protected String removeOperationIdPrefixDelimiter = "_";
+    protected int removeOperationIdPrefixCount = 1;
     protected boolean skipOperationExample;
 
     protected final static Pattern JSON_MIME_PATTERN = Pattern.compile("(?i)application\\/json(;.*)?");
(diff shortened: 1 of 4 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -3725,9 +3737,11 @@ public CodegenOperation fromOperation(String path,
         String operationId = getOrGenerateOperationId(operation, path, httpMethod);
         // remove prefix in operationId
         if (removeOperationIdPrefix) {
-            int offset = operationId.indexOf('_');
-            if (offset > -1) {
-                operationId = operationId.substring(offset + 1);
+            String[] componenets = operationId.split("[" + removeOperationIdPrefixDelimiter + "]");
```
wing328, line 3740: Suggestion: adding a 1-liner comment explaining what the code block does

wing328:

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
ygalblum: Done

wing328: 👍

modules/openapi-generator/src/test/java/org/openapitools/codegen/DefaultCodegenTest.java
```diff
@@ -3560,4 +3560,55 @@ public void testBooleansSetForIntSchemas() {
...
+        codegen.processOpts();
+
+        path = "/dotDelimiter";
+        operation = openAPI.getPaths().get(path).getGet();
+        co = codegen.fromOperation(path, "GET", operation, null);
+        assertEquals(co.operationId, "getAll");
+
+        codegen.additionalProperties().put(CodegenConstants.REMOVE_OPERATION_ID_PREFIX, "True");
+        codegen.additionalProperties().put(CodegenConstants.REMOVE_OPERATION_ID_PREFIX_DELIMITER, "_");
+        codegen.additionalProperties().put(CodegenConstants.REMOVE_OPERATION_ID_PREFIX_COUNT, 2);
```
wing328, line 3596: Thanks for adding the tests. Can we also add a test in which the index is out of bound (e.g. 10 in this case) ?

ygalblum: Done

wing328:
@ygalblum thanks for the PR, which looks good to me 👍

Issue #9669, opened 2021-06-03
[BUG][python] deepcopy of model gives KeyError: '_data_store'


mfmarche:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [x] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

I ran into this as I was attempting to use dataclasses.asdict(obj). I tracked it down to use of a deepcopy call. While copying the openapi model object, I believe the default deepcopy is missing the object attributes.

##### openapi-generator version

master

##### OpenAPI declaration file content or url

I used the sample petstore api in: samples/openapi3/client/petstore/python

##### Generation Details

N/A

##### Steps to reproduce

```
from copy import deepcopy
from petstore_api.model.banana import Banana
first = Banana(5.2)
second = deepcopy(first)

<snip>
KeyError: '_data_store'
```

##### Suggest a fix

I came up with a solution to this by updating the model_utils.py, specifically the OpenApiModel, adding:

```
    def __copy__(self):
        cls = self.__class__
        result = cls.__new__(cls)
        result.__dict__.update(self.__dict__)
        return result

    def __deepcopy__(self, memo):
        cls = self.__class__
        result = cls.__new__(cls)
        for k, v in self.__dict__.items():
            setattr(result, k, deepcopy(v, memo))
        return result
```

Note that this worked only for a Normal object without a discriminator. I ran into a separate issue (might be unrelated?), where the same test with a discriminator gave the following:

```
from copy import deepcopy
from petstore_api.model.mammal import Mammal
deeepcopy(Mammal())

<snip>
petstore_api.exceptions.ApiValueError: Cannot deserialize input data due to missing discriminator. The discriminator property 'className' is missing at path: ()
```

Could there be other pitfalls with implementation of the deepcopy above?

mfmarche:
[...]

Issue #9296, opened 2021-04-20
[BUG][python] Cannot instantiate class without required read-only parameters for POST endpoint


gbmarc1:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [x] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

@spacether 

Python generator

##### Description

I have a POST endpoint and the request body and response body use the same schema definition as below. `Dataset` has multiple `readOnly` parameters that must be present in the response, but should not be provided in the request.

I can instantiate the `Dataset` model providing `None` to the parameters that are readOnly and without verifying the model's input (`_check_type=False`). However, post request still fails (404 bad request) because params are still present with None values.

##### openapi-generator version

5.1.1

##### OpenAPI declaration file content or url

```yaml
paths:
  /datasets:
    post:
      operationId: v1_inference_datasets_post
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Dataset'
        description: The dataset to be registered
        required: true
      responses:
        '201':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Dataset'
          description: Confirmation of dataset registration
      summary: Register a dataset
      tags:
        - datasets
 
components:
  schemas:
    Dataset:
      additionalProperties: false
      example:
        statusLastChange: 2000-01-23T04:56:07+00:00
        modifiedOn: 2000-01-23T04:56:07+00:00
        name: name
        inactiveOn: 2000-01-23T04:56:07+00:00
        description: description
        mediaType: text/plain
        id: 0
        createdOn: 2000-01-23T04:56:07+00:00
        status: active
      properties:
        id:
          description: ID of the dataset
          readOnly: true
          type: integer
        name:
          description: User-defined name of the dataset
          maxLength: 128
          type: string
        description:
          description: User-defined description of the dataset
          maxLength: 256
          type: string
        mediaType:
[...]

PR #9735, opened 2021-06-09
fix: handle deepcopy of openapi objects

mfmarche:
- addresses issue #9669
- Add __deepcopy__ and __copy__ to OpenApiModel
- pass discriminator inside deepcopy if exists
- add test cases for deepcopy of models

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
@@ -13,3 +13,22 @@
     def __getattr__(self, attr):
         """get the value of an attribute using dot notation: `instance.attr`"""
         return self.{{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}(attr)
+
+    def __copy__(self):
+        cls = self.__class__
+        if self.get("_spec_property_naming", False):
+            return cls._new_from_openapi_data(**self.__dict__)
+        else:
+            return new_cls.__new__(cls, **self.__dict__)
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+
+        if self.get("_spec_property_naming", False):
+            new_inst = cls._new_from_openapi_data()
+        else:
+            new_inst = cls.__new__(cls)
+
+        for k, v in self.__dict__.items():
+            setattr(new_inst, k, deepcopy(v, memo))
+        return new_inst
--- a/modules/openapi-generator/src/main/resources/python/model_utils.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model_utils.mustache
@@ -1,6 +1,7 @@
 {{>partial_header}}
 
 from datetime import date, datetime  # noqa: F401
+from copy import deepcopy
 import inspect
 import io
 import os
@@ -223,8 +224,13 @@ class OpenApiModel(object):
             self_inst = super(OpenApiModel, cls).__new__(cls)
             self_inst.__init__(*args, **kwargs)
 
-        new_inst = new_cls.__new__(new_cls, *args, **kwargs)
-        new_inst.__init__(*args, **kwargs)
+        if kwargs.get("_spec_property_naming", False):
+            # when true, implies new is from deserialization
+            new_inst = new_cls._new_from_openapi_data(*args, **kwargs)
+        else:
+            new_inst = new_cls.__new__(new_cls, *args, **kwargs)
+            new_inst.__init__(*args, **kwargs)
+
         return new_inst
 
 
--- a/samples/client/petstore/python/petstore_api/model_utils.py
+++ b/samples/client/petstore/python/petstore_api/model_utils.py
@@ -9,6 +9,7 @@
 
 
 from datetime import date, datetime  # noqa: F401
+from copy import deepcopy
 import inspect
 import io
 import os
@@ -186,6 +187,26 @@ class OpenApiModel(object):
         """get the value of an attribute using dot notation: `instance.attr`"""
         return self.__getitem__(attr)
 
+    def __copy__(self):
+        cls = self.__class__
+        if self.get("_spec_property_naming", False):
+            return cls._new_from_openapi_data(**self.__dict__)
+        else:
+            return new_cls.__new__(cls, **self.__dict__)
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+
+        if self.get("_spec_property_naming", False):
+            new_inst = cls._new_from_openapi_data()
+        else:
+            new_inst = cls.__new__(cls)
+
+        for k, v in self.__dict__.items():
+            setattr(new_inst, k, deepcopy(v, memo))
+        return new_inst
+
+
     def __new__(cls, *args, **kwargs):
         # this function uses the discriminator to
         # pick a new schema/class to instantiate because a discriminator
@@ -295,8 +316,13 @@ class OpenApiModel(object):
             self_inst = super(OpenApiModel, cls).__new__(cls)
             self_inst.__init__(*args, **kwargs)
 
-        new_inst = new_cls.__new__(new_cls, *args, **kwargs)
-        new_inst.__init__(*args, **kwargs)
+        if kwargs.get("_spec_property_naming", False):
+            # when true, implies new is from deserialization
+            new_inst = new_cls._new_from_openapi_data(*args, **kwargs)
+        else:
+            new_inst = new_cls.__new__(new_cls, *args, **kwargs)
+            new_inst.__init__(*args, **kwargs)
+
         return new_inst
 
 
--- a/samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
+++ b/samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
@@ -9,6 +9,7 @@
 
 
 from datetime import date, datetime  # noqa: F401
+from copy import deepcopy
 import inspect
 import io
 import os
@@ -186,6 +187,26 @@ class OpenApiModel(object):
         """get the value of an attribute using dot notation: `instance.attr`"""
         return self.__getitem__(attr)
 
+    def __copy__(self):
+        cls = self.__class__
+        if self.get("_spec_property_naming", False):
+            return cls._new_from_openapi_data(**self.__dict__)
+        else:
+            return new_cls.__new__(cls, **self.__dict__)
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+
+        if self.get("_spec_property_naming", False):
+            new_inst = cls._new_from_openapi_data()
+        else:
+            new_inst = cls.__new__(cls)
+
+        for k, v in self.__dict__.items():
+            setattr(new_inst, k, deepcopy(v, memo))
+        return new_inst
+
+
     def __new__(cls, *args, **kwargs):
         # this function uses the discriminator to
         # pick a new schema/class to instantiate because a discriminator
@@ -295,8 +316,13 @@ class OpenApiModel(object):
             self_inst = super(OpenApiModel, cls).__new__(cls)
             self_inst.__init__(*args, **kwargs)
 
-        new_inst = new_cls.__new__(new_cls, *args, **kwargs)
-        new_inst.__init__(*args, **kwargs)
+        if kwargs.get("_spec_property_naming", False):
+            # when true, implies new is from deserialization
+            new_inst = new_cls._new_from_openapi_data(*args, **kwargs)
+        else:
+            new_inst = new_cls.__new__(new_cls, *args, **kwargs)
+            new_inst.__init__(*args, **kwargs)
+
         return new_inst
 
 
--- a/samples/openapi3/client/extensions/x-auth-id-alias/python/x_auth_id_alias/model_utils.py
+++ b/samples/openapi3/client/extensions/x-auth-id-alias/python/x_auth_id_alias/model_utils.py
@@ -9,6 +9,7 @@
 
 
 from datetime import date, datetime  # noqa: F401
+from copy import deepcopy
 import inspect
 import io
 import os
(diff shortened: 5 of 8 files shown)


mfmarche:
Also note, I believe this resolves #8579

spacether:
Closing and re-opening for travis

modules/openapi-generator/src/main/resources/python/model_templates/method_from_openapi_data_composed.mustache
```diff
@@ -48,7 +48,7 @@
             '_visited_composed_classes': self._visited_composed_classes,
         }
         composed_info = validate_get_composed_info(
-            constant_args, kwargs, self)
+            constant_args, kwargs, self, from_openapi_data=True)
```
spacether, line 51: Why are we passing in from_openapi_data=True here?
Isn't the _from_openapi_data method only used when receiving responses from the server.
When that is done _from_openapi_data is invoked which means __init__ is NOT invoked.
So this code path is for client side instantiation where from_openapi_data is False, right?

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
```diff
@@ -12,4 +12,21 @@
 
     def __getattr__(self, attr):
         """get the value of an attribute using dot notation: `instance.attr`"""
-        return self.{{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}(attr)
\ No newline at end of file
+        return self.{{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}(attr)
+
+    def __copy__(self):
+        cls = self.__class__
+        result = cls.__new__(cls)
```
spacether, line 19: Is this copying a payload instantiated from client data `cls(**kwargs)`
Or one received from the server `cls._from_openapi_data(**kwargs)`?
If so should we not use __new__ and use one of the above options?

mfmarche: that seems reasonble, ideally just _from_openapidata should be used, since we are copying the "data", and don't want the type checking (ie. checking for read-only would not be required).

spacether: Ah so then because we are option2, will be not need the changes to the __init__ signature because that's a different code path?

modules/openapi-generator/src/main/resources/python/model_templates/method_from_openapi_data_composed.mustache
mfmarche: That is a good question. I suspect the orignal intention was that the method from_openapi_data is data from server. However, it can also be used to instantiate objects where you might mocking a server (ie. generating test code, generating an object with some data that needs its read-only data set). In this particular case aboe that you raised, this is a copy. Since we know the object was rendered already (from whereever, client or server), we just need to copy it. If we used the normal __init__ method, if there is read-only data, it will raise an exception. Rather, this this case, we just need it copied, as is, without any checking. This is why the from_openapi_data is choosen.

mfmarche: Since this is the generated code for from_openapi_data, we need to let the method validate_get_composed_info know this, since it eventually calls get_allof_instances, what then may need to instantiate allOf objects.  While instantiating the allOf objects, it needs to instantiate the using the appropriate init method (__init__, or __from_openapi_data).

With regard to your comment on from client or server (and using from_openapi_data), _from_openapi_data can also be used to instantiate objects where you might mocking a server (ie. generating test code, generating an object with some data that needs its read-only data set).

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
spacether: So how about changing this to use _from_openapi_data rather than __new__?

mfmarche: agree, i will make that change.

modules/openapi-generator/src/main/resources/python/model_utils.mustache
```diff
@@ -224,7 +225,8 @@ class OpenApiModel(object):
             self_inst.__init__(*args, **kwargs)
 
         new_inst = new_cls.__new__(new_cls, *args, **kwargs)
-        new_inst.__init__(*args, **kwargs)
+        # use _from_openapi_data to handle readOnly discriminator
+        new_inst = new_inst._from_openapi_data(*args, **kwargs)
```
spacether, line 229: Once we get rid of the new invocation in the copy method can this code be reverted?
If not then I think there need to be two code paths here.
- For from_openapi_data == True, invoke _from_openapi_data
- For from_openapi_data == False invoke __new__ then __init__

mfmarche: its the latter, its still needed (the openapi_data check). Agree, it would be more accurate to selectively call the correct init call. I will make that change.

mfmarche: Since this is inside a __new__, trying to figure out if this is called from __init__ or from_openapi_data is tricky since __new__ only has a single arg (the class name).

mfmarche: It seems the mechansim of validating the read-only vars at construction time (via init vs. from_openapi_data) leaves this code in a bit of a bind.

spacether: Can we use kwargs['_spec_property_naming'] to decide which code path to use?
I think that
- _spec_property_naming = True is for the from_server use case / which is the from_openapi_data use case
- _spec_property_naming = False is the client side instantiation use case

mfmarche: thats a good idea, I hadn't realized that _spec_property_naming had that property. However a click glance through the code, I don't see _spec_property_naming ever be se to True (its default is False in the _from_openapi_data). Is there another code path that is not obvious that sets this to True?

spacether: Our code sets it to true here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/api_client.mustache#L340
when values are received from the server

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
spacether: FYI: one can use the self._spec_property_naming to determine which code path to use when copying a self instance.

modules/openapi-generator/src/main/resources/python/model_utils.mustache
mfmarche: my concern is, this spec_property_naming is set via the deserialize call.
If I instantiate an object with _from_openapi_data(), and then copy that, I think that means now that I have to manually set this parameter (_spec_property_naming).

spacether: Hmm that's true. Should all _from_openapi_data invocations hard code _spec_property_naming = True?

mfmarche: possibly, though I'm not sure. My understanding is, _from_openapi_data is similiar to __init__, so vars are "pythonic". So that means it shouldn't be _spec_property_naming. I believe this tells me that we can't use_spec_property_naming for the purpose that I would like.

I don't quite appreciate the purpose of the _from vs. init. Couldn't the same thing be achieved and keep a single constructor initializer and just an option for checking read-only? I guess that is another discussion separate from this one.

spacether: _from_openapi data is used to instantiate models with data received from the server. When making model instances with readonly parameters the fn signature needed to make a model instances must contain different parameters if the data came from the server vs from the client. You can read more about it here: https://github.com/OpenAPITools/openapi-generator/issues/9296#issuecomment-825768917
That conversation is separate from this one.
_from_openapi_data is used when instantiating models from the server so I'm pretty sure that we could depend on _spec_property_naming being true in it. We could throw an ApiValueError if _spec_property_naming is the arguments in _from_openapi_data.

spacether:
@mfmarche this PR just landed:
https://github.com/OpenAPITools/openapi-generator/pull/10155
Which uses cls._from_openapi_data() or cls() when making composed instances.
Are you still interested in merging this feature?
How about rebasing on master, your update should be simpler now and should depend upon using _spec_property_naming like that PR did.

mfmarche:
Hi @spacether PR  updated and rebased with changes as per review comments. Thanks.

samples/client/petstore/python/petstore_api/model_utils.py
```diff
@@ -186,6 +187,24 @@ def __getattr__(self, attr):
         """get the value of an attribute using dot notation: `instance.attr`"""
         return self.__getitem__(attr)
 
+    def __copy__(self):
+        cls = self.__class__
+        return cls._new_from_openapi_data(**self.__dict__)
```
spacether, line 192: Why is copy assuming a from server context? Why not use self._spec_property_naming to pick if we should use `__new__` or `_new_from_openapi_data`?
This same question applies to __deepcopy__

mfmarche: If I have an object, I would assume that should be able to copy it as is, without consideration for read-only args/etc. Now I can't think of a case where that would occur, but I'm not sure I have a grasp of all the ways objects can be created at moment. That was my reasoning, but if you still feel the logic you have described should be used, I will update to that. thanks.

spacether: You will be able to copy an instance of the model that is made using client side data, and you can do the sam with an instance of the model made with data from the server, but those contexts should allow slightly different data to be stored in the instances. Changing _spec_property_naming changes where the model says its data came from, which we should not be doing.
Please make the suggested change, thanks.

mfmarche: I have updated based on your recommendation.

A question regarding the the petstore_api tests, how exactly are those run? I typically run those manually, but I don't see a CI job that actually runs those?

samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
```diff
@@ -186,6 +187,30 @@ def __getattr__(self, attr):
...
+            return cls._new_from_openapi_data(**self.__dict__)
+        else:
+            return new_cls.__new__(cls, **self.__dict__)
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        kwargs = {}
+        if self.discriminator and self.discriminator.keys():
+            key = list(cls.discriminator.keys())[0]
+            kwargs[key] = self.get(key)
```
spacether, line 202: Why are we grabbing the discriminator?
Won't it already be in the new instance because it is defined in the classes?

samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
```diff
@@ -186,6 +187,30 @@ def __getattr__(self, attr):
...
+
+        if self.get("_spec_property_naming", False):
+            new_inst = cls._new_from_openapi_data(**kwargs)
+        else:
+            new_inst = cls.__new__(cls, **kwargs)
+
+        for k, v in self.__dict__.items():
+            setattr(new_inst, k, deepcopy(v, memo))
```
spacether, line 210: How about grabbing all of these properties before passing them into 
_new_from_openapi_data and _new_from_openapi_data?
```
kwargs = {}
for k, v in self.__dict__.items():
    kwargs[key] = deepcopy(v, memo)

if self.get("_spec_property_naming", False):
    new_inst = cls._new_from_openapi_data(**kwargs)
else:
    new_inst = cls.__new__(cls, **kwargs)
```

samples/client/petstore/python/petstore_api/model_utils.py
spacether: They were run by travis, I need to switch them over to CircleCi because we hit a free limit usage on Travis.
Merging this is blocked by that fix.
Thank you for making the changes.

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
```diff
@@ -12,4 +12,27 @@
...
-        return self.{{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}(attr)
\ No newline at end of file
+        return self.{{#attrNoneIfUnset}}get{{/attrNoneIfUnset}}{{^attrNoneIfUnset}}__getitem__{{/attrNoneIfUnset}}(attr)
+
+    def __copy__(self):
+        cls = self.__class__
+        if self.get("_spec_property_naming", False):
+            return cls._new_from_openapi_data(**self.__dict__)
+        else:
+            return new_cls.__new__(cls, **self.__dict__)
```
spacether, line 22: Do we need to call init?

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
```diff
@@ -12,4 +12,27 @@
...
+        cls = self.__class__
+        kwargs = {}
+        if self.discriminator and self.discriminator.keys():
+            key = list(cls.discriminator.keys())[0]
+            kwargs[key] = self.get(key)
+
+        if self.get("_spec_property_naming", False):
+            new_inst = cls._new_from_openapi_data(**kwargs)
+        else:
+            new_inst = cls.__new__(cls, **kwargs)
```
spacether, line 34: We need to call init here, right?

modules/openapi-generator/src/main/resources/python/model_templates/methods_shared.mustache
mfmarche: Doesn't the __new__ definition already do that? 

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/model_utils.mustache#L227

samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
mfmarche: The reason setattr is used rather than kwargs, is that the dictionary set of values (such as the properties), are stored in __dict__["_data_store"]. In order to pass as kwargs, I would need to interpret the data (handle the _data_store as a special case), and pass that in as the kwargs. Rather, using setinst sets the key/value as is without any interpretation, a more resilent solution IMO.

samples/client/petstore/python_disallowAdditionalPropertiesIfNotPresent/petstore_api/model_utils.py
mfmarche: I think that now new_from_openapi_data is used, it does the correct thing with the discriminator so this is no longer needed as you suggest. I will remove this.

spacether:
@mfmarche can you rebase this on master?
https://github.com/OpenAPITools/openapi-generator/pull/10354
just landed which turns python testing back on in circleci.

spacether (approved):
Thanks for rebasing; this looks great!

PR #9742, opened 2021-06-10
[C++][Pistache] Fix compile break, error handling

JulianGmp:
Here I am fixing bugs I myself have created.

#### Fix a compile break for validation code:

The overload for the models `validate(std::stringstream&, const std::string&)` was marked as private, which lead to model classes being unable to use the helper when validating other classes. The function was made public, with a summary comment noting that it is not of much use for the user.

#### Fix error handling in `handleParsingException`:

The defaultl implementation of `handleParsingException` handles exceptions based on their type. In order to check the type, the caught exception is rethrown in a `try` block. However, it is important to rethrow using `throw;` and not with `throw exVariable;`.
The latter does not preserve derived class information (See https://stackoverflow.com/questions/2360597/c-exceptions-questions-on-rethrow-of-original-exception). This lead to all runtime exceptions being treated like generic exceptions and raised an internal server error (500).

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [X] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request. @ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @martindelille (2018/03) @muttleyxd (2019/08)

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-pistache-server/api-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-pistache-server/api-source.mustache
@@ -36,11 +36,13 @@ void {{classname}}::setupRoutes() {
 std::pair<Pistache::Http::Code, std::string> {{classname}}::handleParsingException(const std::exception& ex) const noexcept
 {
     try {
-        throw ex;
+        throw;
     } catch (nlohmann::detail::exception &e) {
         return std::make_pair(Pistache::Http::Code::Bad_Request, e.what());
     } catch ({{helpersNamespace}}::ValidationException &e) {
         return std::make_pair(Pistache::Http::Code::Bad_Request, e.what());
+    } catch (std::exception &e) {
+        return std::make_pair(Pistache::Http::Code::Internal_Server_Error, e.what())
     }
 }
 
--- a/modules/openapi-generator/src/main/resources/cpp-pistache-server/api-header.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-pistache-server/api-header.mustache
@@ -42,13 +42,15 @@ private:
 
     /// <summary>
     /// Helper function to handle unexpected Exceptions during Parameter parsing and validation.
-    /// May be overriden to return custom error formats.
+    /// May be overriden to return custom error formats. This is called inside a catch block.
+    /// Important: When overriding, do not call `throw ex;`, but instead use `throw;`.
     /// </summary>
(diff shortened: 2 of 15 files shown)


wing328:
@JulianGmp thanks for the PR. I'll merge it tomorrow if there's no feedback/question from anyone.

modules/openapi-generator/src/main/resources/cpp-pistache-server/api-source.mustache
```diff
@@ -36,11 +36,13 @@ void {{classname}}::setupRoutes() {
...
 {
     try {
-        throw ex;
+        throw;
     } catch (nlohmann::detail::exception &e) {
         return std::make_pair(Pistache::Http::Code::Bad_Request, e.what());
     } catch ({{helpersNamespace}}::ValidationException &e) {
         return std::make_pair(Pistache::Http::Code::Bad_Request, e.what());
+    } catch (std::exception &e) {
+        return std::make_pair(Pistache::Http::Code::Internal_Server_Error, e.what())
```
stangerm2, line 45: I'm new here but I think this needs a `;`?
`[ 96%] Building CXX object CMakeFiles/api-server.dir/model/User.cpp.o
/home/matt/src/openapi-generator/out/pis-petstore/api/StoreApi.cpp: In member function ‘virtual std::pair<Pistache::Http::Code, std::__cxx11::basic_string<char> > org::openapitools::server::api::StoreApi::handleParsingException(const std::exception&) const’:
/home/matt/src/openapi-generator/out/pis-petstore/api/StoreApi.cpp:54:85: error: expected ‘;’ before ‘}’ token
   54 |         return std::make_pair(Pistache::Http::Code::Internal_Server_Error, e.what())
`

wing328: @stangerm2 thanks for catching the issue. I wonder if you can file a PR with the suggested fix?

JulianGmp: That is an oversight on my part. I did build the generated server, I must have removed the semicolon afterwards by accident?

I can also file a small PR to fix this later today.

JulianGmp: On another note, I don't think there is a pipeline that builds the generated example, or is there?

That could catch a mistake like this before it gets merged.

PR #9763, opened 2021-06-14
Adding Api per operation option to JaxRS and Spring server generators.

### LINKED ISSUE #9219: [REQ] (Server side) Java Class per endpoint

The way the API interfaces are generated created Low cohesion with Clean Architecture business logic approaches.

I'm always frustrated when trying to implement the interface generated by the `openapi-generator` when my business logic is following clean architecture principles where each User Case is done in an independent class. As the generated API contains many methods, my controller implementation will depend on many classes what makes it have low cohesion and coupling with different use case classes (also mappers, DTO, etc).

I would like a similar behavior to the one we have with the `useTags` option.

It could be a property called `interfacePerEndpoint` that generates a class/interface per each endpoint named with the OperationId or a "generic name" with a method called `execute` or `run` or `do` etc, that contains the related annotations and receive the information.

The usage of `useTags` and creating a tag per endpoint looks to give the result I expect in the generated code, but the problem with this approach is the Swagger-UI that is generated, the visual that comes from a spec with this approach has a lot of nonmeaningful groups (One per tag) with only one endpoint.

The other approach that we are thinking of is to create the implementation without extending from the generated Api interface and using the generated Models, but with this approach could miss updates on the specification and we will be misusing what openapi-generator provides us.

pablobastidasv:
Added the option `apiPerOperation` which work similar to `useTags` but instead of generate a Java class per each annotation, this will tells to Spring and JaxRS server generators to create a Class per each operation (_url + method_).

2 new unit tests were added, one per each generator.

fix #9219

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

--

cc. @bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger @karismann @Zomzog @lwlee2608 @nmuesch

### DIFF
--- a/docs/generators/java-msf4j.md
+++ b/docs/generators/java-msf4j.md
@@ -11,6 +11,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |additionalModelTypeAnnotations|Additional annotations for model type(class level annotations)| |null|
 |allowUnicodeIdentifiers|boolean, toggles whether unicode identifiers are allowed in names or not, default is false| |false|
 |apiPackage|package for generated api classes| |org.openapitools.api|
+|apiPerOperation|Create interface and controller per operation| |false|
 |artifactDescription|artifact description in generated pom.xml| |OpenAPI Java|
 |artifactId|artifactId in generated pom.xml. This also becomes part of the generated library's filename| |openapi-jaxrs-server|
 |artifactUrl|artifact URL in generated pom.xml| |https://github.com/openapitools/openapi-generator|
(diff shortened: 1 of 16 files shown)


wing328:
FYI

https://github.com/openapitools/openapi-generator/blob/master/docs/customization.md#openapi-normalizer

please use the rule `SET_TAGS_TO_OPERATIONID` in the openapi normalizer to have API per operationId

Issue #7518, opened 2020-09-27
[REQ][PHP] depend on meta packages providing the HTTP implementation instead of directly on Guzzle


dkarlovi:
### Is your feature request related to a problem? Please describe.

Currently, the packages generated by `php` depend on Guzzle, like so
```json
{
    "require": {
        "php": ">=7.2",
        "ext-curl": "*",
        "ext-json": "*",
        "ext-mbstring": "*",
        "guzzlehttp/guzzle": "^6.2"
    }
}
```

This is functional, but not ideal, making it harder to swap out HTTP clients, like for example [Symfony HTTP Client](https://symfony.com/doc/current/http_client.html) if you're using Symfony.

Being able to choose your desired HTTP client is very desirable in modern PHP apps since you might have existing clients already, setup to log requests via telemetry, use caching, etc.

## Describe the solution you'd like

Instead of relying on Guzzle directly, depend on a `psr/http-client-implementation` (AKA [PSR 18](https://www.php-fig.org/psr/psr-18/)). This is a metapackage [which implementations can declare they provide](https://packagist.org/providers/psr/http-client-implementation), all the major ones already do (including Guzzle, Symfony, etc).

## Describe alternatives you've considered

Alternatively, the client implementation to select might be `php-http/async-client-implementation` (since the client is expected to be async). The [offered implementations are fewer](https://packagist.org/providers/php-http/async-client-implementation), but the major ones are still there. Result is the same: developer being able to inject their desired client.

This could also be done both ways: depend on the regular client, check if it's one of the async ones at runtime and allow async ops if so.

dkarlovi:
This became more important since the used features got deprecated, for example

```
  344    Call to deprecated function GuzzleHttp\Psr7\build_query():            
         build_query will be removed in guzzlehttp/psr7:2.0. Use Query::build  
         instead.
```

grongor:
Hey @dkarlovi , just wanted to ask you if you started working on this or if this is just an issue for reference. (or maybe @ybelenko ? Saw your name in other php-related stuff). We at https://www.cdn77.com are now in need of this, so ... if you or somebody else isn't working on this, I'd look into it.

ybelenko:
@grongor I'm definitely unaware of new PSR-18 and have no unfinished PRs about this issue 🤣

I prefer not to touch current PHP client generator, because it was developed long time ago before I joined community. The only exception is when I see obvious bugs like #8481

dkarlovi:
[...]

PR #9772, opened 2021-06-15
[PHP] 6.0.x make php implementation depend on meta packages for http client

olexiyk:
Instead of relying on Guzzle directly, depend on a psr/http-client-implementation (AKA PSR 18). This is a metapackage which implementations can declare they provide, all the major ones already do (including Guzzle, Symfony, etc).

Inspired by discussion: https://github.com/OpenAPITools/openapi-generator/issues/7518

### PR checklist
 
- [ x ] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ x ] Pull Request title clearly describes the work in the pull request and Pull Request description provides details about how to validate the work. Missing information here may result in delayed response from the community.
- [ x ] Run the following to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) and update samples:
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [ x ] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master`, `5.1.x`, `6.0.x`
- [ x ] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request.

@jebentier (2017/07), @dkarlovi (2017/07), @mandrean (2017/08), @jfastnacht (2017/09), @ackintosh (2017/09) ❤️, @ybelenko (2018/07), @renepardon (2018/12)


modules/openapi-generator/src/main/resources/php/ApiException.mustache
```diff
@@ -108,4 +110,4 @@ class ApiException extends Exception
     {
         return $this->responseObject;
     }
-}
+}
```
dkarlovi, line 113: Newline missing.

modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -588,11 +658,11 @@ use {{invokerPackage}}\ObjectSerializer;
                 $httpBody = new MultipartStream($multipartContents);
 
             } elseif ($headers['Content-Type'] === 'application/json') {
-                $httpBody = \GuzzleHttp\json_encode($formParams);
+                $httpBody = json_encode($formParams);
 
             } else {
                 // for HTTP post (form)
-                $httpBody = \GuzzleHttp\Psr7\build_query($formParams);
+                $httpBody = Query::build($formParams);
```
dkarlovi, line 665: Is this still required?

modules/openapi-generator/src/main/resources/php/api.mustache
```diff
@@ -637,41 +707,85 @@ use {{invokerPackage}}\ObjectSerializer;
...
+        string $operationHost,
+        string $resourcePath,
+        array $queryParams
+    ): UriInterface {
+        $host = parse_url($operationHost, PHP_URL_HOST);
+        $scheme = parse_url($operationHost, PHP_URL_SCHEME);
+        $basePath = parse_url($operationHost, PHP_URL_PATH);
+        $port = parse_url($operationHost, PHP_URL_PORT);
+        $user = parse_url($operationHost, PHP_URL_USER);
+        $password = parse_url($operationHost, PHP_URL_PASS);
```
dkarlovi, line 776: Can call `parse_url` without the second param once and get an array of tokens.

modules/openapi-generator/src/main/resources/php/composer.mustache
```diff
@@ -26,11 +26,18 @@
         "ext-curl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "guzzlehttp/guzzle": "^6.2"
+        "php-http/discovery": "^1.14",
+        "psr/http-factory": "^1.0",
```
dkarlovi, line 30: `http-factory-implementation`, no?

modules/openapi-generator/src/main/resources/php/composer.mustache
```diff
@@ -26,11 +26,18 @@
...
+        "php-http/client-common": "^2.3",
+        "guzzlehttp/psr7": "^1.8",
```
dkarlovi, line 32: Since this is still required, we haven't moved away from Guzzle. :thinking: IIUC, it's only used for the single method call, maybe we don't need it at all.

dkarlovi (changes requested):
This looks great, thanks for working on this! <3

Some questions I'd like to discuss:

modules/openapi-generator/src/main/resources/php/api.mustache
olexiyk: guzzlehttp/psr7 advices to use a different method from the same package instead
> @deprecated build_query will be removed in guzzlehttp/psr7:2.0. Use Query::build instead.

modules/openapi-generator/src/main/resources/php/composer.mustache
olexiyk: `guzzlehttp/psr7` is a different from `guzzlehttp/guzzle`, which was the main target, so we got rid of the http client

But I think it is possible to use something else for `guzzlehttp/psr7`
We use 2 classes from it
```
use GuzzleHttp\Psr7\MultipartStream;
use GuzzleHttp\Psr7\Query;
```
Instead of `GuzzleHttp\Psr7\MultipartStream` we could use MultipartStreamBuilder from  https://github.com/php-http/multipart-stream-builder, also see https://docs.php-http.org/en/latest/components/multipart-stream-builder.html
But as far as I understand, there is no metapackage for that, we need to decide which PSR-7 MultipartStream builder we are going to use

As for `GuzzleHttp\Psr7\Query` I don't know if there is an alternative. It does something similar to what `http_build_query`, maybe https://packagist.org/packages/league/uri-query-parser?

dkarlovi: `guzzlehttp/psr7` provides the `psr/http-factory-implementation` so IMO we should at least look into trying to avoid using it directly, if possible. The reason is it doesn't allow us to swap out that implementation, so we might get the same issues as with the HTTP client, but from a different direction. :smile: 

We don't need to do it here, this is a great first step toward a HTTP agnostic package builder IMO. :+1:

modules/openapi-generator/src/main/resources/php/composer.mustache
dkarlovi: Just realized, we don't require `psr/http-client-implementation`?

modules/openapi-generator/src/main/resources/php/composer.mustache
olexiyk: `guzzlehttp/psr7` is a bit confusing. It provides 
* PSR-7: HTTP message interfaces implementation
* PSR-17: HTTP Factories implementation
* and common utility methods 

We only use these common utility methods directly: `GuzzleHttp\Psr7\MultipartStream` and `GuzzleHttp\Psr7\Query`, they are not part of PSRs

Classes from `php-http/discovery` make the discovery process and it may or may not use guzzle's PSR implementation

We could discuss if we want to replace these utility methods as I suggested earlier, but I would say that it's a bit out of scope, because we don't use any PSR implementations directly

modules/openapi-generator/src/main/resources/php/composer.mustache
olexiyk: I will add these, and also found a few direct dependencies that were not directly required but came as transient dependencies

olexiyk:
Some tests are failing and I figured out the reason, I don't have a solution for it yet.

`http://petstore.swagger.io/v2` is a base url we are running tests against
any request to that host redirects to `https://petstore.swagger.io/v2`

The redirect plugin from here https://github.com/php-http/plugins/blob/master/src/RedirectPlugin.php#L197
when doing a redirect is updating method from POST to GET. In the end, the response to the second request is 405 Method not allowed.

I read a bit about that and to me it looks that probably RedirectPlugin should not change POST to GET, but that is a common implementation bug

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3

> If the 302 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued.
> 
> Note: RFC 1945 and RFC 2068 specify that the client is not allowed
> to change the method on the redirected request. However, most
> existing user agent implementations treat 302 as if it were a 303
> response, performing a GET on the Location field-value regardless
> of the original request method. The status codes 303 and 307 have
> been added for servers that wish to make unambiguously clear which
> kind of reaction is expected of the client.

As well seen at
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

> Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents conform here - you can still find this type of bugged software out there. It is therefore recommended to set the 302 code only as a response for GET or HEAD methods and to use 307 Temporary Redirect instead, as the method change is explicitly prohibited in that case.

I am not sure on how to solve that issue yet, I will get back to the PR after one week

modules/openapi-generator/src/main/resources/php/composer.mustache
arrilot: guzzlehttp/psr7 v2 was released yesterday. https://github.com/guzzle/psr7/releases/tag/2.0.0

olexiyk:
Could you take a look again? The issue with redirect was fixed by using an updated `php-http/client-common` package. `guzzlehttp/psr7` is now either ^1.8 or ^2.0 because the used classes are the same in both

I have seen that build was failing but could not really understand what exactly was wrong

dkarlovi:
@wing328 is the issue with Petstore tests fixed? I remember it's quite flaky.

hanovruslan:
Hi! Does this issue have some forecasts for release?

Argentum88:
@dkarlovi @wing328 Are there reasons for the merge delay?

thomaspicquet:
it would be great to have this merged, let us know if help is needed

wing328:
@olexiyk can you please resolve the merge conflicts when you've time?

wing328:
UPDATE: I've filed https://github.com/OpenAPITools/openapi-generator/pull/16368 to add the implementation via the library option (e.g. --library psr-18 via CLI)

wing328:
Included https://github.com/OpenAPITools/openapi-generator/pull/16368 in v7.0.0 release.

Issue #1317, opened 2018-10-25
[cpp-pistache-server] Enumerations not generated


brianwaters3:
##### Description

Currently the enumeration definition below generates a class with no members.  The attached file archive.zip contains a complete definition (with several enumerations) that demonstrate this problem.

[Archive.zip](https://github.com/OpenAPITools/openapi-generator/files/2515830/Archive.zip)

##### openapi-generator version

v 3.3.0

##### OpenAPI declaration file content or url

```yaml
    SupportedGADShapes:
      anyOf:
        - type: string
          enum:
            - POINT
            - POINT_UNCERTAINTY_CIRCLE
            - POINT_UNCERTAINTY_ELLIPSE
            - POLYGON
            - POINT_ALTITUDE
            - POINT_ALTITUDE_UNCERTAINTY
            - ELLIPSOID_ARC
        - type: string
```

##### Command line used for generation

java -jar /home/brian/code/git/openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar generate -i updated/TS29122_MonitoringEvent.yaml -g cpp-pistache-server --config __config__ -o ./code/MonitoringEvent/server/cpp-pistache-server

##### Steps to reproduce

1. Create a blank working directory
2. Unzip Archive.zip to the newly created directory.
3. Generate the source using the provided command making path adjustments for your system.
4. The class SupportedGADShapes in ./code/MonitoringEvent/server/cpp-pistache-server/model/SupportedGADShapes.h has no members.

##### Suggest a fix/enhancement

The attached file SupportedGADShapes.zip contains the .h and .cpp definitions for the enumeration. 
 Note that an enumeration named "INVALID" has been added in the header and is used when no default is specified.

[SupportedGADShapes.cpp-pistache-server.zip](https://github.com/OpenAPITools/openapi-generator/files/2515849/SupportedGADShapes.cpp-pistache-server.zip)

```
...
    /////////////////////////////////////////////
    /// SupportedGADShapes members
    enum eSupportedGADShapes
    {
       INVALID,
       POINT,
       POINT_UNCERTAINTY_CIRCLE,
       POINT_UNCERTAINTY_ELLIPSE,
       POLYGON,
       POINT_ALTITUDE,
       POINT_ALTITUDE_UNCERTAINTY,
       ELLIPSOID_ARC
    };

    eSupportedGADShapes getValue() const;
    void setValue(eSupportedGADShapes const value);

protected:
[...]

PR #9786, opened 2021-06-16
Continuing from #1317 and its PRs for pistache server string enum code generation

mfyuce:
* A class that has an `anyOf` specification, in cpp side will have no members: instead it should have a member having the type `stringenumclassname_anyOf`
* Thus, Its `==` operator is not present or wrongly filled
* A string enum, should have a better usage, hence the `setEnumValue`
* this PR, is a brigde between `stringenumclassname_**anyOf**` and `stringenumclassname`
* `anyOf` specification is not just about `Enums`, so a better handling is needed from mustache templates, hence new template model parameter `isStringEnumContainer`

### PR checklist
 
- [ X] Read the [contribution guidelines](https://github.com/openapitools/openapi-generator/blob/master/CONTRIBUTING.md).
- [ X] Pull Request title clearly describes the work in the pull request and Pull Request description provides details about how to validate the work. Missing information here may result in delayed response from the community.
- [X ] Run the following to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) and update samples:
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@ravinikam (2017/07) @stkrwork (2017/07) @etherealjoy (2018/02) @MartinDelille (2018/03) @muttleyxd (2019/08)


modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
```diff
@@ -60,21 +60,22 @@ bool {{classname}}::validate(std::stringstream& msg, const std::string& pathPref
...
         {{> model-validation-body }}
     }
-    {{/hasValidation}}{{/isArray}}
-    {{/vars}}
-    {{/isEnum}}
-
+    {{/hasValidation}}{{/isArray}}{{/vars}}{{/isEnum}}{{#isStringEnumContainer}}{{#anyOf}}{{#-first}}
+    if (!m_value.validate(msg))
+    {
+        success = false;;
```
muttleyxd, line 66: double semicolon

muttleyxd (approved):

modules/openapi-generator/src/main/resources/cpp-pistache-server/model-source.mustache
mfyuce: Thanks. It is fixed.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2634,6 +2634,14 @@ public int compare(CodegenProperty one, CodegenProperty another) {
                 postProcessModelProperty(m, prop);
             }
         }
+        if(!m.isEnum
+                && m.anyOf.size()>1
+                && m.anyOf.contains("std::string")
```
wing328, line 2639: I think `std::string` is C++ only.

What about adding `x-is-string-enum-container` instead in the postProcessModel operation in the C++ pistache server generator?

mfyuce: Thanks.

- Removed "isStringEnumContainer" and its setter from "DefaultCodeGen"
- added the setter to "CppPistacheServerCodegen.java" making it add a new entry to "vendorExtensions" as "x-is-string-enum-container"

Now it is pistache cpp server only. Making it global to all servers...; you may guide me..

wing328: If other generators need the same, they can reuse the same code block in the postProcessModel method.

PR #9799, opened 2021-06-17
Use Kotlin nullable operator for nullable properties

josh-bridge:
The previous check of isReadOnly was not applicable to whether a value should be null or not. As far as I can tell that is more to do with a RESTful design principal than to do with whether the code should present a models values as nullable. 
Initially I thought the `?` should be removed if the value is set to required = true but it is important to note that **required != non-null**, i.e. the field can be required but the value might still be null (if nullable = true).

The default null value will no longer be supplied if the value is not nullable.

Also removed escaping of default string values (this would cause any default string value to be written to the file as 
`val test = &quot;value&quot;` instead of `val test = "value"`.

This would clearly be a breaking change so have set it to a major release branch. Please let me know if you think it could actually go into a minor release.

[kotlin committee] 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


kuFEAR:
Why do all properties become required in your samples and without default '= null' ?  If you look on kotlinx.serialization you will see that if the field is not in model(empty field, not null) it crashes on runtime.

josh-bridge:
> Why do all properties become required in your samples and without default '= null' ? If you look on kotlinx.serialization you will see that if the field is not in model(empty field, not null) it crashes on runtime.

The default value of the `nullable` option in the OpenAPI schema is `false`, as shown below
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schemaNullable

If the value by default should not be nullable, then the fields themselves should also not be nullable right? And to your point about it crashing, that should be intended value to passing a null value to a field which is not specified to be nullable

kuFEAR:
https://github.com/Kotlin/kotlinx.serialization/issues/1196 In last version kotlinx.serialization has new flag `explicitNulls ` for fixing this behaviour.
Now I use this generator with kotlinx.serialization and it generate `val field: Type? = null` properties if it "optional" and it works well, without crashes:
```kotlin
fun provideKotlinxSerialization(): Json = Json {
        registerCommonJvmAdapters()
        coerceInputValues = true
        ignoreUnknownKeys = true
        isLenient = true
    }
```

PR #9807, opened 2021-06-18
conditional serialization add only those property which user wants to…

Ghufz:
ConditionalSerialization is exposed as flag when it sets to true it generates the shouldSerialize<popertyName>() method for each property along with flag which detect that property has been changes or set, if the property has been changed then the flag is set to true and ShouldSerialize<PrpopertyName> return true, which allows the serialization that configured property in request body. 

-This gives a control to the user to decide what are the properties can be set in request body irrespective of sending all the other property which user does not configured.

We face the issue mostly in the case of update the MO, where user wants to configure only one property but in the request all the nulllable properties set to null. which leads to inconsistency.

-This solved another use case where user is using the latest SDK which has newly added property to the existing object and the server side is not updated(working with older openAPIDoc). When user tries to send the request to the server the newly added property also become a part of request body which server does not know abut it, it leads to exception. ConditionalSerialization will not include those newly added property unless and until it is being configured by the user. 

Here is the example :- 
Let say we have Name object which has two properties FirstName and LastName

when user wants to create new Name object the request would be like 
```
{
     "FirstName": "Tom",
      "LastName" : "Harry"
}
```
once above object is created let say user wants to update only "LastName" so the payload as per the deafult Csharp SDK would be 
```
{
  "FirstName" :  null,
  "LastName" :  "Xyz"
}
```
When the above request sents to the server, it updates the firstName as null along with LastName which leads to inconsistency.
when we generate the C# SDK using the conditionalSerialization flag the above request cant be sent like below

```
{
     "LastName" : "Xyz"
}
```
the above request sends only those property which user has configured. 

Note :- there is no change in default code generation.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
@wing328


wing328:
cc @mandrean (2017/08) @frankyjuang (2019/09) @shibayan (2020/02) @Blackclaws (2021/03) @lucamazzanti (2021/05)

wing328:
Remove some trailing spaces and update the samples via https://github.com/OpenAPITools/openapi-generator/pull/9807/commits/14cc4821bb3919e30d53a1eccc0bdc4abca4e85f

roderickgreen:
Thanks for the feature @Ghufz. I was excited to see this because I have a similar issue to resolve. Unfortunately this doesn't cover my use case. I am making heavy use of model construction via ctors, and the ctors don't write to the properties. So after I enabled this option, none of my properties get serialized.  Do you have any thoughts about this from the work you put in? If the parameterless ctor was public I could initialize the models with object initializers, but that might break some of the IsRequired checking.

Issue #9792, opened 2021-06-16
[REQ] Java Client Specify Bearer in each call


cyrusv:
### Is your feature request related to a problem? Please describe.
I am writing a Java app that will use a generated Java Client. I'd like to use the pooled connection, but I need to specify a different Bearer token in each request.

## Describe the solution you'd like
I'd like to update the mustache templates for Java client code to make an additional method signature for each client call that takes an Authentication, which, when set, will take precedence over the one defined on the instance. For example, :
```
public List<String> myMethod(MyMethodRequest myMethodRequest, Authentication auth) throws ApiException
```

I'd update all signatures in the *Api.java files to support the existing sig and new sig, but all deeper files I think can just support the new way. I'd pass down the param and just put some simple override logic in `updateParamsForAuth` method.

## Describe alternatives you've considered
To support this feature, the method signature must change for the calls (changing client instance itself is not threadsafe). The other method sig I considered:
 - Pass in a Map<String, String> of headers to override. This one is nice bc it is more general than just doing this for Authentication implementations, but I had some trouble deciding how the override logic would work since the various functions in the call stack do different things to different headers, I didn't want to make them more complicated. It would be relatively straightforward to just keep the logic as-is, and right before sending the request apply all these overrides to the request headers. Not a bad option.

Ultimately I think `Map<String, String> headers` and `Authentication auth` are both good choices, but I leaned towards `Authentication` since it is more structured and has fewer magic strings.

## I am 100% ready to write the PR, as long as I get some feedback that the strategy is generally correct and someone is willing to work through review process with me!

PR #9820, opened 2021-06-21
[JAVA CLIENT] Add Support to Override Headers for each Request

cyrusv:
My first contribution -- hoping everything is set up ok! 

Resolves https://github.com/OpenAPITools/openapi-generator/issues/9792. Basically, I want to use a shared connection, but I want to specify different JWT for each request. This has to be done by overloading the calls (some calls have no params or primitive params, so I need to have a new param). I implemented as a headers map to be more generally useful.

I added a  `Map<String, String> headers` to the client calls, and passed it down the stack.

While I was fixing, I found it difficult to reason about when these headers were being modified, because the `String[] authNames` parameter was kind of goofy, and all the request headers were not modified in the same place -- I changed the code to now do all the header changes/mutations in one place. This makes it easier to know for certain that any headers intended to override are in fact added after all other mutations to the headers.

Hoping for feedback from Java reviewers:
@bbdouglas @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger @karismann @Zomzog  @lwlee2608  @nmuesch 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

Also I'm an openapi newbie, so would certainly appreciate some help understanding the test failures in CI?

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/api.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/api.mustache
@@ -96,7 +96,7 @@ public class {{classname}} {
     {{#isDeprecated}}
     @Deprecated
     {{/isDeprecated}}
-    public{{/vendorExtensions.x-group-parameters}}{{#vendorExtensions.x-group-parameters}}private{{/vendorExtensions.x-group-parameters}} okhttp3.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}final ApiCallback _callback) throws ApiException {
+    public{{/vendorExtensions.x-group-parameters}}{{#vendorExtensions.x-group-parameters}}private{{/vendorExtensions.x-group-parameters}} okhttp3.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}Map<String, String> headers, final ApiCallback _callback) throws ApiException {
         Object localVarPostBody = {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
 
         // create path and map variables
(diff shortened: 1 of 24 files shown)


cyrusv:
@MosheElisha  or @wing328  -- any help here?  I will fix the tests, but I want to be sure the strategy is generally acceptable before I sink more time into it.

wing328:
@cyrusv thanks for the PR. I'll try to review tomorrow or later this week and let you know.

MosheElisha:
@cyrusv Thanks for the PR.

The approach I take when I need to control headers per request is to use ThreadLocal and interceptor.
When using the client you do something like:

```
try (RequestModifier requestModifier = new RequestModifier()
        .addHeader("Authorization", authToken)) {

  myClient.deleteBook(bookId);

}
```
RequestModifier stores the headers in a ThreadLocal map and an interceptor reads those headers from the thread and adds them to the request.
[RequestModifier.zip](https://github.com/OpenAPITools/openapi-generator/files/6790255/RequestModifier.zip)

I have added RequestModifier code. The implementation of the RequestModifier and interceptor implementation is different for every `library` but it is pretty straight forward.

modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -96,7 +96,7 @@ public class {{classname}} {
     {{#isDeprecated}}
     @Deprecated
     {{/isDeprecated}}
-    public{{/vendorExtensions.x-group-parameters}}{{#vendorExtensions.x-group-parameters}}private{{/vendorExtensions.x-group-parameters}} okhttp3.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}final ApiCallback _callback) throws ApiException {
+    public{{/vendorExtensions.x-group-parameters}}{{#vendorExtensions.x-group-parameters}}private{{/vendorExtensions.x-group-parameters}} okhttp3.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}Map<String, String> headers, final ApiCallback _callback) throws ApiException {
```
wing328, line 99: > Map<String, String> headers, 

My opinion is that I prefer not to add an additional parameter for every methods.

Have you considered using the default headers in the ApiClient object to achieve what you want?  

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache#L529

cyrusv: Thanks, @wing328  -- I had considered this, but that wouldn't be a threadsafe way to do it in a case where multiple threads are sharing the instance.

cyrusv:
Thanks @MosheElisha, I didn't know about Request Modifiers. Let me see if they can solve my problem.

PR #9834, opened 2021-06-23
Opt-in solution to remove unnecessary model for inline schema composition

### LINKED ISSUE #3100: [BUG] [JAVA] Java generation of 'extra' class with allOf
**Description**

I am using the pet store example to create java classes with inheritance. In the newest release (4.0.1) I found a bug. When I generate code with the OpenApi declaration file it generates an extra class.

image

The class 'CatAllOf' I did not expect. It looks the same as 'Cat' except that it does not extends Pet.

**OpenAPI declaration file:**

```
openapi: 3.0.1
info:
  title: Article Event
  description: 'Article Event schema'
  version: 1.0.0
paths:
  /:
    get:
      responses:
        200:
          description: OK
          content: {}
components:
  schemas:
    TextChunk:
      type: object
      discriminator:
        propertyName: textChunkType
      properties:
        parent:
          type: string
    Dog:
      allOf: 
      - $ref: '#/components/schemas/TextChunk'
      - type: object
    Cat:
      allOf:
      - $ref: '#/components/schemas/TextChunk'
      - type: object
        properties:
          bark:
            type: string
```

**openapi-generator version**
I used the gradle plugin with version 4.0.1.

**older version**
I tested the same thing in the version 4.0.0-beta3 and there I did net see the 'extra' class. In other words, I got the expected behavior. It seems the bug was introduced with the latest release.

### LINKED ISSUE #5171: [BUG][typescript-angular] allOf without discriminator results in unnecessary imports

leo-sale:
fix #3100 fix #5171
# Description
When there is a `ComposedSchema` with inline schemas, the children are flatten. What is does is it's creating a new schema, by taking the parent's name and adding `_allOf`, `_anyOf` or `_oneOf`, or taking the `title` value and appending all the properties defined in the inline schema. Then the inline schema becomes just a reference to the newly created schema which is added to the `imports` of the parent. (example in this [issue](https://github.com/OpenAPITools/openapi-generator/issues/5171))

This is alright when we have a `discriminator` in our inline schema and what we wish for is **inheritance** but when all we want is **composition** it's just creating unused schemas.

A possible quick fix to prevent these unnecessary schemas is to ignore them in the `.openapi-generator-ignore` but it does not remove the schemas from the `allModels` instance meaning that you can have reference to them in your generated files (like some imports or exports, hence possibly breaking the build).

# Solution
The proposed solution is to generate the new models **only** when it's for inheritance and just adding the properties of the inline schema to the parent (overriding the parent's properties when there is a name's conflict) without generating a new model. This solution comes with an opt-in CLI parameter : `--allow-inline-schemas`, to ease the merge of this fix.
```
java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate \
   -i https://raw.githubusercontent.com/openapitools/openapi-generator/master/modules/openapi-generator/src/test/resources/3_0/petstore.yaml \
   -g php \
   -o /var/tmp/php_api_client
   --allow-inline-schemas
```

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -248,6 +248,9 @@ public class Generate extends OpenApiGeneratorCommand {
     @Option(name = {"--legacy-discriminator-behavior"}, title = "Support legacy logic for evaluating discriminators", description = CodegenConstants.LEGACY_DISCRIMINATOR_BEHAVIOR_DESC)
     private Boolean legacyDiscriminatorBehavior;
 
+    @Option(name = {"--allow-inline-schemas"}, title = "Allow inline schemas without generating new models", description = CodegenConstants.ALLOW_INLINE_SCHEMAS)
+    private Boolean allowInlineSchema;
+
     @Option(name = {"--minimal-update"},
         title = "Minimal update",
         description = "Only write output files that have changed.")
@@ -416,6 +419,10 @@ public class Generate extends OpenApiGeneratorCommand {
 
         if (strictSpecBehavior != null) {
             configurator.setStrictSpecBehavior(strictSpecBehavior);
+        }
+
+        if (allowInlineSchema != null) {
+            configurator.setAllowInlineSchemas(allowInlineSchema);
         }
 
         if (globalProperties != null && !globalProperties.isEmpty()) {
--- a/modules/openapi-generator-cli/src/test/java/org/openapitools/codegen/cmd/GenerateTest.java
+++ b/modules/openapi-generator-cli/src/test/java/org/openapitools/codegen/cmd/GenerateTest.java
@@ -430,4 +430,10 @@ public class GenerateTest {
         verify(configurator).toContext();
         verifyNoMoreInteractions(configurator);
     }
+
+    @Test
+    public void testAllowInlineSchemasTrue() {
+        setupAndRunGenericTest("--allow-inline-schemas");
+        verify(configurator).setAllowInlineSchemas(true);
+    }
 }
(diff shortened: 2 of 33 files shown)


leo-sale:
Hey @wing328, could you have a look at this PR ? It has almost been 2 months but there is still no review 😢
Thank you 😃

djbreen7:
Eagerly awaiting this to be merged

RoundbootyGuilliman:
Anyone? Why is this PR ignored for half a year when this is a widely requested change?

efriandika:
Hi @wing328 If you don't mind, please have a look at this PR.

magentaRE:
any updates on this?

wing328:
can you try the PR I filed instead? You can find the details in https://github.com/OpenAPITools/openapi-generator/issues/3100#issuecomment-1255043072

PR #9840, opened 2021-06-24
Fix issue where skipIfSpecIsUnchanged was ignored when using a classpath input spec

### LINKED ISSUE #5805: skipIfSpecIsUnchanged doesn't work if the spec is not in the project
https://github.com/OpenAPITools/openapi-generator/blob/c224cf484b020a7f5997d883cf331715df3fb52a/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java#L451

When checking the original hash, you don't read the spec file when it is in a dependency rather then in the project itself. Later on, when calculating the new one, you take care to read it wherever it might be. This way you have the hash stored, but you never use it and the API is generated every time. This is very annoying for me, because my Eclipse keeps on generating sources and building the new ones and so on. 

Could you fix this or do you want me to do it?

dhutchison:
The `skipIfSpecIsUnchanged` configuration setting to the maven plugin did not work when the input spec came from a classpath resource, which could lead to infinite build loops when the plugin was used in eclipse.

Fix #5805

The `CodeGenMojo` appears to have already had support for creating hashes of input specs from different locations, but an incorrect `if` check was preventing these non-filesystem locations having hashes created. 

Two additional test methods have been added to test that a spec from the classpath can respect the `skipIfSpecIsUnchanged` setting. 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
+++ b/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java
@@ -458,7 +458,7 @@ public class CodeGenMojo extends AbstractMojo {
                 }
             }
 
-            if (Boolean.TRUE.equals(skipIfSpecIsUnchanged) && inputSpecFile.exists()) {
+            if (Boolean.TRUE.equals(skipIfSpecIsUnchanged)) {
                 File storedInputSpecHashFile = getHashFile(inputSpecFile);
                 if (storedInputSpecHashFile.exists()) {
                     String inputSpecHash = null;
--- a/modules/openapi-generator-maven-plugin/src/test/java/org/openapitools/codegen/plugin/CodeGenMojoTest.java
+++ b/modules/openapi-generator-maven-plugin/src/test/java/org/openapitools/codegen/plugin/CodeGenMojoTest.java
@@ -16,6 +16,13 @@
 
 package org.openapitools.codegen.plugin;
 
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Comparator;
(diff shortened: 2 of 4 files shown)


dhutchison:
Hi @wing328, any chance you could approve a first-time workflow so the checks can complete on this?

wing328:
"OpenAPI Generator / Samples up-to-date (pull_request)" failed already fixed in the master.

wing328:
I ran `mvn --no-snapshot-updates --quiet clean compile -f modules/openapi-generator-maven-plugin/src/test/resources/classpath/pom.xml` at the root folder of this project but got the following errors:
```
[ERROR] failed to read resource listing
java.lang.RuntimeException: Could not find petstore-on-classpath.yaml on the classpath
```

Tried with `mvn --no-snapshot-updates --quiet clean compile -f pom.xml` in the folder `modules/openapi-generator-maven-plugin/src/test/resources/classpath/` but no luck either.

Am I testing the change correctly?

dhutchison:
It is tested by the test case file `CodeGenMojoTest.java`. As it relies on the classpath, as-is that pom file will only work when used through the test case, as the input spec location is relative to the base directory of the maven plugin module. (When I use this approach in real projects it would be coming from a dependency jar, but I couldn't get that to work with the existing CodeGenMojoTest setup - it doesn't seem to be able to do dependency resolution)

If it is important that the test pom files run independently I could probably modify it to use a URL location instead of a classpath location (both those cases would not work before the fix, but should after the fix).

dhutchison:
Are there any changes required to the test to allow this to be merged?

wing328:
@dhutchison sorry for the delay. The PR has been merged. Thanks for the contribution.

PR #6773, opened 2020-06-25
[SWIFT5] Add the ability to turn uniqueItems array into Set in swift.

cerupcat:
Add the ability to turn uniqueItems array into Set in swift.

CC Swift5 technical committee @jgavris (@ehyche @Edubits @jaz-ah @4brunu

PR #9849, opened 2021-06-25
Add support for generating a Set for a list with uniqueItems set to true

### LINKED ISSUE #9848: [REQ] [Kotlin] Use kotlin.Set for an array with uniqueItems set to true
When defining an array with uniqueItems set to true, I expect that the type is a `Set`. But right now it's always a `List`.

Add support for the `Set` type.

Right now I'm using I'm using custom templates to generate this. I think this would be useful to be part of the default templates.

marcdejonge:
To fix #9848

This specifically adds support for a Set type when a model with an array has `uniqueItems` set to true. This only changes the KotlinClientGenerator template, since that one was missing specific support for it. The tests run agains all 5 generators, but the rest seem to work fine.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


marcdejonge:

Is this the correct solution? I'd much rather go back to the `{{{dataType}}}` solution, but then it would break the kotlinx-serializer I think.

marcdejonge:
> Instead of changing the template, what about changing the java file?
> #6773
> Here is an example of something similar in the swift client.

I agree, but the template already doesn't use the dataType at all. The code already contains the logic for using a Set instead of a List (for the kotlin server templates it works fine). I don't know how else to solve this.

4brunu:
Thanks, let's go with this solution 👍

PR #9861, opened 2021-06-29
[cleanup] erefactor/AutoRefactor - TestNG asserts

cal101:
AutoRefactor cleanup 'TestNGAssert' applied by erefactor:

Refactors to a proper use of TestNG assertions.

For AutoRefactor see https://github.com/JnRouvignac/AutoRefactor
For erefactor see https://github.com/cal101/erefactor

--------

Dear Project Team!

This pull request is provided in the hope that it will be useful.

Changes are manually reviewed after they passed the test suite of
your project. Only successful reviews become project PRs.
Errors found are reported upstream to support the cleanup providers.

But please do a careful review of your own as usual because tools
have bugs and bugs in the changes may have been overlooked.

For scalability and out of fun and laziness, the creation of commits
and PRs and parts of the change handling are automated.

If you are not interested or want to discuss issues just comment the
PR as usual.

Have fun,
   Cal

### DIFF
--- a/modules/openapi-generator/src/test/java/org/openapitools/codegen/haskellservant/HaskellModelTest.java
+++ b/modules/openapi-generator/src/test/java/org/openapitools/codegen/haskellservant/HaskellModelTest.java
@@ -24,6 +24,5 @@ public class HaskellModelTest {
 
     @Test(description = "convert a haskell model with dots")
     public void modelTest() {
-        Assert.assertEquals(true, true);
     }
 }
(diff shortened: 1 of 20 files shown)


modules/openapi-generator/src/test/java/org/openapitools/codegen/haskellservant/HaskellModelTest.java
```diff
@@ -24,6 +24,5 @@
 
     @Test(description = "convert a haskell model with dots")
     public void modelTest() {
-        Assert.assertEquals(true, true);
```
nxtstep, line 27: this whole (test) class seems to be obsolete then?

wing328: Good catch. We will revise it with better model tests for the Haskell Servant generator later.

PR #9958, opened 2021-07-16
[typescript-rxjs] add rxjs 7 support

denyo:
This PR adds support for rxjs 7 (https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md) which introduces some breaking interface changes but brings both better minification and improved performance.

Also type imports are refactored to type-only imports.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


macjohnny:
@denyo thanks for your contribution. would there be a possibility to support RxJS 7 in a backwards compatible way? maybe introduce an option or version flag, similar to https://github.com/OpenAPITools/openapi-generator/blob/c4b98041391a0a94e1d668addac0e6de6bf2a25b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptAngularClientCodegen.java#L121?

macjohnny:
@denyo ping

denyo:
@macjohnny I won't get the chance to make it backwards compatible.
Therefore I would suggest to change the target branch to the next major version.

macjohnny:
@wing328 what is your opinion here? should we break the existing rxjs generator, or should we create a new rxjs7 generator?

TiFu:
Quick question: Why does this PR affect 4,359 files? Looks like some unrelated commits have been included.

macjohnny:
@TiFu I guess this is because of the changed target branch

trumbitta:
@macjohnny would it possible to release this as is and let people stuck with rxjs 6 (rxjs 8 is coming) just point to an old version of the generator by themselves?

Polm1:
Following 👀

macjohnny:
According to https://github.com/OpenAPITools/openapi-generator#11---compatibility, we can have breaking changes only for a major release.
If an option is introduced, we can also habe it for the current minor or patch release

macjohnny:
@denyo can you rebase onto the 6.0 branch to avoid the unrelated changes?

jopelle1:
I opened a new PR fixing this one https://github.com/OpenAPITools/openapi-generator/pull/11364 : 
- Rebased on 6.0x
- Implemented option to switch to rxjs 7

denyo:
@macjohnny done.

For everyone else waiting, this is the docker image built from this branch: [denyo/openapi-generator-cli:6.0.0-test](https://hub.docker.com/repository/docker/denyo/openapi-generator-cli/general)

macjohnny:
I would suggest to close this PR in favor of https://github.com/OpenAPITools/openapi-generator/pull/11364, which we could have in 5.4, and thus avoid merge conflicts in the future.
Any objections?

denyo:
> I would suggest to close this PR in favor of #11364, which we could have in 5.4, and thus avoid merge conflicts in the future. Any objections?

The only concern I have is that complexity will increase in the generator with every new version of rxjs, especially within templates, e.g. by adding a toggle or extending the flag for every version and at the same time users need to update their flags/toggles, too. 

From a user's perspective I would expect to get the latest version by default and have an opt-in for older versions if I really wanna be backwards compatible. At least this is what I am used to.

However the philosophy of this project might be different.

macjohnny:
@denyo we can remove the rxjs<7 support for the 6.0 release, which is an easy thing to do an removes the complexity in the template

macjohnny:
since the next release will be 6.0.0, lets merge this PR

PR #9989, opened 2021-07-20
[swift5 client] fix filename for binary downloads

gretzki:
### Problem:

When a file download is handled within URLSessionImplementations,
the filename is retrieved via the header _Content-Disposition_ part of the response.
The header is looked for an entry starting with _filename=_. 
But the search is skipped too early, since the loop is exited if the first iteration does not match.

### Example:
Try to download a file with a swift5 client for a spec like this:
```
  /PublicDocument/Download:
    get:
      tags:
        - Public Area
      description: 'Document Download for Public Documents. Login not Required'
      parameters:
        - $ref: '#/components/parameters/queryId'
      responses:
        '200':
          description: 'OK'
          content:
            application/pdf:
              schema:
                type: string
                format: binary
```

When downloading from a server, the response header could have an answer like so:

```
Content-Disposition: attachment; filename=downloaded.pdf; filename*=UTF-8''downloaded.pdf
```

There, the server says to use downloaded.pdf as filename. But the actual implementation of the swift5 client will exit after checking _attachment_ and won't find `filename=` in the next section (separated by semicolon).

### Solution:
Simply replacing _break_ with _continue_ in the matching logic inside the loop fixes the algorithm.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
@jgavris, @ehyche, @Edubits, @jaz-ah, @4brunu

### DIFF
--- a/modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache
@@ -277,7 +277,7 @@ private var managerStore = SynchronizedDictionary<String, Alamofire.SessionManag
 
             let filenameKey = "filename="
             guard let range = contentItem.range(of: filenameKey) else {
-                break
+                continue
             }
 
             filename = contentItem
--- a/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
@@ -270,7 +270,7 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
 
             let filenameKey = "filename="
             guard let range = contentItem.range(of: filenameKey) else {
-                break
+                continue
             }
 
             filename = contentItem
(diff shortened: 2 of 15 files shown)


4brunu:
@gretzki Thanks for your contribution.
Could you please also fix this in the AlamofireImplementations? 

https://github.com/OpenAPITools/openapi-generator/blob/1d648223e6fdc1adad8967db05d86db9fc49a581/modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache#L276-L288

And to generate the sample projects, please run and commit the changes.
```
./mvnw clean package 
./bin/generate-samples.sh bin/configs/swift5-*
```

Thanks

gretzki:
@4brunu sure. Done and done.
I realised it's also a bug for the swift4 client codegen but didn't fix it, since it's not relevant for the above mentioned client swift5. If you think it should be handled here too, I'm happy to fix that part too.

4brunu:
Swift4 client is now deprecated, so I'm not sure if it's worth the effort.
I think for now I prefer to focus on the Swift5 client, unless there is still demand for the Swift4 client.

4brunu:
CI is failing, but it's not related to this PR.
Looks good to me this PR 🙂

PR #10017, opened 2021-07-22
[typescript-fetch] correct oneOf model to `or` not `concat` when handling enums

### LINKED ISSUE #5202: [BUG][Typescript-Fetch] Incorrect toJSON & fromJSONTyped functions
#### Bug Report Checklist

##### Description

For a property defined using `oneOf`, invalid methods are generated for `<Model>FromJSONTyped` and for `<Model>ToJSON`:

```js
export function TodosFromJSONTyped(json: any, ignoreDiscriminator: boolean): Todos {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    return {
        
        'url': json['url'],
        'firstUrl': !exists(json, 'first_url') ? undefined : json['first_url'],
        'lastUrl': !exists(json, 'last_url') ? undefined : json['last_url'],
        'prevUrl': !exists(json, 'prev_url') ? undefined : string | UrlListFromJSON(json['prev_url']),
        'nextUrl': !exists(json, 'next_url') ? undefined : string | UrlListFromJSON(json['next_url']),
        'createUrl': !exists(json, 'create_url') ? undefined : json['create_url'],
        'findTodoUrl': json['find_todo_url'],
        'count': json['count'],
        'page': json['page'],
        'totalCount': json['total_count'],
        'totalPages': json['total_pages'],
        'items': ((json['items'] as Array<any>).map(TodoFromJSON)),
    };
}
```

```js
export function TodosToJSON(value?: Todos | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return {
        
        'url': value.url,
        'first_url': value.firstUrl,
        'last_url': value.lastUrl,
        'prev_url': string | UrlListToJSON(value.prevUrl),  // THIS IS INVALID TYPESCRIPT
        'next_url': string | UrlListToJSON(value.nextUrl), // THIS IS INVALID TYPESCRIPT
        'create_url': value.createUrl,
        'find_todo_url': value.findTodoUrl,
        'count': value.count,
        'page': value.page,
        'total_count': value.totalCount,
        'total_pages': value.totalPages,
        'items': ((value.items as Array<any>).map(TodoToJSON)),
    };
}
```

### The interface for the above type `Todos`
<details>
  <summary>Click to expand</summary>
  
```js
export interface Todos {
    /**
     * Link for the current page of results in the series.
     * @type {string}
     * @memberof Todos
     */
    url: string;
    /**
     * Links to first pages of results in a series.
     * @type {string}
     * @memberof Todos
     */
    firstUrl?: string;
    /**
     * Link to the last page of results in a series.
     * @type {string}
     * @memberof Todos
     */
    lastUrl?: string;
    /**
     * List of links to previous pages of results in a series.  The first element in the array is the exact previous page before the current record, etc. 
     * @type {string | UrlList}
     * @memberof Todos
     */
    prevUrl?: string | UrlList;
    /**
     * List of links to next pages of results in a series.   The first element in the array is the exact next page after the current record, etc. 
     * @type {string | UrlList}
     * @memberof Todos
     */
    nextUrl?: string | UrlList;
    /**
     * The url used to create a Todo. If not present, the current user is not authorized to create.
     * @type {string}
     * @memberof Todos
     */
    createUrl?: string;
    /**
     * A url template to be used to find a particular todo. 
     * @type {string}
     * @memberof Todos
     */
    findTodoUrl: string;
    /**
     * Number of records on the current page.
     * @type {number}
     * @memberof Todos
     */
    count: number;
    /**
     * Number of the current page. 1-index based.
     * @type {number}
     * @memberof Todos
     */
    page: number;
    /**
     * Total number of records for this series.
     * @type {number}
     * @memberof Todos
     */
    totalCount: number;
    /**
     * Total number of pages for this series.
     * @type {number}
     * @memberof Todos
     */
    totalPages: number;
    /**
     * 
     * @type {Array<Todo>}
     * @memberof Todos
     */
    items: Array<Todo>;
}
```

</details>

`4.2.3` via 

```
yarn list v1.21.0
└─ @openapitools/openapi-generator-cli@1.0.10-4.2.3
```

```yaml
  schemas:
    urlValue:
      type: string
      description: A single url.
    urlList:
      type: array
      description: A list of urls.
      items:
        type: string

...

# For a property

        prev_url:
          description: |
            List of links to previous pages of results in a series.

            The first element in the array is the exact previous page before the current record, etc.
          oneOf:
            - $ref: '#/components/schemas/urlValue'
            - $ref: '#/components/schemas/urlList'
```

`./node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator generate -g typescript-fetch -i ./openapi.yaml -o .`

**CONFIG:** - `"generateAliasAsModel": true`

### LINKED ISSUE #6513: [BUG] typescript-fetch allOf doesn't handle discrimination by property.
##### Description

OpenApi generator does not generate valid code for unions discriminated by property name. Unions (oneOf) work correctly when discriminated by a property value, however the generated code seems to try and combine the property lists from all members of the union. If each member of the union have disjoint sets of required properties, then the union is discriminated, and should be generated as such. 

openapi-generator 4.3.1

```json
{
    "openapi": "3.0",
    "info": {
        "title": "Test",
        "version": "0.1"
    },
    "servers": [{
        "url": "https://localhost"
    }],
    "paths": {
        "/": {
            "get": {
                "operationId": "dummyOperation",
                "responses": {
                    "200": {
                        "description": "Successful operaion",
                        "content": {
                            "application/json": {
                                "schema": {"$ref": "#/components/schemas/aOrB"}
                            }
                        }
                    }
                }
            }
        }
    },
    "components": {
        "schemas": {
            "a": {
                "type": "object",
                "required": ["a"],
                "properties": {
                    "a": {
                        "type": "string"
                    }
                }
            },
            "b": {
                "type": "object",
                "required": ["b"],
                "properties": {
                    "b": {
                        "type": "string"
                    }
                }
            },
            "aOrB": {
                "oneOf": [
                    {"$ref": "#/components/schemas/a"},
                    {"$ref": "#/components/schemas/b"}
                ]
            }
        }
    }
}
```

openapi-generator generate -i test.json -g typescript-fetch --additional-properties=typescriptThreePlus --generate-alias-as-model -o test

##### Steps to reproduce

After generation, the model file for aOrB is:

export function AOrBFromJSONTyped(json: any, ignoreDiscriminator: boolean): AOrB {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    return { ...AFromJSONTyped(json, true), ...BFromJSONTyped(json, true) };
}

export function AOrBToJSON(value?: AOrB | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return { ...AToJSON(value), ...BToJSON(value) };
}

This fails because A and B both contain 'required' parameters, so they cannot be combined by spreading as above. 

##### Related issues/PRs

There are similar PR, but they don't seem to address exactly the same issue:
#6376 #5202 #4626 #1563 #927

##### Suggest a fix

The generated functions need to discriminate on properties as well as values, something like:

```typescript
export function AOrBFromJSONTyped(json: any, ignoreDiscriminator: boolean): AOrB {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    if (json['a'] != undefined) {
        return AFromJSONTyped(json, true);
    } else if (json['b'] != undefined) {
        return BFromJSONTyped(json, true);
    }
}

export function AOrBToJSON(value?: AOrB | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    if (value.a != undefined) {
        return AToJSON(value);
    } else if (value.b != undefined) {
        return BToJSON(value);
    }
}
```

prince-chrismc:
This was report in 2019 https://github.com/OpenAPITools/openapi-generator/pull/4387#issuecomment-557541270 shortly after the original PR the author noted it was not tested... I do not know how to add test code but if you point me in the right direction I can try!

still broken #5202
resolves #6513

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


samples/client/petstore/typescript-fetch/builds/enum/models/ObjectOneOfEnum.ts
```diff
@@ -0,0 +1,68 @@
...
+    return ObjectOneOfEnumFromJSONTyped(json, false);
+}
+
+export function ObjectOneOfEnumFromJSONTyped(json: any, ignoreDiscriminator: boolean): ObjectOneOfEnum {
+    if ((json === undefined) || (json === null)) {
+        return json;
+    }
+    return {
+        
+        'combinedStringEnum': !exists(json, 'combined-string-enum') ? undefined : StringEnum | StringEnumTwoFromJSON(json['combined-string-enum']),
```
prince-chrismc, line 51: This is still incorrectly generated

samples/client/petstore/typescript-fetch/builds/enum/models/ObjectOneOfEnum.ts
```diff
@@ -0,0 +1,68 @@
...
+export function ObjectOneOfEnumToJSON(value?: ObjectOneOfEnum | null): any {
+    if (value === undefined) {
+        return undefined;
+    }
+    if (value === null) {
+        return null;
+    }
+    return {
+        
+        'combined-string-enum': StringEnum | StringEnumTwoToJSON(value.combinedStringEnum),
```
prince-chrismc, line 64: This is still incorrectly generated

prince-chrismc:
I do not know how to split `datatype` when it's a combination

https://github.com/OpenAPITools/openapi-generator/blob/fe9636e9d93dd371d0defedee4dba962031fbb98/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache#L75
https://github.com/OpenAPITools/openapi-generator/blob/fe9636e9d93dd371d0defedee4dba962031fbb98/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache#L124

samples/client/petstore/typescript-fetch/builds/enum/models/StringOneOf.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function StringOneOfFromJSON(json: any): StringOneOf {
+    return StringOneOfFromJSONTyped(json, false);
+}
+
+export function StringOneOfFromJSONTyped(json: any, ignoreDiscriminator: boolean): StringOneOf {
+    if ((json === undefined) || (json === null)) {
+        return json;
+    }
+    return StringEnumFromJSONTyped(json, true) || StringEnumTwoFromJSONTyped(json, true);
```
prince-chrismc, line 39: Previously this came out as a object spread of the wrong type `{...FromJSONTyped(json, true), ...FromJSONTyped(json, true) }`

amakhrov: is it valid for an enum to include an empty string (`""`) as a value? If so, we might want to be stricter here and use `??` instead of `||`

amakhrov: however, apart from an empty string, this `||` will never work at all. `StringEnumFromJSONTyped` always returns some value (because a null check is done just before it - line 36), so the second part, `StringEnumTwoFromJSONTyped` will never be executed

amakhrov (commented):
@prince-chrismc nice work here!

Could you please add an example of `oneOf` with object types, not string enums? I believe it is a more common case, and will better surface potential problems, if any

samples/client/petstore/typescript-fetch/builds/enum/models/StringOneOf.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function StringOneOfToJSON(value?: StringOneOf | null): any {
+    if (value === undefined) {
+        return undefined;
+    }
+    if (value === null) {
+        return null;
+    }
+    return StringEnumToJSON(value as StringEnum) || StringEnumTwoToJSON(value as StringEnumTwo);
```
amakhrov, line 49: the second part after `||` will never work.
It's not a big deal for string enums, which don't really convert values and only do type assertion. However, for more complex data types it's a potential problem - right?

prince-chrismc: It didn't cross my mind, our docs only use string enums. I test it out tomorrow and see!

samples/client/petstore/typescript-fetch/builds/enum/models/StringOneOf.ts
prince-chrismc: `""` are valid in OAS so `??` is required ... it's also a problem for boolean and numbers so good catch.

I wrongly assumed casting `json as {{classname}}` would return falsy but that's not the case!

prince-chrismc:
> Could you please add an example of `oneOf` with object types, not string enums? I believe it is a more common case, and will better surface potential problems, if any

Can you take a look at my commit https://github.com/OpenAPITools/openapi-generator/pull/10017/commits/4b40ca203d3d47b2814559ccab76bcda6a5b73bf I got the following error

```
Root: Y:\repository\openapi-generator
[pool-1-thread-7] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\sagas-and-records)▒
[pool-1-thread-9] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\with-interfaces)▒
[pool-1-thread-2] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\default)▒
[pool-1-thread-8] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\typescript-three-plus)▒
[pool-1-thread-5] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\multiple-parameters)▒
[pool-1-thread-6] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\prefix-parameter-interfaces)▒
[pool-1-thread-3] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\enum)▒
[pool-1-thread-1] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\default-v3.0)▒
[pool-1-thread-11] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\without-runtime-checks)▒
[pool-1-thread-4] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\es6-target)▒
[pool-1-thread-10] Generating typescript-fetch (outputs to Y:\repository\openapi-generator\samples\client\petstore\typescript-fetch\builds\with-npm-version)▒
[pool-1-thread-3] Generation failed for typescript-fetch: (RuntimeException) Could not process model 'ObjectEnum'.Please make sure that your schema is correct!
java.lang.RuntimeException: Could not process model 'ObjectEnum'.Please make sure that your schema is correct!
        at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:499)
        at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:875)
        at org.openapitools.codegen.cmd.GenerateBatch$GenerationRunner.run(GenerateBatch.java:232)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.NullPointerException
        at org.openapitools.codegen.utils.ModelUtils.isBinarySchema(ModelUtils.java:647)
        at org.openapitools.codegen.utils.ModelUtils.isFileSchema(ModelUtils.java:659)
        at org.openapitools.codegen.languages.TypeScriptFetchClientCodegen.getTypeDeclaration(TypeScriptFetchClientCodegen.java:295)
        at org.openapitools.codegen.languages.TypeScriptFetchClientCodegen.addAdditionPropertiesToCodeGenModel(TypeScriptFetchClientCodegen.java:305)
        at org.openapitools.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:2659)
        at org.openapitools.codegen.languages.TypeScriptFetchClientCodegen.fromModel(TypeScriptFetchClientCodegen.java:449)
        at org.openapitools.codegen.languages.TypeScriptFetchClientCodegen.fromModel(TypeScriptFetchClientCodegen.java:37)
        at org.openapitools.codegen.DefaultGenerator.processModels(DefaultGenerator.java:1250)
        at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:494)
```

prince-chrismc:
❤️ Thank you for the review, it's clear after 6 months I am just starting my typescript journey... 8yrs of C++ was easier 🤣 

Greatly appreciate your help!

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
```diff
@@ -1,10 +1,16 @@
 {{>modelEnumInterfaces}}
 
-export function {{classname}}FromJSON(json: any): {{classname}} {
+export function {{classname}}FromJSON(json: any): {{classname}} | null {
+    if ((json === undefined) || (json === null)) {
+        return json;
+    }
+    if(!Object.values({{classname}}).includes(json as {{classname}})) {
+        return null;
+    }
```
prince-chrismc, line 9: Are these runtime checks allowed?

amakhrov:
> Can you take a look at my commit 4b40ca2 I got the following error

Values of `enum` cannot be objects.

What I meant was trying `oneOf` with regular object types, not with enums at all.

prince-chrismc:
> What I meant was trying `oneOf` with regular object types

They do not work sadly, they also fall victim to https://github.com/OpenAPITools/openapi-generator/pull/10017#issuecomment-885211922

amakhrov:
> hey do not work sadly, they also fall victim to #10017 (comment)

I feel it means the serialization/deserialization logic should be loosened (type-wise) - the code should try to (de-)serialize all possible properties from **intersection** (`A & B`), despite the fact the real type we work with is the union `A | B`.

prince-chrismc:
I am a little out of my depth, but AFAIK It does that for properties... it looks like the "type" is not well parsed... but I did not look into that section of the tool.

However your suggestion makes sense... we originally user the angular generator and that had no issue... and the templates seem much less complicated https://github.com/OpenAPITools/openapi-generator/blob/bf89f40263c6a57a76697c551d2c30c651812f7a/modules/openapi-generator/src/main/resources/typescript-angular/modelOneOf.mustache

amakhrov:
The main difference between typescript-fetch and typescript-angular (apart from using a different underlying http library, of course :) ) is that typescript-fetch produces code for run-time serializing/deserializing. When doing so for nested data structures, it needs to understand at run time which model it's handling at each point, in order to call a corresponding serializer.

On the other hand, code generated by typescript-angular passes the api data through as is, so it doesn't face the same challenge with oneOf schemas.

prince-chrismc:
Is it possible to move ahead with this PR? 🙏 Sadly if the work gets anymore complicated I wouldn't be able to help fix the issue

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
```diff
@@ -1,10 +1,16 @@
 {{>modelEnumInterfaces}}
 
-export function {{classname}}FromJSON(json: any): {{classname}} {
+export function {{classname}}FromJSON(json: any): {{classname}} | null {
```
amakhrov, line 3: this is gonna be a breaking change. if an enum property is marked as required / non-nullable in the spec, all consumers expect that it's defined.
after this change some apps will likely stop compiling.

why do yo need it anyway? you can keep the original code here  - return the enum value as is, without any processing

modules/openapi-generator/src/main/resources/typescript-fetch/modelOneOf.mustache
```diff
@@ -31,7 +31,7 @@ export function {{classname}}FromJSONTyped(json: any, ignoreDiscriminator: boole
     }
 {{/discriminator}}
 {{^discriminator}}
-    return { {{#oneOf}}...{{{.}}}FromJSONTyped(json, true){{^-last}}, {{/-last}}{{/oneOf}} };
+    return {{#oneOf}}{{{.}}}FromJSONTyped(json, true){{^-last}} || {{/-last}}{{/oneOf}};
```
amakhrov, line 34: this works for enums - but breaks the object models (the original code made more senses for objects).
perhaps need some condition here to process enums and objects differently.  might need to create a flag for that in the java model, if it's not available yet

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
prince-chrismc: Typescript gave a me a warning about the implicit return null. I believe it can be removed

amakhrov: that's because this PR changed the code inside the function, narrowing down `json` type to `null` before returning it

modules/openapi-generator/src/main/resources/typescript-fetch/modelOneOf.mustache
prince-chrismc: I think my lack of understanding is preventing us from moving forward so please bare with me.

I added an example of a `oneOf` using two objects

https://github.com/OpenAPITools/openapi-generator/pull/10017/files#diff-5a7c250c7d123900a98f6cb955465d37c8a93c4744a5d7e3798340fe76dbbd63

Why was is template not used to render the object?

I think it used https://github.com/OpenAPITools/openapi-generator/blob/bf89f40263c6a57a76697c551d2c30c651812f7a/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache#L78 instead

amakhrov: great question...

amakhrov: Ah, right. It's because your `ObjectCombinedOneOf` model is not `oneOf`. It's a regular object model with a single property.
This property, in turns, is defined as `oneOf` - but since it's inlined, there is no a separate top-level definition for that.
Probably try something like:
```
ObjectCombinedOneOf:
    oneOf:
        - $ref: '#/components/schemas/PatternObject'
        - $ref: '#/components/schemas/PatternObjectTwo'          
```

samples/client/petstore/typescript-fetch/builds/one-of/models/ObjectCombinedOneOf.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function ObjectCombinedOneOfToJSON(value?: ObjectCombinedOneOf | null): any {
+    if (value === undefined) {
+        return undefined;
+    }
+    if (value === null) {
+        return null;
+    }
+    return PatternObjectToJSON(value as PatternObject) ?? PatternObjectTwoToJSON(value as PatternObjectTwo);
```
prince-chrismc, line 49: Finally 🎉 Got that trouble code generated!

Before it was `{ ...TypeOneFromJson(value), ...TypeTwoFromJson(value) }` which was actually an all of? 🤔

AFAIK, With these changes we now have "its was completely parsed as Type one" or "its not TypeOne and completely parsed as TypeTwo"

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
prince-chrismc: Playing around with this more... it previously returned a value that was no valid. so if you passed `7` and your enum was `[0, 1, 2]` you would get back 7 which is not the right type.

modules/openapi-generator/src/main/resources/typescript-fetch/modelOneOf.mustache
prince-chrismc:
```suggestion
    return {{#oneOf}}{{{.}}}FromJSONTyped(json, true){{^-last}} ?? {{/-last}}{{/oneOf}};
```

samples/client/petstore/typescript-fetch/builds/one-of/models/ObjectCombinedOneOf.ts
amakhrov: > we now have "its was completely parsed as Type one" or "its not TypeOne and completely parsed as TypeTwo"

erhm, not exactly. `TypeOneFromJson` only returns `null` if input is null, otherwise it returns an object So the second parsing (`PatternObjectTwoToJSON`) will never happen

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
amakhrov: well, we trust the api response matches the spec. otherwise the code is gonna break anyway, since we don't do full shape verification

prince-chrismc:
🤔 Hmm

> since we don't do full shape verification

It seems like that would be required for this to work out as you point out in https://github.com/OpenAPITools/openapi-generator/pull/10017#discussion_r688746670 other wise I've made wierd code worse 🤣 

There's the fact

> typescript-fetch produces code for run-time serializing/deserializing https://github.com/OpenAPITools/openapi-generator/pull/10017#issuecomment-894415995

Should continue down this rabbit hole? Or is this too aggressive of a change?

modules/openapi-generator/src/main/resources/typescript-fetch/modelEnum.mustache
prince-chrismc: The challenge is older clients being used with newer backends which just weren't aware of the existence of such values

amakhrov: I would say, `null` is as unexpected as a bogus enum value (if the property is not marked as nullable).
Again, i honestly believe it's way out of scope of the generated api client.

amakhrov:
Probably the most correct way to address your case is to handle primitive values differently from objects.
> perhaps need some condition here to process enums and objects differently. might need to create a flag for that in the java model, if it's not available yet

So object models will keep deserializing using an object spread (`{...oneOfFirstFromJson(data), ...oneOfSecondFromJson(data)}`), while enums will just pass the value through as is

prince-chrismc:
Thanks for all you help.

PR #10021, opened 2021-07-23
[JAVA][FEIGN] Automatically retry request that fail due to a 401 or 403

hugo-ma-alves:
Hello

This PR adds the capability of automatically retry the requests that fail due to a 401 or 403. Sometimes these errors indicate that the token has expired.
This may happen if we have a valid token in memory, but it expires before the request is sent. After reaching the server the request will fail either with a 401 or 403.

For this case the Feign client will renew the token and retry to send the request one more time. If it fails on the second attempt it thrown an exception as before.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/Java/libraries/feign/ApiClient.mustache
+++ b/modules/openapi-generator/src/main/resources/Java/libraries/feign/ApiClient.mustache
@@ -32,9 +32,18 @@ import feign.form.FormEncoder;
 import feign.jackson.JacksonDecoder;
 import feign.jackson.JacksonEncoder;
 import feign.slf4j.Slf4jLogger;
-import {{invokerPackage}}.auth.*;
+import {{invokerPackage}}.auth.HttpBasicAuth;
+import {{invokerPackage}}.auth.HttpBearerAuth;
+import {{invokerPackage}}.auth.ApiKeyAuth;
+
 {{#hasOAuthMethods}}
+import {{invokerPackage}}.auth.ApiErrorDecoder;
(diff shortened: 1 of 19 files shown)


modules/openapi-generator/src/main/resources/Java/libraries/feign/ApiClient.mustache
```diff
@@ -55,6 +64,10 @@ public class ApiClient {
                 .client(new OkHttpClient())
                 .encoder(new FormEncoder(new JacksonEncoder(objectMapper)))
                 .decoder(new JacksonDecoder(objectMapper))
+                {{#hasOAuthMethods}}
+                .errorDecoder(new ApiErrorDecoder())
+                .retryer(new Retryer.Default(0, 0, 2))
```
borsch, line 69: probably make sense to extract this to instance variable or protected method so that end-user can change/override this values

hugo-ma-alves: I believe the end-user is already able to customize the feignBuilder. Either by using the getFeignBuilder that retrieves the default builder, or he can override it with the setFeignBuilder.

Do you think I should add a specific method to disable/enable the retryer?

wing328: @hugo-ma-alves My take is to try with what you've so far. Thanks for the PR.

@borsch thanks for the review.

PR #10066, opened 2021-07-31
Change configuration for typescript-fetch

### LINKED ISSUE #10065: [BUG] Change configuration parameters for typescript-fetch

Yes. Previously, we used to be able to specify how username and password are fetched during API calls. The `ConfigurationParameters` was defined as:

```typescript
interface ConfigurationParameters {
  // other fields
  username?: string | () => string;
  password?: string | () => string;
}
```

In this previous interface, we used to be able to specify configuration as follows:
```typescript
const configuration = new Configuration({
  username: () => getUsernameFromLocalStorage(),
  password: () => getPasswordFromLocalStorage(),
});
```

But in the new versions, the `ConfigurationParameters` interface has been changed, where it only accepts a string for username and password:
```typescript
interface ConfigurationParameters {
  // other fields
  username?: string;
  password?: string;
}
```
[Link to the generated code](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-fetch/builds/default/runtime.ts#L130)

This has made it harder to login and logout the users as login means we specify new username and password and logout means we delete any username and password.

A solution would be to revert back to the old `ConfigurationParameters` interface.

Another solution would be to generate Configuration like it is on [`angular`'s generated code](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/typescript-angular-v9-provided-in-root/builds/default/configuration.ts#L3):
```typescript
export interface ConfigurationParameters {
    /**
     *  @deprecated Since 5.0. Use credentials instead
     */
    apiKeys?: {[ key: string ]: string};
    username?: string;
    password?: string;
    /**
     *  @deprecated Since 5.0. Use credentials instead
     */
    accessToken?: string | (() => string);
    basePath?: string;
    withCredentials?: boolean;
    encoder?: HttpParameterCodec;
    /**
     * The keys are the names in the securitySchemes section of the OpenAPI
     * document. They should map to the value used for authentication
     * minus any standard prefixes such as 'Basic' or 'Bearer'.
     */
    credentials?: {[ key: string ]: string | (() => string | undefined)};
}
```
Here, `username` and `password` have been deprecated and use only string, but `credentials` can use functions, which allows us to specify username and password when api is being called.

I have been using this hack to get around this problem:

```typescript
const httpBasicInterceptor = new (class implements Middleware {
  async pre(context: RequestContext): Promise<FetchParams | void> {
    const username = AuthStorage.getUsername();
    const password = AuthStorage.getPassword();

    const oldHeaders = context.init.headers;
    let newHeaders;

    const AUTHORIZATION_KEY = 'Authorization';
    const basicAuth = 'Basic ' + btoa(`${username}:${password}`);

    if (oldHeaders !== undefined) {
      if (typeof oldHeaders == 'object') {
        newHeaders = {
          ...oldHeaders,
          [AUTHORIZATION_KEY]: basicAuth,
        };
      } else if (Array.isArray(oldHeaders)) {
        newHeaders = oldHeaders as string[][];
        newHeaders.push([AUTHORIZATION_KEY, basicAuth]);
      } else {
        newHeaders = oldHeaders as Headers;
        newHeaders.append(AUTHORIZATION_KEY, basicAuth);
      }
    }

    return {
      url: context.url,
      init: {
        ...context.init,
        headers: newHeaders,
      },
    };
  }
})();
const privateConfigParameters: ConfigurationParameters = {
  ...basicConfigParameters,
  middleware: [httpBasicInterceptor],
  username: '',
  password: '',
};
const privateConfig = new Configuration(privateConfigParameters);
```

But this is complex and harder to read. Also, there might be some circumstances where this solution might fail to work.

## Additional context

No other context required to understand this problem.

kpldvnpne:
fixes #10065
 
### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
@@ -121,8 +121,8 @@ export interface ConfigurationParameters {
     fetchApi?: FetchAPI; // override for fetch implementation
     middleware?: Middleware[]; // middleware to apply before/after fetch requests
     queryParamsStringify?: (params: HTTPQuery) => string; // stringify function for query strings
-    username?: string; // parameter for basic security
-    password?: string; // parameter for basic security
+    username?: string | (() => string); // parameter for basic security
+    password?: string | (() => string); // parameter for basic security
     apiKey?: string | ((name: string) => string); // parameter for apiKey security
     accessToken?: string | Promise<string> | ((name?: string, scopes?: string[]) => string | Promise<string>); // parameter for oauth2 security
     headers?: HTTPHeaders; //header params we want to use on every request
@@ -149,11 +149,19 @@ export class Configuration {
     }
 
     get username(): string | undefined {
-        return this.configuration.username;
+        const username = this.configuration.username;
+        if (typeof username == 'function') {
+            return username();
+        }
+        return username;
     }
 
     get password(): string | undefined {
-        return this.configuration.password;
+        const password = this.configuration.password;
+        if (typeof password == 'function') {
+            return password();
+        }
+        return password;
     }
 
     get apiKey(): ((name: string) => string) | undefined {


macjohnny:
@kpldvnpne thanks for your contribution

Please re-generate the samples according to https://github.com/OpenAPITools/openapi-generator/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Issue #10020, opened 2021-07-23
[BUG] [typescript-angular] Add "header" to reserved words


JohnTraDolta:
##### Description
Having a property in the swagger.json called "header". This result in a client with a parameter called header, later in that function a variable with the same name is instantiated

##### Solution 
Add header as a reserved word to enable the auto renaming functionality

JohnTraDolta:
I can do a pr if anyone can point me in the right direction. 👍

wing328:
@JohnTraDolta thanks for offering help to fix this. Can you please PM me via Slack when you've time?

https://join.slack.com/t/openapi-generator/shared_invite/enQtNzAyNDMyOTU0OTE1LTY5ZDBiNDI5NzI5ZjQ1Y2E5OWVjMjZkYzY1ZGM2MWQ4YWFjMzcyNDY5MGI4NjQxNDBiMTlmZTc5NjY2ZTQ5MGM

PR #10091, opened 2021-08-04
Fix issue 10020: [BUG] [typescript-angular] Add "header" to reserved words

JohnTraDolta:
Renamed the local variabels as agreed upon with @wing328, as a solution to #10020


wing328:

wing328:

amakhrov (approved):
Thanks for your contribution @JohnTraDolta !
I like the approach here.

There is a smallish issue - some of the original variables names are listed as "reveserved" [here](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java#L114). But the new names are not listed. So it's possible to get a conflict with those now (although rather unlikely, it would be a regression).

However, perhaps it would be better to have a separate cleanup PR across various typescript generators

PR #10119, opened 2021-08-10
[csharp-netcore] Fixed typo in httpclient generated code comment

Havunen:
Fixes typo in csharp-netcore generated code comment


wing328:
@Havunen thanks for the fix. I'll update the samples in another commit.

PR #10152, opened 2021-08-13
[CSharp-Netcore][Go] Optimize time taken to generate CSharp code

vvb:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

This PR greatly reduces the time taken for Csharp-netcore code generation.
toModelName() method is invoked numerous number of times and it was accummulating to a lot of time.
This PR adds memoization to CSharp and DefaultCodeGen toModelName() methods.
This has reduced the C# SDK generation time for our rather large openapi spec from ~2 hours to 4 mins.
This has reduced the Golang SDK generation time for our rather large openapi spec from ~1h30m hours to 4 mins.

The generated code is the same as before.

spec: https://cdn.intersight.com/components/an-apidocs/1.0.9-4430/model/intersight-openapi-v3-1.0.9.4430.yaml


Blackclaws:
Just out of curiosity if you're seeing generation times of 2 hours how large is your api definition file? I've been running the generator on what I would consider intermediate sized apis and it never took more than a minute.

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -2322,6 +2325,12 @@ public String toApiName(String name) {
      */
     @Override
     public String toModelName(final String name) {
+        if (schemaKeyToModelNameCache.containsKey(name)) {
+            return schemaKeyToModelNameCache.get(name);
+        }
+
+        schemaKeyToModelNameCache.put(name, camelize(modelNamePrefix + "_" + name + "_" + modelNameSuffix));
```
Blackclaws, line 2332: DRY. Assign ```
camelize(modelNamePrefix + "_" + name + "_" + modelNameSuffix)```
 to a variable and use it for both the add to cache and the return.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -1058,6 +1068,8 @@ public String toModelName(String name) {
             name = "model_" + name; // e.g. 200Response => Model200Response (after camelize)
         }
 
+        schemaKeyToModelNameCache.put(origName, camelize(name));
```
Blackclaws, line 1071: See previous comment

vvb:
> Just out of curiosity if you're seeing generation times of 2 hours how large is your api definition file? I've been running the generator on what I would consider intermediate sized apis and it never took more than a minute.

Our spec is very large, over 350k lines, 5000+ model objects. You can check it out at https://cdn.intersight.com/components/an-apidocs/1.0.9-4430/model/intersight-openapi-v3-1.0.9.4430.yaml

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
vvb: Done

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractCSharpCodegen.java
vvb: Done

Blackclaws:
Ok, yeah that is indeed a spec size that not a lot of people will have. That being said I have no issue with this pull request and it looks good to me, performance enhancements are always welcome even if they are mostly relevant to edge cases :)

Blackclaws:
I'm wondering if this seems to be an issue in multiple generators, whether this caching shouldn't be going into the default codegen somehow

vvb:
> I'm wondering if this seems to be an issue in multiple generators, whether this caching shouldn't be going into the default codegen somehow

so my initial fix was that - https://github.com/OpenAPITools/openapi-generator/pull/10149/commits/84525659a6a45bf5763442559f40a673d413a6ce But after some pointers from @wing328 it was clear that toModelName() is also invoked at many other places and so a better fix was to memoize toModelName itself.

devhl-labs:
If it took two hours before, how long does it take after this change?

wing328:
> I'm wondering if this seems to be an issue in multiple generators, whether this caching shouldn't be going into the default codegen somehow

Good point. I'll see what I can do to address the problem in the default codegen.

wing328:
> If it took two hours before, how long does it take after this change?

Down to a few minutes if I'm not mistaken.

vvb:
> If it took two hours before, how long does it take after this change?

1h56m -> 4m

wing328:
LGTM. I may submit another PR later to try to address similar problem for all generators.

PR #10157, opened 2021-08-15
modified tow pom.xml

i-Taozi:
We noticed a problem with two pom.xml files in this project.
Each of them has a SNAPSHOT dependency. The  SNAPSHOT dependency will require downloading the latest component every time a build is made, which will definitely increase the build time. Therefore, we have modified the versions of the two dependency versions as follows:

①
Relative paths:  openapi-generator/samples/server/petstore/jaxrs-cxf-test-data/pom.xml
Dependency: { Name: org.openapitools/openapi-generator, Version Specifier: 5.2.0-SNAPSHOT }   ------> 5.2.0

②
Relative paths:  openapi-generator/samples/server/petstore/kotlin/vertx/pom.xml
Dependency: {Name: org.jetbrains.kotlinx/kotlinx-coroutines-core, Version Specifier: RELEASE}  ------> 1.5.1-native-mt

### DIFF
--- a/samples/server/petstore/jaxrs-cxf-test-data/pom.xml
+++ b/samples/server/petstore/jaxrs-cxf-test-data/pom.xml
@@ -233,7 +233,8 @@
     <dependency>
       <groupId>org.openapitools</groupId>
       <artifactId>openapi-generator</artifactId>
-      <version>5.2.0-SNAPSHOT</version>
+	  <!-- <version>5.2.0-SNAPSHOT</version>-->
+      <version>5.2.0</version>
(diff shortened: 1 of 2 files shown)


wing328:
@i-Taozi thanks for the PR. Please update the mustache template files in https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources instead and follow step 3 in the PR checklist to update the samples. 

Let us know if you need any help.

wing328:
Closing it as there's no update. Reply if you need help.

PR #10176, opened 2021-08-17
[kotlin] Recognize more json media types in response

sheepdreamofandroids:
Instead of matching the mediatype against "application/json" it now matches with "application/*json" so vendor specific json mediatypes like "application/vnd.shopping.cart.v2+json" will also be recognized and deserialized. It doesn't map mediatypes to different models.
This would accept malformed mediatypes as well but that doesn't seem harmful.
I'm using a similarly patched template in a project.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
    @jimschubert
    @dr4ke616
    @karismann
    @Zomzog
    @andrewemery
    @4brunu
    @yutaka0m

### DIFF
--- a/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
+++ b/modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
@@ -151,8 +151,10 @@ import com.squareup.moshi.adapter
                     }
                 }.build()
             }
+            // recognize custom json variants like "application/vnd.shopping.cart.v2+json"
+            mediaType.startsWith("application/") && mediaType.endsWith("json") ->
             {{#jvm-okhttp3}}
-            mediaType == JsonMediaType -> RequestBody.create(
+            RequestBody.create(
                 {{#moshi}}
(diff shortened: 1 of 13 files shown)


sheepdreamofandroids:
I don't think the failure in AppVeyor is related to my changes.

4brunu:
Hi, thanks for opening this PR.
The CI failure is not related to this PR.
I have no experience vendor specific json media types, so it's hard for me to review it.
Looking at the code it makes sense, and I understand the reasoning behind it.
Did you tested this with any vendor specific json media type?

sheepdreamofandroids:
Hi @4brunu . My only experience with vendor specific mediatypes is an in-house api using one.

There is some documentation here: https://www.rfc-editor.org/rfc/rfc6838.html#section-4.2.8 which suggests that the used condition should work. Alternatively maybe even only checking for a "/json" or "+json" suffix would be enough. But I prefer another PR if that is ever necessary.

This PR is actually a subset of the changes we made to the templates that are running in production here. The other changes are about the same change for the request body and sending requests fully asynchronous. Those will be separate PR's.

sheepdreamofandroids:
I see that I made a mistake when copying some of the changes. 
I'll create a new PR that corrects those.

Issue #10175, opened 2021-08-17
[BUG] [cpp-ue4] formParams not handled in json requests


leith-bartrich:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

When generating ue4 client code, formParams generated for a standard Django auth-token request endpoint do not generate usable code and instead erroneously print the error: "Form parameter ({{baseName}}) was ignored, cannot be used in JsonBody".

The correct behavior seems to instead be, to construct json object from the formParams and pass that as the body.  This satisfies Django's standard server implementation of this standard Django api.

e.g.

the development HEAD version generates this request code:

```
// Default to Json Body request
	if (Consumes.Num() == 0 || Consumes.Contains(TEXT("application/json")))
	{
		UE_LOG(Logfpipelite_client, Error, TEXT("Form parameter (username) was ignored, cannot be used in JsonBody"));
		UE_LOG(Logfpipelite_client, Error, TEXT("Form parameter (password) was ignored, cannot be used in JsonBody"));
		UE_LOG(Logfpipelite_client, Error, TEXT("Form parameter (token) was ignored, cannot be used in JsonBody"));
	}
```

When it seems it should instead be constructing a json object from the formParams.

I have a pull request authored that fixes this behavior and will submit it shortly.

e.g.

my proposed fix (to be submitted shortly) generates this request code:

```
	// Default to Json Body request
	if (Consumes.Num() == 0 || Consumes.Contains(TEXT("application/json")))
	{
            // Form parameters
            FString JsonBody;
            JsonWriter Writer = TJsonWriterFactory<>::Create(&JsonBody);
            Writer->WriteObjectStart();
            Writer->WriteIdentifierPrefix(TEXT("username"));
            WriteJsonValue(Writer, Username);
            Writer->WriteIdentifierPrefix(TEXT("password"));
[...]

PR #10177, opened 2021-08-17
[cpp-ue4] fix for generating formParams in json requests

leith-bartrich:
Issue 10175 describes a potential bug and potential remedy.  This pull request is the basic fix I referenced in the issue.

https://github.com/OpenAPITools/openapi-generator/issues/10175

Simply put:  if there is no body parameter defined, it tries to generate a json body from formParams.  If a formParam isn't required, it uses IsSet from TOptional<> before writing the parameter.

See the referenced issue for the example schema that seems to require this template fix.

To be clear, this isn't really tested against a wide array of OpenAPI schemas.  And there are further issues with the generator.  But as a first pass, this at least gets a standard Django api-token-auth to be accepted by a Django server, and doesn't ruin other end points in my sample schema.

tagging @Kahncode as requested.

I'd want some more eyes on this at the very least.  And my conceptual question from the issue still stands.  I presume this to be a template problem, rather than a generator problem.  But that might not be correct.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-operations-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-operations-source.mustache
@@ -170,28 +170,28 @@ void {{classname}}::{{operationIdCamelCase}}Request::SetupHttpRequest(const FHtt
 		HttpRequest->SetHeader(TEXT("Content-Type"), TEXT("application/json; charset=utf-8"));
 		HttpRequest->SetContentAsString(JsonBody);
 		{{/bodyParams.0}}
-        {{^bodyParams.0}}
-            // Form parameters
-            FString JsonBody;
-            JsonWriter Writer = TJsonWriterFactory<>::Create(&JsonBody);
-            Writer->WriteObjectStart();
-            {{#formParams}}
-                {{#required}}
-            Writer->WriteIdentifierPrefix(TEXT("{{baseName}}"));
-            WriteJsonValue(Writer, {{paramName}});
-                {{/required}}
-                {{^required}}
-            if ({{paramName}}.IsSet()){
-                Writer->WriteIdentifierPrefix(TEXT("{{baseName}}"));
-                WriteJsonValue(Writer, {{paramName}}.GetValue());
-            }
-                {{/required}}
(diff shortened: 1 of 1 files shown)


Kahncode:
Looks good to me.

wing328:
Submitted https://github.com/OpenAPITools/openapi-generator/pull/10194 to update the samples

Kahncode:
Hello, we actually have issues created from this pull request. I'd love to discuss it further.

As far as I understood, a request cannot have body and form params at the same time. But then again, form params should never be generated as json body.

The error happens when a request has a file param, as this cannot be translated into a json body.

What is the expected behavior according to you @leith-bartrich ?

Kahncode:
Hello @leith-bartrich. Thanks for the thorough answer.

There are very valid points here, in particular, no matter what the standard says, we have to support as many real world cases as we reasonably can. 

I fixed the issue by simply omitting the file parameters when performing the fallback to form parameters generating a json body.

I do think that, this is in fact a generator issue, as it should not give both form and body parameters in the same request, as you may either send a json payload or a binary form, and not both at the same time. But regardless, the template must handle it gracefully.

> Also, when this was written, the generated client code automatically preferred the JSON-RPC implementation when really an application should be able to choose based on application logic.

Absolutely great point. This could be made as an option in the generator, although nobody has requested it. There are also many formats such as xml that are not supported in this template.

> My question would be; do you have a schema that says the following: This endpoint only takes a JSON-RPC request, and that request contains a file parameter. If so, I wonder if that's even a legal schema endpoint?

Yes, hence the error with your change. Any RPC with a file parameter as form parameter and no body parameters would attempt to compile illegal code.

https://github.com/OpenAPITools/openapi-generator/pull/15068

I improved the template to only attempt to fill the body with the form parameters if no body parameters are defined. Alongside ignoring files (and printing an error) altogether, I believe this should solve most real world cases.

If you have the chance, please review those changes and perhaps confirm that they work with your use case, as I would not want to break anything.

https://github.com/OpenAPITools/openapi-generator/pull/15068

PR #10188, opened 2021-08-18
[dart-dio-next] improve api client

ahmednfwela:
Improves api client generation

This PR generalizes many things that were hardcoded in the `dart-dio-next` generator, and introduces some improvements and bug fixes

1. [General] Added `isContainer` to `CodegenResponse` and called the missing `setUniqueItems`
2. [General] Changed `debugModels`, `debugOperations` behavior to debug the files AFTER post processing, making sure the developer sees the actual hash that is being fed to mustache files.
3. mapped `file` and `binary` to `MultipartFile` instead of `Uint8List` and let dio handle the rest
4. use `inlineEnums` when possible instead of just the underlying type
5. improved & generalized serializing response types
6. introduce `connectTimeout` and `receiveTimeout` as  parameters to the api client file
7. split repetitive generation into separate mustache files

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


ahmednfwela:
blocked by https://github.com/OpenAPITools/openapi-generator/issues/10189

ahmednfwela:
CC @jaumard (2018/09) @josh-burton (2019/12) @amondnet (2019/12) @sbu-WBT (2020/12) @kuhnroyal (2020/12) @agilob (2020/12)

ahmednfwela:
@wing328 this PR also includes improvements to the core modules

samples/openapi3/client/petstore/dart-dio-next/petstore_client_lib_fake/doc/FakeApi.md
```diff
@@ -520,7 +520,7 @@ final int int32 = 56; // int | None
 final int int64 = 789; // int | None
 final double float = 3.4; // double | None
 final String string = string_example; // String | None
-final Uint8List binary = BINARY_DATA_HERE; // Uint8List | None
+final MultipartFile binary = BINARY_DATA_HERE; // MultipartFile | None
```
kuhnroyal, line 523: This is probably not correct. That is why the special handling existed.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
```diff
@@ -116,35 +119,15 @@ class {{classname}} {
...
-      headers: _response.headers,
-      isRedirect: _response.isRedirect,
-      requestOptions: _response.requestOptions,
-      redirects: _response.redirects,
-      statusCode: _response.statusCode,
-      statusMessage: _response.statusMessage,
-      extra: _response.extra,
-    );{{/returnType}}{{^returnType}}
-    return _response;{{/returnType}}
+    return _response;{{/returnType}}{{#returnType}}{{#responses}}
```
kuhnroyal, line 124: Something go lost here?

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
```diff
@@ -43,11 +46,11 @@ class {{classname}} {
...
+    {{#required}}{{^isNullable}}{{^defaultValue}}required {{/defaultValue}}{{/isNullable}}{{/required}}{{>serialization/built_value/variable_type}}{{^isNullable}}{{^required}}?{{/required}}{{/isNullable}} {{paramName}}{{^isContainer}}{{>serialization/built_value/param_default_type}}{{/isContainer}},{{/isQueryParam}}{{#isHeaderParam}}
+    {{#required}}{{^isNullable}}{{^defaultValue}}required {{/defaultValue}}{{/isNullable}}{{/required}}{{>serialization/built_value/variable_type}}{{^isNullable}}{{^required}}?{{/required}}{{/isNullable}} {{paramName}}{{^isContainer}}{{>serialization/built_value/param_default_type}}{{/isContainer}},{{/isHeaderParam}}{{#isBodyParam}}
+    {{#required}}{{^isNullable}}required {{/isNullable}}{{/required}}{{>serialization/built_value/variable_type}}{{^isNullable}}{{^required}}?{{/required}}{{/isNullable}} {{paramName}},{{/isBodyParam}}{{#isFormParam}}
+    {{#required}}{{^isNullable}}required {{/isNullable}}{{/required}}{{>serialization/built_value/variable_type}}{{^isNullable}}{{^required}}?{{/required}}{{/isNullable}} {{paramName}},{{/isFormParam}}{{/allParams}}
```
kuhnroyal, line 53: The `{{^isNullable}}{{^required}}?{{/required}}{{/isNullable}}` change doesn't make sense to me.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
```diff
@@ -116,35 +119,15 @@ class {{classname}} {
...
-      data: _responseData,
-      headers: _response.headers,
-      isRedirect: _response.isRedirect,
-      requestOptions: _response.requestOptions,
-      redirects: _response.redirects,
-      statusCode: _response.statusCode,
-      statusMessage: _response.statusMessage,
-      extra: _response.extra,
-    );{{/returnType}}{{^returnType}}
-    return _response;{{/returnType}}
```
kuhnroyal, line 145: I would like the general try/catch/response handling to stay here, independent of built_value. If the `return_type` doesn't fit then we should fix that.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
```diff
@@ -116,35 +119,15 @@ class {{classname}} {
...
+    return _response;{{/returnType}}{{#returnType}}{{#responses}}
+{{#isDefault}}{{>serialization/built_value/deserialize_response_type}}{{/isDefault}}{{#is2xx}}{{>serialization/built_value/deserialize_response_type}}{{/is2xx}}{{/responses}}{{/returnType}}
   }
 
   {{/operation}}
 }
+{{#inlineEnums}}
+{{>serialization/built_value/enum_inline}}
```
kuhnroyal, line 131: These enums will cause collisions for some of my projects and likely for others as well.
I think enum handling for parameters should be a separate PR.

kuhnroyal (changes requested):

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
ahmednfwela: No, it's exactly the same as the old handling, I just moved it into a separate mustache file

samples/openapi3/client/petstore/dart-dio-next/petstore_client_lib_fake/doc/FakeApi.md
ahmednfwela: I think dio should be able to handle all files as MultipartFile, no matter where

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
ahmednfwela: but this generates the same result as the old way

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
kuhnroyal: Yes but there can be enums with the same name but different values for multiple operations in on API class.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
kuhnroyal: Probably because we have no example which is `nullable` but also `required`. This is however possible.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
ahmednfwela: I think we can solve that by prepending the name of the api class to the name of the enum, what do you think ?

kuhnroyal: Probably the name of the operation? That one might be long but I don't see another way.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartDioNextClientCodegen.java
```diff
@@ -278,78 +279,217 @@ public void postProcessModelProperty(CodegenModel model, CodegenProperty propert
         }
     }
 
+    private void fixEnumParam(String opName,IJsonSchemaValidationProperties param) {
```
josh-burton, line 282: any of these 'fix' methods need some nice documentation in my opinion. A simple description of what this method does and what if fixes.

ahmednfwela: will do, but these are to compensate for incorrect behavior in the core modules

josh-burton: ok, definitely in need of some good documentation then! Would be good to explain what they compensate for and at what point that could be removed.

ahmednfwela: I added some comments, let me know if more are needed

modules/openapi-generator/src/main/resources/dart/libraries/dio/serialization/built_value/simple_type.mustache
```diff
@@ -0,0 +1 @@
+{{#isEnum}}{{{enumName}}}{{/isEnum}}{{^isEnum}}{{{datatypeWithEnum}}}{{^datatypeWithEnum}}{{{dataType}}}{{/datatypeWithEnum}}{{/isEnum}}
```
kuhnroyal, line 1: This does not look specific to built_value. We should try to keep generic handling outside of the `built_value` folder so that it can be reused for the upcoming `json_serializable` support.

modules/openapi-generator/src/main/resources/dart/libraries/dio/serialization/built_value/variable_type.mustache
```diff
@@ -1 +1 @@
-{{#isContainer}}{{baseType}}<{{#isMap}}String, {{/isMap}}{{#items}}{{>serialization/built_value/variable_type}}{{/items}}>{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}{{/isContainer}}{{#isNullable}}?{{/isNullable}}
\ No newline at end of file
+{{#isContainer}}{{baseType}}<{{#isMap}}String, {{/isMap}}{{#items}}{{>serialization/built_value/variable_type}}{{/items}}>{{/isContainer}}{{^isContainer}}{{>serialization/built_value/simple_type}}{{/isContainer}}{{#isNullable}}?{{/isNullable}}
```
kuhnroyal, line 1: Same here, this is generic.

samples/openapi3/client/petstore/dart-dio-next/petstore_client_lib_fake/doc/FakeApi.md
kuhnroyal: This is not handled by `dio` this is handled by `built_value` serialization which doesn't have a serializer for `MultipartFile`. Not sure it works out of the box with `Uint8List` but one could add a custom Base64 serializer for it. This is not possible with `MultipartFile` as this is an abstraction around an async stream.

modules/openapi-generator/src/main/resources/dart/libraries/dio/api.mustache
kuhnroyal: I would like the general try/catch/response handling to stay here, independent of built_value. If the return_type doesn't fit then we should fix that. This block should be reusable for `json_serializable`.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/DartDioNextClientCodegen.java
```diff
@@ -278,78 +279,224 @@ public void postProcessModelProperty(CodegenModel model, CodegenProperty propert
...
-                    param.baseType = param.baseType.replace("Uint8List", "MultipartFile");
-                    op.imports.add("MultipartFile");
-
-                    if (SERIALIZATION_LIBRARY_BUILT_VALUE.equals(library)) {
-                        boolean skipFormModel = Boolean.parseBoolean(GlobalSettings.getProperty(CodegenConstants.SKIP_FORM_MODEL, "true"));
-                        if (param.isFormParam && param.isContainer && !skipFormModel) {
-                            // Because of skipFormModel=false, there is a model class generated which has
-                            // "BuiltList<Uint8List>" as property and it requires the correct
-                            // serializer imports to be added in order to compile.
-                            addBuiltValueSerializerImport("Uint8List");
```
kuhnroyal, line 303: This whole block is now missing, we don't have a sample for `skipFormModel` but I don't think it will compile atm.

kuhnroyal:
I really think you should strip the enum changes out of this PR so we can concentrate on the other changes.
This is already very complex and we are missing samples for a lot of the things that are being touched which makes these changes pretty risky.

ahmednfwela:
> I really think you should strip the enum changes out of this PR so we can concentrate on the other changes.
> This is already very complex and we are missing samples for a lot of the things that are being touched which makes these changes pretty risky.

I actually didn't intend for the enum changes at the start, but it came as a side effect for generalizing variable types with a single mustache

samples/openapi3/client/petstore/dart-dio-next/petstore_client_lib_fake/doc/FakeApi.md
ahmednfwela: I just noticed that there is a model "FormatTest" that has a `binary` property.
and looking at `csharp-netcore`, they have these type mappings
```java
typeMapping.put("file", "System.IO.Stream");
typeMapping.put("binary", "byte[]");
```
which for our case should be
```java
typeMapping.put("file", "MultipartFile");
typeMapping.put("binary", "Uint8List");
```

kuhnroyal: Yea but I think a we also have a form param in the sample which is `binary`, not sure I remember that correctly.

ahmednfwela: actually looking at it more:
`csharp-netcore` also generates `System.IO.Stream` for `binary` formats (seems that `binary` aliases into `file` for backward compatibility I think?)
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/csharp-netcore/OpenAPIClient-net5.0/src/Org.OpenAPITools/Model/FormatTest.cs#L143

which is also not correct, and looking at OAS3 specification: https://swagger.io/docs/specification/describing-responses/#response-that-returns-a-file

it appears that the `FormatTest` model itself is wrong here, as having an embedded property that returns `format: binary` is illegal.

So that leaves us with responses with ONLY a file as a response, and in that case the response shouldn't get de-serialized by anything at all, and instead let dio return the file stream.

similar to the download method here:
https://github.com/flutterchina/dio/blob/20506b496d41d7c2cc71d323e22634e99dc3ef21/dio/lib/src/entry/dio_for_native.dart#L84

ahmednfwela: @wing328 do you think this schema is in line with the specification?  https://github.com/OpenAPITools/openapi-generator/blob/158649aff9a87b13b448880d7ca6a4ac35fadcc2/modules/openapi-generator/src/test/resources/3_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1510

kuhnroyal: `file` is the legacy/OAS2 type and gets mapped to `binary` now or the other way around. 

> it appears that the FormatTest model itself is wrong here, as having an embedded property that returns format: binary is illegal.

This is probably right, I am all for removing this from the sample and replace it (if it is not already there) with `format: byte`. This should be done in a separate PR.

> So that leaves us with responses with ONLY a file as a response, and in that case the response shouldn't get de-serialized by anything at all, and instead let dio return the file stream.

Yes, we could do that and return a stream, ideally we give the user both options. One with `ResponseType.byte` and one `$operationDownload` with `ResponseType.stream`.  But this should also be a separate PR.

ahmednfwela:
I will try splitting them into multiple PRs, one for the core fixes, one for the generalization changes WITHOUT the inline enums, and one with them

Issue #10164, opened 2021-08-16
[BUG] [typescript-fetch] Compilation failure in typescript build


IodizedGabe:
#### Bug Report Checklist

- [ ] Have you provided a full/minimal spec to reproduce the issue?
- [ ] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

The `typescript-fetch` client fails to compile.

Compilation failure:
```
    src/ext/runtime.ts:152:9 - error TS2322: Type '((input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>) | undefined' is not assignable to type '(input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>'.
      Type 'undefined' is not assignable to type '(input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>'.

    152         return this.configuration.fetchApi;
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

##### openapi-generator version

Version: v5.2.1
Regression: Yes

##### Generation Details

CLI used:
```
docker run -v "${PWD}:/local" openapitools/openapi-generator-cli:v5.2.1 generate -i /local/<specfile> -g typescript-fetch
```

##### Steps to reproduce

See above

##### Related issues/PRs

This was introduced in: https://github.com/OpenAPITools/openapi-generator/commit/c05ec99182c80993a218a841853ba6412e6a6c16

This causes an error because this config value may be `undefined`:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L122

##### Suggest a fix
The fallback as seen here: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L122

Could be added to the getter.

wing328:
@IodizedGabe can you please submit a PR with the suggested fix? Thanks..

[...]

PR #10200, opened 2021-08-19
[typescript-fetch] fix definition of fetchApi in configuration (#10164)

javier-garcia-meteologica:
Fix definition of fetchApi in configuration to solve typescript compilation errors (#10164)

@wing328 @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny @topce @akehir @petejohansonxo  @amakhrov

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
@@ -138,7 +138,7 @@ export class Configuration {
     }
 
     get fetchApi(): FetchAPI {
-        return this.configuration.fetchApi;
+        return this.configuration.fetchApi || fetch;
     }
 
     get middleware(): Middleware[] {
--- a/samples/client/petstore/typescript-fetch/builds/default-v3.0/runtime.ts
+++ b/samples/client/petstore/typescript-fetch/builds/default-v3.0/runtime.ts
@@ -148,7 +148,7 @@ export class Configuration {
         return this.configuration.basePath != null ? this.configuration.basePath : BASE_PATH;
     }
 
-    get fetchApi(): FetchAPI {
+    get fetchApi(): FetchAPI | undefined {
         return this.configuration.fetchApi;
     }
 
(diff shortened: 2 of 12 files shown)


japrescott:
this wont work as it will fail everywhere else as fetchAPI could be undefined. We should revert back to defaulting to `window.fetch`;

`return this.configuration.fetchApi || window.fetch.bind(window);`

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -137,7 +137,7 @@ export class Configuration {
         return this.configuration.basePath != null ? this.configuration.basePath : BASE_PATH;
     }
 
-    get fetchApi(): FetchAPI {
+    get fetchApi(): FetchAPI | undefined {
         return this.configuration.fetchApi;
```
japrescott, line 141: this should be changed to `return this.configuration.fetchApi || window.fetch.bind(window);` and not typed as if it could be undefined

japrescott (changes requested):

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
javier-garcia-meteologica: But BaseApi [already provides a fallback here](https://github.com/javier-garcia-meteologica/openapi-generator/blob/fix_typescript_fetch_fetchApi/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L78), so that every call to `this.fetchApi` has the global `fetch` as a fallback.

If a default value is implemented in the configuration, that other fallback should be removed?

amakhrov: The previous PR (https://github.com/OpenAPITools/openapi-generator/pull/10072) had good reasons to avoid `window` , in order to support Node environment (as in universal rendering)

javier-garcia-meteologica: I've changed the code to address both issues. `fetchApi` returns always a defined `FetchAPI` and uses global `fetch` as fallback

javier-garcia-meteologica: Oh, I see  this has already been fixed in  #10216, I will close this PR

Issue #10205, opened 2021-08-19
[BUG] [cpp-ue4] response json parsing logging a error to unreal console when there is no actual parsing problem


leith-bartrich:
#### Bug Report Checklist

- [x] Have you provided a full/minimal spec to reproduce the issue?
- [x] Have you validated the input using an OpenAPI validator ([example](https://apidevtools.org/swagger-parser/online/))?
- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?
- [x] Have you searched for related issues/PRs?
- [x] What's the actual output vs expected output?
- [ ] [Optional] Sponsorship to speed up the bug fix or feature request ([example](https://github.com/OpenAPITools/openapi-generator/issues/6178))

##### Description

The cpp-ue4 template creates code that in turn, creates and logs errors about parsing errors, that are not actually parsing errors.

The actual error log statement is generated in the API source like so:

```c++
void OpenAPIOrgbuilderApi::HandleResponse(FHttpResponsePtr HttpResponse, bool bSucceeded, Response& InOutResponse) const
{
	InOutResponse.SetHttpResponse(HttpResponse);
	InOutResponse.SetSuccessful(bSucceeded);

	if (bSucceeded && HttpResponse.IsValid())
	{
		InOutResponse.SetHttpResponseCode((EHttpResponseCodes::Type)HttpResponse->GetResponseCode());
		FString ContentType = HttpResponse->GetContentType();
		FString Content;

		if (ContentType.IsEmpty())
		{
			return; // Nothing to parse
		}
		else if (ContentType.StartsWith(TEXT("application/json")) || ContentType.StartsWith("text/json"))
		{
			Content = HttpResponse->GetContentAsString();

			TSharedPtr<FJsonValue> JsonValue;
			auto Reader = TJsonReaderFactory<>::Create(Content);

			if (FJsonSerializer::Deserialize(Reader, JsonValue) && JsonValue.IsValid())
			{
				if (InOutResponse.FromJson(JsonValue))
					return; // Successfully parsed
			}
		}
		else if(ContentType.StartsWith(TEXT("text/plain")))
		{
			Content = HttpResponse->GetContentAsString();
			InOutResponse.SetResponseString(Content);
			return; // Successfully parsed
		}

[...]

PR #10221, opened 2021-08-21
[cpp-ue4] bug-fix issue 10205: response json parsing logging a error to unreal console when there is no actual parsing problem

leith-bartrich:
This is my proposed fix to #10205

In summary, rather than walk all vars, it walks required vars and non-required (optional) vars.  And within the required vars, it differentiates between writeOnly and non-writeOnly vars.  It allows optional or writeOnly vars to fail to parse (because they're absent) and not trigger a big red error log indication of a parsing failure.  To be clear:  The generated code should not yell in red about an error, when a server is actually following the schema.  The balance here, is that it tries to parse the vars anyway, but doesn't yell unless the schema truly indicates their absence is critical.

There are many deficiencies here.  e.g.  A TryGetJsonValue call doesn't differentiate between a parse failure versus the absence of a value, in its return.  Which forces a lowest common denominator solution here.  e.g. It's not necessarily providing much runtime feedback where it could.  However, I think both of those deficiencies are actually a design change, rather than a bug fix.  They could be handled in a different issue and and discussed there.

Here, I'm only seeking to keep the generated code from logging errors for missing vars, that probably should be missing.

I moved the comments to keep the code looking nicer, since my original bug submission.

This change hasn't been tested against a large array of schemas.  But it has been tested against the simple schema provided in the bug, against a dirt simple Django backend.

The code generated from the samples seems sane to me. But I'm not as familiar with that schema.  Of note:  the pet.yml model is really the only one that marks certain fields required for a response.  However, this is why I think it should instead be logging more useful information to Verbose and VeryVerbose on a per field basis, rather than declare a ParsingFailure on the whole response object.

tagging: @Kahncode

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


leith-bartrich:
A rebase to development HEAD and subsequent rebuild and generation shows the Docs and Samples CI issue might be with development HEAD rather than this request.

evidence after doing a rebase and generation locally:

```
git diff
diff --git a/samples/server/petstore/haskell-yesod/.openapi-generator/FILES b/samples/server/petstore/haskell-yesod/.openapi-generator/FILES
index 1f055028bf6..08edcd6331e 100644
--- a/samples/server/petstore/haskell-yesod/.openapi-generator/FILES
+++ b/samples/server/petstore/haskell-yesod/.openapi-generator/FILES
@@ -1,6 +1,5 @@
 .dir-locals.el
 .gitignore
-.openapi-generator-ignore
 README.md
 app/DevelMain.hs
 app/devel.hs
@@ -23,8 +22,5 @@ src/Settings.hs
 src/Settings/StaticFiles.hs
 stack.yaml
 static/.gitkeep
-test/Handler/PetSpec.hs
-test/Handler/StoreSpec.hs
-test/Handler/UserSpec.hs
 test/Spec.hs
 test/TestImport.hs
```

likely culprit: #10193

wing328:
@leith-bartrich the build failure issue due to outdated doc/samples has been fixed in the master. Please rebase or merge latest master into your branch to resolve the issue.

wing328:
FYI. https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword contains keywords to close an issue automatically when the PR is merged.

leith-bartrich:
Rebased, rebuilt, regenerated and pushed.  The docs and samples passed fine this time.  I'm unclear as to if the ci/circleci failure is meaningful.  It seems to be the parallel test that failed?

trying out your suggested auto issue link:
Fixes #10205 

Also, @wing328 , Do you want me to squash commits in this kind of pull requests?  Or leave them broken out?  I already squashed some silly ones that were just me fixing white-space mistakes.  But what about intermediate changes like moving comments around and such?

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
```diff
@@ -175,9 +175,20 @@ bool {{classname}}::FromJson(const TSharedPtr<FJsonValue>& JsonValue)
...
+    {{/isWriteOnly}}
+    {{^isWriteOnly}}
+	//Required and not write-only.  Its absence is likely a problem.
 	ParseSuccess &= TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
-    {{/vars}}
+    {{/isWriteOnly}}
+	{{/requiredVars}}
+	{{#optionalVars}}
+	//Optional.  Its absence would be okay.
+	TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
```
Kahncode: Normally this is handed by the TOptional used with optional vars, so this line should not have to be changed. Can you confirm?

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
```diff
@@ -175,9 +175,20 @@ bool {{classname}}::FromJson(const TSharedPtr<FJsonValue>& JsonValue)
...
 	bool ParseSuccess = true;
 
-    {{#vars}}
+	{{#requiredVars}}
+    {{#isWriteOnly}}
+	//Required but write-only. Its absence is reasonable when parsing JSON response.  
+	TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
+    {{/isWriteOnly}}
+    {{^isWriteOnly}}
+	//Required and not write-only.  Its absence is likely a problem.
```
Kahncode: I would rather not have the comments in the generated code to be honest.

Kahncode (changes requested):

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
```diff
@@ -175,9 +175,20 @@ bool {{classname}}::FromJson(const TSharedPtr<FJsonValue>& JsonValue)
 
 	bool ParseSuccess = true;
 
-    {{#vars}}
+	{{#requiredVars}}
```
Kahncode, line 178: WriteJson uses #vars and #required to perform something similar here, if possible i'd like to keep things consistent, so if it's possible to iterate on all the vars with #vars rather than #requiredVars and #optionalVars, this would put me more at ease, and be more consistent with the place those variables are declared in the first place. Alternatively, change where those variables are declared in the header, but I'd suggest the first approach.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
```diff
@@ -175,9 +175,20 @@ bool {{classname}}::FromJson(const TSharedPtr<FJsonValue>& JsonValue)
...
+    {{#isWriteOnly}}
+	//Required but write-only. Its absence is reasonable when parsing JSON response.  
+	TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
+    {{/isWriteOnly}}
+    {{^isWriteOnly}}
+	//Required and not write-only.  Its absence is likely a problem.
 	ParseSuccess &= TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
```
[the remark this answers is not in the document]
leith-bartrich: Just to clarify:  the red highlight on the & symbol  in the full &= operator is a syntax highlighting thing with GitHub.  If you look the generated sample for the petshop that is part of this commit, you'll see it does generate the full &= operator.

If that's not what you're commenting on, please clarify, I'm confused.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
leith-bartrich: Noted.  Will change and test.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
leith-bartrich: Noted.  Will change.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
leith-bartrich: Implemented request change and tested.  Works fine.  Will push shortly.

leith-bartrich:
Requested changes made, generated samples and docs, and pushed.

One outstanding request remains based on what may be a misunderstanding on syntax highlighting.

Testing locally on my codebase fixes the errant behavior in a satisfactory manner.

Of note: the petstore sample doesn't seem to show any change here because it doesn't seem to use any writeOnly fields in its components.  Which is likely why this case never came up.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
```diff
@@ -175,9 +175,19 @@ bool {{classname}}::FromJson(const TSharedPtr<FJsonValue>& JsonValue)
 
 	bool ParseSuccess = true;
 
-    {{#vars}}
+	{{#vars}}
+	{{#required}}
+	{{#isWriteOnly}}
+	TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
+	{{/isWriteOnly}}
+	{{^isWriteOnly}}
```
Kahncode, line 183:
```suggestion
	{{^isWriteOnly}} ParseSuccess &= {{/isWriteOnly}} TryGetJsonValue(*Object, TEXT("{{baseName}}"), {{name}});
```

Slightly more concise version, otherwise the change looks great.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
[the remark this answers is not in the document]
Kahncode: Indeed, ignore my previous comment, i'm not used to github reviews.

modules/openapi-generator/src/main/resources/cpp-ue4/model-source.mustache
leith-bartrich: change made locally, tested and pushed.

leith-bartrich:
Made requested change for conciseness.  Tested against local codebase.  Regenerated samples and docs.  Pushed.

Kahncode (approved):

Kahncode:
Looks good for me. Thanks for the contribution!

PR #10249, opened 2021-08-24
Libcurl CMake adaptations

ahmedyarub:
@zhemant  @ityuhui @michelealbano 

Remove default build type and library shared flag
Add OpenSSL using imported targets on supported CMake versions
PreTarget and PostTarget scripts where the former one is used to add new source files and run additional find_package() commands, and the latter can be used to add optional components such as whether to include building examples and also to customize the installation command.

Related PR which I'm going to close in favor of this later: https://github.com/kubernetes-client/c/pull/75

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [X] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master` (5.3.0), `6.0.x`

### DIFF
--- a/modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
+++ b/modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.6)
+cmake_minimum_required (VERSION 2.6...3.10.2)
 project (CGenerator)
 
 cmake_policy(SET CMP0063 NEW)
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
```diff
@@ -75,10 +81,15 @@ set(HDRS
 
 )
 
-# Add library with project file with projectname as library name
-add_library(${pkgName} SHARED ${SRCS} ${HDRS})
+include(PreTarget OPTIONAL)
+
+# Add library with project file with project name as library name
+add_library(${pkgName} ${SRCS} ${HDRS})
```
ityuhui, line 87: Why need remove `SHARED` ?

ahmedyarub: The SHARED/STATIC types should only be used if we know in advance that the library works when compiled as one and not as the other. By removing the flag I could add the -DBUILD_SHARED=FALSE flag to the CMake command and compile it as a static library which worked for me without any problem.
Is there any restriction for compiling as a static library?

PR #10262, opened 2021-08-25
[typescript-node] Set model default values

sarumont:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -309,7 +309,7 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
         return toApiFilename(name);
     }
 
-@Override
+    @Override
     protected void addAdditionPropertiesToCodeGenModel(CodegenModel codegenModel, Schema schema) {
         super.addAdditionPropertiesToCodeGenModel(codegenModel, schema);
         Schema additionalProperties = getAdditionalProperties(schema);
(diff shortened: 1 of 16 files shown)


wing328:
Can you please follow step 3 in the PR checklist to update the samples so that the CI can verify the change?

sarumont:
Updated - thought I had done that already!

macjohnny:
This change seems to set undefined als a default value for fields that are not optional, see the circle CI failure

sarumont:
Ah, yes...there was another mod we have in our fork of these templates. There are two options here, though:

1. set the default for `defaultValue` to Java `null` in the generator code, causing `{{#defaultValue}}` to only hit when it is specified in the OAPI spec.
2. only set the default value for non-required fields

We went with the first approach, which I think is the cleaner option (as you may want to set defaults for required fields). I'll make that mod here and revert later if the latter option is preferred.

macjohnny:
Option 1 sounds good

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -319,4 +319,37 @@ protected void addAdditionPropertiesToCodeGenModel(CodegenModel codegenModel, Sc
...
+            }
+            return null;
+        } else if (ModelUtils.isNumberSchema(p) || ModelUtils.isIntegerSchema(p)) {
+            if (p.getDefault() != null) {
+                return p.getDefault().toString();
+            }
+            return null;
+        } else if (ModelUtils.isStringSchema(p)) {
+            if (p.getDefault() != null) {
+                return "'" + (String) p.getDefault() + "'";
```
macjohnny, line 347: Nit: the string should be escaped

sarumont: That's copy-pasta from `AbstractTypeScriptCodegen` - I'll fix in both places. :)

sarumont: Actually, fixed in `super` and refactored the node codegen to call rather than copy.

macjohnny (approved):
LGTM

PR #10283, opened 2021-08-29
Documentation of Configuration/ConfigurationParameter for consolidated TypeScript generator

TiFu:
As requested by @rgomezp in #6341, I have added a bit more elaborate comments on the `ConfigurationParameters` and `Configuration` objects. 

Just an initial draft for now, but happy to already get some feedback on how we can further improve the documentation to make it more user-friendly.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

CC @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)


rgomezp:
This is great @TiFu  thank you.

Two comments:
1. You refer several times to the "spec", but where is this spec? As someone who is new to this library, I struggled to become familiar with the API for this generator since the docs didn't have any info on the specific configuration API. 
2. Could some of this info also be added to the documentation? Regardless, these new type annotations certainly help loads.

bodograumann:
The spec is the OpenAPI file that you use to generate the code, @rgomezp .
Maybe speaking about the OpenAPI definition would be more on-point, because spec can also mean the OpenAPI 2/3 specification itself.

modules/openapi-generator/src/main/resources/typescript/README.mustache
```diff
@@ -28,3 +28,45 @@ _unPublished (not recommended):_
...
+    baseServer: your_api.servers[0], // First server is default
+    authMethods: authConfig, // No auth is default
+    promiseMiddleware: [new Test()],
+}
+
+// Convert to actual configuration
+const config = your_api.createConfiguration(configurationParameters);
+
+// Use configuration with your_api
+const api = new your_api.PetApi(config);
```
wujibear, line 71: Would be great to also see how you would make a call off of this `api` instance.
It seems like we'd do something like `api.getUsers()`, but I'm getting tripped up by POST requests.

It seems like you'd do something like: 
```
// Seems like object arguments need to be namespaced, and that's based on the model name?
// If there's an option to not-namespace params I'd love to know it
api.createUser({createUserRequest: { email: 'email', password: 'password'}})
```

This seems to work for me if the SDK is making a call from one localhost port to another, but in my prod env it strips out the request payload. I imagine that's because I'm missing some kind of config option, and seeing how you do this would be super helpful information.

TiFu: I have added another couple lines with an example in the latest commit. Hope that explains it a bit more. There's also some examples in the `tests` folder under `samples/openapi3/client/petstore/typescript/` - hope this helps with debugging your problem.

TiFu:
Failures are unrelated. 

Given that this only touches files in the consolidated typescript generator & only changes comments, will go ahead and merge this.

PR #10296, opened 2021-08-31
[Rust][Client] Default version to spec version if not provided as property

Foorack:
If no version number is provided as property then it will try identify the version number automatically from the OpenAPI specification. If no version exist there either then it will still fallback to "1.0.0". This code is taken [from the typescript generator](https://github.com/OpenAPITools/openapi-generator/blob/5.2.x/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractTypeScriptClientCodegen.java#L279), is very short and is severely tested. This mirrors behavior in e.g. typescript.

### PR checklist
 
- [X] Run the following to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) and update samples = No change
- [X] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): master

@frol @farcaller @richardwhiuk @paladinzh

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
@@ -263,7 +263,7 @@ public class RustClientCodegen extends DefaultCodegen implements CodegenConfig {
         // If none of them is provided then fallback to default version
         if (additionalProperties.containsKey(CodegenConstants.PACKAGE_VERSION)) {
             setPackageVersion((String) additionalProperties.get(CodegenConstants.PACKAGE_VERSION));
-        } else if (openAPI.getInfo() != null && openAPI.getInfo().getVersion() != null) {
+        } else if (openAPI != null && openAPI.getInfo() != null && openAPI.getInfo().getVersion() != null) {
         	setPackageVersion(openAPI.getInfo().getVersion());
         }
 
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
```diff
@@ -259,10 +259,12 @@ public void processOpts() {
             setPackageName("openapi");
         }
 
+        // If no version is provided in additional properties, version from API specification is used.
+        // If none of them is provided then fallback to default version
         if (additionalProperties.containsKey(CodegenConstants.PACKAGE_VERSION)) {
             setPackageVersion((String) additionalProperties.get(CodegenConstants.PACKAGE_VERSION));
-        } else {
-            setPackageVersion("1.0.0");
```
wing328, line 265: can we still keep the else clause as the default package version setting?

Foorack: I mean, *we can*, but it is unnecessary. The default version is already set [above](https://github.com/OpenAPITools/openapi-generator/blob/16c9d4047d953fd09694c1affe517871979f3c39/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java#L52) in the same file.

Although I personally don't mind and can revert if you want.

Foorack:
Both Drone and CircleCI are passing, while Bitrise are "timing out after 2700 seconds" building the Swift5 client. It seems like there is a partial service degradation on Bitrise's side.

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/RustClientCodegen.java
wing328: ok. let's go with your approach.

PR #10303, opened 2021-09-01
[kotlin-client] OkHttp call is now non-blocking

sheepdreamofandroids:
All functions in the kotlin-client are marked suspend but the final call to OkHttp was blocking the thread. This is no longer the case.

Note that this can have impact on integration tests that verify backend calls because now it can happen that one call fails and therefore cancels the whole hierarchy of coroutines which possible cancels other backend calls before they even started. That IS desired behaviour though.

This code is a subset of a few patches that are running in production.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@jimschubert, @dr4ke616, @karismann, @Zomzog, @andrewemery, @4brunu, @yutaka0m


modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
```diff
@@ -326,7 +333,16 @@ import com.squareup.moshi.adapter
             headers.forEach { header -> addHeader(header.key, header.value) }
         }.build()
 
-        val response = client.newCall(request).execute()
+        val response: Response = suspendCoroutine { continuation ->
```
4brunu, line 336: @sheepdreamofandroids I think here we should use a `suspendCancellableCoroutine` instead of `suspendCoroutine`.

https://github.com/coil-kt/coil/blob/0af5fe016971ba54518a24c709feea3a1fc075eb/coil-base/src/main/java/coil/util/Extensions.kt#L45-L51

4brunu (commented):
Hey, thanks for this contribution, this is indeed a good idea.
I left some comments with improvements.
Also all those changes should only be applied when the user uses kotlin coroutines, so they should be inside the following if. 
```
{{#useCoroutines}}
changes here for the projects that use kotlin coroutines
{{/useCoroutines}}
{{^useCoroutines}}
changes here for the projects that don't use kotlin coroutines
{{/useCoroutines}}
{{/doNotUseRxAndCoroutines}}
```

wing328:

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
```diff
@@ -326,7 +333,16 @@ import com.squareup.moshi.adapter
...
+            client.newCall(request).enqueue(object : Callback {
+                override fun onFailure(call: Call, e: IOException) {
+                    continuation.resumeWithException(e)
```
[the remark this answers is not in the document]
sheepdreamofandroids: If the call is cancelled because the coroutine is cancelled as per your next comment, the resumeWithException shouldn't do anything.
On the other hand, if the call is ever cancelled for another reason, you do want to report the exception and resume, or else the coroutine will never terminate.
See https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
sheepdreamofandroids: I did this except for the check on call.isCancelled.

I added checks around the coroutine imports. The other imports may not be necessary all the time but they come from the same very common dependencies as other unconditional imports. I prefer to keep the template simple.

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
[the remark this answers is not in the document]
4brunu: Thanks for the explanation, looks good to me 👍

modules/openapi-generator/src/main/resources/kotlin-client/libraries/jvm-okhttp/infrastructure/ApiClient.kt.mustache
sheepdreamofandroids: The additional value is that an ongoing http call will be cancelled in case the result is not needed anymore. In itself that is valuable.

It seems to me that the time window where you can cancel an ongoing http call is very short, unless there is a huge request body so it will rarely make a difference.

The additional complexity is quite low though so my gut feeling says it's worth it.

4brunu: Thanks 👍

4brunu (approved):
For me this PR is ready to merge 👍

PR #10325, opened 2021-09-04
Give enums valid names when specified with x-enum-varnames

### LINKED ISSUE #3246: [BUG] x-enum-varnames is underspecified
##### Description

Support for the `x-enum-varnames` extension was added in #917 to support specifying variable names for enumeration values (e.g. for numeric values, as in https://github.com/OpenAPITools/openapi-generator/issues/893#issuecomment-416617460).  The documentation from #2010 suggests that the names will undergo language-specific normalization (as enum values would), but they do not, which can result in broken code.

The issue has been present since #917 was merged.  Tested on `master`.

```yaml
openapi: '3.0.2'
info:
  title: x-enum-varnames example
  version: '1.0.0'
components:
  schemas:
    WeatherType:
      type: integer
      enum:
      - 0
      - 1
      - 2
      x-enum-varnames:
      - Sunny
      - Partly Cloudy
      - Rainy
paths:
  /weather:
    get:
      responses:
        default:
          description: Current weather
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/WeatherType'
```

`openapi-generator-cli.jar generate -i openapi.yaml -g java -o bad-x-enum-varnames-java`

##### Steps to reproduce

1.  Run the above command.
2.  Try to compile the generated code, which will fail due to `Partly Cloudy` being an invalid variable name.

##### Suggest a fix

The appropriate fix depends on the desired semantics of `x-enum-varnames`:

1.  If `x-enum-varnames` should be a natural language name, then the values should be converted to identifiers in the same way as enum values (as suggested in the documentation), which would fix the above example.
2.  If `x-enum-varnames` must be a valid identifier (in every language which can be generated) then the documentation should be updated to clarify that.  (Ideally the code generation would also fail more gracefully.)

A problem with the first option is that it prevents an existing use case of overriding the generator convention, as described in https://github.com/swagger-api/swagger-codegen/issues/7466#issuecomment-483602884, and could change the generated API by changing the generated names (depending on how identifier normalization is done).

A problem with the second option is that different languages have different casing conventions for enumeration values and there's no way to satisfy them all.  Should the spec use `PARTLY_CLOUDY`, `PartlyCloudy`, or `partlyCloudy` for `x-enum-varnames`?  Any choice will be unconventional in several languages.  Also, different languages have different rules for valid identifiers and requiring spec authors to comply with all of them is burdensome.

Thanks for considering,
Kevin

UnsolvedCypher:
This changes how x-enum-varnames behaves. Issue #3246 already describes the problem quite well, but to summarize, the current codegen just takes the value specified by x-enum-varnames and sets the enum variable name to it. This causes the generation of bad code when the specified name isn't a valid variable name, such as because it contains spaces. Additionally, different languages have different conventions for enum names, so for you may want `Partly cloudy` to generate `PARTLY_CLOUDY` in Java but `PartyCloudy` in TypeScript.

This change now passes the user-specified name to the language-specific `toEnumVarName` function for processing, so that the generated variable name is legal for the target language, and so that it conforms to language conventions or any other generator-specific settings that have been set. This is what issue #3246 refers to as "option 1" for a fix.

If there is any other information I can provide, please let me know 😃

Fix #3246

### PR checklist
 
- [x ] Pull Request title clearly describes the work in the pull request and Pull Request description provides details about how to validate the work. Missing information here may result in delayed response from the community.
- [x ] Run the following to [build the project](https://github.com/OpenAPITools/openapi-generator#14---build-projects) and update samples:
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [ x] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master` (5.3.0), `6.0.x`
- [x ] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request.

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
@@ -5726,14 +5726,18 @@ public class DefaultCodegen implements CodegenConfig {
 
     protected void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions, String dataType) {
         if (vendorExtensions != null) {
-            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-varnames", "name");
-            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-descriptions", "enumDescription");
+            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-varnames", "name", dataType);
+            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-descriptions", "enumDescription", dataType);
         }
     }
 
-    private void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions, String extensionKey, String key) {
+    private void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions, String extensionKey, String key, String dataType) {
         if (vendorExtensions.containsKey(extensionKey)) {
             List<String> values = (List<String>) vendorExtensions.get(extensionKey);
+            if (extensionKey.equals("x-enum-varnames"))
+            {
+                values = values.stream().map(v -> toEnumVarName(v, dataType)).collect(Collectors.toList());
+            }
             int size = Math.min(enumVars.size(), values.size());
             for (int i = 0; i < size; i++) {
                 enumVars.get(i).put(key, values.get(i));
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -874,7 +874,7 @@ WeatherType:
     - Rainy
 ```
 
-In the example for the integer value `42`, the description will be `Blue sky` and the name of the enum item will be `Sunny` (some generators changes it to `SUNNY` to respect some coding convention).
+In the example for the integer value `42`, the description will be `Blue sky` and the name of the enum item will be `Sunny`, `SUNNY`, or `sunny` depending on the enum conventions of the target language. The name will be modified to be a valid enum name if necessary, for example spaces might become underscores.
 
 ### ObjC
 #### x-objc-operationId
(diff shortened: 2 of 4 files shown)


joey-yoonsung (approved):
Thanks 👍 That's what I want!

modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java
```diff
@@ -5726,14 +5726,18 @@ protected String getEnumDefaultValue(String defaultValue, String dataType) {
...
+            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-varnames", "name", dataType);
+            updateEnumVarsWithExtensions(enumVars, vendorExtensions, "x-enum-descriptions", "enumDescription", dataType);
         }
     }
 
-    private void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions, String extensionKey, String key) {
+    private void updateEnumVarsWithExtensions(List<Map<String, Object>> enumVars, Map<String, Object> vendorExtensions, String extensionKey, String key, String dataType) {
         if (vendorExtensions.containsKey(extensionKey)) {
             List<String> values = (List<String>) vendorExtensions.get(extensionKey);
+            if (extensionKey.equals("x-enum-varnames"))
```
agilob, line 5737: Can extensionKey be null?

UnsolvedCypher: It can't be null- the only uses of this function are a few lines up where `extensionKey` is specified to be either `x-enum-varnames` or `x-enum-descriptions`.

UnsolvedCypher:
Hi @wing328 could you please take a look at this when you have a chance? It would be great to get it merged!

UnsolvedCypher:
Hi @wing328, I believe this is good to merge. Please let me know if you would like any changes

PR #10326, opened 2021-09-04
Windows adaptations and installer configuration

ahmedyarub:
@zhemant  @ityuhui  @michelealbano 

Cast to correct type when applicable to be compilable on C++ compilers
Add conditional installation of configuration files

1- Adaptations for Windows compilation
2- Installer with CMake configuration

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [X] File the PR against the [correct branch](https://github.com/OpenAPITools/openapi-generator/wiki/Git-Branches): `master` (5.3.0), `6.0.x`


modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
```diff
@@ -92,8 +92,54 @@ target_link_libraries(${pkgName} ${CURL_LIBRARIES} )
 
 include(PostTarget.cmake OPTIONAL)
 
-#install library to destination
-install(TARGETS ${pkgName} DESTINATION ${CMAKE_INSTALL_PREFIX})
+if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in)
```
ityuhui, line 95: How the `${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in` is generated ?

In my env, this file doesn't exist, at this case, how to execute the original `install()` commands after your PR ?

modules/openapi-generator/src/main/resources/C-libcurl/api-header.mustache
```diff
@@ -8,6 +8,10 @@
 {{#imports}}{{{import}}}
 {{/imports}}
 
+#undef stdin
+#undef stderr
+#undef stdout
```
ityuhui, line 13: Are these lines to fix compilation errors on Windows ？ Must it be in this files? （There will be many generated files affected.）

ahmedyarub: I'm not sure whether this is related to the compiler or some standard libraries. An ideal solution would change the name of reserved identifier from something like stdin to _stdin to avoid clashes. Do you have any alternative solution?

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
ahmedyarub: I think that I need one more day to take care of these points. Thank you for the patience.

ahmedyarub: It should be created manually just like this https://github.com/kubernetes-client/c/blob/4221f70114de95773d3ef53e8cafe357f454b705/kubernetes/Config.cmake.in
But you are right, we should have a fallback when the file does not exist.

modules/openapi-generator/src/main/resources/C-libcurl/api-header.mustache
ityuhui: What's the problem if these `undef` don't exist ? The build on my linux(Ubuntu) works well.

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
ityuhui: Is the file `Config.cmake.in` is used for `find_package()` by users ?
https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html
Then the name should be `<PackageName>Config.cmake`, so I think the name `${pkgName}Config.cmake.in` is better than `Config.cmake.in` here.

modules/openapi-generator/src/main/resources/C-libcurl/api-header.mustache
ahmedyarub: for example if I remove it from kubernetes/model/v1_container.h I get the following errors around "int stdin"

```
E:\cpp\k8s_c\kubernetes\model\v1_container.h(49): error C2143: syntax error: missing ')' before 'constant'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(49): error C2032: '__acrt_iob_func': function cannot be member of struct 'v1_container_t'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(49): error C2059: syntax error: ')'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(58): error C2059: syntax error: '}'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(75): error C2143: syntax error: missing ')' before 'constant'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(75): error C2059: syntax error: ')'
E:\cpp\k8s_c\kubernetes\model\v1_container.h(83): error C2059: syntax error: ')'
```

ahmedyarub: It looks like the error happens only in model files so I removed the directives from api files

ityuhui: OK. I got it.  `stdin` maybe a keyword of Microsoft VC++ and libraries, but it happens in OpenAPI spec of kubernetes API server.
```yaml
        "stdin": {
          "description": "Whether this container should allocate a buffer for stdin in the container runtime. If this is not set, reads from stdin in the container will always result in EOF. Default is false.",
          "type": "boolean"
        },
```

I think the right solution is adding `stdin/stdout/stderr` to reservedWords ( in the function `setReservedWordsLowerCase`). They will be escaped automatically.
Reference: https://github.com/OpenAPITools/openapi-generator/pull/8205

ahmedyarub: Removed the #undef directives and made the suggested changes

modules/openapi-generator/src/main/resources/C-libcurl/model-header.mustache
```diff
@@ -19,6 +19,7 @@ typedef struct {{classname}}_t {{classname}}_t;
 #include "{{{.}}}.h"
 {{/imports}}
 
+
```
ityuhui, line 22: Remove this line please.

ahmedyarub: Done

modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/CLibcurlClientCodegen.java
```diff
@@ -245,7 +245,10 @@ public CLibcurlClientCodegen() {
                         "final",
                         "override",
                         "transaction_safe",
-                        "transaction_safe_dynamic")
+                        "transaction_safe_dynamic",
+                        "stdin",
```
ityuhui, line 249: Suggest adding an annotation before "stdin" like
```
// cpp reserved keywords
```

e.g.
```
// VC++ reserved keywords
```

ahmedyarub: Done

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
```diff
@@ -86,14 +87,66 @@ include(PreTarget.cmake OPTIONAL)
...
+        get_filename_component(HDR_DIRECTORY ${HDR_FILE} DIRECTORY)
+        get_filename_component(ABSOLUTE_HDR_DIRECTORY ${HDR_DIRECTORY} ABSOLUTE)
+        file(RELATIVE_PATH RELATIVE_HDR_PATH ${CMAKE_CURRENT_SOURCE_DIR} ${ABSOLUTE_HDR_DIRECTORY})
+        install(FILES ${HDR_FILE} DESTINATION include/${RELATIVE_HDR_PATH})
+    endforeach()
+
+    include(CMakePackageConfigHelpers)
+    write_basic_package_version_file(
+            "${CMAKE_CURRENT_BINARY_DIR}/${pkgName}/${pkgName}ConfigVersion.cmake"
+            VERSION 0.0.1
```
ityuhui, line 119: Is this version (0.0.1) for this project ?
How to decide/update this version (For now it is hard-coded) ?

ahmedyarub: Yes it is. I think that I'll set its value in some variable and make it configurable in PreTarget script. What do you think?

samples/client/petstore/java/apache-httpclient/docs/FakeApi.md
```diff
@@ -586,7 +586,7 @@ public class Example {
         Float _float = 3.4F; // Float | None
         String string = "string_example"; // String | None
         File binary = new File("/path/to/file"); // File | None
-        LocalDate date = LocalDate.now(); // LocalDate | None
+        LocalDate date = new LocalDate(); // LocalDate | None
```
ityuhui, line 589: I think we don't need commit code change for the sample of other languages.

ahmedyarub: I just regenerated the docs and this was the result.

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
ityuhui: Yes. it's better to make this value configurable. For the sample (without a PreTarget), it can be 0.0.1 by default.

Kubernetes C client is planning version now. Thank you for your contribution !

docs/generators/scala-play-server.md
```diff
@@ -83,7 +83,6 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 <li>case</li>
 <li>catch</li>
 <li>class</li>
```
ityuhui: I think we dont't need change doc for other languages here.

samples/client/petstore/java/apache-httpclient/docs/FakeApi.md
ahmedyarub: Change reverted and updated the required file only

docs/generators/scala-play-server.md
ahmedyarub: Change reverted and updated the required file only

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
ahmedyarub: Done

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
```diff
@@ -80,20 +81,73 @@ set(HDRS
 
 )
 
+set(VERSION 0.0.1)
```
ityuhui, line 84: Could you please move `set(VERSION 0.0.1)` near after `set(pkgName "{{projectName}}")` (line 30) ? 
I think it's clear for user.

ityuhui: And add an annotation to tell user the "VERSION" can be changed by some other file.

ahmedyarub: Done

modules/openapi-generator/src/main/resources/C-libcurl/CMakeLists.txt.mustache
```diff
@@ -27,6 +28,7 @@ else()
 endif()
 
 set(pkgName "{{projectName}}")
+set(VERSION 0.0.1)
```
ityuhui, line 31: And add an annotation to tell user the "VERSION" can be changed by some other file.

ahmedyarub: Done

ityuhui (approved):
Nice code !
It is good to me.

PR #10338, opened 2021-09-07
Revert "Improved rendering of python function params and docstrings in IDEs"

spacether:
Reverts OpenAPITools/openapi-generator#10331


spacether:
Reverting because tests broke

PR #10344, opened 2021-09-07
[cpp-ue4] [bugfix] for "generated code doesn't properly check for connection failures"

### LINKED ISSUE #10315: [BUG] [cpp-ue4]
#### Bug Report Checklist

- [x] Have you [tested with the latest master](https://github.com/OpenAPITools/openapi-generator/wiki/FAQ#how-to-test-with-the-latest-master-of-openapi-generator) to confirm the issue still exists?

##### Description

The generated code for handling REST requests and responses doesn't properly check for connection failures.  Also, when it defaults into a connection failure scenario later on, it erroneously injects a HTTP response code of 408.  Unreal's own files say that response means:
```c++
		// the server timed out waiting for the request.
		RequestTimeout = 408,
```

this can be seen at the end of the "HandleResponse" method in the Api object.  Quoting the template:

```c++
	// By default, assume we failed to establish connection
	InOutResponse.SetHttpResponseCode(EHttpResponseCodes::RequestTimeout);
```

This code is reached when the HTTPRespone is null, or bSuccess is false.  In the case of a connection failure, the server never responded but we got here anyway; and now, the REST protocol layer is pretending that the server sent a response and a code that it never sent.  This behavior is very problematic.

For my part, I was trying to catch the 408 response by debugging my server code because of course, an http response code that is reserved for an actual server to send, is being sent by the server right?  Nope.  Eventually I debugged Django deeply enough to suspect that the connection was being rejected and thought to check for client side code making up the 408 error out of thin air.  Perhaps it was Apple's WebKit client?  Nope, it was the generated REST code from the template.

What likely should be happening is: before constructing a response, the client should probably check the request for its status.

from the existing template:
```c++
void {{classname}}::On{{operationIdCamelCase}}Response(FHttpRequestPtr HttpRequest, FHttpResponsePtr HttpResponse, bool bSucceeded, F{{operationIdCamelCase}}Delegate Delegate) const
{
	{{operationIdCamelCase}}Response Response;
	HandleResponse(HttpResponse, bSucceeded, Response);
	Delegate.ExecuteIfBound(Response);
}
```

The HttpRequest has a GetStatus() method which could be leveraged as such before actually trying to "HandleResponse" when a response actually never happened:
```c++
void OrgbuilderApi::OnUpdateLegalEntityResponse(FHttpRequestPtr HttpRequest, FHttpResponsePtr HttpResponse, bool bSucceeded, FUpdateLegalEntityDelegate Delegate) const
{
	//handling a response that isn't an actual response would be a problem.
	switch (HttpRequest->GetStatus())
	{
	case EHttpRequestStatus::Failed :
		{
			UpdateLegalEntityResponse Response;
			Response.SetHttpResponseCode(EHttpResponseCodes::Unknown);
			Response.SetResponseString("Client reports request connection failed: " + HttpRequest->GetURL());
			Response.SetSuccessful(false);
			Delegate.ExecuteIfBound(Response);
		}; break ;
	case EHttpRequestStatus::Failed_ConnectionError :
		{
			UpdateLegalEntityResponse Response;
			Response.SetHttpResponseCode(EHttpResponseCodes::Unknown);
			Response.SetResponseString("Client reports request connection failed but could be retried: " + HttpRequest->GetURL());
			Response.SetSuccessful(false);
			Delegate.ExecuteIfBound(Response);
		}; break ;
	default :
		{
			//we likely have an actual response to handle.
			UpdateLegalEntityResponse Response;
			HandleResponse(HttpResponse, bSucceeded, Response);
			Delegate.ExecuteIfBound(Response);
		};
	}
}
```

I'm not thrilled with this solution.  Because we're still technically telling the downstream developer that a response happened.  Though I'm much happier at least setting the response to Unknown (0) and passing along a response string that describes this as a client talking about a client connection error.

```c++
// status code not set yet
		Unknown = 0,
```

lastly, we'd likely change the fall through in HandleResponse to something like this, to keep the general structure in place:

```c++
	// By default, assume we failed to establish connection
	InOutResponse.SetHttpResponseCode(EHttpResponseCodes::Unknown);
	if (bSucceeded)
	{
		InOutResponse.SetResponseString(TEXT("The client didn't report a connection failure, but it received no response."));
	} else
	{
		InOutResponse.SetResponseString(TEXT("The client didn't report a connection failure, but the request was not succesful."));
	}
```

But really we probably need a more descriptive interface here where a downstream developer can be informed of client connection failures, as a separate matter from handling server responses.

development HEAD

any valid file will generate this code.

##### Generation Details

any cpp-ue4 generated api will generate this code.

##### Steps to reproduce

- generate any api with cpp-ue4 to see the code.
- If trying to create this client-server condition, one might try putting up a firewall, which should reject client connection requests.

##### Related issues/PRs

none

##### Suggest a fix

see above.

tagging @Kahncode

leith-bartrich:
fixes #10315 

As described in the original bug report, there were issues with response handling.  Specifically surrounding connection failures and what was eventually termed an "opinionated part of the handling."

This pull request restructures the building of the response and disambiguates parts of it.  It also documents heavily the purpose of the published and processed items in the response.  From the commits themselves:

- fixed likely mistake in assigning RetryManager field.
- reworked response handling to better incorporate client connection errors and more granular information.
- differentiated various strings and returns in the response between http and rest.  e.g. RestRequestSuccesful and HttpRequestSuccesful are two different things.
- removed an instance of the client setting an arbitrary server response.
- renamed HandleRespons to TryBuildResponse.  The response may not have actually happened.
- Added HttpRequest to the Response.
- Renamed internal operationResponse to operationProcessRequestComplete.  Mirror's Unreal's naming for what it's handling.
- converted some white-space spaces to tabs.
- changed rest response content and description to TOptional.
- reordering and documenting Response data.

This is a breaking changeset for downstream developers.  The disambiguation of parts of the response will very likely break application code that uses it.  And unfortunately, I suspect that's necessary.

Also, I do not have the resources to test it extensively.  I have tested it on my local code-base with plenty of artificially induced client and server failure situations, to check robustness.  But I don't have a breadth of APIs and servers to test against.  I think a change as extensive as this, requires more eyes and testing than I can provide.

Looking at the diff for TryBuildResponse is probably not as useful as looking at the full version of it. The changes are substantial.  Similarly I'd suggest looking at the whole response object with the comments in place.

The overall approach to the TryBuildResponse method is now one of publishing and processing, in a less than "opinionated" manner.  In hopes that it publishes and processes everything that one would need, without requiring diving deeper into the Unreal HTTP layer.  But ultimately it provides the access to that layer if needed.

For reference, I'm providing some example code here of how I use this data in my own project.

First, a template method for general REST request and response handling structure (using my BP wrapper classes, but close enough to the underlying objects to be understood):
```c++
	template<typename U, typename V, typename  W>
	UFUNCTION(BlueprintCallable, Category = "URESTView")
	void DoHandleRESTRequest(U * request, W * api, TFunction<void (V*)> on_success, TFunction<void(V*)> on_connection_fail, TFunction<void(V*)> on_unknown_fail )
	{
		request->OnResponseEvent.AddLambda([this,on_success, on_connection_fail, on_unknown_fail](U * request, V * response)
		{
			if (!response->IsHttpRequestSuccessful())
			{
				if (response->IsHttpClientConnectionError())
				{
					this->LogError("The client couldn't connect.");
					on_connection_fail(response);
				} else
				{
					this->LogError("HTTP request failed.");
					on_unknown_fail(response);
				}
				return;
			}
			if (!response->IsRestRequestSuccessful())
			{
				this->LogError("The REST request failed.  HTTP Response Code: " + FString::FromInt(response->GetHttpResponseCode()) );
				if (response->RestResponseDescriptionIsSet())
				{
					this->LogInformation("REST Response Description: " + response->GetRestResponseDescription());
				}
				if (response->RestResponseContentIsSet())
				{
					this->LogInformation("REST Response Content: " + response->GetRestResponseContent());
				}
				on_unknown_fail(response);
				return;
			}
			if (!response->IsRestContentFullyParsed())
			{
				this->LogWarning("The REST response didn't parse fully as per the Schema.");
			}
			on_success(response);
			return;
		});
		request->ExecuteRequest(api);
	}
```

and further, a use of this templated method in a ClientView object, from my codebase:
```c++
void ULegalEntitiesManagerView::RequestUpdateLegalEntity(UFPLLegalEntity* entry)
{
	auto request = NewObject<UFPLOrgbuilderApiUpdateLegalEntityRequest>();
	request->SetId(entry->GetId().ToString());
	request->SetLegalEntity(entry);
	this->LogInformation("Requesting commit of LegalEntity: " + entry->GetId().ToString());
	DoHandleRESTRequest<UFPLOrgbuilderApiUpdateLegalEntityRequest,UFPLOrgbuilderApiUpdateLegalEntityResponse,UFPLOrgbuilderApi>(request,this->api,
		[this](auto response)
		{
			this->LogInformation("Successful commit of LegalEntity.");
			auto returned = response->GetContent();
			this->OnReplaceLegalEntity.Broadcast(returned);
			this->OnShowManager.Broadcast();
		},
		[this](auto response)
		{
			this->LogInformation("The server never received the commit. It's safe to retry.");
			this->OnShowManager.Broadcast();
		},
		[this](auto response)
		{
			this->LogInformation("The server may have received and processed the update.  But it's safe to retry the commit.");
			this->OnShowManager.Broadcast();
		}
	);
	this->DoCancelableRequest("Updating LegalEntitity: " + entry->GetId().ToString(),[request](){request->CancelRequest();});
	
}
```

Of note in these client usage examples is that most information is being published back to the user.  But a client application layer could also make use the same information to execute more complex behavior.

One thing this adjustment has made fairly clear is that this part of the generated code will likely required substantial rework at a later time.

A schema can define different "content" or return types for an endpoint depending on the server response code.  And as it sits, this (and the prior) implementation will make the mistake of trying to parse the first specified response in the schema, every time, no matter the response code.  Which works well enough to get things working for simple Schemas.

Further, it also won't handle common rest styled errors either.  There are two points in the comments were I've noted that we could expand to check for and parse some well known error types.  But those kinds of expansion are beyond the scope of this fix.

It's debatable as to if the TryBuildResponse method should try to parse Content from a server response that doesn't match the response code in the schema.  I have left the behavior in place where it'll try anyway and gracefully fail to do so.  Which doesn't match a pedantic reading of the Schema, but is instead very lenient towards the server and Schema.

Finally, I apologize for some ugliness to the diff.  The codebase is clearly a tab indented set, and I took the opportunity to include some fixes where some lines in the files I was working in were erroneously space intended.  So, if you can't figure out why some lines seem to be changing as they don't relate to the changes above and don't seem substantive, consider that it may be a whitespace fix.  If this is a serious issue, I can go back and squash out those fixes.

For now, this is a draft pull request.  Largely because I think it needs more testing before it could be merged.  Also, because I have not addressed the issue of how to (or if to) signal these kinds of breaking changes.

tagging @Kahncode 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-header.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-header.mustache
@@ -32,17 +32,17 @@ public:
 	{{#operations}}{{#operation}}class {{operationIdCamelCase}}Request;
 	class {{operationIdCamelCase}}Response;
 	{{/operation}}{{/operations}}
-    {{#operations}}{{#operation}}DECLARE_DELEGATE_OneParam(F{{operationIdCamelCase}}Delegate, const {{operationIdCamelCase}}Response&);
-    {{/operation}}{{/operations}}
-    {{#operations}}{{#operation}}{{#description}}/* {{{.}}} */
+	{{#operations}}{{#operation}}DECLARE_DELEGATE_FourParams(F{{operationIdCamelCase}}Delegate, const {{operationIdCamelCase}}Response&, FHttpRequestPtr, FHttpResponsePtr, bool bHttpSuccdeded);
+	{{/operation}}{{/operations}}
+	{{#operations}}{{#operation}}{{#description}}/* {{{.}}} */
 	{{/description}}FHttpRequestPtr {{operationIdCamelCase}}(const {{operationIdCamelCase}}Request& Request, const F{{operationIdCamelCase}}Delegate& Delegate = F{{operationIdCamelCase}}Delegate()) const;
-    {{/operation}}{{/operations}}
+	{{/operation}}{{/operations}}
 private:
-    {{#operations}}{{#operation}}void On{{operationIdCamelCase}}Response(FHttpRequestPtr HttpRequest, FHttpResponsePtr HttpResponse, bool bSucceeded, F{{operationIdCamelCase}}Delegate Delegate) const;
-    {{/operation}}{{/operations}}
+	{{#operations}}{{#operation}}void On{{operationIdCamelCase}}ProcessRequestComplete(FHttpRequestPtr HttpRequest, FHttpResponsePtr HttpResponse, bool bSucceeded, F{{operationIdCamelCase}}Delegate Delegate) const;
+	{{/operation}}{{/operations}}
 	FHttpRequestRef CreateHttpRequest(const Request& Request) const;
 	bool IsValid() const;
-	void HandleResponse(FHttpResponsePtr HttpResponse, bool bSucceeded, Response& InOutResponse) const;
+	void TryBuildResponse(FHttpRequestPtr HttpRequest, FHttpResponsePtr HttpResponse, bool bHttpSucceeded, Response& InOutResponse) const;
 
 	FString Url;
 	TMap<FString,FString> AdditionalHeaderParams;
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-header.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-header.mustache
@@ -32,7 +32,7 @@ public:
 	{{#operations}}{{#operation}}class {{operationIdCamelCase}}Request;
 	class {{operationIdCamelCase}}Response;
 	{{/operation}}{{/operations}}
-	{{#operations}}{{#operation}}DECLARE_DELEGATE_FourParams(F{{operationIdCamelCase}}Delegate, const {{operationIdCamelCase}}Response&, FHttpRequestPtr, FHttpResponsePtr, bool bHttpSuccdeded);
+	{{#operations}}{{#operation}}DECLARE_DELEGATE_OneParam(F{{operationIdCamelCase}}Delegate, const {{operationIdCamelCase}}Response&);
 	{{/operation}}{{/operations}}
 	{{#operations}}{{#operation}}{{#description}}/* {{{.}}} */
 	{{/description}}FHttpRequestPtr {{operationIdCamelCase}}(const {{operationIdCamelCase}}Request& Request, const F{{operationIdCamelCase}}Delegate& Delegate = F{{operationIdCamelCase}}Delegate()) const;
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-operations-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-operations-source.mustache
@@ -294,7 +294,7 @@ void {{classname}}::{{operationIdCamelCase}}Response::SetHttpResponseCode(EHttpR
 	{{#isDefault}}
 	default:
 	{{/isDefault}}
-		SetResponseString(TEXT("{{message}}"));
+		SetRestResponseDescription(TEXT("{{message}}"));
 		break;
 	{{/responses}}
 	}
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
@@ -161,7 +161,7 @@ void {{classname}}::On{{operationIdCamelCase}}ProcessRequestComplete(FHttpReques
 {
 	{{operationIdCamelCase}}Response Response;
 	TryBuildResponse(HttpRequest, HttpResponse, bHttpSucceeded, Response);
-	Delegate.ExecuteIfBound(Response, HttpRequest, HttpResponse, bHttpSucceeded);
+	Delegate.ExecuteIfBound(Response);
 }
 
 {{/operation}}
--- a/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
+++ b/modules/openapi-generator/src/main/resources/cpp-ue4/api-source.mustache
@@ -47,7 +47,7 @@ bool {{classname}}::IsValid() const
 
 void {{classname}}::SetHttpRetryManager(FHttpRetrySystem::FManager& InRetryManager)
 {
-	if(RetryManager != &GetHttpRetryManager())
+	if(&InRetryManager != RetryManager)
 	{
 		DefaultRetryManager.Reset();
 		RetryManager = &InRetryManager;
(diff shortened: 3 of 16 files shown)


Kahncode:
Hello,

This is definitely a lot to process but it goes in the right direction. I do believe we should make the response objects lighter by providing accessors to the underlying data rather than copying it directly as booleans in the response, i.e. the request success, raw response and errors can be directly accessed without being copied. We should also aim to not break the existing public API of Response, which is not necessary. I also think the general structure and naming of the response handling methods needs not change as much.

To your later points about the generator not handling different kinds of response, this is absolutely the case and would require a much larger rework.

One extra comment to this post is that your templated method is BlueprintCallable, as far as I know this was rejected by UHT?

leith-bartrich:
The UHT one is easy to answer as an aside.  I noticed while I was publishing it here that its fairly nonsensical but since that's how its currently working I left it in, warts and all.  I suspect UHT is ignoring it because it's a template.  And I suspect the actual template preprocessing isn't actually writing that part out in the code to compile, as it's not technically part of the method.  I'll likely remove it (that macro usage) from my code-base shortly for that reason.  And yea, had UHT complained about it, I'd have noticed it is a silly thing to have put there.

leith-bartrich:
> I do believe we should make the response objects lighter by providing accessors to the underlying data rather than copying it directly as booleans in the response, i.e. the request success, raw response and errors can be directly accessed without being copied.

Will take a look at this.  It'll take a couple of days.  I'm away tomorrow but I'll be getting to it.

> We should also aim to not break the existing public API of Response, which is not necessary.

I'll take a look at this.  Just off the top of my head:  I believe the original issue that made me disambiguate was that a single field "ResponseString" was being used multiple times in ambiguous ways.  First it was set to the actual http response content. Then later it was being set to the Schema's description of the response code in some cases.  And in practice, that wipes out the (probably) useful return body, replaced by a static string defined by the Schema file.  I split them up to at least hold them separately (as they're both useful) and needed to give them both more meaningful names.  The problem as I see it, is that downstream developers are probably assuming that "ResponseString" is one of two things.  But we don't know which they're assuming and actually, it's neither of those things anyway.  Ideally we'd just mark it deprecated and have it get the http response body as a compromise.  Is that something we can do in unreal code that's not a UFUNCTION?  I've never tried.  https://en.cppreference.com/w/cpp/language/attributes/deprecated. I'll try it at least.

> I also think the general structure and naming of the response handling methods needs not change as much.

I'll take a look. iirc:  It's not critical since it's all internal.  So I felt the threshold to changing them was lower.  I can change them back if need be.

leith-bartrich:
going a different direction.

Kahncode:
I thought this was going pretty much in the right direction, I'm wondering what would be your other approach. Cheers.

PR #10361, opened 2021-09-09
[Typescript] Improve exception for unknown responses

bodograumann:
For http errors that are not explicitly defined in the api specification, currently the consolidated typescript generator does not retain the body. Instead a generic `[object Object]` string is provided for it in the error message. This is probably an oversight.

Instead we now set the message string and response body attributes on the exception object separately.
Additionally I have added a heuristic which tries to parse the response as JSON, falls back first to plain text, then to binary and then to undefined.

I have added a new `deleteNonExistantPet` integration test to the default typescript/node petstore sample.

This is a template-only change.

Steps:
- Clean up tests
- Change tests to async-await
- Use different pets for each test
- Add test for unknown error response code
- Make body of unexpected error responses available
- Regenerate all consolidated typescript samples

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

CC @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01) @topce (2018/10) @akehir (2019/07) @petejohansonxo (2019/11) @amakhrov (2020/02)

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript/api/exception.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript/api/exception.mustache
@@ -8,7 +8,7 @@
  *
  */
 export class ApiException<T> extends Error {
-    public constructor(public code: number, public body: T) {
-        super("HTTP-Code: " + code + "\nMessage: " + JSON.stringify(body))
+    public constructor(public code: number, message: string, public body: T) {
+        super("HTTP-Code: " + code + "\nMessage: " + message + "\nBody: " + JSON.stringify(body))
     }
 }
--- a/modules/openapi-generator/src/main/resources/typescript/http/http.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript/http/http.mustache
@@ -290,22 +290,18 @@ export class ResponseContext {
 
     /**
      * Use a heuristic to get a body of unknown data structure.
-     * Try JSON with fall back to plain text and binary.
+     * Return as string if possible, otherwise as binary.
      */
-    public async getBodyAsAny(): Promise<any> {
+    public getBodyAsAny(): Promise<string | {{{fileContentDataType}}} | undefined> {
         try {
-            const body = await this.body.text();
-            try {
-                return JSON.parse(body);
-            } catch {}
-            return body;
+            return this.body.text();
         } catch {}
 
         try {
             return this.body.binary();
         } catch {}
 
-        return undefined;
+        return Promise.resolve(undefined);
     }
 }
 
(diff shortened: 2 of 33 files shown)


modules/openapi-generator/src/main/resources/typescript/api/exception.mustache
```diff
@@ -8,7 +8,7 @@
  *
  */
 export class ApiException<T> extends Error {
-    public constructor(public code: number, public body: T) {
-        super("HTTP-Code: " + code + "\nMessage: " + JSON.stringify(body))
+    public constructor(public code: number, message: string, public body: T) {
```
agilob, line 11: that's not backwards compatible, is it?

bodograumann: Yes there are some slight breaking changes here. (Luckily the generator is still marked as experimental)

1. If you used the constructor in user code, that will have to be changed. (That seems unlikely thought)
2. The `body` property of the exception now really contains the response body. Previously it sometimes contained the message as ApiException<string> and sometimes the body.

All-in-all I feel these are justified breakages.
Maybe it could make sense to put it in v6.0. Is there a timeline on that?

amakhrov (approved):
LGTM

modules/openapi-generator/src/main/resources/typescript/http/http.mustache
```diff
@@ -287,6 +287,26 @@ export class ResponseContext {
...
+
+    /**
+     * Use a heuristic to get a body of unknown data structure.
+     * Try JSON with fall back to plain text and binary.
+     */
+    public async getBodyAsAny(): Promise<any> {
+        try {
+            const body = await this.body.text();
+            try {
+                return JSON.parse(body);
```
macjohnny, line 299: I suggest not to parse the json here, but simply pass it as a string to the exception

bodograumann: In my use cases data is predominantly sent as json. So I thought this might make things easier. If it cannot be parsed as json the heuristic would still fall back to plain string.
What is the reason you would discourage json parsing, @macjohnny ?
What about only doing it when there is a suitable Content-Type header set?

bodograumann: How do you want to go ahead with this @macjohnny .
I would love to get this moving.
If you feel strongly about removing the JSON parsing, I‘ll do the change.

macjohnny: Sorry, forgot to reply.
Since it is for debugging purposes, we should avoid heuristic parsing, and simply log the string payload

bodograumann: Ok, that makes sens. I removed the heuristic.

bodograumann:
Rebased on master.

PR #10363, opened 2021-09-09
draft1

anichari:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/resources/python/model.mustache
+++ b/modules/openapi-generator/src/main/resources/python/model.mustache
@@ -25,7 +25,9 @@ from {{packageName}}.exceptions import ApiAttributeError
 {{#imports}}
 {{#-first}}
 
-def lazy_import(zzz_class_name_to_class=dict()):
+def lazy_import(zzz_class_name_to_class=None):
+    if not zzz_class_name_to_class:
+        zzz_class_name_to_class = dict()
     {{{.}}}
 {{/-first}}
 {{^-first}}
@@ -35,7 +37,7 @@ def lazy_import(zzz_class_name_to_class=dict()):
 {{/-first}}
 {{#-last}}
     {{{.}}}
-return zzz_class_name_to_class{{/-last}}
+    return zzz_class_name_to_class{{/-last}}
 {{/imports}}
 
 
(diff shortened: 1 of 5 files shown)


modules/openapi-generator/src/main/resources/python/model_templates/classvars.mustache
```diff
@@ -92,24 +92,24 @@
         """
 {{#imports}}
 {{#-first}}
-        lazy_import()
+        class_name_to_class = lazy_import()
```
anichari, line 95: This line and its buddies dont get created if there are no imports obviously. And further down in 100, 103 etc, when I try to use the calling class' name, the dict is non-existent since 95 is not there.

modules/openapi-generator/src/main/resources/python/model.mustache
```diff
@@ -25,9 +25,17 @@ from {{packageName}}.exceptions import ApiAttributeError
 {{#imports}}
 {{#-first}}
 
-def lazy_import():
+def lazy_import(zzz_class_name_to_class=dict()):
+    {{{.}}}
```
spacether, line 29: dicts are mutable. If you modify tgiz dict it stays modified. For that reason it's a bad practice to set argument defaults to empty dict. The second time you run the function the dict will not be empty. How about moving the empty dict assignment into the function?

modules/openapi-generator/src/main/resources/python/model_templates/classvars.mustache
spacether: That is expected. Are there models that are missing imports?

modules/openapi-generator/src/main/resources/python/model_templates/classvars.mustache
```diff
@@ -130,7 +161,12 @@
 {{/mappedModels}}
         val = {
 {{#mappedModels}}
+{{#imports}}
```
spacether, line 164: This may not work. It looks like it is looping over every import.
You probably want:
{{#imports}}
{{#-first}}
{{/-first}}
{{/imports}}

wing328:
Closing it as there's no update.

PR #10373, opened 2021-09-10
[kotlin-spring] ResponseEntity option

frecco75:
This PR adds `ResponseEntity` option.
By default the generator (before this PR) generates code with 200 return code.

Given this schema
```
openapi: "3.0.1"
info:
  version: 1.0.0
  title: Users

paths:
  /users/{userId}:
    get:
      parameters:
        - in: path
          name: userId
          required: true
          schema:
            type: string
      responses:
        200:
          description: user found
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
        404:
          description: user not found
  
components:
  schemas:
    User:
      type: object
      required:
        - name
      properties:
        name:
          type: string
```

It generates by default this code
```
fun usersUserIdGet(@ApiParam(value = "", required=true) @PathVariable("userId") userId: kotlin.String
): ResponseEntity<User> {
        return ResponseEntity(service.usersUserIdGet(userId), HttpStatus.valueOf(200))
    }
```
which doesn't respect the schema responses code.

Providing the option `responseEntity=false` it generates 
```
fun usersUserIdGet(@ApiParam(value = "", required=true) @PathVariable("userId") userId: kotlin.String
): ResponseEntity<User> {
        return service.usersUserIdGet(userId)

    }
```

and for the service
```
interface UsersApiService {
    fun usersUserIdGet(userId: kotlin.String): ResponseEntity<User>
}
```


wing328:
cc @jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10) @4brunu (2019/11) @yutaka0m (2020/03)

4brunu:
First of all thanks for this PR.
Could you please apply this setting to a sample project to see the change in practice?
Thanks

frecco75:
@4brunu Yes I have pushed a sample project [here](https://github.com/frecco75/openapigenerator.sample).

The use case is basic (and we can handle this with Spring Exception Handling), but in general it's nice to handle directly with response entities to provide custom status codes, specific headers etc.

4brunu:
Sorry if I mis communicate, I mean create a sample project here in `samples/server/petstore/kotlin-springboot-blablabla/...` to ensure that this feature continues to work as expected in the future.

frecco75:
> Sorry if I miscommunicate, I mean create a sample project here in `samples/server/petstore/kotlin-springboot-blablabla/...` to ensure that this feature continues to work as expected in the future.

done [d9a93d6](https://github.com/OpenAPITools/openapi-generator/pull/10373/commits/d9a93d68b4e9aa947d91562daee4e25b4a4015ce)

PR #10374, opened 2021-09-10
[kotlin-spring] kotlin-builder support

frecco75:
This PR adds the support of KotlinBuilder
(See https://github.com/Pozo/mapstruct-kotlin).
@KotlinBuilder is interesting when we share models with java projects, because each annotated class will have a corresponding Builder class.

With this PR, each generated model will be annotated with `@KotlinBuilder`.

Given the schema
```
openapi: "3.0.1"
info:
  version: 1.0.0
  title: Users

paths:
  /users:
    get:
      responses:
        200:
          description: users
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/User'

components:
  schemas:
    User:
      type: object
      required:
        - name
      properties:
        name:
          type: string

```

When we compile with the option `kotlinBuilder=true` it generates
```
@KotlinBuilder data class User(
    @ApiModelProperty(example = "null", required = true, value = "")
    @field:JsonProperty("name", required = true) val name: kotlin.String
)
```

Note :  for the moment the [Lombok compiler plugin](https://kotlinlang.org/docs/lombok.html) doesn't support @Builder, so this library is useful

### DIFF
--- a/docs/generators/kotlin-spring.md
+++ b/docs/generators/kotlin-spring.md
@@ -18,6 +18,7 @@ These options may be applied as additional-properties (cli) or configOptions (pl
 |gradleBuildFile|generate a gradle build file using the Kotlin DSL| |true|
 |groupId|Generated artifact package's organization (i.e. maven groupId).| |org.openapitools|
 |interfaceOnly|Whether to generate only API interface stubs without the server files.| |false|
+|kotlinBuilder|use kotlin builder| |false|
 |library|library template (sub-template)|<dl><dt>**spring-boot**</dt><dd>Spring-boot Server application.</dd></dl>|spring-boot|
 |modelMutable|Create mutable models| |false|
 |modelPackage|model package for generated code| |org.openapitools.model|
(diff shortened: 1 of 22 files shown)


wing328:
cc @jimschubert (2017/09) ❤️, @dr4ke616 (2018/08) @karismann (2019/03) @Zomzog (2019/04) @andrewemery (2019/10) @4brunu (2019/11) @yutaka0m (2020/03)

4brunu:
First of all thanks for this PR.
Looking at https://github.com/Pozo/mapstruct-kotlin it seems to be deprecated.
Shouldn't we prefer https://github.com/mapstruct/mapstruct instead?
Or it doesn't give support for what you need?

frecco75:
@4brunu the original repo `mapstruct` doesn't provide @Builder generation and is a library to generates objects mappings.

The expected solution should be to use Lombok's builder, but it's not (for the moment) supported in Kotlin.
https://kotlinlang.org/docs/lombok.html

4brunu:
Could you please apply this setting to a sample project to see the change in practice?

wing328:
@frecco75 thanks for the PR. When you've time, can you please PM me via Slack when you've time? https://join.slack.com/t/openapi-generator/shared_invite/zt-12jxxd7p2-XUeQM~4pzsU9x~eGLQqX2g

frecco75:
I decline this Pull request because Kotlin ecosystem is now working on Lombok support, which provide a Builder.
https://youtrack.jetbrains.com/issue/KT-46959/Kotlin-Lombok-Support-generated-builders-Builder

See official documentation:
https://kotlinlang.org/docs/lombok.html

PR #10416, opened 2021-09-18
[gradle]: allow setting custom engines other than handlebars

PhilippHeuer:
The openapi-gradle generator plugin has a strict check on the `engine` param to only set the parameter if `handlebars` is set as value. Otherwise the parameter is ignored. This is currently blocking anyone from specifiying custom template engines (https://openapi-generator.tech/docs/templating/#custom-engines) when using the official gradle plugin.

This change will set the `engine` param to the user-provided value, even if it's not `handlebars`.

The engine param was initially added in: https://github.com/OpenAPITools/openapi-generator/pull/5686

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt
+++ b/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt
@@ -644,6 +644,8 @@ open class GenerateTask : DefaultTask() {
             engine.ifNotEmpty { value ->
                 if ("handlebars".equals(value, ignoreCase = true)) {
                     configurator.setTemplatingEngineName("handlebars")
+                } else {
+                    configurator.setTemplatingEngineName(value)
                 }
             }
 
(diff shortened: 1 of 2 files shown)


wing328:
>  This change will set the engine param to the user-provided value, even if it's not handlebars.

Thanks for the PR. What other values are allowed?

Shall we validate the input to only allow `handlebars`, `mustache` ?

PhilippHeuer:
> > This change will set the engine param to the user-provided value, even if it's not handlebars.
> 
> Thanks for the PR. What other values are allowed?
> 
> Shall we validate the input to only allow `handlebars`, `mustache` ?

When working with a custom engine (https://openapi-generator.tech/docs/templating/#custom-engines) - example https://github.com/twitch4j/nitro-codegen/blob/master/template-engine/src/main/java/com/github/twitch4j/codegen/engine/PebbleEngineAdapter.java#L26 - it's important that users can provide the custom identifier value (`peb` in the example), otherwise it's not possible to use custom template engine adapters with the gradle plugin.
Therefore any value can be valid, this depends on the identifiers used in custom template engine adapters - this makes value validation hard.

If users provide a invalid value than the openapi-generator will return this kind of error:
```
Exception in thread "main" java.lang.RuntimeException: Couldn't load template engine adapter my-custom-engine. Available options:
mustache
handlebars
```
Therefore i don't think we need this kind of validation in the gradle plugin, as we already get good error messages from the generator.

wing328:
@PhilippHeuer thanks for the PR and the explanation. Makes sense to me.

PR #10462, opened 2021-09-23
[bash] force content length in empty post

Ricordel:
In the case where we are making empty posts, use -d '' to force curl to set the Content-Length header to 0. Some backends (among which Microsoft Kestrel) are sensitive about POST without a Content-Length set.

This pull request adds a `-d ''` option to curl's command-line in such a situation

I don't have an extensive test setup to provide, for a simple change like this I'm not sure it's necessary. Please let me know if I should add things.

As this is bash: CC @frol, @bkryza, @kenjones-cisco


wing328:
LGTM but I've not tested it locally

wing328:
CI failures not related to this PR.

Issue #10477, opened 2021-09-27
[REQ] Feature Request Description


javier-garcia-meteologica:
### Is your feature request related to a problem? Please describe.

In javascript/typescript throwing anything but an error is considered an antipattern. Only errors should be thrown because:
  - Only errors can provide a reliable stacktraces
  - Makes error handling easier, since 99.99% of caught "things" will have a common interface `{ message: string }`.
  
Right now typescript-fetch throws a bare `Response` which is undesirable.
https://github.com/OpenAPITools/openapi-generator/blob/60b29e1f8e267e90d4d81f30d3fb3fe69161e59b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L38-L42

## Describe the solution you'd like

Typescript-fetch should throw a custom error in which the response is just a property, just like it does with RequiredError.
https://github.com/OpenAPITools/openapi-generator/blob/60b29e1f8e267e90d4d81f30d3fb3fe69161e59b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L104-L109

E.g.

```ts
        const response = await this.fetchApi(url, init);
        if (response.status >= 200 && response.status < 300) {
            return response;
        }
        throw new ResponseError(response, 'Response returned an error code');

...

export class ResponseError extends Error {
    name: "ResponseError" = "ResponseError";
    constructor(public response: string, msg?: string) {
        super(msg);
    }
}
```

Finrod927:
Hi!

This change has been shipped in version 6.0 and I'm facing an issue since I moved to this version.

On 4xx requests, our API always return a custom Error Schema with some internal error code, class and messages. It helps the front-end to handle correctly these errors and display a precise error message to the user.

```ts
    async function addUser(user: User) {
      try {
        const newUserId = await api.administration.addUser({ user })

        // Display success message
      } catch (e) {
        // e contains error information from API, used to display the error message to the end user
      }
    }
```

But since this change, I cannot get the response body anymore to have these error information.

```ts
    async function addUser(user: User) {
      try {
        const newUserId = await api.administration.addUser({ user })

        // Display success message
      } catch (e) {
        // e only contains a basic object with a message "Response returned an error code"
      }
    }
```

Can the new `ResponseError` class be updated to contains also the request Response?

Thx :)

macjohnny:
@javier-garcia-meteologica you can file a PR to add this information to the exception

Finrod927:
Hi!

[...]

PR #10478, opened 2021-09-27
[typescript-fetch]: use ReponseError instead of throwing a response (#10477)

javier-garcia-meteologica:
### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

@TiFu @taxpon @sebastianhaas  @kenisteward  @Vrolijkx @macjohnny @topce @akehir @petejohansonxo @amakhrov

### DIFF
--- a/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
+++ b/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
@@ -39,7 +39,7 @@ export class BaseAPI {
         if (response.status >= 200 && response.status < 300) {
             return response;
         }
-        throw response;
+        throw new ResponseError(response, 'Response returned an error code');
     }
 
     private createFetchParams(context: RequestOpts, initOverrides?: RequestInit) {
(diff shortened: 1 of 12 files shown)


macjohnny:
Thanks for your contribution.
Since this is a breaking change, please file against the 6.0 branch.
@amakhrov what is your opinion here? Should we introduce an option?

amakhrov:
> Should we introduce an option?

I don't think so. Migrating existing code should be trivial enough.

modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache
```diff
@@ -101,6 +101,13 @@ export class BaseAPI {
     }
 };
 
+export class ResponseError extends Error {
```
amakhrov, line 104: note that this comes with some gotchas and generally won't work if your TS compile target is `es5`.

e.g. see https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work

javier-garcia-meteologica: As a workaround, you can match by error name if your code needs to be compiled to es5

```ts
if (err.name === 'ResponseError') {
   ...
}
```

mkusaka:
@macjohnny @amakhrov Is there any possibility this change will be included in version 6.0?
Currently, it seems that this pr is not included to 6.0 milestone.
Thanks.

macjohnny:
@amakhrov are you fine with this change?

amakhrov:
@macjohnny yes, it looks reasonable, and is definitely better than throwing an arbitrary response object

PR #10481, opened 2021-09-27
Generated example for composed schema

SergiiKusii:
* Generated min items for array example

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPythonCodegen.java
+++ b/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPythonCodegen.java
@@ -20,6 +20,7 @@ import com.github.curiousoddman.rgxgen.RgxGen;
 import io.swagger.v3.oas.models.examples.Example;
 import io.swagger.v3.oas.models.media.ArraySchema;
 import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.parameters.Parameter;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -355,6 +356,18 @@ public abstract class AbstractPythonCodegen extends DefaultCodegen implements Co
                 LOGGER.warn("allDefinitions not defined in toExampleValue!\n");
             }
         }
+        else if (schema instanceof ComposedSchema)
+        {
+            ComposedSchema cs = (ComposedSchema) schema;
+
+            if (cs.getAllOf() != null) {
+                return toExampleValueRecursive(cs.getAllOf().get(0), includedSchemas, indentation);
+            } else if (cs.getAnyOf() != null) { // anyOf
+                return toExampleValueRecursive(cs.getAnyOf().get(0), includedSchemas, indentation);
+            } else if (cs.getOneOf() != null) { // oneOf
+                return toExampleValueRecursive(cs.getOneOf().get(0), includedSchemas, indentation);
+            }
+        }
         if (ModelUtils.isDateSchema(schema)) {
             example = "datetime.datetime.strptime('1975-12-30', '%Y-%m-%d').date()";
             return example;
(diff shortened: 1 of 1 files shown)


modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractPythonCodegen.java
```diff
@@ -355,6 +356,18 @@ private String toExampleValueRecursive(Schema schema, List<String> includedSchem
                 LOGGER.warn("allDefinitions not defined in toExampleValue!\n");
             }
         }
+        else if (schema instanceof ComposedSchema)
+        {
+            ComposedSchema cs = (ComposedSchema) schema;
+
+            if (cs.getAllOf() != null) {
```
spacether, line 363: What about the use case when allOf + oneOf + anyOf are combined?
Your code assumes that only one will be used when all 3 can be used in combination.

SergiiKusii: We need only 1 example here so the first chosen example will be correct. It is the simplest approach that allows covering most cases with combined properties (especially with discriminator and allof https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#allof). 

if properties/items/enums are defined on the cs composed schema it will be handled because we have recursion here toExampleValueRecursive

If we want to handle all examples it is a very different task.

I agree about the warning, maybe it would be useful here.

spacether: > if properties/items/enums are defined on the cs composed schema it will be handled because we have recursion here toExampleValueRecursive

Not sure if you are thinking of the same schema as I am. Here is what I mean for properties/items/additionalProeprties/enums + composition.
Your code doesn't handle this use case:
```
ComposedSchema:
    properties:
        name:
            type: string
    allof:
        - $ref: components.../Box
Box:
    properties:
        width:
            type: integer
```

SergiiKusii: it does but your use case has no example so it does not show anything. This function handles an example only
I added this logic to be able to handle use-cases like this
```
ComposedSchema:
    properties:
        name:
            type: string
    allOf:
        - $ref: #/components/schemas/Box
Box:
    type: object
    properties:
        width:
            type: integer
    example: {"width": 17}
```
it will go into `Box` and will take the example

SergiiKusii: or following schema 
```
ComposedSchema:
    properties:
        box:
            description: "some property with allof"
            allOf:
                - $ref: #/components/schemas/Box
Box:
    type: object
    properties:
        width:
            type: integer
    example: {"width": 17}
```

spacether: Right but it will be missing the name property. What if name is required in ComposedSchema?

SergiiKusii: yes, correct. I agreed with adding a warning from the beginning :)

In order to cover more use cases we can
-  try to somehow merge examples for allof
-  iterate through oneOf while an example will be found
-  iterate through anyOf while an example will be found

SergiiKusii: from your example, what result do you expect?

spacether: For ComposedSchema w/ PermissiveName + ConstrainedName I expect:
```
{
    "name": "some_string"
}
```
For ComposedSchema w/ Box (assuming that name is required in ComposedSchema) I expect:
```
{
    "name": "some_string",
    "width": 1
}
```
A most robust solution would allow us to combine all schemas into one schema and then generate a sample for that if it is possible. One can do horrible things like:
type: string
enum:
- 0
For which no data can ever meet the requirement.

SergiiKusii: but this example you should specify in yaml
```
{
    "name": "some_string",
    "width": 1
}
```
and everything will work.

spacether: I think that if one specified it as an example in ComposedSchema it would work. That should probably be verified.
From what I've seen most people don't specify examples, so the point of this code is to autogen them when we can, right?

SergiiKusii: yes, exactly

spacether: Great. It sounds like we agree. Can you update your code so it only handles the use case I described and returns null and logs a warning otherwise? Could you also add a test verifying that it works with your use case and returns null and logs a warning otherwise?

PR #10541, opened 2021-10-06
Bump dart2 code compatibility to Dart 2.14 and introduce sound null-safety

jld3103:
@jaumard @josh-burton @amondnet @sbu-WBT @kuhnroyal @agilob @ahmednfwela

I actually started working on this a few weeks ago, before any of the current PRs against the dart2 generator were made.
I'm unsure what to do now, but I want to put out my work, so that we can have a great dart2 generator.
Anyway I update the Dart version to the latest stable and thus implemented null-safety. I wasn't able to actually test the generated code yet, but there are no errors shown anymore (due to the project I'm generated a client for using `oneOf`).

Related PRs:
https://github.com/OpenAPITools/openapi-generator/pull/10532
https://github.com/OpenAPITools/openapi-generator/pull/10536

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 


agilob:
This PR does more than adding null safety to dart generator, I think it should be split into a few smaller PRs. A few general comments:

* I would go as far as dart 2.14, but something older, this might be too big change for some
* ignore and remove all changes related to json_serializable, consider this generator gone. Ideally, #10532 would be merged before this one
* we tend to keep changes in dart-dio and dart-dio-next and dart in separate PRs
* with removal of generated test files this is even more braking change, this risks deleting user developed tests

modules/openapi-generator/src/main/resources/dart2/api_helper.mustache
```diff
@@ -4,16 +4,16 @@ class QueryParam {
   const QueryParam(this.name, this.value);
 
   final String name;
-  final String value;
+  final String? value;
```
agilob, line 7: can query have name but no value? and for the line below, it has to be empty string not `null` string, as end user won't be able to distinguish intentional `null` string from no value. I think both fields should be not-null in this class

modules/openapi-generator/src/main/resources/dart2/api_exception.mustache
```diff
@@ -6,9 +6,9 @@ class ApiException implements Exception {
   ApiException.withInner(this.code, this.message, this.innerException, this.stackTrace);
 
   int code = 0;
-  String message;
-  Exception innerException;
-  StackTrace stackTrace;
+  String? message;
```
agilob, line 9: another design question, when it makes sense to raise ApiException without a message?

modules/openapi-generator/src/main/resources/dart2/auth/oauth.mustache
```diff
@@ -3,7 +3,7 @@
 class OAuth implements Authentication {
   OAuth({this.accessToken});
 
-  String accessToken;
+  String? accessToken;
```
agilob, line 6: by making this not-null you will avoid null check below. does it make sense to use OAuth without access token? Technically yes, but practically that's a sec vuln on endpoint?

modules/openapi-generator/src/main/resources/dart2/api_client.mustache
```diff
@@ -66,10 +61,10 @@ class ApiClient {
     String path,
     String method,
     List<QueryParam> queryParams,
-    Object body,
+    Object? body,
     Map<String, String> headerParams,
```
agilob, line 65: `headerParams` and `formParams` are nullable in other openapi generators, so I would prefer to keep it this way for language agnostic users

agilob:
This branch was only passing tests because github actions have not kicked in and many tests didn't run...

PR #10610, opened 2021-10-16
fix: #1466 additionalProperties works now in spring generator

### LINKED ISSUE #1466: [java] Combining "properties" and "additionalProperties" in one class generates broken Jackson mapping.
##### Description

If using both "properties" and "additionalProperties", the generated class extends HashMap and contains plain properties. Like this:
```java
public class Example extends HashMap<...>  {
  @JsonProperty("a")
  private String a;
  ...
}
```

Such a class cannot be meaningfully deserialised using Jackson. Jackson puts all the fields as key/value pairs into the HashMap ignoring the plain fields. In provided example getA() would return "null", while get("a") contains the value.

Used version: openapi-generator-maven-plugin : 3.3.3

##### OpenAPI declaration file

```yaml
Example:
  type: object
  properties:
    a:
    type: string
  additionalProperties: true
```

I would suggest to generate such a class (which has both "properties" and "additionalProperites") as:
```java
public class Example {
  private String a;
  private HashMap<> otherProperties;
  ...
}
```

Similar suggestion in "swagger-codegen" project:
https://github.com/swagger-api/swagger-codegen/issues/5187

benfonty:
When additional properties is true, use an additionalProperties attribute and @JsonAnySetter and @JsonAnyGetter instead of inherit from Map. I followed what was done for java.jersey2 in https://github.com/OpenAPITools/openapi-generator/pull/6518.

This fixes #1466 only for spring target.

cc @nmuesch 

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

### DIFF
--- a/samples/server/petstore/spring-mvc-j8-async/src/main/java/org/openapitools/model/AdditionalPropertiesAnyType.java
+++ b/samples/server/petstore/spring-mvc-j8-async/src/main/java/org/openapitools/model/AdditionalPropertiesAnyType.java
@@ -11,16 +11,13 @@ import java.time.OffsetDateTime;
 import javax.validation.Valid;
 import javax.validation.constraints.*;
 
-<<<<<<< HEAD
 
 import java.util.*;
 
-=======
 import java.util.Map;
 import java.util.HashMap;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
->>>>>>> be1ba96d0a1... fix: #1466 additionalProperties works now in spring generator
 /**
  * AdditionalPropertiesAnyType
  */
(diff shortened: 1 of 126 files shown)


bfrggit:
Q1: What if `additionalProperties` is not boolean, (i.e., the additional properties have a certain type of values)? Does this also work?
Q2: Can this be done for the Java generator too?

benfonty:
> Q1: What if `additionalProperties` is not boolean, (i.e., the additional properties have a certain type of values)? Does this also work? 

Yes, if the type is string for example, additionalAttributes has type Map<String, String> instead of Map<String, Object>.

> Q2: Can this be done for the Java generator too?

This can be done for all java generators, but the code might differ from generator to generator, as the template are not always the same. quote https://github.com/OpenAPITools/openapi-generator/pull/6518/files#diff-e939f76e33b14e4de9645853ba5bdeb050b5fa0b6676693ff900fd60aaf08528R375

jameswynn:
The failed bitrise test looks like an intermittent failure. Is that all that is blocking this PR?

benfonty:
> The failed bitrise test looks like an intermittent failure. Is that all that is blocking this PR?

I don't know. There were two failed builds with the same error that seems indeed intermittent. I am not sure what to do to fix it.

jameswynn:
@cbornet The normal build passes, but bitrise chokes. Any advice?

PR #10782, opened 2021-11-04
[Python] remove default content-type in the request

### LINKED ISSUE #8116: [REQ] Fix default Content-Type Header in Python client

Currently in the Python client (at least, I have not verified for other generated clients) when there is no `requestBody` key specified for a given path in the spec file the generated client will automatically set the `Content-Type` header to `application/json`. The biggest issue with this is that an empty request body is NOT valid JSON. Therefore the client effectively sends an invalid request to the server each time there is no request body specified. This is not a problem most of the time, however I have encountered issues with this when running against a Rails server. 

The `Content-Type` header should default to either `text/plain` or should be omitted if it is unnecessary.

kevchentw:
Before this fix, the content-type of the request defaults to "application/json".

With this fix, we no longer set such default.

fix #8116, relate to https://github.com/OpenAPITools/openapi-generator/pull/10769/

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 

cc @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11) @tomplus (2018/10) @Jyhess (2019/01) @arun-nalla (2019/11) @spacether (2019/11)

### DIFF
--- a/modules/openapi-generator/src/main/resources/python-legacy/api.mustache
+++ b/modules/openapi-generator/src/main/resources/python-legacy/api.mustache
@@ -250,10 +250,11 @@ class {{classname}}(object):
         {{/hasProduces}}
         {{#hasConsumes}}
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = local_var_params.get('_content_type',
-            self.api_client.select_header_content_type(
-                [{{#consumes}}'{{{mediaType}}}'{{^-last}}, {{/-last}}{{/consumes}}],
-                '{{httpMethod}}', body_params))  # noqa: E501
+        content_types_list = self.api_client.select_header_content_type(  # noqa: E501
+            [{{#consumes}}'{{{mediaType}}}'{{^-last}}, {{/-last}}{{/consumes}}])  # noqa: E501
+            [{{#consumes}}'{{{mediaType}}}'{{^-last}}, {{/-last}}{{/consumes}}])  # noqa: E501
+        if content_types_list:
+                header_params['Content-Type'] = content_types_list
 
         {{/hasConsumes}}
         # Authentication setting
--- a/modules/openapi-generator/src/main/resources/python-legacy/api.mustache
+++ b/modules/openapi-generator/src/main/resources/python-legacy/api.mustache
@@ -252,7 +252,6 @@ class {{classname}}(object):
         # HTTP header `Content-Type`
         content_types_list = self.api_client.select_header_content_type(  # noqa: E501
             [{{#consumes}}'{{{mediaType}}}'{{^-last}}, {{/-last}}{{/consumes}}])  # noqa: E501
-            [{{#consumes}}'{{{mediaType}}}'{{^-last}}, {{/-last}}{{/consumes}}])  # noqa: E501
         if content_types_list:
                 header_params['Content-Type'] = content_types_list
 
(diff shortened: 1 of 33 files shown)


modules/openapi-generator/src/main/resources/python-legacy/api.mustache
```diff
@@ -247,8 +247,10 @@ class {{classname}}(object):
         {{/hasProduces}}
         {{#hasConsumes}}
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
+        header_list = self.api_client.select_header_content_type(  # noqa: E501
```
spacether, line 250: How about `content_types_list = self....`
header_list sounds like it can contain any kind of headers
I understand that it was historically named that

spacether:
Can you add a test that checks that 1 or two content types are set when they are included in the spec?
Should we allow users to select a specific content type that is present in the given content_types?

kevchentw:
> Can you add a test that checks that 1 or two content types are set when they are included in the spec? Should we allow users to select a specific content type that is present in the given content_types?

Sure. Let me add some tests. 

I think we cannot let the user to specific content type right now. It defaults to use the first one. Maybe we can create another issue for this?

kevchentw:
@spacether 
I think those tests are enough?

https://github.com/OpenAPITools/openapi-generator/blob/3f3f8376b0ee363f37ec0f5a359d341b77462055/samples/client/petstore/python/tests/test_api_client.py#L104-L123

spacether:
This PR and https://github.com/OpenAPITools/openapi-generator/pull/10686#pullrequestreview-810638082 are both seeking to solve the same problem

modules/openapi-generator/src/main/resources/python-legacy/api_client.mustache
```diff
@@ -552,7 +552,7 @@ class ApiClient(object):
         :return: Content-Type (e.g. application/json).
         """
         if not content_types:
-            return 'application/json'
```
spacether, line 555: I am concerned that this could be a breaking change for servers/client that depended upon this functionality

spacether:
> @spacether I think those tests are enough?
> 
> https://github.com/OpenAPITools/openapi-generator/blob/3f3f8376b0ee363f37ec0f5a359d341b77462055/samples/client/petstore/python/tests/test_api_client.py#L104-L123

Thank you for adding those tests

modules/openapi-generator/src/main/resources/python-legacy/api_client.mustache
wing328: > I am concerned that this could be a breaking change for servers/client that depended upon this functionality.
How about adding a generator argument useJsonAsDefaultContentType and default it to True.

We've done something similar in other generators (e.g. Ruby) to remove such default behaviour (which was added before I worked on this project). No complains so far.

wing328:
@kevchentw can you please resolve the merge conflicts when you've time and then PM me via Slack?

https://join.slack.com/t/openapi-generator/shared_invite/zt-12jxxd7p2-XUeQM~4pzsU9x~eGLQqX2g

PR #10790, opened 2021-11-05
[swift5] Reuse URLSessions

### LINKED ISSUE #8562: [REQ] [swift5] Reuse URLSessions for requests

After doing a little research on best practices for URLSession, I stumbled upon this: https://developer.apple.com/videos/play/wwdc2018/714/?time=1646

> Now let's see how using fewer URLSession objects can help reduce latency.
> All the benefits of connections we use that we just discussed in the previous slides are applicable only if you use the same URLSession object to create your tasks.
> It's also important to know that every URLSession object has a connection pool and when you create multiple of these URLSession objects, you don't get any benefit of connection to use.
> It's also important to note that the URLSession objects are fairly expensive to create and have a non-trivial memory footprint.

- I'd suggest having a dictionary of `[Configuration: URLSession]` that will return you an existing session object for your configuration. 
- Configuration must be a hash of `URLSessionConfiguration`, and optional `URLSessionDelegate` and `OperationQueue` as [you can initialize](https://developer.apple.com/documentation/foundation/urlsession/1411597-init) your `URLSession` with those parameters.

- We could leave it as it is now, since you can achieve the same thing by subclassing the request builder.
  - Also, possible gains may not be noticeable since today's hardware is extremely fast.

## Additional context

- the client with Alamofire option doesn't reuse url sessions either
  - here we're creating a new `SessionManager` object for each request, which under the hood of Alamofire 4.x, creates a new `URLSession`:  https://github.com/OpenAPITools/openapi-generator/blob/150e24dc553a8ea5230ffb938ed3e6020e972faa/modules/openapi-generator/src/main/resources/swift4/AlamofireImplementations.mustache#L64

DavidHorvath:
resolve #8562

Creating new URLSession for every request isn't optimal and it is not recommended by Apple.
Also, `NSPOSIXErrorDomain Code=28 "No space left on device"` will get thrown after too many URLSessions get created. 

> https://developer.apple.com/videos/play/wwdc2018/714/?time=1646
> https://developer.apple.com/videos/play/wwdc2017/709/
> https://developer.apple.com/forums/thread/84663
> https://stackoverflow.com/questions/67318867/error-domain-nsposixerrordomain-code-28-no-space-left-on-device-userinfo-kcf/67507327#67507327

I've hashed the session's configuration, credentials and sessionDelegate with Hasher so we can reuse URLSessions with the same settings.

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [x] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request. @4brunu

### DIFF
--- a/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
@@ -19,8 +19,7 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
     }
 }
 
-// Store the URLSession to retain its reference
-private var urlSessionStore = SynchronizedDictionary<Int, URLSession>()
+{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} typealias {{projectName}}APIChallengeHandler = ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))
 
 // Store the URLSession's delegate to retain its reference
 private let sessionDelegate = SessionDelegate()
@@ -29,7 +28,7 @@ private let sessionDelegate = SessionDelegate()
 private let defaultURLSession = URLSession(configuration: .default, delegate: sessionDelegate, delegateQueue: nil)
 
 // Store current taskDidReceiveChallenge for every URLSessionTask
-private var challengeHandlerStore = SynchronizedDictionary<Int, ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))>()
+private var challengeHandlerStore = SynchronizedDictionary<Int, {{projectName}}APIChallengeHandler>()
 
 // Store current URLCredential for every URLSessionTask
 private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
@@ -39,7 +38,7 @@ private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
     /**
      May be assigned if you want to control the authentication challenges.
      */
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} var taskDidReceiveChallenge: ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))?
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} var taskDidReceiveChallenge: {{projectName}}APIChallengeHandler?
 
     /**
      May be assigned if you want to do any of those things:
--- a/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
@@ -20,7 +20,19 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
 }
 
 // Store the URLSession to retain its reference
-private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
+private var urlSessionStore = SynchronizedDictionary<Int, URLSession>()
+
+// Store the URLSession's delegate to retain its reference
+private let sessionDelegate = SessionDelegate()
+
+// Store the URLSession to retain its reference
+private let defaultURLSession = URLSession(configuration: .default, delegate: sessionDelegate, delegateQueue: nil)
+
+// Store current taskDidReceiveChallenge for every URLSessionTask
+private var challengeHandlerStore = SynchronizedDictionary<Int, ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))>()
+
+// Store current URLCredential for every URLSessionTask
+private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
 
 {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class URLSessionRequestBuilder<T>: RequestBuilder<T> {
 
@@ -47,12 +59,7 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
      configuration.
      */
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func createURLSession() -> URLSession {
-        let configuration = URLSessionConfiguration.default
-        configuration.httpAdditionalHeaders = buildHeaders()
-        let sessionDelegate = SessionDelegate()
-        sessionDelegate.credential = credential
-        sessionDelegate.taskDidReceiveChallenge = taskDidReceiveChallenge
-        return URLSession(configuration: configuration, delegate: sessionDelegate, delegateQueue: nil)
+        return defaultURLSession
     }
 
     /**
@@ -94,10 +101,7 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
     }
 
     override {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) {
-        let urlSessionId = UUID().uuidString
-        // Create a new manager for each request to customize its request header
         let urlSession = createURLSession()
-        urlSessionStore[urlSessionId] = urlSession
 
         guard let xMethod = HTTPMethod(rawValue: method) else {
             fatalError("Unsupported Http method - \(method)")
@@ -123,13 +127,16 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
             }
         }
 
-        let cleanupRequest = {
-            urlSessionStore[urlSessionId]?.finishTasksAndInvalidate()
-            urlSessionStore[urlSessionId] = nil
-        }
-
         do {
             let request = try createURLRequest(urlSession: urlSession, method: xMethod, encoding: encoding, headers: headers)
+
+            var taskIdentifier: Int?
+             let cleanupRequest = {
+                 if let taskIdentifier = taskIdentifier {
+                     challengeHandlerStore[taskIdentifier] = nil
+                     credentialStore[taskIdentifier] = nil
+                 }
+             }
 
             let dataTask = urlSession.dataTask(with: request) { data, response, error in
 
@@ -159,11 +166,14 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
                 onProgressReady?(dataTask.progress)
             }
 
+            taskIdentifier = dataTask.taskIdentifier
+            challengeHandlerStore[dataTask.taskIdentifier] = taskDidReceiveChallenge
+            credentialStore[dataTask.taskIdentifier] = credential
+
             dataTask.resume()
 
         } catch {
             apiResponseQueue.async {
-                cleanupRequest()
                 completion(.failure(ErrorResponse.error(415, nil, nil, error)))
             }
         }
@@ -398,25 +408,20 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
     }
 }
 
-private class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
-
-    var credential: URLCredential?
-
-    var taskDidReceiveChallenge: ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))?
-
+private class SessionDelegate: NSObject, URLSessionTaskDelegate {
     func urlSession(_ session: URLSession, task: URLSessionTask, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
 
         var disposition: URLSession.AuthChallengeDisposition = .performDefaultHandling
 
         var credential: URLCredential?
 
-        if let taskDidReceiveChallenge = taskDidReceiveChallenge {
+        if let taskDidReceiveChallenge = challengeHandlerStore[task.taskIdentifier] {
             (disposition, credential) = taskDidReceiveChallenge(session, task, challenge)
         } else {
             if challenge.previousFailureCount > 0 {
                 disposition = .rejectProtectionSpace
             } else {
-                credential = self.credential ?? session.configuration.urlCredentialStorage?.defaultCredential(for: challenge.protectionSpace)
+                credential = credentialStore[task.taskIdentifier] ?? session.configuration.urlCredentialStorage?.defaultCredential(for: challenge.protectionSpace)
 
                 if credential != nil {
                     disposition = .useCredential
--- a/samples/client/petstore/swift5/default/PetstoreClient/Classes/OpenAPIs/URLSessionImplementations.swift
+++ b/samples/client/petstore/swift5/default/PetstoreClient/Classes/OpenAPIs/URLSessionImplementations.swift
@@ -19,15 +19,26 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
     }
 }
 
+public typealias PetstoreClientAPIChallengeHandler = ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))
+
+// Store the URLSession's delegate to retain its reference
+private let sessionDelegate = SessionDelegate()
+
 // Store the URLSession to retain its reference
-private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
+private let defaultURLSession = URLSession(configuration: .default, delegate: sessionDelegate, delegateQueue: nil)
+
+// Store current taskDidReceiveChallenge for every URLSessionTask
+private var challengeHandlerStore = SynchronizedDictionary<Int, PetstoreClientAPIChallengeHandler>()
+
+// Store current URLCredential for every URLSessionTask
+private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
 
 open class URLSessionRequestBuilder<T>: RequestBuilder<T> {
 
     /**
      May be assigned if you want to control the authentication challenges.
      */
-    public var taskDidReceiveChallenge: ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))?
+    public var taskDidReceiveChallenge: PetstoreClientAPIChallengeHandler?
 
     /**
      May be assigned if you want to do any of those things:
@@ -47,12 +58,7 @@ open class URLSessionRequestBuilder<T>: RequestBuilder<T> {
      configuration.
      */
     open func createURLSession() -> URLSession {
-        let configuration = URLSessionConfiguration.default
-        configuration.httpAdditionalHeaders = buildHeaders()
-        let sessionDelegate = SessionDelegate()
-        sessionDelegate.credential = credential
-        sessionDelegate.taskDidReceiveChallenge = taskDidReceiveChallenge
-        return URLSession(configuration: configuration, delegate: sessionDelegate, delegateQueue: nil)
+        return defaultURLSession
     }
 
     /**
@@ -94,10 +100,7 @@ open class URLSessionRequestBuilder<T>: RequestBuilder<T> {
     }
 
     override open func execute(_ apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) {
-        let urlSessionId = UUID().uuidString
-        // Create a new manager for each request to customize its request header
         let urlSession = createURLSession()
-        urlSessionStore[urlSessionId] = urlSession
 
         guard let xMethod = HTTPMethod(rawValue: method) else {
             fatalError("Unsupported Http method - \(method)")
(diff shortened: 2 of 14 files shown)


4brunu:
@DavidHorvath Hey, first of all, thanks for this contribution.
Everything looks good to me.
Just one question, did you tested this in a project of yours?
Thanks

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -52,7 +52,22 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
...
+
+        var hasher = Hasher()
+        hasher.combine(configuration)
+        hasher.combine(credential)
+        hasher.combine(sessionDelegate)
+
+        let hash = hasher.finalize()
+
+        if let session = urlSessionStore[hash] {
+            return session
```
4brunu, line 64: @DavidHorvath I was testing this locally, but it's not reusing sessions.
I set a breakpoint here, but it never hit the breakpoint.
The problem seems to be related to `hasher.combine(sessionDelegate)`

DavidHorvath: @4brunu Hi, yeah you are right. I've tried it before adding the sessionDelegate to the Hasher. 
Sorry about that. I will update my PR

DavidHorvath: @4brunu Updated it

DavidHorvath: @4brunu But I need to update the sample codes as well so just another minute

4brunu: No problem 🙂 I'm just wondering if this is safe or not without the SessionDelegate.
Imagine the following scenario.
The user makes a request to the endpoint A.
Then he repetes the same request.
The second request will invoque the first delegate, not the second one.
Or am I thinking wrong?

DavidHorvath: I think we won't have any issue with this, because we only have one `requestBuilderFactory` and in that one `sessionDelegate` which is implemented privately in the `URLSessionImplementations` as well. 
So it won't change

DavidHorvath: Meanwhile I've updated my PR with the sample codes as well

DavidHorvath:
@4brunu Yeah, we use it in two of our projects without any issue. 
We had the `NSPOSIXErrorDomain Code=28 "No space left on device"` error and this solved it as well.

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
4brunu: We only have one `requestBuilderFactory`, but the `requestBuilderFactory` generates a new instance of `URLSessionRequestBuilder` and `URLSessionDecodableRequestBuilder` for each request, which means that each request we have a different `sessionDelegate`.

DavidHorvath: I see. Let me think about a solution

DavidHorvath: @4brunu 

So right now
- `RequestBuilderFactory` creates a new instance of `URLSessionRequestBuilder` and `URLSessionDecodableRequestBuilder` for every request
- Every time a new `URLSessionRequestBuilder` is created for a request, it creates a new `URLSession` and a new `SessionDelegate)` which handles the request challenges

But if we don’t want to create a new URLSession for each request
- We can change the SessionDelegate(URLSessionDelegate) to a TaskDelegate(URLSessionTaskDelegate) and just use that for every request

**If we want to handle credential requests in a session-level we would need to implement (URLSessionDelegate)**
urlSession(_:didReceive:completionHandler:)
https://developer.apple.com/documentation/foundation/urlsessiondelegate/1409308-urlsession

**The implementation we have currently is for requests (URLSessionTaskDelegate)**
urlSession(_:task:didReceive:completionHandler:)
https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/1411595-urlsession

https://developer.apple.com/documentation/foundation/url_loading_system/handling_an_authentication_challenge

What do you think? If you want I can add a new commit for you to check what I'm thinking about

One thing is not clear for me, what would be the preferred way to assign `taskDidReceiveChallenge` & `taskCompletionShouldRetry`? 
Can you please clarify that for me?
Thanks

4brunu: Hi @DavidHorvath,

> One thing is not clear for me, what would be the preferred way to assign taskDidReceiveChallenge & taskCompletionShouldRetry?


I would say maybe in the init method of the subclass of  `RequestBuilder`.

For example here, if there was an init method.


https://github.com/OpenAPITools/openapi-generator/blob/0bb4f186ec7e776efbe370be89798ed2f659c9ff/samples/client/petstore/swift5/urlsessionLibrary/SwaggerClientTests/SwaggerClient/BearerDecodableRequestBuilder.swift#L22


I'm not to worried about `taskCompletionShouldRetry`.

My main concern is with `taskDidReceiveChallenge` and the credential handle, that could be address with your solution of replacing the SessionDelegate.

If you could add a commit to show your idea of replacing the `URLSessionDelegate` with the `URLSessionTaskDelegate` that would be nice.
I think that maybe the could be something on the right track, but looking at the concrete implementation of it would make things easier.

Thanks for this PR, this is a nice improvement 👍

DavidHorvath: @4brunu I've added 2 new commits:

- Add separate TaskDelegate as data task's delegate
- Store URLSession's delegate to keep reference

**Handle Task/Request Challenge**

1. Renamed `SessionDelegate` to `TaskDelegate`. The current delegate implementation `urlSession(_:task:didReceive:completionHandler: )` is in the [URLSessionTaskDelegate](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/1411595-urlsession)
so I changed the delegates to `URLSessionTaskDelegate` as well. (This is inherited from `URLSessionDelegate` so we can use the same TaskDelegate for the URLSession later as well)
2. Instantiate a new private TaskDelegate for the URLSessionRequestBuilder, set the credentials and taskDidReceiveChallenge and set it as the `dataTask.delegate` so we can handle the challenge for each request.

**Handle Session Challenge**

- Later if we want/need, we can have a separate SessionDelegate class which implements URLSessionDelegate and [ urlSession(_:didReceive:completionHandler:)](https://developer.apple.com/documentation/foundation/urlsessiondelegate/1409308-urlsession) to handle session level challenges
- **The documentation says:** `If you do not implement this method, the session calls its delegate’s urlSession(_:task:didReceive:completionHandler:) method instead.` 
- So to handle session level challenges (until we implement a separate SessionDelegate) we can use the TaskDelegate for the URLSession as well

So what I did:
1. I've created a new `sessionDelegateStore` to store the delegates for the URLSessions
2. If we don't have a stored URLSession, I create a new TaskDelegate instance for the new URLSession, set the credentials and taskDidReceiveChallenge (as before) and store it in the sessionDelegateStore

Questions/Notes:
- Is it okay that I renamed `SessionDelegate` for `TaskDelegate` even if we use it for the URLSession as well? Basically it is a task delegate but as the documentation says if we don't have the session delegate implementation it will use the task delegate to handle the challenges
- Currently both the URLSession's challenge and the DataTask's challenge will call the same `taskDidReceiveChallenge`. If we want to have a separate `SessionDelegate` we will need another new closure like `sessionDidReceiveChallenge`

What do you think about it? Please let me know if I missed something

4brunu: Hey,

Overall this looks good to me, setting the delegate on the `URLSessionTask` is a great idea.
But I found one big problem, the `delegate` of `URLSessionTask` is only available on iOS 15 an above.

Looking at `URLSessionTask` gave me another idea.
The `URLSessionTask` has a property `taskIdentifier`.
What if we have only one `URLSession` and only one `URLSessionDelegate` reusable for all the requests.
Inside the `URLSessionDelegate` instead of having only one `taskDidReceiveChallenge`, we would have a dictionary of them, and the `URLSessionTask.taskIdentifier` would be the key for it.
Then, when the method `func urlSession(_ session: URLSession, task: URLSessionTask, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void)` get's called, we would get the callback using the `task.taskIdentifier`.

What do you think of this?
Or do you have a better idea to workaround the iOS 15 limitation?

DavidHorvath: 😕 Yeah, you are right. I missed that this is from iOS15 only...

I think your solution can work. I will try it out

DavidHorvath: @4brunu 

I just added a new commit. Please check its okay.
If you think this is the right way I will squash my commits for the changes and generate new sample codes

DavidHorvath: Also, what do you think about adding a typealias for the closure type to be more readable?
`public typealias {{projectName}}APIChallengeHandler = ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))`

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -49,10 +55,25 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func createURLSession() -> URLSession {
         let configuration = URLSessionConfiguration.default
         configuration.httpAdditionalHeaders = buildHeaders()
```
4brunu, line 57: @DavidHorvath since now the `URLSession` is shared between all the requests, could you please remove this line?

DavidHorvath: Something like this?
```    
open func createURLSession() -> URLSession {
        var hasher = Hasher()
        hasher.combine(buildHeaders())
        hasher.combine(credential)

        let hash = hasher.finalize()

        if let session = urlSessionStore[hash] {
            return session
        }

        let configuration = URLSessionConfiguration.default
        configuration.httpAdditionalHeaders = buildHeaders()

        let sessionDelegate = SessionDelegate()
        sessionDelegate.credential = credential

        let session = URLSession(configuration: configuration, delegate: sessionDelegate, delegateQueue: nil)
        urlSessionStore[hash] = session
        sessionDelegateStore[hash] = sessionDelegate

        return session
    }
```

wing328: > @DavidHorvath since now the URLSession is shared between all the requests, could you please remove this line?

Just want to share a bit more this in other API client design.

In some clients (e.g. Java, C#), the object "ApiClient" is shared by all the Api classes (e.g. PetApi) by default. One can also create a new ApiClient instance to initialise the Api classes, e.g.
```
ApiClient apiClient = new ApiClient();
PetApi api = new PetApi(apiClient);
```

If each ApiClient instance has its own URLSession, then one can choose whether to share the same URLSession or have a separate URLSession per Api class.

Not sure if such design/approach works best in Swift so just to share more to help this work forward.

DavidHorvath:
@4brunu I've just updated my PR with the asked changes

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -19,15 +19,23 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
...
 }
 
+public typealias {{projectName}}APIChallengeHandler = ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))
+
 // Store the URLSession to retain its reference
-private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
+private var urlSessionStore = SynchronizedDictionary<Int, URLSession>()
+
+// Store the URLSession's delegate to retain its reference
+private var sessionDelegateStore = SynchronizedDictionary<Int, SessionDelegate>()
```
4brunu, line 28: @DavidHorvath since we will have only one URLSession and one SessionDelegate, I think we can replace those two line with
```
 // Store the URLSession's delegate to retain its reference
private let sessionDelegate = SessionDelegate()

// Store the URLSession to retain its reference
private let defaultURLSession = URLSession(configuration: .default, delegate: sessionDelegate, delegateQueue: nil)
```

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -19,15 +19,23 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
     }
 }
 
+public typealias {{projectName}}APIChallengeHandler = ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))
```
4brunu, line 22: @DavidHorvath can you please replace `public` with `{{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}}`?

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -49,10 +55,25 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func createURLSession() -> URLSession {
         let configuration = URLSessionConfiguration.default
         configuration.httpAdditionalHeaders = buildHeaders()
+
+        var hasher = Hasher()
```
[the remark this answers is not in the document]
4brunu: @DavidHorvath I think here we can simplify this function to 
```
    internal func createURLSession() -> URLSession {
        return defaultURLSession
    }
```

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -19,15 +19,23 @@ class URLSessionRequestBuilderFactory: RequestBuilderFactory {
...
+
 // Store the URLSession to retain its reference
-private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
+private var urlSessionStore = SynchronizedDictionary<Int, URLSession>()
+
+// Store the URLSession's delegate to retain its reference
+private var sessionDelegateStore = SynchronizedDictionary<Int, SessionDelegate>()
+
+// Store current taskDidReceiveChallenge for every URLSessionTask
+private var challengeHandlerStore = SynchronizedDictionary<Int, {{projectName}}APIChallengeHandler>()
```
4brunu, line 31: @DavidHorvath since the SessionDelegate is now shared, we also need to store the credential associated with the task identifier
```
// Store current URLCredential for every URLSessionTask
 private var credentialStore = SynchronizedDictionary<Int, URLCredential >()
```

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -159,11 +179,14 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
                 onProgressReady?(dataTask.progress)
             }
 
+            taskIdentifier = dataTask.taskIdentifier
+            challengeHandlerStore[dataTask.taskIdentifier] = taskDidReceiveChallenge
```
4brunu, line 183: @DavidHorvath we also need to store the credential associated with the task identifier here
```
credentialStore[dataTask.taskIdentifier] = credential
```

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -123,14 +141,16 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
...
-            urlSessionStore[urlSessionId] = nil
-        }
-
         do {
             let request = try createURLRequest(urlSession: urlSession, method: xMethod, encoding: encoding, headers: headers)
 
+            var taskIdentifier: Int?
+             let cleanupRequest = {
+                 if let taskIdentifier = taskIdentifier {
+                     challengeHandlerStore[taskIdentifier] = nil
```
4brunu, line 150: @DavidHorvath we also need to cleanup the credential associated with the task identifier here
```
credentialStore[taskIdentifier] = nil
```

4brunu:
@DavidHorvath thanks for making the changes, I left some comments

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
4brunu: Hi @wing328,
Here is works somehow the same way, but with the naming a bit different.

The main problem we were facing is that we are using for every request a new URLSession instance, and when you have many URLSession instances the iOS will throw an error.
`NSPOSIXErrorDomain Code=28 "No space left on device"`

So we are changing things a bit so that we have only one URLSession instance to all requests (although this is configurable and the user can change this if he wants).

So this will work more similar to the other generators, at least kotlin, that have an ApiClient instance per request, but all of them reuse the same OkHttp instance that is the equivalent of URLSession in iOS.

wing328: Thanks for the explanation 👍

DavidHorvath:
@4brunu Thanks for the review! I've fixed your comments

4brunu:
@DavidHorvath Thanks for your help fixing this issue. 
Did you test those changes in your project?
Do they fix the error `NSPOSIXErrorDomain Code=28 "No space left on device"`?

samples/client/petstore/swift5/default/PetstoreClient/Classes/OpenAPIs/URLSessionImplementations.swift
```diff
@@ -123,14 +126,17 @@ open class URLSessionRequestBuilder<T>: RequestBuilder<T> {
...
         do {
             let request = try createURLRequest(urlSession: urlSession, method: xMethod, encoding: encoding, headers: headers)
 
+            var taskIdentifier: Int?
+             let cleanupRequest = {
+                 if let taskIdentifier = taskIdentifier {
+                     challengeHandlerStore[taskIdentifier] = nil
+                     credentialStore[taskIdentifier] = nil
+                 }
+             }
```
4brunu, line 138: @DavidHorvath could you please move this block of text to line 128, just before the `do {} catch {}`

4brunu:
@DavidHorvath I have tested this on my local project and it seems that everything is working.
Could you please also test this in your project?
Before those changes, were you able to reproduce the error `NSPOSIXErrorDomain Code=28 "No space left on device"`?
Could you test to see if that error is still happening?

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -398,25 +407,20 @@ private var urlSessionStore = SynchronizedDictionary<String, URLSession>()
     }
 }
 
-private class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
-
-    var credential: URLCredential?
-
-    var taskDidReceiveChallenge: ((URLSession, URLSessionTask, URLAuthenticationChallenge) -> (URLSession.AuthChallengeDisposition, URLCredential?))?
-
+private class SessionDelegate: NSObject, URLSessionTaskDelegate {
```
4brunu, line 410: @DavidHorvath I think here it should be `URLSessionDelegate`, or am I wrong?
Before it was `URLSessionDelegate, URLSessionDataDelegate` but I think `URLSessionDelegate` should be enough?
I would like to have your feedback on this please.

DavidHorvath: The reason I changed the delegate to `URLSessionTaskDelegate` is because we actually use it to handle the challenges
[urlSession(_:task:didReceive:completionHandler:)](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/1411595-urlsession) 

`URLSessionTaskDelegate` inherits from `URLSessionDelegate` and if we don't implement the session level challenge handler [urlSession(_:didReceive:completionHandler:)](https://developer.apple.com/documentation/foundation/urlsessiondelegate/1409308-urlsession), according to the documentation
`If you do not implement this method, the session calls its delegate’s urlSession(_:task:didReceive:completionHandler:) method instead.`

4brunu: Thanks, looks good to me 👍

DavidHorvath: We have to add `URLSessionTaskDelegate` otherwise this delegate won't be called. 
The reason it worked before is because `URLSessionDataDelegate` was used which is inherited from `URLSessionTaskDelegate` so the delegate was called

DavidHorvath:
> @DavidHorvath I have tested this on my local project and it seems that everything is working.
> Could you please also test this in your project?
> Before those changes, were you able to reproduce the error `NSPOSIXErrorDomain Code=28 "No space left on device"`?
> Could you test to see if that error is still happening?

I haven't tested the latest version in our project yet but I will later today.
 I will get back to you

DavidHorvath:
@4brunu 
I've tested it on our project. Everything is working and I don't get the error message! 🚀

4brunu:
Nice, thanks for your help 😊

DavidHorvath:
> Nice, thanks for your help 😊

@4brunu Thanks for your support

PR #10824, opened 2021-11-09
Use parallel classes test over reusing forks

agilob:
I played a bit with surefire config locally. `mvn test` with config from master branch runs in 418 seconds on my laptop so this was my baseline. I tried a few different things, experimented with configuration on parallel classes, methods etc. and got to this point where `mvn test ` now completes in 280 seconds

From surefire doesn't really encourage to use fork parallelism:

>forkCount=1/reuseForks=false executes each test class in its own JVM process, one after another. It creates the highest level of separation for the test execution, but it would probably also give you the longest execution time of all the available options. Consider it as a last resort.

testng had to be downgraded to support `parallel` flag in surefire, but that will be fixed in the next version of surefire:
https://stackoverflow.com/questions/68143219/maven-surefire-failsafe-testng-7-4-0-parallel-error-void-org-testng-xml-xmlsuit

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,13 +176,12 @@
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
                     <printSummary>true</printSummary>
-                    <forkedProcessExitTimeoutInSeconds>120</forkedProcessExitTimeoutInSeconds>
+                    <parallel>classes</parallel>
+                    <threadCountClasses>3</threadCountClasses>
                     <useSystemClassLoader>false</useSystemClassLoader>
                     <junitArtifactName>none:none</junitArtifactName>
                     <testNGArtifactName>org.testng:testng</testNGArtifactName>
                     <argLine>@{argLine} -XX:+StartAttachListener</argLine>
-                    <forkCount>3</forkCount>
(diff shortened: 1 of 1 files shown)


agilob:
@wing328 @jimschubert

pom.xml
```diff
@@ -1573,7 +1572,7 @@
         <swagger-core.version>2.1.2</swagger-core.version>
         <swagger-parser-groupid.version>io.swagger.parser.v3</swagger-parser-groupid.version>
         <swagger-parser.version>2.0.26</swagger-parser.version>
-        <testng.version>7.4.0</testng.version>
+        <testng.version>7.3.0</testng.version>
```
wing328, line 1575: @agilob thanks for the PR. What's the issue in using v7.4.0?

agilob: with testng 7.4 and surefire dont play well with `parallel` flag https://stackoverflow.com/questions/68143219/maven-surefire-failsafe-testng-7-4-0-parallel-error-void-org-testng-xml-xmlsuit

wing328: 👌

Issue #10817, opened 2021-11-09
[BUG][PHP] ObjectSerializer.php throws ValueError when handling \SplFileObject type


robertsjumis:
@wing328 

##### Description

`/samples/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php`

ObjectSerializer.php throws ValueError when handling \SplFileObject type @ line 336. This is because SplFileObject can be passed to settype(), which then throws the said error.

There is no reason to handle 'object', 'mixed', '\DateTime' in line 336, because they're handled before. 

Also - there is no reason for line 336 to handle '\SplFileObject', because this specific object type is handled below.

The said object types should be deleted from the array.

##### openapi-generator version

5.3.0. - 6.0.0.

##### OpenAPI declaration file content or url

.../samples/client/petstore/php/OpenAPIClient-php/lib/ObjectSerializer.php

Line 336

##### Generation Details

php client

##### Steps to reproduce

1. Find an endpoint that returns a PDF file.
2. Generate PHP client that makes a GET request to the said endpoint.
3. Use client to call the endpoint.

##### Suggest a fix

replace line 336 with: 
`if (in_array($class, ['array', 'bool', 'boolean', 'byte', 'double', 'float', 'int', 'integer', 'number', 'string', 'void'], true)) {`

wing328:
@robertsjumis Can you please file a PR with the suggested fix? Thanks.

PR #10846, opened 2021-11-12
[PHP][BUG]Fix ValueError exception in ObjectSerializer.php

robertsjumis:
@wing328

Link to opened issue: https://github.com/OpenAPITools/openapi-generator/issues/10817


wing328:
Thanks for the PR. You will need to update the mustache template instead: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/php/ObjectSerializer.mustache

PR #10855, opened 2021-11-14
[swift5] Cancellable requests

### LINKED ISSUE #8658: [REQ] [Swift5] Cancellation of requests

Swift 5 generated code does not support **cancellation of requests**. This is the case for all libraries you can configure using `responseAs` option (PromiseKit, RxSwift, Result, Combine) and even the standard completion handler.

### Combine and RXSwift

Reactive frameworks have the concept of cancellation built-in.

The Combine generator uses `Future`, which does not support cancellation and is primarily designed to wrap APIs based on completion handlers. Similarly, the RxSwift `Observable` is created in a way the task is cannot be disposed during its execution.

### PromiseKit

PromiseKit does not have the concept of cancellation built-in as of version 6.x and therefore there is not a strong need to do any updates in here.

If the generated API call methods would return URL session data task, the user would be able to control the lifecycle of the request.

```swift
@discardableResult
open class func addPet(
    pet: Pet,
    apiResponseQueue: DispatchQueue = OpenAPIClientAPI.apiResponseQueue,
    completion: @escaping ((_ data: Pet?, _ error: Error?) -> Void)
) -> URLSessionTask
```

Discardable result is added to keep backward compatibility, so no warnings are thrown when you regenerate the code.

### Combine and RxSwift

We would neet to change the `execute` method on `RequestBuilder` to also return the data task, so we can properly wrap the data task in a `Publisher` or an `Observable`.

- *Copy the templates and rewrite them to fulfill our use-case and to have expanded feture set.* (We did and it works fine for us. I believe this is core feature of client-server communication and many of the other users could benefit from such feature.)
- *Do not care about cancellation.* (We want to optimize the data requirements and it makes sense to end some requests in case we no longer need to finish loading them. This is the case for most of the `GET` requests or more generally for all the requests which are idempotent.)

## Additional context

What do you think about this approach? I'm willing to help with the implementation, I see many ways how the Swift generator could be improved and in my opinion this is the primary feature we miss in the standard generator templates.

DavidHorvath:
resolve #8658

Previously the generated clients didn't support **cancellation of requests**

## I've added request cancellation for:
- Both URLSession & Alamofire library
- Combine configuration
- RxSwift configuration
- Result configuration
- AsyncAwait configuration

For these clients I have updated the RequestBuilder's execute function so it returns an optional `URLSessionDataTask`.
I've also added `@discardableResult` to keep backward compatibility.

### Default
I've added to return the `URLSessionDataTask` so the user will be able to control the request's lifecycle.
``` swift
    @discardableResult
    open class func addPet(body: Pet, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue, completion: @escaping ((_ data: Void?, _ error: Error?) -> Void)) -> URLSessionDataTask? {
        return addPetWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
            switch result {
            case .success:
                completion((), nil)
            case let .failure(error):
                completion(nil, error)
            }
        }
    }
```

### Result
I've added to return the `URLSessionDataTask` so the user will be able to control the request's lifecycle.
``` swift
    @discardableResult
    open class func addPet(body: Pet, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue, completion: @escaping ((_ result: Swift.Result<Void, ErrorResponse>) -> Void)) -> URLSessionTask? {
        return addPetWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
            switch result {
            case .success:
                completion(.success(()))
            case let .failure(error):
                completion(.failure(error))
            }
        }
    }
```

### Combine
I've implemented [handeEvents(receiveCompletion:](https://developer.apple.com/documentation/combine/fail/handleevents(receivesubscription:receiveoutput:receivecompletion:receivecancel:receiverequest:)) where I cancel the returned `URLSessionDataTask` when the stream is cancelled
``` swift
open class func addPet(body: Pet, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) -> AnyPublisher<Void, Error> {
        var dataTask: URLSessionDataTask?
        return Future<Void, Error> { promise in
            dataTask = addPetWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
                switch result {
                case .success:
                    promise(.success(()))
                case let .failure(error):
                    promise(.failure(error))
                }
            }
        }
        .handleEvents(receiveCancel: {
            dataTask?.cancel()
        })
        .eraseToAnyPublisher()
    }
```

### RxSwift
I've added to cancel the returned `URLSessionDataTask` in `Disposables.create` when the stream is cancelled
``` swift
    open class func addPet(body: Pet, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) -> Observable<Void> {
        return Observable.create { observer -> Disposable in
            let dataTask = addPetWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
                switch result {
                case .success:
                    observer.onNext(())
                case let .failure(error):
                    observer.onError(error)
                }
                observer.onCompleted()
            }
            
            return Disposables.create {
                dataTask?.cancel()
            }
        }
    }
```

### Async Await
I've added to cancel the returned `URLSessionDataTask` in `withTaskCancellationHandler.onCancel:` 
``` swift
    open class func addPet(body: Pet, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) async throws {
        var task: URLSessionTask?
        return try await withTaskCancellationHandler {
            try Task.checkCancellation()
            return try await withCheckedThrowingContinuation { continuation in
                guard !Task.isCancelled else {
                  continuation.resume(throwing: CancellationError())
                  return
                }

                task = addPetWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
                    switch result {
                    case .success:
                        continuation.resume(returning: ())
                    case let .failure(error):
                        continuation.resume(throwing: error)
                    }
                }
            }
        } onCancel: { [task] in
            task?.cancel()
        }
    }
```

### PR checklist
 
  ```
  ./mvnw clean package 
  ./bin/generate-samples.sh
  ``` 
  Commit all changed files. 
- [x] If your PR is targeting a particular programming language, @mention the [technical committee](https://github.com/openapitools/openapi-generator/#62---openapi-generator-technical-committee) members, so they are more likely to review the pull request. @4brunu

### DIFF
--- a/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
@@ -56,8 +56,11 @@ import Vapor
             headers[header] = value
         }
     }
-
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) { }
+    {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}
+    @discardableResult{{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}-> URLSessionDataTask? {{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}{
+        {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}return nil{{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}
+    }
 
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} func addHeader(name: String, value: String) -> Self {
         if !value.isEmpty {
--- a/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/APIs.mustache
@@ -56,10 +56,10 @@ import Vapor
             headers[header] = value
         }
     }
-    {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}
-    @discardableResult{{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}-> URLSessionDataTask? {{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}{
-        {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}return nil{{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}{{/useURLSession}}
+
+    @discardableResult
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) -> URLSessionTask? {
+        return nil
     }
 
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}public{{/nonPublicApi}} func addHeader(name: String, value: String) -> Self {
--- a/modules/openapi-generator/src/main/resources/swift5/api.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/api.mustache
@@ -199,22 +199,32 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) async throws{{#returnType}} -> {{{returnType}}}{{/returnType}} {
+        var task: URLSessionTask?
+        return try await withTaskCancellationHandler {
+            try Task.checkCancellation()
+            return try await withCheckedThrowingContinuation { continuation in
+                guard !Task.isCancelled else {
+                  continuation.resume(throwing: CancellationError())
+                  return
+                }
 
-        return try await withCheckedThrowingContinuation { continuation in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
-                switch result {
-                {{#returnType}}
-                case let .success(response):
-                    continuation.resume(returning: response.body!)
-                {{/returnType}}
-                {{^returnType}}
-                case .success:
-                    continuation.resume(returning: ())
-                {{/returnType}}
-                case let .failure(error):
-                    continuation.resume(throwing: error)
+                task = {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+                    switch result {
+                    {{#returnType}}
+                    case let .success(response):
+                        continuation.resume(returning: response.body!)
+                    {{/returnType}}
+                    {{^returnType}}
+                    case .success:
+                        continuation.resume(returning: ())
+                    {{/returnType}}
+                    case let .failure(error):
+                        continuation.resume(throwing: error)
+                    }
                 }
             }
+        } onCancel: { [task] in
+            task?.cancel()
         }
     }
 {{/useAsyncAwait}}
--- a/modules/openapi-generator/src/main/resources/swift5/api.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/api.mustache
@@ -55,9 +55,10 @@ extension {{projectName}}API {
      */
     {{#isDeprecated}}
     @available(*, deprecated, message: "This operation is deprecated.")
-    {{/isDeprecated}}
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, completion: @escaping ((_ data: {{{returnType}}}{{^returnType}}Void{{/returnType}}?, _ error: Error?) -> Void)) {
-        {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+    {{/isDeprecated}}{{^useAlamofire}}
+    @discardableResult{{/useAlamofire}}
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, completion: @escaping ((_ data: {{{returnType}}}{{^returnType}}Void{{/returnType}}?, _ error: Error?) -> Void)){{^useAlamofire}} -> URLSessionDataTask?{{/useAlamofire}} {
+        {{^useAlamofire}}return {{/useAlamofire}}{{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
             switch result {
             {{#returnType}}
             case let .success(response):
@@ -122,8 +123,8 @@ extension {{projectName}}API {
     @available(*, deprecated, message: "This operation is deprecated.")
     {{/isDeprecated}}
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> Observable<{{{returnType}}}{{^returnType}}Void{{/returnType}}> {
-        return Observable.create { observer -> Disposable in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+        {{#useURLSession}}return {{/useURLSession}}Observable.create { observer -> Disposable in
+            {{#useURLSession}}let dataTask = {{/useURLSession}}{{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
                 switch result {
                 {{#returnType}}
                 case let .success(response):
@@ -138,7 +139,10 @@ extension {{projectName}}API {
                 }
                 observer.onCompleted()
             }
-            return Disposables.create()
+            
+            return Disposables.create{{^useURLSession}}(){{/useURLSession}}{{#useURLSession}} {
+                dataTask?.cancel()
+            }{{/useURLSession}}
         }
     }
 {{/useRxSwift}}
@@ -157,8 +161,9 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> AnyPublisher<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> {
+        {{#useURLSession}}var dataTask: URLSessionDataTask?{{/useURLSession}}
         return Future<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> { promise in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+            {{#useURLSession}}dataTask = {{/useURLSession}}{{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
                 switch result {
                 {{#returnType}}
                 case let .success(response):
@@ -172,7 +177,11 @@ extension {{projectName}}API {
                     promise(.failure(error))
                 }
             }
-        }.eraseToAnyPublisher()
+        }{{#useURLSession}}
+        .handleEvents(receiveCancel: {
+            dataTask?.cancel()
+        }){{/useURLSession}}
+        .eraseToAnyPublisher()
     }
     #endif
 {{/useCombine}}
--- a/modules/openapi-generator/src/main/resources/swift5/api.mustache
+++ b/modules/openapi-generator/src/main/resources/swift5/api.mustache
@@ -55,10 +55,10 @@ extension {{projectName}}API {
      */
     {{#isDeprecated}}
     @available(*, deprecated, message: "This operation is deprecated.")
-    {{/isDeprecated}}{{^useAlamofire}}
-    @discardableResult{{/useAlamofire}}
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, completion: @escaping ((_ data: {{{returnType}}}{{^returnType}}Void{{/returnType}}?, _ error: Error?) -> Void)){{^useAlamofire}} -> URLSessionDataTask?{{/useAlamofire}} {
-        {{^useAlamofire}}return {{/useAlamofire}}{{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+    {{/isDeprecated}}
+    @discardableResult
+    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, completion: @escaping ((_ data: {{{returnType}}}{{^returnType}}Void{{/returnType}}?, _ error: Error?) -> Void)) -> URLSessionTask? {
+        return {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
             switch result {
             {{#returnType}}
             case let .success(response):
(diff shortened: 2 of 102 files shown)


modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -157,8 +161,9 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> AnyPublisher<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> {
+        {{#useURLSession}}var dataTask: URLSessionDataTask?{{/useURLSession}}
         return Future<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> { promise in
```
DavidHorvath, line 165: @4brunu I have one question about this.
Don't we want to wrap this `Future publisher` into a [Deferred publisher](https://developer.apple.com/documentation/combine/deferred)?
So it would work like Rx's `Single` by having it defer its execution until it receives a subscriber.

4brunu:
Hey @DavidHorvath, this is another great PR, thanks 👍 
Here are some thoughts.

Instead of returning an `URLSessionDataTask` I think we should return a `URLSessionTask` because it's less specialised.

We should also support cancellation in Alamofire, by replace the following code 
```        
if let encoding = encoding {
    let request = makeRequest(manager: manager, method: xMethod, encoding: encoding, headers: headers)
    if let onProgressReady = self.onProgressReady {
        onProgressReady(request.uploadProgress)
    }
    processRequest(request: request, managerId, apiResponseQueue, completion)
    return request.task
}

return nil
```
In the following class
https://github.com/OpenAPITools/openapi-generator/blob/f596b32316aea8486ceaa822647185621fae652a/modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache#L138-L144

Those two changes should a lot of conditions in the templates.

We should also support request cancelation in `result` by returning the `URLSessionTask` and `async/await` by using the `Task` but we need to explore this a bit more, because I don't have much experience with `async/await`.
https://www.hackingwithswift.com/quick-start/concurrency/how-to-cancel-a-task
https://desiatov.com/swift-structured-concurrency-introduction/
https://betterprogramming.pub/swiftui-concurrency-understand-cooperative-task-cancellation-b71cc69483e0

modules/openapi-generator/src/main/resources/swift5/api.mustache
4brunu: I don't have much experience with `RxSwift` or `Combine`.
Is the current `RxSwift` integration lazy (defer its execution until it receives a subscriber)?

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
```diff
@@ -56,8 +56,11 @@ import Vapor
             headers[header] = value
         }
     }
-
-    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) { }
+    {{#useURLSession}}{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}
```
4brunu, line 59: @DavidHorvath do we need those conditions?
I don't we should return the `URLSessionDataTask` in every configuration, and then in the `api.mustache`we add the conditions, case by case.

4brunu:
Looks like the way to cancel an async/await function is https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#cancellation-handlers

Here is the example in the like above.

```
func download(url: URL) async throws -> Data? {
  var urlSessionTask: URLSessionTask?

  return try withTaskCancellationHandler {
    return try await withUnsafeThrowingContinuation { continuation in
      urlSessionTask = URLSession.shared.dataTask(with: url) { data, _, error in
        if let error = error {
          // Ideally translate NSURLErrorCancelled to CancellationError here
          continuation.resume(throwing: error)
        } else {
          continuation.resume(returning: data)
        }
      }
      urlSessionTask?.resume()
    }
  } onCancel: {
    urlSessionTask?.cancel() // runs immediately when cancelled
  }
}
```

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: @4brunu Yes. RxSwift Observable's only call when you subscribe to the stream.
In Combine publishers are working the same way except Future. We can only achieve the same if we would wrap Future in Deferred.

So currently if we just write this:
``` swift
let newPet = Pet(id:<id>, category: <category>, name: <name>, photoUrls: [])

let addPetPublisher = PetAPI.addPet(body: newPet)
```
The request will be called without subscribing to the stream with `.sink`

4brunu: This is the way it is, because when I introduced support for `Combine` I didn't have much experience with it, and Future looked like a good fit.
I think we can also make `Combine` lazy, because that would make the API more flexible, by only starting the request when the user want's.
What do you think?

modules/openapi-generator/src/main/resources/swift5/APIs.mustache
DavidHorvath: I've only added this condition because I only knew how I need to cancel request for URLSession with Default, Combine or RxSwift configurations and I didn't want to update the others 🙂 

But with your comments and links above, I have a direction to do it.
So if I update the other configurations as well, we won't need this

4brunu: Sure, and you did great, because honestly in the past I looked at this big task and felt a bit discouraged.
If you need any help please let me know 👍

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: Yes, I think we should do that 👍

4brunu: Can this behaviour change create any bugs for the users that already implement Combine in their projects?
For example a network request that never happens because of this change?

DavidHorvath: Unfortunately, yes 😕

DavidHorvath: If somebody used it without subscribing to the stream, it won't be called

4brunu: Hum, let's keep that for a different PR then, and keep this one with just the cancellation, to avoid any friction to merge it.

DavidHorvath:
> Looks like the way to cancel an async/await function is https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#cancellation-handlers
> 
> Here is the example in the like above.
> 
> ```
> func download(url: URL) async throws -> Data? {
>   var urlSessionTask: URLSessionTask?
> 
>   return try withTaskCancellationHandler {
>     return try await withUnsafeThrowingContinuation { continuation in
>       urlSessionTask = URLSession.shared.dataTask(with: url) { data, _, error in
>         if let error = error {
>           // Ideally translate NSURLErrorCancelled to CancellationError here
>           continuation.resume(throwing: error)
>         } else {
>           continuation.resume(returning: data)
>         }
>       }
>       urlSessionTask?.resume()
>     }
>   } onCancel: {
>     urlSessionTask?.cancel() // runs immediately when cancelled
>   }
> }
> ```

@4brunu Unfortunately this is not working the way it suppose to 😕 
I get a `Reference to captured var 'urlSessionTask' in concurrently-executing code` compile error.
https://forums.swift.org/t/how-to-cancel-a-publisher-when-using-withtaskcancellationhandler/49688

DavidHorvath:
@4brunu
I've added a new commit where I 
- Changed URLSessionDataTasks to URLSessionTasks
- Added support for request cancellation for alamofire library based on your comments
- Return the task in Result configuration

Still missing:
- Async/Await (I commented above that unfortunately that solution isn't working, I need to investigate further)

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -190,8 +199,9 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) async throws{{#returnType}} -> {{{returnType}}}{{/returnType}} {
+
         return try await withCheckedThrowingContinuation { continuation in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).executapiResponseQueue) { result in
```
4brunu, line 204: @DavidHorvath I think this change was not intentional

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -170,11 +171,13 @@ private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
             credentialStore[dataTask.taskIdentifier] = credential
 
             dataTask.resume()
-
+{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}
```
4brunu, line 174: @DavidHorvath I think this logic should be removed from here, and be handled in the `api.mustache`, as already is

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
```diff
@@ -170,11 +171,13 @@ private var credentialStore = SynchronizedDictionary<Int, URLCredential>()
...
+            return dataTask{{/useAsyncAwait}}{{/useResult}}{{/usePromiseKit}}{{/useVapor}}
         } catch {
             apiResponseQueue.async {
                 completion(.failure(ErrorResponse.error(415, nil, nil, error)))
-            }
+            }{{^useVapor}}{{^usePromiseKit}}{{^useResult}}{{^useAsyncAwait}}
```
4brunu, line 179: @DavidHorvath I think this logic should be removed from here, and be handled in the api.mustache, as already is

DavidHorvath: Yeah we don't need that, sorry, I left that in... I'll update in a minute

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
DavidHorvath: Yeah we don't need that, sorry, I left that in... I'll update in a minute

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: Oh... yeah it was unintentional

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
4brunu: No problem 🙂 Thanks for your help

DavidHorvath: Updated

modules/openapi-generator/src/main/resources/swift5/libraries/urlsession/URLSessionImplementations.mustache
DavidHorvath: Updated

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: Updated

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -157,8 +161,9 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> AnyPublisher<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> {
+        var task: URLSessionTask?
         return Future<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> { promise in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+            {{#useURLSession}}dataTask = {{/useURLSession}}{{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
```
4brunu, line 166: @DavidHorvath I think here we can remove the checks on the URLSession, because now is alamofire also supports cancellation

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -172,7 +177,11 @@ extension {{projectName}}API {
                     promise(.failure(error))
                 }
             }
-        }.eraseToAnyPublisher()
+        }{{#useURLSession}}
```
4brunu, line 180: @DavidHorvath I think here we can remove the checks on the URLSession, because now is alamofire also supports cancellation

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -122,8 +123,8 @@ extension {{projectName}}API {
     @available(*, deprecated, message: "This operation is deprecated.")
     {{/isDeprecated}}
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> Observable<{{{returnType}}}{{^returnType}}Void{{/returnType}}> {
-        return Observable.create { observer -> Disposable in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+        {{#useURLSession}}return {{/useURLSession}}Observable.create { observer -> Disposable in
```
4brunu, line 126: @DavidHorvath I think here we can remove the checks on the URLSession, because now is alamofire also supports cancellation

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -138,7 +139,10 @@ extension {{projectName}}API {
                 }
                 observer.onCompleted()
             }
-            return Disposables.create()
+            
+            return Disposables.create{{^useURLSession}}(){{/useURLSession}}{{#useURLSession}} {
```
4brunu, line 143: @DavidHorvath I think here we can remove the checks on the URLSession, because now is alamofire also supports cancellation

DavidHorvath: Oh I was neglectful... Sorry about this and thanks for your comments!

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: fixed

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: fixed

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: fixed

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: fixed

4brunu:
@DavidHorvath I think I got `async/await` working with cancellation
```
    @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
    open class func fakeOuterBooleanSerialize1(body: Bool? = nil, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) async throws -> Bool {
        var task: URLSessionTask?
        return try await withTaskCancellationHandler {
            try await withCheckedThrowingContinuation { continuation in
                task = fakeOuterBooleanSerializeWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
                    switch result {
                    case let .success(response):
                        continuation.resume(returning: response.body!)
                    case let .failure(error):
                        continuation.resume(throwing: error)
                    }
                }
            }
        } onCancel: { [task] in
            task?.cancel()
        }
    }
```

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -190,6 +199,7 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) async throws{{#returnType}} -> {{{returnType}}}{{/returnType}} {
+
         return try await withCheckedThrowingContinuation { continuation in
```
4brunu, line 203: @DavidHorvath  Can you please try this?
```
    {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) async throws{{#returnType}} -> {{{returnType}}}{{/returnType}} {
        var task: URLSessionTask?
        return try await withTaskCancellationHandler {
            try await withCheckedThrowingContinuation { continuation in
                task = {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
                    switch result {
                    {{#returnType}}
                    case let .success(response):
                        continuation.resume(returning: response.body!)
                    {{/returnType}}
                    {{^returnType}}
                    case .success:
                        continuation.resume(returning: ())
                    {{/returnType}}
                    case let .failure(error):
                        continuation.resume(throwing: error)
                    }
                }
            }
        } onCancel: { [task] in
            task?.cancel()
        }
    }
```

4brunu: I hope I got it right, because I didn't tested it, but I tested the swift code that I send I minute ago

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -157,8 +161,9 @@ extension {{projectName}}API {
     {{/isDeprecated}}
     @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
     {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue) -> AnyPublisher<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> {
+        var task: URLSessionTask?
         return Future<{{{returnType}}}{{^returnType}}Void{{/returnType}}, Error> { promise in
-            {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
+            dataTask = {{operationId}}WithRequestBuilder({{#allParams}}{{paramName}}: {{paramName}}{{^-last}}, {{/-last}}{{/allParams}}).execute(apiResponseQueue) { result in
```
4brunu, line 166: @DavidHorvath this `dataTask` doesn't match with `task` variable bellow, could you please rename this to `task`?

modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache
```diff
@@ -78,7 +78,8 @@ private var managerStore = SynchronizedDictionary<String, Alamofire.Session>()
         return manager.request(URLString, method: method, parameters: parameters, encoding: encoding, headers: HTTPHeaders(headers))
     }
 
-    override {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) {
+    @discardableResult
+    override {{#nonPublicApi}}internal{{/nonPublicApi}}{{^nonPublicApi}}open{{/nonPublicApi}} func execute(_ apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, _ completion: @escaping (_ result: Swift.Result<Response<T>, ErrorResponse>) -> Void) -> URLSessionTask {
```
4brunu, line 82: @DavidHorvath here the return type should be `URLSessionTask?`

modules/openapi-generator/src/main/resources/swift5/api.mustache
```diff
@@ -220,8 +230,9 @@ extension {{projectName}}API {
     {{#isDeprecated}}
     @available(*, deprecated, message: "This operation is deprecated.")
     {{/isDeprecated}}
+    @discardableResult
     open class func {{operationId}}({{#allParams}}{{paramName}}: {{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{{datatypeWithEnum}}}_{{operationId}}{{/isContainer}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{/isEnum}}{{^required}}? = nil{{/required}}{{^-last}}, {{/-last}}{{/allParams}}{{#hasParams}}, {{/hasParams}}apiResponseQueue: DispatchQueue = {{projectName}}API.apiResponseQueue, completion: @escaping ((_ result: Swift.Result<{{{returnType}}}{{^returnType}}Void{{/returnType}}, ErrorResponse>) -> Void)) {
```
4brunu, line 234: @DavidHorvath here we are missing the return type `URLSessionTask?`

modules/openapi-generator/src/main/resources/swift5/libraries/alamofire/AlamofireImplementations.mustache
DavidHorvath: Fixed

modules/openapi-generator/src/main/resources/swift5/api.mustache
DavidHorvath: It wasn't very straightforward at first how to do it but finally I was able to test our generated async await APIClient.
I've displayed the response on the UI or a `Cancelled` text if the task had been canceled.

``` swift
    func load() async throws -> String {
        let fetchTask = Task { () -> String in
            do {
                let response = try await TestAPI.asyncTest()
                if Task.isCancelled { return "Cancelled" }

                return response
            } catch {
                return "Error"
            }
        }

        fetchTask.cancel()

        return await fetchTask.value
    }

    open class func asyncTest(apiResponseQueue: DispatchQueue = TestAPI.apiResponseQueue) async throws -> String {
        var task: URLSessionTask?
        return try await withTaskCancellationHandler {
            try await withCheckedThrowingContinuation { continuation in
                task = testWithRequestBuilder().execute(apiResponseQueue) { result in
                    switch result {
                    case let .success(response):
                        continuation.resume(returning: response.body!)
                    case let .failure(error):
                        continuation.resume(throwing: error)
                    }
                }
            }
        } onCancel: { [task] in
            print("Task Cancelled")
            task?.cancel()
        }
    }

```

And there was a `Task Cancelled` console message and a `Cancelled` text on the UI as well, so it works! 🙂 🚀

modules/openapi-generator/src/main/resources/swift5/api.mustache
4brunu: @DavidHorvath can you please also fix this one please?

modules/openapi-generator/src/main/resources/swift5/api.mustache
4brunu: Nice, good to know 👍

4brunu:
@DavidHorvath I don't want to be annoying, but I got one last improvement that it would be nice, but if you prefer I can do it in a separated PR.
In the async/await we could check for cancellation, like so:
```
    @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
    open class func call123testSpecialTags(body: Client, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) async throws -> Client {
        var task: URLSessionTask?
        return try await withTaskCancellationHandler {
            try Task.checkCancellation() // HERE
            return try await withCheckedThrowingContinuation { continuation in // also add return in the beginning of the line
                guard !Task.isCancelled else { // and HERE
                  continuation.resume(throwing: CancellationError())
                  return
                }
                task = call123testSpecialTagsWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
                    switch result {
                    case let .success(response):
                        continuation.resume(returning: response.body!)
                    case let .failure(error):
                        continuation.resume(throwing: error)
                    }
                }
            }
        } onCancel: { [task] in
            task?.cancel()
        }
    }
```
What do you think?
I got this idea from here
https://www.pointfree.co/episodes/ep154-async-refreshable-composable-architecture (in the first exercise that use withTaskCancellationHandler)

DavidHorvath:
> @DavidHorvath I don't want to be annoying, but I got one last improvement that it would be nice, but if you prefer I can do it in a separated PR. In the async/await we could check for cancellation, like so:
> 
> ```
>     @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
>     open class func call123testSpecialTags(body: Client, apiResponseQueue: DispatchQueue = PetstoreClientAPI.apiResponseQueue) async throws -> Client {
>         var task: URLSessionTask?
>         return try await withTaskCancellationHandler {
>             try Task.checkCancellation() // HERE
>             return try await withCheckedThrowingContinuation { continuation in // also add return in the beginning of the line
>                 guard !Task.isCancelled else { // and HERE
>                   continuation.resume(throwing: CancellationError())
>                   return
>                 }
>                 task = call123testSpecialTagsWithRequestBuilder(body: body).execute(apiResponseQueue) { result in
>                     switch result {
>                     case let .success(response):
>                         continuation.resume(returning: response.body!)
>                     case let .failure(error):
>                         continuation.resume(throwing: error)
>                     }
>                 }
>             }
>         } onCancel: { [task] in
>             task?.cancel()
>         }
>     }
> ```
> 
> What do you think? I got this idea from here https://www.pointfree.co/episodes/ep154-async-refreshable-composable-architecture (in the first exercise that use withTaskCancellationHandler)

@4brunu Sure I can add that as well

4brunu (approved):
Looks good to me 👍 
@DavidHorvath Thank you very much for the two awesome PRs, that were very impactful in the Swift Generator, and for your quality PRs.
Thanks you also for your patience and for doing the required changes to merge the PRs.
Are you planning to do open more PRs? 😃 
Feel free to contribute any time you want 👍

PR #10913, opened 2021-11-21
Add option to skip case conversion for api operation names

matthewjchou:
Added a flag to skip any case conversion for operationId
ie. operationId: 'addPet' does not change to 'add_pet' when using a python generator

Added test cases for checking that the flag sets the correct variables in WorkFlowSettings/CodegenConfig
Added test case for checking that the operationId is not modified when the flag is set in the PythonClientTest test file

### DIFF
--- a/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
+++ b/modules/openapi-generator-cli/src/main/java/org/openapitools/codegen/cmd/Generate.java
@@ -227,6 +227,11 @@ public class Generate extends OpenApiGeneratorCommand {
             description = "Skips the default behavior of validating an input specification.")
     private Boolean skipValidateSpec;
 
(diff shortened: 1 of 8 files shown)


wing328:
> ie. operationId: 'addPet' does not change to 'add_pet' when using a python generator

`addPet` for example doesn't conform to the style guide: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

matthewjchou:
> > ie. operationId: 'addPet' does not change to 'add_pet' when using a python generator
> 
> `addPet` for example doesn't conform to the style guide: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

I know, it was a request in an issue (with multiple +1's) asking for the feature to bypass automatic style-guide conformity

wing328:
sorry I don't think we can accept this change based on the reason mentioned above.

